PR #2, opened 2015-02-09
Added ASL licence headers to files

paliwalashish:
Added the Licence header to most of the files. Some config and data files remain. Also need to add rat check in pom.xml to detect file without ASL header, which shall do in separate PR


spmallette:
Thanks for doing the bulk of this work.  Nice to have that bit of administrative work complete.

paliwalashish:
Thanks ! It was a good way of browsing the code. We need namespace migration to com.apache.tinkerpop as well, want me to take care of it?

PR #3, opened 2015-02-10
moved namespace to org.apache

paliwalashish:
Moved file to org.apache namespace, includes doc/test changes. Used IDE to refactoring, and saw files as Delete and Added, so the history might not be maintained. Pls see if it's a concern before committing this.


spmallette:
as you said, it looks like the history isn't maintained as a result of the IDE refactoring.  i think we'd like to try to preserve that if we could.  i don't think we will merge this as it is.  we'd like to see what we can do to protect the history.

paliwalashish:
Yeah sure ! We need to move files manually. Shall create a new PR once done. mv or git mv preserves the history. Shall use them

paliwalashish:
Shall generate a new one

PR #6, opened 2015-02-11
namespace migration to org.apache

paliwalashish:
Moved namespace to org.apache using git mv command to preserve the history. Pls verify and commit.


paliwalashish:
Shall I cancel this request, as there is a similar PR and integrated with Travis build?

spmallette:
thanks for giving this a shot - I think you can close this as we've done the namespace migration in the other PR.

PR #43, opened 2015-03-26
TINKERPOP3-602 Provide osgi metadata for tinkerpop core artifacts.

splatch:


mfranklin:
LGTM, but the branch isn't currently mergeable.

splatch:
I will update this PR next week after getting another - related to maven cleanup. I think in this order it may be better to go.

PR #48, opened 2015-04-07
Make gremlin-groovy a provided dependency of tinkergraph-gremlin

jkschneider:
Not every use of Tinkergraph is Groovy based.

### DIFF
--- a/tinkergraph-gremlin/pom.xml
+++ b/tinkergraph-gremlin/pom.xml
@@ -35,6 +35,7 @@ limitations under the License.
             <groupId>org.apache.tinkerpop</groupId>
             <artifactId>gremlin-groovy</artifactId>
             <version>${project.version}</version>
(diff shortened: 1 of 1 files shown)


spmallette:
Use of `<optional>` is interesting here. `TinkerGraph` only depends on `gremlin-groovy` because of the  plugin infrastructure.  In the case where `TinkerGraph` is used as a plugin, it would be expected that `gremlin-groovy` would be provided by the infrastructure that was hosting it (i.e `gremlin-console` and `gremlin-server`). 

In that sense, would `provided` scope be better for `gremlin-groovy`?

jkschneider:
@spmallette - great point... the PR has been updated.

spmallette:
Thanks @jkschneider - took me a bit longer than expected to merge this, but it's in there now.

PR #53, opened 2015-04-14
TINKERPOP3-632 Enforce gremlin.tests env var in AbstractGremlinSuite

mhfrantz:


mhfrantz:
Hold off on this, as I just saw the request to rename the env var, and I still need to patch the README.  Any feedback on my approach thus far would be appreciated.

mhfrantz:
OK, this is ready for review.

I tried for a bit to include the parameterized tests that do not descend from AbstractGremlinSuite.  However, that proved challenging because the logic for parsing gremlin_tests is part of the gremlin-test module, but we have parameterized unit tests in gremlin-core (which does not depend on gremlin-test).  So, the "run only this test" mode works for tests that derive from gremlin-test harnesses.

mhfrantz:
Renamed env var to uppercase per @dalaro suggestion

PR #55, opened 2015-04-17
TINKERPOP3-639 selectList

mhfrantz:
This PR includes #54 and puts those changes in context.  In particular, the new `Path` API (`getLast` and `getList`) are used to implement `select` and `selectList`.

I'm still open to renaming `selectList` to `selectPath`.

The unit tests in `SelectListTest` are based on the ones in `SelectTest`.

I added a couple of tests to `SelectTest` because there weren't any tests that demonstrated the list result.


mhfrantz:
Replaced by #56

PR #56, opened 2015-04-22
TINKERPOP3-639 selectList (v4)

mhfrantz:
Replaces PR #55.  Since that PR, integration tests now pass.  One bug (in computing the traverser requirements) was fixed, and some tests were disabled for `COMPUTER` mode.

PR #57, opened 2015-04-24
TINKERPOP3-639 selectList (v5)

mhfrantz:
Replaces #56 

Removed `SparsePath` stuff, since that class no longer exists.

Removed `PATH_ACCESS` requirement, since that was removed as well.

Please advise on whether we should call this `selectList` or `selectPath` or `selectMulti`.

Please advise on whether the `select` changes should be separated from this PR.


mhfrantz:
Stale

PR #58, opened 2015-05-01
Add `Profileable` interface to enable vendors to receive a `Step's MutableMetrics`.

rjbriody:


mbroecheler:
Wouldn't it be better to return 0 in getCount instead of null? That would make this easier to use and 0 seems a reasonable default.

spmallette:
@rjbriody this didn't force any change to serialization right?  all good there?

rjbriody:
@spmallette correct, no changes to serialization.

rjbriody:
@mbroecheler I debated 0 vs null and decided on 0 for the following reasons: 
* the counts are an arbitrary map and this behavior seems consistent w/ map semantics
* overloading 0 as "really zero" or "not defined" leads to ambiguity

What do you think? I'm not dead set on it.

mbroecheler:
Yes, the only downside to return 0 is that you cannot differentiate between a true 0 and "there isn't a count". However, in most cases where you are counting stuff, you don't need to differentiate that. What would be the use case for "has this count actually been set" in metrics?

rjbriody:
The Traversal Steps populate 2 counts that are expected by the parent TraversalMetrics. These are for the num traversers and num elements emitted. Metrics that correspond to something other than a step (such as will be the case when Titan adds nested Metrics for DB or index access) will not have values for these keys, so saying that their values are 0 would be missleading.

PR #60, opened 2015-05-08
TINKERPOP3-670 tail step (v1)

mhfrantz:
https://issues.apache.org/jira/browse/TINKERPOP3-670

This is implemented as specified in JIRA, with a couple more tests for good measure.

As specified, the behavior for `tail(local, 1)` is special if the input is a `List`, in that it emits the element and not a one-element `List`.  That makes it different from `limit(local, 1)`, but more like `select("x")` (for a single step label.

In the spirit of "containers only when necessary", should `limit(local, 1)` be changed to emit the element rather then a one-element `List`?  I filed https://issues.apache.org/jira/browse/TINKERPOP3-673 to think about that.

I had some OOM problems running the full integration test suite, but all of the new tests are disabled for COMPUTER for various reasons.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
@@ -559,8 +559,16 @@ public interface GraphTraversal<S, E> extends Traversal<S, E> {
         return this.range(scope, 0, limit);
     }
 
+    public default GraphTraversal<S, E> tail() {
+        return this.tail(1);
+    }
+
     public default GraphTraversal<S, E> tail(final long limit) {
         return this.tail(Scope.global, limit);
+    }
+
+    public default GraphTraversal<S, E> tail(final Scope scope) {
+        return this.tail(scope, 1);
     }
 
     public default GraphTraversal<S, E> tail(final Scope scope, final long limit) {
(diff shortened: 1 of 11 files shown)


mhfrantz:
@okram This is one of the things you said I could work on.  Any feedback before I rebase against master?

okram:
Can you demonstrate `TailGlobalStep` working with bulks? For instance, do a `repeat()` to get some bulk going and then make sure your `tail()` returns expected results. As it stands, I don't think you take that into account.

Did you add your `GraphTraversal` methods to `__` (anonymous traversal)?

dkuppitz:
I had a quick look - shouldn't `TailGlobalStep` be a `CollectingBarrier`?

mhfrantz:
As I mention in the ticket, I don't think `CollectingBarrierStep` is optimal because it collects all of the items.  But I will see how it handles bulk, and try to address that in my next PR.

okram:
Yea, I don't think `CollectiveBarrierStep` is right... For bulk, please see how `RangeGlobalStep` handles bulks.

mhfrantz:
I'm looking in more detail at how `RangeGlobalStep` works.  It seems there are a few places where this exact step is a special case (via `instanceof` checks, e.g. in `TraverserMapReduce` as well as in several strategies).

The way that `Bypassing` is used also has to be considered.

I also see that `ComputerVerificationStrategy` looks for specific types of steps, and `TailGlobalStep` might belong there.

It feels like some refactoring is warranted.  `RangeGlobalStep` and `CollectingBarrierStep` each represent some class of steps to which `TailGlobalStep` may or may not belong.  Perhaps `Bypassing` is one such aspect, but are there others?

Maybe `TailGlobalStep` is really a new animal, but it feels like it ought to be composed from pieces that mean "it is like a barrier, but not exactly CollectingBarrierStep" and "it is like a range, but not exactly RangeGlobalStep".

I'm still coming up to speed on the design, and I'm not yet totally lost, but any pointers would be appreciated.  I can appreciate the tension between a fully a generalized "step aspects" API and one that assumes that the number of fundamental steps is small and so will scale with special case logic.

okram:
`RangeGlobalStep` and `TailGlobalStep` could inherit from some abstract class. However, I think getting `TailGlobalStep` built is best first or else the PR will be too big and too hard to review. After that studying the two implementation should yield the generalization. Moreover, both steps are very similar to `DedupStep` as well. They are like "lazy barriers."

`TraverserMapReduce` is responsible for dealing with the end of a traversal. It does any final computations like ranging, sorting, etc. And yes, `TailGlobalStep` would fit there as well. However, I can connect it up to `TraverserMapReduce` -- its not that hard.

And yes -- `ComputerVerificationStrategy` will need to complain if `TailGlobalStep` is used mid-traversal. It can only be used as the end step and thus, its processing relegated to `TraverserMapReduce`.

mhfrantz:
Replaced with PR #61

PR #65, opened 2015-05-16
TINKERPOP3-670 tail step doc

mhfrantz:
Also adds doc for `limit(local)` and `range(local)`, since they are related.

BTW, the documentation tool chain is really nice!  I like being able to write Groovy samples and have them executed during preprocessing.


mhfrantz:
@okram This is the doc for the `tail` step that you requested.

PR #70, opened 2015-06-06
TINKERPOP3-713 Remove dependency on groovy-console which contains creative commons licenses

graben1437:
Updates two pom files - for gremlin-groovy and gremlin-driver to remove the groovy-all and use specific groovy files needed.

Also removed jline dependency from gremlin-groovy - not needed and also caused version conflicts.


graben1437:
fixing pull request for TINKERPOP3-713 to include just the files for that change.  Will close this pull request and submit a new one.

PR #81, opened 2015-06-15
TINKERPOP3-727 doc update for __.in() and __.as()

pluradj:


dkuppitz:
Note, that this is only an *issue* in Groovy (and other languages that have `as` and `in` as a keyword). In Java it just works:

```java
g.V().repeat(in()).times(2).match("a",
    as("a").in("knows").as("b"),
    as("b").out("created")).select("b")
```

Whether you can use the shortcut or not, always depends on the language that you're using.

pluradj:
Good point, Daniel. I'll make additional updates to the wording.

PR #91, opened 2015-07-22
TINKERPOP3-779: RFC for coalesce retain path

mhfrantz:
I tried to make some progress on this.  First, I added unit test to reproduce the bug.

I then changed the child traversals from local to global, based on the fact that BranchStep shows its branch traversals as global.  I'm not entirely sure what this does, and it had no effect on the test.

Next, I tried to reimplement CoalesceStep using ComputerAwareStep (instead of FlatMapStep).  This seems necessary, since the child traversals can do whatever they want.  My test now passes in "standard" mode, but fails in "computer" mode.

In "computer" mode, I'm not sure how to implement the sequential "first successful child" logic.  What's in there now is "follow all paths" logic.  If it appears that I am on the right track, I can work more on this.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/CoalesceStep.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/CoalesceStep.java
@@ -60,7 +60,7 @@ public final class CoalesceStep<S, E> extends FlatMapStep<S, E> implements Trave
     }
 
     @Override
-    public List<Traversal.Admin<S, E>> getLocalChildren() {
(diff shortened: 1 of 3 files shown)


okram:
The problem with OLAP (global children) is that you can't do stuff like (as you say) "first successful child" as everything is executing in parallel and you don't know for which input what child output you are referring too. I suspect you could get fancy with path analysis, but then bulking goes away.

We really need to make it so that global and local children behave "the same" as there are other steps that currently can't work in OLAP (e.g. `intersect()`, etc.).

okram:
If this is not happening, please close the PR.

mhfrantz:
This PR is apparently not how to accomplish the goal.

PR #94, opened 2015-07-31
RFC: Add "For Committers" documentation

mhfrantz:


okram:
@spmallette -- if you are happy with this, can you merge? Also, can you remove "cool-new-feature" :) as there is nothing hip or cool about what we do (its computer science).

PR #98, opened 2015-09-30
Made correction to fix TINKERPOP3-855.

PommeVerte:
This fixes the issue with Json submitted sasl arguments not being in `byte[]` format.
It is not BC breaking. Tests seem to pass though I wouldn't mind confirmation here.

I also added changes to the documentation to reflect the new functionality.


PommeVerte:
It's worth going through the issue as I've made an arbitrary decision on the JSON input type and this needs to be reviewed.

PR #100, opened 2015-10-02
Changed TinkerGraphSerializer to reuse existing Kryo instance

christianbellinaef:
Let the tinker graph serializer reuse an existing kryo instance instead of creating a new one.

Tests assert that the same kryo instance is being used inside the Gryo reader/writer classes.

https://groups.google.com/forum/#!topic/gremlin-users/apFfQW3DkK4


spmallette:
thanks for doing this - about to merge to master.

PR #101, opened 2015-10-03
TINKERPOP3-850: Reduce Graph.addVertex overload ambiguity

mhfrantz:
Since this is a breaking change for vendors, do we want to merge this into tp30 or master?

Also, how can we verify that it addresses the initial problem with Scala and overloading?


mhfrantz:
Replaced by #102 on master, since breaking changes should not go into 3.0.2.

PR #102, opened 2015-10-06
TINKERPOP3-850 Reduce Graph.addVertex overload ambiguity (v1)

mhfrantz:
https://issues.apache.org/jira/browse/TINKERPOP3-850


mhfrantz:
Replaced by #104

okram:
Why is this method needed?

https://github.com/RedSeal-co/incubator-tinkerpop/blob/chore/TINKERPOP3-850-Reduce-Graph-addVertex-overload-ambiguity_104440766.v2/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/Graph.java#L145

mhfrantz:
@okram Because the variadic overload confuses Scala in the 1-arg case, but in many places, we built an `Object[]`, so we still need that capability.

PR #106, opened 2015-10-13
TINKERPOP3-843

okram:
Test cases passed locally.

https://issues.apache.org/jira/browse/TINKERPOP3-843


okram:
I VOTE +1. I ran the test suite and checked Spark/Giraph integration tests just for a "few tests" to make sure it was all connected correctly.

spmallette:
+1 - standard test suite is happy and did a manual test (the message i was looking for is present):

```text
gremlin> graph = GraphFactory.open('conf/hadoop/hadoop-gryo.properties')
==>hadoopgraph[gryoinputformat->gryooutputformat]
gremlin> g = graph.traversal(computer(SparkGraphComputer))
==>graphtraversalsource[hadoopgraph[gryoinputformat->gryooutputformat], sparkgraphcomputer]
gremlin> g.V().count()
...
WARN  org.apache.tinkerpop.gremlin.spark.process.computer.SparkGraphComputer  - /home/graphie/titan09/titan/lib/hadoop-gremlin-3.0.1-SNAPSHOT.jar does not reference a valid directory -- proceeding regardless
==>6
```

spark-gremlin/src/test/java/org/apache/tinkerpop/gremlin/spark/process/computer/SparkHadoopGraphProvider.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+@GraphProvider.Descriptor(computer = SparkGraphComputer.class)
+public final class SparkHadoopGraphProvider extends HadoopGraphProvider {
+
+    @Override
+    public Map<String, Object> getBaseConfiguration(final String graphName, final Class<?> test, final String testMethodName, final LoadGraphWith.GraphData loadGraphWith) {
+        final Map<String, Object> config = super.getBaseConfiguration(graphName, test, testMethodName, loadGraphWith);
+        config.put("mapreduce.job.reduces", 4);
+        /// spark configuration
+        config.put("spark.master", "local[4]");
```
dkuppitz: I think `local[*]` would make a better default value (`[*]` will use as many CPUs as available, `[4]` will always use 4, regardless of how many CPUs are actually available).

okram: This is a test case and I want things to run fast, but still test threading. Dedicating to many threads is an overdose.

Note that this ticket has nothing to do with this. You see this as "new" cause I renamed the test case class. It has always been `local[4]`. You can make a new ticket if you think that we should change this.

dkuppitz:
+1, after inspecting the fraction of the committed code that was actually relevant to resolve the ticket.

Congrats, that's the 3rd +1, which means you can finally merge.

PR #107, opened 2015-10-13
BulkLoaderVertexProgram for generalized batch loading across graphs

dkuppitz:
`BulkLoaderVertexProgram` now supports

* several inputs (`HadoopGraph` with all its supported `InputFormat`s and any TinkerPop3 enabled graph, like TinkerGraph or Neo4j)
* any TinkerPop3 enabled graph, even if it only supports single threaded access

Samples are provided in the docs. Tests were added to ensure proper functionality of `BulkLoaderVertexProgram`s configuration options.

### DIFF
--- a/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/computer/bulkloading/BulkLoaderVertexProgramTest.java
+++ b/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/computer/bulkloading/BulkLoaderVertexProgramTest.java
@@ -24,11 +24,15 @@ import org.apache.tinkerpop.gremlin.LoadGraphWith;
 import org.apache.tinkerpop.gremlin.process.AbstractGremlinProcessTest;
 import org.apache.tinkerpop.gremlin.structure.Graph;
 import org.apache.tinkerpop.gremlin.structure.util.GraphFactory;
+import org.apache.tinkerpop.gremlin.util.iterator.IteratorUtils;
+import org.junit.After;
 import org.junit.Test;
 
+import java.io.File;
 import java.lang.reflect.Field;
 
 import static org.apache.tinkerpop.gremlin.LoadGraphWith.GraphData.MODERN;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
@@ -37,16 +41,30 @@ import static org.junit.Assert.assertTrue;
  */
 public class BulkLoaderVertexProgramTest extends AbstractGremlinProcessTest {
 
+    final static String TINKERGRAPH_LOCATION = "/tmp/tinkertest.kryo";
+
     private BulkLoader getBulkLoader(final BulkLoaderVertexProgram blvp) throws Exception {
         final Field field = BulkLoaderVertexProgram.class.getDeclaredField("bulkLoader");
         field.setAccessible(true);
         return (BulkLoader) field.get(blvp);
     }
 
-    private Graph getTargetGraph() {
+    private Configuration getWriteGraphConfiguration() {
         final Configuration configuration = new BaseConfiguration();
         configuration.setProperty(Graph.GRAPH, "org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerGraph");
-        return GraphFactory.open(configuration);
+        configuration.setProperty("gremlin.tinkergraph.graphLocation", TINKERGRAPH_LOCATION);
+        configuration.setProperty("gremlin.tinkergraph.graphFormat", "gryo");
+        return configuration;
+    }
+
+    private Graph getWriteGraph() {
+        return GraphFactory.open(getWriteGraphConfiguration());
+    }
+
+    @After
+    public void cleanup() {
+        final File graph = new File(TINKERGRAPH_LOCATION);
+        assertTrue(!graph.exists() || graph.delete());
     }
 
     @Test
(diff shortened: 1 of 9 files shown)


okram:
Why are these needed in TinkerGraph's pom.xml?

```
+        <dependency>
+            <groupId>org.neo4j</groupId>
+            <artifactId>neo4j-tinkerpop-api</artifactId>
+            <version>0.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.neo4j</groupId>
+            <artifactId>neo4j-tinkerpop-api-impl</artifactId>
+            <version>0.1-2.2</version>
+            <scope>test</scope>
+        </dependency>
```

dkuppitz:
Good catch! I actually had a TODO item to remove those dependencies, but ... wanted to see if RTC works out :).

okram:
I VOTE +1 -- assuming @dkuppitz has the docs building fully and integration tests passing.

dkuppitz:
FYI: Yes, integration tests pass and docs are looking good.

dkuppitz:
After learning that I can vote on my own PR and doing a lot more tests than anybody else is likely going to do...

Vote: +1

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/computer/bulkloading/BulkLoaderVertexProgramTest.java
```diff
@@ -60,23 +78,43 @@ public void shouldUseIncrementalBulkLoaderByDefault() throws Exception {
...
         target.vertices().forEachRemaining(v -> assertFalse(v.property(loader.getVertexIdProperty()).isPresent()));
     }
 
-    // TODO: once Neo4j supports concurrent connections, write a real integration test that leverages BLVP
-    // TODO: also, once Neo4j can be used, remove the tinkergraph-gremlin dependency from hadoop-gremlin and clean up the existing tests
+    @Test
+    @LoadGraphWith(MODERN)
+    public void shouldOverwriteExistingElements() throws Exception {
+        final BulkLoaderVertexProgram blvp = BulkLoaderVertexProgram.build()
+                .userSuppliedIds(true)
```
spmallette: @dkuppitz can you explain why this test passes when Neo4j is the graph you are writing to?  Your documentation says:

> |`userSuppliedIds(boolean)` |Whether to use the id's from the source graph as id's in the target graph. If set to `true`, `vertexIdProperty` will be ignored. Note, that the target graph must support user supplied identifiers. | `false`

but Neo4j doesn't support user supplied ids.  Seems like you should have a:

```java
@FeatureRequirement(featureClass = Graph.Features.EdgeFeatures.class, feature = Graph.Features.EdgeFeatures.FEATURE_USER_SUPPLIED_IDS)
```

Am I missing something obvious here?

dkuppitz:
> Am I missing something obvious here?

Probably. The test uses `TinkerGraph`.

spmallette:
Oh - so in the case i was specifying, it reads from Neo4j and writes to TinkerGraph.  Now I get it - it was something obvious.  

Looking at the asserts in the test - they largely just assert counts.  Is it possible to use:

https://github.com/apache/incubator-tinkerpop/blob/master/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/structure/io/IoTest.java#L643

to assert the entire Modern graph rather than just the counts. Note that there's some options there to ignore the ids and stuff.  If you could do that, then that should validate the work you have in `IncrementalBulkLoader` which I don't think has tests at all.  

I also think adding a test of "The Crew" is worthwhile to assure the meta/multi property stuff is solid.

dkuppitz:
Nice idea. The problem is, that `assertToyGraph` also checks edge ids, which will never be preserved by BLVP. I can add some more "manual" assertions though and especially a test case for the Crew graph.

spmallette:
i'm open to you editing that "assert" to allow for your case where ids are ignored.  As long a the existing test cases still assert the ids, i'm cool with whatever you have to change there.  That assert function is as complete as you can get and would be fairly bullet proof for what you're doing.  If you don't want to touch it, that's fine, but might be less work that new "reasonable" manual assertions.

dkuppitz:
Wow, multi-valued properties really suck. I added a test case for the Crew graph, only to realize that multi-valued properties aren't handled properly by BLVP. I thought it's easy to fix, but it's not.

The default implementation of `IncrementalBulkLoader` doesn't know anything about a schema. Now, if it's about to write a property, how should it know, whether it should overwrite an existing value or simply append the new value? I don't think there's a general solution and multi-valued properties need to be handled domain specific.

`IncrementalBulkLoader` should always assume single-valued properties.

Thoughts?

spmallette:
Here's the pattern used in `Attachable`:

https://github.com/apache/incubator-tinkerpop/blob/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/Attachable.java#L333-L335

basically, you can query the graph for the cardinality of the property:

```java
graph.features().vertex().getCardinality(k)
```

I have no idea what that check will do to the speed of your load, but perhaps we worry about optimizing later (unless you see an easy way to avoid lots of repeated checks to this method that we get stuck with in `Attachable`).

okram:
Eek. That sucks. Well, each `VertexProperty` has a unique `id`. Thus, can't you treat it like an edge? You can have multiple edges with the same label between two vertices.... How do you solve that?

dkuppitz:
`IncrementalBulkLoader` supports only one edge with the same label between 2 particular vertices. It's actually the same problem - if the behavior is not sufficient, you'll need to provide a domain specific `BulkLoader` implementation.

I can add that to the docs, but we need a default behavior for multi-valued properties. I still tend to support only single-valued properties.

Regarding @spmallette's suggestion: As you can see in `TheCrew`, cardinalitites can be mixed; `name` is single-valued, `location` is multi-valued.

okram:
@dkuppitz -- I think the default behavior should be multi-property. If the original dataset has two of the same property, well, add the second... Or am I missing something?

dkuppitz:
Imagine a partially loaded `TheCrew` graph where some of the locations are missing and some names are wrong. We know that names should be replaced as they are single valued (or maybe we decided to add middle names, who knows), but locations should be appended (if they don't exist yet; note that this is the distinction between `list` and `set`).

`IncrementalBulkLoader` can only replace OR append, it can't make a decision based on the property name (or anything else). Incremental loading over `TheCrew` graph requires domain specific knowledge.

okram:
Huh. Is it possible for someone to do a builder to provide "schema" ? Not that this should be in bulk loader vertex program, but in like `Schema` object. dah -- @pietermartin ...

dkuppitz:
Well, we talked about a rule based BL, that would solve the problem. However, that's something to consider for 3.1.0, not 3.0.2.

okram:
Okay. Well perhaps we have the "easy version" now. And hopefully if we do something like `Schema` in the future, we can do something like:

```
blvp.writeGraph(conf).schema(Schema.build().....create())
```

If you don't provide a schema, then the default behavior is what you (@dkuppitz) think is reasonable.

spmallette:
i'm ok with this working as-is (replace OR append) for 3.0.2 with the idea that it will be improved for 3.1.0.  I'm just glad to know the limitation before we hung it out there in release so it can be documented clearly.  So based on that are there any other commits to look at?  are you amending the test changes based on what we talked about earlier?

okram:
@dkuppitz --- two things to be aware of (though you might already, just to be clear).

1. Vertex properties and edges have unique ids.
2. `Graph.features().getCardinality()`

spmallette:
VOTE: +1 - that's 3 +1s - ready to merge.

PR #108, opened 2015-10-15
Enhance match()-step and where()-step documentation

jbmusso:
Result from the Gremlin console for the added `match()`-step query:
```groovy
:> g.V().match(__.as('creators').out('created').has('name', 'lop').as('projects'),__.as('projects').in('created').has('age', 29).as('cocreators')).select('creators','cocreators').by('name')
==>{creators=marko, cocreators=marko}
==>{creators=josh, cocreators=marko}
==>{creators=peter, cocreators=marko}
```

### DIFF
--- a/docs/src/the-traversal.asciidoc
+++ b/docs/src/the-traversal.asciidoc
@@ -779,6 +779,20 @@ g.V().match(
       select('a','c').by('name')
 ----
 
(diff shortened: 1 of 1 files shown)


okram:
Given that this is such a minor change, I will merge to `tp30` and declare CTR.

PR #109, opened 2015-10-16
Use Epoll based netty eventloop when possible

anagav:
This patch checks if the the OS is linux and uses netty epoll based event loops which perform better than NIO based event loops.

More details at:
http://netty.io/wiki/native-transports.html

Not sure if you guys have already tried it out. I saw a reduced GC pressure using epoll.

### DIFF
--- a/gremlin-server/pom.xml
+++ b/gremlin-server/pom.xml
@@ -140,6 +140,7 @@ limitations under the License.
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
             </plugin>
+
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-failsafe-plugin</artifactId>
@@ -147,6 +148,29 @@ limitations under the License.
         </plugins>
     </build>
     <profiles>
+        <profile>
+            <id>linux</id>
+            <activation>
+                <os>
+                    <family>Linux</family>
+                </os>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-failsafe-plugin</artifactId>
+                        <configuration>
+                            <systemPropertyVariables>
+                                <gremlin.server.epoll>true</gremlin.server.epoll>
+                            </systemPropertyVariables>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
+
         <profile>
             <!--
               This profile will include neo4j for purposes of transactional testing within Gremlin Server.
--- a/gremlin-server/pom.xml
+++ b/gremlin-server/pom.xml
@@ -151,9 +151,10 @@ limitations under the License.
         <profile>
             <id>linux</id>
             <activation>
-                <os>
-                    <family>Linux</family>
-                </os>
+                <activeByDefault>false</activeByDefault>
+                <property>
+                    <name>EPOLL</name>
+                </property>
             </activation>
             <build>
                 <plugins>
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GremlinServer.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GremlinServer.java
@@ -18,33 +18,28 @@
  */
 package org.apache.tinkerpop.gremlin.server;
 
-import org.apache.tinkerpop.gremlin.server.op.OpLoader;
-import org.apache.tinkerpop.gremlin.server.util.LifeCycleHook;
-import org.apache.tinkerpop.gremlin.server.util.MetricManager;
-import org.apache.tinkerpop.gremlin.server.util.ServerGremlinExecutor;
-import org.apache.tinkerpop.gremlin.server.util.ThreadFactoryUtil;
 import io.netty.bootstrap.ServerBootstrap;
 import io.netty.buffer.PooledByteBufAllocator;
-import io.netty.channel.Channel;
-import io.netty.channel.ChannelFuture;
-import io.netty.channel.ChannelFutureListener;
-import io.netty.channel.ChannelOption;
-import io.netty.channel.EventLoopGroup;
+import io.netty.channel.*;
+import io.netty.channel.epoll.EpollEventLoopGroup;
+import io.netty.channel.epoll.EpollServerSocketChannel;
 import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.channel.socket.nio.NioServerSocketChannel;
 import io.netty.util.concurrent.GenericFutureListener;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 import io.netty.util.internal.logging.Slf4JLoggerFactory;
+import org.apache.commons.lang3.SystemUtils;
+import org.apache.tinkerpop.gremlin.server.op.OpLoader;
+import org.apache.tinkerpop.gremlin.server.util.LifeCycleHook;
+import org.apache.tinkerpop.gremlin.server.util.MetricManager;
+import org.apache.tinkerpop.gremlin.server.util.ServerGremlinExecutor;
+import org.apache.tinkerpop.gremlin.server.util.ThreadFactoryUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.lang.reflect.UndeclaredThrowableException;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 
 /**
  * Start and stop Gremlin Server.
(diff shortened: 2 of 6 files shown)


spmallette:
I did want to try this out.  Nice to know that you saw positive improvement for doing so.  Did you happen to run the  gremlin-server integration tests on your end to make sure everything works nice there?

```text
mvn clean install
cd gremlin-server
mvn verify -DskipIntegrationTests=false
```

spmallette:
I gave this some more thought.  I'm a little concerned that when we do this, tests will be bound to the OS that we're executing on. I don't imagine real problems with that, but ughhh.....if there are it's going to stink.  

I wonder if we shouldn't have a configuration that forces regular java NIO even on LInux.  Then in the tests we have a whole class of tests that run with that setting turned on so that if we happen to be on linux we get coverage there.

Do you have any thoughts on this?

anagav:
@spmallette  I have successfully run the integration test after some fixes :). 

It makes sense to to have separate tests for linux and non-linux env to get coverage of using epoll event loops. Let me work on that.

Also maybe have a travis setup to run tests on multiple platforms.

Also will it be helpful to have travis build for the project? if so I can set it up.

mhfrantz:
@anagav We currently use Apache's Jenkins resources, but we don't have it working exactly the way we want it to.  If you have Travis know-how, then that would work.  I believe there was once a working Travis CI, at least for unit tests.  The goal would be to run integration tests in the CI, too.  If we start doing platform-specific things on purpose (or in case we do so accidentally), then we should cover that with the CI.

spmallette:
> It makes sense to to have separate tests for linux and non-linux env to get coverage of using epoll event loops. Let me work on that.

Do you have a sense of how you want to implement that?  

> Also maybe have a travis setup to run tests on multiple platforms.

There are some issues with using travis in the full way we would like.  I suppose if we restricted travis to just:

```text
mvn clean install
cd gremlin-server
mvn verify -DskipIntegrationTests=false
```

The build would likely not time out.  But we couldn't run the full integration suite over the entire project. I think it takes too long.

> also will it be helpful to have travis build for the project? if so I can set it up.

@anagav thank you for offering. :) 

@mhfrantz think you could work with @anagav on something here? i know you've been thinking about our CI situtation.  I ask because he'll need some help navigating apache infra with whatever we come up with. btw, that discussion should probably happen on the dev mailing list and not here.  @anagav are you on our dev mailing list?  if not, it would be great if you could join:

http://mail-archives.apache.org/mod_mbox/incubator-tinkerpop-dev/

anagav:
Do you have a sense of how you want to implement that?
I am still figuring it out. Do you have any suggestions?

I just subscribed to the mailing list.

spmallette:
> I am still figuring it out. Do you have any suggestions?

I think we should add a configuration to manage epoll or not epoll somehow.  Maybe you add something to `Settings` called `useEpollIfPossible` that is `true` or `false` (default as `false`).  Then in your code, you use that setting on startup to use epoll if on linux (otherwise just use java nio) or if directly set to `false` just use java nio regardless of OS.  

Now for tests, we could add an environment variable, like `GREMLIN_SERVER_EPOLL` which is `true` or `false`.  in `AbstractGremlinServerIntegration` test you check for that variable (defaulted to `false`) in the `overrideSettings` method.  If `true` then flip the `Setting` to `true` and return it.  That will make all the tests work against Gremlin Server configured in "epoll mode".

Lastly, we add a new maven profile to the gremlin-server `pom.xml` that adds your `GREMLIN_SERVER_EPOLL` environment variable to the failsafe plugin config so that the integration tests will get that.  In this way we can optionally run those tests as needed.  I think it is sufficient to not constantly run those tests, but if we decide that is so, we could move it out of a profile and into a second failsafe execution.

This change would also entail an update to the documentation to describe the `epollIfPossible` setting.

What do you think?  Still interested in working on it? Sad that you got sucked into this? :wink: 

> I just subscribed to the mailing list.

cool - welcome.  feel free to start up a thread on the CI issue.  @mhfrantz and i can will join in from there.

anagav:
@spmallette I have already added (first commit :) )  a variable in Settings class which can be overwritten from the config file. 

The setting looks like: 
`
+  public boolean useEpollEventLoop = false;
`

So the decision to use epollevent loop would be `epollSettingEnabled && IS_OS_LINUX` .

Adding a profile based test sounds good to me. I will try to get it done over weekend and hopefully should have something monday.

spmallette:
Sounds like a good start and glad you agree with the approach.  Thanks for going the extra step with this pull request.

anagav:
@spmallette  I have added code based on your comments. 
* I have added maven profile activation based on os
* enable the setting in `AbstractGremlinServerIntegrationTest`
* use this setting while running test to use epoll (if not present fallback to NIO).

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GremlinServer.java
```diff
@@ -99,11 +114,16 @@ public GremlinServer(final Settings settings) {
     public GremlinServer(final ServerGremlinExecutor<EventLoopGroup> serverGremlinExecutor) {
         this.serverGremlinExecutor = serverGremlinExecutor;
         this.settings = serverGremlinExecutor.getSettings();
+        this.isLinuxOS = settings.useEpollEventLoop && SystemUtils.IS_OS_LINUX;
```
spmallette: I think that it would be worth doing a `logger.warn()` after this line if someone tries to "useEpoll" but are not on linux.  I'd further add a `logger.info()` to indicate the event loop ultimately used.

gremlin-server/pom.xml
```diff
@@ -148,6 +149,29 @@ limitations under the License.
     </build>
     <profiles>
         <profile>
+            <id>linux</id>
+            <activation>
+                <os>
+                    <family>Linux</family>
```
spmallette: Doesn't this mean that if I'm on linux, i'll always be testing epoll as the profile "activates" when the os is linux?  I think we'd want the flexibility to be able to test java nio on linux, which we don't get on this approach.  In other words, instead of activating by os, it should activate when requested with:

```xml
<activation>
  <activeByDefault>false</activeByDefault>
  <property>
    <name>epoll</name>
  </property>
</activation>
```

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GremlinServer.java
```diff
@@ -71,21 +74,33 @@
...
     private final ExecutorService gremlinExecutorService;
     private final ServerGremlinExecutor<EventLoopGroup> serverGremlinExecutor;
+    private final boolean isLinuxOS;
 
     /**
      * Construct a Gremlin Server instance from {@link Settings}.
      */
     public GremlinServer(final Settings settings) {
         this.settings = settings;
+        this.isLinuxOS = settings.useEpollEventLoop && SystemUtils.IS_OS_LINUX;
```
spmallette: minor nit....can you please rename `isLinuxOS` to something more fitting, like `epollEnabled`?

spmallette:
This is shaping up into a nice PR.  You probably expected to get to this last, but just as a reminder, the docs need to be updated as well with your new configuration option: `useEpollEventLoop`.  Please add it to the table here:

https://github.com/apache/incubator-tinkerpop/blob/master/docs/src/gremlin-applications.asciidoc#configuring

anagav:
@spmallette I have updated the code based on your comments. I will update the doc and send a different PR.

Please let me know if there are any more fixes.

anagav:
@spmallette I have updated the doc :smile:

spmallette:
Created an issue to track this against: 

https://issues.apache.org/jira/browse/TINKERPOP3-901

spmallette:
Thanks for making those changes - getting into nuts and bolts now.  I'm testing your changes locally.  In doing so, I think you can back out your change to the gremlin-driver pom.xml that adds:

```xml
<dependency>
  <groupId>io.netty</groupId>
  <artifactId>netty-transport-native-epoll</artifactId>
  <version>${netty.version}</version>
</dependency>
```

epoll seems to come with `netty-all` which is already present.  Tests seem to work for me when I remove that extra dependency, but I thought I'd check with you first to make sure I'm not missing anything.  

Please validate that everything works on your end without that change and if so, amend your PR.  At that point, I will ask other committers to review and we can work toward merging.

anagav:
@spmallette done

spmallette:
VOTE: +1

This PR feels ready for general review from committers.  Here's a summary for those who haven't been following along:

* The PR adds a configuration to Gremlin Server to use a linux-only netty feature that has shown to improve performance.
* @anagav was good enough to work tests such that we can run Gremlin Server integration tests with and without this feature by way of a secondary maven profile.  In this way we can run tests on any platform.

From my testing perspective, I successfully did:

```text
mvn clean install
cd gremlin-server
mvn verify -DskipIntegrationTests=false
mvn verify -DskipIntegrationTests=false -Plinux
```

Even if you're not testing on linux it would be good to try the `-Plinux` option as it will validate the fallback to java NIO and still work.  

I'll handle the merge of this PR (when the VOTE is complete) and update changelog in the process.

okram:
I reviewed the code and the code looks clean. Basically, `if(onLinux) doLinuxOptimizedSocket/Threads`. That compounded with @spmallette's expertise in this area of the code base and his review of the PR:

VOTE +1.

gremlin-server/pom.xml
```diff
@@ -148,6 +149,30 @@ limitations under the License.
     </build>
     <profiles>
         <profile>
+            <id>linux</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+                <property>
+                    <name>EPOLL</name>
```
pluradj: Should add a note in the contributing doc about the `EPOLL` profile, similar to `includeNeo4j` https://github.com/apache/incubator-tinkerpop/blob/master/CONTRIBUTING.asciidoc

spmallette: good point - i might ultimately adjust naming/docs and such once we merge.  @anagav i can handle that on my end.  @pluradj is that cool for you for purpose of voting?

pluradj:
Reviewed the code. Tested with and without EPOLL on Linux, tested on non-Linux platforms also.

VOTE: +1

spmallette:
Thanks @anagav - nice PR.

okram:
Great @pluradj, thorough.

anagav:
Thanks @spmallette for merging the PR.

PR #110, opened 2015-10-16
TINKERPOP3-858 Cleanup after failed plugin install

pluradj:
https://issues.apache.org/jira/browse/TINKERPOP3-858 see manual testing steps in the JIRA

### DIFF
--- a/gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/commands/InstallCommand.groovy
+++ b/gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/commands/InstallCommand.groovy
@@ -53,6 +53,12 @@ class InstallCommand extends CommandSupport {
             final def pluginsThatNeedRestart = grabDeps(dep)
             return "Loaded: " + arguments + (pluginsThatNeedRestart.size() == 0 ? "" : " - restart the console to use $pluginsThatNeedRestart")
         } catch (Exception ex) {
+            if (!(ex instanceof IllegalStateException)) {
+                // IllegalStateException is thrown if a module with the same name is already installed.
+                final def uninstall = new UninstallCommand(shell, mediator)
+                final List<String> module = Collections.singletonList(artifact.getArtifact())
+                uninstall.execute(module)
+            }
             return ex.message
         }
     }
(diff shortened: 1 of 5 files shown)


spmallette:
VOTE: +1

gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/commands/InstallCommand.groovy
```diff
@@ -53,6 +53,12 @@ class InstallCommand extends CommandSupport {
             final def pluginsThatNeedRestart = grabDeps(dep)
             return "Loaded: " + arguments + (pluginsThatNeedRestart.size() == 0 ? "" : " - restart the console to use $pluginsThatNeedRestart")
         } catch (Exception ex) {
+            if (!(ex instanceof IllegalStateException)) {
+                // IllegalStateException is thrown if a module with the same name is already installed.
+                final def uninstall = new UninstallCommand(shell, mediator)
+                final List<String> module = Collections.singletonList(artifact.getArtifact())
+                uninstall.execute(module)
```
mhfrantz: Can uninstall throw an exception?

pluradj:
@dkuppitz `DependencyGrabber.copyDependenciesToPath()` throws two exceptions explicitly, `IOException` and `IllegalStateException`. `RuntimeException` occurs if there is an error grabbing grapes. This is the initial bug report, and its failed directory should be rolled back.

`IOException` occurs if there are problems creating the directories under `ext`. Rollback should occur here, but most likely there isn't even a directory to delete.

`IllegalStateException` occurs if plugin module already exists. For example, `tinkergraph-gremlin` is installed by default out of the box. If the code doesn't catch the `IllegalStateException`, `tinkergraph-gremlin` would get *uninstalled* if the user tried to install it a second time.

gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/commands/InstallCommand.groovy
pluradj: No, I didn't see any exceptions declared or thrown in the class.

dkuppitz:
+1

okram:
If @pluradj votes +1, then there will be three. Thus, @pluradj  -- if you think your work is solid, then +1 it and tally the vote and merge. Good luck.

pluradj:
Thanks for the comments. I've centralized the rollback logic into DependencyGrabber, and added a couple unit tests.

pluradj:
Vote: +1

okram:
You have 3 +1 votes. Please merge and if this is going into tp30, don't forget to merge to master as well. Good luck.

PR #112, opened 2015-10-19
TINKERPOP3-863: Turn off bulking -- or is there something more general? (hope not).

okram:
https://issues.apache.org/jira/browse/TINKERPOP3-863

* Added `GraphTraversalSource.withBulk(boolean)`, where `true` is default.
* Added `Traversal.Admin.addRequirement()` to allow for non-step based assignment of traverser requirements.
* Removed `PathIdentityStep` as we can now `addRequirement(PATH)` at `withPath()`.
* Updated the docs demonstrating how `withBulk(false)` no longer creates weird behaviors around energy duplication.
* Added a new suite of traverser species that support `TraverserRequirement.ONE_BULK` (code duplication is key to ensure a small footprint in traverser serialization in OLAP).
* Added a new test case to `SackTest` that proves that bulking turned off does not create weird behavior.
* Added new test case to ensure that `DefaultTraverserGeneratorFactory` is producing the right traverser species for particular types of traversals.
* Updated `GryoMapper` with new traverser serializations.
* Fixed up the various pre-existing `TraverserGenerators` so that their requirements are in alphabetical order and thus, easier to map to their name.

Testing notes:
* Full docs built and succeeded.
* Full integration tests passed.

Will updated CHANGELOG on merge.


okram:
I VOTE +1 on this. A master piece of code. You have truly outdone yourself Marko, bravo.

mhfrantz:
I know you said you'd update the CHANGELOG on merge, but I think that would help set the context here.  What are the new use cases that this change allows?  There is a lot of conversation in the JIRA ticket, but if you had to give an executive summary in the CHANGELOG, could you maybe add that as a comment to this PR, and then copy-paste it into the CHANGELOG later?  Something about modeling conservative processes, etc.  I feel like I'm this close [holds finger close to thumb] to understanding how this changes everything.

okram:
Bulk and sack do not play well together when you are dealing with "energy flows." Please see the existing 3.0.1-SNAPSHOT docs for a simple example:

```
gremlin> g.withSack(1.0f,sum).V(1).local(outE('knows').barrier(normSack).inV()).in('knows').barrier().sack() //(5)
==>1.0
==>1.0
gremlin> g.withSack(1.0f,sum).V(1).local(outE('knows').barrier(normSack).inV()).in('knows').barrier().sideEffect{it.setBulk(1)}.sack() //(6)
==>1.0
```

```
5. The barrier() merges the two traversers at vertex 1 into a single traverser whose sack is 1.0.

6. There is now a single traverser with bulk of 2 and sack of 1.0 and thus, setting the bulk to 1 yields the expected 1.0.
```

Now look at what changes I made to the docs and see how you no longer need `sideEffect{it.setBulk(1)}`.

mhfrantz:
VOTE +1

spmallette:
VOTE: +1 (that's 3)

PR #114, opened 2015-10-20
Fix empty buffer return upon buffer capacity exceeded

christianbellinaef:
Discussion: https://groups.google.com/forum/#!topic/gremlin-users/zWGAn-oS5PQ

Return the underlying byte array output stream which will always contain the full serialized object.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ser/GryoMessageSerializerV1d0.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ser/GryoMessageSerializerV1d0.java
@@ -225,7 +225,7 @@ public final class GryoMessageSerializerV1d0 implements MessageSerializer {
         ByteBuf encodedMessage = null;
         try {
             final Kryo kryo = kryoThreadLocal.get();
-            try (final OutputStream baos = new ByteArrayOutputStream()) {
+            try (final ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
                 final Output output = new Output(baos, bufferSize);
 
                 // request id - if present
@@ -244,9 +244,9 @@ public final class GryoMessageSerializerV1d0 implements MessageSerializer {
                 if (size > Integer.MAX_VALUE)
                     throw new SerializationException(String.format("Message size of %s exceeds allocatable space", size));
 
+                output.flush();
                 encodedMessage = allocator.buffer((int) size);
-                if (size > bufferSize) output.flush();
-                encodedMessage.writeBytes(output.toBytes());
+                encodedMessage.writeBytes(baos.toByteArray());
             }
 
             return encodedMessage;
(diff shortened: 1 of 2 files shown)


adrianluisgonzalez:
+1

okram:
Awesome. Can you please answer the following questions:

1. Have you run full "mvn clean install" and it passes?
2. Is there a test case you can write to validate that without this, there is a problem and with it, it is solved? (this way, in the future, if we tweak something, a test says "no no.")

spmallette:
Now that I look at this more closely, I feel like there was a reason for that being there (as the code was the way you had it before hand).  I think I wrote a test that validates that, so on top of `mvn clean install` you should also be sure that the integration tests for gremlin-server succeed:

```text
mvn clean install
cd gremlin-server
mvn verify -DskipIntegrationTests=false
```

With respect to Marko's question about adding a test case, you might be able to write a unit test here:

https://github.com/apache/incubator-tinkerpop/blob/b8dbfc34f523f60be4c341699414fb5e28e8e6aa/gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/ser/GryoMessageSerializerV1d0Test.java

If you can only re-create it top-down from the Gremlin Server side, then let me know and I can get you oriented on where/how to do that.

Finally, relating to the code change itself, shouldn't the same changes be made in "request" serialization here (where code is quite similar):

https://github.com/englishtown/incubator-tinkerpop/blob/serializer-returns-empty-buffer-fix/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ser/GryoMessageSerializerV1d0.java#L293

christianbellinaef:
I added test coverage and also applied the changes to the response method. Mvn clean install, and integration tests both pass.

Looking at the Kryo docs, I see that output.toBytes() gets the [bytes currently in the output's *buffer*](http://javadox.com/com.esotericsoftware.kryo/kryo/2.23.0/com/esotericsoftware/kryo/io/Output.html#toBytes()) not the bytes in the underlying output stream being written to.  

Calling flush clears the buffer and writes the data, and since we are currently returning the buffer and not the stream, that would explain the empty return message.

okram:
The code looks clean to me and I like the test case. Note that this will effect Spark as it uses Gryo for its internal serialization. However, you said you ran `-DskipIntegrationTests=false` so thats great. When the final merge happens, I'll re-run the integration tests locally just to be certain.

VOTE +1 (binding).

spmallette:
Created an issue in JIRA to tie this to:

https://issues.apache.org/jira/browse/TINKERPOP3-903

@okram note that i think @christianbellinaef only ran Gremlin Server integration tests.  I wouldn't anticipate problems though with spark as the code that was changed is related to serializing Gremlin Server request/response messages which I don't think have much to do with spark.

One of these days I'm going to know when to flush and when not to :toilet: 

In the future, please watch your IDE settings - as a point of code style TinkerPop code doesn't use wildcarded imports:

https://github.com/apache/incubator-tinkerpop/pull/114/files#diff-c68ab1009a5e6ce5daaf5f10349a34eeR52

I'll fix that up on the merge and update CHANGELOG.

I ran `mvn clean install` and integration tests for Gremlin Server and they seem to work ok.

VOTE: +1 (binding)

@pluradj do you have some time to give us the final vote we need to get this merged?

pluradj:
Built successfully with `mvn clean install -DcleanGrapes`.
Manually ran a couple subgraph() tests with Gremlin Server and Gremlin Console (remote).
Concur with the comments on the wildcard imports. Other than that, looks good to me.
Thanks, Christian!
Vote: +1 (binding)

spmallette:
@christianbellinaef this has been merged - thanks for doing this.

PR #115, opened 2015-10-20
TINKERPOP3-898: Rename ConjuctionP and ConjuctionStep to ConnectiveP and ConnectiveStep

okram:
https://issues.apache.org/jira/browse/TINKERPOP3-898

In logic, "conjunction" implies AND. The general term for both AND and OR is "connective."

* Renamed all `ConjunctionXXX` to `ConnectiveXXX`.
* This is a breaking change but very internal to TinkerPop steps and given that we are not exposing the "instruction set" paradigm yet, its best this happens now than later.
* I ran `mvn clean install` and everything built. This does NOT effect docs nor integration testing.
* I updated the `CHANGELOG.asciidoc`.

I VOTE +1 (binding).


dkuppitz:
Vote: +1

Very simple refactoring, nothing to complain about.

spmallette:
VOTE: +1 (binding)

Ran `mvn clean install` - no problems....you have 3 +1s - time to merge

PR #116, opened 2015-10-21
TINKERPOP3-861: Solve "The Number Problem" for Operator (and follow on operators)

dkuppitz:
I implemented `NumberHelper`, which a) determines the highest common `Number` type for a list of given `Number`s and b) provides simple operations like `min`, `max`, `add`, `sub`, etc..

The `NumberHelper` is now used in all the `Operator` enums and ensures that the result type is always the highest common `Number` type and thus results are always accurate.

**Before**

```
Operator.mul.apply(1, 0.5d) ==> 0
Operator.add.apply(1, 0.5d) ==> 1
```

**After**

```
Operator.mul.apply(1, 0.5d) ==> 0.5d
Operator.add.apply(1, 0.5d) ==> 1.5d
```

Furthermore `Operator` now works for any `Number` type (e.g. `BigInteger`, `BigDecimal` and any custom `Number` implementation).


okram:
This is really solid work. The code is clean, the test cases are thorough. 

* I am assuming you ran `mvn clean install`.
* Be sure to update CHANGELOG on merge.
* Are there other areas of the codebase that can leverage `NumberHelper`? (perhaps email dev@ and lets discuss there after merge).
* This is potentially *breaking* if someone was relying on 0.5 * 1 to be 0. :) ... Regardless add this to the upgrade docs too so people know.

VOTE +1.

spmallette:
@dkuppitz do you mind amending the PR to include the upgrade docs?  @okram raises a good point on that - would like to see that in place prior to voting if that's cool.

dkuppitz:
@spmallette I updated CHANGELOG and the upgrade docs.

Also, since forgot it in my initial comment:

My vote: +1

pluradj:
Built cleanly with `mvn clean install`.
Ran a few manual tests against Operator from the Gremlin Console.
Reviewed the doc updates.
Learned something new about TinkerPop today.

Vote: +1

PR #120, opened 2015-10-24
TINKERPOP3-903 - amend a test

spmallette:
Add a test to validate that results of all sizes stream properly.

This is an amendment to 

https://issues.apache.org/jira/browse/TINKERPOP3-903 

which was already merged and closed.  As @rjbriody was describing the problem he was seeing, it would be cool if he could review this test to see if it is sufficient for covering the problems he had.  I'd like to have more tests in this area if he can think of holes in this one.  We can vote to merge once we get his :+1:

### DIFF
--- a/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
+++ b/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
@@ -20,8 +20,11 @@ package org.apache.tinkerpop.gremlin.server;
 
 import org.apache.commons.lang.exception.ExceptionUtils;
 import org.apache.tinkerpop.gremlin.TestHelper;
-import org.apache.tinkerpop.gremlin.driver.*;
 import org.apache.tinkerpop.gremlin.driver.Channelizer;
+import org.apache.tinkerpop.gremlin.driver.Client;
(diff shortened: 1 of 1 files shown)


dkuppitz:
Looks like the script does a lot of things that are not really required for the test. I think we should save some CPU cycles wherever we can, since the tests already take long enough.

```groovy
def rand = new Random()
def ii = 0
def ids = [graph.addVertex().id()]
while (ii++ < size) {
  def v = graph.addVertex()
  def u = graph.vertices(ids.get(rand.nextInt(ids.size()))).next()
  v.addEdge('linked', u)
  ids += [v, u]*.id()
}
g.V()
```

In my opinion that should be sufficient.

spmallette:
I'll let @rjbriody say if your change is ok @dkuppitz but I think having properties were an important part of what was required in the test.  Note that this test only runs on TinkerGraph in the integration tests for gremlin server so it's pretty fast (seconds).

dkuppitz:
Ok, there's nothing else wrong with the test, I just wanted to point out, that it *probably* does a lot of unnecessary stuff. If it's needed, cool.

Vote: +1

spmallette:
I like that @dkuppitz looked at this with a discerning eye and questioned the motives of the test, but I guess i could have CTR'd this one.  As we're already in the vote, @okram can you just have a look real quick and vote?  

Since @rjbriody is good with the test and it passes nicely:

VOTE: +1

okram:
VOTE +1.

PR #126, opened 2015-10-28
TINKERPOP-857: Add GraphComputer.config(key,value)

okram:
https://issues.apache.org/jira/browse/TINKERPOP3-857

Added `GraphComputer.config(key,value)`. This is NOT a universal method. That is, each `GraphComputer` can decide, for itself, what each key/value should do. Moreover, what works for `TinkerGraphComputer` doesn't necessarily work for `GiraphGraphComputer`. This is here so users can fluently add configuration options to their `GraphComputer` above and beyond what is specified in their `Graph.configuration()`. 

Run test suite. VOTE +1.


spmallette:
+1 - `mvn clean install` seems good to go.

dkuppitz:
Looks good to me, although it's not clear to me why it's tagged as a breaking change.
Anyway, +1, can be merged.

PR #127, opened 2015-10-28
TINKERPOP3-918: ComputerVerificationStrategy is too restrictive

okram:
@dkuppitz found a restrictive verification that has now been loosened up. I added the two examples that Kuppitz demonstrated that should be allowed but were not. In the process, I found a bug in how MapReduce comparators work and fixed it. 

I ran `mvn clean install`. I also did Spark integration testing and tested the known tests that this effects in Giraph.

When merged, I will updated CHANGELOG.

VOTE +1.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/TraversalHelper.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/TraversalHelper.java
@@ -20,6 +20,8 @@ package org.apache.tinkerpop.gremlin.process.traversal.util;
 
 import org.apache.tinkerpop.gremlin.process.traversal.Step;
 import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
+import org.apache.tinkerpop.gremlin.process.traversal.lambda.ElementValueTraversal;
+import org.apache.tinkerpop.gremlin.process.traversal.lambda.TokenTraversal;
 import org.apache.tinkerpop.gremlin.process.traversal.step.HasContainerHolder;
 import org.apache.tinkerpop.gremlin.process.traversal.step.Scoping;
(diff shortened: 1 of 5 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/TraversalHelper.java
```diff
@@ -57,6 +60,23 @@
     private TraversalHelper() {
     }
 
+    public static boolean isNotBeyondElementId(final Traversal.Admin<?, ?> traversal) {
```
spmallette: I think some javadoc here in `TraversalHelper` would be smart.  These methods are very useful in strategy development which is externally facing to providers/users.  `TraversalHelper` also has a good body of unit tests around it at this point - adding some coverage for this new method would be good too:

https://github.com/apache/incubator-tinkerpop/blob/master/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/util/TraversalHelperTest.java

dkuppitz:
Code looks great.

+1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/TraversalHelper.java
okram: We really need to get this method hidden as if you note me response. I don't think we have all case covered. This really should be in 3.2.0 when we have a slimmed down instruction set. What I will do on merge is move this to ComputerVerifcationStrategy and make it private. Then come 3.2.0 we can move it to TraversalHelper.

spmallette: Fair enough.

spmallette:
+1 

RESULT: +1 - 3 (binding)

PR #130, opened 2015-10-31
TINKERPOP3-899: Bump to the latest version of Neo4j.

okram:
https://issues.apache.org/jira/browse/TINKERPOP3-899

This has been implemented and tested with `-DincludeNeo4j`. Note the `<exclusions>` I had to do in `<scope>test</scope>` as Neo4j doesn't use enforcer-plugin and they have self conflicts. I picked the highest version number of the self-conflict as the version to use.

Again, tested with `mvn clean install -DincludeNeo4j`.

My VOTE is +1.

### DIFF
--- a/neo4j-gremlin/src/main/java/org/apache/tinkerpop/gremlin/neo4j/structure/Neo4jHelper.java
+++ b/neo4j-gremlin/src/main/java/org/apache/tinkerpop/gremlin/neo4j/structure/Neo4jHelper.java
@@ -19,15 +19,14 @@
 package org.apache.tinkerpop.gremlin.neo4j.structure;
 
 import org.apache.tinkerpop.gremlin.structure.Direction;
-import org.apache.tinkerpop.gremlin.structure.Element;
-import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.neo4j.tinkerpop.api.Neo4jNode;
-import org.neo4j.tinkerpop.api.Neo4jRelationship;
 
 /**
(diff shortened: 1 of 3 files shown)


neo4j-gremlin/src/main/java/org/apache/tinkerpop/gremlin/neo4j/structure/Neo4jHelper.java
```diff
@@ -45,13 +44,16 @@ public static boolean isDeleted(final Neo4jNode node) {
...
+                return true;
+            else
+                throw e;
         }
     }
 
-    public static boolean isNotFound(RuntimeException ex) {
-        return ex.getClass().getSimpleName().equals("NotFoundException");
+    public static boolean isNotFound(final RuntimeException ex) {
+        return ex.getClass().getSimpleName().equals(NOT_FOUND_EXCEPTION);
```
dkuppitz: There's probably a good reason for the `toString()` comparison, but wouldn't it be easier/cleaner to do this:

```java
try {
    node.getKeys();
    return false;
} catch (NotFoundException e) {
    return true;
}
```

okram: `NotFoundException` is in Neo4j and that is not an Apache2 class :|.

dkuppitz:
I see. Ugly, but functional.

+1

PommeVerte:
Tested (int. tests too). Everything seems to be working.

+1

PR #133, opened 2015-11-03
TINKERPOP3-904: BulkLoaderVertexProgram optimizations

dkuppitz:
`BulkLoaderVertexProgram` now uses `EventStrategy` to monitor what the underlying `BulkLoader` implementation does (e.g. whether it creates a new vertex or returns an existing). This way we don't need modify method signatures and introduce breaking changes.

Full integrations testsuite passed*

VOTE: +1

_* I had to exclude Giraph integration tests, since it no longer works on my machine_


okram:
I reviewed the code and it looked good. It doesn't touch anything that anyone else works on so its an island unto @dkuppitz world and thus, safe from my perspective. If Kuppitz is happy with the work and test pass. Good.

VOTE +1.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/bulkloading/BulkLoaderVertexProgram.java
```diff
@@ -407,4 +415,84 @@ public boolean requiresVertexPropertyAddition() {
...
+
+    static class BulkLoadingListener implements MutationListener {
+
+        private long counter;
+        private boolean isNewVertex;
+
+        public BulkLoadingListener() {
+            this.counter = 0L;
+            this.isNewVertex = false;
+            ;
```
spmallette: you have an empty code line here - just something to remove before you merge.

spmallette:
This should be a good test of `EventStrategy` as it is not used internally anywhere else in the code base.  Is there any change required to the reference docs?  

if no docs are required then +1 from me:

RESULT: +1 (3 binding)

dkuppitz:
> Is there any change required to the reference docs?

Yes, I thought about it last night. I will add a comment in the docs, since users now have to ensure that a custom `BulkLoader` implementation uses things like `g.addV()` instead of `graph.addVertex()`, otherwise the `EventStrategy` would be blind and assume that nothing has ever been created. It wouldn't break aything, but it would prevent BLVP from using the optimizations.

spmallette:
I guess there is no way to enforce that the custom `BulkLoader` implementation would use a `TraversalSource` instead of `Graph`.  That's definitely something `BulkLoader` javadocs and reference docs should have especially if it isn't cleanly enforced.

PR #134, opened 2015-11-03
TINKERPOP3-768: MatchStep in OLAP should be smart about current vertex.

okram:
https://issues.apache.org/jira/browse/TINKERPOP3-768

The previous implementation of MatchStep was not biased to patterns that start at the current vertex. In OLTP, this is good. In OLAP, this is really really really bad as you are incuring a massive overhead with inter process communication. Thus, optimization, in OLAP, will sort the patterns first by whether the traverser needs to leave the current object location and then by TraversalType and then by multiplicity of the pattern. In the process, I cleaned up a few things here and there in MatchStep.

Ran `mvn clean install` and all is good.

VOTE +1.


dkuppitz:
Code looks clean. I assume that the full integration test suite also passes.

VOTE: +1

okram:
@dkuppitz --- Spark integration tests passed. Giraph is still running (but the changes won't effect the integration test suites).

PommeVerte:
Tests pass for me as well. it all looks clean. 

VOTE: +1

PR #138, opened 2015-11-06
TINKERPOP3-923 - Add the "Getting Started" Tutorial

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP3-923

You can see the finished product here:

http://tinkerpop.incubator.apache.org/docs/3.1.0-SNAPSHOT/tutorials-getting-started.html

I'll polish up spelling/grammar during code freeze week.  At this point, I think the VOTE is on the "tutorial pattern", general content, and property doc generation.

VOTE: +1


PommeVerte:
I'm really liking the structure here. Thought it was clever to add all of those tutorials in a row like this. No need for an index. 

As far as the implementation goes I'm guessing this means new tutorials will require configuring `sourceDocumentName`?  I don't really remember everything that came out of the discussions around the folder structure for these but I'm guessing this is preferable to a "drop in" solution? 
I'm ok with this either way so my VOTE is +1

dkuppitz:
I didn't go through all of it in every little detail, but - although it looks like would take the reader more than 30 minutes - I like he structure.

VOTE: +1

spmallette:
@dkuppitz it depends on how you wan to measure the minutes.  in pure "reading" time, this is under 30 minutes.  "comprehension" time, trying all the examples, etc. - that's probably over 30 minutes.  i was shooting for "reading" time (which i've tested)

PR #140, opened 2015-11-07
TINKERPOP3-885 Revision of #113

spmallette:
this is a revision of the work done here: #113 - the other had two +1s already but I thought that given the added changes it should be reviewed yet again.  this commit - a806c7ec0a4fc5baaa1d684a647d0e5084f9c4cb - has the description of the changes since #113 

you can test with:

```text
mvn clean install -DincludeNeo4j
```

Note that I tested against Titan as well to validate the changes so that should give us added confidence that this change will work for everyone.


dkuppitz:
Nice and solid code.

VOTE: +1

PommeVerte:
Like the changes. This looks nice.
Tests pass.
Definite +1 from me.

PR #142, opened 2015-11-10
TINKERPOP3-953 Artifact equality is not evaluating properly

n-tran:
https://issues.apache.org/jira/browse/TINKERPOP3-953

Corrected bug in artifact.equal where a local variable was overriding a member variable with the same name.


spmallette:
This fixed a minor bug and added some good test coverage.  Fairly trivial code wise.  Merged via CTR.  Thanks

PR #143, opened 2015-11-11
TINKERPOP3-955 Registered HashMap$Node to GryoMapper.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP3-955

Note that this class had to be registered via reflection as it is a private class of HashMap. Added tests to validate that the change works with the code specified in the issue. 

Please review as it pertains to area of serialization that you are familiar with.  To test with gremlin-server you need to run the integration tests:

```text
mvn clean install
mvn verify -pl gremlin-server -DskipIntegrationTests=false
```

VOTE: +1

btw, we _could_ postpone this to 3.1.1 if we wanted - i think. there is a workaround to the bug in that the person just doesn't have to `next()` the result to get it to serialize.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoMapper.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoMapper.java
@@ -174,6 +174,22 @@ public final class GryoMapper implements Mapper<Kryo> {
         private static final Class LINKED_HASH_MAP_ENTRY_CLASS = m.entrySet().iterator().next().getClass();
 
         /**
+         * The {@code HashMap$Node} class comes into serialization play when a {@code Map.entrySet()} is
+         * serialized.
+         */
+        private static final Class HASH_MAP_NODE;
+
+        static {
+            // have to instantiate this via reflection because it is a private inner class of HashMap
+            final String className = HashMap.class.getName() + "$Node";
+            try {
+                HASH_MAP_NODE = Class.forName(className);
+            } catch (Exception ex) {
+                throw new RuntimeException("Could not access " + className, ex);
+            }
+        }
+
+        /**
          * Note that the following are pre-registered boolean, Boolean, byte, Byte, char, Character, double, Double,
          * int, Integer, float, Float, long, Long, short, Short, String, void.
          */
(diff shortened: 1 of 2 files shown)


okram:
+1 ... this should have been a CTR. This is wasting people's time/effort.

spmallette:
i'm really don't agree - during code freeze, all commits should be under scrutiny. outside of code freeze i would have CTR'd.  also, i don't think anyone who wants a review should feel reticent to ask for one in or out of code freeze regardless of complexity.

PommeVerte:
Straightforward enough, Tests pass VOTE: +1

pluradj:
Code looks fine. Tests pass. Wondering if other private inner classes like this one are lurking...
VOTE: +1

spmallette:
i suspect more @pluradj - would be nice to get some more serialization tests in play to root out other problems.  I made the tests i supplied here integration tests to make sure the driver was working properly end to end, but i'll tack in a few more here as unit tests for additional coverage:

https://github.com/apache/incubator-tinkerpop/tree/master/gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/ser

note that we relaxed the need for registration in 3.1.0 by adding a config to flip that switch, but it's better to register than not and registration remains the default.

spmallette:
Added the extra tests and then found that GraphSON with embedded types failed as well.  Here's the commit with the bonus tests and GraphSON serializer for `Map.Entry`:

https://github.com/apache/incubator-tinkerpop/commit/e35e6c86d93cbb7e8c6618212d63eb3020eed96d

It basically just treats each entry as its own `Map`.

pluradj:
Nice catch, Stephen.

spmallette:
i'm not done catching either........ :(

pluradj:
https://www.youtube.com/watch?v=s-eEV7b4UCo

spmallette:
haha - i'm an old head though so saturday morning cartoons for me that involved catching stuff was restricted to the Wile E Coyote trying to snag the Roadrunner: 

https://www.youtube.com/watch?v=Gq_bjaI0NTo

.....and still funny!

spmallette:
Ok - I think I got this finalized now with: https://github.com/apache/incubator-tinkerpop/commit/cb5ea9010374d41fe6347237d42304d3eb00093f

The integration tests pointed me to the failure.  I made some other improvements in the process which are outlined in the commit message.  

@rjbriody if there's any chance you could build and test to your satisfaction before we release this thing, that would be awesome.

rjbriody:
LGTM. +1

PR #144, opened 2015-11-16
TINKERPOP3-957: Improve speed of addV()

twilmes:
https://issues.apache.org/jira/browse/TINKERPOP3-957

I profiled the basic addV traversal and made a number of spot optimizations based upon the top 5 or so bottlenecks.  These included:
* Converting a number of stream calls into basic for loops
* Cutting down on the allocation of new unmodifiable step collections by wrapping the steps once on init
* Avoiding object allocation through the use of Collections.arrayContains

I did not end up making any changes to the addV or addE steps themselves because the majority of the run time is not coming from the execution of those steps themselves.

Performance results on my box of running 100,000 addVs
Before: 341 ms
After: 224 ms

Ran mvn clean install without issue.  I did not add or update any tests because the spots that were updated already had coverage.

I just learned the hard way that I cannot rewrite git history in the apache repo so there are 2 extra commits associated with this due to a rebase and an inability to force a push.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -22,10 +22,11 @@ TinkerPop 3.1.0 (A 187 On The Undercover Gremlinz)
 
 image::https://raw.githubusercontent.com/apache/incubator-tinkerpop/master/docs/static/images/gremlin-gangster.png[width=185]
 
+[[release-3.1.0-incubating]]
 TinkerPop 3.1.0 (Release Date: November 16, 2015)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-This release also includes changes from <<3-0-1-incubating, 3.0.1-incubating>> and <<3-0-2-incubating, 3.0.2-incubating>>.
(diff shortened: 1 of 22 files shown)


okram:
I ran `mvn clean install` and everything passed. I reviewed the code and these are very simple changes to our `TraversalHelper`/`DefaultTraversal` classes. For example, replacing `Stream` with `for-loop` is always good. Given that these methods are so core to everything traversal, making these as fast as possible is smart. Thus, this is a really useful PR. Hopefully we can expose more performance problem areas and address them much like this ticket.

VOTE +1.

spmallette:
do we keep this ticket open even after merging this PR?

twilmes:
One thought I had was to create a new discovery type ticket to go in and do a more thorough analysis across a wider range of traversal uses to pull out other areas we can make similar improvements that will decrease traversal execution overhead.  That could then lead to some specific, targeted tickets.

dkuppitz:
Nice one. `addV` is a lot faster then before; still a lot slower than `addVertex`, but that's expected.

VOTE: +1

PR #147, opened 2015-11-20
TINKERPOP3-978 Native GraphSON Serialization for Gremlin Server

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP3-978

We already had this feature for Gryo but it wasn't built for GraphSON as the original idea was to mostly support serialization over `:remote`.  That was a bit short-sighted in thinking so I've added the same capability for GraphSON as will be especially useful to non-JVM languages.

you can test with `mvn clean install` followed by the Gremlin Server integration tests:

```text
```

my vote is +1

@PommeVerte @jbmusso @davebshow - this one may be interesting to you and your driver work

### DIFF
--- a/tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerIoRegistry.java
+++ b/tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerIoRegistry.java
@@ -18,18 +18,44 @@
  */
 package org.apache.tinkerpop.gremlin.tinkergraph.structure;
 
+import org.apache.tinkerpop.gremlin.structure.Edge;
+import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.apache.tinkerpop.gremlin.structure.io.AbstractIoRegistry;
+import org.apache.tinkerpop.gremlin.structure.io.GraphReader;
+import org.apache.tinkerpop.gremlin.structure.io.GraphWriter;
 import org.apache.tinkerpop.gremlin.structure.io.IoRegistry;
+import org.apache.tinkerpop.gremlin.structure.io.graphson.GraphSONIo;
+import org.apache.tinkerpop.gremlin.structure.io.graphson.GraphSONTokens;
+import org.apache.tinkerpop.gremlin.structure.io.graphson.GraphSONUtil;
 import org.apache.tinkerpop.gremlin.structure.io.gryo.GryoIo;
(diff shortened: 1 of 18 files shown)


PommeVerte:
ok, so I wanted to be thorough with this. Hence de delay.
- `mvn clean install` passes
- `mvn verify -DskipIntegrationTests=false -pl gremlin-server` passes
- wrote some driver side tests and they work properly. but shouldn't `config: { useMapperFromGraph: graph }` be added to the default configuration as well? 

If there's some drawback to adding the above to the default `gremlin-server.yaml` and the PR should stay as is then:

VOTE: +1

If there isn't any noticeable drawback I'd be tempted to squeeze that in before merging.

spmallette:
Good idea @PommeVerte - updated the packaged config files to include `useMapperFromGraph`.

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerIoRegistry.java
```diff
@@ -86,4 +113,145 @@ public TinkerGraph read(final Kryo kryo, final Input input, final Class<TinkerGr
...
+
+                if (!jsonParser.getCurrentName().equals(GraphSONTokens.EDGES))
+                    throw new IOException(String.format("Expected a '%s' key", GraphSONTokens.EDGES));
+
+                jsonParser.nextToken();
+                edges = (List<Map<String, Object>>) deserializationContext.readValue(jsonParser, ArrayList.class);
+            } else {
+                final Map<String, Object> graphData = deserializationContext.readValue(jsonParser, HashMap.class);
+                vertices = (List<Map<String,Object>>) graphData.get("vertices");
+                edges = (List<Map<String,Object>>) graphData.get("edges");
```
pluradj: nit... `GraphSONTokens.VERTICES` and `GraphSONTokens.EDGES` here

pluradj:
Did a clean install, then the Gremlin Server integration tests, then a code inspection.

VOTE: +1

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerIoRegistry.java
spmallette: oops - yes @pluradj - good find - will fix before merge.

spmallette:
Fixed with the above commit @pluradj - merging

PR #149, opened 2015-11-20
Fix fo TINKERPOP3-945

velo:
Added root cause to exceptions where a root cause is available

To make the decision of where to add root cause, I looked the impl (neo4j, hadoop, spark and orientDB) and add it when the root cause is available.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/Graph.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/Graph.java
@@ -364,9 +364,13 @@ public interface Graph extends AutoCloseable, Host {
             public static IllegalArgumentException variableValueCanNotBeNull() {
                 return new IllegalArgumentException("Graph variable value can not be null");
             }
-
+            
             public static UnsupportedOperationException dataTypeOfVariableValueNotSupported(final Object val) {
-                return new UnsupportedOperationException(String.format("Graph variable value [%s] is of type %s is not supported", val, val.getClass()));
+            	return dataTypeOfVariableValueNotSupported(val, null);
+            }
+
+            public static UnsupportedOperationException dataTypeOfVariableValueNotSupported(final Object val, final Exception rootCause) {
+                return new UnsupportedOperationException(String.format("Graph variable value [%s] is of type %s is not supported", val, val.getClass()), rootCause);
             }
         }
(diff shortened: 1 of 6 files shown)


spmallette:
it's nice that you have "root cause" as optional to avoid breaking change.  in that way, it is optional for providers to pass a cause and if they just want a vanilla exception they can use it that way.  I guess that opens up the opportunity for there to be some inconsistency in the exceptions as one provider might choose to include cause and the other might not have a cause in the first place to throw....that might not matter i suppose.

I also wonder if all exceptions should have this option.  It's a bigger change, but perhaps all exception methods should be deprecated and instead return some sort of `ExceptionBuilder<T extends Throwable>`.  That way, the provider could do something like:

```text
throw Property.Exceptions.buildDataTypeOfPropertyValueNotSupported(value).cause(root).create();
```

or 

```text
throw Property.Exceptions.buildDataTypeOfPropertyValueNotSupported(value).create();
```

as the need was required.  Then this would be consistently applied.  Of course, perhaps this specific issue is isolated as a problem and we aren't hiding "native" causes and messages elsewhere.  If that's the case, then this pull request is the "simple" thing to do.

velo:
I think is a nice idea, but that would have way bigger impact them what I did.

I can do it anyway, just say a word.

spmallette:
I gave this some thought - your change solves a pretty specific problem so perhaps we should go with it for 3.1.1-incubating.  The change I'd suggested is pretty big and probably should be discussed on the dev list before we dive into it. I do think it's a nice idea though but perhaps we reserve that for 3.2.0-incubating (which has not yet started development).  

In the mean time we can add changes like the one you supplied as needed.  Hopefully we won't need too many more before 3.2.0-incubating as it will just more methods/exceptions for us to deprecate later.

VOTE: +1

velo:
FWIW, I rebased this so travis would pass

PommeVerte:
looks good to me.
VOTE: +1

okram:
VOTE +1. @spmallette -- will you handle the merge and any documentation on backwards compatibility? Also, if this is going to break for 3.1.1, then we should delay before merging when we get 3.2.0 set up.

spmallette:
This change doesn't introduce breaks.  @velo  kept the old method signatures and provided a couple overloads.  providers can choose to use the new methods or not.

PR #150, opened 2015-11-20
TINKERPOP3-984

velo:
Made possible to use GraphProvider to convert ids into Strings

The idea is to prevent this:
````
Script1.groovy: 1: unexpected char: '#' @ line 1, column 5.
   g.V(#11:0).next()
````

### DIFF
--- a/gremlin-groovy-test/src/main/groovy/org/apache/tinkerpop/gremlin/groovy/loaders/SugarLoaderTest.groovy
+++ b/gremlin-groovy-test/src/main/groovy/org/apache/tinkerpop/gremlin/groovy/loaders/SugarLoaderTest.groovy
@@ -78,11 +78,11 @@ class SugarLoaderTest extends AbstractGremlinTest {
         assertEquals(6, g.V.count.next())
         assertEquals(6, g.V.out.count.next())
         assertEquals(6, g.V.out.name.count.next())
-        assertEquals(2, g.V(1).out.out.name.count.next());
-        g.V(1).next().name = 'okram'
-        assertEquals('okram', g.V(1).next().name);
-        g.V(1).next()['name'] = 'marko a. rodriguez'
-        assertEquals(["okram", "marko a. rodriguez"] as Set, g.V(1).values('name').toSet());
+        assertEquals(2, g.V(graphProvider.convertId(1, Vertex.class)).out.out.name.count.next());
+        g.V(graphProvider.convertId(1, Vertex.class)).next().name = 'okram'
+        assertEquals('okram', g.V(graphProvider.convertId(1, Vertex.class)).next().name);
+        g.V(graphProvider.convertId(1, Vertex.class)).next()['name'] = 'marko a. rodriguez'
+        assertEquals(["okram", "marko a. rodriguez"] as Set, g.V(graphProvider.convertId(1, Vertex.class)).values('name').toSet());
         assertEquals(29, g.V.age.is(eq(29)).next())
     }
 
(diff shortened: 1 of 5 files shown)


okram:
This is not how this should be done. There is a bug in the `GremlinGroovyScriptEngineOverGraphTest.java` test. Please see how we do it in the, for example, `GroovyGroupTest.groovy` in `gremlin-groovy-test`. Thus, instead of changing `GraphProvider`, we need to change `GremlinGroovyScriptEngineOverGraphTest` to use the standard model.

velo:
ok, seems to be a good alternative... I will send another PR in a few

velo:
question, shall I do this for ALL engine.eval() or just for the ones that pass ID as parameters?

okram:
Now that I think it through more -- you can probably just do a `Bindings` and then `engine.eval("g.V(id).blah.blah",bindings)` where `bindings.set("id",convertToId("marko"))`.

velo:
Yeap, that is what I did, just re-running tests and as soon I get a green light I will update this PR

spmallette:
I wondered why Titan never had problems with the Groovy environment tests - doesn't look like they are implemented - i guess that explains it.

velo:
Travis is trying to run rvm... I guess it is not that reliable right now...

okram:
If you are happy with your PR @velo, we can all do our respective reviews and VOTE. Just please tell us when you are ready.

velo:
@okram feel free to call voters upon this PR

okram:
I reviewed the code and ran `mvn clean install -DincludeNeo4j` and `BUILD SUCCESSFUL`. Yes, @velo , we simply had a set of cases that we not using bindings in an intelligent way. Thanks for fixing them up.

When we merge this, I will make a TinkerGraph provider/test that uses an object id that does not have a nice `toString()` so we make sure this doesn't pop up again.

VOTE +1.

spmallette:
I merged this locally and did a `mvn clean install -DincludeNeo4j` and `mvn clean install -DskipIntegrationTests=false -pl tinkergraph-gremlin` to at least exercise the integration tests for groovy.  Of course, neither of these is a perfect tests here because neither Neo4j nor TinkerGraph use complex objects for identifiers.  If OrientDB is good with this change then this should be good to go.

VOTE: +1

okram:
At @spmallette --- Once we merge this, I will make a `TinkerGraphUUIDProvider` that will allow us to make sure all our tests are good for OrientDB (and others) moving forward. This way, we will be confident OrientDB will work.

velo:
Nice @okram

dkuppitz:
Tests pass and code looks good. However, I wonder why you've used `graphProvider.convertId(1, Vertex.class)` instead of `convertToVertexId("marko")`. But I guess I'm missing something here, since @okram mentioned that he reviewed the code and he's fine with it. Hence:

VOTE: +1

velo:
@dkuppitz the original code was ````g.V(1)```` them I just used graphProvider to turn this into a proper ID

dkuppitz:
I saw that, but I still think that `convertToVertexId("marko")` is the cleaner (cleanest?) way to do it, especially if the test uses the modern graph.

okram:
The problem is that these are not MODERN tests. When I merge, I'll fiddle to see if we can make these full modern. Else, I think we have to go with what @velo has. I'll handle this tomorrow and report back.

dkuppitz:
> The problem is that these are not MODERN tests.

[It looks pretty MODERN](https://github.com/velo/incubator-tinkerpop/blob/TINKERPOP3-984/gremlin-groovy-test/src/main/groovy/org/apache/tinkerpop/gremlin/groovy/loaders/SugarLoaderTest.groovy#L75) to me.

okram:
Ah, you are right. Good find. Okay, tomorrow on merge, I'll clean it up along with adding `TinkerGraphUUIDProvider` so we don't have to rely on OrientDB to test this.

PR #152, opened 2015-11-21
Fixed feature check in shouldPersistDataOnClose()

ellitron:
GraphTests.shouldPersistDataOnClose checked that *edges* supported String type properties before adding String type properties to *vertices*. Changed this check to be consistent with the operation being performed.

Additional Notes:
1) This unit test adds properties to vertices, but does not *explicitly* make the check for Graph.Features.VertexFeatures.FEATURE_ADD_PROPERTY support in its feature requirements. The feature requirement set SIMPLE does not include Graph.Features.VertexFeatures.FEATURE_ADD_PROPERTY, but rather includes the feature Graph.Features.VertexPropertyFeatures.FEATURE_ADD_PROPERTY, which is a check for being able to add meta properties. Therefore, since meta properties cannot be supported without basic add property support for vertices, the check for adding properties to vertices is *implicit*, not explicit. 
2) Although Graph.Features.VertexPropertyFeatures.FEATURE_ADD_PROPERTY is included in the feature set requirements for this unit test, this unit test does not actually depend on meta property support. This unit test will not run for graphs that support persistence, but do not support meta properties.

### DIFF
--- a/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/structure/GraphTest.java
+++ b/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/structure/GraphTest.java
@@ -1205,7 +1205,7 @@ public class GraphTest extends AbstractGremlinTest {
 
         final Vertex v = graph.addVertex();
         final Vertex u = graph.addVertex();
-        if (graph.features().edge().properties().supportsStringValues()) {
+        if (graph.features().vertex().properties().supportsStringValues()) {
             v.property(VertexProperty.Cardinality.single, "name", "marko");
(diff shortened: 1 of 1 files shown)


spmallette:
going to merge this by CTR. it is a simple one liner fix.  

@ellitron hopefully you've been able to subcribe to the dev list at this point and can talk more about the tests there in this context you've raised.  I'd started a thread on this there:

https://pony-poc.apache.org/thread.html/Zn7fzlygehgxjc5

ellitron:
@spmallette Got subscribed, was confirmation was in my spam box... then in my attempt to reply to the thread it looks like I may have created a new thread entirely.... rats

PR #154, opened 2015-11-23
TINKERPOP3-988 Change SparkGraphComputer executor

dalaro:
https://issues.apache.org/jira/browse/TINKERPOP3-988 describes the problem and my justification of this change.

ForkJoinPool.commonPool shouldn't be used here, but I'm also not certain that creating one single-threaded executor per `submit` call is a reasonable alternative.  My assumption is that `SparkGraphComputer.submit` is an operation with pretty heavyweight processing implications, so this isn't something you'd want to fire off ten thousand times a second, and the overhead of one thread per call isn't that bad.  I can rework it if you think this is not acceptable though.

### DIFF
--- a/spark-gremlin/src/main/java/org/apache/tinkerpop/gremlin/spark/process/computer/SparkGraphComputer.java
+++ b/spark-gremlin/src/main/java/org/apache/tinkerpop/gremlin/spark/process/computer/SparkGraphComputer.java
@@ -54,6 +54,8 @@ import org.apache.tinkerpop.gremlin.spark.structure.io.OutputRDD;
 import java.io.File;
 import java.io.IOException;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


okram:
I have no idea why this is good or bad but if @dalaro says it should be as he has it, I'm all for it. Also, to answer @dalaro -- no, this is not a heavily threaded method. This is batch processing and thus, you won't be "firing off ten thousand times a second."

@dalaro -- QUESTION: Should we do the same for `GiraphGraphComputer`? If so, can you update your PR for that as well. Also, what about `TinkerGraphComputer`?

I pulled into a local branch. Tested `SparkGraphComputerProcessIntegrationTest`. It passed fully. I also ran the `mvn clean install` to test other non-integration tests for Spark. They all passed as well.

VOTE +1.

spmallette:
hmmm - I can see where @dalaro is coming on this one though I didn't know that it would cause problems or whether it would cause problems elsewhere or just Spark.  

VOTE +1 - pending dan's answer on `Giraph/TinkerGraphComputer` which both have `ForkJoinPool` usage.

dalaro:
I think this problem also applies to GiraphGraphComputer.  Consider the following sequence of operations in SparkGraphComputer when `null != System.getSecurityManager()`.

* submit passes a task to CompletableFuture.supplyAsync (served by the forkjoin common pool)
* submitted task calls FileSystem.get
* FileSystem.get calls a bunch of Hadoop permission guts and eventually gets to UserGroupInformation.newLoginContext
* UserGroupInformation.newLoginContext calls setContextClassLoader on the current thread
* setContextClassLoader throws SecurityException if `null != System.getSecurityManager()` and the task is running in a common forkjoin pool thread, killing the task

The FileSystem.get call is visible in SGC's source.  The succeeding calls happen inside FileSystem.get and aren't visible in TP's source.  But GGC calls FileSystem.get with the same parameter signature as SGC, so they should both be affected by this problem.

I don't know whether the same problem applies to TinkerGraphComputer, but my guess is no, unless TinkerGraphComputer's internals touch the context classloader or something else that does, like Hadoop.

I'll rework the PR to apply to both SGC and GGC (but not TGC).

dkuppitz:
Full test suite passed for me.

VOTE: +1

okram:
We will merge this when @dalaro submits the same fix for `GiraphGraphComputer`.

dalaro:
I'll open a new PR.  Don't merge this one.  I extracted the common executor-related logic into a helper shared between the Giraph and Spark computers, but much more importantly, my first attempt at this leaks executors.  Nowhere does my patch call `shutdown` or `shutdownNow`.   The `newSingleThreadExecutor` factory method that I invoked internally sets the core-pool size to 1, so the idle thread could never die, even after its submission-related work is all done.  I resolved this in an incoming PR with a try-finally construct where the try block constructs the executor and the finally block invokes `shutdown`.

dalaro:
This is superseded by https://github.com/apache/incubator-tinkerpop/pull/164.

PR #158, opened 2015-11-27
TINKERPOP3-912 Improve the ability to embed Gremlin Server with Channelizer injection

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP3-912

I folded in a couple of changes related to general refactoring that were above and beyond the scope described in the JIRA ticket.  Most notably was the removal of more code related the deprecated "rebinding" term.  

I thought this ticket was going to be much bigger when i started, but it seemed to be less so for 3.1.1 anyway. Anyway, it is pretty deep in the guts of Gremlin Server, so reviewers should focus mostly on ensuring that there are no regressions as a result of this refactoring work.

I tested with:

```text
mvn clean install
```

VOTE: +1

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
@@ -310,7 +310,7 @@ public abstract class Client {
          */
         @Override
         public Client alias(String graphOrTraversalSource) {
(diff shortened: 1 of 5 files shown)


twilmes:
Tests & `mvn verify -DskipIntegrationTests=false -pl gremlin-server` all passed without issue.  Code updates looked straightforward.

VOTE +1

PommeVerte:
Integration tests pass for me. I ran the lower level driver test I have here and they all passed as well.

VOTE: +1

PR #162, opened 2015-12-01
TINKERPOP3-995 Add Authenticator.newSaslNegotiator(InetAddress)

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP3-995

This change is non-breaking despite the deprecation. Reviewers should focus on validating that there are no regressions:

```text
mvn clean install
```

VOTE: +1

### DIFF
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/auth/AllowAllAuthenticator.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/auth/AllowAllAuthenticator.java
@@ -18,6 +18,7 @@
  */
 package org.apache.tinkerpop.gremlin.server.auth;
 
+import java.net.InetAddress;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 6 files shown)


gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/auth/AllowAllAuthenticator.java
```diff
@@ -41,7 +42,16 @@ public AuthenticatedUser authenticate(final Map<String, String> credentials) thr
         return AuthenticatedUser.ANONYMOUS_USER;
     }
 
+    /**
+     * @deprecated As of release 3.1.1-incubating, replaced by {@link #newSaslNegotiator(InetAddress)}.
+     * @see <a href="https://issues.apache.org/jira/browse/TINKERPOP3-995">TINKERPOP3-995</a>
```
twilmes: This is about as tiny as it comes, but could the '3' be dropped to be consistent with the latest name update?  Can't remember if they were ever going to drop the redirect.  I'm sure there are others in the code so maybe instead, at some point a mass update could be made if we wanted.

twilmes:
All the tests passed for me and the doc updates & code look good.

VOTE +1

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/auth/AllowAllAuthenticator.java
spmallette: i'm not sure that we need to worry about the redirect dropping - i guess we could discuss a mass find/replace of that stuff. i was going to leave the old stuff as-is in my thinking this morning, but maybe we shouldn't. probably something to discuss on the list.

okram:
VOTE +1.

PR #163, opened 2015-12-01
TINKERPOP3-1012 generalize BLVP vertex id types

dalaro:
Change the type of two vertex ID variable declarations from Long to Object.  As mentioned in the JIRA, his should be reviewed.  I'm not sure this is the right change.


dkuppitz:
Yep, that's how it should be.

VOTE: +1

okram:
Easy enough. VOTE +1. 

@dkuppitz -- when we get the third VOTE, can you be responsible for merging to master/ and CHANGELOG stuffs?

dkuppitz:
Yep, will do.

spmallette:
VOTE: +1

PR #165, opened 2015-12-01
TINKERPOP3-1013: Traverser tags as a safer way of using path labels

okram:
https://issues.apache.org/jira/browse/TINKERPOP3-1013

We have had a hack in `MatchStep` this whole time where we use path labels to store branch history. This is not only semantically awkward (as labels are added that were not provided via `as()`), it also means we will not be able to do https://issues.apache.org/jira/browse/TINKERPOP3-736. Also, this will allow us to cleanly implement `IntersectStep` and `SymetricDifferenceStep` (down the road).

The implementation is really simple. `Traverser.getTags()`. This method returns a mutable set and thus, we can remove, add, size, contains, etc. tags associated with a traverser.

Ran `mvn clean install` and all is golden.

VOTE +1.


spmallette:
Ran `mvn clean install` - all good. 

VOTE +1

twilmes:
Tests passed and code looks good.

VOTE +1

PR #166, opened 2015-12-02
TINKERPOP3-860 Bindings applied to the PluginAcceptor should appear to Gremlin Server

spmallette:
Bindings from plugins were being pushed into the ScriptEngine context but were not being merged into the bindings used on eval.  They had to be merged together.  In doing this, I also found a bug where plugins could conflict with one another (i.e. the first plugin would install, then the second would come along an blow out what the previous one had done) - fixed that.  Added test cases to cover these scenarios.

I tested with:

```text
mvn clean install
mvn verify -DskipIntegrationTests=false -pl gremlin-groovy
```

and also tested Gremlin Server manually:

```text
gremlin> :remote connect tinkerpop.server conf/remote.yaml
==>Connected - localhost/127.0.0.1:8182
gremlin> :> hdfs.toString()
==>org.apache.hadoop.fs.LocalFileSystem@7154cf94
```

A manual test requires:

```text
bin/gremlin.sh -i org.apache.tinkerpop hadoop-gremlin 3.1.1-SNAPSHOT
```

and editing the gremlin server yaml file to include `tinkerpop.hadoop` in the "plugins" key:

```yaml
plugins:
  - tinkerpop.tinkergraph
  - tinkerpop.hadoop
```


spmallette:
i think the CI failure is related to this: 

https://issues.apache.org/jira/browse/TINKERPOP3-1006

and not my changes in particular.

spmallette:
Travis is happy now - votes anyone? :)

dkuppitz:
Code / configurations look good. Also the logs in Travis are still verbose enough to see what's going on. Nice.

VOTE: +1

okram:
VOTE +1

PR #167, opened 2015-12-02
TINKERPOP3-1006 Random error during builds: shouldReloadClassLoaderWhileDoingEvalInSeparateThread()

spmallette:
Adjust tests to be a bit more bulletproof with respect to thread timing. Not sure if i've fixed anything with this change as these tests have never failed for me on my system.  I would have CTR'd this but I'd like to see if there is any validation that this is working correctly now. I guess runs in Travis and those who have macs would be the best judge of whether or not this issue is resolved.

VOTE +1


okram:
I never had this test fail for me. Thus, your loosening only makes it LESS likely to fail for me :). 

VOTE +1.

PommeVerte:
This correctly builds here. I haven't had problems recently so I can't really tell if this is fixed or not. I suspect that travis will complain if it isn't. This loosens things a bit so VOTE: +1 from me.

spmallette:
Travis looks happy with this - so we'll assume fixed - until it isn't.  merging.

PR #169, opened 2015-12-03
TINKERPOP3-982: valuesDecr, valuesIncr, keysDecr, and valuesDecr is lame.

okram:
https://issues.apache.org/jira/browse/TINKERPOP3-982

I made it so the `Column` enum implements `Function<Object,Object>` and where `Column.keys` and `Column.values` project out the respective "column" for `Map`, `Map.Entry`, `Path`, etc. What is cool about this is that now we no longer need `SelectColumnStep` -- `select(keys) == map(keys)`. Of course, like `where(traversal) == filter(traversal)`, I left `select(column)` in `GraphTraversal` and don't plan to remove it. 

Given that this is a minor version push (3.1.1), I will NOT deprecate `Order.valueDecr`/etc. until 3.2.0 and have the following ticket to remind me.  https://issues.apache.org/jira/browse/TINKERPOP-1021

Finally, `mvn clean install` and a full `publish-docs.sh` are happy happy long time.

VOTE +1.


twilmes:
Looks good, tests pass.

VOTE +1

dkuppitz:
Nice one.

VOTE: +1

okram:
This was merged.

PR #170, opened 2015-12-04
Additions to hadoop-gremlin conf files

vtslab:
Hi Marko,

Following your request in:
https://groups.google.com/forum/#!topic/gremlin-users/aE1plgCMhTs
please find the pull request asked for. 

However, I feel there is more to it than just the properties files: 
 - gremlin.sh should accept HADOOP_HOME, SPARK_HOME and YARN_HOME env variables (in addition to HADOOP_GREMLIN_LIBS) and set the CLASS_PATH accordingly. E.g., see:
https://github.com/vtslab/incubator-tinkerpop/blob/3.1.0-hdp-2.3.2.0-2950/gremlin-console/bin/gremlinhdp.sh
 - user documentation could mention the spark-yarn option
 - a spark-yarn test case that can run on the build server

I could prepare an additional pull request for some of these issues.

Regards,    Marc

### DIFF
--- a/hadoop-gremlin/conf/hadoop-graphson.properties
+++ b/hadoop-gremlin/conf/hadoop-graphson.properties
@@ -28,14 +28,43 @@ gremlin.hadoop.jarsInDistributedCache=true
 # the vertex program to execute
 gremlin.vertexProgram=org.apache.tinkerpop.gremlin.process.computer.ranking.pagerank.PageRankVertexProgram
 
-####################################
-# SparkGraphComputer Configuration #
-####################################
+############################################
+# SparkGraphComputer Configuration (Local) #
+############################################
 spark.master=local[4]
 
-#####################################
-# GiraphGraphComputer Configuration #
-#####################################
+############################################
+# SparkGraphComputer Configuration (Yarn) #
+############################################
+#spark.master=yarn-client
+# app name in the cluster's history server
+#spark.app.name=TinkerPop
+# faster startup if many cluster users ask for the default 4040 port
+#spark.ui.port=4050
+# needed if the default JAVA_HOME on the cluster is not jdk1.8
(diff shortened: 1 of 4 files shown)


spmallette:
Thanks for submitting this - it will be reviewed/merged closer to the release of 3.1.1 as there is some other review of the hadoop configs that package with the distribution that needs to be done first.

spmallette:
@okram are we now looking at this one for 3.1.2 at this point?

okram:
Yes. This is more for @dkuppitz as will get getting GiraphGraphComputer working on the Blades and once he gets it solid, he will be able to go through each recommended property and decide whether what we need and not need.

spmallette:
@okram @dkuppitz is there anything we want to do with this one? work it into documentation somehow?

dkuppitz:
One day I tried to use `yarn-client` and got all kind of weird errors. I will try again.

hadoop-gremlin/conf/hadoop-graphson.properties
```diff
@@ -28,14 +28,43 @@ gremlin.hadoop.jarsInDistributedCache=true
...
+#spark.executor.extraJavaOptions=-Dhdp.version=2.3.2.0-2950
+# optional parameters for tuning your TinkerPop application
+#spark.executor.instances=10
+#spark.executor.memory=4g
+#spark.executor.cores=2
+# needed on a secure cluster to allow for multiple computations
+# on a single Kerberos ticket
+#gremlin.spark.persistContext=true
+
+# Below copy all properties from your spark-client's spark-defaults.conf
```
dkuppitz: Is that necessary? Recommended? The purpose of `*-defaults.conf` files is to set up defaults which you don't want to specify somewhere else, no?

spmallette:
It doesn't look as though we ever had much success with tests on this and the branch is now conflicted. I'm going to close this PR in the interest of tidying up.  @vtslab if you think there's something that we should not be ignoring here, please feel free to submit a fresh PR that we can evaluate. Thanks for your time.

PR #174, opened 2015-12-09
TINKERPOP-1035 Improved consistency of transaction management in gremlin server.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1035

Sessionless requests now have special handling for `GraphTraversal` return results.  Gremlin Server now force iterates those results (when they include `Mutating` steps) to a `List`.  It then commits and then iterates that `List`. For all other results the commit occurs in the same spot, but doesn't force iterate which allows results to stream as they always have.

Tested manually and with:

```text
mvn clean install
```

VOTE: +1


twilmes:
Tests pass and the doc updates were good.  I think this will help clarify things for folks.

VOTE +1

okram:
VOTE +1.

PR #176, opened 2015-12-09
TINKERPOP-930 Added a strictTransactionManagement setting to Gremlin Server.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-930

Basically forces the user to pass the aliases argument.  Gremlin Server then uses that to determine the scope of the transactions to close.  This setting is false by default so as to be backward compatible.

Tested with:

```text
mvn clean install
mvn verify -DskipIntegrationTests=false -DincludeNeo4j -pl gremlin-server
```

VOTE +1


okram:
VOTE +1.

dkuppitz:
* `mvn clean install` worked
* integration tests worked

VOTE: +1

PR #178, opened 2015-12-11
TINKERPOP-954 Consistent test directory usage

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-954

Got all hard-coded directories in tests using refined methods in `TestHelper`.  That basically gets everything behind the `/target/test-case-data` directory in a consistent fashion.  I also went a step further and used the class name passed to `TestHelper` (formerly just used to orient the `/target` directory) to include a subdirectory structure which then yields `/target/test-case-data/ClassName` so that it's easy to find the test data output in case it needs to be reviewed.

Ran the full test suite with integration tests and all passed.

VOTE +1


dkuppitz:
* `mvn clean install`: worked
* integration tests: worked

VOTE: +1

twilmes:
Yep, tests are good here.

VOTE: +1

PR #179, opened 2015-12-11
TINKERPOP-891 Refactored the sandboxing abstractions for Gremlin Server

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-891

Deprecated the `SandboxExtension` and replaced it with the `AbstractSandboxExtension`.  Took existing implementations and had them extend from the new `AbstractSandboxExtension`.  Added a new "useful" implementation called `FileSandboxExtension` which gets its white list configuration from a file.  Updated appropriate docs and wrote a number of tests for the new `FileSandboxExtension`.

Both unit and integration tests pass.  Also tested manually by configuring Gremlin Server to use the different sandboxes.  

VOTE: +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,8 @@ TinkerPop 3.1.1 (NOT OFFICIALLY RELEASED YET)
 * Fixed a `SparkGraphComputer` sorting bug in MapReduce that occurred when there was more than one partition.
 * Fixed a `NullPointerException` bug in `PeerPressureVertexProgram` that occurred when an adjacency traversal was not provided.
 * Improved Transaction Management consistency in Gremlin Server.
+* Added `FileSandboxExtension` which takes a configuration file to white list methods and classes that can be used in `ScriptEngine` execution.
+* Deprecated `SandboxExtension` and `SimpleSandboxExtension` in favor of `AbstractSandboxExtension` which provides better abstractions for those writing sandboxes.
 * Fixed a long standing "view merge" issue requiring `reduceByKey()` on input data to Spark. It is no longer required.
 * Added `Spark` static object to allow "file system" control of persisted RDDs in Spark.
 * Added a Spark "job server" to ensure that persisted RDDs are not garbage collected by Spark.
--- a/docs/src/reference/gremlin-applications.asciidoc
+++ b/docs/src/reference/gremlin-applications.asciidoc
@@ -940,6 +940,10 @@ this documentation.  Consult the latest link:http://docs.groovy-lang.org/latest/
 for information on the differences. It is important to understand the impact that these configuration will have on
 submitted scripts before enabling this feature.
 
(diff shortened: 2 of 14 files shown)


twilmes:
Looks good.

VOTE: +1

dkuppitz:
* `mvn clean install`: worked
* integration tests: worked

VOTE: +1

dpitera:
@spmallette I am curious why this Pull Request got rid of the `methodBlackList`? I ask because I want to extend the `FileSandboxExtension` here to support blacklisting-- would you prefer I did the extension at that level, created another `FileSandboxWithBlacklistingExtension`, or did the extension at the `AbstractSandboxExtension`?

spmallette:
I don't see a reference to `methodBlackList` in this PR, but if we were to just reduce the question to why do we have whitelisting and no blacklisting then I think I could probably answer that.  I'd rather not support blacklisting in TinkerPop, as it just seems to lead people into thinking they have a secure solution when they soon learn that they'd forgotten yet another harmful entry to blacklist. 

I think that the whitelist works really well in TinkerPop, because the base list of classes required to execute Gremlin is small (and really that's all we care about from TinkerPop's perspective). Whitelisting tends to work best in cases like this as it assumes everything is bad except for this small, easy to maintain list.  Since whitelisting fits this situation so well, a blacklist feels a bit useless - extra code without purpose which we try to avoid. 

Obviously, we do have the `SimpleSandboxExtension` which does some basic blacklisting but it's mostly for demonstration and for basic protection from the worst of the worst `System.exit()`. 

Does that make sense?

dpitera:
> Whitelisting tends to work best in cases like this as it assumes everything is bad except for this small, easy to maintain list.

Agreed. Which is what leads me to find myself in a situation where even things like `"^java\\.lang\\.String"` must be whitelisted to be called. 

However....["^java\\.lang\\.String#getBoolean\\("](http://docs.oracle.com/javase/6/docs/api/java/lang/Boolean.html#getBoolean(java.lang.String)) must be blacklisted because it leaks implementations details about the underlying System.

I believe this is a prime example for situations where the best filter is something like:
`!methodBlackList.any { descriptor =~ it } && methodWhiteList.any { descriptor =~ it }`.

Would you agree?

> I don't see a reference to methodBlackList in this PR

This is because the methodBlackList of which I speak is part of the deprecated sadnbox extension classes

spmallette:
That's a reasonable example. Could someone not just "blacklist" by whitelisting though? I just mean that the whitelisting system is all regex based. You could add negation to the whitelist and get rid of your troublesome method:

```text
java\.lang\.Boolean\#(?!getBoolean\(String\)).*
```

That allows everything else on `Boolean` except `getBoolean(String)`.

dpitera:
I will have to investigate what you've just claimed, but if this is true, then yes you are right... I'll get back to you when I've confirmed. Thanks for the suggestion!

PR #180, opened 2015-12-16
TINKERPOP-1045 Validate bindings on the server to ensure that they were String keys

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1045.

Also, fixed a problem where error messages were not always propagating back to the client (usually around validation).  They were being sent back in the result field rather than in the statusMessage field.

Tested with: 

```text
mvn clean install && mvn verify -DskipIntegrationTests=false -pl gremlin-server
```
and did manual testing as well with the console - see the JIRA ticket for console output.


dkuppitz:
* `mvn clean install`: worked
* integration tests: worked

VOTE: +1

ShaunakDas88:
Run the following on this branch, and still get a hang on the client-side:

gremlin> client.submit('graph.toString()').all().get()
==>result{object=tinkergraph[vertices:0 edges:0] class=java.lang.String}
gremlin> script = "for(int i = 0; i < 10000000; i++){ graph.addVertex('label', i) }"
==>for(int i = 0; i < 10000000; i++){ graph.addVertex('label', i) }
gremlin> client.submit(script).all().get()

spmallette:
@ShaunakDas88 ok - so, just to be clear, do you mean that OOME is not hanging the client?

ShaunakDas88:
Yes, OOME does not hang the client anymore. Sorry for the confusion, it looks as though it was hanging, but the message just took some time to get relayed from the server.

twilmes:
`mvn clean install` and integration tests are good.  Code updates look good.

VOTE: +1

PR #181, opened 2015-12-16
TINKERPOP-1026 BVLP should store vertex IDs as String

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1026

BLVP (in particular `IncrementalBulkLoader`) now stores the `toString()` version of vertex IDs in the temporary `bulkLoader.vertex.id` property. I also added another `BulkLoader` implementation (`OneTimeBulkLoader`), to a) be able to add more test cases and b) verify that the `toString()` update doesn't interfere with something else.

Successfully ran `mvn clean install` and integration tests.

VOTE: +1

### DIFF
--- a/docs/src/reference/the-graphcomputer.asciidoc
+++ b/docs/src/reference/the-graphcomputer.asciidoc
@@ -338,7 +338,7 @@ writeGraphConf.setProperty("gremlin.tinkergraph.graphFormat", "gryo")
 writeGraphConf.setProperty("gremlin.tinkergraph.graphLocation", "/tmp/tinkergraph.kryo")
 modern = TinkerFactory.createModern()
 blvp = BulkLoaderVertexProgram.build().
-           keepOriginalIds(false).
+           bulkLoader(OneTimeBulkLoader).
            writeGraph(writeGraphConf).create(modern)
 modern.compute().workers(1).program(blvp).submit().get()
 graph = GraphFactory.open(writeGraphConf)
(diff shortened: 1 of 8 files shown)


spmallette:
I assume that `OneTimeBulkLoader` is something users might use?  or was it really just for those two internal use cases you listed there?  If it's just internal, perhaps you shouldn't make the class visible to people?  

or if it is something that users might use, do you need to update the reference docs to discuss how it is used?  maybe some javadocs is enough? both?

dkuppitz:
Updated.

* added notes in the reference docs
* made one BLVP sample code snippet use the new `OneTimeBulkLoader`
* verified that `process-docs.sh` still works
* added javadocs for `OneTimeBulkLoader`

spmallette:
looks like you angered the travis gods. doesn't look like the error is related to your changes though.

dkuppitz:
Hm, worked for me locally. Can we somehow retry Travis builds w/o doing another commit?

spmallette:
Maybe push an empty commit? 

```text
git commit --allow-empty
```

docs/src/reference/the-graphcomputer.asciidoc
```diff
@@ -364,10 +364,11 @@ It's recommended to tune this property for the target graph and not use the defa
 |`writeGraph(String)` | Sets the path to a `GraphFactory` compatible configuration file for the target graph. | _none_
 |========================================
 
-NOTE: `BulkLoaderVertexProgram` comes with a default `BulkLoader` implementation, namely `IncrementalBulkLoader`. It
-will work for the most use-cases, but has one limitation though: It doesn't support multi-valued properties.
-`IncrementalBulkLoader` will handle every property as a single-valued property. A custom `BulkLoader` implementation
-has to be used if the default behavior is not sufficient.
+NOTE: `BulkLoaderVertexProgram` uses the `IncrementalBulkLoader` by default. The other option is the `OneTimeBulkLoader`,
+which doesn't store any temporary IDs in the `writeGraph` and thus should only be used for initial bulk loads. Both
+implementations should cover the mahority of use-cases, but have a limitation though: They don't support multi-valued
```
spmallette: spelling mistake - should be "majority" :smile:

spmallette:
`mvn clean install` is good to go and code/docs look good:

VOTE +1

twilmes:
Tests and code look good.

VOTE +1

PR #183, opened 2015-12-16
TINKERPOP-1051 Add note in best practice docs about gremlin server heap setting

jeromatron:


spmallette:
Hate to nit/pick but "-xmx" should be "-Xmx" and could you change uses of "max" to "maximum" to avoid the abbreviation.  

Hopefully, travis will be happy with his next build of this PR - seemed like the failure has nothing to do with the content of your PR.

jeromatron:
Bah you're right.  I fixed that typo with this follow-up commit.

jeromatron:
Changed max to maximum as well.

PR #184, opened 2015-12-22
TINKERPOP-983 - Provide a way to track open Graph instances in tests

n-tran:
Doing a test PR to see if Travis will build, there might be something wrong with my local environment. If Travis doesn't pass, I will rescind my PR.

Refactor tests to use member variables for `graphProvider`
Added Java docs

https://issues.apache.org/jira/browse/TINKERPOP-983

### DIFF
--- a/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/AbstractGremlinTest.java
+++ b/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/AbstractGremlinTest.java
@@ -133,8 +133,14 @@ public abstract class AbstractGremlinTest {
     public void tearDown() throws Exception {
         if (null != graphProvider) {
             graphProvider.clear(graph, config);
+
+            // All GraphProvider objects should be an instance of ManagedGraphProvider, as this is handled by GraphManager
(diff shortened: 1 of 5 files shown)


spmallette:
Not sure why travis didn't pick this up for build, but I pulled the changes and it failed for me too. It's a subtle problem, but the fix is to add this method to your `ManagedGraphProvider`:

```java
@Override
public String getWorkingDirectory() {
    return innerGraphProvider.getWorkingDirectory();
}
```

Without that, HDFS gets the wrong directory and the assert fails. Please try that and see if it fixes your problem with the build.  If all good, please update your PR and we can review in full.

spmallette:
All good with:

```text
mvn clean install -DincludeNeo4j && mvn verify -DskipIntegrationTests=false -DincludeNeo4j
```

The approach makes sense to me.  Not sure how else it could be done short of placing the burden on providers to ensure that open `Graph` implementations were tracked. 

VOTE +1

okram:
VOTE +1.

twilmes:
Yeah, looks good.

VOTE: +1

PR #186, opened 2015-12-24
Osgify

gdelafosse:
The goal of this request is to start bundleizing tinkerpop.
In this request, I have bundleized the following modules : 
- gremlin-core
- gremlin-groovy
- tinkerpop-gremlin
  They are enough to start playing with TinkerPop in an OSGI environment.

To be the less intrusive as possible, I had to create a new module called gremlin-osgi-deps which aggregate and export packages from tinkerpop dependencies which are not bundles for the moment.
As a result, the only impact my request has, is to add entries in the MANIFESTs to make those modules valid bundles.

To illustrate the usage of those bundles, I'm developping a Wisdom-Framework extension (https://github.com/gdelafosse/wisdom-tinkerpop).


okram:
I, personally, do not want OSGi support in TinkerPop. Over the years, we never have anyone that maintains such pushes, no one understands its use, there are no test cases, ... Is it possible for this to be maintained by a separate community project (e.g. tinkerpop-osgi-deps) and NOT part of Apache TinkerPop?

gdelafosse:
I understand.

I propose I remove tinkerpop-osgi-deps from your repository, and I create one repository into my group for it. My plan is to contribute and make those non-bundled dependencies valid OSGI bundles, so that we no longer need tinkerpop-osgi-deps.

But, what about the "bundle" packaging required to make tinkerpop jars OSGI bundles? Do you accept to keep them? If you are afraid with maintaining the maven-bundle-plugin configuration, I propose I check it regularly. I need them for my applications that use [wisdom-framework](https://github.com/wisdom-framework/wisdom) and TinkerPop, so I'll be vigilant...

gdelafosse:
I removed gremlin-osgi-deps from the pull request and created this project in my repositories : https://github.com/gdelafosse/gremlin-osgi-deps.

Would you accept to package your projects as bundles?

spmallette:
I don't know much about OSGi to properly review/vote on this, so at the moment, I'm +0.  

I think I find this situation similar to trying to accept a PR where you don't have a clue as to what the code in the PR is doing (you can't say if it's "right" or "wrong").  I also know from dealing with past pull requests on OSGi (various TinkerPop 2.x projects and Titan) that just altering the pom.xml is only the start of OSGi support - it's never been clear to me what else would need to be done for "full" support, how we would maintain that, how we would keep from breaking that, how it would affect the project as a whole, how it would change design decisions, etc.  Answers in this area have been somewhat vague and full of variety going from person to person.  

btw, i've known "bundles" to behave weirdly with Grape sometimes (grape not finding dependencies).  it would be very bad for the plugin system if introducing a "bundle" packaging type messed up things up there.  i think that would have to be tested very carefully if we made this change.

Can other committers who are newer to the TinkerPop game comment on this issue? @pluradj or @twilmes ?  do you folks have any experience here?

gdelafosse:
The maven-bundle-plugin just modifies the MANIFEST file with its own entries. I don't think it impacts any other frameworks or tools.

pluradj:
I worked briefly with OSGi several years ago, but I don't remember too much about it.

The code changes remaining in this PR seem minimal, but like the others have mentioned, there aren't any documentation or tests. If all that this PR does is modify the manifest file, maybe all we need is some committer documentation to verify what to look for and what needs to be updated if packages/dependencies change?

IMO Grapes behaves weirdly *in general* so if OSGi makes Grapes unstable, that would not be a good story.

spmallette:
> IMO Grapes behaves weirdly in general so if OSGi makes Grapes unstable, that would not be a good story.

haha and agreed

gdelafosse:
Hi all,
concerning Grapes, it doesn't seem to use the jar MANIFEST file, so I don't think it will be disturbed by my change.

Concenring the TinkerPop dependencies, I've created a pull request for each of them to turn them into OSGI bundle, so that gremlin-osgi-deps won't be required anymore.
See
https://github.com/svenkubiak/jBCrypt/pull/4
https://github.com/jcabi/jcabi-log/pull/69
https://github.com/jcabi/jcabi-aspects/pull/201
https://github.com/jcabi/jcabi-manifests/pull/29
https://github.com/carrotsearch/hppc/pull/5
https://github.com/javatuples/javatuples/pull/4

Concerning testing, I'm implementing an it test for each impacted sub module to check that they are valid bundles installable into an OSGI container thanks to PaxExam. I will join those tests to this pull request soon.

Concerning the documentation, where can I document how to install TinkerPop into an OSGI container?

okram:
In terms of documentation, perhaps in the "Gremlin Applications" section at the end with a new subsection would be best:

http://tinkerpop.apache.org/docs/3.1.0-incubating/#gremlin-applications

Or is this more of an "Implementations" thing? Again, I'm don't understand OSGi so I don't really what this provides, so its more up to you to decide where in the docs this should go. ??

So before we review this again for PR merge, we are waiting for the aforementioned projects to turn their respective projects in to OSGi bundles?

gdelafosse:
The reason why I want to osgify your projects is because I'm building graph applications with the wisdom-framework/wisdom which is a web-stack based on OSGI.
Anyway, it could be interesting for anybody that wants to integrate TinkerPop artifacts into their OSGI container.
You can consider OSGI as a Java application container, where application can be installed and removed dynamically (without restarting the JVM).

So to my mind it's more an "implementation" thing.

Concerning the dependencies, my pull requests are being accepted, but some projects are less reactive than other. So, waiting for all to be accepted, released and published into maven central, I will still maintain gdelafosse/gremlin-osgi-deps. In my documentation I'll explain how to install TinkerPop and the required dependencies into an OSGI container.

To conclude, I think you can review the PR as soon as I implemented the tests.

gremlin-core/pom.xml
```diff
@@ -37,7 +41,7 @@ limitations under the License.
         <dependency>
             <groupId>org.yaml</groupId>
             <artifactId>snakeyaml</artifactId>
-            <version>1.15</version>
+            <version>${snakeyaml.version}</version>
```
spmallette: any reason you turned these into properties? doesn't seem necessary as they aren't really used anywhere outside of gremlin-core.  i think we tend to only use properties in the root pom.xml when there's a need to keep a value consistent that might be referenced in multiple places.

gremlin-shaded/pom.xml
```diff
@@ -24,6 +24,67 @@ limitations under the License.
     </parent>
     <artifactId>gremlin-shaded</artifactId>
     <name>Apache TinkerPop :: Gremlin Shaded</name>
+    <packaging>bundle</packaging>
+    <properties>
+        <!--
+        As this jar is shaded, the maven-bundle-plugin is not able to process exported and imported packages from generated classes.
+        So we must explicitely tell which packages are exported and imported by this bundle to the maven-bundle-plugin.
+        -->
+        <osgi.import>
```
spmallette: how do we know what should be added to `<osgi.import>` and `<osgi.export>`?  this goes back to my earlier question of how to "maintain" this stuff over time?

pom.xml
```diff
@@ -147,6 +147,19 @@ limitations under the License.
         <log4j-silent.properties>file:target/test-classes/log4j-silent.properties</log4j-silent.properties>
 
         <muteTestLogs>false</muteTestLogs>
+
+        <!--
+        Configuration properties for the OSGi maven-bundle-plugin
+        -->
+        <osgi.embed></osgi.embed>
```
spmallette: again, not sure i'm seeing the need for properties here - any reason you don't just add this stuff directly to `<instructions>` of the `<maven-bundle-plugin>`

spmallette:
> Concerning testing, I'm implementing an it test for each impacted sub module to check that they are valid bundles installable into an OSGI container thanks to PaxExam. I will join those tests to this pull request soon.

That's interesting - you would be the first person to ever offer to supply tests for OSGi.  

Another addition to this PR that I think is important would be some documentation for TinkerPop Committers that serves as some reminder as to how to "maintain" these configuration and their meaning.  To me this is more internal developer docs - something that would go in this section somewhere:

https://github.com/apache/incubator-tinkerpop/blob/ef40dce314c1f9ee48661227d992beef8cad64d0/docs/src/dev/developer/contributing.asciidoc#for-committers

From my limited understanding of OSGi, I would say user docs should be updated as part of "Gremlin Applications" section of the user docs. I guess there would be a new sub-section there of some sort.

gremlin-core/pom.xml
gdelafosse: This version was initially used in the module gremlin-osgi-deps that I just removed (see https://github.com/gdelafosse/incubator-tinkerpop/commit/f29d01339514837ae4cc98240bc70f53b754aff7).
So indeed it's no more usefull here, but I usually centralize all the version in the dependencyManagment in the parent pom to see them all at a glance.
I'll put it back to its initial location as it's not your rule.

gremlin-shaded/pom.xml
gdelafosse: Gremlin Shaded was not the most easy to bundleize, mainly because of the package relocation processed while packaging this jar. So I had to tell it explicitely what to do.
To list those instructions I checked the imported and exported packages of my gremlin-osgi-deps project, and put the instructions corresponding to the gremlin-shaded packages directly in this project. As a result, gremlin-shaded is a correct OSGI bundle, and I'm able to get rid of the gremlin-osgi-deps (as soon as third parts dependencies are OSGI bundles too).

To answer your question, the integration test I'm implementing will detect if you miss to update to the osgi.import and osgi.export.

Sounds tricky but actually this gremlin-shaded project sounds tricky to me too. Could you explain why you have to shade those dependencies instead of using them directly?

spmallette: > To list those instructions I checked the imported and exported packages of my gremlin-osgi-deps project, and put the instructions corresponding to the gremlin-shaded packages directly in this project.

You say you check the imported/exported packages of your gremlin-osgi-deps project and got the information from there.  But when that project "goes away", how would you know what should be added to `<osgi.import>` and `<osgi.export>`.  In other words, how did gremlin-osgi-deps know what to add?

> Could you explain why you have to shade those dependencies instead of using them directly?

Shading is a technique for avoiding dependency conflict and it is especially useful for packages that are widely used and are known to introduce incompatible breaks between versions.  By shading, we can bind TinkerPop to any version of a library we want and upgrade it at our own pace without worrying about conflict with other system we might depend on or what users might be using.

gdelafosse: I see. That solves the typical trouble that OSGI solves. In OSGI you can load several versions of the same third library and each bundle tells which version it wants to use.
I understand your concern about this huge list of packages to maintain. I thought that the unit test will detect and tell if some packages must be added to the <osgi.import> and <osgi.export>.
But I'm struggling withe maven-shade-plugin and maven-bundle-plugin to make this bundle automatically without having to specify the list of packages to import and to export. I'm trying what's suggested in this [ticket](https://issues.apache.org/jira/browse/FELIX-1184) without any success. I don't give up...

gdelafosse:
You should enjoy this update : 
- I simplify the gremlin-shaded pom thanks to the solution explained [here](https://issues.apache.org/jira/browse/FELIX-1184)
The idea is to call sequentially the maven-shade-plugin to shade the dependencies, the maven-jar-plugin to unpack the shaded jar, and eventually the maven-bundle-plugin to compute the OSGI MANIFEST entries and repack the jar.
- I implemented an integration test to test that germlin-core is a valid OSGI bundle.
Unfortunately I don't manage to implement the same test for tinkerpop-gremlin because when I add my test dependencies it breaks the existing test (hamcrest Matchers classes are not found). I don't understand for the moment how the tinkerpop-gremlin tests work...

ruizale:
Hi, I just have a rapid question.
it's this pull request going to be merged at some point?
I'm just checking because I already use tinkerpop version 2.6 inside an OSGi application and I would like to upgrade and use latest version

PR #187, opened 2015-12-28
Update README.asciidoc with correct hyperlinks

MarkHoekstra:


spmallette:
thanks for noticing/fixing that.

PR #191, opened 2016-01-04
TINKERPOP-938 Add a "clear SNAPSHOT jars" section to the process-docs.sh.

dkuppitz:
Clear the local Grapes directory and HDFS in `bin/process-docs.sh`.
Also added an option (`--no-clean`) to avoid the cleanup.

VOTE: +1


okram:
I validated by having tinkerpop grapes and doing `ls` while it was starting up the script. grapes disappeared.

VOTE +1.

spmallette:
can you please add some notes to the developer docs for `--no-clean`:

https://github.com/apache/incubator-tinkerpop/blob/master/docs/src/dev/developer/contributing.asciidoc#building-and-testing

also, i know i wrote "--no-clean" in the JIRA issue, but perhaps it would be better as "--noClean" as we used the camel case form for "--dryRun".  As a result, "--noClean" would be more consistent.  Does that make sense?

dkuppitz:
Made sense. I modified the option name and added notes to the developer docs.

spmallette:
VOTE +1

PR #195, opened 2016-01-08
Do not include TraversalMetrics cap in profile results.

rjbriody:
https://issues.apache.org/jira/browse/TINKERPOP-1078

If the last step in a Traversal that is being profiled is ```cap(TraversalMetrics.METRICS_KEY)``` then that step is not profiled and not included as a row in the TraversalMetrics results.

Includes tests.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/ProfileStep.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/ProfileStep.java
@@ -134,6 +134,12 @@ public final class ProfileStep<S> extends AbstractStep<S, S> implements MapReduc
         final List<Step> steps = traversal.getSteps();
         for (int ii = 0; ii + 1 < steps.size(); ii = ii + 2) {
             Step step = steps.get(ii);
+
+            if (!(steps.get(ii+1) instanceof ProfileStep)) {
+                // If the next step is not a ProfileStep then we are done initializing.
+                return;
+            }
+
             ProfileStep profileStep = (ProfileStep) steps.get(ii + 1);
 
             // Create metrics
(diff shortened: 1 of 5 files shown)


okram:
VOTE +1.

okram:
Actually, why are we doing this? I don't think we should just rely on `cap(metrics)` and cut that out. What about `cap(metrics,m)`, etc. I say we leave this information in there. We start slicing and dicing and getting clever, it will turn into problems.

rjbriody:
> What about cap(metrics,m)

I do confirm that the size of the cap keys is 1, so the only cap will be the metrics.

I get what you're saying but I think this falls on the "safe side of the fence" with regards to being fancy. It's not that fancy. Right now, nearly all profiled traversals will end like this:

```.profile().cap(TraversalMetrics.METRICS_KEY)```

...and having the SideEffect step show up in the profile is bad for 2 reasons. 1) it's annoying but mostly 2) it alters the timing results for the other steps.

rjbriody:
For example, check out the last row in the profile step docs http://tinkerpop.apache.org/docs/3.1.0-incubating/#profile-step

The 'SideEffectCapStep([~metrics])' row really shouldn't be there.

okram:
I would say we move this to 3.2.0. It changes the user experience and if people are using the Profile object programmatically, it breaks their code.

Next, with 3.2.0 we are going to be changing how `ProfileStep` works so perhaps we see how that flushes out... 

Thoughts?

rjbriody:
That's fine.

What does "move this to 3.2.0" actually mean in terms of this PR?

okram:
Nothing. We haven't created a 3.2.0 branch yet as we are working on 3.1.1. We will soon. We can just let this PR sit till then. If it slips through the cracks, ping it up again. Thanks Bob.

okram:
VOTE +1. When this gets merged, I'll handle tweaking it to support the `Memory`-model in https://github.com/apache/incubator-tinkerpop/pull/243

rjbriody:
@okram make sure you merge this one first https://github.com/apache/incubator-tinkerpop/pull/242, as it will be easier to deal w/ conflicts in this PR.

okram:
@rjbriody --- is this PR still valid? Also, I really think we should keep all the information in there. I believe (moving forward too) that `explain()` and `profile()` should have the same steps in them.  Meaning, we don't hide information to make it "look good." This means, in OLAP, we need to make it so `TraversalVertexProgramStep` is shown, etc. We can do all that moving forward and thus, I don't think this PR is good anymore (from both a its so far behind from master/ and from a we should not hide steps perspective).

rjbriody:
I'll post my comments in the original ticket so it's easier to follow along if we scrap this PR.

spmallette:
Had a talk with @rjbriody - this PR can be closed as its stale at this point, but the JIRA ticket should stay open.

PR #196, opened 2016-01-08
TINKERPOP-1072: Allow the user to set persistence options using StorageLevel.valueOf()

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1072

I always thought Spark had some configuration like `default.storageLevel` and then when a user did `cache()` it would do that default. I was wrong. `cache()` is always `MEMORY_ONLY`. I made it so you can specify the storage level for both persisted RDDs and runtime job RDDs and thus now (internally) use `persist(STORAGE_LEVEL)` where `MEMORY_ONLY` is the default.

Test cases, docs, and Spark integration tests pass.

VOTE +1.


spark-gremlin/src/test/java/org/apache/tinkerpop/gremlin/spark/structure/io/PersistedInputOutputRDDTest.java
```diff
@@ -54,6 +56,44 @@
...
+                            .traversal(GraphTraversalSource.build().engine(ComputerTraversalEngine.build().computer(SparkGraphComputer.class)),
+                                    "gremlin-groovy",
+                                    "g.V()").create(graph)).submit().get();
+            ////////
+            assertTrue(Spark.hasRDD(Constants.getGraphLocation(rddName)));
+            assertEquals(StorageLevel.fromString(storageLevel), Spark.getRDD(Constants.getGraphLocation(rddName)).getStorageLevel());
+            assertTrue(Spark.hasRDD(Constants.getMemoryLocation(rddName, Graph.Hidden.hide("traversers"))));
+            assertEquals(StorageLevel.fromString(storageLevel), Spark.getRDD(Constants.getMemoryLocation(rddName, Graph.Hidden.hide("traversers"))).getStorageLevel());
+            assertEquals(counter * 2, Spark.getRDDs().size());
+            //System.out.println(SparkContextStorage.open().ls());
```
twilmes: Looks like there was a lingering debug println here that could be removed.

twilmes:
Looking good.  With or without that commented out println I'm a +1

dkuppitz:
* `mvn clean install`: passed
* integration tests: almost passed (only 1 error that's fixed in `master/`)

Vote: +1

PR #198, opened 2016-01-12
TINKERPOP-1035 Better Consistency in Gremlin Server Transaction Management

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1035

A second set of improvements on the transaction management system.  This round of changes undoes some of the work previously entered for this ticket as the concepts weren't as sound as they should have been.  In this PR, I've removed some of the special handling traversals had which I think would have confused people and pushed serialization more cleanly into the `GremlinExecutor` which ensure that it occurs in the same thread of execution as the original transaction.  This prevents "extra" transactions from being opened which could be problematic for graphs that incur "high" expense for that activity.  In another attempt to ensure that transactions were more tightly controlled, I always checked for an open transaction before actually issuing a commit/rollback. 

Tested with:

```text
mvn clean install && mvn verify -DskipIntegrationTests=false -pl gremlin-server
```

VOTE +1


dkuppitz:
* `mvn clean install`: passed
* integration tests: passed

VOTE: +1

okram:
`mvn clean install` ... VOTE +1.

PR #201, opened 2016-01-14
TINKERPOP-943 Warn if Gremlin Server is running prior to generating docs

dkuppitz:
Tried with Gremlin Server running and the preprocessor was cancelled with a nice error message. Tried with Gremlin server not running and and the preprocessing worked as usual.

VOTE:  +1

### DIFF
--- a/docs/preprocessor/preprocess.sh
+++ b/docs/preprocessor/preprocess.sh
@@ -36,6 +36,11 @@ do
   fi
 done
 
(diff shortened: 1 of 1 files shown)


spmallette:
looks nice to me:

```text
$ bin/process-docs.sh
The port 8182 is required for Gremlin Server, but already in use. Be sure to close the application that currently uses the port before processing the docs.
```

VOTE +1

okram:
VOTE +1

PR #204, opened 2016-01-19
TINKERPOP-1089: Order.shuffle implementation is too fragile

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1089

This required some refactoring. We have lots of `Comparator` wrappers everywhere so things get a little nasty. However, I added some solid test cases that ensure that the order violation doesn't happen with `Order.shuffle`.

VOTE +1.


okram:
BTW: I did Spark integration testing and the start of Giraph integration testing that checks order-stuff in `GraphComputerTest`.

dkuppitz:
* `mvn clean install`: passed
* `integration tests`: passed
* `manual test`: passed

Code looks good.

VOTE: +1

spmallette:
VOTE +1

PR #205, opened 2016-01-19
TINKERPOP-997 - Feature and FeatureRequirement adjustments

spmallette:
As they were very closely related, this PR covers:

https://issues.apache.org/jira/browse/TINKERPOP-997
https://issues.apache.org/jira/browse/TINKERPOP-998

it deprecates a `Feature` that was basically a duplicate of another.  It also opens up a good number of test cases to graphs that don't support meta-properties as there was a mis-definition of the `SIMPLE` requirement set.

Tested with `mvn clean install -DincludeNeo4j` and all was good. Also temporarily modified the TinkerGraph feature for meta-property support and manually noted that the right test cases opened up and closed down based on this change.  Would be nice to get some verification from other graphs that this change is all good.  

@jdellithorpe please take a look at this change when you get a moment, as you were the one who noted the problem in the first place.  Looks like we get about 90 tests opened up as a result of this.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -30,6 +30,7 @@ TinkerPop 3.1.1 (NOT OFFICIALLY RELEASED YET)
 * `DefaultTraversal` has a well defined `hashCode()` and `equals()`.
 * Added serializers to Gryo for `java.time` related classes.
 * Integrated `NumberHelper` in `SackFunctions`.
+* Deprecated `VertexPropertyFeatures.supportsAddProperty()` which effectively was a duplicate of `VertexFeatures.supportsMetaProperties`.
 * The Spark persistence `StorageLevel` can now be set for both job graphs and `PersistedOutputRDD` data.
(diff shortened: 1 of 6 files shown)


okram:
VOTE +1

twilmes:
Looks good.  VOTE +1

dkuppitz:
Just to make that time spent on integration tests not too worthless:

* `mvn clean install`: passed
* integration tests: passed

VOTE: +1

spmallette:
thanks @dkuppitz - the real test will be when providers start running their tests and stuff starts popping off. hopefully a few get a chance next week to try out all the tests that opened up as a result of these changes.

jdellithorpe:
This looks good to me. Just to be clear, now supporting meta-properties means specifically supporting the *addition* of meta-properties. Removing meta-properties is a separate feature, not implied by FEATURE_META_PROPERTIES.

spmallette:
yeah - we still have `VertexPropertyFeatures.supportsRemoveProperty()` hanging out there.

PR #206, opened 2016-01-22
TINKERPOP-818 Consider a P.type()

dkuppitz:
I'm not happy with the naming of the opposite of `P.type()`: `P.notType()`
Any suggestions are welcome and the refactoring should be pretty easy. But so far ...

* `mvn clean install`:  passed
* integration tests: passed

VOTE: +1


okram:
I don't think types should be comparable. Its equals or not equals. Next, there should NOT be `notType()` that is `not(type(Class))`. Also, are there any good reasons for this? Use cases?

dkuppitz:
> there should NOT be `notType()` that is `not(type(Class))`

I fully agree, BUT I had to [implement negate()](https://github.com/apache/incubator-tinkerpop/blob/11845b7a0408f910fed96a261369c315b3a39773/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Compare.java#L190-L193).

> Use cases?

We've seen use cases in the past I just can't provide one off my head.

okram:
You implemented `Compare` not `P`. If we do `type`, it needs to be a `P`-predicate. (and still assuming we want this...). I just haven't seen anyone (nor myself) ever need this functionality. And its bad to add functionality with limited use cases. Perhaps we wait until there is a stress for it AND we know all the ways in which it will be used. For example -- by doing this, we are assuming the Java-type system. This may be bad for Gremlin VM in other languages.

dkuppitz:
Do you suggest to implement the logic directly in `P`? So far the pattern has always been: implement the `BiPredicate` (`Compare` or `Contains`) and use it in `P`. For other language we could allow the type to be a `String`: `g.V()...is(type("java.lang.String"))`

okram:
I don't see how `type` is a comparable? Its not sortable.

okram:
Couple of notes. 

* Why is `Type` an enum?
* `Type` will need to implement serializable.
* Is `Type` the best name? Perhaps `TypePredicate`? We should really hide `Type` as the only way to get to it is via `P`.
* I would not do the `String` model just yet. Stick with just a `Class` parameter until we think through fully what this means for other languages.

spmallette:
@dkuppitz  This PR kinda died  - is it worth keeping it open?

dkuppitz:
Probably not. As @okram said "limited use cases" and then there's an open question on how to handle those things properly in different languages. Since nobody asked for it within the last 5-6 months nor did I need it, we should in fact "wait until there is a stress for it". I will close the PR and then delete the branch tomorrow if there are no objections.

spmallette:
cool - please close up the JIRA issue too if the intention is to never do this at all.

PR #207, opened 2016-02-01
TINKERPOP-1041

velo:
I decided to take a look into Tinkerpop-1041 and by doing so I end up setting up the project to build on a windows cloud CI and also fix a few problems I was able to spot on my machine.

This is not 100% done yet, just decided to give it some visibility before investing more time

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,23 @@
+version: '{build}'
+
+environment:
+  matrix:
+    - JAVA_HOME: C:\Program Files\Java\jdk1.8.0
+
+os: Windows Server 2012
+
+install:
+  - ps: |
+      Add-Type -AssemblyName System.IO.Compression.FileSystem
+      if (!(Test-Path -Path "C:\maven" )) {
+        (new-object System.Net.WebClient).DownloadFile('http://www.us.apache.org/dist/maven/maven-3/3.2.5/binaries/apache-maven-3.2.5-bin.zip', 'C:\maven-bin.zip')
+        [System.IO.Compression.ZipFile]::ExtractToDirectory("C:\maven-bin.zip", "C:\maven")
+      }
+  - cmd: SET PATH=C:\maven\apache-maven-3.2.5\bin;%JAVA_HOME%\bin;%PATH%
+  - cmd: SET MAVEN_OPTS=-XX:MaxPermSize=2g -Xmx4g
+  - cmd: SET JAVA_OPTS=-XX:MaxPermSize=2g -Xmx4g
+  - cmd: mvn --version
+  - cmd: java -version
+
+build_script:
+  - mvn clean install -B -Dci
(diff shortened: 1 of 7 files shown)


gremlin-groovy/src/test/java/org/apache/tinkerpop/gremlin/groovy/engine/ScriptEnginesTest.java
```diff
@@ -194,8 +195,8 @@ public void shouldFailUntilImportExecutes() throws Exception {
         threadEvalAndTriggerImport.join();
         threadImport.join();
 
-        assertTrue(successes.intValue() > 0);
-        assertTrue(failures.intValue() >= 500);
+        assertTrue("Success: " + successes.intValue() + " - Failures: " + failures.intValue(), successes.intValue() > 0);
```
spmallette: I tend to convert `assertTrue()` to hamcrest based asserts like `assertThat(success.intValue(), isGreaterThan(0))` which yields a better output on failure which i assume is what you are after here.

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/TestHelper.java
```diff
@@ -50,6 +52,7 @@
 public final class TestHelper {
 
     private static final String SEP = File.separator;
+    private static final char URL_SEP = '/';
```
spmallette: ah! the class lookup is by URL so you need standard URL pathing to handle that (and not the windows file separator). cool

spmallette:
Thanks for digging into this.  Can you please talk about what the "windows" maven profile is for?

I'd also like to know more about the appveyor config.  I guess you have that building against your fork with your appveyor account?

spmallette:
Sorry - i also meant to mention - can you please also rename this PR to "TINKERPOP-1041" so it hooks into JIRA?

velo:
the windows profile is to work around
https://issues.apache.org/jira/browse/HADOOP-10969
and
https://issues.apache.org/jira/browse/SPARK-6961

velo:
BTW, @spmallette  can you enable appveyor on this project as well?
So we would know how does the build look like on windows
https://ci.appveyor.com/project/velo/incubator-tinkerpop

spmallette:
@pluradj just thought i'd grab your attention on this issue....

@velo i can't just enable it as apache owns the repo - apache infra would have to enable it i guess.  we'll have to discuss on the mailing list anyway.  i'll bring it up

velo:
@spmallette thanks

appveyor.yml
```diff
@@ -0,0 +1,23 @@
...
+  matrix:
+    - JAVA_HOME: C:\Program Files\Java\jdk1.8.0
+
+os: Windows Server 2012
+
+install:
+  - ps: |
+      Add-Type -AssemblyName System.IO.Compression.FileSystem
+      if (!(Test-Path -Path "C:\maven" )) {
+        (new-object System.Net.WebClient).DownloadFile('http://www.us.apache.org/dist/maven/maven-3/3.2.5/binaries/apache-maven-3.2.5-bin.zip', 'C:\maven-bin.zip')
```
steveloughran: https considered more secure

velo: https://www.us.apache.org/dist/maven/maven-3/3.3.9/binaries/apache-maven-3.3.9-bin.zip 
is considered an insecure connection

steveloughran: really? I'd grab it from ASF Central, or think about checking sha1 checksums (extra work)

spmallette: +1 for https however that would work

PR #208, opened 2016-02-02
TinkerFactory.createClassic -> createModern

leifurhauks:
fixed reference to TinkerFactory.createClassic that should refer instead to TinkerFactory.createModern, which creates the version of TinkerGraph with vertex labels.


spmallette:
I think this was meant to be `createClassic()` - note the use of float values for the "weight" property.  

@okram any recollection on why we used classic here at all?  would it be better to demonstrate graph structure with modern since it includes vertex labels?

leifurhauks:
Ah, good point. So it's not quite equivalent to either classic or modern.

spmallette:
Short of there being some consensus to convert "classic" to "modern" in the docs, I don't think there's anything to change here.  Going to close this.

PR #210, opened 2016-02-04
TINKERPOP-962: Provide "vertex query" selectivity when importing data in OLAP.

okram:
TINKERPOP-962: Provide "vertex query" selectivity when importing data in OLAP.

https://issues.apache.org/jira/browse/TINKERPOP-962

(For TinkerPop 3.2.0 -- Breaking Change for GraphComputer Implementations)

This feature enables us to push down a `GraphFilter` predicate to the underlying OLAP graph system. For instance, if `g.V().count()` is executed by `SparkGraphComputer`, then there is no reason to load all the edges, simply push down a `GraphFilter`-predicate that filters out edges. For graph database providers like Titan, they can simply only send up the subset of the graph that is required for the OLAP job instead of filtering on the OLAP cluster machines. In the future, we will provide `GraphFilterTraversalStrategy` which will analyze the traversal and automatically generate a `GraphFilter` so the user is blind to which subsets of the full graph are actually being accessed by the OLAP engine.

This pull request yields a breaking change for graph system providers that have their own `GraphComputer` implementation. There are two new methods on `GraphComputer` and one new method on `GraphReader`.

```
GraphComputer vertices(Traversal<Vertex,Vertex> vertexFilter)
GraphComputer edges(Traversal<Vertex,Edge> edgeFilter)
GraphReader.readVertex(InputStream inputStream, GraphFilter graphFilter)
```

TinkerPop provides a `GraphFilter` object that does a lot of the heavy lifting so at minimum, the graph system provider simply needs to `GraphFilter.isLegal()` the vertices and edges it loads. Note that if the graph system provider relies on `GiraphGraphComputer` or `SparkGraphComputer`, then there is no change on their part unless they want to leverage the `GraphFilter` locally before sending their data to Giraph or Spark (an optimization that can be done at a later date without impacting users).

There was a host of changes that took place for this feature to be created. When merged, the `CHANGELOG.txt` will have the following new items:

```
* Added `GraphFilter` to support filtering out vertices and edges that won't be touched by an OLAP job.
* Added `GraphComputer.vertices()` and `GraphComputer.edges()` for `GraphFilter` construction (*breaking*).
* `SparkGraphComputer`, `GiraphGraphComputer`, and `TinkerGraphComputer` all support `GraphFilter`.
* Added `GraphComputerTest.shouldSupportGraphFilter()` which verifies all filtered graphs have the same topology.
* Added `GraphFilterAware` interface to `hadoop-gremlin/` which tells the OLAP engine that the `InputFormat` handles filtering.
* `GryoInputFormat` and `ScriptInputFormat` implement `GraphFilterAware`.
* Added `GraphFilterInputFormat` which handles graph filtering for `InputFormats` that are not `GraphFilterAware`.
* Fixed a bug in `TraversalUtil.isLocalStarGraph()` which allowed certain illegal traversals to pass.
* Added `TraversalUtil.isLocalVertex()` to verify that the traversal does not touch incident edges.
* `GraphReader` IO interface now has `Optional<Vertex> readGraph(InputStream, GraphFilter)`. Default `UnsupportOperationException`.
* `GryoReader` does not materialize edges that will be filtered out and this greatly reduces GC and load times.
* Created custom `Serializers` for `SparkGraphComputer` message-passing classes which reduce graph sizes significantly.
```

Ran `mvn clean install` and integration tests. Passed.

VOTE +1.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/GraphFilter.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/GraphFilter.java
@@ -47,10 +47,10 @@ public final class GraphFilter implements Cloneable, Serializable {
     private Traversal.Admin<Vertex, Vertex> vertexFilter = null;
     private Traversal.Admin<Vertex, Edge> edgeFilter = null;
 
-    protected boolean allowNoEdges = false;
-    protected Direction allowedEdgeDirection = Direction.BOTH;
-    protected Set<String> allowedEdgeLabels = new HashSet<>();
-    protected boolean allowAllRemainingEdges = false;
+    private boolean allowNoEdges = false;
+    private Direction allowedEdgeDirection = Direction.BOTH;
+    private Set<String> allowedEdgeLabels = new HashSet<>();
+    private boolean allowAllRemainingEdges = false;
 
     public void setVertexFilter(final Traversal<Vertex, Vertex> vertexFilter) {
         if (!TraversalHelper.isLocalVertex(vertexFilter.asAdmin()))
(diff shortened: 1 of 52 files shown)


okram:
The different between no `GraphFilter` and `GraphFilter`.

```
Application ID	Name	Cores	Memory per Node	Submitted Time	User	State	Duration
app-20160205091408-0001	Apache TinkerPop's Spark-Gremlin	80	10.0 GB	2016/02/05 09:14:08	tinkerpop	FINISHED	13 min
app-20160205090220-0000	Apache TinkerPop's Spark-Gremlin	80	10.0 GB	2016/02/05 09:02:20	tinkerpop	FINISHED	6.8 min
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/GraphFilter.java
```diff
@@ -0,0 +1,193 @@
...
+            this.allowNoEdges = true;
+        else if (this.edgeFilter.getStartStep() instanceof VertexStep) {
+            this.allowedEdgeLabels.clear();
+            this.allowedEdgeLabels.addAll(Arrays.asList(((VertexStep) this.edgeFilter.getStartStep()).getEdgeLabels()));
+            this.allowedEdgeDirection = ((VertexStep) this.edgeFilter.getStartStep()).getDirection();
+            //this.allowAllRemainingEdges = 1 == this.edgeFilter.getSteps().size();
+        }
+    }
+
+    /*public void compileFilters() {
```
twilmes: Did you mean to comment this out and remove calls to it? Figure you meant to but I recognized it from a previous commit.

okram: I was using `compileFilters()` at first, but then didn't need them. I left it commented out just in case its something that might pop up in the future. `GraphFilter` is very new and over this 3.2.0 push, I'm sure we will tweak here and there. Just want people to know its around.

Regarding the `allowRemainingEdges`. Again, its something I didn't need, but can easily add if we find a good use case.

dkuppitz:
* `mvn clean install`: passed
* integration test (incl. Neo4j): passed

VOTE: +1

twilmes:
Tests and code look good.

VOTE: +1

PR #214, opened 2016-02-05
TINKERPOP-1131: TraversalVertexProgram traverser management is inefficient memory-wise.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1131

This will go into TinkerPop 3.1.2. I will then upmerge it to master/ (3.2.0) and when https://github.com/apache/incubator-tinkerpop/pull/210 is merged to master/ (3.2.0), we will have a lean-mean OLAP processing machine!

I realized that I was copying `TraverserSets` instead of draining one into the other. By draining, for every traverser put into one set, it is removed from the other. In the worst case, we could have up to 3 sets of equivalent size all in memory at one vertex in OLAP. If you do any sort of `outE()`-type traversal, thats ALOT of data. Moreover, I reorganized the flow of processing as previously I was determining if messages needed to be sent right after received messages! Pointless waste of CPU cycles. Finally, I create a "bulking model" where by I try and fill a `Step` with as many traversers as I can before I have to drain it for message passing. Prior, it was one traverser at a time -- again, this can lead to significant inefficiencies.

`mvn clean install` -- integration tests (Giraph still running, but it will work as if one `TraversalVertexProgram` test passes then the meat is right). 

VOTE +1.

### DIFF
--- a/giraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/giraph/process/computer/GiraphComputation.java
+++ b/giraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/giraph/process/computer/GiraphComputation.java
@@ -26,6 +26,7 @@ import org.apache.tinkerpop.gremlin.hadoop.structure.io.ObjectWritable;
 import org.apache.tinkerpop.gremlin.hadoop.structure.io.VertexWritable;
 import org.apache.tinkerpop.gremlin.process.computer.VertexProgram;
 import org.apache.tinkerpop.gremlin.process.computer.util.ComputerGraph;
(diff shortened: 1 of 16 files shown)


okram:
Oh, the CHANGELOG will be:

```
* Optimized memory-usage in `TraversalVertexProgram`.
```

dkuppitz:
* `mvn clean install`: passed
* integration tests: passed

VOTE: +1

dkuppitz:
Reevaluated my previous tests.

Still: +1

spmallette:
+1 - did `mvn clean install` and some basic review of the code itself.

PR #218, opened 2016-02-09
TINKERPOP-1106 Better handled commit and serialization errors in Gremlin Server.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1106

Both types of errors were causing problems.  For commit errors, the last message from iteration was already sent back to the client as a terminating message so failure on commit after that was ignored by the driver because it had already received the terminator for that request.  Serialization errors didn't break the result iteration loop - they were being handled, but the result iteration was allowed to continue. I'm not so sure that those types of failures weren't leaking transactions either - that should be fixed now too.

This stuff is hard to write tests for with Gremlin Server as I dont' know how to force raise a commit exception for Neo4j.  I was able to test externally with graphs that were easier to force that condition.

I will update CHANGELOG after the merge for this one.

Tested with: `mvn clean install && mvn verify -pl gremlin-server -DskipIntegrationTests=false`

VOTE: +1


dkuppitz:
Code looks good, `mvn clean install` passed...

VOTE: +1

okram:
Looked at code. VOTE +1.

PR #219, opened 2016-02-09
Added another `edge()` method in `ScriptElementFactory`

dkuppitz:
Added another `edge()` method in `ScriptElementFactory` that allows to set explicit edge ids.

### DIFF
--- a/docs/src/reference/implementations.asciidoc
+++ b/docs/src/reference/implementations.asciidoc
@@ -1551,6 +1551,7 @@ Vertex vertex(Object id); // get or create the vertex with the given id
 Vertex vertex(Object id, String label); // get or create the vertex with the given id and label
 Edge edge(Vertex out, Vertex in); // create an edge between the two given vertices
 Edge edge(Vertex out, Vertex in, String label); // create an edge between the two given vertices using the given label
+Edge edge(Vertex out, Vertex in, Object id, String label); // create an edge between the two given vertices using the given edge id and label
 
 An appropriate `parse()` for the above adjacency list file is:
 
--- a/hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/structure/io/script/ScriptRecordReader.java
+++ b/hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/structure/io/script/ScriptRecordReader.java
@@ -138,5 +138,9 @@ public final class ScriptRecordReader extends RecordReader<NullWritable, VertexW
         public Edge edge(final Vertex out, final Vertex in, final String label) {
             return out.addEdge(label, in);
         }
+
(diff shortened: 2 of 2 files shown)


okram:
Why do you use a different order to your methods? Why not copy the parameterization from `Vertex`?

```
id, ?vertex?, label, vertex, properties...
```

okram:
I studied `ScriptElementFactory` and am thinking that API is poorly designed. Here are my concerns:

```
protected class ScriptElementFactory {

        private final StarGraph graph;

        public ScriptElementFactory() {
            this.graph = StarGraph.open();
        }

        public Vertex vertex(final Object id) {
            return vertex(id, Vertex.DEFAULT_LABEL);
        }

        public Vertex vertex(final Object id, final String label) {
            final Iterator<Vertex> vertices = graph.vertices(id);
            return vertices.hasNext() ? vertices.next() : graph.addVertex(T.id, id, T.label, label);
        }

        public Edge edge(final Vertex out, final Vertex in) {
            return edge(out, in, Edge.DEFAULT_LABEL);
        }

        public Edge edge(final Vertex out, final Vertex in, final String label) {
            return out.addEdge(label, in);
        }
    }
```

`ScriptElementFactory` should be a `GraphWrapper` so that the APIs are consistent.

```
ScriptGraph.addVertex()
ScriptVertex.addEdge()
```

Its weird that you diverge from the standard pattern for seemingly no reason except the `vertex(id,label)`-method logic.

okram:
I don't even know why you have `ScriptElementFactory` now. Why not just make available a static `getOrCreate()` and get rid of `ScriptElementFactory` ?

dkuppitz:
I can think about another API for TP 3.2, but this PR is for TP 3.1.2.

okram:
I was looking more into this. I think we should `Deprecate` ScriptElementFactory. Why not just use `StarGraph` directly? Your `vertex()`-logic is identical to `StarGraph.addVertex()`. Note that if its the `StarVertex`, it comes right back. If its not a new `StarAdjacentVertex` is created and returned.

I would recommend.

1. Do not add more methods to `ScriptElementFactory`.
2. Deprecate `ScriptElementFactory` in TP 3.2.0.
3. Update to use `StarGraph` directly, but if the user's script has a method for `ScriptElementFactory`, then use that method.

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1144

dkuppitz:
What's the remaining lifetime of something that is marked as deprecated? Can we deprecate ScriptElementFactory for 3.1.2 and remove it in 3.2.0?

okram:
No. We leave it around indefinitely. We may do a "big purge" at 3.5.0.

dkuppitz:
I've found no way to inject the `graph` as a global variable. I can either add another `parse` method parameter, which would break existing code, or I can add `ScriptElementFactory::getGraph()`.

PR #220, opened 2016-02-09
TINKERPOP-1137 Deprecate ScriptElementFactory and make star graph globally available

dkuppitz:
Previous PR + discussion: https://github.com/apache/incubator-tinkerpop/pull/219


dkuppitz:
* `mvn clean install`: passed
* `bin/process-docs.sh`: passed
* integration tests: passed

VOTE: +1

okram:
This is looking much better. Given that this is backwards compatible (nice Exception catching trick BTW), can you make at least one of the script.groovy in the old model so we know that backwards compatibility with Deprecated ScriptElementFactory is preserved. When we ultimately drop ScriptElementFactory, we then fix up the old script.groovy.

dkuppitz:
Reverted the changes in `grateful-dead-script-input.groovy`, so it's using the old model. `script-input.groovy` uses the new model.  Both scripts are used in unit tests, thus each case is covered.

okram:
VOTE +1.

twilmes:
Looks good
VOTE +1

PR #221, opened 2016-02-09
clear graph after IoTest.shouldReadWriteSelfLoopingEdges test runs.

pietermartin:


spmallette:
@pietermartin thanks for raising this issue and for offering a fix.  i have a more complete solution to the problem on this #223 - can you please take a look and see if it works for you and sqlg? 

If it works, I think you can close this PR.

spmallette:
dah - there was a secondary problem at play here which i just noticed - fixed with this: e982a0c1b018604e13755ccad65e308b77018fc3

this seems to work for titan now after this fix so i assume the problem is now resolved. @pietermartin if you can confirm, that would be great. not that this work is for 3.1.2-SNAPSHOT on the tp31 branch

pietermartin:
Fixed on tp31 3.1.2-SNAPSHOT

PR #223, opened 2016-02-10
TINKERPOP-1146 Use GraphProvider.clear() to auto-clear graphs created in tests

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1146

Was formerly just calling `Graph.close()` which wasn't enough to actually empty graphs of data. This led to test errors for some graph providers.  This change actually calls `GraphProvider.clear()` which by definition should empty a graph of data at the end of a test.

To test you technically need a graph that is persistent (e.g. Neo4j). Ran `mvn clean install -DincludeNeo4j` without problems.

VOTE +1

### DIFF
--- a/docs/src/upgrade/release-3.1.x-incubating.asciidoc
+++ b/docs/src/upgrade/release-3.1.x-incubating.asciidoc
@@ -22,6 +22,32 @@ image::https://raw.githubusercontent.com/apache/incubator-tinkerpop/master/docs/
 
 *A 187 On The Undercover Gremlinz*
 
+TinkerPop 3.1.2
+---------------
+
+*Release Date: NOT OFFICIALLY RELEASED YET*
+
+Please see the link:https://github.com/apache/incubator-tinkerpop/blob/3.1.1-incubating/CHANGELOG.asciidoc#tinkerpop-312-release-date-XXXXXXXXXXXXXXXXXXXXXXXXXX[changelog] for a complete list of all the modifications that are part of this release.
+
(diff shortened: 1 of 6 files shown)


docs/src/upgrade/release-3.1.x-incubating.asciidoc
```diff
@@ -22,6 +22,32 @@ image::https://raw.githubusercontent.com/apache/incubator-tinkerpop/master/docs/
...
+
+Graph System Providers
+^^^^^^^^^^^^^^^^^^^^^^
+
+GraphProvider.clear() Semantics
++++++++++++++++++++++++++++++++
+
+The semantics of the various `clear()` methods on `GraphProvider` didn't really change, but it would be worth reviewing
+their implementations to ensure that implementations can be called successfully in an idempotent fashion. Multiple
+calls to `clear()` may occur for a single test on the same `Graph` instance, as `3.1.1-incubating` introduced an
```
twilmes: Should `3.1.1-incubating` be `3.1.2-incubating` here?

okram:
VOTE +1. I will be so happy to see the "randomness" removed from our integration testing. Thanks @spmallette.

docs/src/upgrade/release-3.1.x-incubating.asciidoc
spmallette: It should be as-is - that feature came in at that version, but came with the described limitation

twilmes:
VOTE +1

PR #228, opened 2016-02-13
Tp31  - fix for gremlin server and console to open Neo4j 2.3.* databases

leemit:
Neo4j 2.3.0 changed the format of the neostore.nodestore.db 

This causes Neo4j databases created w/ Gremlin-server (e.g.  /tmp/neo4j) and neo4j databases created by the neo4j server from neo4j.com to be incompatible with each other.   Also Gremlin-server/console cannot open databases created by the newer Neo4j code (2.3.0 and higher)

The fix is two one line each in the gremlin pom files:
   gremlin-server/pom.xml
   neo4j-gremlin/pom.xml


okram:
Good find. This will go out in TinkerPop 3.2.0.

PR #229, opened 2016-02-13
slf4j-log4j12 / log4j is only required for testing (TINKERPOP-1151)

ceefour:
Issue: https://issues.apache.org/jira/browse/TINKERPOP-1151

### DIFF
--- a/gremlin-core/pom.xml
+++ b/gremlin-core/pom.xml
@@ -66,6 +66,7 @@ limitations under the License.
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-log4j12</artifactId>
             <version>${slf4j.version}</version>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>


spmallette:
I guess we shouldn't have log4j be a standard dependency for `gremlin-core`, given use of slf4j, but I'm not sure making it a test dependency is right.  we package log4j by default in the `gremlin-console` and `gremlin-server`.  maybe those slf4j implementations should be added to those project directly and they are made a test dependency here?

ceefour:
Yes, the "app" projects should have the runtime dependencies marked as "scope=runtime", and the "library" projects should have these dependencied marked as "scope=test"

spmallette:
Cool - that makes sense - can you please amend your PR appropriately and take all the modules into account in doing so?

spmallette:
Also, when you do that, can you please target this change at the `tp31` branch? I think we'd want to see this change for `3.1.2-incubating`.

ceefour:
Replaced by #233

PR #231, opened 2016-02-18
TINKERPOP-1165:Tooling Support: Compile with -parameters

rjbriody:
https://issues.apache.org/jira/browse/TINKERPOP-1165

Add -parameters to maven compile args.

This can be manually verified w/ something like this
```
System.out.println(GraphTraversal.class.getDeclaredMethods()[0].getParameters()[0].getName());
```

Used to print "arg0". Should now print "column".

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -308,6 +308,9 @@ limitations under the License.
                     <configuration>
                         <source>1.8</source>
                         <target>1.8</target>
+                        <compilerArgs>
+                            <arg>-parameters</arg>
+                        </compilerArgs>
                     </configuration>
                 </plugin>
                 <plugin>


spmallette:
Any idea of what side-effects this produces? I guess I'm trying to get at what the negative aspects of this configuration might be (having never used this configuration before).

zackurey:
@spmallette Off the top of my head:
1.  Maybe security through obfuscation?  But this is a open source library and one can easily know which method does what based on matching the version of the library with the type signatures
2.  This will make class files a little larger since before Java8 only the method signature was stored.  Also this does appear to only be supported by Java8, which is fine since Tinkerpop requires it:  https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javac.html.  Note, not there for Java7.

I do see some mention of the larger memory footprint issue here:  http://docs.oracle.com/javase/tutorial/reflect/member/methodparameterreflection.html.  So that is a consideration.  In this case, with gremlin-core the difference seems to be about 1K when looking at the uncompressed class file size difference.

spmallette:
Thanks @zackurey - appreciate that added information.  I'm going to just float this PR on the dev mailing list to make sure folks are aware of it.  If no objections, we'll go ahead and merge.

spmallette:
@rjbriody can you please re-target this PR a the `tp31` branch instead of `master`?  i'll merge it as soon as you do...sorry didn't notice that until today.

rjbriody:
I can do that. Will you then take care of merging the necessary tp31 changes to master?

spmallette:
yes - i will merge that down to master - no problem there.

rjbriody:
@spmallette https://github.com/apache/incubator-tinkerpop/pull/236

PR #232, opened 2016-02-18
TINKERPOP-1168 Switch plugins in docs preprocessor

dkuppitz:
Built the docs, verified that the output is still looking good.

VOTE: +1


okram:
Epic. Thank you. VOTE +1.

spmallette:
This fixes the problem but leaves the docs oddly organized.  The `implementations-neo4j.asciidoc` file isn't just "neo4j". It's basically the first half of the "implementations" section which includes "Graph System Provider Requirements" and "TinkerGraph-Gremlin".  

I guess we could fix this as part of:

https://issues.apache.org/jira/browse/TINKERPOP-937

but I think it would be better to just split `implementations-neo4j.asciidoc` into three files on this PR for each of the separate sections.  Either way - your call.

VOTE +1

dkuppitz:
I hope that's just a temporary solution. In an ideal world we would merge those 2 files back together pretty soon (as soon as we bumped out Spark dependency).

PR #235, opened 2016-02-22
TINKERPOP-1172 Fixed bug with driver connecting to an initially dead server.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1172

The `ConnectionPool` wasn't getting initialized properly when the host was dead which then lead to a null pointer exception when trying to reconnect.  Also fixed the possibility of a connection leak if the pool was partially initialized but the host died in the middle of that. Those partial connections were never properly closed.

I performed multiple successful runs of `mvn clean install && mvn verify -pl gremlin-server -DskipIntegrationTests=false`.

VOTE +1


pluradj:
Compiled, ran gremlin-server integration tests, and inspected the code.

VOTE +1

okram:
VOTE +1

PR #238, opened 2016-02-24
TINKERPOP-1107 Added interpreter mode for ScriptEngine

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1107

Configuration for the `ScriptEngine` that allows the user to send `def` top level variables in a script and have them be treated as "global".  Traditionally, for a script like this:

```groovy
def x = 1
int y = 2
z = 3
```

the variables `x` and `y` would be local to that script only while `z` would be preserved globally for future scripts. Users can find that confusing especially when using issuing script to Gremlin Server in a "session".  When enabling "interpreter mode", all of the above variables assignments will be treated as global.

VOTE +1


spmallette:
I've rebased this PR now that groovy has been bumped to 2.4.6. Still all good and ready for review/vote

spmallette:
@velo we aren't really having much luck with appveyor.  don't think we've seen a clean build on any branch/pr since we merged your original PR. thoughts?

velo:
```
Failed tests: 
ProfileTest$Traversals>ProfileTest.g_V_sideEffectXThread_sleepX10XX_sideEffectXThread_sleepX5XX_profile:173 Duration should be at least the length of the sleep (59ms): 51 
```

This same test keep breaking.... this test is intermittent for me as well locally.

spmallette:
i don't think it's the same test though:

https://ci.appveyor.com/project/ApacheSoftwareFoundation/incubator-tinkerpop/build/185

this got well past that spot. take a look at some of the other jobs too.

velo:
That other I don't know what is about...

okram:
VOTE +1.

dkuppitz:
Intentionally I've never used `def` or alike for global variables, but being able to do that will make some of my code snippets look a lot better. Nice!

Vote: +1

PR #243, opened 2016-02-26
TINKERPOP-1166, TINKERPOP-1164, TINKERPOP-1057, TINKERPOP-1162

okram:
#243 https://issues.apache.org/jira/browse/TINKERPOP-1166
https://issues.apache.org/jira/browse/TINKERPOP-1164
https://issues.apache.org/jira/browse/TINKERPOP-1057
https://issues.apache.org/jira/browse/TINKERPOP-1162

This PR has all these tickets combined because in order to solve any I needed to make sure all these ticket's respective solutions played well with one another. In summary, GraphComputer `Memory` has come to the center stage as a replacement for `MapReduce`. The benefit is that `Memory` can compute/reduce in parallel with a VertexProgram's execution run. This has the following benefits:

* Reductions happen within the vertex program and thus, no subsequent rescan of the full graph needed to generate sideEffect. For example -- `g.V().count()` is one pass through the graph, not two.
* Because reductions happen within the vertex program, barrier steps are no longer the cut-off point for an OLAP traversal. An OLAP traversal can now have multiple reducing barrier steps within it (e.g. `max()`, `groupCount()`, `min()`, `group()`, etc.). Its all one job.
* Vertex compute keys can be marked transient and are automagically removed from the resultant graph. This is good for removing "scratch data." (e.g. PageRankVertexProgram.EDGE_COUNTS).
* Memory compute keys can be marked transient and are automagically remove from the result memory. This is good for removing "scratch data." (e.g. VOTE_TO_HALT).
* Memory compute keys can be declared to NOT broadcast and thus, no be sent to the workers on each iteration. Workers can still send data, but just can not read data.

Finally, one of the major side-effect benefits of this work is that numerous traversals that were considered "illegal" by `ComputerVerificationStrategy` are no longer illegal. The only types of traversals that are illegal in OLAP are those that have `by()`-modulators that go beyond the local star graph or are path-based (`select()` and `path()`) and go beyond the element id in their `by()`-modulations.

This work creates breaking changes for both users (trivial) and providers (intense). However, for providers, its only those providers that have their own custom `GraphComputer` implementation. If they use `SparkGraphComputer` or `GiraphGraphComputer`, no work is required of them.

CHANGELOG

```
* Added `MemoryComputeKey` for specification of `Memory` keys in `VertexProgram`. (*breaking*)
* Added `VertexComputeKey` for specification of vertex compute properties in `VertexProgram`. (*breaking*)
* Added `and`, `or`, and `addAll` to `Operator`.
* `Memory` API changed to support setting and adding values for reduction. (*breaking*)
* `Memory` keys can be marked as broadcast and only those values are sent to workers on each iterator. 
* `Memory` keys can be marked transient and thus deleted at the end of the OLAP job.
* Vertex compute keys can be marked transient and thus deleted at the end of the OLAP job.
* `VertexProgram` API changed to support `MemoryComputeKey` and `VertexComputeKey`. (*breaking*)
* `TraversalVertexProgram` able to execute OLAP and OLTP traversal sections dynamically within the same job.
* Removed `FinalGet` interface as all post processing of reductions should be handled by the reducing step explicitly. (*breaking*)
* Greatly simplified all `ReducingBarrierStep` implementations as they no longer require `MapReduce` in OLAP.
* Nearly all steps in OLAP that used `MapReduce` now use `Memory` to do their reductions which expands the list of legal traversals.
* `GroupStep` simplified with `GroupHelper.GroupMap` no longer being needed. Related to the removal of `FinalGet`.
* OLAP side-effects that are no longer generated by `MapReduce` are simply stored in `ComputerResult.Memory` w/ no disk persistence needed. (*breaking*)
* Added `Generate` step interface which states that there could be a final generating phase to a side-effect or reduction (e.g. `GroupStep`).
* `Barrier` step interface is the means by which non-parallel steps communicate with their counterparts in OLAP.
* Added `MemoryComputing` step interface which states that the step uses `MemoryComputeKeys` for its computation in OLAP.
```

UPGRADE-PROVIDERS

```
GraphComputer updates to semantics and API
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Providers that have a custom `GraphComputer` implementation will have a lot to handle. Note that if the graph system simply uses `SparkGraphComputer` or `GiraphGraphComputer` provided by TinkerPop, then no updates are required.

`Memory` updates:

* Any `BinaryOperator` can be used for reduction and is made explicit in the `MemoryComputeKey`.
* `MemoryComputeKeys` can be marked transient and must be removed from the resultant `ComputerResult.memory()`.
* `MemoryComputeKeys` can be specified to not broadcast and thus, must not be available to workers to read in `VertexProgram.execute()`.
* The `Memory` API has been changed. No more `incr()`, `and()`, etc. Now its just `set()` (setup/terminate) and `add()` (execute).

See TINKERPOP-1166:https://issues.apache.org/jira/browse/TINKERPOP-1166

Operational semantic test cases have been added to `GraphComputerTest` to ensure that all the above behaviors are implemented correctly.

Providers that have a custom `ReducingBarrierStep` implementation will need to adjust their implementation slightly to accommodate a new API that reflects the `Memory` updates above. This should be a simple change. Note that `FinalGet` no longer exists and such post-reduction processing is handled by the reducing step (via `Generating` interface).

See TINKERPOP-1164:https://issues.apache.org/jira/browse/TINKERPOP-1164

```

*UPGRADE-USERS*

```
VertexProgram and MemoryComputeKey and VertexComputeKey
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Users that have custom `VertexProgram` implementations will need to change their implementations to support the new `VertexComputeKey` and `MemoryComputeKey` classes. In the `VertexPrograms` provided by TinkerPop, these changes were trivial, taking less than 5 minutes to make the updates.

* `VertexProgram.getVertexComputeKeys()` returns a `Set<VertexComputeKey>`. No longer a `Set<String>`. Use `VertexComputeKey.of(String key,boolean transient)` to generate a `VertexComputeKey`. Transient keys were not supported in the past, so to make the implementation semantically equivalent, the boolean transient should be false.
* `VertexProgram.getMemoryComputeKeys()` returns a `Set<MemoryComputeKey>`. No longer a `Set<String>`. Use `MemoryComputeKey.of(String key, BinaryOperator reducer, boolean broadcast, boolean transient)` to generate a `MemoryComputeKey`. Broadcasting and transients were not supported in the past so to make the implementation semantically equivalent, to boolean broadcast should be true and the boolean transient should be false.

See TINKERPOP-1162:https://issues.apache.org/jira/browse/TINKERPOP-1162

*SparkGraphComputer and GiraphGraphComputer Persistence*
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Most of the `MapReduce`-based steps in `TraversalVertexProgram` have been removed and replaced using the new `Memory`-reduction model. `MapReduce` jobs always created a persistence footprint (e.g. in HDFS). Memory data was never persisted to HDFS. As such, there will not be data on the disk that is accessible. For instance, no more `~reducing`, `~traversers`, and specially named side-effects such as those from `groupCount('m')`. The data is still accessible via `ComputerResult.memory()`, its just simply does not have a corresponding on-disk representation.
```


dylanht:
When I tried to build this, maven complained the now capitalized "D" in the GroupSideEffectStepV3D0Test.java filename:

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/GroupSideEffectStepV3D0Test.java

Changing it back to lower case fixed it and the build goes through.

spmallette:
If that naming is the problem with that test it should get fixed on the tp31 branch as it's failing appveyor over there:

https://issues.apache.org/jira/browse/TINKERPOP-1179

okram:
@dylanht  -- dah. Sometimes git does that. I did a class rename and it accidently made it "D". Then I renamed it to "d" and it still always gets fuggled. @spmallette -- you have had this problem before. What is the fix?

spmallette:
i don't remember offhand - i'll have to take a look. will fix on the tp31 side though so you'll probably want to rebase your branch when that happens.

okram:
```
[INFO] Reactor Summary:
[INFO]
[INFO] Apache TinkerPop .................................. SUCCESS [3.658s]
[INFO] Apache TinkerPop :: Gremlin Shaded ................ SUCCESS [2.210s]
[INFO] Apache TinkerPop :: Gremlin Core .................. SUCCESS [35.088s]
[INFO] Apache TinkerPop :: Gremlin Test .................. SUCCESS [10.526s]
[INFO] Apache TinkerPop :: Gremlin Groovy ................ SUCCESS [38.056s]
[INFO] Apache TinkerPop :: Gremlin Groovy Test ........... SUCCESS [5.694s]
[INFO] Apache TinkerPop :: TinkerGraph Gremlin ........... SUCCESS [3:08.700s]
[INFO] Apache TinkerPop :: Hadoop Gremlin ................ SUCCESS [4:58.574s]
[INFO] Apache TinkerPop :: Spark Gremlin ................. SUCCESS [6:52.110s]
[INFO] Apache TinkerPop :: Giraph Gremlin ................ SUCCESS [2:18:15.161s]
[INFO] Apache TinkerPop :: Neo4j Gremlin ................. SUCCESS [18:09.462s]
[INFO] Apache TinkerPop :: Gremlin Driver ................ SUCCESS [8.172s]
[INFO] Apache TinkerPop :: Gremlin Server ................ SUCCESS [11:11.713s]
[INFO] Apache TinkerPop :: Gremlin Console ............... SUCCESS [1:08.896s]
[INFO] Apache TinkerPop :: Gremlin Archetype ............. SUCCESS [0.083s]
[INFO] Apache TinkerPop :: Archetype - TinkerGraph ....... SUCCESS [5.695s]
[INFO] Apache TinkerPop :: Archetype - Server ............ SUCCESS [10.165s]
[INFO] BUILD SUCCESS
[INFO] Total time: 3:05:44.437s
[INFO] Finished at: Mon Feb 29 14:01:04 MST 2016
[INFO] Final Memory: 97M/810M
```

VOTE +1.

spmallette:
Note the commit above that fixes the badly named test. easy to fix with `git mv`:

```text
$ git mv -f gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/GroupSideEffectStepV3D0Test.java gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/GroupSideEffectStepV3d0Test.java
```

dkuppitz:
Travis is happy, docs built properly on my local machine,

VOTE: +1

spmallette:
VOTE +1 - full build with integration tests

PR #245, opened 2016-02-29
TINKERPOP-1185 - only load/instantiate the plugins listed in plugins.txt

bdeggleston:


dkuppitz:
Nice. Now I can actually simplify the docs pre-processor again.

VOTE: +1

spmallette:
Removing the `ServiceLoader` is a bigger change than it immediately appears.  The `ServiceLoader` provides functionality that allows the console to show the user the list of plugins that are present but not "activated":

```text
gremlin> :plugin list
==>tinkerpop.server[active]
==>tinkerpop.gephi
==>tinkerpop.utilities[active]
==>tinkerpop.sugar
==>tinkerpop.credentials
==>tinkerpop.tinkergraph[active]
```

With this change in place you get:

```text
gremlin> :plugin list
==>tinkerpop.server[active]
==>tinkerpop.utilities[active]
==>tinkerpop.tinkergraph[active]
```

Of course, as we are learning, the notion of "active" is somewhat blurry because an "inactive" plugin may still load some aspect of a class during instantiation by `ServiceLoader`.  An "inactive" plugin is really one that doesn't have its plugin methods called to perform imports, init the console environment, etc.

I guess the question is whether or not its valuable to know what plugins are available but not active. If it was determined that we could just drop that capability, there'd be some additional code/docs to amend to this PR. If we wanted to try to keep this functionality, i guess a custom `ServiceLoader` could work - it would just need to do the same thing as the current `ServiceLoader` but not throw an exception if a class could not be found - though that would have its own dangers in some sense.  

@bdeggleston do you see any other ways to implement this without losing the functionality of listing available plugins as it works now?

bdeggleston:
I had a feeling this was too easy :)

So, as far as I can tell, there's no way to automagically detect `GremlinPlugin` implementations without loading and inspecting every class on the classpath. Internally, this is what `ServiceLoader` is doing, which causes static initialization of every class it loads.

What do you think about deferring the classpath scanning until `:plugin list` is run? I think the autodiscovery is useful, but it doesn't need to run on startup.

spmallette:
The idea of "deferring" is interesting, but doesn't that just push the failure to when someone does an `:install` or `:plugin list`? won't the console come crashing down at that point? or is there something to be done to prevent that?

bdeggleston:
It would also defer any failures for `:plugin list`, yes. `:plugin install` should be ok, assuming it only attempts to load the plugin class being installed.

To answer your question about preventing the failures during `:plugin list`: we should be able to catch exceptions, but we couldn't prevent any side effects (file io, etc) that occur during class initialization. That's another problem with deferring, `:plugin list` could end up having a lot of unintended side effects depending on what's on the classpath.

So then I guess it boils down to whether you think it's better to auto discover plugins, or better to skip loading unused classes.

spmallette:
I don't think there are any easy solutions here. I'll comment in more detail in JIRA but I'm not seeing how this gets done without considerable change to our approach to plugins. As long as we use `ServiceLoader` and require an instance of a plugin its going to load things that are inactive. I think we should probably just close this PR.

PR #246, opened 2016-02-29
Make the GraphManager instance injectable into ServerGremlinExecutor

xenji:
This refers to a discussion from the gremlin-users mailing list:
[https://groups.google.com/forum/#!topic/gremlin-users/xYzCAQExvvM](https://groups.google.com/forum/#!topic/gremlin-users/xYzCAQExvvM)

It is currently not possible to use a standard way to inject a
fully configured graph. The way presented in this gist is only
a hack and should not be taken as the default way.
[https://gist.github.com/xenji/e469abde2e0b80aadc4e](https://gist.github.com/xenji/e469abde2e0b80aadc4e)

Adding a new constructor parameter is, from my point of view,
the most flexible and compatible change to enable the injection
without breaking old code.

I've run the whole test suite of the server before pushing without
failures.

### DIFF
--- a/docs/src/reference/gremlin-applications.asciidoc
+++ b/docs/src/reference/gremlin-applications.asciidoc
@@ -517,8 +517,8 @@ CompletableFuture<List<Result>> results = client.submit("[1,2,3,4]").all();  <3>
 
 CompletableFuture<ResultSet> future = client.submitAsync("[1,2,3,4]"); <4>
 
-Map<String,Object> params = new HashMap<>()
-params.put("x",4)
+Map<String,Object> params = new HashMap<>();
+params.put("x",4);
 client.submit("[1,2,3,x]", params); <5>
 ----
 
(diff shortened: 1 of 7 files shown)


spmallette:
I think this might be a good idea.  Can you please do a the following:

1. Rebase this against the `tp31` branch - I think it can safely go into 3.1.2 (rather than the 3.2.x line of code)
2. Instead of modifying the constructor, can you please just add a new one and keep the old one? That way we don't break anyone's code. If the users chooses to not supply `GraphManager` you can just instantiate from `Settings`.

Let's see what it looks like from that perspective.

xenji:
Ok, I'll do the rebase asap. I got two questions regarding your implementation advice. 

The first:
When I got you right, you want me to remove the parameter from the "old" constructor and create a new one with the GraphManager as parameter, right? What I do not get is: When I create a new constructor with the GraphManager as parameter, leaving the old one as it is, how should I influence the "bootstrap" process given in the old constructor? Setting the member before calling the other constructor does not work by design and duplicating all the code seems no good option to me. Can you point me in the right direction?

The second:
The intention behind the PR was - at least from my point of view - exactly what you described. I wanted the original signature of the constructor behave the same way as before. What did I do wrong to not achieve that goal?

spmallette:
oops - i just misread the diff.  i think what you have is fine. it remains non-breaking it what it does. never mind what i said about that. let's just get this PR lined up against tp31 and it should be good to go.

xenji:
I close this one in favor of #250, which is directly based on tp31.

PR #248, opened 2016-03-01
TINKERPOP-956 Connection errors tend to force a complete close of the channel

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-956

Implemented a method of refreshing a connection when notification was received that it was closed on the server-side. As a result the driver should do a better job of recovering in the face of a server-side closing of a channel.  

Had a good build with `mvn clean install && mvn verify -DskipIntegrationTests=false -DincludeNeo4j`

VOTE +1


okram:
VOTE +1.

dkuppitz:
VOTE: +1

PR #251, opened 2016-03-04
TINKERPOP-1192 & TINKERPOP-1193 & TINKERPOP-951 & TINKERPOP-958: The future is now.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1192
https://issues.apache.org/jira/browse/TINKERPOP-1193
https://issues.apache.org/jira/browse/TINKERPOP-951
https://issues.apache.org/jira/browse/TINKERPOP-958

Summary: `TraversalSideEffects` and `GraphComputer` `Memory` are aligned. The are unified via `MemoryTraversalSideEffects` which gives us the flexibility to have a single `TraversalSideEffects` across an arbitrary number of OLAP jobs and finally makes it so that OLTP and OLAP use of sideEffects are consistent. This comes at the cost of a few API changes to `TraversalSideEffects`  and some semantics changes. For 99% of users, they won't notice. Only those that make heavy use of `sideEffect{x = ...}` may have to change their usage.

```
g.V().out("created").
  group("m").by(label).
  pageRank(1.0).
    by("pageRank").
    by(inE()).
    times(1).
  in("created").
  group("m").by("pageRank").
    cap("m")
```

The result is:

[{2.0=[v[4], v[4], v[4], v[4]], 1.0=[v[6], v[6], v[6], v[1], v[1], v[1]], software=[v[5], v[3], v[3], v[3]]}]
		
Do you see why this traversal is so cool? Check out what it is doing ---

	1. Normal out("created").group(). Fine and dandy all the software vertices are grouped.
	2. But then we kick off another OLAP job that computes a single-step spreading activation starting at the software vertices.
	3. Then, from the software vertices, it goes to the people who created that software and groups those people by their pageRank value.

First, (already known), this traversal compiles to 3 OLAP jobs and it "just works." Slammin'.
Second, as of this afternoon, TraversalSideEffects (GraphComputer.Memory) are preserved across OLAP jobs and across TraversalVertexPrograms.
	--- it truly is just one traversal.

------

CHANGELOG

```
* Fixed a severe `Traversal` cloning issue that caused inconsistent `TraversalSideEffects`.
* `TraversalSideEffects` remain consistent and useable across multiple chained OLAP jobs.
* Added `MemoryTraversalSideEffects` which backs a `TraversalSideEffects` by `Memory` in OLAP.
* `TraversalSideEffects` are now fully functional in OLAP save that an accurate global view is possible after an iteration.
* Updated the `TraversalSideEffects` API to support registered reducers and updated `get()`-semantics. (*breaking*)
* Split existing `profile()` into `ProfileStep` and `ProfileSideEffectStep`. 
* The `profile()`-step acts like a reducing barrier and emits `TraversalMetrics` without the need for `cap()`. (*breaking*)
* Added `LocalBarrier` to allow traversers to remain distributed during an iteration so as to limit cluster traffic.
* An OLAP-based `Barrier` synchronization bug has been fixed.
```

UPDATE

```
TraversalSideEffect Update
^^^^^^^^^^^^^^^^^^^^^^^^^^

There were significant changes to `TraversalSideEffect` both at the semantic level and at the API level. Users that have traversals of the form `sideEffect{…}` that use global side-effects should read the following carefully. If the user only uses side-effects via non-lambda steps (e.g. `groupCount("m")`) and with `g.withSideEffects()`, then the changes below will not effect them. Providers will not be effected by the changes save any tests cases that use `Traversal.getSideEffects().get()`. This method no longer returns `Optional<V>`, but instead just the value `<V>`.

TraversalSideEffect Get API Change
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

`TraversalSideEffects` can now logically operate within a distributed OLAP environment. In order to make this possible, it is necessary that each side-effect be registered with a reducing `BinaryOperator`. This binary operator will combine distributed updates into a single global side-effect at the master traversal. Many of the methods in `TraversalSideEffect` have been `Deprecated`, but they are backwards compatible save that `TraversalSideEffects.get()` no longer returns an `Optional`, but instead throws an `IllegalArgumentException`. While the `Optional` semantics could have remained, it was deemed best to directly return the side-effect value to reduce object creation costs and because all side-effects must be registered apriori and therefore there is never a reason why an unknown side-effect key would be used.

TraversalSideEffect Registration Requirement
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

All `TraversalSideEffects` must be registered upfront. This is because, in OLAP, side-effects map to `Memory` compute keys and as such, must be declared prior to the execution of the `TraversalVertexProgram`. If a user's traversal creates a side-effect mid-traversal, it will fail. The traversal must use `GraphTraversalSource.withSideEffect()` to declare the side-effects it will use during its execution lifetime.

TraversalSideEffect Add Requirement
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

In a distributed environment, a side-effect can not be mutated and be expected to exist in the mutated form at the final, aggregated, master traversal. For instance, if a side-effect "myCount" references a `Long`, the `Long` can not be updated directly via `sideEffects.set("myCount", sideEffects.get("myCount") + 1)`. Instead, it must rely on the reducer to do the merging and thus, the `Step` must do `sideEffect.add("mySet",1)` where the registered reducer is `Operator.sum`. Note that `Traverser.sideEffect(key,value)` will use `TraversalSideEffect.add()`. Thus, the below will increment "a". If no operator was provided, then the operator is assumed `Operator.assign` and the final result of "a" would be 1.

`g.withSideEffect("a",0,sum).V().out().sideEffect(t -> t.sideEffect("a",1))`

See link:https://issues.apache.org/jira/browse/TINKERPOP-1192[TINKERPOP-1192]

ProfileStep Update and GraphTraversal API Change
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

The `profile()`-step has been refactored into 2 steps -- `ProfileStep` and `ProfileSideEffectStep`. Users who previously used the `profile()` in conjunction with `cap(TraversalMetrics.METRICS_KEY)` can now simply omit the cap step. Users who retrieved `TraversalMetrics` from the side-effects after iteration can still do so, but will need to specify a side-effect key when using the `profile()`. For example, `profile("myMetrics")`.

See link:https://issues.apache.org/jira/browse/TINKERPOP-958[TINKERPOP-958]
```

VOTE +1.


okram:
```
[INFO] Reactor Summary:
[INFO]
[INFO] Apache TinkerPop .................................. SUCCESS [4.845s]
[INFO] Apache TinkerPop :: Gremlin Shaded ................ SUCCESS [2.346s]
[INFO] Apache TinkerPop :: Gremlin Core .................. SUCCESS [41.846s]
[INFO] Apache TinkerPop :: Gremlin Test .................. SUCCESS [11.712s]
[INFO] Apache TinkerPop :: Gremlin Groovy ................ SUCCESS [42.031s]
[INFO] Apache TinkerPop :: Gremlin Groovy Test ........... SUCCESS [6.604s]
[INFO] Apache TinkerPop :: TinkerGraph Gremlin ........... SUCCESS [4:01.081s]
[INFO] Apache TinkerPop :: Hadoop Gremlin ................ SUCCESS [5:18.703s]
[INFO] Apache TinkerPop :: Spark Gremlin ................. SUCCESS [7:54.731s]
[INFO] Apache TinkerPop :: Giraph Gremlin ................ SUCCESS [2:22:15.673s]
[INFO] Apache TinkerPop :: Neo4j Gremlin ................. SUCCESS [18:03.868s]
[INFO] Apache TinkerPop :: Gremlin Driver ................ SUCCESS [9.911s]
[INFO] Apache TinkerPop :: Gremlin Server ................ SUCCESS [11:23.461s]
[INFO] Apache TinkerPop :: Gremlin Console ............... SUCCESS [1:10.067s]
[INFO] Apache TinkerPop :: Gremlin Archetype ............. SUCCESS [0.082s]
[INFO] Apache TinkerPop :: Archetype - TinkerGraph ....... SUCCESS [6.484s]
[INFO] Apache TinkerPop :: Archetype - Server ............ SUCCESS [10.505s]
[INFO] BUILD SUCCESS
[INFO] Total time: 3:12:24.450s
[INFO] Finished at: Thu Mar 03 22:44:08 MST 2016
[INFO] Final Memory: 99M/813M
~/software/tinkerpop/tinkerpop3$
```

dkuppitz:
Docs don't build properly (error in `the-traversal.asciidoc`).

okram:
Bug in docs fixed and I also did some more work making our docs (OLAP stuff) better.

Published:
  http://tinkerpop.apache.org/docs/3.2.0-SNAPSHOT/reference/

dkuppitz:
Worked!

VOTE: +1

spmallette:
good with `mvn clean install`

VOTE +1

PR #252, opened 2016-03-04
TINKERPOP-1184 Fixed bug where session close was happening an extra time.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1184

If a session was closed explicitly by the client, the scheduled job to close the session on expiration was not being cancelled leading to a second closing. The second closing really wasn't doing anything "bad" but it did generate extra logging that was confusing if the user had expected the session to already be closed.  Refactored some of the session "kill" operations to make them a bit more thread safe.

Tested with ` mvn clean install -DskipTests && mvn verify -pl gremlin-server -DskipIntegrationTests=false -DincludeNeo4j`

VOTE +1


okram:
VOTE +1.

dkuppitz:
Well, I guess that when I discovered that, I had thousands of background tasks that were waiting to kill something that did no longer exist. Thus it kinda did something bad in that it was wasting a ton of resources.

VOTE: +1

spmallette:
true - maybe more bad than i gave it credit for - tons of scheduled jobs just hanging around.

PR #254, opened 2016-03-09
TINKERPOP-1180: Add more optimized binary operators to Operator

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1180

This is simple. Added `Operator.sumLong` which bypasses using `NumberHelper`. Smart when you know you are summing longs as in `CountGlobalStep`. I also tweaked the initial capacity of the singleton maps of the `GroupXXXSteps`. As we need more Operators, we can just CTR add them. Its trivial.

`mvn clean install`.

VOTE +1.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/GroupSideEffectStep.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/GroupSideEffectStep.java
@@ -77,7 +77,7 @@ public final class GroupSideEffectStep<S, K, V> extends SideEffectStep<S> implem
 
     @Override
     protected void sideEffect(final Traverser.Admin<S> traverser) {
-        final Map<K, Object> map = new HashMap<>();
+        final Map<K, Object> map = new HashMap<>(1);
         final K key = TraversalUtil.applyNullable(traverser, keyTraversal);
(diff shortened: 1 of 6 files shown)


spmallette:
VOTE +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/GroupSideEffectStep.java
```diff
@@ -77,7 +77,7 @@ public void modulateBy(final Traversal.Admin<?, ?> kvTraversal) {
 
     @Override
     protected void sideEffect(final Traverser.Admin<S> traverser) {
-        final Map<K, Object> map = new HashMap<>();
+        final Map<K, Object> map = new HashMap<>(1);
```
dkuppitz: If you expect only 1 entry, then you should also set the load factor to 1.0.

```java
map = new HashMap<>(1, 1.0f);
```

okram: That line of code is as you have it in master/. Was done in another PR. Also, the load factor can't be 1.0 as that map will presumably grow on reduction.

dkuppitz:
VOTE: +1

PR #256, opened 2016-03-09
TINKERPOP-1096 TINKERPOP-1097 Session Aliases, Console Sessions and Remote Console

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1096
https://issues.apache.org/jira/browse/TINKERPOP-1097

This PR introduces several things:

1. `SessionedClient.alias()` is now implemented so that you can do aliases on a session-based connection
1. The console's `:remote connect` now allows you to append "session" to the end of it so that it starts that remote with a session
1. The console now has `:remote console` which allows the console to be put into a mode where all scripts are sent to remote and the `:>` is no longer required

Tested the console manually in an abundance of ways but here's an example:

```text
gremlin> :remote connect tinkerpop.server conf/remote.yaml session
==>Connected - localhost/127.0.0.1:8182-[98bc7cf9-1c7b-451a-acbd-124a53a811d3]
gremlin> :remote config alias a g
==>a=g
gremlin> :remote console
==>All commands will now be sent to Gremlin Server - [localhost/127.0.0.1:8182]-[98bc7cf9-1c7b-451a-acbd-124a53a811d3] - type ':remote console' to exit
gremlin> x = 1
==>1
gremlin> y = 2
==>2
gremlin> x + y
==>3
gremlin> a.V()
==>v[0]
==>v[1]
==>v[3]
gremlin> :remote config alias b g
==>b=g
gremlin> b.V()
==>v[0]
==>v[1]
==>v[3]
```

Note that this is a non-breaking change. It is completely compatible with what we had before and the `:>` still works as it did before. This might yet be useful for some users as they might want to work with remote data in a local way and mix local/remote evaluations. Also, some plugins won't work well without it - like the gephi plugin.

Builds/tests nicely with: `mvn clean install && mvn verify -pl gremlin-server,gremlin-console -DskipIntegrationTests=false -DincludeNeo4j`

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/incubator-tinkerpop/master/docs/
 TinkerPop 3.1.2 (NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Added the `:remote console` option which flips the Gremlin Console into a remote-only mode where all script evaluation is routed to the currently configured remote, which removes the need to use the `:>` command.
 * Added `allowRemoteConsole()` to the `RemoteAcceptor` interface.
 * The `:remote` for `tinkerpop.server` now includes an option to establish the connection as a "session".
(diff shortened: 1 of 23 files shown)


okram:
VOTE +1. This is baller.

dkuppitz:
It even supports multi-line scripts! Epic.

VOTE: +1

PR #261, opened 2016-03-15
Update slf4j to 1.7.19

ceefour:


spmallette:
There were a number of good fixes to SLF4j since the version we were using. This was a good upgrade. Merged.

PR #262, opened 2016-03-15
Update hadoop to 2.7.2

ceefour:


ceefour:
It's weird that hadoop-client and hadoop-hdfs do not agree on netty version. This should be fixable.

ceefour:
Root issue reported to Hadoop project: https://github.com/apache/hadoop/pull/84

steveloughran:
Why the two versions? hdfs netty went in later, MR netty hasn't (yet) updated, then there's the fear of breaking things downstream. Hadoop 2.8 splits hdfs client libs out into their own JAR; there's no dependency in netty 4 in hadoop-hdfs-client at all.

spmallette:
@okram i think you have to lead the review on this hadoop stuff.

okram:
VOTE +1. 

I merged locally (had to tweak a few other things -- docs and Manifest stuff, but works). This appears okay for `tp31` and an upmerge to `master`.

spmallette:
VOTE +1

dkuppitz:
VOTE: +1

okram:
Note that I also verified that this still works with Hadoop 2.7.1 by building the docs with Hadoop 2.7.1 running.

okram:
Note that I also just merged master/ and all is good in the (proverbial) hood.

PR #276, opened 2016-03-24
TINKERPOP-1109 Make Gremlin Console better suited for system level installs

dkuppitz:
`gremlin.sh` now handles user-defined `ext/` directories (`tinkerpop.ext`). If the user specified directory doesn't exist, it will

a) create it and
b) copy the contents of `gremlin-console/ext` to it

### DIFF
--- a/gremlin-console/src/main/bin/gremlin.sh
+++ b/gremlin-console/src/main/bin/gremlin.sh
@@ -22,6 +22,17 @@
 set -e
 set -u
 
+DIR="$( cd -P "$( dirname "$0" )" && pwd )"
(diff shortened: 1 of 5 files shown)


spmallette:
Just pushed in a minor change to not bother to default to user home since we can always expect `tinkerpop.ext` to be set or defaulted. even if it isn't it now defaults to the local `ext` directory.

Tested this manually a bunch of different ways at this point and console integration tests work well still (though i'm not sure they hit the code paths that were changed).

VOTE +1

okram:
VOTE +1

dkuppitz:
VOTE: +1

PR #281, opened 2016-03-28
Docker Integration

dkuppitz:
Provided Docker script that allows the execution of several build tasks within a Docker container.


PommeVerte:
VOTE +1

I think this is great to have around. It will streamline the workflow. 
I also think it will help with travis testing. I know travis has support for docker, I don't know how well that's integrated for custom containers but if it's possible to setup we might just get integration tests running on multiple operating systems, which would be awesome.

spmallette:
VOTE +1 - tried a couple commands. looks good

dkuppitz:
Forgot my own vote in the new PR, hence...

VOTE: +1

PR #282, opened 2016-03-28
TINKERPOP-1223: Allow jars in gremlin.distributedJars to be read from HDFS

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1223

If a `HADOOP_GREMLIN_LIBS` directory does not reference a local directory, then check for that directory in HDFS. If its there, copy it to a `java.io.tmp` directory and then process the `jars` from there.

CHANGELOG

```
* `HADOOP_GREMLIN_LIBS` can now reference a directory in HDFS and will be used if the directory does not exist locally.
```  

`mvn clean install` and tested manually via the Gremlin Console.

VOTE +1.

### DIFF
--- a/hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/process/computer/AbstractHadoopGraphComputer.java
+++ b/hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/process/computer/AbstractHadoopGraphComputer.java
@@ -211,7 +211,7 @@ public abstract class AbstractHadoopGraphComputer implements GraphComputer {
             File file = new File(localDirectory);
             if ((Boolean.valueOf(System.getProperty("is.testing", "false")) || !file.exists()) && fileSystem.exists(new Path(localDirectory)) && fileSystem.isDirectory(new Path(localDirectory))) {
                 final File tempDirectory = new File(System.getProperty("java.io.tmpdir") + "/" + hadoopGremlinLibsRemote);
-                if (!tempDirectory.exists()) assert tempDirectory.mkdir();
+                if (!tempDirectory.exists()) assert tempDirectory.mkdirs();
                 final String tempPath = tempDirectory.getAbsolutePath() + "/" + new File(localDirectory).getName();
                 final RemoteIterator<LocatedFileStatus> files = fileSystem.listFiles(new Path(localDirectory), false);
                 while (files.hasNext()) {
--- a/hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/process/computer/AbstractHadoopGraphComputer.java
+++ b/hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/process/computer/AbstractHadoopGraphComputer.java
@@ -18,6 +18,10 @@
  */
 package org.apache.tinkerpop.gremlin.hadoop.process.computer;
 
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.LocatedFileStatus;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.RemoteIterator;
 import org.apache.hadoop.util.ReflectionUtils;
 import org.apache.tinkerpop.gremlin.hadoop.Constants;
 import org.apache.tinkerpop.gremlin.hadoop.structure.HadoopGraph;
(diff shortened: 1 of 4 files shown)


hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/process/computer/AbstractHadoopGraphComputer.java
```diff
@@ -195,4 +202,27 @@ public boolean supportsDirectObjects() {
...
+
+    //////////
+
+    public static File copyDirectoryIfNonExistent(final FileSystem fileSystem, final String localDirectory) {
+        try {
+            final String hadoopGremlinLibsRemote = "hadoop-gremlin-" + Gremlin.version() + "-libs";
+            File file = new File(localDirectory);
+            if ((Boolean.valueOf(System.getProperty("is.testing", "false")) || !file.exists()) && fileSystem.exists(new Path(localDirectory)) && fileSystem.isDirectory(new Path(localDirectory))) {
+                final File tempDirectory = new File(System.getProperty("java.io.tmpdir") + "/" + hadoopGremlinLibsRemote);
+                if (!tempDirectory.exists()) assert tempDirectory.mkdir();
```
dkuppitz: `mkdir()` should be `mkdirs()` since `java.io.dir` can be set by the user and we shouldn't make assumptions about existing directory structures.

hadoop-gremlin/src/test/java/org/apache/tinkerpop/gremlin/hadoop/process/computer/AbstractHadoopGraphComputerTest.java
```diff
@@ -0,0 +1,81 @@
...
+        final FileSystem fs = FileSystem.get(new Configuration());
+        if (!new File(System.getProperty("java.io.tmpdir") + "/tmp").exists())
+            assertTrue(new File(System.getProperty("java.io.tmpdir") + "/tmp").mkdir());
+        File tempFile1 = new File(System.getProperty("java.io.tmpdir") + "/tmp/test1.txt");
+        File tempFile2 = new File(System.getProperty("java.io.tmpdir") + "/tmp/test2.txt");
+        assertTrue(tempFile1.createNewFile());
+        assertTrue(tempFile2.createNewFile());
+        assertTrue(tempFile1.exists());
+        assertTrue(tempFile2.exists());
+        if (fs.exists(new Path("target/testing")))
```
spmallette: Please use `TestHelper.makeTestDataPath`. It places test data under standard "test data" directories in `/target` which basically creates a path like: `/target/TestClassName/`. This helps avoid naming collisions and makes it really easy to find test data files in the event of failures.

hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/process/computer/AbstractHadoopGraphComputer.java
```diff
@@ -195,4 +202,27 @@ public boolean supportsDirectObjects() {
...
     }
+
+    //////////
+
+    public static File copyDirectoryIfNonExistent(final FileSystem fileSystem, final String localDirectory) {
+        try {
+            final String hadoopGremlinLibsRemote = "hadoop-gremlin-" + Gremlin.version() + "-libs";
+            File file = new File(localDirectory);
+            if ((Boolean.valueOf(System.getProperty("is.testing", "false")) || !file.exists()) && fileSystem.exists(new Path(localDirectory)) && fileSystem.isDirectory(new Path(localDirectory))) {
+                final File tempDirectory = new File(System.getProperty("java.io.tmpdir") + "/" + hadoopGremlinLibsRemote);
```
spmallette: Please replace hardcoded path separator with platform independent call to `System.getProperty("file.separator")`

dkuppitz:
VOTE: +1

hadoop-gremlin/src/test/java/org/apache/tinkerpop/gremlin/hadoop/process/computer/AbstractHadoopGraphComputerTest.java
okram: I can not do this because of how HDFS uses the local file system. This is why this test case was so hard to write. I have to use `java.io.tmpdir`.

hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/process/computer/AbstractHadoopGraphComputer.java
okram: Done.

hadoop-gremlin/src/test/java/org/apache/tinkerpop/gremlin/hadoop/process/computer/AbstractHadoopGraphComputerTest.java
okram: What I will do is use a `getClass().toString()`-directory name off of `target/` to be safe.

spmallette: i thought you could use TestHelper so i was a little sloppy with the test path. the data actually goes in `target/test-case-data/TestClassName` - can you make that one final change? sorry

spmallette:
VOTE +1

PR #284, opened 2016-03-29
TINKERPOP-1237: ProjectMap: For the Love of Die Faterland

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1237

Added `GraphTraversal.project()` which allows is like the inverse of `select()`. Instead of pulling from this path history, you are pushing into the future. While this can be done with `match()`, `ProjectStep` does not require `LABELED_PATHS` and thus, is more efficient (though, its constrained to local children `by()`-modulation).

EXAMPLE:

```
gremlin> g.V().out("created").
           project("a","b").
             by("name").
             by(in("created").count()).
           order().by(select("b"),decr).
           select("a")
==>lop
==>lop
==>lop
==>ripple
```

CHANGELOG

```
* Added `GraphTraversal.project()` to allow projecting out a `Map<String,E>` given the current traverser and an arbitrary number of `by()`-modulators.
```

Docs build, `mvn clean install` and Giraph (at `ProjectTest`) tested.

VOTE +1.

### DIFF
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -1376,6 +1376,24 @@ metrics = t.getSideEffects().get('metrics')
 
 For traversal compilation information, please see <<explain-step,`explain()`>>-step.
 
+[[project-step]]
+Project Step
+~~~~~~~~~~~~
+
+The `project()`-step (*map*) projects labeled attributes of the current object into a `Map<String,Object>`. It is similar
+to <<select-step,`select()`>>-step, save that instead of retrieving and modulating historic traverser state, it modulates
+the current state of the traverser.
+
+[gremlin-groovy,modern]
+----
+g.V().out('created').
+           project('a",'b').
+             by('name').
+             by(__.in('created').count()).
+           order().by(select('b'),decr).
+           select('a')
+----
+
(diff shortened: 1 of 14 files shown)


dkuppitz:
VOTE: +1

docs/src/reference/the-traversal.asciidoc
```diff
@@ -1376,6 +1376,24 @@ metrics = t.getSideEffects().get('metrics')
...
+~~~~~~~~~~~~
+
+The `project()`-step (*map*) projects labeled attributes of the current object into a `Map<String,Object>`. It is similar
+to <<select-step,`select()`>>-step, save that instead of retrieving and modulating historic traverser state, it modulates
+the current state of the traverser.
+
+[gremlin-groovy,modern]
+----
+g.V().out('created').
+           project('a",'b').
```
twilmes: Small typo, `'a"` -> `'a'`

twilmes:
VOTE: +1

docs/src/reference/the-traversal.asciidoc
okram: I just fixed. I built the docs, then reformatted the text and I caused that. The docs are rebuilding right now... then I will push the typo fix.

PR #287, opened 2016-04-06
Fix NPE in ObjectWritable.toString

dalaro:


okram:
There were other issues around NPE. I fixed those and added an `ObjectWritableTest`. Can you please close this PR.

dalaro:
Thanks, Marko.  I think the compareTo change that accompanied the toString fix in fa2d626951f799a0a50191f9653a17b5c18cdc20 is a regression, but the discussion about it is ongoing in https://issues.apache.org/jira/browse/TINKERPOP-1251 anyway, so I'll kill this PR.

PR #288, opened 2016-04-07
Fix TINKERPOP-1252

greedy:
It is now ensured that if there was any problem committing or rolling
back a transaction then the ThreadLocal binding to a Neo4j transaction
will be cleared.

### DIFF
--- a/neo4j-gremlin/src/main/java/org/apache/tinkerpop/gremlin/neo4j/structure/Neo4jGraph.java
+++ b/neo4j-gremlin/src/main/java/org/apache/tinkerpop/gremlin/neo4j/structure/Neo4jGraph.java
@@ -332,24 +332,22 @@ public final class Neo4jGraph implements Graph, WrappedGraph<Neo4jGraphAPI> {
 
         @Override
         public void doCommit() throws TransactionException {
-            try {
-                threadLocalTx.get().success();
+            try (Neo4jTx tx = threadLocalTx.get()) {
+                tx.success();
             } catch (Exception ex) {
                 throw new TransactionException(ex);
             } finally {
-                threadLocalTx.get().close();
                 threadLocalTx.remove();
             }
         }
 
         @Override
         public void doRollback() throws TransactionException {
-            try {
-                threadLocalTx.get().failure();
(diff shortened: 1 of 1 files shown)


spmallette:
So you're saying that there is some case where `threadLocalTx.get().close();` fails in the `finally` clause and then we don't get a nice call to `threadLocalTx.remove();`. Doesn't your change introduce a new problem though, where a fail on a call to `threadLocalTx.get().success();` or `threadLocalTx.get().failure();` would mean that  `threadLocalTx.get().close();`  would never get called?

Also, is this only a problem for 3.2.0? seems like this fix should be targetted at 3.1.2 and the tp31 branch, no?

greedy:
Yeah, that's what I saw happening. threadLocalTx.remove() was skipped because threadLocalTx.get().close() failed.

I did think about using two levels of finally blocks to guarantee that close is called. The Neo4j API documentation doesn't really say much about which of these methods can fail and in which ways. I figured that if either one of those fail then the transaction should be considered dead. From looking through the implementations when I was trying to figure out what was going on, I saw that success() and failure() just set a boolean flag so they shouldn't fail. I wouldn't have any problem with cascaded finally blocks to make sure everything is getting called; if that seems better to you, I can update the pull request to do that.

It does indeed impact the 3.1 series as that is where I actually encountered it. It should be applied to both so I marked both series as affected in the JIRA issue. I wasn't sure if I should create two pull requests: this one and one for the tp31 branch; or if a committer would just cherry-pick this fix on to tp31 if/when it lands on master.

I did run the neo4j test suite for the change and they passed.

spmallette:
> I wouldn't have any problem with cascaded finally blocks to make sure everything is getting called; if that seems better to you, I can update the pull request to do that.

It's not pretty, but I think you should. In that way, we can be sure everything gets called and if neo4j ever changes that code to provide a way to fail, we'll be covered.

> I wasn't sure if I should create two pull requests: this one and one for the tp31 branch

Please just re-target this PR to tp31 and we'll handling merging it appropriately. Thanks

greedy:
Closing this PR to open a new one targeting tp31

PR #291, opened 2016-04-15
TINKERPOP-1263 Pass SASL mechanism through to server

mike-tr-adamson:


spmallette:
Ran `$ mvn clean install && mvn verify -pl gremlin-server -DskipIntegrationTests=false -DincludeNeo4j` and all is good

VOTE +1

PommeVerte:
Hey @mike-tr-adamson .

I had a look at [`RFC 4422 section 3.2`](https://tools.ietf.org/html/rfc4422#section-3.2) and it seems to imply that `mechanism` negotiation implementations are essentially our choice. I was wondering why you decided to prepend it to the payload, rather than say, add it to the payload, or negotiate it after authentication as pointed out in the RFC (to avoid downgrade attacks for users who want to extend the server capabilities here)

mike-tr-adamson:
@PommeVerte I'm not sure what prepending it or adding it makes any difference to the negotiation. I prepended it so it was (potentially) the first element in the payload that could be checked without having to read the whole payload. As you said, it was a choice. 

I believe that the RFC says that a negotiation can happen after authentication but that doesn't provide any real protection against downgrade. I'd prefer that this sort of protection was left to 3rd party implementations in the form of additional authorization checks against mechanism usage.

PommeVerte:
Hey Mike. Sounds fair enough. I wasn't pointing fingers, just wanted to understand the thought process a bit better. 

Just to explicit what I meant by the difference between prepending and adding. The payload is constructed of a `wrapper` which is basically a `byte` and a `mimetype` at this point for binary payloads. And a `message` to be deserialized (via `GraphSONMessageSerializer`, `GryoMessageSerializer`, etc.). 
Prepending is basically adding the mechanism to the `wrapper`.
Adding would be including it in the `message` as an argument. Along side the `sasl` credential argument for example (or some other variation)

The difference here is that with the curent implementation, existing drivers will need to be upgraded to support this feature change in the payload, along the lines of the changes you made in this PR on the Java Driver. 

The first point here is that this can be an issue with drivers that separate the `wrapper` building logic from the `message` building. ie: the `wrapper` is always built the same way and reused for all operations. Adding the mechanism for _only_ the authentication `message` becomes a special use case in the `wrapper`. Depending on the implementation this can either get ugly, or be less trivial than expected. (`PHP`, `javascript` and `python` drivers fall in this category)

The second point is that if the mechanism were included as a `message` argument instead, then most drivers out there either already support this feature or have mechanisms in place to extend authentication `message` building. So we basically provide the feature keys in hand for users of those drivers. And if the drivers really have no support for what was just mentioned, it's a trivial upgrade. 

I tried to keep this clear but it's not one of my fortes. Let me know if it wasn't clear, or if there's some conflicting logic in implementing this in the message rather than the wrapper.

Cheers.

spmallette:
I didn't realize from my review that this was a breaking change to the drivers/protocol. @mike-tr-adamson do you see any problem in trying to make this a backward compatible change by implementing as @PommeVerte suggested?

spmallette:
The more that I think about this the more I think we'd make it easier for driver devs to handle this change if the modification came as @PommeVerte suggested.

mike-tr-adamson:
I have changed the ```Handler::channelRead0``` to add the mechanism as an argument to the initial SASL message. These means that the ```SaslAuthenticationHandler``` is now unchanged. 

I agree that this is a more elegant solution and allows 3rd party auth implementations to use or not use the mechanism at their leisure. Note that it is down to the auth implementation to handle the case of no mechanism being passed. It shouldn't be a requirement.

spmallette:
thanks @mike-tr-adamson but travis took a compilation error:

```text
[ERROR] /home/travis/build/apache/incubator-tinkerpop/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Handler.java:[92,35] cannot find symbol
  symbol:   method arrArg(java.lang.String,saslClient[...] null)
  location: variable messageBuilder of type org.apache.tinkerpop.gremlin.driver.message.RequestMessage.Builder
```

Is there something to fix here?

mike-tr-adamson:
@spmallette Yes, apologies. My tidying up, over-tidied the code. I've pushed again with something that will hopefully compile.

spmallette:
I re-ran all the tests as i did on the last vote - all good. I'll re-iterate my VOTE +1

PommeVerte:
This looks really good. Simple enough. VOTE +1 from me.

okram:
VOTE +1.

spmallette:
Not sure why but I had to close this PR manually. Looks like the merge went as it should so.....maybe just some broken automation.  I updated changelog and dev/upgrade docs as well.

PR #295, opened 2016-04-27
TINKERPOP-1272 Added init-tp-spark.sh to console distribution.

spmallette:
This shell script was referenced in the docs but was not contained in the console zip. Moved the source location of the script to spark-gremlin as that seemed consistent with how things were being done for external configuration files (i.e. hadoop-gremlin config stuff). 

A pretty simple change and would have CTRd but wasn't sure if moving `init-tp-spark.sh` as I did would be "bad" for some reason.

VOTE +1


okram:
Thanks for handling this. VOTE +1.

twilmes:
VOTE: +1

dkuppitz:
VOTE: +1

PR #296, opened 2016-04-27
TINKERPOP-1232: Write a tutorial demonstrating the 3 ways to write a Gremlin language variant.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1232

A tutorial demonstrating how to embed Gremlin in a host language. The tutorial explains how to do this for both Jython and Python. The reflection-based model proposed should easily map over to other languages such as Ruby, JavaScript, and PHP. As long as this model is sufficient for these languages, then I think we have a solid tutorial.

Also, I went through all the `index.asciidoc` files in our `docs/` and made the Apache TinkerPop logo HREF to http://tinkerpop.apache.org and added `*x.y.z*` so they all have a version number.

The tutorial has been published here for review:
  http://tinkerpop.apache.org/docs/3.2.1-SNAPSHOT/tutorials/gremlin-language-variants/

VOTE +1 (`bin/publish-docs.sh` worked)


spmallette:
This line doesn't seem completely accurate to me:

> Jython is a JSR-223 ScriptEngine implementation that enables the evaluation of Python on the Java virtual machine. 

Jython provides a JSR-223 implementation but isn't Jython on its own much more than that?

Something else I hadn't thought of before. We refer to all of this generated code as `gremlin-python` but there's actually two distinct sets of classes at play - one that works directly with the JVM and one that creates groovy scripts for submission over `RemoteConnection`. Seems like there should be some distinction in naming there whether at the class level or the module level. Maybe this doesn't matter for the tutorial, but does matter on implementation of TINKERPOP-1278 ?

Anyway, despite my thoughts/questions - this is an good tutorial and an important one for organizing the community into something coherent around these issues.

VOTE +1

okram:
A changed "is a" to "provides". 

Regarding naming, there are two classes: `gremlin-jython.py` (which is all JVM-based) and `gremlin-python.py` (which is all `String` construction based). Thus,we DO have a distinction in naming. I would suspect any JVM-based language is automagically using the JVM model -- e.g. Gremlin-Clojure, Gremlin-Scala, ?Gremlin-Rhino? :(, Gremlin-Jython, Gremlin-JRuby, Gremlin-Groovy...

spmallette:
> Regarding naming, there are two classes: gremlin-jython.py (which is all JVM-based) and gremlin-python.py 

oops - missed that. 

> I would suspect any JVM-based language is automagically using the JVM model 

agreed

dkuppitz:
`docker/build.sh -d` successfully finished. Everything's looking good.

VOTE: +1

PR #297, opened 2016-04-28
TINKERPOP-1268 Interactive and Executor Mode for Console

spmallette:
This PR covers:

https://issues.apache.org/jira/browse/TINKERPOP-1268
https://issues.apache.org/jira/browse/TINKERPOP-1157
https://issues.apache.org/jira/browse/TINKERPOP-1155
https://issues.apache.org/jira/browse/TINKERPOP-1156

I think that the console is working with a nice level of consistency now. We no longer have to warn people that the script they give as an init script is different than a they give to `-e` - it all executes in the same environment so no more discrepancies.

You can see some examples in the documentation for how `-e` and `-i` are used.  The shell script simplified a bit - @dkuppitz  if you see a better way to deal with the `getopts` thing please let me know. I need to track `-l` in the gremlin.sh because we use it to dynamically set some log4j stuff and other debug natured things.  `getopts` kinda feels like overkill but - it's working so.....

I tested windows and it seemed to work.

Please give it a shot and see if you can break it. I tried to test as many combinations as I could think of but i may have fell short somewhere.

VOTE + 1


dkuppitz:
Looks good to me. I prefer this approach for option parsing:

https://github.com/apache/incubator-tinkerpop/blob/master/docker/scripts/build.sh#L40-L50

However, it requires that you handle all possible options and show a usage message if an option is unknown. That probably doesn't fit well in `gremlin.sh`, as the Java side is supposed to handle all "unknown" options and you likely don't want to keep the shell script and the Java side in sync whenever an option is added / removed.

VOTE: +1

spmallette:
yeah - i thought about adding all the options to `getopts` and i did not like that idea especially since the console also supports long variants of each of the items: 

```text
bin/gremlin.sh -e script.groovy
bin/gremlin.sh --execute=script.groovy
```

was getting too complicated to try to maintain that.

spmallette:
I mean to write this earlier....note that the addition of:

```xml
        <dependency>
            <groupId>commons-cli</groupId>
            <artifactId>commons-cli</artifactId>
            <version>1.2</version>
        </dependency>
```

does affect the binary distribution for the console but it does not affect LICENSE/NOTICE.  The NOTICE file included in commons-cli is boilerplate apache and therefore does not need to be included in our NOTICE.

gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/Console.groovy
```diff
@@ -284,27 +315,98 @@ class Console {
...
+    private void executeInShell(final List<String> scriptAndArgs) {
+        final String scriptFile = scriptAndArgs[0]
         try {
-            final File file = new File(initScriptFile)
-            file.eachLine { line ->
+            // check if this script comes with arguments. if so then set them up in an "args" bundle
+            if (scriptAndArgs.size() > 1) {
+                List<String> args = scriptAndArgs.subList(1, scriptAndArgs.size())
+                groovy.execute("args = [\"" + args.join('\",\"') + "\"]")
+            }
```
pluradj: Should declare `args` as either `null` or `[]` here if none are passed in, otherwise a script allowing 0 or more args would hit an error for `args` not declared.

pluradj:
Looks good, Stephen. I reviewed the code, reviewed the docs, tested on Mac and Windows.

VOTE: +1

gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/Console.groovy
dkuppitz: That's done in [L166](https://github.com/apache/incubator-tinkerpop/pull/297/files#diff-94e9f2f26def3fb716d93349f670434aR166).

pluradj: I'm suggesting this
```
            if (scriptAndArgs.size() > 1) {
                List<String> args = scriptAndArgs.subList(1, scriptAndArgs.size())
                groovy.execute("args = [\"" + args.join('\",\"') + "\"]")
            } else {
                groovy.execute("args = []")
            }
```

so that this input groovy script
```
println args
```

doesn't fail like this
```
/bin/gremlin.sh -e /tmp/foo.groovy
Error in /tmp/foo.groovy at [1: println args] - No such property: args for class: groovysh_evaluate
groovy.lang.MissingPropertyException: No such property: args for class: groovysh_evaluate
```

PR #302, opened 2016-05-03
Bump SLF4J to 1.7.21

analytically:
Releases 1.7.19 and 1.7.20 suffer from a memory leak. See http://www.slf4j.org/news.html.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -140,7 +140,7 @@ limitations under the License.
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <skipIntegrationTests>true</skipIntegrationTests>
         <skipPerformanceTests>true</skipPerformanceTests>
-        <slf4j.version>1.7.19</slf4j.version>
+        <slf4j.version>1.7.21</slf4j.version>
         <hadoop.version>2.7.2</hadoop.version>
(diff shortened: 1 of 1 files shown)


spmallette:
Thanks for pointing this out. I think we should make this fix available on 3.1.x and 3.2.x. Can you please re-target this pull request for the `tp31` branch?  

Also please update the LICENSE files here:

* https://github.com/apache/incubator-tinkerpop/blob/tp31/gremlin-console/src/main/static/LICENSE#L227-L228
* https://github.com/apache/incubator-tinkerpop/blob/tp31/gremlin-server/src/main/static/LICENSE#L226-L228

Please also validate that the LICENSE file in the slf4j jars have not changed from the following:

* https://github.com/apache/incubator-tinkerpop/blob/tp31/gremlin-console/src/main/static/licenses/slf4j
* https://github.com/apache/incubator-tinkerpop/blob/tp31/gremlin-server/src/main/static/licenses/slf4j

If they did, please update those as well.

analytically:
Done, #303.

spmallette:
Do you still plan to re-submit this PR? What you had looked good on #303 but it didn't base off of the `tp31` branch.

analytically:
#306

PR #303, opened 2016-05-04
Bump SLF4J to 1.7.21

analytically:
Releases 1.7.19 and 1.7.20 suffer from a memory leak. See http://www.slf4j.org/news.html.


spmallette:
Thanks but you created your PR against "master" again. When you create your PR, make sure to select the "tp31" branch from the drop down.

PR #306, opened 2016-05-05
Bump SLF4J to 1.7.21 to fix memory leak.

analytically:


spmallette:
this looks good - merged to `tp31` and `master`. CTR

PR #307, opened 2016-05-07
TINKERPOP-1295 Precompile ScriptInputFormat scripts once during initialization of ScriptRecordReader

twilmes:
This update precompiles an input script and then reads the input file using the compiled script instead of repeatedly calling the engine.eval().  This should cut down on the time spent repeatedly eval-ing the input script. 

I ran a quick and dirty benchmark on my measly macbook with SparkGraphComputer, 2 workers.  `g.V().count()` on a test file with 250,000 vertices and a simple `.groovy` script to read it in.

    Average of 10 runs
    -------------------------
    before (TP 3.2.0 - engine.eval): 14975.7 ms
    after (TP-1295 - w/ compiled script): 10163.6 ms

`mvn clean install` success

VOTE: +1

### DIFF
--- a/hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/structure/io/script/ScriptRecordReader.java
+++ b/hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/structure/io/script/ScriptRecordReader.java
@@ -41,6 +41,8 @@ import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.apache.tinkerpop.gremlin.structure.util.star.StarGraph;
 
 import javax.script.Bindings;
+import javax.script.Compilable;
+import javax.script.CompiledScript;
 import javax.script.ScriptEngine;
 import javax.script.ScriptException;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


okram:
Perfect. This will come in handy with all my Friendster benchmarking that uses `ScriptInputFormat`.

Did a code review, makes perfect sense.

VOTE +1.

spmallette:
This is an interesting change. I was surprised to see that much difference between `compile()` and `eval()`. `eval()` ultimately calls compile but does cache the compile script for future use. I would have thought that would have narrowed the gap you managed to get between the two calls. There's a fair bit of code in `eval()` that gets executed with or without the cache as compared to `compile` so perhaps that has something to do with it.  As the execution here is about usage of the same script over and over there is little need for `eval()` so this change makes sense to me.

I will only note that this forces the `ScriptEngine` we use to for `ScriptRecordReader` to implement `Compileable`. Not all implementations do that and it is not a required interface to the best of my recollection. That probably doesn't matter too much as we've currently hardcoded the groovy script engine in here.

Ran `mvn clean install` to success.

VOTE +1

PommeVerte:
`mvn clean install` runs for me as well and the changes make sense. :) 

VOTE +1

PR #308, opened 2016-05-10
TINKERPOP-1297 Updated the gephi plugin to work on Gephi 0.9.x

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1297

Also Improved output of HTTP request errors to gephi. It is likely that this change will not allow Gephi to work with 0.8.x anymore.

Tested manually and ran tests with: 

```text
mvn clean install && mvn verify -pl gremlin-console -DskipIntegrationTests=false
```

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -30,6 +30,7 @@ TinkerPop 3.2.1 (NOT OFFICIALLY RELEASED YET)
 * Added tests to ensure that threaded transactions cannot be re-used.
 * `GraphFilter` helper methods are now more intelligent when determining edge direction/label legality.
 * Added `GraphFilterStrategy` to automatically construct `GraphFilters` via traversal introspection in OLAP.
+* Updated the Gephi Plugin to support Gephi 0.9.x.
 * Increased the testing and scope of `TraversalHelper.isLocalStarGraph()`.
 * Changed signature of `get_g_VXlistXv1_v2_v3XX_name` and `get_g_VXlistX1_2_3XX_name` of `VertexTest` to take arguments for the `Traversal` to be constructed by extending classes.
 * Added `VertexProgramInterceptor` interface as a general pattern for `GraphComputer` providers to use for bypassing `GraphComputer` semantics where appropriate.
--- a/docs/src/reference/gremlin-applications.asciidoc
+++ b/docs/src/reference/gremlin-applications.asciidoc
@@ -1630,18 +1630,18 @@ This plugin imports the appropriate classes for managing the credentials graph.
 Gephi Plugin
 ~~~~~~~~~~~~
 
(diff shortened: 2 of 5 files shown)


okram:
VOTE +1.

pluradj:
I was able to get the graph to render in Gephi. I'm running Gephi 0.9.1 on Mac. Applying the layout and fiddling with the viz properties worked fine.

Running the visual traversal didn't seem to do anything. I didn't see any event changes flowing through the graph. Here's is the Gremlin Console for those steps:

```
gremlin> :remote config visualTraversal graph
==>Connection to Gephi - http://localhost:8080/workspace2 with stepDelay:1000, startRGBColor:[0.0, 1.0, 0.5], colorToFade:g, colorFadeRate:0.7, startSize:3.0,sizeDecrementRate:0.33
gremlin> traversal = g.V(2).in().out('knows').has('age', gt(30)).outE('created').has('weight', gt(0.5d)).inV();null
==>null
gremlin> :> traversal
==>v[5]
==>false
```

Some other comments for the docs:
1. The default workspace is now `workspace1`. This should be updated in a couple spots.
2. In the traversal example, the trailing `;null` is important to prevent the auto-iteration, right? If so, the docs should call this out specifically as a footnote.
3. `visualTraversal` should appear in the `:remote config` parameters table.
4. There doesn't seem to be a good way to see the existing configuration values. `:remote config visualTraversal graph` displays all config values (except `visualTraversal`), but maybe it might be better if for example `:remote config workspace` with no arguments would return the current value.
5. Along the same lines, `:remote config help` might be nice to let the user know what parameters can be configured.

spmallette:
I've rebased, made some changes and force pushed...

> Running the visual traversal didn't seem to do anything. I didn't see any event changes flowing through the graph. 

You are using "g" to spawn your `Traversal`. You need to use "vg" which was created by the previous command. I made that more clear in the documentation. It is actually documented in the `visualTraversal` config option (see below)

> 1. The default workspace is now workspace1. This should be updated in a couple spots.

fixed

> 2. In the traversal example, the trailing ;null is important to prevent the auto-iteration, right? If so, the docs should call this out specifically as a footnote.

added that

> 3. visualTraversal should appear in the :remote config parameters table.

it is documented but i think there was a formatting problem with asciidoc and it was getting cut out on generation. that should be fixed in my push.

> 4. There doesn't seem to be a good way to see the existing configuration values. :remote config visualTraversal graph displays all config values (except visualTraversal), but maybe it might be better if for example :remote config workspace with no arguments would return the current value.
> 5. Along the same lines, :remote config help might be nice to let the user know what parameters can be configured.

Please create JIRA issues for these two if you think they are important. I just wanted to get Gephi working on 0.9 for purposes of this body of work.

pluradj:
Thanks for the updates. Looks good.

VOTE: +1

pluradj:
Opened https://issues.apache.org/jira/browse/TINKERPOP-1303 for the help suggestions

PR #317, opened 2016-05-20
TINKERPOP-1003 Setting up latest/current links for bins and docs.

lewismc:
This issue addresses https://issues.apache.org/jira/browse/TINKERPOP-1003

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -25,9 +25,9 @@ Documentation
 TinkerPop3 provides graph computing capabilities for both graph databases (OLTP) and graph analytic systems (OLAP).
 
 * link:http://tinkerpop.apache.org/[homepage (downloads)]
-* link:http://tinkerpop.apache.org/docs/3.2.1-SNAPSHOT/reference/[reference documentation]
-* link:http://tinkerpop.apache.org/javadocs/3.2.1-SNAPSHOT/core/[core javadoc]
-* link:http://tinkerpop.apache.org/javadocs/3.2.1-SNAPSHOT/full/[full javadoc]
+* link:http://tinkerpop.apache.org/docs/current/reference/[reference documentation]
+* link:http://tinkerpop.apache.org/javadocs/current/core/[core javadoc]
+* link:http://tinkerpop.apache.org/javadocs/current/full/[full javadoc]
 
 Building and Testing
 ~~~~~~~~~~~~~~~~~~~~
@@ -42,7 +42,7 @@ The zip distributions can be found in the following directories:
 . `gremlin-server/target`
 . `gremlin-console/target`
 
-Please see the link:http://tinkerpop.apache.org/docs/3.2.1-SNAPSHOT/dev/developer/#_contributing[CONTRIBUTING.asciidoc] file for more detailed information and options for building, test running and developing TinkerPop.
+Please see the link:http://tinkerpop.apache.org/docs/current/dev/developer/#_contributing[CONTRIBUTING.asciidoc] file for more detailed information and options for building, test running and developing TinkerPop.
 
 Get Started
 ~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


spmallette:
@dkuppitz since this touches stuff you know the most about i think it would be useful to have your feedback on this one.

lewismc:
ack, thanks anyone for review.

dkuppitz:
Well, it looks good to me, but unfortunately I can't test it, since @Humbedooh never managed to set up my SVN account (no write permission). I'm tired of asking him over and over again.

lewismc:
:(

spmallette:
weird - i never noticed that you replied to my post in jira @lewismc - i'll just say that i basically agree with what you wrote there. we can push off the SEO stuff to another ticket. i'm fine with that. i guess what i'm still concerned about here is the situation where someone does a `publish-docs.sh` for 3.2.1-SNAPSHOT, and then someone else turns around and does one for 3.1.3-SNAPSHOT. In that case, what's in "current" is just constantly being shifted around. I also don't think we we want to promote a SNAPSHOT as "current'. We should make "current" our most recent production release that we primarily promote on the web site - so our "current", if we did this today would be a copy of 3.2.0-incubating. 

I think that there should be something in the script that doesn't bother to update "current" if the version is SNAPSHOT. I don't know what we do about the actual release versions to prevent 3.1.6 from one day overwriting 3.3.0. See the problem? any thoughts?

dkuppitz:
Oh right, I didn't think about snapshot and parallel release lines. Now I also remember how I wanted to solve it: we actually shouldn't get rid of the version directories altogether. We should keep doing that but then add a bit of extra logic that  makes a recursive copy of the latest version directory (maybe symlinks would work as well?).

```
rm -rf current/
cp -R ${LATEST} current/
```

Determining `LATEST` is easy once you have all the version directories:

```
daniel@cube /tmp/test $ mkdir 3.1.1 3.1.2 3.1.2-SNAPSHOT
daniel@cube /tmp/test $ mkdir 3.2.0 3.2.1 3.2.2 3.2.3-SNAPSHOT
daniel@cube /tmp/test $ ls
3.1.1  3.1.2  3.1.2-SNAPSHOT  3.2.0  3.2.1  3.2.2  3.2.3-SNAPSHOT
daniel@cube /tmp/test $ ls | grep -v SNAPSHOT | sort -rV | head -n1
3.2.2
daniel@cube /tmp/test $
```

lewismc:
@dkuppitz so do you want to make a quick update to this script so we can refer to ${LATEST} as the highest release before the SNAPSHOT?

dkuppitz:
Here's what I think it should be: https://gist.github.com/dkuppitz/a24127c0d82e8bfc8e68b52466ffc689
Note, that I've only added [L61-L73](https://gist.github.com/dkuppitz/a24127c0d82e8bfc8e68b52466ffc689#file-publish-docs-sh-L61-L73), everything else is unchanged.

@spmallette Can you give that a try?

spmallette:
just to update what's happening here - i tried to test, but we have doc generation failures on master atm. there is discussion in the tinkerpop hipchat on what could be wrong. @dkuppitz was able to isolate the failure to this commit: 

https://github.com/apache/incubator-tinkerpop/commit/c890cebad5231cdab8744878bebb4ea363689a3a

once that's fixed i can resume testing on this.

spmallette:
ok - finally was able to test in full. the work @dkuppitz  did in his gist is good. now we have:

http://tinkerpop.apache.org/docs/current/reference/

@lewismc given that the fix for this ultimately came from kuppitz, i think he should go ahead and just do a PR from his end. thanks for your effort on this and helping to limit the scope of this issue to simplify our way to a solution. i guess we can now re-visit your other PR that helped spawn this one and get that merged. I will probably look to deal with that tomorrow. If you feel like closing this from your end, cool - otherwise i'll just close it tomorrow. thanks again!

lewismc:
> @lewismc given that the fix for this ultimately came from kuppitz, i think he should go ahead and just do a PR from his end. thanks for your effort on this and helping to limit the scope of this issue to simplify our way to a solution. i guess we can now re-visit your other PR that helped spawn this one and get that merged. I will probably look to deal with that tomorrow. If you feel like closing this from your end, cool - otherwise i'll just close it tomorrow. thanks again!

Thank you v much for review and discussion. Glad to hear that the issue is driving towards a resolution :)

PR #320, opened 2016-05-25
TINKERPOP-1312 .count().is(0) is not properly optimized

dkuppitz:
Tweaked `RangeByIsCountStrategy`. The patterns ...

* `....count().is(0)`
* `....count().is(lt(1))`
* `....count().is(lte(0))`

... are now replaced by `not(...)`.

VOTE: +1

### DIFF
--- a/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/RangeByIsCountStrategyTest.java
+++ b/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/RangeByIsCountStrategyTest.java
@@ -22,6 +22,7 @@ import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
 import org.apache.tinkerpop.gremlin.process.traversal.TraversalEngine;
 import org.apache.tinkerpop.gremlin.process.traversal.TraversalStrategies;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.__;
+import org.apache.tinkerpop.gremlin.process.traversal.step.filter.NotStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.filter.RangeGlobalStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.filter.TraversalFilterStep;
 import org.apache.tinkerpop.gremlin.process.traversal.util.DefaultTraversalStrategies;
@@ -36,7 +37,16 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 
-import static org.apache.tinkerpop.gremlin.process.traversal.P.*;
+import static org.apache.tinkerpop.gremlin.process.traversal.P.eq;
+import static org.apache.tinkerpop.gremlin.process.traversal.P.gt;
+import static org.apache.tinkerpop.gremlin.process.traversal.P.gte;
+import static org.apache.tinkerpop.gremlin.process.traversal.P.inside;
+import static org.apache.tinkerpop.gremlin.process.traversal.P.lt;
+import static org.apache.tinkerpop.gremlin.process.traversal.P.lte;
+import static org.apache.tinkerpop.gremlin.process.traversal.P.neq;
+import static org.apache.tinkerpop.gremlin.process.traversal.P.outside;
+import static org.apache.tinkerpop.gremlin.process.traversal.P.within;
+import static org.apache.tinkerpop.gremlin.process.traversal.P.without;
 import static org.junit.Assert.assertEquals;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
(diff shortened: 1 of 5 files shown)


okram:
Can you add some more complex tests? For instance:

```
out().count().is(0).as("a")
```

Also, `RangeByIsCountStrategy` doesn't have `applyPrior()` of `AdjacentToIncidentStrategy`. Does that matter? That is:

```
out().count().is(0) ->
outE().count().is(0) ->
not(outE())
```

We want `not(outE())` NOT `not(out())`. Can you verify that such compilations happen?

Also, why do you make a distinction between computer and standard engines for this strategy?

I think you should copy that pattern in the other strategy tests that ensure proper `__()` compilations. We can go crazy on adding patterns to make sure compilations are as expected. E.g.:

```
  @Parameterized.Parameters(name = "{0}")
    public static Iterable<Object[]> generateTestParameters() {

        return Arrays.asList(new Traversal[][]{
                {__.outE().count(), __.outE().count()},
                {__.bothE("knows").count(), __.bothE("knows").count()},
                {__.properties().count(), __.properties().count()},
                {__.properties("name").count(), __.properties("name").count()},
                {__.out().count(), __.outE().count()},
                {__.in().count(), __.inE().count()},
                {__.both().count(), __.bothE().count()},
                {__.out("knows").count(), __.outE("knows").count()},
                {__.out("knows", "likes").count(), __.outE("knows", "likes").count()},
                {__.filter(__.out()), __.filter(__.outE())},
                {__.where(__.not(__.out())), __.where(__.not(__.outE()))},
                {__.where(__.out("knows")), __.where(__.outE("knows"))},
                {__.values().count(), __.properties().count()},
                {__.values("name").count(), __.properties("name").count()},
                {__.where(__.values()), __.where(__.properties())},
                {__.and(__.out(), __.in()), __.and(__.outE(), __.inE())},
                {__.or(__.out(), __.in()), __.or(__.outE(), __.inE())},
                {__.out().as("a").count(), __.outE().count()},   // TODO: is this good?
                {__.where(__.as("a").out("knows").as("b")), __.where(__.as("a").out("knows").as("b"))}});
    }
```

dkuppitz:
> why do you make a distinction between computer and standard engines for this strategy's test case?

I was wondering why we initially chose another pattern for this test class; I couldn't figure it out, but then still just followed the pattern that we started with. I'll change that.

dkuppitz:
Updated. Regarding the order of `RangeByIsCountStrategy` and `AdjacentToIncidentStrategy`, it doesn't really matter. We either go:

`out().count().is(0) -> outE().count().is(0) -> not(outE())`

or:

`out().count().is(0) -> not(out()) -> not(outE())`

In fact it's the former, but either way we would always end up with the same step sequence.

okram:
Sorry to be a pest, but could you add more complex tests still! :)

```
outE().count().is(0).inV()
both().both().count().is(0).as("a")
```

For the latter, should it be `not(both().both())` ? What are the semantics in longer chains -- also, what about local children and global children. All your test cases are basically "the same" in terms of just `out().*`.

dkuppitz:
> For the latter, should it be `not(both().both())`?

Since we only look at it in conjunction with `count().is(something)`, we don't need to differentiate between single steps and longer chains.

`both().not(both())` is the same thing as `not(both().both())`.

It makes a difference if something comes after `is(0)`, but in that case we don't use `not()`:

`both().both().count().is(0).both() -> both().both().limit(1).count().is(0).both()`

okram:
When all is in place, can you provide a benchmark on Grateful Dead graph of some query in master/ vs. this branch and their runtimes? So to be certain its all as expected. Even provide a `explain()` of the traversal you are benchmarking so we see how it compiled?

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/RangeByIsCountStrategyTest.java
```diff
@@ -142,6 +142,7 @@ public void doTest(final Traversal traversal, final Traversal optimized) {
                     {__.out().count().is(without(2, 6, 4)), __.out().limit(6).count().is(without(2, 6, 4))},
                     {__.map(__.count().is(0)), __.map(__.limit(1).count().is(0))},
                     {__.flatMap(__.count().is(0)), __.flatMap(__.limit(1).count().is(0))},
+                    {__.filter(__.count().is(0)), __.filter(__.not(__.identity()))},
```
okram: `filter(not(_.identity))` the same as `not()` ?

okram:
Sweet benchmark. VOTE +1.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/RangeByIsCountStrategyTest.java
dkuppitz: It is the same, but IMO it's not RangeByIsCountStrategy's job to optimize that.

spmallette:
Nice tests and benchmark stuff  - VOTE +1

dkuppitz:
Not sure why the merge was not recognized automatically, but it's merged into `tp31` and `master`.

PR #321, opened 2016-05-26
TINKERPOP-1308 Allow Serialization to "reference"

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1308

Added a "lite" serializer option for gryo in Gremlin Server. This serializer will write graph elements to "referenced" rather than "detached". The default serialization behavior for Gremlin Server has not changed as a result of this work.  Using the "lite" serialization is not the default, but an option to be made on the request.

Everything is good with:

```text
```

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/incubator-tinkerpop/master/docs/
 TinkerPop 3.2.1 (NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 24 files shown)


PommeVerte:
Had a look through the changes. Noticed you add a 'lite' `path` Serializer, does `tree` get serialized properly as well? I kind of expect it to since it leverages `vertex`/`edge` serialization but thought I would double check that with you.

Appart from that one question it's looking tight.

spmallette:
yes - i would expect it to though i don't know that i explicitly tried. as it is implemented at the serializer level, the behavior should apply to graph elements embedded in other objects (like `Tree`). so is that a +1 from you? :)

PommeVerte:
Yeah that's a +1 from me :)

dkuppitz:
VOTE: +1

PR #322, opened 2016-05-27
TINKERPOP-1196 Fix for Result.one() which could have blocked indefinitely.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1196

Result.one() sometimes would hang under rare but possible conditions. This change should prevent that from happening now as it removes the chance of await() futures from being created while a flush of existing waiting futures is occurring. When that happened it seemed to create the possibility where an await() future could get created but never been completed.

I would have CTR'd this as it is a small change but I was hoping for reviewers to do:

```text
```

so as to validate that the problem is resolved on their systems. This problem didn't typically occur on mine - @okram this seemed to happen a lot to you during test runs so it would be great if you could give it a try.

VOTE +1


dkuppitz:
20 runs in a row, all of them succeeded.

VOTE: +1

pluradj:
I inspected the code, and got 5 successful runs on my Mac.

VOTE: +1

PR #326, opened 2016-06-02
TINKERPOP-1322 Added ConfigurationCustomizerCompiler

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1322

This `CustomizerCompiler` implementation makes it possible to set low-level configurations in the `GremlinGroovyScriptEngine` on the Groovy `CompilerConfiguration` which ultimately controls script compilation settings used by the classloader. The most likely feature that someone would use here is the setting of the `ScriptBaseClass` which will let executing scripts inherit some base functionality.

All good with:

```text
```

Also performed a manual test with Gremlin Server that worked nicely.

VOTE +1


okram:
VOTE +1.

dkuppitz:
VOTE: +1

PR #329, opened 2016-06-03
TINKERPOP-1318: use mockito-core instead of mockito-all to avoid hamcrest conflict

pluradj:
https://issues.apache.org/jira/browse/TINKERPOP-1318

### DIFF
--- a/gremlin-test/pom.xml
+++ b/gremlin-test/pom.xml
@@ -51,11 +51,23 @@ limitations under the License.
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>${junit.version}</version>
+            <exclusions>
+                <exclusion>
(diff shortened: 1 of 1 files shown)


okram:
Do integration tests pass? ... that will be all I care about. @spmallette will have deeper thoughts on the matter as he uses Mockito.

pluradj:
Yes, I ran integration tests with `-DincludeNeo4j`

okram:
VOTE +1.

PommeVerte:
This change is simple enough and shouldn't impact anything (short of an incompatibility between mockito and the version of hamcrest here, but the tests pass so it should be ok.)

VOTE +1

pluradj:
VOTE: +1

PR #334, opened 2016-06-09
TINKERPOP-1331 HADOOP_GREMLIN_LIBS can only point to local file system

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1331

It's now possible to have DFS paths in `HADOOP_GREMLIN_LIBS`.

VOTE: +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/incubator-tinkerpop/master/docs/
 TinkerPop 3.2.1 (NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Allow DFS paths in `HADOOP_GREMLIN_LIBS`.
 * Added a safer serializer infrastructure for use with `SparkGraphComputer` that uses `KryoSerializer` and the new `GryoRegistrator`.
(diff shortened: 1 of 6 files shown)


spmallette:
Note that the travis integration is broken here because of the migration from incubator (i.e. the link is just broken). One job passed and one failed - here's the passing run:

https://travis-ci.org/apache/tinkerpop/builds/136484066

The failing run was due to a failure in the gremlin server tests, but i'm guessing that to be an internal failure of travis and not something related to this PR since we did get a green build out of this.

VOTE +1

okram:
I don't see any test cases. I would like to see some tests in {{HadoopGremlinSuite}} that shows that using {{hdfs://}} and {{file://}} in {{HADOOP_GREMLIN_LIBS}} works. Also, that its backwards compatible with the {{x/y/z:a/b/c:blah/blah}} as previous.

dkuppitz:
The failing Travis test run contains a massive amount of exceptions in `RemoteGraphProcessComputerTest` which ultimately led to:

```
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).

The job has been terminated
```

I don't think that's in any way related to my PR.

okram:
Do the reference docs need to be updated with your new work? If so, can you update the docs in the PR too (please)?

dkuppitz:
Added a note in `implementations-hadoop.ascidoc`.

okram:
VOTE +1

PR #336, opened 2016-06-15
TINKERPOP-1320 fix GremlinGroovyScriptEngineFileSandboxTest resource loading

pluradj:
https://issues.apache.org/jira/browse/TINKERPOP-1320
Passed `mvn clean install -DincludeNeo4j` and integration tests cleanly.
May be CTR worthy. Let me know.
VOTE: +1

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoWriter.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoWriter.java
@@ -131,7 +131,7 @@ public final class GryoWriter implements GraphWriter {
     public void writeProperty(final OutputStream outputStream, final Property p) throws IOException {
         final Output output = new Output(outputStream);
         writeHeader(output);
-        kryo.writeObject(output, DetachedFactory.detach(p, true));
+        kryo.writeObject(output, DetachedFactory.detach(p));
         output.flush();
     }
 
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoWriter.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoWriter.java
@@ -131,7 +131,7 @@ public final class GryoWriter implements GraphWriter {
     public void writeProperty(final OutputStream outputStream, final Property p) throws IOException {
         final Output output = new Output(outputStream);
         writeHeader(output);
-        kryo.writeObject(output, DetachedFactory.detach(p));
+        kryo.writeObject(output, DetachedFactory.detach(p, true));
         output.flush();
     }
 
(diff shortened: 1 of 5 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoWriter.java
```diff
@@ -131,7 +131,7 @@ public void writeVertexProperty(final OutputStream outputStream, final VertexPro
     public void writeProperty(final OutputStream outputStream, final Property p) throws IOException {
         final Output output = new Output(outputStream);
         writeHeader(output);
-        kryo.writeObject(output, DetachedFactory.detach(p, true));
```
okram: Is there any way that this may lead to a serialization format that is not backwards compatible?

spmallette: that's weird as a standalone issue - @pluradj  what did this change have to do with TINKERPOP-1320 ?

pluradj: Let me pull out this one change and open up a separate issue for it. I'll expand more in the issue on this one.

spmallette: cool - something doesn't make sense in that {{GryoWriter}} - very weird looking.

spmallette:
I didn't re-test this but both travis builds are good and the code looks right. Just a side note - @pluradj I didn't review this but note that we have a bit of an ugly reality where `TestHelper` has a twin over in `gremlin-core/test`. There are test classes in there that need that kind of functionality and since `gremlin-core` can't depend on `gremlin-test` I just copied the file. :grimacing: Note sure if any of your changes needs to be propagated over there or not, but probably something you should look at before you finalize this one and merge. Pending that little investigation:

VOTE +1

pluradj:
Can you expand on that more, re: the duplication on `TestHelper.java`? I'd bumped into the two classes previously but haven't spent any time trying to figure out the reason why there are two of them.

spmallette:
`gremlin-core` just have tests that need the some of the same exact functions that are in `gremlin-test`, but the dependencies are such that `gremlin-core` can't use the `TestHelper` in `gremlin-test` and `gremlin-test` can't use the `TestHelper` in `gremlin-core` so there are two of them. 

I suppose we could move `TestHelper` out of `gremlin-core/test` to the main source code as a utility??? then there could be one. Or we publish `gremlin-core/test` as an artifiact which i kinda hate. Or we create a new dependency all together for `gremlin-core` and `gremlin-test` to depend on. All of the answers sorta stink i guess. Up until now, copying one file seemed to smell the best of all of them....

spmallette:
@pluradj is there anything left to do on this? were my remarks on `TestHelper` clear? are you just waiting for a final vote at this point?

pluradj:
@spmallette I'll get this finished up before the weekend

dkuppitz:
VOTE: +1

PR #339, opened 2016-06-16
TINKERPOP-939 Added docs for Neo4j HA configuration.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-939

I manually tested this with a local 3 node cluster both in Gremlin Console and in Gremlin Server. Everything seems to work fine. I basically just wrote some basic docs for how to setup but since TinkerPop is really just a pass through for configuration options, most of the docs for "how" are on the Neo4j side.

I don't know that there is a way to do testing for this on our current framework or if there is a ton of value to doing so as this is all pass through - we don't even use a separate "HA Graph instance" like we did in Blueprints. Thoughts welcome on that matter....

Personally, I think the docs are enough for now to close this old issue. If we really want tests we could open a new issue to write some "integration" tests that do some basic checks and would stand apart from the standard framework. So based on that....

VOTE +1

### DIFF
--- a/docs/src/reference/implementations-neo4j.asciidoc
+++ b/docs/src/reference/implementations-neo4j.asciidoc
@@ -259,3 +259,77 @@ gremlin.neo4j.directory=/tmp/neo4j
 gremlin.neo4j.conf.node_auto_indexing=true
 gremlin.neo4j.conf.relationship_auto_indexing=true
 ----
+
+High Availability Configuration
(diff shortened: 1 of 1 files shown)


okram:
Beautiful. Thank you for doing testing that and finally getting that ticket done. Crazy it worked the whole time we just didn't know... VOTE +1.

dkuppitz:
VOTE: +1

PR #340, opened 2016-06-16
add RawExpression to gremlin_python

leifurhauks:
As described on dev@tinkerpop.apache.org thread gremlin_python GLV

### DIFF
--- a/docs/src/reference/gremlin-variants.asciidoc
+++ b/docs/src/reference/gremlin-variants.asciidoc
@@ -157,3 +157,37 @@ g.V().both()[1]
 g.V().both().name
 ----
 
+Inserting arbitrary expressions with Gremlin-Python
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To make use of APIs provided by your graph system vendor that aren't part of
+the gremlin traversal API, you can use `RawExpression`. `RawExpression` allows
+you to construct arbitrary expressions that collect bindings. When you pass
+strings and bindings to the `RawExpression` constructor, the arguments will
+be concatenated in the script that is sent to gremlin-server.
+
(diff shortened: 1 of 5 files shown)


okram:
How do we use this again? This is just for the `GeoPoint`-style adding classes? Can you provide an example so we can add it to the docs. See:

https://github.com/apache/tinkerpop/blob/TINKERPOP-1278/docs/src/reference/gremlin-variants.asciidoc

leifurhauks:
> This is just for the GeoPoint-style adding classes?

More generally, it can be used to create expressions that make use of APIs, presumably provided by the graph vendor, that aren't strictly part of gremlin. My immediate use case is to call static methods on Titan's Geoshape class.

I'll add an example to the asciidoc.

leifurhauks:
@okram , I realized the current implementation of bindings as dicts is breaking 2/3 compatibility. It also makes it bit hairy to extract just the key in `RawExpression`. 

I would suggest using a tuple (e.g. `('symbol', value)` ) or a simple class that just has two data members. May I revise the bindings implementation?

okram:
Please. Use tuple though instead of introducing a new class. Thanks.

spmallette:
@leifurhauks do you know why travis is not happy? the build took some errors it seems. Can you please look into that as you push your next change?

leifurhauks:
I believe the failures are due to the fact that the aenum package (introduced in 8ee2904 ) is not installed in the CI environment. As a result, the gremlin_python module errors on import.

Since 146f476, the test suite adds the contents of PYTHONPATH to sys.path, but aenum still needs to be downloaded and built. One possibility would be to create a virtualenv in the working dir, install aenum to that virtualenv, and add it to PYTHONPATH.

spmallette:
ah - didn't realize that build was hosed in the TINKERPOP-1278 branch itself.

PR #343, opened 2016-06-17
restructured gremlin-python package

davebshow:
This PR restructures the Gremlin-Python package to unify packages, modules, and classes under the gremlin_python namespace as discussed on the mailing list. It also provides a minor fix to ensure Python 2/3 compatibility, as well as relative imports for convenience and flexibility.


okram:
When I run the test suite, I get failures.

```
  File "src/main/jython/gremlin_python/groovy_translator.py", line 23, in <module>
    from .gremlin_python import P
ValueError: Attempted relative import in non-package
```

Easy solution I can do or would you like to fix up your PR?

davebshow:
This would probably be because the way tests are set up the modules are getting executed as a script instead of as a package. I don't know how to fix this off the top of my head--I haven't familiarized myself with how tests are being run, and I unfortunately don't have time to figure it out right now. Ideally, Gremlin-Python would be installed using pip and used normally as a package.

okram:
This is how the Python modules are loaded into Jython.

https://github.com/apache/tinkerpop/blob/TINKERPOP-1278/gremlin-variant/src/test/java/org/apache/tinkerpop/gremlin/python/PythonProvider.java#L39-L62

Recommendations to make it more "package like" ?

okram:
This just changed to here:

https://github.com/apache/tinkerpop/blob/TINKERPOP-1278/gremlin-variant/src/test/java/org/apache/tinkerpop/gremlin/python/JythonScriptEngineSetup.java#L36-L59

davebshow:
Sorry, I hurried this because I was leaving town and have not had much access to the internet. Due to the many conflicts it will be simpler to close and create a new PR. I will make sure all tests pass before submitting.

spmallette:
Thanks @davebshow - things are still moving around a little bit. @okram do you agree that dave should hold off for a little bit until we stabilize a bit further? perhaps a week or so?

okram:
Yea -- its a little crazy on that branch right now. It will stabilize (be near the end) by the end of next week.

PR #344, opened 2016-06-22
gremlin_python: generate/build/deps before tests

leifurhauks:
Code generation is triggered by Maven before tests are run; dependencies
(currently aenum, requests) are installed, and the generated package is built
into ${project.build.testOutputDirectory}/Lib, which is added to the
classpath for the tests via the python.home property. Jython tests
import the gremlin_python package instead of importing its modules
directly.

### DIFF
--- a/gremlin-variant/pom.xml
+++ b/gremlin-variant/pom.xml
@@ -100,6 +100,39 @@
         </testResources>
         <plugins>
             <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>exec-maven-plugin</artifactId>
+                <version>1.2.1</version>
+                <executions>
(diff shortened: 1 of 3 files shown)


leifurhauks:
I think this should fix the CI test failures.

On the subject of the python dependencies (aenum and requests), should they be bundled in the jar? aenum has a 3-clause BSD [license](https://bitbucket.org/stoneleaf/aenum/src/c9dc3b8fc3e3755a919ba4357ab9ed9b22131af1/aenum/LICENSE?at=default&fileviewer=file-view-default), and requests uses the ASL.

okram:
Can you add the license boilerplate? 

BTW: Just pulled your patch and building locally now.

PR #349, opened 2016-06-28
TINKERPOP-1349: RepeatUnrollStrategy should unroll loops while maintaining equivalent semantics.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1349

`RepeatUnrollStrategy` is a Standard-only strategy that will unroll a `repeat()` into a linear form if and only if it has a known loop amount (e.g. `times(2)`). That is `repeat(out()).times(2)` becomes `out().barrier().out().barrier()`. The `barrier()` insertions are necessary as `repeat()` itself is a barrier step and also, the benefit of bulking is leveraged. `RepeatUnrollStrategy` removes the need for `RepeatStep`, `RepeatEndStep`, and `LoopTraversal`.

`RepeatUnrollStrategyTest` ensures that unrolling is faster than not-unrolling. Here are some runtimes. The first item in the Pair is the runtime and the second item is the amount of unique objects being processed. With more unique objects, the faster it gets. Note that at 10 unique elements (tiny traversal), sometimes rolled is faster than unrolled and vice versa (i.e. they are equal).

```
rolled:[0.22843720299999998, 10]	unrolled:[0.174451055, 10]
rolled:[0.514145509, 100]		unrolled:[0.212532769, 100]
rolled:[1.4944357259999999, 1000]	unrolled:[0.441714987, 1000]
rolled:[28.502658626, 10000]		unrolled:[12.936112649999998, 10000]
```

Also in this ticket, I fixed a bug in `BranchStep` where children were not being integrated and thus, susceptible to clone-problems. Moreover, I added a timing test to `IdentityRemovalStrategyTest`. @spmallette, do you know how to make a test run once and only once in a `Parameterized` test case? If so, can you update `IdentityRemoveStrategyTest` and `RepeatUnrollStrategyTest` accordingly?

---

```
CHANGELOG

* Added `RepeatUnrollStrategy` to linearize a `repeat()`-traversal if loop amount is known at compile time.
* Fixed a bug in `BranchStep` around child integration during `clone()`.
* Fixed a bug in `AbstractStep` around label set cloning.
```

VOTE +1.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalStrategies.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalStrategies.java
@@ -30,6 +30,7 @@ import org.apache.tinkerpop.gremlin.process.traversal.strategy.optimization.Matc
 import org.apache.tinkerpop.gremlin.process.traversal.strategy.optimization.OrderLimitStrategy;
 import org.apache.tinkerpop.gremlin.process.traversal.strategy.optimization.PathProcessorStrategy;
 import org.apache.tinkerpop.gremlin.process.traversal.strategy.optimization.RangeByIsCountStrategy;
+import org.apache.tinkerpop.gremlin.process.traversal.strategy.optimization.RepeatUnrollStrategy;
 import org.apache.tinkerpop.gremlin.process.traversal.strategy.verification.ComputerVerificationStrategy;
 import org.apache.tinkerpop.gremlin.process.traversal.strategy.verification.StandardVerificationStrategy;
 import org.apache.tinkerpop.gremlin.process.traversal.util.DefaultTraversalStrategies;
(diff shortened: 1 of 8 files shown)


okram:
NOTE: This can not go into `tp31/` because `NoOpBarrierStep` is new to TinkerPop 3.2.0.

dkuppitz:
VOTE: +1

spmallette:
Travis is failing - looks like an issue from Rat - missing an Apache header in `TraversalStrategyPerformanceTest`. 

@okram I don't think there's a way to get a Parameterized test to run a single test once. It pretty much treats the whole test case as parameterized when you use that feature. Using `Enclosed` as you did is the only way to make that work afaik.

@pietermartin do you intend to vote on this one? you had a fair number of comments/questions on the JIRA ticket....

pietermartin:
@spmallette No I do not intend to vote. The questions were about understanding why the barrier steps were needed and then to understand why barrier/bulking has such a significant performance impact. I have not read the `RepeatUnrollStrategy` code nor ran any tests for it.

spmallette:

VOTE +1

dkuppitz:
Same here, `docker/build.sh -t -i -n` reported success over night. All changes after my first vote were included in this test run.

PR #353, opened 2016-06-29
TINKERPOP-1341 Add missing classes to GryoRegistrator

dalaro:
https://issues.apache.org/jira/browse/TINKERPOP-1341

GryoRegistrator is missing registrations for three classes registered
in GryoSerializer:

* void
* Void
* CompactBuffer[]

This commit adds all three to GryoRegistrator.getExtraRegistrations().

So far, CompactBuffer[] is the only omission known to cause practical
problems, as detailed by Dylan Bethune-Waddell on the mailing list and
in the referenced issue.  It's not clear whether omitting Void and
void causes practical problems, but they're trivial, so the cost
associated with registering them is tiny.  Might as well.


okram:
VOTE +1.

okram:
Crazy. I never new `void.class` existing in Java. ?!

spmallette:
VOTE +1

dkuppitz:
VOTE: +1

PR #354, opened 2016-06-30
TINKERPOP-1354 Added more invalid binding keys to Gremlin Server validation

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1354

These "invalid" keys are reserved terms for Gremlin Server as they are statically imported enums and shouldn't be used as binding keys. You get some less than easy to understand error messages if those keys are used. 

I would have CTR'd but wanted to see if anyone had other suggestions for additional validations at play. Also, this change is "breaking" in the sense that users who were somehow successfully using some of these newly reserved keys on previous versions (not fully sure if that was even possible) will have to update their code. I don't think this is a massive problem for someone to fix, so while "breaking" it doesn't seem massively detrimental and shouldn't be widely problematic.

Builds with `mvn clean install -DskipTests && mvn verify -pl gremlin-server -DskipIntegrationTests=false`

VOTE +1


robertdale:
BUILD SUCCESS

dkuppitz:
VOTE: +1

okram:
VOTE +1

PR #355, opened 2016-07-04
TINKERPOP-1355 Design HasContainer for extension

BrynCooke:


dkuppitz:
I noticed a flaw while reading through the code. It was not introduced by this PR, but it's probably a good time to get it from the table.

```
gremlin> g.V().hasId(within(new ArrayList()))
0
Display stack trace? [yN] N
gremlin> g.V().hasId(without(new ArrayList()))
0
Display stack trace? [yN]
```

A `HasContainer` with an empty collection will always throw an exception because of [line 58 in HasContainer.java](https://github.com/BrynCooke/incubator-tinkerpop/blob/ca8a4f9fb8a9bc1d69aa10870d75076b8fa906b5/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/HasContainer.java#L58), where we simply assume that a collection will always contain at least one element.

dkuppitz:
I just implemented my proposed changes. Now we no longer get exceptions, but unexpected results:

```
gremlin> g.V().hasId(within(new ArrayList()))
==>v[1]
==>v[2]
==>v[3]
==>v[4]
==>v[5]
==>v[6]
gremlin> g.V().hasId(without(new ArrayList()))
==>v[1]
==>v[2]
==>v[3]
==>v[4]
==>v[5]
==>v[6]
```

This should probably go into a separate ticket.

BrynCooke:
Also the conversion of the collection to an array is dangerous. Who knows what the collection is actually backed by? It should use an iterator and get the first element if it exists.

dkuppitz:
Yea, lots of issues in this code area. `enforceHomogenousCollectionIfPresent` also has the potential to throw NPE's (if the first element of a non-empty collection is `null`). I will create another ticket for those issues.

For this PR:

VOTE: +1

okram:
Clean changes. VOTE +1.

spmallette:
VOTE +1

PR #357, opened 2016-07-07
TINKERPOP-1278 : Make implicit relative imports explicit

leifurhauks:
Implicit relative imports break python3 compatibility. This commit
replaces implicit relative imports with explicit relative imports, which
are supported by both python 2 (>= 2.6) and python 3.


okram:
I tried to merge this but there were two problem:

1.) the PR references old code and thus, lots of conflicts.
2.) graph_traversal and traversal are auto-generated and thus editing the .py does nothing.
3.) Jython complains in JythonScriptEngineSetup (tests don't pass)

If you can fix this PR accordingly, I'd be happy to merge.

leifurhauks:
Brought this branch up to date and fixed the code generators.

leifurhauks:
Added WIP to the title of the PR while I sort out some silly mistakes I made with the conflict resolution...

leifurhauks:
Okay, should be good now

PR #361, opened 2016-07-10
TINKERPOP-1360: compare vertex property ids as Long

pluradj:
https://issues.apache.org/jira/browse/TINKERPOP-1360

Verified with:

```
mvn clean install && mvn verify -q -DskipIntegrationTests=false -pl spark-gremlin
```

And via Docker
```
docker/build.sh -t -i -n
```

@spmallette @dkuppitz Please try out multiple times in your environments since the error only happens 1 out of 4 scenarios (based on 2 different `RANDOM.nextBooolean()` calls). I manually forced each scenario in by setting specific boolean values in `HadoopGraphProvider` and `SparkHadoopGraphProvider`.

### DIFF
--- a/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/PropertiesTest.java
+++ b/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/PropertiesTest.java
@@ -100,9 +100,8 @@ public abstract class PropertiesTest extends AbstractGremlinProcessTest {
             ids.add(vertexProperty.id());
             assertEquals("name", vertexProperty.key());
             assertEquals(convertToVertex(graph, vertexProperty.value()).values("name").next(), vertexProperty.value());
-            assertEquals(convertToVertex(graph, vertexProperty.value()).properties("name").next().id(), vertexProperty.id());
-            // assertEquals(convertToVertex(graph, vertexProperty.value()), vertexProperty.element());
-            // assertEquals(convertToVertexId(graph, vertexProperty.value()), vertexProperty.element().id());
+            // compare the ids as longs. assumes modern graph uses numeric ids only (long/integer).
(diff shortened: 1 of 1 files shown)


okram:
VOTE +1. I think this could have been CTR'd as well given the simplicity of the change.

pluradj:
Perhaps, but [CTR](https://issues.apache.org/jira/browse/TINKERPOP-1360?focusedCommentId=15369381&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15369381) got us to this point. I'd rather have this reviewed than have something [ominous](https://lists.apache.org/thread.html/ff501e4391bd2e180f46891873b2fa7833c697742dd5dadd551bcca4@%3Cdev.tinkerpop.apache.org%3E) happen :unamused:

dkuppitz:
Dunno, is this really a fix or do we simply ignore a flaw? I mean why do we have to cast result types at all?

pluradj:
In the failing scenario, you have GraphSONRecordReader pulling in the graph and using Integer as the id type (via Jackson in GraphSONReader). It is getting compared against a graph from ToyGraphInputRDD which uses Long as the id type (via TinkerGraph default id manager). `equals()` comparison of the same value `6` fails because the types are different.

```java.lang.AssertionError: expected: java.lang.Integer<6> but was: java.lang.Long<6>```

Switching to `tinkerpop-modern-typed.json` instead of `tinkerpop-modern.json` doesn't help the situation since the id type isn't stored.

dkuppitz:
I see.. I guess it can be considered to be a "fix" then ;)

VOTE: +1

robertdale:
Java is brain damaged when comparing numbers. Don't blame JSON.

pluradj:
VOTE: +1

pluradj:
Merged, closing https://github.com/apache/tinkerpop/commit/90a4cdad129d0eecbd67da626d40c21872b596e0

PR #363, opened 2016-07-20
TINKERPOP-1379 remove excess bulk in tail buffer

pluradj:
https://issues.apache.org/jira/browse/TINKERPOP-1379


pluradj:
I've come up with a cleaner test case. I'll close this PR and open a new one.

PR #366, opened 2016-07-30
TINKERPOP-1379 remove excess bulk in tail buffer

pluradj:
https://issues.apache.org/jira/browse/TINKERPOP-1379

VOTE: +1

### DIFF
--- a/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/TailTest.java
+++ b/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/TailTest.java
@@ -28,6 +28,7 @@ import org.junit.runner.RunWith;
 
 import static org.apache.tinkerpop.gremlin.LoadGraphWith.GraphData.MODERN;
 import static org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.__.both;
+import static org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.__.in;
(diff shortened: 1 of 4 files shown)


pluradj:
Test case `g.V().repeat(__.in().out()).times(3).tail(7).count()` against the modern graph returns `-62`. Fails all the way back to 3.0.x.

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/TailTest.java
```diff
@@ -120,6 +123,19 @@ public void g_V_repeatXbothX_timesX3X_tailX7X() {
         assertEquals(7, counter);
     }
 
+    /** Scenario: Global scope, using repeat (excess BULK) */
+    @Test
+    @LoadGraphWith(MODERN)
+    public void g_V_repeatXin_outX_timesX3X_tailX7X_count() {
+        final Traversal<Vertex, Long> traversal = get_g_V_repeatXin_outX_timesX3X_tailX7X_count();
+        printTraversalForm(traversal);
+        long count = 0L;
```
okram: This is as easy as:

```
checkResults(Collections.singleList(7L), traversal)
```

It will also verify `hasNext() == false` at the end.

okram:
VOTE +1. 

Note that I added a comment for how to make the test case simpler.

spmallette:
All good with a quick `mvn clean install`

VOTE +1  - pending fix @okram suggested for the test code and what looks like the need for a rebase given conflicts that are likely on changelog.

PR #367, opened 2016-08-02
TINKERPOP-1350 was never quite fixed in 3.1.3.

spmallette:
Unfortunately, more testing revealed that the fix for the problem was only masked by the changes in 3.1.3. 

Changed response encoding to not use the session executor when the session has an error condition it is trying to serialize. This should be fine as there is no need to serialized an error condition as part of a transaction and thus no need to have the session thread to do it. That in turn frees up the worker executor to serialize and cancel long run jobs in the session. Removed recommendations for submitting parallel requests on a session from docs.

Works in full with `mvn clean install && mvn verify -pl gremlin-server -DskipIntegrationTests=false`

VOTE +1


twilmes:
Code and doc updates look good.

VOTE: +1

okram:
VOTE +1.

spmallette:
Rebased and force pushed to resolve merge conflicts on tp31. merged

PR #368, opened 2016-08-03
TINKERPOP-1394: fixed dev list and jira links

robertdale:


spmallette:
merged via CTR - thanks

PR #370, opened 2016-08-03
Added missing feature requirement

seanbarzilay:
This test uses MetaProperties but didn't require the feature.


spmallette:
Thanks for re-targeting - merged via CTR

PR #371, opened 2016-08-03
Added new recipe for Traversal Induced Values.

spmallette:
Just a documentation addition. I learned that it's not so good to CTR these as there is sometimes good feedback on Gremlin code that should be incorporated and a PR is the best way to go about gathering that feedback.

VOTE +1

### DIFF
--- a/docs/src/recipes/traversal-induced-values.asciidoc
+++ b/docs/src/recipes/traversal-induced-values.asciidoc
@@ -1,0 +1,83 @@
+////
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+////
+[[traversal-induced-values]]
+Traversal Induced Values
+------------------------
+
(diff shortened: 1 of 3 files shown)


spmallette:
Please ignore the failing test:

 IdentityRemovalStrategyTest$PerformanceTest>TraversalStrategyPerformanceTest.shouldBeFaster:107 null

that's an issue raised elsewhere: https://issues.apache.org/jira/browse/TINKERPOP-1390

twilmes:
I might be missing something silly here, but It looks like you need to commit `traversal-induced-values.asciidoc`.  I'm only seeing the updates to the changelog and index on this branch.

spmallette:
Sorry - forgot to add the file @twilmes - fixed up the commit and force pushed.

twilmes:
Looks good, nice choice for a recipe.

VOTE: +1

okram:
VOTE +1.

This got me to thinking that we really need to start to extend other steps (beyond the `Mutating`) to support traversal parameterization.

docs/src/recipes/traversal-induced-values.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+<1> Find the "marko" `Vertex` and label it as "marko".
+<2> Traverse out on the "knows" edges to the adjacent `Vertex` and label it as "person".
+<3> Filter the incoming "person" vertices. It is within this filter, that the traversal induced values are utilized.
+The inner `select` grabs the "marko" vertex and the current "friend". The `by` modulator extracts the "age" from both
+of those vertices which yields a `Map` with two keys, "marko" and "friend", where the value of each is the "age".
+<4> The `Map` produced in the previous step can then be filtered with `where` to only return a result if the "friend"
+age is greater than the "marko" age. If this is successful, then the `filter` step from the previous line will succeed
+and allow the "friend" vertex to pass through.
+
+This traversal could also be written declaratively with `match` step as follows:
```
pluradj: Should there be any discussion of the possible performance differences between the 2 approaches? Or is that mostly a system-specific discussion?

spmallette: I thought about going into more detail there, but opted not to. I kinda thought that if we wanted that sort of content that it didn't belong in a recipe but in reference docs. It seemed a little out of scope of the recipe.

PR #373, opened 2016-08-04
TINKERPOP-1151 Made a number of changes to logging dependencies.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1151

Log4j is now generally a test dependency except for gremlin-server and gremlin-console where they need to be shipped as part of the binary distribution. In that case, they are optional scope for those who for some reason depend on those libs.

I tested this a bunch of different ways:

* `mvn clean install` shows the expected log messages 
* `mvn clean install -DskipIntegrationTests` shows the expected log messages
* Gremlin Console displays log messages that aren't hidden by the default config and changes to that config allow logs to show in full
* Gremlin Server displays expected log messages
* Looked at the zip distributions and they obviously had the appropriate log4j jars
* Tested builds of both archetype outputs and logging is good within those

Note that I didn't do this work for tp31 because this isn't really a true bug fix and it messes with dependencies which didn't seem right given how we've characterized the work to occur on that branch going forward.

Anything else i missed where logging would be an issue? if not then VOTE +1 for me


spmallette:
Rebased and force pushed to resolve conflicts on master.

twilmes:
Looking good.  VOTE:  +1

PommeVerte:
Code looks straight forward. VOTE +1

PR #375, opened 2016-08-05
Finished renaming distribution from 'apache-' to 'apache-tinkerpop-'

joshsh:
Fixed gremlin.sh and assembly descriptor, updated documentation


spmallette:
hey @joshsh - wow....totally missed all those.  this would have all bit me on release day so thanks for that. 

can i ask that you please retarget your pull request for the tp31 branch? we need these fixes in the 3.1.x line of code where applicable. Or perhaps it's a PR to each tp31 and master? seems like i see a few fixes that only apply on master. what do you think?

joshsh:
Sure.  Here: https://github.com/apache/tinkerpop/pull/376

Note that the distros for giraph-gremlin, spark-gremlin, and hadoop-gremlin don't use the apache-tinkerpop- prefix.  I haven't changed them.

PR #377, opened 2016-08-11
TINKERPOP-1397 StarGraph filtering with self edges

dalaro:
See:

* old PR #372
* JIRA issue https://issues.apache.org/jira/browse/TINKERPOP-1397


okram:
This solution makes sense to me -- no new `SelfEdge` class and we now have a test case. Thanks.

VOTE +1.

dkuppitz:
VOTE: +1

spmallette:
VOTE +1

PR #379, opened 2016-08-14
added simple WebSocketRemoteConnection

davebshow:
Also: Made RemoteConnection into actual abstract base class, fixed typo in remote_graph.

This PR starts the process of adding a websocket driver to the gremlin-python GLV. It is basically working, except I am not sure how you submit the `Bytecode` object and `traversal_source` to the server, since the `Bytecode` object does not serialize to JSON. As is, the the `RESTRemoteConnection` has the same problem. I'm sure there are just a few things missing of which I am not aware. Also, I am not sure how the `traversal_source` should be passed. My initial thought was to pass it as an alias, but this is obviously not correct. Anyway, let me know how to proceed with this and I will move forward.


spmallette:
Build failure in travis is due to rat checks - new websocket_remote_connection.py needs apache license header

PR #380, opened 2016-08-15
TINKERPOP-1402 Added IoBuilder.onMapper() method.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1402

Deprecated IoBuilder.registry() in the process of doing that. This should provide greater flexibility to Graph providers who can now not only supply a custom IoRegistry but also access other non-generalized features of a Mapper.Builder instance.

Tested with `mvn clean install`. I did a one-time test before changing TinkerGraph to stop using the deprecated `registry` method and the build worked fine so I think that even though we can't directly test the old `registry` method anymore, it still should be working ok if provides elect to stick with it.

VOTE +1


spmallette:
Note that the build failure in travis is that `IdentityRemovalStrategyTest$PerformanceTest>TraversalStrategyPerformanceTest.shouldBeFaster:118 null` nonsense

dkuppitz:
VOTE: +1

twilmes:
VOTE: +1

PR #381, opened 2016-08-16
TINKERPOP-1405 - profile() doesn't like withPath()

twilmes:
Fixed a small bug in `StandardVerificationStrategy` that caused verification to fail when `withPath` was used in conjunction with `ProfileStep`.  Updated `StandardVerificationStrategyTest` to test traversals with `ProfileStep` and added a test to confirm verification failed on a reducing barrier nested within a repeat while I was at it.

This issue did not arise in TP31 so I took a look and the TP31 `StandardVerificationStrategy` does not have the `ProfileStep` and some other verification verification logic in it that is in master.  Because of this, I pointed this PR only at master but can re-point to TP31 and add the profile step verification in if you guys think that makes sense.

The Travis build appears to have stalled out while downloading artifacts.  Tests passed local with `mvn clean install`

```
[INFO] BUILD SUCCESS
[INFO] Total time: 24:25 min
[INFO] Finished at: 2016-08-16T08:43:01-05:00
[INFO] Final Memory: 140M/843M
```

VOTE: +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -33,6 +33,7 @@ TinkerPop 3.2.2 (NOT OFFICIALLY RELEASED YET)
 * Added methods to retrieve `Cluster` settings in `gremlin-driver`.
 * Fixed a severe bug in `SubgraphStrategy`.
 * Deprecated `SubgraphStrategy.Builder.vertexCriterion()/edgeCriterion()` in favor of `vertices()/edges()`.
+* Fixed a small bug in `StandardVerificationStrategy` that caused verification to fail when `withPath` was used in conjunction with `ProfileStep`.
 
 [[release-3-2-1]]
 TinkerPop 3.2.1 (Release Date: July 18, 2016)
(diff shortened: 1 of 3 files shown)


okram:
Nice find. I really do hate `RequirementsStep`.

VOTE +1.

dkuppitz:
Lots of calls to ` traversal.asAdmin().getEndStep()`. Maybe it's smarter to call it only once..? More like this:

```
// The ProfileSideEffectStep must be the last step, 2nd last step when accompanied by the cap step,
// or 3rd to last when the traversal ends with a RequirementsStep.
Step endStep;
if (TraversalHelper.hasStepOfClass(ProfileSideEffectStep.class, traversal) &&
        !((endStep = traversal.asAdmin().getEndStep()) instanceof ProfileSideEffectStep ||
                (endStep instanceof SideEffectCapStep && endStep.getPreviousStep() instanceof ProfileSideEffectStep) ||
                (endStep instanceof RequirementsStep && (
                        endStep.getPreviousStep() instanceof SideEffectCapStep ||
                        endStep.getPreviousStep() instanceof ProfileSideEffectStep)))) {
    throw new VerificationException("When specified, the profile()-Step must be the last step or followed only by the cap()-step and/or requirements step.", traversal);
}
```

^ I also removed all the negations, which makes the chained conditions a bit easier to read IMO.

But all in all this PR looks good to me.

VOTE: +1

twilmes:
Good suggestions Daniel.  I pushed the cleaned up code.  Tests were good on my side but I'll wait for Travis to finish up and merge if all is good.

PR #383, opened 2016-08-18
TINKERPOP-1246 Added help message to console

robertdale:


okram:
Hm. Not a fan. :| +0.

PommeVerte:
I'm personally ok with the concept. It seems pretty mainstream and will definitely help some folk. It's the kind of thing that you don't really pick up from reading the documentation (at least I don't) I know it's there from keeping up with the mailing list but otherwise I wouldn't have a clue.

PR #384, opened 2016-08-18
TINKERPOP-1037 Made life more colorful

robertdale:
Stephen did all the hard work.
- Added option -C to disable ANSI colors just in case.
- Colorized the gremlin ascii art.
- Added bold red to some, but probably not all, error messages
- Snuck in the help text for TINKERPOP-1246

### DIFF
--- a/gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/Console.groovy
+++ b/gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/Console.groovy
@@ -72,7 +72,7 @@ class Console {
     public static final String PREFERENCE_ITERATION_MAX = "max-iteration"
     private static final int DEFAULT_ITERATION_MAX = 100
     private static int maxIteration = DEFAULT_ITERATION_MAX
-    
+
     public static final String PREF_GREMLIN_COLOR = "gremlin.color"
     def gremlinColor = { Preferences.get(PREF_GREMLIN_COLOR, "reset") }
 
@@ -102,13 +102,13 @@ class Console {
 
     public static final String PREF_RESULT_PROMPT_COLOR = "result.prompt.color"
     def resultPromptColor = { Preferences.get(PREF_RESULT_PROMPT_COLOR, "reset") }
-    
+
     public static final String PREF_EMPTY_RESULT_IND = "empty.result.indicator"
     def emptyResult = { Preferences.get(PREF_EMPTY_RESULT_IND, "null") }
 
     public static final String PREF_INPUT_PROMPT = "input.prompt"
     def inputPrompt = { Preferences.get(PREF_INPUT_PROMPT, "gremlin>") }
-    
+
     public static final String PREF_RESULT_PROMPT = "result.prompt"
     static def resultPrompt = { Preferences.get(PREF_RESULT_PROMPT, "==>") }
 
@@ -127,7 +127,7 @@ class Console {
      */
     @Deprecated
     public Console(final String initScriptFile) {
-        this(new IO(System.in, System.out, System.err), initScriptFile.size() != null ? [initScriptFile] : null, true)
+        this(new IO(System.in, System.out, System.err), initScriptFile.size() != null ? [initScriptFile]: null, true)
     }
 
     public Console(final IO io, final List<String> scriptAndArgs, final boolean interactive) {
@@ -143,12 +143,17 @@ class Console {
 
         maxIteration = Preferences.get(PREFERENCE_ITERATION_MAX, DEFAULT_ITERATION_MAX.toString()).toInteger()
         Preferences.addChangeListener(new PreferenceChangeListener() {
-            @Override
-            void preferenceChange(PreferenceChangeEvent evt) {
-                if (evt.key == PREFERENCE_ITERATION_MAX)
-                    maxIteration = Integer.parseInt(evt.newValue)
-            }
-        })
+                    @Override
+                    void preferenceChange(PreferenceChangeEvent evt) {
+                        if (evt.key == PREFERENCE_ITERATION_MAX && null != evt.newValue)
+                        try {
+                            maxIteration = Integer.parseInt(evt.newValue)
+                        } catch (NumberFormatException e) {
+                            io.out.println(ansiRender(errorColor,"Unable to convert '${evt.newValue}' to integer. Using default ${DEFAULT_ITERATION_MAX}"))
+                            maxIteration = DEFAULT_ITERATION_MAX
+                        }
+                    }
+                })
 
         final Mediator mediator = new Mediator(this)
 
@@ -310,10 +315,10 @@ class Console {
             }
         }
     }
-    
+
     def colorizeResult = { object ->
         if (object instanceof Vertex) {
-             return ansiRender(vertexColor, object.toString())
+            return ansiRender(vertexColor, object.toString())
         } else if (object instanceof Edge) {
             return ansiRender(edgeColor, object.toString())
         } else if (object instanceof Iterable) {
@@ -433,10 +438,14 @@ class Console {
             if (!interactive) System.exit(1)
         }
     }
-    
+
     def ansiRender = { color, text -> Ansi.ansi().render(String.format("@|%s %s|@", color.call(), text)).toString() }
 
     public static void main(final String[] args) {
+
+//        ExpandoMetaClass.enableGlobally()
+        org.apache.tinkerpop.gremlin.console.Preferences.expandoMagic()
+
         // need to do some up front processing to try to support "bin/gremlin.sh init.groovy" until this deprecated
         // feature can be removed. ultimately this should be removed when a breaking change can go in
         IO io = new IO(System.in, System.out, System.err)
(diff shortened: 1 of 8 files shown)


spmallette:
that looks like all the same code i've written in the past...............................

robertdale:
Like I said, you did all the hard work. I just added color.  Try this in your console:  

gremlin> "@|bold,green Hello |@ "
==>Hello

robertdale:
Screenshot:  screenshot_2016-08-18_13-59-58

dkuppitz:
Hm, now the console has some colors, that's nice, but `TINKERPOP-1037` is actually about output coloring (which means to me, that results should have a bit of syntax coloring, right?). What do we need to do in order to have colored results (e.g. Strings in red, numbers in bllue, etc. et.c)?

okram:
Whoa. This is really cool. However, I think we should have a `.gremlin_preferences` style file where we do this:

```
gremlin.console.errorMessage=red
gremlin.console.prompt=white
gremlin.console.character=green
etc.
```

Then people can tweak to their liking and if they don't like it, they can go NOT have that file and it will be as it was before.

alaam:
+1 for having this be a preferences file

robertdale:
There are some things that are out of our control because they are inherited from groovy.

robertdale:
Would you really want a separate preferences (properties) file or rather reuse the existing preferences mechanism (:set gremlin.console.errorMessage red) ?

spmallette:
+1 for use of `:set`

okram:
+1 for `:set`. Look at our other `sets` to get a consistent naming convention, please.

robertdale:
screenshot_2016-08-18_18-54-19

## Color codes:
-    BLACK
-         RED
-         GREEN
-         YELLOW
-         BLUE
-         MAGENTA
-         CYAN
-         WHITE

### Background Colors
-         BG_BLACK
-         BG_RED
-         BG_GREEN
-         BG_YELLOW
-         BG_BLUE
-         BG_MAGENTA
-         BG_CYAN
-         BG_WHITE

### Attributes
-         RESET
-         INTENSITY_BOLD
-         INTENSITY_FAINT
-         ITALIC
-         UNDERLINE
-         BLINK_SLOW
-         BLINK_FAST
-         BLINK_OFF
-         NEGATIVE_ON
-         NEGATIVE_OFF
-         CONCEAL_ON
-         CONCEAL_OFF
-         UNDERLINE_DOUBLE
-         UNDERLINE_OFF
-         BOLD
-         FAINT

spmallette:
nice @robertdale - a final request from my end would be documentation. perhaps a word or two added to the docs about these settings. I guess the reference docs might be the best place for that:

https://github.com/apache/tinkerpop/blob/f35f733663029701af538fc527e1f5ab563eb07d/docs/src/reference/gremlin-applications.asciidoc#console-commands

then maybe an announcement of the feature in upgrade docs for users:

https://github.com/apache/tinkerpop/blob/f35f733663029701af538fc527e1f5ab563eb07d/docs/src/upgrade/release-3.2.x-incubating.asciidoc#upgrading-for-users

and if you want to really be complete an entry in changelog:

https://github.com/apache/tinkerpop/blob/f35f733663029701af538fc527e1f5ab563eb07d/CHANGELOG.asciidoc#tinkerpop-322-not-officially-released-yet

@dkuppitz brought up the full scope of the ticket, but i think we can save the bigger picture for some other time. at least the pattern for colors is in place and working.

I still haven't tested this out myself or reviewed the code just yet to see what you did differently than me to get this to work. My mind is still kinda reeling from that.  :disappointed:

robertdale:
I think a mockup of the desired coloring effect would be nice to have.  Currently elements just have toString called on them. I added a vertex.color and edge.color for demonstration purposes. Here's what it would look like. Let me know if that's worth committing.
screenshot_2016-08-18_21-00-22

dkuppitz:
Awesome! I don't like the color choice, but it's good to see that it works. How did you do it? Did you overwrite the respective `toString()` methods?

robertdale:
I'm definitely not the right person to pick a color scheme. Open for suggestions or we can figure it out later. Committed the vertex, edge, path coloring so you can see.

dkuppitz:
Oh nice, I like that.

VOTE: +1

robertdale:
Attached groovy script that prints the color palette.  usage: groovy jansi-colors.txt 

[jansi-colors.txt](https://github.com/apache/tinkerpop/files/427016/jansi-colors.txt)

alaam:
+1

okram:
Thoughts:

1. I think there is a bit of an overdose of options here. Keep it simple.

2. We have to realize that not everyone has a black background as their terminal. I use a light brown. Some people use white. Thus, I don't think the default should be to use color. You can `:set` colors if you want and perhaps we even have "standard themes" -- `:set black-background-standard` (or whatever the naming convention that we are using for other `:set` commands so we are consistent in our naming of parameters.

If you guys want to go with edge/vertex/etc. being able to be colored, please don't have that be part of the "standard templates." Keep things clean and simple and not a "disco club" coming at you.

robertdale:
Disco disco!  ANSI is enabled by default. This is actually not a change. It was this way before this PR.  No color scheme by default. If you want colors, you have to set them yourself. Added colorizing for String, Number, and values within Iterables.

okram:
Great. What do you think about `:set color.template=black-background`, `:set color.template=white-background`, etc. And then from there, those "MACRO" command will trigger a series of set error, message, gremlin, etc.

robertdale:
I think I'll leave templating for another jira ticket. :satisfied:   Currently, themes (really any `:set` commands) can be saved into files and loaded with `:load my-gremlin.theme`.

robertdale:
I'm done with changes. Carry on.

okram:
VOTE +1.

pluradj:
@robertdale did you test this on Windows?

spmallette:
@pluradj sucks the wind out of the PR - hahaha

pluradj:
Good news, it works on Windows.

gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/Console.groovy
```diff
@@ -66,12 +72,48 @@ class Console {
     public static final String PREFERENCE_ITERATION_MAX = "max-iteration"
     private static final int DEFAULT_ITERATION_MAX = 100
     private static int maxIteration = DEFAULT_ITERATION_MAX
+    
+	public static final String PREF_GREMLIN_COLOR = "gremlin.color"
+    def gremlinColor = { Preferences.get(PREF_GREMLIN_COLOR, "reset") }
+
```
pluradj: Got a mix of tab and space indenting going on here. The [dev docs](http://tinkerpop.apache.org/docs/current/dev/developer/#_code_style) don't call for one over the other, but most Java/Groovy files in the repo use spaces only.

PommeVerte:
awesome!! VOTE: +1 from me as well

pluradj:
The colorization stuff looks good.

One small problem I've run into is that you can't set a prompt string that contains spaces. For example, I tried `:set result.prompt 'result => '` and got this error message:

```Command ':set' requires arguments: <name> [<value>]```

Setting the preference containing spaces from `groovysh` works fine. It appears to be specific to Gremlin Console, and it is reproducible in 3.2.1 without this PR. Would be fine to handle this bug as a separate JIRA.

**Comment by pluradj** [comment]
VOTE: +1

File: gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/GremlinGroovysh.groovy [hunk 2]
```diff
@@ -45,12 +46,13 @@ class GremlinGroovysh extends Groovysh {
...
-        if (args.size() == 0) return null
+        final List<String> linetokens = parseLine(l)
+        if (linetokens.size() == 0) return null
 
-        def cmd = registry.find(args[0])
+        def cmd = registry.find(linetokens[0])
 
-        if (cmd != null && args.size() > 1 && parsedArgs != null) {
+        if (cmd != null && linetokens.size() > 1 && parsedArgs != null) {
+            List<String> args = CommandArgumentParser.parseLine(line, parsedArgs == null ? 1 : -1)
```
robertdale: groovysh added this line 1 month after this code was written.

robertdale:
The previous commit to fix :set with quoted spaces did break :remote commands because the parser stripped the quotes. Since :set command is the only one affected, I made a quick hack for it while leaving everything else alone. If this is not acceptable, just let me know and I'll back it out.

spmallette:
someone might want to run:

```
mvn clean install -pl gremlin-console -DskipIntegrationTests=false
```

before merge (now that there are sufficient votes to do so)

pluradj:
I retested with the latest commits. Setting a preference value with spaces worked, remotes worked nicely with `conf/remote-objects.yaml`. Nice work @robertdale.

`mvn clean install -pl gremlin-console -DskipIntegrationTests=false` was successful.

One other thought here even though I've already voted... it might be better if the preference keys were all prefixed with `gremlin.` since Groovy preferences are [globally shared](https://github.com/apache/groovy/blob/master/src/main/org/codehaus/groovy/tools/shell/util/Preferences.java#L34) with other groovysh-based consoles. That preference problem may have hit us before with [`interpreterMode`](https://github.com/thinkaurelius/titan/issues/1172).

robertdale:
@pluradj I noticed that too. I was thinking that metaclass could be used to intercept calls to groovy shell's Preferences class and gremlin console could have its own Java Preferences node. This would give it full separation from normal 'groovysh'.  I believe that currently max-iteration is the only gremlin-specific pref. So if we do decide to use namespaces, then max-iteration should be deprecated but supported with and without namespace.

PR #390, opened 2016-08-25
GraphSON 2.0 Deser tweaks and improvements

newkek:
There's 2 commits: 

- the first brings a massive improvement to the GraphSON 2.0 deserialization, avoiding a lot of useless copies done on the JSON parser when parsing to detect a type in the payload. There's in almost every case no copy done anymore, thus reducing the memory impact and improves performances.
- second fixes a bug I had on my machine that was causing a StackOverflow exception when deserializing a `Tree` from GraphSON2.0. The issue has been pinned down to understanding that there could be a never-ending recursive reference when creating a `JavaType` to be indexed in the GraphSONTypeIdResolver. All detailed in comments in the code but now I have a special case for when a `Tree` is registered that force a non-cyclic reference type. It is still unclear to me why this happens to me but not all the time for @spmallette for example. In any case this fixes it for me and should be fixing for everybody..

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONTypeDeserializer.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONTypeDeserializer.java
@@ -115,12 +115,13 @@ public class GraphSONTypeDeserializer extends TypeDeserializerBase {
             if (jsonParser.getCurrentToken() == JsonToken.START_OBJECT) {
                 buf.writeStartObject();
                 String typeName = null;
-                boolean valueCalled = false;
+                boolean valueDetected = false;
+                boolean valueDetectedFirst = false;
 
                 for (int i = 0; i < 2; i++) {
                     String nextFieldName = jsonParser.nextFieldName();
                     if (nextFieldName == null) {
-                        // empty map or less than 2 fields.
+                        // empty map or less than 2 fields, go out.
                         break;
                     }
                     if (!nextFieldName.equals(this.propertyName) && !nextFieldName.equals(this.valuePropertyName)) {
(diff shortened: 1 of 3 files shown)


PommeVerte:
Is the difference in behavior on `Tree` between both of you because of some lingering older version of jackson (shaded) in your local maven repo? Thought this should have been fixed a while back and travis hasn't been chocking so far?
Also, are we certain `Tree<? extends Element>` is going to fit all needs? I don't have much visibility here, dunno how this plays into other implementations.

Cheers.

newkek:
@PommeVerte I've seen indeed some issues on the Jackson GH issue tracker that were looking like what I see with `Tree` but all were resolver in Jackson 2.7.6, and even when removing my local "gremlin-shaded" jar and updating the Jackson databind to 2.7.6 I still see the issue on my computer... Have you had that kind of problem already?

PommeVerte:
Hmmm. I haven't seen this crop up since it was fixed by upgrading Jackson to `2.7.2`. Though originally my issue was related to the serializing and not deserializing so there may be something else at play. 

What's weird is that in travis' environment this doesn't seem to be an issue. 
Have you tried with the docker image? Do you still get errors there?

newkek:
Seems like it is only on my machine... I think Stephen tried on the docker image ?
Yes what I have seen so far is it happens at the moment we call `deserializationContext.findContextualValueDeserializer(typeFromId)` in `GraphSONTypeDeserializer` where the `typeFromId` is the type that has been constructed at init time with `TypeFactory.defaultInstance().constructType(Tree.class)`. I also see that for some reason I don't get, it happens to me only when the [Enums types](https://github.com/apache/tinkerpop/blob/TINKERPOP-1278/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONModule.java#L131-L140) from the Traversal serialzers have been constructed as well. If I comment those lines, I've got no problem when doing constructType(Tree.class) like all the rest..

spmallette:
i merged the first of these two commits via cherry pick. let's sit on the other one for a bit.

spmallette:
I think I understand the Tree issue better (not the solution, just the context of the PR). Merged the second commit as well. Made note of the limitation to deal with later.

PR #392, opened 2016-08-29
TINKERPOP-1419 : Throw a proper exception when no available host in a SessionedClient

davidclement90:
When a SessionedClient is initialized and no host is available in the cluster,
A proper exception is thrown instead of an IndexOutOfBoundsException


spmallette:
Thanks for this. Merged via CTR to tp31 and then master.

PR #393, opened 2016-08-29
Add pytest-runner in order to properly build gremlin-python dependencies.

davebshow:
Use `pytest-runner` in order to invoke `pytest` using the `setup.py test` command and properly build dependencies. Otherwise, tests will fail unless all dependencies are installed locally. This PR also specifies versions for `tornado` and `aenum`, as well as removing the custom `PyTest` command from `setup.py` (unneeded due to `pytest-runner`).

### DIFF
--- a/gremlin-python/pom.xml
+++ b/gremlin-python/pom.xml
@@ -323,9 +323,8 @@
                                 <configuration>
                                     <outputDirectory>${project.build.testOutputDirectory}</outputDirectory>
                                     <libraries>
-                                        <param>aenum</param>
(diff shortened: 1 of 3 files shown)


davebshow:
@spmallette I think when you run the tests in the virtualenv you may have trouble with the site-package test suites being invoked (specifically wheel). To remedy this, I added lib/lib64 to the patterns that should be avoided in test discovery. Because setting the norecursedirs option resets the defaults, I included the standard defaults as well. Hopefully this will avoid further problems.

PR #396, opened 2016-09-01
TINKERPOP-1383 publish-docs.sh might publish to current too early

dkuppitz:
Added a separate script to update the "current" docs.


dkuppitz:
CTR'ed.

PR #400, opened 2016-09-06
Added JitPack Maven repository to Grape configuration

szarnyasg:
https://issues.apache.org/jira/browse/TINKERPOP-1430

### DIFF
--- a/docker/resources/groovy/grapeConfig.xml
+++ b/docker/resources/groovy/grapeConfig.xml
@@ -9,6 +9,7 @@
       <ibiblio name="codehaus" root="http://repository.codehaus.org/" m2compatible="true"/>
       <ibiblio name="central" root="http://central.maven.org/maven2/" m2compatible="true"/>
       <ibiblio name="java.net2" root="http://download.java.net/maven/2/" m2compatible="true"/>
+      <ibiblio name="jitpack" root="https://jitpack.io/" m2compatible="true"/>
       <ibiblio name="hyracs-releases" root="http://obelix.ics.uci.edu/nexus/content/groups/hyracks-public-releases/" m2compatible="true"/>
       <ibiblio name="local" root="file:${user.home}/.m2/repository/" m2compatible="true"/>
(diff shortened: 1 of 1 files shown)


spmallette:
Thanks @szarnyasg 

@dkuppitz how has this worked along the way? is it because we had jbcrypt in our local .m2?

robertdale:
I would say yes and probably through building the project.  I had to add the repo to my own project pom. That updates my local .m2 repo. Once that has it, grapes can pull from there.

dkuppitz:
Hmm, good question. I'm not an expert when it comes to dependency management, but it sounds like a reasonable explanation to me. However, from time to time I tend to delete my local .m2 directory, so it's a bit surprising that I never ran into this issue.

robertdale:
Do you also clean your groovy/grapes dir?  artifacts will be copied there from maven. afterwards, it won't reference the maven artifacts unless they are also missing from grapes.

dkuppitz:
Rarely. But here's another thing: The Docker build always starts with a clean environment - no .m2, no grapes - and it always worked for me.

robertdale:
Ok. Didn't notice it mentioned docker and I can't speak to that. However, it is very repeatable on other systems.  Given no other configuration...
```
rm -rf ~/.groovy
rm -rf ~/.m2
$ ./bin/gremlin.sh 

         \,,,/
         (o o)
-----oOOo-(3)-oOOo-----
plugin activated: tinkerpop.server
plugin activated: tinkerpop.utilities
plugin activated: tinkerpop.tinkergraph
gremlin> :install org.apache.tinkerpop neo4j-gremlin 3.2.1
==>Error grabbing Grapes -- [unresolved dependency: com.github.jeremyh#jBCrypt;jbcrypt-0.4: not found]
gremlin> 
```

szarnyasg:
Closed: not a real issue, refer to the docs in http://tinkerpop.apache.org/docs/current/reference/#gremlin-applications.

PR #405, opened 2016-09-12
TINKERPOP-1391 issue with where filter

dkuppitz:
Enclose filter steps and side-effect steps in `RangeByIsCountStrategy` when `NotStep` is used to wrap part of the traversal.

VOTE: +1


spmallette:
Looks like this goes to tp31 and master - tp31 should get the CHANGELOG entry then, right?

dkuppitz:
You're right. Updated both PRs.

spmallette:

VOTE +1

okram:
VOTE +1.

PR #411, opened 2016-09-13
TINKERPOP-1444: Benchmark bytecode->Traversal creation and implement GremlinServer cache if necessary.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1444

GremlinServer caching `Map<Bytecode,Traversal>` is only powerful (really powerful) when bindings are not involved (a solid 50x+ increase in speed). However, given that production systems will typically have bindings in their traversals, then such a cache doesn't help. 

In its place, a update to `JavaTranslator` is provided that makes use of an internal static `METHOD_CACHE` so as to make reflection faster. The speed up is provided below:

```
Prior to method caching:
  Bytecode->JavaTranslator call    : 0.13305238800000002
After method caching:
  Bytecode->JavaTranslator call    : 0.054839154
```

CHANGELOG

```
* Added a static method cache to `JavaTranslator` to reduce the cost of Java reflection.
```

@spmallette --- if you think that a `Map<Bytecode,Traversal>` cache IS necessary for GremlinServer,  then  please VOTE -1 and we can collaborate on this branch to also include a bytecode-cache.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/JavaTranslator.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/JavaTranslator.java
@@ -106,8 +106,8 @@ public final class JavaTranslator<S extends TraversalSource, T extends Traversal
         //////////////////////////
         // populate method cache for fast access to methods in subsequent calls
         final Map<String, List<Method>> methodCache = delegate instanceof TraversalSource ?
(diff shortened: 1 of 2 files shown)


spmallette:
I'm thinking that the cache hit count would be pretty low given the fact that bytecode contains parameters. since the hit count will be low, we'll just have the overhead of a cache without much value. it would be sweet if we could decouple the traversal from parameters so that a cache would be more effective, but i don't know what's involved with that offhand.

spmallette:
VOTE +1

dkuppitz:
All good, here's the missing

VOTE: +1

PR #412, opened 2016-09-13
TINKERPOP-1442 Improved session cleanup on client close [tp31]

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1442

While not a perfect implementation, a long run job blocking a close request from the client will now at least get an attempt at interruption rather than consuming the thread indefinitely (or until timeout whenever that is). 

All unit tests are good including gremlin-server integration tests. 

VOTE +1


PommeVerte:
VOTE: +1

davebshow:
This branch builds successfully using mvn clean install. VOTE +1

PR #413, opened 2016-09-14
TINKERPOP-1442 Improved session cleanup on client close [master]

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1442

see #412 for more description

All unit tests and gremlin-server integration tests are passing on this branch.

VOTE +1


PommeVerte:
VOTE +1 this is a nice fix to have. Unit tests passed for me too.

davebshow:
Stop accepting transactions after a close makes sense to me. This branch builds successfully using mvn clean install. VOTE +1

robertdale:
I'm just going to throw my $0.02 out there. Maybe I don't understand the use case, but this seems strange and unexpected to me.  I don't understand why or how a completely separate client would affect another.  Isn't one of the reasons for using sessions is that a client can come in and out while retaining state?  How would a client safely disconnect and reconnect later without killing the session and other clients?

PR #415, opened 2016-09-14
TINKERPOP-1448 gremlin-python should be Python 2/3 compatible

davebshow:
https://issues.apache.org/jira/browse/TINKERPOP-1448

This PR adds code to promote Python 2/3 compatibility in the gremlin-python GLV. The main change here involves adding a module, `compat`, that provides 2/3 compatible type info.

`compat` defines a class `long` for Python 3. `long` inherits from `int` and behaves like an integer. The advantage of creating this class is it allows application developers to type an integer as `long` in Python 3 code, which will result in serialization as `int64` in GraphSON.

`compat` also provides values for standard library `types` module that are not included in Python 3 (`IntType`, `FloatType`, and `LongType`).

Finally, this PR uses `six.get_function_code` for introspection, as this API changed with Python 3.

While all tests have been updated accordingly, when building with `mvn clean install -pl gremlin-python -DglvPython` tests are only run against the local machine's default version of Python. Therefore, tests are not run on both versions by default, however, they *will* now pass if a user's default Python interpreter is Python 3.


okram:
I like this. It works for me against Python 2.7.10. The one thing I think would be nice is to move the `compat.py` code into `statics.py`. Why? One less file. The `compat.py` code is all about static definitions of numeric types and thus is about `statics.py`.

VOTE +1.

davebshow:
Sounds good to me. I'll make the changes.

gremlin-python/src/main/jython/gremlin_python/driver/remote_connection.py
```diff
@@ -42,7 +42,7 @@ def traversal_source(self):
 
     @abc.abstractmethod
     def submit(self, bytecode):
-        print "sending " + bytecode + " to GremlinServer..."
+        print("sending " + bytecode + " to GremlinServer...")
```
spmallette: Is use of `print()` "good"? in Java we wouldn't `System.out.println()` we would use a logging framework of some sort (i.e. log4j). Is python similar in that respect? no big deal for this PR - just asking for future reference.

spmallette:
@davebshow what are the risks of building python 2 and not 3 or the other way around? will we easily break compatibility such that not building against both on a constant basis will cause problems? or can building both be saved for "integration tests" which run less frequently? 

put in more context, is it worth extending the build time of `mvn clean install` to cover both python 2 and 3? 

I'm sorta hoping the answer is "no" and that we can make running "both" part of integration tests only. 

Also, what does python 2/3 compatibility mean for deployment? does anything change in how this stuff goes up to pypi?

On the whole, this is most excellent btw. We really needed this. I do think that we should try to figure out how all the test/deploy stuff gets settled before merging though. I suggest we further wait for #416 to merge to master so that you can rebase this branch. Then I can work with you to tweak the pom.xml from there based on whatever is settled on for test/deploy issues.

gremlin-python/src/main/jython/gremlin_python/driver/remote_connection.py
davebshow: No, typically I would use Python's `logging` module. As the above is part of an abstract class and never gets called, I left is as such. In the future we can set up some nice logging for gremlin-python.

davebshow:
Maintaining compatibility should be pretty simple, but as you can see from the PR, there are little things that can go wrong. I think the main issue we will run into with this library is numeric typing, and I think we are well on our way to dealing with it.

Regarding the mvn build, I think we can just have it run against the default Python installation for now, which is a bit more risky, but we can mitigate the risk by running CI builds for multiple versions of Python. I also like to test locally, I just spin up a server manually and run `python setup.py test` against multiple versions of Python. This kind of multiple environment testing is made easier using the tool `tox`, maybe we could look at adding this as a requirement.

This shouldn't affect the deployment process at all. The only difference is now we can add Python 3 to the package metadata.

gremlin-python/src/main/jython/gremlin_python/driver/remote_connection.py
spmallette: yes - can you please create a JIRA issue for "nice logging"? thanks

spmallette:
ok - fair enough

> The only difference is now we can add Python 3 to the package metadata.

can you tack that into this PR please?

gremlin-python/src/main/jython/gremlin_python/driver/remote_connection.py
davebshow: will do. I'll be opening serveral JIRA issues tomorrow morning.

davebshow:
1c5e698 add simple package metadata.

davebshow:
VOTE +1

spmallette:
Ran tests on Python 2.7 with:

```test
mvn clean install -pl gremlin-python -DglvPython
```

VOTE +1

PR #416, opened 2016-09-14
TINKERPOP-1449 Streamline gremlin-python build

spmallette:
Provides implementation for both: 

https://issues.apache.org/jira/browse/TINKERPOP-1449
https://issues.apache.org/jira/browse/TINKERPOP-1431

I'm really happy with this PR. Cleaned up a lot of weird/inconsistent things in the gremlin-python build. Most importantly, I got rid of `-DglvPython` - which I somehow manage to forget to include on more occasions that i care to admit to. The solution is pretty simple - if you have an environment setup to build gremlin-python then you add a `.glv` file to the root of `gremlin-python` (hidden from git). If maven sees that file, it knows that it's OK to activate the profile that builds it.

You can still obviously do:

```text
mvn clean install -pl gremlin-server -Pglv-python
```

which directly activates the profile by its name.

Also note that I bumped to the latest version of the gmavenplus-plugin and centralized its configuration in the parent pom.xml.

All docs are updated to reflect the change to how `gremlin-python` is now built. Tests all seem to run ok. I even did a manual test to deploy python artifacts to pypitest and it worked nicely with:

```text
mvn clean deploy -pl gremlin-python -Dpypi -DskipTests
```

Docs are now generating with `-DskipTests` as well.

VOTE: +1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ benchmarks/
 __pycache__/
 *.py[cdo]
 __version__.py
+.glv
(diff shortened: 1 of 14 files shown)


davebshow:
Worked like a charm. Vote +1

PR #417, opened 2016-09-14
TINKERPOP-1330: by()-modulation for where()

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1330

 Added `by()`-modulation support to `where()` predicate-based steps. Added 3 solid `WhereTest` cases to verify proper functioning. Ensured proper `hashCode()` construction in `WhereStepTest`. Also, optimized `TraversalRing` to return `null` if there are no traversals in the ring and thus, `TraversalUtil.applyNullable()` can be leveraged instead which
 is more efficient than using `IdentityTraversal`. Finally, there was a severe bug in Gremlin-Python that made a complex `WhereTest` fail because `P.and` and `P.or` nesting was 
 reversed! I have fixed Gremlin-Python `P` in this PR.

---

 Here is an example of the new `where().by()`-model.

```
// give me "a" and "b" is "a" knows "b" and "a" is older than "b".
gremlin> g.V().as("a").out("knows").as("b").
......1>   where("a",gt("b")).by("age").
......2>   select("a","b").by("name")
==>[a:marko,b:vadas]


gremlin>  g.V().as("a").outE("created").as("b").
......1>   inV().as("c").
......2>   in("created").as("d").
......3>     where("a", lt("b").or(gt("c")).and(neq("d"))).
......4>       by("age").
......5>       by("weight").
......6>       by(__.in("created").values("age").min()).
......7>   select("a", "c", "d").by("name")
==>[a:josh,c:lop,d:marko]
==>[a:josh,c:lop,d:peter]
==>[a:peter,c:lop,d:marko]
==>[a:peter,c:lop,d:josh]
```

In the second query:
	a -> "age"
	b -> "weight"
	c -> in("created")...
	d -> "age" // TraversalRings are round-robin structures

Pretty insane-o.

If people start using `where()-by()` heavily, I believe there is an easy translation to `match()` and as such a `TraversalStrategy` would enable us to get the benefits of `match()`-steps runtime query optimizer.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,10 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.2.3 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Fixed a bug in Gremlin-Python `P` where predicates reversed the order of the predicates.
+* Fixed a naming bug in Gremlin-Python where `P._and` and `P._or` should be `P.and_` and `P.or_`. (*breaking*)
+* `where()` predicate-based steps now support `by()`-modulation.
+* `TraversalRing` returns an null if it does not contain traversals (previously `IdentityTraversal`).
 * Fixed a `JavaTranslator` bug where `Bytecode` instructions were being mutated during translation.
 * Added `Path` to Gremlin-Python with respective GraphSON 2.0 deserializer.
 * Added missing `InetAddress` to GraphSON extension module.
--- a/docs/src/recipes/traversal-induced-values.asciidoc
+++ b/docs/src/recipes/traversal-induced-values.asciidoc
@@ -40,37 +40,21 @@ g.V(marko).out('knows').has('age', gt(marko.value('age'))).values('name')
 
 The downside to this approach is that it takes two separate traversals to answer the question. Ideally, there should
 be a single traversal, that can query "marko" once, determine his `age` and then use that for the value supplied to
-filter the people he knows. In this way the _value_ for the `age` filter is _induced_ from the `Traversal` itself.
+filter the people he knows. In this way the _value_ for the `age` in the `has()`-filter is _induced_ from the `Traversal`
+itself.
 
 [gremlin-groovy,modern]
 ----
-g.V().has('name','marko').as('marko').         <1>
-  out('knows').as('friend').                   <2>
-  filter(select('marko','friend').by('age').   <3>
-         where('friend', gt('marko'))).        <4>
+g.V().has('name','marko').as('marko').      <1>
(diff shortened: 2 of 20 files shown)


dkuppitz:
Regarding the `match()` translation, it's indeed very easy:

```
gremlin> g.V().as("a").outE("created").as("b").
......1>  inV().as("c").in("created").as("d").
......2>    match(
......3>      __.as("a").values("age").as("a_by"),
......4>      __.as("b").values("weight").as("b_by"),
......5>      __.as("c").in("created").values("age").min().as("c_by"),
......6>      __.as("d").values("age").as("d_by"),
......7>    ).
......8>  where("a_by", lt("b_by").or(gt("c_by")).and(neq("d_by"))).
......9>  select("a", "c", "d").by("name")
==>[a:josh,c:lop,d:marko]
==>[a:josh,c:lop,d:peter]
==>[a:peter,c:lop,d:marko]
==>[a:peter,c:lop,d:josh]
```

I will make a code review and vote tomorrow.

spmallette:
I think we did a pretty bad job on upgrade docs in 3.2.2 so i'm kinda on the lookout for things that we might add to them - this might be a nice thing to call attention to in "user" section imo.

spmallette:
Could you please update the Traversal Induced Values recipe?

https://github.com/apache/tinkerpop/blob/master/docs/src/recipes/traversal-induced-values.asciidoc

okram:
@spmallette I updated `traversal-induced-values.asciidoc`. Wow -- `where().by()` is super powerful. That nested `where(select())`-clause you had in there was naaaasty. 

Given how much easier some things are to express, I'm realizing that people will be wanting to use `where().by()`  more often than not and given that `where().by()` is a `PathProcessor` now, OLAP will not allow any `by()` beyond the element id! Eek. No fear, we have `PathProcessorStrategy` that knows how to "flatten" `select().by()` accordingly. With that, I think another ticket should provide two things:

1. `PathProcessorStrategy` should be updated to support `where().by()` flattening.
2. Create a new strategy called `WhereByMatchStrategy` which turns `where().by()` patterns into `match()`-steps. (@dkuppitz's idea)

Now, if (1) above happens then (2) above won't do anything. Why would (1) happen, but not (2). The complexity of the `where()`-step's `by()`-clause. If the `by()` clause is within the bounds of the local star graph, then (1) will trigger. Else, it won't and (2) would trigger.  Next, we will have to see how these two strategies interact with `MatchPredicateStrategy` as this folds `where()`-clauses into `match()` to get the runtime optimizer benefits... Its all very complex :D

spmallette:
Doc update looks good. I'd still like a short section in the upgrade docs but other than that - All tests pass with `docker/build.sh -t -n -i`

VOTE +1

dkuppitz:
Code looks great, although the Python-related fixes are a bit misplaced in this PR. Anyhow,

VOTE: +1

okram:
The Python-related fixes are necessary or else some of the `WhereTests` fail. Python has a bad ordering to the scoping variables with nested `and()`/`or()`. Thus, for the test suite to pass I had to fix a bug in Python.

PR #418, opened 2016-09-15
Fixed issue in NumberSerializer that could cause integer out of bounds

admiralFedora:
In python, unless a number is specified with an L, it will be an int whether it's a 32 or 64 bit integer. When a 64 bit integer is serialized, it could potentially be labeled as an int32 which would cause an integer out of bounds error.

### DIFF
--- a/gremlin-python/src/main/jython/gremlin_python/structure/io/graphson.py
+++ b/gremlin-python/src/main/jython/gremlin_python/structure/io/graphson.py
@@ -181,7 +181,7 @@ class NumberSerializer(GraphSONSerializer):
     def _dictify(self, number):
         if isinstance(number, bool):  # python thinks that 0/1 integers are booleans
             return number
-        elif isinstance(number, long) or (number > 2147483647): # in python all numbers are int unless specified otherwise
+        elif isinstance(number, long) or (abs(number) > 2147483647): # in python all numbers are int unless specified otherwise
             return _SymbolHelper.objectify("Int64", number)
         elif isinstance(number, int):
             return _SymbolHelper.objectify("Int32", number)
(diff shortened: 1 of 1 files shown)


gremlin-python/src/main/jython/gremlin_python/structure/io/graphson.py
```diff
@@ -181,7 +181,7 @@ class NumberSerializer(GraphSONSerializer):
     def _dictify(self, number):
         if isinstance(number, bool):  # python thinks that 0/1 integers are booleans
             return number
-        elif isinstance(number, long):
+        elif isinstance(number, long) or (number > 2147483647): # in python all numbers are int unless specified otherwise
```
leifurhauks: is graphson's Int32 unsigned?

spmallette: i guess the expectation is a signed integer

leifurhauks: Maybe the absolute value of the integer should be checked here.

admiralFedora: Whoops. Thanks for reminding me.

PommeVerte:
What's the test status on this GLV? Could it be possible to add tests for this case?

spmallette:
We definitely can do native python tests now. That stuff works. As it stands we have no unit tests for the serializers. We could do with some for sure. I'm content to accept this small PR though for what it is without a battery of tests. The change is pretty small. 

Longer term though, I think we really need to work on the python test suite so that when little PRs come it's easy for folks to just drop in a test or two for their changes.

davebshow:
I agree, the tests need work. I'm hoping to get there this weekend. I think though, as Stephen said, this can be merged pretty safely.

PR #422, opened 2016-09-15
TINKERPOP-1409 Print no result line if empty.result.indicator is empty string.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1409

This seems like a better use of the empty string setting than to actually just print an empty string line.

```text
gremlin> graph = TinkerGraph.open()
gremlin> graph.close()
==>null
gremlin> graph = TinkerGraph.open()
gremlin> :set empty.result.indicator ""
gremlin> graph.close()
gremlin> 
```

Pretty small change, but, I didn't CTR this in case someone didn't really like this behavior.  If there are -1s I can just kill this and close the JIRA currently hanging out there as I only kept it open because I saw this dangling issue I wanted to implement.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -32,6 +32,7 @@ TinkerPop 3.2.3 (Release Date: NOT OFFICIALLY RELEASED YET)
 * `TraversalRing` returns a `null` if it does not contain traversals (previously `IdentityTraversal`).
 * Fixed a `JavaTranslator` bug where `Bytecode` instructions were being mutated during translation.
 * Added `Path` to Gremlin-Python with respective GraphSON 2.0 deserializer.
+* If `empty.result.indicator` is set to an empty string, then no "result line" is printed in the console.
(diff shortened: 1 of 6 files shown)


dkuppitz:
VOTE: +1

okram:
Cool. VOTE +1.

okram:
Acctually, I think the parameterization should be `null.result.indicator` as you check `object == null` and thus, its not "empty," but `null`.

spmallette:
I ended up naming it `result.indicator.null` as "result." was the prefix for other similar preferences:

```text
gremlin> graph = TinkerGraph.open()
gremlin> graph.close()
==>null
gremlin> :set result.indicator.null nil
gremlin> graph = TinkerGraph.open()
gremlin> graph.close()
==>nil
gremlin> :set result.indicator.null ""
gremlin> graph = TinkerGraph.open()
gremlin> graph.close()
gremlin> 
```

PR #425, opened 2016-09-16
Fixed behavior of has() methods

dkuppitz:
Made `has[Id|Label|Key|Value]` compatible with the old varargs method signature and added tests to ensure the compatibility.

VOTE: +1


spmallette:
Please adapt your "INFO" email to dev into the Upgrade Docs. I think it's important to explain the history of what happened here for 3.2.3 as we missed the change for the Upgrade Docs in 3.2.2.

PommeVerte:
Docs were clear, change is straightforward and tested this manually:
VOTE +1

spmallette:
good: "all `has()` method overloads behave like before"
 
VOTE +1

PR #426, opened 2016-09-16
TINKERPOP-1437 Added tests for dedup(Scope,String...) overload in DedupTest

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1437

Pretty simple - just added a couple of tests - but didn't CTR as I'd feel better with a couple of extra eyes on this as we've had bad tests slip into the suite in the past.  

Runs nicely with `mvn clean install` - doing a `docker/build.sh -t -i -n` now. If anything bad happens I'll post back here.

VOTE +1

### DIFF
--- a/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/DedupTest.java
+++ b/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/DedupTest.java
@@ -23,6 +23,7 @@ import org.apache.tinkerpop.gremlin.process.AbstractGremlinProcessTest;
 import org.apache.tinkerpop.gremlin.process.GremlinProcessRunner;
 import org.apache.tinkerpop.gremlin.process.traversal.Order;
 import org.apache.tinkerpop.gremlin.process.traversal.Path;
+import org.apache.tinkerpop.gremlin.process.traversal.Scope;
 import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.__;
 import org.apache.tinkerpop.gremlin.structure.T;
@@ -38,6 +39,9 @@ import java.util.Set;
 
 import static org.apache.tinkerpop.gremlin.LoadGraphWith.GraphData.MODERN;
 import static org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.__.bothE;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.isIn;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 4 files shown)


gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/DedupTest.java
```diff
@@ -241,6 +279,16 @@ public void g_V_both_both_dedup_byXoutE_countX_name() {
 
     public static class Traversals extends DedupTest {
         @Override
+        public Traversal<Vertex, String> get_g_V_out_in_valuesXnameX_fold_dedupXlocalX_unfold() {
+            return g.V().out().in().values("name").fold().dedup(Scope.local).unfold();
+        }
+
+        @Override
+        public Traversal<Vertex, Map<String, String>> get_g_V_out_in_asXxX_in_asXyX_selectXx_yX_byXnameX_fold_dedupXlocal_x_yX_unfold() {
```
okram: You named the method wrong. Should be `get_g_V_out_asXxX_in_asXyX.....`

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/DedupTest.java
```diff
@@ -73,6 +77,40 @@
...
+        assertEquals(3, names.size());
+        assertThat(names, containsInAnyOrder("marko", "josh", "peter"));
+    }
+
+    @Test
+    @LoadGraphWith(MODERN)
+    public void g_V_out_in_asXxX_in_asXyX_selectXx_yX_byXnameX_fold_dedupXlocal_x_yX_unfold() {
+        final Traversal<Vertex, Map<String, String>> traversal = get_g_V_out_in_asXxX_in_asXyX_selectXx_yX_byXnameX_fold_dedupXlocal_x_yX_unfold();
+        final List<Map<String,String>> names = traversal.toList();
+        assertEquals(6, names.size());
```
okram: You can use the `checkResult` and `MapMaker` to make this cleaner if you like. Check `WhereTest` for examples:

```
    @Test
    @LoadGraphWith(MODERN)
    public void g_V_asXaX_outEXcreatedX_asXbX_inV_asXcX_inXcreatedX_asXdX_whereXa_ltXbX_orXgtXcXX_andXneqXdXXX_byXageX_byXweightX_byXinXcreatedX_valuesXageX_minX_selectXa_c_dX() {
        final Traversal<Vertex, Map<String, String>> traversal = get_g_V_asXaX_outEXcreatedX_asXbX_inV_asXcX_inXcreatedX_asXdX_whereXa_ltXbX_orXgtXcXX_andXneqXdXXX_byXageX_byXweightX_byXinXcreatedX_valuesXageX_minX_selectXa_c_dX();
        printTraversalForm(traversal);
        checkResults(makeMapList(3,
                "a", "peter", "c", "lop", "d", "josh",
                "a", "peter", "c", "lop", "d", "marko",
                "a", "josh", "c", "lop", "d", "marko",
                "a", "josh", "c", "lop", "d", "peter"), traversal);
    }
```

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/DedupTest.java
```diff
@@ -73,6 +77,40 @@
...
+    public void g_V_out_in_valuesXnameX_fold_dedupXlocalX_unfold() {
+        final Traversal<Vertex, String> traversal = get_g_V_out_in_valuesXnameX_fold_dedupXlocalX_unfold();
+        final List<String> names = traversal.toList();
+        assertEquals(3, names.size());
+        assertThat(names, containsInAnyOrder("marko", "josh", "peter"));
+    }
+
+    @Test
+    @LoadGraphWith(MODERN)
+    public void g_V_out_in_asXxX_in_asXyX_selectXx_yX_byXnameX_fold_dedupXlocal_x_yX_unfold() {
```
okram: This has a bad method name --- see my other comment and then fix it here too.

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/DedupTest.java
```diff
@@ -73,6 +77,40 @@
...
+
+    public abstract Traversal<Vertex, Map<String, String>> get_g_V_out_in_asXxX_in_asXyX_selectXx_yX_byXnameX_fold_dedupXlocal_x_yX_unfold();
+
+    @Test
+    @LoadGraphWith(MODERN)
+    public void g_V_out_in_valuesXnameX_fold_dedupXlocalX_unfold() {
+        final Traversal<Vertex, String> traversal = get_g_V_out_in_valuesXnameX_fold_dedupXlocalX_unfold();
+        final List<String> names = traversal.toList();
+        assertEquals(3, names.size());
+        assertThat(names, containsInAnyOrder("marko", "josh", "peter"));
```
okram: Note we have `checkResults()`... e.g.

```
    @Test
    @LoadGraphWith(MODERN)
    public void g_V_whereXnotXoutXcreatedXXX_name() {
        final Traversal<Vertex, String> traversal = get_g_V_whereXnotXoutXcreatedXXX_name();
        printTraversalForm(traversal);
        checkResults(Arrays.asList("vadas", "lop", "ripple"), traversal);
    }
```

okram:
Made a few comments to ensure your test code is consistent with the others. Besides minor cosmetics -- VOTE +1.

dkuppitz:
VOTE: +1

Question: For future tests, should we always prefer Hamcrest over JUnit?

spmallette:
I typically prefer JUnit for all simple assertions except for `assertTrue` and `assertFalse`. Those two really stink compared to hamcrest and `assertThat(something(), is(true))` or `assertThat(something(), is(false))` because they print better output than the former that makes it easier to immediately see the problem with the assertion on failure. I convert those `assertTrue` and `assertFalse` as I see opportunity to do so. I also use hamcrest when the assertion logic is complex and there is a better simple one liner operation with hamcrest that makes the assertion more readable with a cleaner output message than junit.  So, no, I don't think we have to prefer Hamcrest to Junit across the board, but we should consider it when it produces a more readable output or more readable code.

PR #427, opened 2016-09-19
TINKERPOP-927 bin/publish-docs.sh should only upload diffs.

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-927

VOTE: +1

PR #428, opened 2016-09-19
TINKERPOP-927 bin/publish-docs.sh should only upload diffs.

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-927

VOTE: +1


spmallette:
I assume #427 doesn't need 120cef6 ?  I don't think the dual merge commits hurt anything from the git history perspective.  I assume that you merged in TINKERPOP-927, realized you needed something else in TINKERPOP-927, committed again in there and then had to merge again to TINKERPOP-927-master? If so you could have `git reset --hard origin/master` on the TINKERPOP-927-master branch and then just remerged TINKERPOP-927 into it which would have given you one merge commit.

This is an awesome and important PR by the way - can't wait to start using this.

spmallette:
dopest +1 ever

okram:
VOTE +1.

PR #429, opened 2016-09-20
TINKERPOP-1457 Fixed Lambda serialization in bytecode for java/groovy.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1457

Standard tests using lambdas are ignored which is why we didn't catch this in the process tests. I added a single test to `GremlinServerIntegrationTest` to just validate that lambdas could be sent, so we at least have a basic smoke test for that functionality.

Also opened up a few tests by removing some `OptOut` entries on `RemoteGraph`. 

All tests pass with: 

```text
$ mvn clean install && mvn verify -pl gremlin-server -DskipIntegrationTests=false
```

Going to kick off a full docker build tonight too.

VOTE +1


dkuppitz:

VOTE: +1

okram:
VOTE +1

PR #432, opened 2016-09-22
Fixed SerializerProvider for GraphSON 2.0.

newkek:
Serialization was broken for unknown objects when creating multiple instances of the GraphSONMapper with different GraphSON versions.
The issue is fixed and I added a test that would be failing if there wasn't this fix.


spmallette:
Merged to master as CTR - all good with: `mvn clean install && mvn verify -pl gremlin-server -DskipIntegrationTests=false`

PR #433, opened 2016-09-22
TINKERPOP-1249 Add keep-alive functionality to websockets Java Driver

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1249

If a connection goes idle, the driver will periodically send a ping to the server to keep the connection alive. 

All good with `mvn clean install && mvn verify -pl gremlin-server -DskipIntegrationTests=false`. Also ran some manual tests and did some memory profiling of the driver to make sure there were no leaks.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -33,6 +33,7 @@ TinkerPop 3.2.3 (Release Date: NOT OFFICIALLY RELEASED YET)
 * `TraversalRing` returns a `null` if it does not contain traversals (previously `IdentityTraversal`).
 * Fixed a `JavaTranslator` bug where `Bytecode` instructions were being mutated during translation.
 * Added `Path` to Gremlin-Python with respective GraphSON 2.0 deserializer.
+* Added "keep-alive" functionality to the Java driver, which will send a heartbeat to the server when normal request activity on a connection stops for a period of time.
 * Renamed the `empty.result.indicator` preference to `result.indicator.null` in Gremlin Console
 * If `result.indicator.null` is set to an empty string, then no "result line" is printed in Gremlin Console.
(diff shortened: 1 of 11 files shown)


robertdale:
Tested a clean console build with 3.2.2 server.  Pinging works as expected.  Closing each Client, a Cluster, and ':remote close' does stop pinging as expected.

Not sure what the expected behavior is here but this does not stop pinging after close. I had expected it to work like Client.  Instead the Cluster must be closed.
```
cluster = Cluster.open('conf/remote-objects.yaml')
rem = DriverRemoteConnection.using(cluster, 'g')
g = EmptyGraph.instance().traversal().withRemote(rem)
g.V()
rem.close()
```

k4rthikr:
How do I know that pings are happening in the background? In other words, how do I test this?

robertdale:
Use your favorite packet sniffer.  I used tcpdump.

spmallette:
@robertdale thanks for testing that. If you pass in your own `Cluster`, then the `DriverRemoteConnection`, expects you to manage it. In this way, a `Cluster` can be shared.

robertdale:
Yup, there it is right there in the javadoc:  When close() is called, the Cluster is left open for the caller to close.

dkuppitz:

VOTE: +1

PommeVerte:
VOTE +1 I get the pinging as well. Looks good.

cjoftheweb:
Why does this wait until a connection is written to before it schedules the keepAliveFuture? Is there some other mechanism to keep connections alive before they've been written to?

spmallette:
@cjoftheweb if there was a reason for that, i don't recall what it was. i suppose that if keep-alive is enabled it should just start doing its work after the websocket handshake is established.

PR #436, opened 2016-09-25
TINKERPOP-944 Deprecate Graph.Exceptions.elementNotFound()

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-944

This exception wasn't used  in the code base at all except for tests. Seems best to just deprecate (in case a graph provider is publically using it). Didn't change the nature of the tests that were using this method, so this should be a non-breaking change.

The change is pretty simple, but I decided not to CTR since "deprecation" was involved. All tests good with `mvn clean install`.

VOTE +1


dkuppitz:
VOTE: +1

okram:
VOTE +1

PR #437, opened 2016-09-25
TINKERPOP-790 TraversalSource and Traversal implement AutoCloseable.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-790

This allowed for `TraversalSource` instances created using `withRemote()` to close out open `RemoteConnection` instances. It also sets up for side-effects created by a "remote" `Traversal` to be cleared from cache on the server. That work is not done yet and is reserved for a different ticket.

Builds with `$ mvn clean install && mvn verify -pl gremlin-server -DskipIntegrationTests=false`

I won't VOTE on this yet as I suspect there might be some feedback from @okram perhaps.

@davebshow note the placeholder for sending the "close" message for side-effects.


davebshow:
Good stuff, I should get there today.

okram:
Good. Was scared at first seeing "new methods" on Traversal and TraversalSource, but I think this is smart. Gotz to release da resourcezzzzzzz...

VOTE +1.

spmallette:
Just rebased on master to clean up conflicts and force pushed.

twilmes:
VOTE: +1

PR #441, opened 2016-09-28
TINKERPOP-1467 Corrected a number of problems in close() operations for the driver [tp31]

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1467

This was more of a commit than I wanted for `tp31`, but `close()` was really messed up. Fixed a number of race conditions and other logic that would allow the driver to hang on close. Also made it so that the `Cluster` makes an attempt to clean up any `Client` instances that it spawns.

Tested with `mvn clean install` and endlessly with `mvn verify -pl gremlin-server -DskipIntegrationTests=false` (basically ran it for a whole day over and over again).

VOTE +1

### DIFF
--- a/docs/src/upgrade/release-3.1.x-incubating.asciidoc
+++ b/docs/src/upgrade/release-3.1.x-incubating.asciidoc
@@ -27,6 +27,23 @@ TinkerPop 3.1.5
 
 *Release Date: NOT OFFICIALLY RELEASED YET*
 
+Please see the link:https://github.com/apache/tinkerpop/blob/3.1.4/CHANGELOG.asciidoc#tinkerpop-315-release-date-XXXXXXXXXXXX[changelog] for a complete list of all the modifications that are part of this release.
+
+Upgrading for Users
+~~~~~~~~~~~~~~~~~~~
+
+Java Driver and close()
+^^^^^^^^^^^^^^^^^^^^^^^
+
+There were a few problems noted around the `close()` of `Cluster` and `Client` instances, including issues that
+presented as system hangs. These issues have been resolved, however, it is worth noting that an unchecked exception
+that was thrown under a certain situation has changed as part of the bug fixes. When submitting an in-session request
+on a `Client` that was closed (or closing) an `IllegalStateException` is thrown. This replaces older functionality
+that threw a `ConnectionException` and relied logic far deeper in the driver to produce that error and had the
+potential to open additional resources despite the intention of the user to "close".
+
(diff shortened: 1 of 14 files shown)


gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
```diff
@@ -1248,6 +1250,73 @@ public void shouldProcessSessionRequestsInOrderAfterTimeout() throws Exception {
...
+        assertEquals(2, session.submit("1+1").all().get().get(0).getInt());
+        assertEquals(2, sessionlessTwo.submit("1+1").all().get().get(0).getInt());
+        assertEquals(2, sessionlessThree.submit("1+1").all().get().get(0).getInt());
+        // dont' send anything on the 4th client
+
+        // close one of these Clients before the Cluster
+        sessionlessThree.close();
+        cluster.close();
+
+        try {
```
dkuppitz: It is not verified, that exceptions are actually thrown. Should be more like:

```
try {
    sessionXyz.submit("1+1").all().get();
    assertTrue(false);
} catch (Exception ex) {
    ....
```

docs/src/upgrade/release-3.1.x-incubating.asciidoc
```diff
@@ -27,6 +27,23 @@ TinkerPop 3.1.5
...
+Please see the link:https://github.com/apache/tinkerpop/blob/3.1.4/CHANGELOG.asciidoc#tinkerpop-315-release-date-XXXXXXXXXXXX[changelog] for a complete list of all the modifications that are part of this release.
+
+Upgrading for Users
+~~~~~~~~~~~~~~~~~~~
+
+Java Driver and close()
+^^^^^^^^^^^^^^^^^^^^^^^
+
+There were a few problems noted around the `close()` of `Cluster` and `Client` instances, including issues that
+presented as system hangs. These issues have been resolved, however, it is worth nothing that an unchecked exception
```
dkuppitz: worth nothing => worth noting

dkuppitz:
As said in a private chat, integration tests were stuck when I ran them first. However, the second attempt succeeded. More investigation is apparently needed, but for this PR:

VOTE: +1

spmallette:
I ran the docker build multiple times today following the report of stuck integration tests. I never ran into the problem, but the adjustments I forced pushed prior to those tests hopefully fixed it.

dkuppitz:
FYI:  3x BUILD SUCCESS in a row.

My vote still holds true.

okram:
VOTE +1.

PR #447, opened 2016-10-04
TINKERPOP-1455: Provide String-based withStrategy()/withoutStrategy() for language variant usage

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1455

Added TraversalSource.withStrategy(String, Object...) and TraversalSource.withoutStrategy(String). This allows Gremlin language variants (e.g. Gremlin-Python) to add and remove strategies from a traversal source. Prior to this moment, strategy addition/removal was via strategy instances (Java objects) and classes (Class objects). That is not language variants friendly.

This branch is backwards compatible. If a `TraversalStrategy` does not have a builder and exposes a static `instance()`-method, then `g.withStrategy('com.blah.MyStrategy')` will work as expected. If a `TraversalStrategy` does have a builder (that is, can be configured), then it should expose a static `create(Configuration)`-method. Note Gremlin language variants could not use `withStrategies()` prior to this and thus, this addition is backwards compatible. Finally, `TraversalSource.withComputer(Object...)` was added as well for the same reasons.

```
>>> g = g.withStrategy('org.apache.tinkerpop.gremlin.process.traversal.strategy.decoration.SubgraphStrategy','vertices',hasLabel('person'),'edges',has('weight',gt(0.5)))
>>> g.V().name.toList()
[u'marko', u'vadas', u'josh', u'peter']
```

The CHANGELOG, gremlin-variants.asciidoc, and upgrade doc have all been updated accordingly.

VOTE +1

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/decoration/PartitionStrategy.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/decoration/PartitionStrategy.java
@@ -18,6 +18,7 @@
  */
 package org.apache.tinkerpop.gremlin.process.traversal.strategy.decoration;
 
+import org.apache.commons.configuration.Configuration;
 import org.apache.tinkerpop.gremlin.process.traversal.P;
 import org.apache.tinkerpop.gremlin.process.traversal.Parameterizing;
 import org.apache.tinkerpop.gremlin.process.traversal.Step;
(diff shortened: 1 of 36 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/decoration/PartitionStrategy.java
```diff
@@ -367,6 +387,26 @@ public Builder partitionKey(final String partitionKey) {
...
+         */
+        public Builder readPartitions(final String... readPartitions) {
+            return this.readPartitions(Arrays.asList(readPartitions));
+        }
+
+        /**
+         * Specifies the partition of the graph to read from.  It is possible to assign multiple partition keys so
+         * as to read from multiple partitions at the same time.
+         *
+         * @deprecated Since 3.2.3. Use {@link Builder#readPartitions} instead
```
spmallette: That link is invalid and as a minor nit, the text we tend to use is "As of release 3.2.3, replaced by {@link....".

okram:
```
[INFO] BUILD SUCCESS
[INFO] Total time: 05:09 h
[INFO] Finished at: 2016-10-04T22:33:42-06:00
[INFO] Final Memory: 157M/1520M
```

dkuppitz:
Code looks good, `docker/build.sh -t -i` succeeded.

VOTE: +1

spmallette:
This is a great PR. Didn't think we'd have this for 3.2.3. Nice

All tests pass with `docker/build.sh -t -i` - final "todo" would be to add a JIRA to track the deprecation for future removal. 

VOTE +1

okram:
If you guys don't mind, I'm going to change the method signatures (the NEW methods added -- nothing from before changed).

{code}
withStrategy(String, Object...) -> withStrategies(Map<String,Object>...)
withoutStrategy(String) -> withoutStrategies(String...)
{code}

The reason this is smart is:

1. Registering strategies is expensive as it costs a `TraversalSource` clone and a `TraversalStrategies` sort. This is why doing all strategy declarations in one method is smart.

2. While creating `Map`s in Java is horrendous, Java users would NOT use these methods. They would use the Java object version of these methods. However, for Gremlin-Python, Groovy, JavaScript -- creating maps is natural and looks good. Also, we don't run into the "multi-key" issues.

Nothing fudamental here, just some asethetics. If you don't object, I will just do this, update docs, and merge.

spmallette:
no problem from me.

dkuppitz:
Cool.

okram:
```
[INFO] BUILD SUCCESS
[INFO] Total time: 05:10 h
[INFO] Finished at: 2016-10-06T02:35:44-06:00
[INFO] Final Memory: 158M/1358M
```

okram:
So this is now done and merged. This branch provided a wealth of testing infrastructure, test cases, and ways of serializing `TraversalStrategy` instances. Now, with all that, I think we should actually NOT have `withStrategies(Map...)` and we should really make `g:TraversalStrategy`. I started going down that road to see how easy it would be and its actually pretty hard given how `GraphSONMapper` works -- and the need to register all strategies by class. Anywho... if we do decide to go down that route then in Gremlin-Python/etc. we would have `TraversalStrategy({config:map,is:this})`.

PR #452, opened 2016-10-10
TINKERPOP-1189 Increased performance of CredentialGraph authentication

robertdale:
Using lowest complexity 4 (was 10).


spmallette:
Merged this via CTR - pretty straightforward - thanks.

PR #453, opened 2016-10-10
TINKERPOP-1492: RemoteStrategy or the RemoteConnection should append a lazy barrier().

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1492

I think that `LazyBarrierStrategy` should add a `barrier()` to the end of a traversal if it believes it will reduce the number of traversers. This is particularly important for `withRemote()` as you can greatly reduce the network traffic. I don't have a good way of knowing if the traversal is being executed remote or not so this will lazy `barrier()` will be added regardless of remote'ing. 

Please review and tell me if you disagree with where I am placing this optimization.

VOTE +1.


spmallette:

VOTE +1

dkuppitz:
Code changes + tests look good.

VOTE: +1

PR #444, opened 2016-09-28
TINKERPOP-1158 gremlin.sh -v emits log4j initialization errors

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1158

Tweaked `gremlin.sh` so that it can be started from anywhere.

Removed all shell scripts that just forwarded the command to another script and added symbolic links instead.

VOTE: +1

PR #456, opened 2016-10-11
TINKERPOP-1493 Groovy project doesn't build on Windows

pauljackson:
Removed support for user.dir property as it was being prepended to a
fully qualified path and the second drive letter was making the path
illegal.

Made sure JarFile instances were being closed so that Groovy could
delete the directory without encountering file locked errors.

### DIFF
--- a/gremlin-groovy/src/main/groovy/org/apache/tinkerpop/gremlin/groovy/util/DependencyGrabber.groovy
+++ b/gremlin-groovy/src/main/groovy/org/apache/tinkerpop/gremlin/groovy/util/DependencyGrabber.groovy
@@ -173,21 +173,24 @@ class DependencyGrabber {
         try {
             def pathToInstalled = extPath.resolve(artifact.artifact + "-" + artifact.version + ".jar")
             final JarFile jar = new JarFile(pathToInstalled.toFile())
-            final Manifest manifest = jar.getManifest()
-            def attrLine = manifest.mainAttributes.getValue("Gremlin-Plugin-Dependencies")
-            def additionalDependencies = [] as Set<URI>
-            if (attrLine != null) {
-                def splitLine = attrLine.split(";")
-                splitLine.each {
-                    def artifactBits = it.split(":")
-                    def additional = new Artifact(artifactBits[0], artifactBits[1], artifactBits[2])
+            try {
+                final Manifest manifest = jar.getManifest()
+                def attrLine = manifest.mainAttributes.getValue("Gremlin-Plugin-Dependencies")
+                def additionalDependencies = [] as Set<URI>
+                if (attrLine != null) {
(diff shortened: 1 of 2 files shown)


spmallette:
Not sure if it matters, but if we only bring this change to master, 3.1.x won't build on windows and the process will have diverged. not sure if we should let that happen. @pluradj do you have an opinion?

pluradj:
Let's get 3.1.x fixed also. I don't see why not.

pluradj:
One problem I'm seeing is that RAT doesn't like these files on Windows:

```
Unapproved licenses:

  bin/gremlin.sh
  gremlin-console/bin/gremlin.sh
```

No complaints from Linux/Mac since they resolve OK with the [symlinks](https://github.com/apache/tinkerpop/pull/444). On Windows, they get extracted as files with content:

```
../gremlin-console/bin/gremlin.sh

../target/apache-tinkerpop-gremlin-console-3.2.3-SNAPSHOT-standalone/bin/gremlin.sh
```

spmallette:
ugh - i dunno guys - we kinda decided a couple weeks ago on dev that we weren't going to go too deep into supporting a windows build. i mean i'd like the basics to work like `mvn clean install`, but not at the risk of disrupting core development work. 

not sure what you can do about the symlinks situation - i guess you could add those to rat config to exclude them, because they won't have the license in them as symlinks or files.

pauljackson:
There’s more differences than just the comments at the head of these files, so you could copy them from 3.2.3 if you are ok with those changes.

pluradj:
RAT exclusion works. A Windows user shouldn't be touching those symbolic links anyway.

```
diff --git a/pom.xml b/pom.xml
index 2e67eda..5351689 100644
--- a/pom.xml
+++ b/pom.xml
@@ -293,6 +293,8 @@ limitations under the License.
                         <exclude>**/_bsp/**</exclude>
                         <exclude>DEPENDENCIES</exclude>
                         <exclude>**/.glv</exclude>
+                        <exclude>bin/gremlin.sh</exclude>
+                        <exclude>gremlin-console/bin/gremlin.sh</exclude>
                     </excludes>
```

@pauljackson Can you get the exclusions added in `pom.xml` to the PR and retarget it to `apache:tp31`?

spmallette:
@pluradj are you going to have votes for these PRs sometime soon? i was sorta waiting for you to be the first to test this out.

pluradj:
I think this one can be closed @pauljackson, and we'll use the other one to merge to all the branches.

pluradj:
I meant that this pull request #456 can be closed @pauljackson. We'll work with the other one #457.

pauljackson:
OK, sorry about that. I closed it now.

-Paul

PR #458, opened 2016-10-18
Message scope initialization in PeerPressureVertexProgram

sjudeng:
These updates resolve issues observed running tinkerpop-3.2.2 tests against Titan. Sorry if this should have been submitted through JIRA.

### Message scope initialization in PeerPressureTest

Failing test was `g_V_peerPressure_byXclusterX_byXoutEXknowsXX_pageRankX1X_byXrankX_byXoutEXknowsXX_timesX2X_group_byXclusterX_byXrank_sumX`

During test execution it was observed at the failing (vote) step that the message scope was `VertexStep(OUT,[knows],edge)`, but this didn't match the message scope, `VertexStep(OUT,edge)`, in the statically initialized set, `VOTE_SCOPE`, returned by `PeerPressureVertexProgram#getMessageScopes`. This caused an error in the custom memory implementation that is resolved by moving the initialization as shown in this PR.

### Vertex id checks in BulkDumperVertexProgramTest

Vertex id checks fail when implementation does not support user-specified ids

### Transaction rollback in TransactionTest

Implementation fails to rollback a transaction that is already closed

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/clustering/peerpressure/PeerPressureVertexProgram.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/clustering/peerpressure/PeerPressureVertexProgram.java
@@ -58,8 +58,6 @@ public class PeerPressureVertexProgram extends StaticVertexProgram<Pair<Serializ
 
     private MessageScope.Local<?> voteScope = MessageScope.Local.of(__::outE);
     private MessageScope.Local<?> countScope = MessageScope.Local.of(new MessageScope.Local.ReverseTraversalSupplier(this.voteScope));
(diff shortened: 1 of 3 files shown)


okram:
Ahhhhhhhh! Interesting. This has been a thorn in our side for some time.

VOTE +1.

spmallette:
@sjudeng do you mind re-targeting this pull request to the tp32 branch? In that way it will get fixed along the 3.2.x line of code. we can merge forward to master on our end so that is available on 3.3.x as well.

sjudeng:
Should be good now. I first tried to do this by reapplying the commit on tp32 before I realized I needed to edit the PR itself.

spmallette:

VOTE +1

dkuppitz:
VOTE: +1

PR #459, opened 2016-10-20
TINKERPOP-1485 Move source for TinkerPop site to source code repo

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1485

This PR adds a publish script for the TinkerPop homepage. The script uses the same pattern as `publish-docs.sh`:

```
$ site/bin/publish-home.sh <username>
```

It first checks out the currently deployed version, then calculates the difference with the local version and finally pushes only changed files to SVN.

VOTE: +1


spmallette:
I tried to use this to update the site for 3.2.3/3.1.5 and it didn't seem to work.  It didn't error, but nothing seemed to go to svn. Also, I see you create a `site/target` directory. I was using the root `/target` for stuff in my scripts. Is there any reason not to make that consistent? i'm fine with your directory local to `site/` btw - that's fine as long as it's all in one place.

spmallette:
forget that last post - that was dumb...................i didn't merge your latest changes (i thought i'd done it last night)

spmallette:
I have used this several times now in relation to 3.1.5/3.2.3 release and it works nicely:

VOTE +1

hopefully it works as nicely for @okram :)

twilmes:
This is awesome, thanks @dkuppitz  VOTE: +1

okram:
I added @robertdale and the "Gremlin implementation of the Gremlin Traversal Machine" blog post and used both `generate-home.sh` and `publish-home.sh` and everything working as expected.

  http://tinkerpop.apache.org/

One nick nack -- if others agree -- I think the homepage should be in the docs directory. `docs/site`. and then `site/bin` scripts should just be in `bin/`. 

Regardless -- VOTE +1.

spmallette:
i had thought about putting it under `/docs` but it didn't seem to fit there as the "site" really included "docs" and not the other way around. i didn't want to up-end everything and move `/docs` to `/site` so i just put it at the root, thinking we might move `/docs` under `/site` at some point in the future. But perhaps that's too gnarly a change for too little logic. I've moved `/site` under `/docs`.

PR #463, opened 2016-10-25
TINKERPOP-919 Added supportsDuplicateMultiProperties()

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-919

That feature allows a graph to specify whether or not it supports multi-properties that allow the same value to be supplied on the same key. Not sure why this wasn't implemented long ago. It was relatively straightforward to put in place. I did take a look at the test suite for existing tests that might have needed this feature included, but I didn't see any.

`mvn clean install -DincludeNeo4j` is all good

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.2.4 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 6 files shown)


okram:
I was a confused for a bit on the method name. Seems (and this is just me) that a better name would be:

```
Features.supportsDuplicateMultiPropertyValues()
```

To me this means same key, duplicate values. The problem with "identical" is that there is no such thing as identical as properties can have meta-properties and be different. Thus, its a key/value identity, not a key/value/properties/id identity. 

Dunno...whateves.

spmallette:
i played with a lot of different words. "duplicate" was one. i don't mind changing it if you like that better.

spmallette:
ok - renamed to use "duplicate" and rebased on tp32

okram:
VOTE +1

dkuppitz:
VOTE: +1

PR #464, opened 2016-10-25
TINKERPOP-1527: Do not override registered strategies in TraversalStrategies.GlobalCache

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1527

`GraphComputer` strategy registration in `TraversalStrategy.GlobalCache` has been a "uh..I think static{}-blocks load once and only once and when the class is used?" ... This assumption lead to a few errant problems that have been rectified and verified in this PR.

1.) `TraversalStrategies.GlobalCache` strategies can be manipulated without fear of a new instance of the `Graph` or `GraphComputer` class over-writing user manipulations. That is, a test case now proves that `static{}` loads once and only once.

2.) `TraversalStrategies.GlobalCache.getStrategies()` is responsible for ensuring that `static{}` has been called for the particular `Graph` or `GraphComputer` class. A test case now proves that `Class.forName()` is sufficient to accomplish this.


okram:
VOTE +1

spmallette:
Not sure why travis is unhappy - looks like some maven nonsense - anyway all tests pass with `docker/build.sh -t -i`

VOTE +1

dkuppitz:
VOTE: +1

PR #472, opened 2016-10-31
TINKERPOP-1508: Add choose(predicate,trueTraversal)

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1508

Added `GraphTraversal.choose(predicate,traversal)` and `GraphTraversal.choose(traversal,traversal)`. These are implemented as `choose(traversal,identity())`. Up to this point, `choose()` has been if/then/else-semantics and now with addition, we have if/then-semantics.

VOTE +1


dkuppitz:
VOTE: +1

PommeVerte:
VOTE +1

PR #473, opened 2016-11-01
TINKERPOP-1372: ImmutablePath should not use Java recursion (call stacks are wack)

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1372

`ImmutablePath` used heavy method-recursion which is expensive in Java to create a new call stack for each recurse. All method-recursion has been replaced with `while(true)`-recursion. Furthermore, was able to get rid of `ImmutablePath.TailPath` with a `public static ImmutablePath TAIL_PATH = new ImmutablePath(null,null,null)`.  This makes things much cleaner and we don't need the package protected `ImmutablePathImpl` interface. Finally, I stole @pietermartin's trick to use direct reference to `Set<String>` labels as the labels are immutable.

Here is a benchmark of a bunch of `match()`-traversals on the Grateful Dead graph where the first two columns are time in milliseconds and the last column is the number of returned results.

```
PREVIOUS   NEW     # RESULTS
----------------------------
[12.676,  12.019,  93]      
[222.123, 177.596, 2952]
[27.187,  35.787,  6]
[80.917,  77.891,  5421]
[189.354, 176.308, 5096]
[14.644,  14.969,  18]
[2.214,   0.908,   3]
[924.093, 777.707, 314932]
```

VOTE +1.


okram:
@pietermartin voted +1 on the dev@ mailing list for this ticket. In the future, @pietermartin, can you VOTE on the GitHub pull request comment thread?

okram:
@pietermartin -- I think my last commit is another optimization you discussed.

spmallette:

VOTE +1

PR #474, opened 2016-11-01
TINKERPOP-1471: IncidentToAdjacentStrategy use hidden marker to avoid repeated recursion.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1471

Hidden labels can be used by strategies as a way to communicate between "layers" of the compilation tree. This model is now applied to `IncidentToAdjacentStrategy` where if the traversal (as a whole) should NOT be processed by `IncidentToAdjacentStrategy`, then each traversal in the tree is "marked." If the traversal can be processed by `IncidentToAdjacentStrategy`, then there is no need to mark the traversals. This removes the need to, for every traversal, do a recursion from the parent traversal looking for invalidating step classes.

VOTE +1.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/decoration/SubgraphStrategy.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/decoration/SubgraphStrategy.java
@@ -105,21 +105,11 @@ public final class SubgraphStrategy extends AbstractTraversalStrategy<TraversalS
         this.vertexPropertyCriterion = null == vertexPropertyCriterion ? null : vertexPropertyCriterion.asAdmin().clone();
 
         if (null != this.vertexCriterion)
-            this.addLabelMarker(this.vertexCriterion);
+            TraversalHelper.applyTraversalRecursively(t -> t.getStartStep().addLabel(MARKER), this.vertexCriterion);
         if (null != this.edgeCriterion)
-            this.addLabelMarker(this.edgeCriterion);
+            TraversalHelper.applyTraversalRecursively(t -> t.getStartStep().addLabel(MARKER), this.edgeCriterion);
         if (null != this.vertexPropertyCriterion)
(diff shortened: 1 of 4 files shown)


dkuppitz:
VOTE: +1

A little refactoring would be nice though:

```
for (final Step<?, ?> step : traversal.getSteps()) {
    if (step instanceof TraversalParent) {
        ((TraversalParent) step).getLocalChildren().forEach(this::addLabelMarker);
        ((TraversalParent) step).getGlobalChildren().forEach(this::addLabelMarker);
    }
}
```
... should be as easy as:

```
TraversalHelper.forEachNestedTraversal(traversal, this::addLabelMarker);
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/decoration/SubgraphStrategy.java
```diff
@@ -105,21 +105,11 @@ private SubgraphStrategy(final Traversal<Vertex, ?> vertexCriterion, final Trave
         this.vertexPropertyCriterion = null == vertexPropertyCriterion ? null : vertexPropertyCriterion.asAdmin().clone();
 
         if (null != this.vertexCriterion)
-            this.addLabelMarker(this.vertexCriterion);
+            TraversalHelper.applyTraversalRecursively(t -> t.getStartStep().addLabel(MARKER), this.vertexCriterion);
```
dkuppitz: `vertexCriterion` can be `null`. In that case `TraversalHelper.applyTraversalRecursively` and/or the consumer would throw a NPE.

okram: ? -- look at line 107 -- `if(null != this.vertexCriteria)`

dkuppitz: Ha, it's getting late here.....

spmallette:

VOTE +1

PR #475, opened 2016-11-01
TINKERPOP-1211: UnfoldStep should unfold arrays.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1211

Unfold step now handles unfolding of arrays.

VOTE +1.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.3.0 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* `UnfoldStep` now supports unfolding of arrays. (*breaking*)
 * Removed all performance tests that were not part of `gremlin-benchmark`.
 * Removed dependency on `junit-benchmarks` and it's related reference to `h2`.
 * Moved the source for the "home page" into the repository under `/site` so that it easier to accept contributions.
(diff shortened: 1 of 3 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/UnfoldStep.java
```diff
@@ -46,6 +47,8 @@ else if (s instanceof Iterable)
             return ((Iterable) s).iterator();
         else if (s instanceof Map)
             return ((Map) s).entrySet().iterator();
+        else if (s.getClass().isArray())
+            return new ArrayIterator((Object[])s);
```
twilmes: This cast will fail if the incoming array is an array of primitives.  I think you could do something like this which is short and to the point but creates an intermediate, albeit short-lived list.
`return (Iterator<E>)Arrays.asList(s).iterator();`

I was able to reproduce the issue when I nested a primitive array inside of a list.  Calling `unfold` twice looks a little silly but maybe someone would do this if they were dealing with nested collections.

```
gremlin> __.inject([0, 1, 2, [3, 4] as int[]]).unfold().unfold()
==>0
==>1
==>2
[I cannot be cast to [Ljava.lang.Object;
Display stack trace? [yN]n
gremlin> __.inject([0, 1, 2, [3, 4]]).unfold().unfold()
==>0
==>1
==>2
==>3
==>4
```

spmallette:

VOTE +1

CHANGELOG.asciidoc
```diff
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.3.0 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* `UnfoldStep` now supports unfolding of arrays. (*breaking*)
```
spmallette: I don't really like the use of "breaking" in the CHANGELOG for our manual entries. The JIRA ticket will have that when we generate that list on release.  Also, we've not done that consistently from the start of CHANGELOG. If something is "breaking" the better way to call attention to it is in the upgrade docs where we can better describe and demonstrate the change for users rather than a single one-liner in CHANGELOG.

PR #476, opened 2016-11-01
TINKERPOP-1292: TinkerGraphComputer VertexProgramInterceptors

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1292

Added `TinkerGraphCountStrategy` which will translate `g.V().count()` and `g.E().count()` into direct calls to the underlying `TinkerGraph.vertices/edges` hash maps (`.size()`). Thus, no need to iterate out the vertices/edges and then count the iteration. O(1) time for count.

--- we can extend on this line of reasoning in this branch if people have other ideas for shortcuts.

* Also did some cleanup of some test suites in Neo4jGraph and TinkerGraph that should have been removed long ago when their respective strategy suites were removed.

VOTE +1.

### DIFF
--- a/neo4j-gremlin/src/main/java/org/apache/tinkerpop/gremlin/neo4j/structure/Neo4jGraph.java
+++ b/neo4j-gremlin/src/main/java/org/apache/tinkerpop/gremlin/neo4j/structure/Neo4jGraph.java
@@ -71,7 +71,6 @@ import java.util.stream.Stream;
 @Graph.OptIn(Graph.OptIn.SUITE_GROOVY_ENVIRONMENT_INTEGRATE)
 @Graph.OptIn(Graph.OptIn.SUITE_GROOVY_ENVIRONMENT_PERFORMANCE)
 @Graph.OptIn("org.apache.tinkerpop.gremlin.neo4j.NativeNeo4jSuite")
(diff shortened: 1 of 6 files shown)


twilmes:
VOTE: +1

neo4j-gremlin/src/main/java/org/apache/tinkerpop/gremlin/neo4j/structure/Neo4jGraph.java
```diff
@@ -71,7 +71,6 @@
 @Graph.OptIn(Graph.OptIn.SUITE_GROOVY_ENVIRONMENT_INTEGRATE)
 @Graph.OptIn(Graph.OptIn.SUITE_GROOVY_ENVIRONMENT_PERFORMANCE)
 @Graph.OptIn("org.apache.tinkerpop.gremlin.neo4j.NativeNeo4jSuite")
-@Graph.OptIn("org.apache.tinkerpop.gremlin.neo4j.process.traversal.strategy.Neo4jStrategySuite")
```
okram: Yes. It is not the model we use for `TraversalStrategy` testing. I converted both Neo4j and TinkerGraph's strategy tests to the model we use in `core/`.

spmallette:

VOTE +1

PR #477, opened 2016-11-02
TINKERPOP-1473: Given PathRetractionStrategy, PathProcessorStrategy can be extended to support partial where() inlining.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1473

Extended `PathProcessStrategy` to support the following translations which expose more traversals to OLAP processing (works around local star graph violations).

```
where(as(a).out().as(b))    => as(xyz).select(a).where(out().as(b)).select(xyz)
where(as(a).out())          => as(xyz).select(a).filter(out()).select(xyz)
```

Next, given that `PathProcessStrategy` is prior to `InlineFilterStrategy` you get this effect:

```
where(as(a).has(z))          => as(xyz).select(a).has(z).select(xyz)
```

You can see why `PathRetractionStrategy` is important. Without it `xyz` (which are UUIDs in practice) would remain in the traverser's path data.

VOTE +1.


spmallette:

VOTE +1

twilmes:
Tests passed and manual testing looked good.

VOTE: +1

PR #479, opened 2016-11-02
TINKERPOP-1434 Block calls on "remote" traversal to get side-effects

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1434

This prevents calls for side-effects that are not yet ready on the server. This was generally a problem in tests, but users could have also run afoul of the issue if they decided to get low-level with the API. It is more likely that they would grab side-effects with a call to `cap()` so this would be a non-issue in that case.

Did multiple runs of `mvn clean install && mvn verify -pl gremlin-server -DincludeNeo4j -DskipIntegrationTests=false` and it passes nicely. I never saw the test failures that triggered the effort to produce a fix for this problem, so @dkuppitz who saw it regularly should give this a review. Also, if travis passes, that would be a good sign - the error was seen there sometimes as well.  Please be sure to use `-DincludeNeo4j` if re-running the tests.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -37,6 +37,7 @@ TinkerPop 3.2.4 (Release Date: NOT OFFICIALLY RELEASED YET)
 * Added `ProjectStep.getProjectKeys()` for strategies that rely on such information.
 * Added `VertexFeatures.supportsDuplicateMultiProperties()` for graphs that only support unique values in multi-properties.
 * Deprecated the "performance" tests in `OptIn`.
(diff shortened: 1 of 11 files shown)


dkuppitz:
Unfortunately I still get the same errors. Looks like they always happen in `RemoteGraphGroovyTranslatorProcessStandardTest`.

dkuppitz:
Finally! 3 good builds in row, using `docker/build.sh -t -i -n`

VOTE: +10

okram:
VOTE +1.

PR #481, opened 2016-11-03
TINKERPOP-1428: profile() throws NPE for union(group, group)

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1428#

`GroupSideEffectStep` is "weird" in that it splits the `valueTraversal` into a pre- and post-traversals where the split is the first `Barrier`-step encountered. This makes it so that its possible for a local child to never be `next()'d` and thus, profiling never initialized for that step. The blanket fix is basically, if a metric is not initialized, don't include it in the profile results. As a side, I fixed a bug in `GroupXXXStep` around empty traverser set values (basically, don't include them).

VOTE +1.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/GroupSideEffectStep.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/GroupSideEffectStep.java
@@ -43,9 +43,9 @@ import java.util.Set;
 public final class GroupSideEffectStep<S, K, V> extends SideEffectStep<S> implements SideEffectCapable<Map<K, ?>, Map<K, V>>, TraversalParent, ByModulating {
 
     private char state = 'k';
-    private Traversal.Admin<S, K> keyTraversal = null;
-    private Traversal.Admin<S, ?> preTraversal = null;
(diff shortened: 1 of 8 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/GroupSideEffectStep.java
```diff
@@ -45,7 +45,7 @@
     private char state = 'k';
     private Traversal.Admin<S, K> keyTraversal = null;
     private Traversal.Admin<S, ?> preTraversal = null;
-    private Traversal.Admin<S, V> valueTraversal = this.integrateChild(__.fold().asAdmin());
+    private Traversal.Admin<S, V> valueTraversal = null;
```
dkuppitz: Wasted CPU cycles: http://www.javapractices.com/topic/TopicAction.do?Id=14

okram: Huh. K. Fixed.

dkuppitz:
Manual tests look good. `docker/build.sh -t -i -n` worked up to the point where it always fails for me in `tp32`.

VOTE: +1

spmallette:
VOTE +1

PR #482, opened 2016-11-08
TINKERPOP-1541: Select should default to Pop.last semantics

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1541

All `Scoping` steps use `Pop.last` semantics (`where()`, `match()`, `dedup()`), except `select()` which uses `Pop.mixed`. This is a wart from TinkerPop2 days. I changed `select()` to default to `Pop.last` and provided a `selectV3d0()` backwards compatibility step (@spmallette -- is that how we should name it? or is it `selectV3d2()` now?).

The benefit of this is that we can now introduce a simpler `Path` data structure if a traversal is all `Path.last` (which is 99% of traversals out there).  That is, if the traversal requirements is `LABELED_PATH` and all pops are `Pop.last`, then we don't need to store "historic" path elements and can have a `MapPath` which is backed by a `Map<String,Object>`. Less data and a simpler/faster equality test for bulking.

@robertdale --

```
gremlin> g.V(1).as('a').select(mixed,'a').as('a').select(mixed,'a')
==>[v[1],v[1]]
gremlin> g.V(1).as('a').select('a').as('a').select('a')
==>v[1]
```

VOTE +1.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,8 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.3.0 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Added `Pop.mixed` instead of using `null` to represent such semantics.
+* `select()`-step now defaults to using `Pop.last` instead of `Pop.mixed`. (*breaking*)
 * Removed previously deprecated `Console` constructor that took a `String` as an argument from `gremlin-console`.
 * Removed previously deprecated `ConcurrentBindings` from `gremlin-groovy`.
 * Removed previously deprecated `ScriptExecutor` from `gremlin-groovy`.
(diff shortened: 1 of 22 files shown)


spmallette:
It should be `selectV3d2()` given that it will yield backward compatibility to 3.2.x. 

Maybe I missed it, but are there any tests for `selectV3d2()` - at least something to exercise the old way of doing things?

okram:
The only difference is `select(Pop.mix,...)` (that is `selectV3d2()`) and there are 6 or so tests that test it in `RangeTest` and `TailTest`. I can explicitly make them use the method `selectV3d2()`, but as you can see, its simply just a call to `select(Pop.mix,...)`. I could make a `SelectV3d2Test` and just copy those range tests over and call those methods explicitly, but seems pointless as unlike `groupV320()`, its not a new step with different semantics, its the same step with just a different argument.

spmallette:
ok - if it's trivial, then perhaps not worrying about.

okram:
When I go to merge, I will change the method name to `selectV3d2()` so if you want to vote.... :) ... Also, I think we need @dkuppitz  and @robertdale  and @mhfrantz to have a look and vote too.

dkuppitz:
I'm cool with that change. I still have to run a `docker/build.sh -t -i -n` though - will do so as soon as I have the resources available (unless someone else volunteers).

dkuppitz:

VOTE: +1

robertdale:
VOTE: +1

PR #483, opened 2016-11-08
TINKERPOP-1547: Two bugs found associated with MatchStep: Path retraction and range count.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1547

Two bugs had always existed because we had commented out tests due to their failure. Fixed the bugs and exposed more tests. Also, some `match()` benchmarks got faster. Finally, I found a bug in `MatchStep.clone()` that only makes itself apparent (and randomly) when there is a `match()`-step within a local child of an OLAP traversal with single processor threading.

VOTE +1.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.2.4 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Added another optimization in `RangeByIsCountStrategy`, that removes `count().is()` altogether if it's not needed.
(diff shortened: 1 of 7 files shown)


twilmes:
Been having problems with traversal interrupted tests failing across many branches but I ran other tests including neo4j and everything passed.  Checked out match retraction updates and they look good.

VOTE: +1

spmallette:
@twilmes can you please send some specifics on what interruption tests were tanking on you and under what circumstances? Maybe just ping me on hipchat when you have a moment rather than go through it here since it seem unrelated to this PR.

dkuppitz:
Please don't merge this PR yet. While testing the changes in `RangeByIsCountStrategy` I realized that we can add a minor tweak to get another great optimization. I'll push my local changes as soon as `docker/build.sh -t -i` reports success.

okram:
@dkuppitz -- note that @spmallette never VOTE'd so when you are done with your testing and updates, just VOTE and I will merge.

dkuppitz:
Changes pushed. `docker/build.sh -t -i -n` succeeded.

VOTE: +1

PR #486, opened 2016-11-10
TINKERPOP-1545 IncidentToAdjacentStrategy is buggy

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1545

Made `SimplePathStep` implement `PathProcessor` and `IncidentToAdjacentStrategy` skip all traversals that contain path processing steps.

The first few test suites of `docker/builder -t -i -n` succeeded and I will let it run over night, but I'm confident, that nothing bad is gonna happen in the other test suites (otherwise I'm going to revert the next line tomorrow).

VOTE: +1

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/IncidentToAdjacentStrategy.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/IncidentToAdjacentStrategy.java
@@ -24,19 +24,17 @@ import org.apache.tinkerpop.gremlin.process.traversal.TraversalStrategy;
 import org.apache.tinkerpop.gremlin.process.traversal.step.LambdaHolder;
 import org.apache.tinkerpop.gremlin.process.traversal.step.map.EdgeOtherVertexStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.map.EdgeVertexStep;
-import org.apache.tinkerpop.gremlin.process.traversal.step.map.PathStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.map.VertexStep;
 import org.apache.tinkerpop.gremlin.process.traversal.strategy.AbstractTraversalStrategy;
+import org.apache.tinkerpop.gremlin.process.traversal.traverser.TraverserRequirement;
 import org.apache.tinkerpop.gremlin.process.traversal.util.TraversalHelper;
 import org.apache.tinkerpop.gremlin.structure.Direction;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.javatuples.Pair;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.Set;
 
 /**
@@ -63,7 +61,6 @@ public final class IncidentToAdjacentStrategy extends AbstractTraversalStrategy<
         implements TraversalStrategy.OptimizationStrategy {
 
     private static final IncidentToAdjacentStrategy INSTANCE = new IncidentToAdjacentStrategy();
-    private static final Set<Class> INVALIDATING_STEP_CLASSES = new HashSet<>(Arrays.asList(PathStep.class, LambdaHolder.class));
 
     private IncidentToAdjacentStrategy() {
     }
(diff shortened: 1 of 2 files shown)


okram:
`PathProcessors` are intended for path analysis steps that use `by()`-modulation. Also, what about `CyclicPathStep`? Perhaps  just itemize those steps explictly for now and now have them implement `PathProcessor`.

However...I guess they are `PathProcessors` that don't project out... hmmm... perhaps do a study of `PathProcessor` use and see if making `SimplePathStep` and `CyclicPathStep` one of those would be a problem.

dkuppitz:
The `PathProcessor` interface has no function for SimpleStep and doesn't add any overhead. AFAIK `PathProcessor` is only used by strategies to easily identify traversals with "special requirements". Both, `SimpleStep` and `CyclicStep`, access the current `traverser.path()` and thus they **process** the **path**, which makes them `PathProcessor`s :).

I can also rely on the result of `getRequirements()`, but this would make the strategy a bit slower and I really don't see why `SimplePath` and `CyclicPath` should not be `PathProcessor`s.

dkuppitz:
You know what, I will revert the change and not implement `PathProcessor` in `SimplePathStep`. Here's why: `IncidentToAdjacentStrategy` should only rely on the `getRequirements()` result, because otherwise this problem would come up again with `VertexProgramSteps`.

dkuppitz:
Force-pushed. Filtering steps based on 2 conditions (type and traversal requirements) was easier than I thought (thanks to `TraversalHelper.anyStepRecursively()`).

okram:
Why do you do a root recursion on every traversal? Why not use the "marker model" ? Also, `getRequirements()` is already recursive off the root traversal steps. I would do this:

1. If root -- search for `SimplePathStep`, `CyclicPathStep`, `LambdaHolder`, `Path`, `Tree`, ?
2. If one exists, mark all child traversals with hidden label.
3. If a child - am I marked, yes? remove the mark and return. No? do the rewrite.

dkuppitz:
> Why not use the "marker model" ?
This PR targets `tp31`.

okram:
Okay. The fear I have about `getRequirements()` and I don't like its use in compilation is that it gets cached. I forget if it gets reset after compilation..... but okay.

dkuppitz:
I really don't see what you mean by "it gets cached". `getRequirements()` as well as `getSelfAndChildRequirements()` look pretty stateless to me.

okram:
Look at line 142 of `DefaultTraversal`.

dkuppitz:
Oh, you mean `Traversal.getTraverserRequirements()`. That's useless anyways (in this case). The strategy has to check every single step.

okram:
If a step is a `TraversalParent`, then it uses that method.

dkuppitz:
I see, that sucks. Looks like there's nothing that resets the cache at any point. IMO the admin methods `addStep` and `removeStep` should invalidate the cache.

dkuppitz:
I think we have 2 options:

1. reset requirements (a simple `this.requirements = null;`) at the end of mutating methods (`addStep` and `removeStep`)
2. don't cache requirements before `applyStrategies()`

The problem with 1 is, that we would have to propagate the reset to all child- and parent-traversals.
The problem with 2 is, that if we don't cache requirements while startegies are being applied, why do we cache them at all? Are requirements used at runtime?

Thoughts?

dkuppitz:
Requirements were not cached in `tp31`. I have a neat solution for `master/`, but not for `tp32`. In `tp32` we can either reset the requirements after each strategy has been applied or we can ignore the caching, which - in the worst case - would make the traversal assume, that it has more requirements than it actually has. This is the only case I can currently think of:

When `IncidentToAdjacentStrategy` replaces `bothE().otherV()` with `both()`, the traversal could lose its `PATH` requirement.

okram:
This is a good find. However, the implementation is a bit gimpy.

`Step.getRequirements()` is already a recursive process as `TraversalParents` will check their children for requirements. I propose we go back to the `INVALIDATING_STEPS` model and just list all the steps that explicitly require `PATH`. That is, add the ones we were missing and keep the new test case you added to `IncidentToAdjacentStrategyTest`.

* `PathStep`
* `TreeStep`
* `CyclicPathStep`
* `SimplePathStep`
* `EdgeOtherVertexStep`
* `TreeSideEffectStep`

...then also add `LambdaHolder`.

dkuppitz:
Replaced by:

* https://github.com/apache/tinkerpop/pull/525 (tp31)
* https://github.com/apache/tinkerpop/pull/526 (tp32 + master)

PR #487, opened 2016-11-11
TINKERPOP-1399 NumberHelper needs to go into util and have a private constructor

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1399

Moved `NumberHelper` into `org.apache.tinkerpop.gremlin.util`. Note that it's not necessary to add a private constructor, since the only constructor that exists, is already private.

VOTE: +1


okram:
VOTE +0.99.

spmallette:
The title of this PR says to make `NumberHelper` have a "private constructor" - i don't see that change. And, as long as that's being done, you could also probably make the class `final` - I assume no one would extend that.

dkuppitz:
It only has a private constructor already. Making it final is a good call though.

spmallette:
oh - sorry - i see it now. i had to expand out the diff. i did a quick `mvn clean install` and all good.

VOTE +1

robertdale:
VOTE: +0.01

PR #488, opened 2016-11-11
TINKERPOP-1545 IncidentToAdjacentStrategy is buggy (tp32)

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1545

Made SimplePathStep implement PathProcessor and IncidentToAdjacentStrategy skip all traversals that contain path processing steps.

VOTE: +1


okram:
VOTE -1. 

I don't think its smart to make `SimplePathStep` a path processor as its not a path processor. I think you best bet for this ticket is to simply just add `SimplePathStep` and `CyclicPathStep` to your list of "invalidating steps."

dkuppitz:
This PR is no longer relevant, as we have a discussion going in https://github.com/apache/tinkerpop/pull/486 (where `SimplePathStep` is no longer a `PathProcessor`).

PR #490, opened 2016-11-13
TINKERPOP-1474 GraphSON decode of Vertex and Edge properties

bjmoor:
If the server sends properties back with the Vertices and Edges, gremlin_python should deserialize the properties and make them available.

This patch will add the properties to the returned Vertex, Edge, VertexProperty or Property, if and only if, the "properties" are returned.

By default, gremlin-server will not return Properties or Labels of elements.  However, when using a DetachedFactory HaltedTraverserStrategy, the server will return these properties.

### DIFF
--- a/gremlin-python/src/main/jython/gremlin_python/structure/io/graphson.py
+++ b/gremlin-python/src/main/jython/gremlin_python/structure/io/graphson.py
@@ -291,7 +291,10 @@ class VertexDeserializer(_GraphSONTypeIO):
 
     @classmethod
     def objectify(cls, d, reader):
-        return Vertex(reader.toObject(d["id"]), d.get("label", ""))
+        v = Vertex(reader.toObject(d["id"]), d.get("label", ""))
+        if "properties" in d:
+            v.properties = reader.toObject(d["properties"])
+        return v;
 
 
 class EdgeDeserializer(_GraphSONTypeIO):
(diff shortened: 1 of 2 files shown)


spmallette:
Thanks for this. Would you mind including tests for `Edge` and `VertexProperty`? Also, could you please edit your PR to have TINKERPOP-1474 as the prefix to the rest of your text in the name as in:

"TINKERPOP-1474 GraphSON decode of Vertex and Edge properties"

This change applies to that ticket I think.

okram:
Dah. I didn't see this ticket. I'm doing this (along with other things) in TINKERPOP-1520. Hmm... perhaps you can provide a PR to that branch ... In particular your test cases.

spmallette:
TINKERPOP-1420 has merged to master. @bjmoor do you want to rebase your PR? note this area of development is still somewhat in flux on master given:

Perhaps the best option would be to target this work at the tp32 branch where GraphSON 2.0 is stable. wdyt?

spmallette:
@okram pinged me on the side and filled me in with some background info i wasn't aware of - the short of it is that a lot of work has been done in this area on the aforementioned tickets and if you'd like to contribute you should probably coordinate with those working on TINKERPOP-1565 at this point.

bjmoor:
It looks like the work being done in TINKERPOP-1565 will be addressing the issue which this PR was attempting to address, and probably in a more effective manner.  Should I just close this PR?

spmallette:
yes - i think you can close this one since the work is already underway elsewhere. thanks for submitting this though.

PR #491, opened 2016-11-14
TINKERPOP-1538 Use non-default port for gremlin server tests

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1538

Used something other than 8182 to run the gremlin server tests to avoid port conflicts if gremlin server happens to be running in the background.

Ran `$ mvn clean install && mvn verify -pl gremlin-server -DskipIntegrationTests=false -DincludeNeo4j && mvn verify -pl gremlin-console -DskipIntegrationTests=false` to success.

VOTE +1


dkuppitz:
Looks good, tests are not affected by other applications using port 8182.

VOTE: +1

okram:
VOTE +1

PR #495, opened 2016-11-15
TINKERPOP-1502 & TINKERPOP-1482: Chained has-containers and AndP stringency

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1502
https://issues.apache.org/jira/browse/TINKERPOP-1482

This ticket accomplishes various things around `GraphTraversal.has()`. 

1.) A write-time optimization is provided where if a `HasStep` is the traversal's end step, simply `HasStep.addHasContainer()`. This reduces the time required by `InlineFilterStrategy` and provider strategies to fold in `HasContainers` and remove `HasSteps`. 

2.) `GraphTraversal.hasXXX()` was tweaked by @dkuppitz a release ago. He used `Object` to determine numerous things like `String`, `P`, etc.  This is bad. We want the `GraphTraversal` API to be explicit so I added stricter typing: e.g. `hasLabel(String,String...)` and `hasLabel(P<String>)`. 

3.) `HasContainer.makeHasContainers()` was a TinkerPop hack trying to help providers and it shouldn't be done. This method turned `AndP(x,y)` into individual `x` and `y` predicates so providers didn't have to analyze `AndP` trees.  We shouldn't do this as the provider will be getting `AndP` compositions elsewhere and will need to be able to handle it. I updated `Neo4jGraphStep` and `TinkerGraphStep` to process `AndP` predicates accordingly for index lookups. Added a note to upgrade docs for other providers --- simple update if they want to use the old "break apart `AndP`"-model.

4.) Found a bug in `InlineFilterStrategy` around `outE().hasLabel()` where `outE('knows').hasLabel('created')` was turned into `outE('knows','created')` and thus, semantically incorrect.

Going to add a few more tests before VOTING.


dkuppitz:
```
gremlin> g.V(1).outE("knows")
==>e[7][1-knows->2]
==>e[8][1-knows->4]
gremlin> g.V(1).outE("knows").hasLabel("created")
==>e[9][1-created->3]
```

This is still wrong. No `knows` edge will have a `created` label. The result should be empty. Furthermore I think that this is wrong:

```
gremlin> g.V(1).outE().hasLabel("knows").hasLabel("created")
==>e[7][1-knows->2]
==>e[8][1-knows->4]
==>e[9][1-created->3]
```

It should all boil down to `hasLabel(eq("knows").and(eq("created")))` / `has(label, eq("knows").and(eq("created")))`. Chained `has()` calls are / should be `and()`'ed, not `or()`'ed.

okram:
What you have in your examples is the "bug" that is fixed in this PR:

```
gremlin> g.V(1).outE('knows').hasLabel('created')
gremlin>
```

Likewise:

```
gremlin> g.V(1).outE().hasLabel("knows").hasLabel("created")
gremlin>
```

```
gremlin> g.V(1).outE().hasLabel("knows")
==>e[7][1-knows->2]
==>e[8][1-knows->4]
gremlin> g.V(1).outE().hasLabel("knows").explain()
==>Traversal Explanation
Original Traversal                 [GraphStep(vertex,[1]), VertexStep(OUT,edge), HasStep([~label.eq(knows)])]
...
InlineFilterStrategy         [O]   [GraphStep(vertex,[1]), VertexStep(OUT,[knows],edge)]
...
Final Traversal                    [TinkerGraphStep(vertex,[1]), VertexStep(OUT,[knows],edge)]
gremlin>
```

dkuppitz:
Alright, looks like I pulled too early, just got some more changes coming through. I will retest the latest.

dkuppitz:
Yea, looks better now. Not sure if we should really keep all the steps in `g.V(1).outE('knows').hasLabel('created').more().bla()`. It's like the Java compiler would keep stuff within a `if (false) { ... }` block. The static analysis already tells us, that nothing will make it through the filter.

You already treat this particular case in a dedicated code block, why not remove everything that comes after it and append a `.not(identity())`?

okram:
We could. However, its such a rare situation ............ too much optimization?

okram:
VOTE +1.

dkuppitz:
Since this PR touches `FilterRankingStrategyTest`s, it would be nice to have this issue fixed as well (and have some tests added to cover this scenario):

```
gremlin> g.V(1).as("a").both().both().has("age").where(neq("a")).by("age")
The property does not exist as the key has no associated value for the provided element: v[5]:age
Display stack trace? [yN]N
gremlin> g.V(1).as("a").both().both().has("age").where(neq("a")).by("age").explain()
==>Traversal Explanation
===========================================================================================================================================================================================================================================
Original Traversal                 [GraphStep(vertex,[1])@[a], VertexStep(BOTH,vertex), VertexStep(BOTH,vertex), TraversalFilterStep([PropertiesStep([age],value)]), WherePredicateStep(neq(a),[value(age)])]
...
FilterRankingStrategy        [O]   [GraphStep(vertex,[1])@[a], VertexStep(BOTH,vertex), VertexStep(BOTH,vertex), WherePredicateStep(neq(a),[value(age)]), TraversalFilterStep([PropertiesStep([age],value)])]
...
Final Traversal                    [TinkerGraphStep(vertex,[1])@[a], VertexStep(BOTH,vertex), VertexStep(BOTH,vertex), WherePredicateStep(neq(a),[value(age)]), NoOpBarrierStep(2500), TraversalFilterStep([PropertiesStep([age],property)])]
```

dkuppitz:
VOTE: +1

spmallette:
VOTE +1

PR #496, opened 2016-11-16
TINKERPOP-1556 allow developers to pass options to docker

robertdale:
This was a much smaller change than I expected. Originally I was adding some sysctl settings to the dockerfiles. In the end I discovered I could control that and much more from docker itself. So I'm going to go ahead and commit this small change.

example:
```
export DOCKER_OPTS="--cpuset-cpus 0-3 --sysctl net.ipv6.conf.all.disable_ipv6=1 --sysctl net.ipv6.conf.default.disable_ipv6=1"

./docker/build.sh -t -n -i
```


dkuppitz:
Isn't  the `DOCKER_OPTS` environment variable used by default? Or is this only true for the daemon?

robertdale:
Technically, neither, not even for the daemon. It's specific to certain distribution startup scripts. So there should be no overlap here.

dkuppitz:
Okay, cool, then...

VOTE: +1

PR #499, opened 2016-11-18
TINKERPOP-1520: Difference between 'has' step generated graphson2.0 in java and python glv implementation

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1520
https://issues.apache.org/jira/browse/TINKERPOP-1454

This branch has done alot around testing and GraphSON.

* We now verify that all translators create Traversals with bytecode identical to the original translated bytecode. (tested against all traversals in ProcessSuite)
* We now verify that Gremlin-Python's GraphSON Bytecode representation is identical to Gremlin-Java's. (tested against all traversals in ProcessSuite)
* We now verify that GraphTraversal is in one-to-one correspondence with Bytecode. That is, has("name") -> [has,name]. (bad ass reflection technique used here)
* There was a GraphSON 2.0 issue with Edge properties. Changed to the same format a VertexProperties. Smaller footprint and only necessary information.
* Added VertexID ("vertex"-field) to VertexProperty in GraphSON 2.0 -- how else will you attach?
* Added Element ("element"-field) to Property in GraphSON 2.0 -- how else will you attach?
* Added TraversalStrategy classes to GryoMapper as we (@spmallette !!! shame on you) simply dropped `withStrategies()` instructions in Gryo Bytecode serialization.
* TranslationStrategy is now smart to remove itself from the Bytecode. No longer do the Translators have to be wary of this (lots of code deleted).
* We now verify that Gryo created bytecode is identical to source bytecode.

...need I go on?

VOTE +1


okram:
Oh -- and I forgot. Gremlin-Python now has serializers for Vertex, Edge, Property, VertexProperty... Thus, you can `g.V(v[1])` from Gremlin-Python to Gremlin-Java.

CHANGELOG.asciidoc
```diff
@@ -26,6 +26,12 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.2.4 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Added "attachment requisite" `VertexProperty.element()` and `Property.element()` data in GraphSON serialization.
+* Added `Vertex`, `Edge`, `VertexProperty`, and `Property` serializers to Gremlin-Python and exposed tests that use graph object arguments.
+* `Bytecode.getSourceInstructions()` and `Bytecode.getStepInstructions()` now returns `List<Instruction>` instead of `Iterable<Instruction>`.
+* Added various `TraversalStrategy` registrations with `GryoMapper`.
+* Fixed a naming mistake in Gremlin-Python: `IdentityRemoveStrategy` is now called `IdentityRemovalStrategy`. (*breaking*)
```
spmallette: I've noticed that in the last few versions you've started adding "breaking" to changelog entries. I've been tying "breaking" to tickets that will then show up in the changelog on release.  We can start doing this from now i guess, but it's not been done consistently in changelog in any way. I guess that's a really minor breaking change which wouldn't warrant a separate ticket. I dunno - do we want to keep doing this and should it be done consistently going forward?

gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GroovyTranslator.java
```diff
@@ -47,8 +49,6 @@
  */
 public final class GroovyTranslator implements Translator.ScriptTranslator {
 
-    private static final boolean IS_TESTING = Boolean.valueOf(System.getProperty("is.testing", "false"));
```
spmallette: is "is.testing" completely gone now?

spmallette:
This is a nice body of work. So you also close this one with this PR?

https://issues.apache.org/jira/browse/TINKERPOP-1454

Any others?

As there are some breaking changes to graphson formatting, can you please add upgrade docs? also, the io documentation needs to be updated and regenerated.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
```diff
@@ -71,4 +80,89 @@ public void graphTraversalShouldHaveMethodsOfAnonymousGraphTraversal() {
             }
         }
     }
+
+    @Test
+    public void graphTraversalMethodsShouldGenerateCorrespondingBytecode() throws Exception {
+        final Random random = new Random();
```
okram: Lets keep it un-seeded. If it breaks down the road, we learn something. If not, then its good.

CHANGELOG.asciidoc
spmallette: mostly don't like it because we didn't do it from the start....my ocd is nagging at me about it now and there's no easy way to go back through and retroactively fix it. if you're cool to drop them, i'm fine with that.

okram:
Yea. I will do the upgrade and io docs. Integration testing Giraph is running right now :| ... once that completes, I will push doc updates.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
spmallette: Ok - I have a body of tests that are kinda like this. I add this:

https://github.com/apache/tinkerpop/blob/master/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerIntegrateTest.java#L375-L379

so that they only run when `-DassertNonDeterministic` is added to the `mvn clean install`. This should probably be one of those tests.

okram: Who ever does `-DassertNonDeterministic`? If we use that then its basically a non-test as I have never used/seen that flag. The point seems that tests that ARE random should be used as much as possible (`mvn clean install`-style) so you can scan the parameter space more and more and more.

okram:
```
[INFO] BUILD SUCCESS
[INFO] Total time: 05:10 h
[INFO] Finished at: 2016-11-18T11:57:52-07:00
[INFO] Final Memory: 158M/1657M
```

newkek:
Indeed there seems to be some breaking changes in the protocol. It seems ok for pure TinkerPop since the parsing code is updated accordingly but will cause problems to implementors, just as a note (maybe only for my own remembering)...

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONSerializersV2d0.java
```diff
@@ -175,8 +177,30 @@ public PropertyJacksonSerializer() {
         public void serialize(final Property property, final JsonGenerator jsonGenerator, final SerializerProvider serializerProvider)
                 throws IOException {
             jsonGenerator.writeStartObject();
-            jsonGenerator.writeObjectField(GraphSONTokens.KEY, property.key());
+            jsonGenerator.writeStringField(GraphSONTokens.KEY, property.key());
             jsonGenerator.writeObjectField(GraphSONTokens.VALUE, property.value());
+            if (property.element() instanceof VertexProperty) {
+                VertexProperty vertexProperty = (VertexProperty) property.element();
+                jsonGenerator.writeObjectFieldStart(GraphSONTokens.ELEMENT);
+                jsonGenerator.writeStringField(GraphSONTokens.VALUETYPE, "g:VertexProperty");
```
newkek: It isn't great to bypass the typing system.. As the type system is dynamic, hardcoding the type can lead to inconsistent types if users overrides the Gremlin standard serializers and types as it is currently possible. But I see why we can't just jsonGenerator.writeObjectField(GraphSONTokens.ELEMENT, property.element()), because that would, well, re-write VertexProperty's properties and so on. We can maybe introduce a "g:Element" type? Or maybe just write a simple Map here where the deserializers are aware of what to do with a "element" field? wdyt?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONSerializersV2d0.java
```diff
@@ -175,8 +177,30 @@ public PropertyJacksonSerializer() {
...
+                jsonGenerator.writeObjectFieldStart(GraphSONTokens.VALUEPROP);
+                jsonGenerator.writeObjectField(GraphSONTokens.ID, vertexProperty.id());
+                jsonGenerator.writeStringField(GraphSONTokens.LABEL, vertexProperty.label());
+                jsonGenerator.writeObjectField(GraphSONTokens.VERTEX, vertexProperty.element().id());
+                jsonGenerator.writeEndObject();
+                jsonGenerator.writeEndObject();
+            } else if (property.element() instanceof Edge) {
+                Edge edge = (Edge) property.element();
+                jsonGenerator.writeObjectFieldStart(GraphSONTokens.ELEMENT);
+                jsonGenerator.writeStringField(GraphSONTokens.VALUETYPE, "g:Edge");
```
newkek: This will write an object of type "g:Edge" but will not contain the same content than what a Edge serializer would have serialized, and so the contents are inconsistent. Here what is missing is the out/inVLabel

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONSerializersV2d0.java
```diff
@@ -299,6 +325,7 @@ public void serialize(final TraversalExplanation traversalExplanation, final Jso
                 intermediate.put(GraphSONTokens.STRATEGY, pair.getValue0().toString());
                 intermediate.put(GraphSONTokens.CATEGORY, pair.getValue0().getTraversalCategory().getSimpleName());
                 intermediate.put(GraphSONTokens.TRAVERSAL, getStepsAsList(pair.getValue1()));
+                intermediate.put(GraphSONTokens.TRAVERSAL, getStepsAsList(pair.getValue1()));
```
newkek: Is this line supposed to be duplicated?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONSerializersV2d0.java
```diff
@@ -459,9 +486,33 @@ public PropertyJacksonDeserializer() {
...
+        }
+
+        @Override
+        public VertexProperty createObject(final Map<String, Object> propData) {
+            return propData.containsKey(GraphSONTokens.VERTEX) ?
+                    new DetachedVertexProperty<>(
+                            propData.get(GraphSONTokens.ID),
+                            (String) propData.get(GraphSONTokens.LABEL),
+                            propData.get(GraphSONTokens.VALUE), (Map<String, Object>) propData.get(GraphSONTokens.PROPERTIES),
+                            new DetachedVertex(propData.get(GraphSONTokens.VERTEX), Vertex.DEFAULT_LABEL, null)) :
```
newkek: Supposing it's a Vertex with VertexProperties that is deserialized (and not only a VertexProperty on its own), will this really make the VertexProperty link to the right parent Vertex? It looks like this will create another one. However, I've recently found a nice way to make a deserializer sort of "context aware" in order to, when a Vertex deserializer deserializes its own VertexProperties, it can give to the VertexProperty deserializers the Vertex object that's "above" them - the VertexProperty's parent. I haven't had time to propose a fix here in TP but I intended to do so soon if you guys are interested

newkek:
I sneaked in and made a few comments I hope that's ok

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
spmallette: I use it from time to time during normal development and it is certainly used before a release. We can't have random in tests where we don't have a clue how to reproduce the error. It leads to tons of wasted time and frustration when it happens. Additionally, we already mostly ignore our automated build systems because of false failure and the aim should be to get rid of things that cause us to doubt failure in there - randomness is one of those things. 

That said, I admit that there are tests that benefit from randomness and this is a good candidate for one. I'd offer two options:

1. In the short term (for this PR), we log the seed that is being used to the build output. In that way, if there is a failure on this test we can edit the code, plug in the seed for the failure and presumably recreate the failure. 
2. In the long term (a new ticket), we look at doing this more generally - maybe `assertNonDeterministic` takes a seed instead of being binary true/false or we use [junit-quickcheck](http://pholser.github.io/junit-quickcheck/site/0.6.1/) which is built precisely for this kind of situation.

wdyt?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONSerializersV2d0.java
okram: You can't always expect `outVLabel` and `inVLabel` to exist. Think about `ReferenceEdge`. It would be too expensive to always include such information. The most important think about the serialization is "can it be attached?" The way GraphSON 2.0 was designed, this was not the case. How do you attach a property? an vertex property? Finally, given the attach requirements of an edge, you have the problem of:

property -> edge -> vertex (which vertex -- in or out?) ...  

The amount of data required to attach a property would be pretty nasty. Hence, the "reference" serialization is smaller.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONSerializersV2d0.java
```diff
@@ -175,8 +177,30 @@ public PropertyJacksonSerializer() {
...
+            jsonGenerator.writeStringField(GraphSONTokens.KEY, property.key());
             jsonGenerator.writeObjectField(GraphSONTokens.VALUE, property.value());
+            if (property.element() instanceof VertexProperty) {
+                VertexProperty vertexProperty = (VertexProperty) property.element();
+                jsonGenerator.writeObjectFieldStart(GraphSONTokens.ELEMENT);
+                jsonGenerator.writeStringField(GraphSONTokens.VALUETYPE, "g:VertexProperty");
+                jsonGenerator.writeObjectFieldStart(GraphSONTokens.VALUEPROP);
+                jsonGenerator.writeObjectField(GraphSONTokens.ID, vertexProperty.id());
+                jsonGenerator.writeStringField(GraphSONTokens.LABEL, vertexProperty.label());
+                jsonGenerator.writeObjectField(GraphSONTokens.VERTEX, vertexProperty.element().id());
```
okram: Because a `VertexProperty` is attached to a `Vertex` and thus, generalized `Element` semantics would only cause a chain of JSON unnecessary when `Vertex` attachement only requires a single ID -- the vertex id.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONSerializersV2d0.java
okram: dunno. never touched that. I removed it in this PR though. pushed.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONSerializersV2d0.java
```diff
@@ -153,13 +152,16 @@ public void serialize(final Edge edge, final JsonGenerator jsonGenerator, final
...
-            final Iterator<Property<Object>> elementProperties = normalize ?
+            final Iterator<Property<Object>> edgeProperties = normalize ?
                     IteratorUtils.list(edge.properties(), Comparators.PROPERTY_COMPARATOR).iterator() : edge.properties();
-            if (elementProperties.hasNext()) {
+            if (edgeProperties.hasNext()) {
                 jsonGenerator.writeFieldName(GraphSONTokens.PROPERTIES);
 
                 jsonGenerator.writeStartObject();
-                elementProperties.forEachRemaining(prop -> safeWriteObjectField(jsonGenerator, prop.key(), prop));
+                while (edgeProperties.hasNext()) {
```
newkek: I had noticed the discrepancy in that VertexProperty's Properties are not serialized as Properties but instead the values directly are serialized and that's an oversight from when I implemented GraphSON2. I understand that the same thing is applied here and it makes a smaller footprint but I think we should rather fix the VertexProperty's properties to be serialized as Map of Properties and not Map of the values, because it reflects directly what is going on underneath and so the deserialization process does not need to any transformation, just set the fields of the newly created object with the rest of the deserialized data, rather than creating. Also it is not consistent because each Vertex, Edge, and VertexProperty derive from Element, which has the same `properties` field, a list of  `? extends Property` so we should come up with the same format for all I believe and here a Vertex would have its properties serialized objects, but the others would have their properties serialized as the values, not sure if that's a major concern but I would advocate for consistency as much as possible

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONSerializersV2d0.java
```diff
@@ -153,13 +152,16 @@ public void serialize(final Edge edge, final JsonGenerator jsonGenerator, final
...
+                    final Property<?> property = edgeProperties.next();
+                    jsonGenerator.writeObjectField(property.key(), property.value());
```
newkek: I had noticed the discrepancy in that VertexProperty's Properties are not serialized as Properties but instead the values directly are serialized and that's an oversight from when I implemented GraphSON2. I understand that the same thing is applied here and it makes a smaller footprint but I think we should rather fix the VertexProperty's properties to be serialized as Map of Properties and not Map of the values, because it reflects directly what is going on underneath and so the deserialization process does not need to any transformation, just set the fields of the newly created object with the rest of the deserialized data, rather than creating. Also it is not consistent because each Vertex, Edge, and VertexProperty derive from Element, which has the same properties field, a list of ? extends Property so we should come up with the same format for all I believe and here a Vertex would have its properties serialized objects, but the others would have their properties serialized as the values, not sure if that's a major concern but I would advocate for consistency as much as possible

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONSerializersV2d0.java
okram: I don't quite follow what you are saying, but if you have a cool idea with "context awareness," by all means, push a patch to this branch.

Here is the thing about being able to attach an element back to the main graph.

* `Vertex`: All you need is the vertex ID.
* `Edge`: Edge ID (but for most dbs, outV id and edge label too)
* `VertexProperty`: vertex property ID, vertex ID (but for optimization in most dbs, label too).
* `Property`: key/value, element id. Given that properties are ONLY on vertex properties and edges, then you need the requisite information about the element for attachment.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONSerializersV2d0.java
okram: Why don't you propose a GraphSON 2 format for Vertex, Edge, Property, VertexProperty in a JSON document that we can look at. From there, without how to implenent the serializers/deserailizers, we can see your arguments separate from code. Then if they are good, we take the respective code patch. Thoughts?

newkek: Also I had waited before correcting the discrepancy mentioned above for the VertexProperty's properties because it's a breaking change and it seems dangerous to make it part of GraphSON 2.**0** in my opinion

newkek: Basically I had written [this for some specs](https://gist.github.com/newkek/15e9933dc4b43674c2c60bc9487fc80e#file-graph-types-asciidoc) (note: not sure the Metrics and TraversalMetrics ones are still up to date but the rest should be).
What I suggest is [the revision here](https://gist.github.com/newkek/15e9933dc4b43674c2c60bc9487fc80e/revisions#diff-5a08d02157ba6ec13bacc40ab3a3a7bc).

okram:
Please put it in a form like: https://github.com/apache/tinkerpop/blob/e09309c1210ad5d7d5f33440d4d8d1470bd3f0e2/docs/src/dev/io/graphson.asciidoc .. What you mean by "data" doesn't make sense to me. If we see:

```json
{ "@type" : "g:Blah",
  "@value": {
    "x" : "y",
    "a" : "b"
  }
}
```

...it will be more clear what you are proposing.

newkek:
https://gist.github.com/newkek/bfc976712f1adcbe76bcea4e34435ea2 - the rest doesn't change

okram:
If we do this, then we would have to change `Edge` back to the same model to be consistent. However, why change it back? We gain the benefit of less verbosity.

newkek:
> If we do this, then we would have to change Edge back to the same model to be consistent

Indeed.

> why change it back?

Because currently it is not consistent, (Detached)Edge/Vertex/VertexProperty all derive from (Detached)Element, for which `properties` have the same structure, but with what's currently on the PR, Edge and VertexProperty don't have their `properties` serialized in the same format than Vertex.
If we unify all we can make generalize the deserialization of an object that derives from Element and make the deserialization code more readable.

In my opinion I don't believe verbosity should be chosen over consistency here, especially since GraphSON 2 is already very very verbose and not really meant to be compact.

okram:
Talking with @spmallette. There are lots of issues with GraphSON 2.0 that both you and I notice. Lets do this. Lets get this ticket merged as there are lots of other work in here and start up a new ticket for proposed GraphSON 2.0 changes that will target the same branch as the merge. I can create a new JIRA and will add the link.

spmallette:
@okram what aspect of this PR should we be considering then for purposes of review? do you have a brief synopsis of that and what is on the table for change later?

spmallette:
> In my opinion I don't believe verbosity should be chosen over consistency here, especially since GraphSON 2 is already very very verbose and not really meant to be compact.

fwiw, the verbosity issue of GraphSON has been kicked around the dev list for some time and the consensus tends to be that very few are really concerned about compactness and by "very few" i mean no one has voiced that concern in those discussions.

okram:

okram:
@spmallette --- we simply can't abandon the GraphSON 2.0 changes I made cause we need to be able to "attach properties" (vertex properties and element properties). However, there are decisions to be made on how to do that (as outlined by @newkek). Moreover, there are more general decisions to be made about GraphSON 2.0 (https://issues.apache.org/jira/browse/TINKERPOP-1565). I think we should simply review this for what it is (ignore discussions of "best GraphSON 2.0" representation and be sure that we do https://issues.apache.org/jira/browse/TINKERPOP-1565 to come to a consensus on GraphSON 2.0 for the next release.

okram:
Another thing we can do that I think would be best is target this to `master/`. Why?

1. There are changes to Gryo we will need to make so ensure "attach-ability."
2. This gives us more leeway around breaking GraphSON 2.0 changes.

spmallette:
If we do that will there be "bad" bugs left in tp32 that won't get fixed?

okram:
This ticket doesn't address a "bug" as much as a consistency issue. The primary issue is that Bytecode and GraphTraversal are not in one-to-one correspondence. You get the same "answer," but for example: `hasId(1)` is represented in bytecode as `has(T.id,1)`. We can just kick this consistency verification to 3.3.0.

newkek:
Thanks for the separate ticket, I'll put comments over there. So All the following items (from the PR desc):

> - There was a GraphSON 2.0 issue with Edge properties. Changed to the same format a VertexProperties. Smaller footprint and only necessary information.
- Added VertexID ("vertex"-field) to VertexProperty in GraphSON 2.0 -- how else will you attach?
- Added Element ("element"-field) to Property in GraphSON 2.0 -- how else will you attach?

Will be handled separately in TINKERPOP-1565, and GraphSON2 will not change in this PR, right? (just trying to understand because it has external impact to change GraphSON2)

okram:
No GraphSON will change in this branch but we will solve TINKERPOP-1562 in the same release as this branch. Thus, this branch is like a "half-ticket" with TINKERPOP-1562's completion being the finale. Note that I also targeted this branch to `master/` so we can have breaking changes to GraphSON 2.0 for TinkerPop 3.3.0.

okram:
Integration tests ran against `master/` (the new target branch).

```
[INFO] BUILD SUCCESS
[INFO] Total time: 04:14 h
[INFO] Finished at: 2016-11-22T23:02:41-07:00
[INFO] Final Memory: 179M/1656M
```

dkuppitz:

VOTE: +1

(don't blame me for voting on half-done work; @okram assured me that we won't have a release before the work, started in this PR, is actually done)

okram:
You are a bad person @dkuppitz.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
okram: Ah yea. Thats smart. Do you want to push that to this branch or tell me how to do it with an example?

spmallette: I can amend the branch

spmallette:
I added the logging of the seed for that test that uses `Random`: 73859cce8563c3951e39fef58b189acb9718e5c3 

It's made for one time use right now - we probably could generalize that a bit for re-use in other tests that use `Random` (and it doesn't make sense to wholly factor out the `Random`).  

VOTE +1

PR #501, opened 2016-11-18
Added TP2-to-TP3 GraphML XSLT to resources

joshsh:
In gremlin-core as suggested by @spmallette

### DIFF
--- a/gremlin-core/src/main/resources/tp2-to-tp3-graphml.xslt
+++ b/gremlin-core/src/main/resources/tp2-to-tp3-graphml.xslt
@@ -1,0 +1,40 @@
+<?xml version="1.0" ?>
+<!-- XSL stylesheet to convert TinkerPop v2 GraphML files for Apache TinkerPop v3 -->
+<xsl:stylesheet version="1.0"
+                xmlns="http://graphml.graphdrawing.org/xmlns"
+                xmlns:graphml="http://graphml.graphdrawing.org/xmlns"
+                xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+                exclude-result-prefixes="graphml">
+    <xsl:output method="xml" indent="yes" omit-xml-declaration="yes"/>
+    <xsl:strip-space elements="*"/>
+
+    <xsl:template match="node()|@*">
+        <xsl:copy>
+            <xsl:apply-templates select="node()|@*"/>
+        </xsl:copy>
+    </xsl:template>
+
+    <xsl:template match="graphml:graphml">
+        <graphml>
+            <key id="labelV" for="node" attr.name="labelV" attr.type="string"/>
+            <key id="labelE" for="edge" attr.name="labelE" attr.type="string"/>
+            <xsl:apply-templates/>
+        </graphml>
+    </xsl:template>
+
+    <xsl:template match="graphml:node">
+        <node>
+            <xsl:apply-templates select="node()|@*"/>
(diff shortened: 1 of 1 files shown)


okram:
Can you please make corresponding JIRA tickets and use the PR naming convention for your PRs? Also, this PR will need documentation as well as an update to the CHANGELOG. Finally, I don't know why this is targeted to `master/` when it seems it can go into `tp31/`.

spmallette:
@joshsh - thanks - note that we still support 3.1.x (mostly for maintenance) on the tp31 branch, 3.2.x on the tp32 branch (for maintenance and non-breaking new features) and then do all dev for the future 3.3.x on master (which is all new stuff including breaking changes). In this case, I don't think it would hurt to put this on tp31 as @okram suggests - it might be a useful little bit of functionality for someone still on  3.1.x and shouldn't cause any conflicts. 

Is there an easy test you could write that just exercises the XSLT a little bit? Perhaps you could just put something in `gremlin-test` in `IoTest` as we have tests in there like `shouldReadLegacyGraphSON` which would be similar to what you have here. Maybe just copy that model:

https://github.com/apache/tinkerpop/blob/492c8e160743bb6d15bc07262a47275d18242eed/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/structure/io/IoTest.java#L438-L452

and use your XSLT instead of the `LegacyGraphSONReader`?

joshsh:
Thanks for the info, @spmallette.  I will also look harder at graph migration as in IoGraphTest.  Note that I do not appear to have the create-issue permission in JIRA.

spmallette:
Anyone who has an account can create issues. In fact, I just checked and I think you should even have right such that you can assign them to yourself (i see your account there as "joshsh"). Can you please check again?

pluradj:
Created https://issues.apache.org/jira/browse/TINKERPOP-1608

gremlin-core/src/main/resources/tp2-to-tp3-graphml.xslt
```diff
@@ -0,0 +1,40 @@
...
+    </xsl:template>
+
+    <xsl:template match="graphml:edge">
+        <edge id="{@id}" source="{@source}" target="{@target}">
+            <data key="labelE">
+                <xsl:value-of select="@label"/>
+            </data>
+        </edge>
+    </xsl:template>
+
```
pluradj: Appears that this transform creates the `labelE` child node, but the existing edge properties are not carried over.

gremlin-core/src/main/resources/tp2-to-tp3-graphml.xslt
```diff
@@ -0,0 +1,40 @@
...
+            <xsl:apply-templates select="node()|@*"/>
+            <data key="labelV">vertex</data>
+        </node>
+    </xsl:template>
+
+    <xsl:template match="graphml:edge">
+        <edge id="{@id}" source="{@source}" target="{@target}">
+            <data key="labelE">
+                <xsl:value-of select="@label"/>
+            </data>
```
pluradj: Adding `<xsl:copy-of select="node()"/>` here seems to work.

joshsh: Thanks for catching that.  My test file had no edge properties.

spmallette:
Closed in favor of #595

PR #502, opened 2016-11-18
Tweaks for broad JDK8 support

joshsh:


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/TraversalExplanation.java
```diff
@@ -106,7 +107,12 @@ public String prettyPrint(final int maxLineLength) {
         if (maxTraversalColumn < 1)
             throw new IllegalArgumentException("The maximum line length is too small to present the " + TraversalExplanation.class.getSimpleName() + ": " + maxLineLength);
         int largestTraversalColumn = Stream.concat(Stream.of(Pair.with(null, this.traversal)), this.strategyTraversals.stream())
-                .map(Pair::getValue1)
+                .map(new Function<Pair<? extends Object,? extends Traversal.Admin<?,?>>, Object>() {
+                    @Override
+                    public Object apply(Pair<? extends Object, ? extends Traversal.Admin<?, ?>> objects) {
+                        return objects.getValue1();
+                    }
+                })
```
dkuppitz: This is pretty verbose. What's wrong with the `Pair::getValue1` method reference?

robertdale:
I like the refactoring.  But why go backwards in java version?  Is this for source or target compatibility?  If it's source, I can't imagine as a dev you don't have the ability to upgrade to something released in the past year. As for target, let alone that java 8 should work on java 8,  the CVEs should scare anyone enough to not run anything less than the latest.

@okram You commented for using Java 1.8.0_40. Can you recall what/why?

> tinkerpop.xml master pom requires the build to be on Java 1.8.0_40+.

spmallette:
There was a bug in the JDK before 40 - it definitely shouldn't be used.

joshsh:
Alrighty, then.  I will just cherry-pick the second of those commits.

PR #504, opened 2016-11-20
TINKERPOP-1548: Traversals can complete before interrupted in TraversalInterruptionComputerTest

twilmes:
I added a short sleep to prevent traversal from finishing before it can be interrupted during `TraversalInterruptionComputerTest`.

This is a small update but I want to make sure this doesn't cause any build issues for folks.  I was getting this failure before the fix regularly.  After this fix, I've had 10 good runs without issue also `docker/build.sh -t -i -n` succeeded.

VOTE: +1

### DIFF
--- a/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalInterruptionComputerTest.java
+++ b/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalInterruptionComputerTest.java
@@ -75,6 +75,17 @@ public class TraversalInterruptionComputerTest extends AbstractGremlinProcessTes
             try {
                 final Traversal traversal = traversalMaker.apply(g).sideEffect(traverser -> {
                     startedIterating.countDown();
+                    try {
+                        // artificially slow down the traversal execution so that
+                        // this test has a chance to interrupt it before it finishes
+                        Thread.sleep(100);
+                    } catch (InterruptedException e) {
+                        // if the traversal is interrupted while in this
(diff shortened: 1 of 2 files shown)


robertdale:
I've seen this on master.

VOTE: +1

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalInterruptionComputerTest.java
```diff
@@ -86,7 +97,7 @@ public void shouldRespectThreadInterruptionInVertexStep() throws Exception {
 
         // total time for test should not exceed 5 seconds - this prevents the test from just hanging and allows
         // it to finish with failure
-        assertThat(startedIterating.await(5000, TimeUnit.MILLISECONDS), CoreMatchers.is(true));
+        assertThat(startedIterating.await(1000, TimeUnit.MILLISECONDS), CoreMatchers.is(true));
```
robertdale: Why not just leave this as 5s?  Seems to work against the fix.  What is the actual timeout time value?

twilmes: When I determined the timing issue was related to the traversal finishing first as opposed to it not having enough time to start I figured I could experiment with lowering this but it was admittedly a shot in the dark at what the "right" number is.

spmallette:
I did several local and docker builds and it works - of course, it never really failed on my system. Travis is happy too where we sometimes saw failure on this so given that:

VOTE +1

PR #507, opened 2016-11-29
TINKERPOP-1573: Bindings don't work in coalesce

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1573

The solution to this problem uses `ThreadLocal`. It sucks, but I don't know of any other way of doing this. Whats funny is that I knew this problem existed as in my notes, the `ThreadLocal` solution is what I came up with back then. We just never had a test case the demonstrated the anonymous traversal bindings problem and thus, I never implemented the "solution."

VOTE +1.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.2.4 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Deprecated `TraversalSource.withBindings()` as it is no longer needed in Gremlin-Java and never was needed for other variants.
 * Fixed a bug in Gremlin-Java `Bytecode` where anonymous traversals were not aware of parent bindings.
 * Fixed a bug in Gremlin-Java GraphSON deserialization around `P.within()` and `P.without()`.
 * Converted Spark process suite tests to "integration" tests.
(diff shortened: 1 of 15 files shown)


robertdale:
`docker/build.sh -t -i -n` passes.

VOTE: +1

dkuppitz:
I don't like the implementation of `Bindings`. Suggestion for an improved implementation: https://gist.github.com/dkuppitz/8bd896d639c3fd8fc27fa33f832e68ef

okram:
@dkuppitz  --- while your model looks more "natural," the problem is that we will create a `Bindings` instance in `ThreadLocal` for every `Bytecode` creation. Given that `Bindings` are rare (especially given that they are only useful for ScriptEngine languages (and thus only when lambdas are used) -- not Gremlin-Java), then with the model I have, we don't create a bunch of `ThreadLocal` values for "no reason."

dkuppitz:
Okay, I see what you're saying. I fixed that and here's another version:
https://gist.github.com/dkuppitz/47bae08ab19c848f663d95c7213da26b

What I'm worried about in your implementation is the `MAP.get()` in every method. Look into the `ThreadLocal` implementation how nasty this call can get.

okram:
So the most frequently called method will be `Bindings.getBoundVariable()`. It is called on every `Bytecode.convertArguments()` which is called on every instruction addition. Next, one of primary benefits of the static nature of `Bindings` (now) is that there is no longer a `transient` reference to them in `Bytecode` as we can statically access the data.

In terms of the cost of `ThreadLocal.get()`. Dunno..... perhaps a `clock()`? I can't imagine it being too crazy given that you will never really have that many entries. ?

dkuppitz:
Fine. I still think the code sucks (you know, usually I'm dazzled by the beauty of the code you write, but this time, maaaan...). Anyways..

VOTE: +1

okram:
I know it sucks. I don't like this any more than you do. Here is the saving grace -- if someone comes up with a better solution, there are only two classes we need to tweak (`Bytecode` and `Bindings`). `Bindings` (besides as a singleton) are never accessed outside of these two classes. Moreover, most of the `Bindings` methods are `protected` and thus, not exposed to users. Only `of` is exposed.

I'll merge and perhaps someone down the road comes up with the better solution. ?.

dkuppitz:
Yea, go ahead and merge it. One day I'll be bored and rewrite this stuff. Maybe around Xmas.

PR #508, opened 2016-11-30
TINKERPOP-1363 Cleanup Docker build script for next major release

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1363

Now that i squashed all my commits into 2 commits, it looks like this was all done in no time, what a shame!

Anyway, Python stuff is now part of the base image (this change requires that you - the tester - rebuild your Docker images; just drop the old ones).

I've also cleaned up the docs, which contained some dead links to a Groovy site that longer exists. Furthermore I updated `grapeConfig.xml` - in the docs and the one used by the Docker build script. Note that your local `grapeConfig.xml` may look different and still work, but I swear, what we have in this PR is the only way to make it work a) in Docker and b) on brand new development machines. Any other configuration will likely lead to failures when you try to install Console plugins.

VOTE: +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.3.0 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Updated Docker build scripts to include Python dependencies (NOTE: users should remove any previously generated TinkerPop Docker images).
 * Added "attachment requisite" `VertexProperty.element()` and `Property.element()` data in GraphSON serialization.
 * Added `Vertex`, `Edge`, `VertexProperty`, and `Property` serializers to Gremlin-Python and exposed tests that use graph object arguments.
 * `Bytecode.getSourceInstructions()` and `Bytecode.getStepInstructions()` now returns `List<Instruction>` instead of `Iterable<Instruction>`.
--- a/docker/resources/groovy/grapeConfig.xml
+++ b/docker/resources/groovy/grapeConfig.xml
@@ -1,15 +1,22 @@
 <ivysettings>
   <settings defaultResolver="downloadGrapes"/>
+  <property name="m2-pattern" value="${user.home}/.m2/repository/org/apache/tinkerpop/[module]/[revision]/[module]-[revision](-[classifier]).[ext]" />
+  <property name="m2-pattern-ivy" value="${user.home}/.m2/repository/org/apache/tinkerpop/[module]/[revision]/[module]-[revision](-[classifier]).pom" />
+  <caches>
+    <cache name="nocache" useOrigin="true" />
+  </caches>
   <resolvers>
     <chain name="downloadGrapes">
+      <filesystem name="local-maven2" checkmodified="true" changingPattern=".*" changingMatcher="regexp" m2compatible="true" cache="nocache">
+        <artifact pattern="${m2-pattern}"/>
+        <ivy pattern="${m2-pattern-ivy}"/>
+      </filesystem>
       <filesystem name="cachedGrapes">
         <ivy pattern="${user.home}/.groovy/grapes/[organisation]/[module]/ivy-[revision].xml"/>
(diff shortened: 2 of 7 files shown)


docker/resources/groovy/grapeConfig.xml
```diff
@@ -1,15 +1,22 @@
...
+        <ivy pattern="${m2-pattern-ivy}"/>
+      </filesystem>
       <filesystem name="cachedGrapes">
         <ivy pattern="${user.home}/.groovy/grapes/[organisation]/[module]/ivy-[revision].xml"/>
         <artifact pattern="${user.home}/.groovy/grapes/[organisation]/[module]/[type]s/[artifact]-[revision].[ext]"/>
       </filesystem>
-      <ibiblio name="codehaus" root="http://repository.codehaus.org/" m2compatible="true"/>
-      <ibiblio name="central" root="http://central.maven.org/maven2/" m2compatible="true"/>
-      <ibiblio name="java.net2" root="http://download.java.net/maven/2/" m2compatible="true"/>
-      <ibiblio name="hyracs-releases" root="http://obelix.ics.uci.edu/nexus/content/groups/hyracks-public-releases/" m2compatible="true"/>
```
spmallette: maybe you learned something new about grape, but why does this work without these entries? if giraph isn't in the local /.m2 or ./groovy then how will it get those dependencies?

robertdale:
Building world `docker/build.sh -t -i -n -d ` passes.
Building docs only `docker/build.sh -d ` passes.
Skimmed reference docs through browser.

VOTE: +1

docker/resources/groovy/grapeConfig.xml
dkuppitz: Giraph is now available in Maven central, hence no reason to keep this exotic repository in the list: https://mvnrepository.com/artifact/org.apache.giraph

twilmes:
`docker/build.sh -t -i -n -d` passed and I took a look at the docs and confirmed they generated correctly.

VOTE: +1

CHANGELOG.asciidoc
```diff
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.3.0 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Updated Docker build scripts to include Python dependencies (NOTE: users should remove any previously generated TinkerPop Docker images).
```
pluradj: should add this note to docs/src/upgrade/release-3.3.x.asciidoc

pluradj:
`docker/build.sh -t -i -n -d` worked for me. I ran a Docker build against `tp31` afterwards, and that worked too.

VOTE: +1

CHANGELOG.asciidoc
spmallette: so far, upgrade docs have really just been for "users" and "providers" not for "TinkerPop Developers" so i'm not sure where something like that would go.  It doesn't fit too well there, but I suppose you could add it to the "users" section and it would make sense. I suppose it would just need to be made clear that the "docker stuff" is for purpose of "building tinkerpop source" or something like that. I doubt many end users really spend much time building tinkerpop, but I could be wrong. 

Anyway, i don't feel strongly about it one way or the other - just my two cents.

dkuppitz: Yea, I didn't add it there, because it's neither for users nor for providers. I could add a `Upgrading for Developers` section... Thoughts?

spmallette: i'm -0.5 on a new section.  the developer docs just need to stay current imo. if someone feels +1.5 then go for it ;)

robertdale: Maybe a reminder to dev list at release time will be sufficient.

dkuppitz: Okay, reminder to dev list sounds good to me (although I think that all the people who actually use the Docker script, participated in the review of this PR). Final word on the note in the upgrade docs? I tend to say we don't need a note there.

spmallette: +1 for a reminder. i keep a list of notes of stuff like that to include in announcements (and usually try to ask around if anyone has anything to add). I'll add this one to my list.

pluradj: I'm fine with leaving it out.

PR #510, opened 2016-12-01
TINKERPOP-1562 Migrate ScriptEngine-related code to gremlin-core

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1562

Apologies in advance - this is a fat, nasty PR and I doubt there would have been a way to break this up into smaller bits of work. Every change I made ended up creating new branches of stuff that needed to be touched. As it turns out `ScriptEngine` stuff was deeply nested into all areas of the code base and unraveling it all into a way that was non-breaking for 3.2.x was non-trivial.

The core of this work focuses on:

* Moving `ScriptEngine` and `GremlinPlugin`  related stuff out of gremlin-groovy down to gremlin-core. 
* Given the above change that meant:
** Taking each existing gremlin-groovy based `GremlinPlugin` (hadoop, giraph, gephi, etc) and developing a parallel plugin based on gremlin-core.
** Making it so that gremlin server/console could utilize either plugin type (gremlin-server does this through a separate configuration option in gremlin-server.yaml and the console does this through a system variable that you could add to gremlin.sh (`-Dplugins=v3d3`)
** Deprecating lots of odds and ends to allow me to move classes around. The deprecation approach was more tedious and leads to some dead code, but it was a nice way to evolve this change more slowly and intelligently. It also taught me that we should be more watchful of "public class" scope. If just a fraction of these classes that I deprecated (which no one probably uses anyway) had been package scope this work would have been much easier.

The testing for this body of work involved:

* `mvn clean install` with integration tests
* `docker/build.sh -t -i -n`
* `process-docs.sh`
* manual testing:
** every plugin in both versions 
** specific testing of the dependency grabber via `:install` for the console and `-i` in gremlin server

For the most part I didn't provide a lot of docs for this because I don't think we need to publish this for 3.2.x. These changes essentially just set the stage for use to use them in full in 3.3.0.  I'll do some separate smaller PRs for 3.3.0 work around this that will include doc updates as I remove the deprecated code I've created here.

VOTE +1


spmallette:
@okram one thing that could use your explict review here - we lost the ability to do wildcard imports given the work you did on `CoreImports` for gremlin-python. Some plugins used a lot of wildcards, specifically the hadoop plugin:

https://github.com/apache/tinkerpop/blob/90f427c8dc11636a5884e6cca3c0a22249d26576/hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/jsr223/HadoopGremlinPlugin.java

I didn't go through hadoop APIs to add all the classes that were wildcarded. can you please give a review to see if anything else should be added there?

hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/jsr223/HadoopGremlinPlugin.java
```diff
@@ -0,0 +1,153 @@
...
+                            VertexWritable.class,
+                            GraphSONInputFormat.class,
+                            GraphSONOutputFormat.class,
+                            GraphSONRecordReader.class,
+                            GraphSONRecordWriter.class,
+                            GryoInputFormat.class,
+                            GryoOutputFormat.class,
+                            GryoRecordReader.class,
+                            GryoRecordWriter.class,
+                            ScriptInputFormat.class,
```
okram: We also need `ScriptOutputFormat.class`. And for safety, `ScriptRecordReader` and `ScriptRecordWriter`.

spmallette: done

dkuppitz:
`docker/build.sh -t -i -n` succeeded. Haven't done a code review, no clue where to start and what to look for in 144 changed files. Anyway, Docker is happy and so am I.

VOTE: +1

okram:
VOTE +1.

PR #511, opened 2016-12-02
TINKERPOP-1539 Create a ComplexTraversalTest with crazy nested gnarly traversals.

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1539

Added 2 more tests, one for OLTP only and one for OLTP+OLAP. More tests should not be added as part of this PR; we can add more later whenever a good traversal appears on the mailing list or somewhere else in the real world.

VOTE: +1


okram:
This is great. We should really be diligent to keep growing the tests here. As our strategies get more complex (and are working in recursive fashions) we really need to have complex, nasty nested traversal test cases.

I think that growing `ComplexTest.Traversals` should be CTR based to make it easy/convenient to expand as much as possible.

VOTE +1.

spmallette:
VOTE +1

PR #513, opened 2016-12-08
TINKERPOP-1582 - Encode Bytecode as a dict, not a string

bjmoor:
https://issues.apache.org/jira/browse/TINKERPOP-1582

Remote connection driver was encoding the Gremlin Bytecode as a json string, rather than as a dict.  This caused the TraversalOpProcessor to not use the gremlin-server configured deserializers, but rather a non-configurable statically declared serializer to decode the Bytecode string.


spmallette:
Thanks - tests all pass. Merged via CTR

PR #515, opened 2016-12-09
TINKERPOP-1581 Gremlin-Python driver connection is not thread safe.

davebshow:
https://issues.apache.org/jira/browse/TINKERPOP-1581

This was a simple fix. The `DriverRemoteConnection.__init__` method no longer ignores the `loop` kwarg passed by user. I added two tests using threads as well. One demonstrates how not passing an `IOLoop`, which was in effect the behavior before this PR, will cause a `RuntimeError`. The other shows how you can now pass an `IOLoop` for each thread to avoid this problem.

VOTE +1


spmallette:
After the recent additional fix from @davebshow this seems to now pass consistently. I've done a number of test runs with docker and my local environments and all is good.

VOTE +1

okram:
VOTE +1

PR #517, opened 2016-12-19
Fix numbers deserialization for GraphSON2.

newkek:
Issue:
There was a small issue with `ObjectMapper` that was preventing one from doing `objectMapper.readValue(jsonString, Integer.class/Double.class)`. Turns out the issue come from the fact that Jackson normally doesn't uses the `GraphSONTypeResolverBuilder` hook I've introduced for the typing system - and calls directly the default deserializers - for numbers values like `Integer` or `Double`, which led the default deserializers to fail since they are facing a Typed object (`{"@type":..., "@value":....}`) and not simply the value.

Fix: 
Adding simple deserializers for `Integer` and `Double` causes Jackson to use the hook for the typing system and hence deserialization works normally.


spmallette:
Merged via CTR to tp32 and master

PR #518, opened 2016-12-21
Honor the equals() contract on Property impls by allowing nulls as valid comparisons.

metlos:
If user code put the properties into some kind of collection together with `null` values, things would blow up.

Note that `ElementHelper.areEqual(Element, Object)` allows for `null` values, but `ElementHelper.areEqual(Property, Object)` does not. I assume the reason for this discrepancy is that properties are never supposed to be null but merely not present.

While this assumption is valid in Tinkerpop impl, I think it cannot be imposed on the user code, which assumes a valid implementation of `equals()` which *should not* blow up on null arguments.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/detached/DetachedProperty.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/detached/DetachedProperty.java
@@ -96,7 +96,7 @@ public class DetachedProperty<V> implements Property<V>, Serializable, Attachabl
     @SuppressWarnings("EqualsWhichDoesntCheckParameterClass")
     @Override
     public boolean equals(final Object object) {
-        return ElementHelper.areEqual(this, object);
+        return object != null && ElementHelper.areEqual(this, object);
     }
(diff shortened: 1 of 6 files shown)


spmallette:
hmmm - i'm not sure why `ElementHelper.areEqual(Property, Object)` needs to throw exceptions for null situations. i'm not so sure that's really how it should work. In studying some of the usage, I don't see where that's really valuable logic. I think the better fix is to just to make `ElementHelper.areEqual(Property, Object)` handle nulls properly like the other `areEqual()` methods.  @okram any problems taking that approach?

okram:
Yea, I would put the null check into `ElementHelper.areEqual()` so all providers have the same semantics. Then I would add a structure test to verifies that null is okay for `equals()` for every element type: vertex, edge, property, vertex property.

spmallette:
Thanks @okram 

@metlos please also target the tp32 branch instead of master.

metlos:
Thanks for the feedback. I'm closing this PR then and will target the reworked patch to tp32.

metlos:
Created #519 with the updates as suggested here.

PR #519, opened 2016-12-21
Make ElementHelper.areEqual(Property, Object) handle nulls…

metlos:
… so that it can be used correctly in equals() methods of Property impls.

Added test methods for additional equality "scenarios" in ElementHelper.

Note that it is not possible to add tests for equals on `Vertex`, `Edge`, `Property` and `VertexProperty` directly as they are interfaces, not concrete classes. Instead I've added additional test methods for `ElementHelper` to test various case of properties and elements. `ElementHelper.areEqual()` methods are suggested to be used for `equals` of implementations of those interfaces.


spmallette:
Thank you - this has been merged via CTR as it was a pretty straightforward change

PR #520, opened 2017-01-03
TINKERPOP-1130 IO Testing

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1130

Added `gremlin-io-test` module under the `gremlin-tools` module. It provides more complete testing of IO related capabilities of GraphSON and Gryo. The testing helps with maintaining compliance of different versions of GraphSON and Gryo. Stubbed in Gryo 3.0.

This change also allows for slightly more automated generation of the `dev/io` docs for GraphSON. If you do: `mvn clean install -Dio` it will generate the asciidoc for the GraphSON sections to `target/dev-docs`. From the content of those files can be cut/paste to the asciidoc files in `dev/io`. That directory also holds a `model.csv` file which has data about the various version compatibilities. Having it in that format makes it pretty easy to see formats are compatible with what versions of specific releases.

VOTE +1


okram:
First off, thank you for doing this -- what an undertaking. Getting I/O solid for 3.3.0 is crucial given the versioning push and the desired serialization changes for GLVs. Before officially voting, can you answer the following questions?

1. Does this verify that we don't accidentally create a breaking IO format change in the same IO version?

2. Does this verify that GraphSON and Gryo (and anything else down the line) have all the same serializers? (that is, are capable of handling all the same data -- e.g. we won't miss that `Pick.none` was not registered in 3.2.1 for GraphSON but it was for Gryo?)

spmallette:
> Does this verify that we don't accidentally create a breaking IO format change in the same IO version?

Yes. We generate GraphSON and Gryo files that are statically checked into source control on each release. Then we make sure that those files can still be read with the current version. For example, we currently are verifying that 3.3.0-SNAPSHOT can read Gryo 1.0 generated in 3.3.0, 3.2.4 and 3.2.3. 

In doing this, I've already encountered inconsistencies which are documented in the `Model` class which defines all the incompatibilities. So, if we do have a break (or something untestable), we can at least document it there. For example, there were a number of bugs in Gryo 1.0 that generated "bad" static files to test against. I made the fixes for 3.2.4, so we can really only test against versions 3.2.4 and later:

https://github.com/apache/tinkerpop/blob/TINKERPOP-1130/gremlin-tools/gremlin-io-test/src/main/java/org/apache/tinkerpop/gremlin/structure/io/Model.java#L100

Note that the incompatibilities in the line referenced above also include "untyped" GraphSON there as untyped GraphSON won't deserialize to specific types either (typically just all coerces to `Map` and `List` without a way to deserialize back to the original type). 

Actually, we still have some todos on this PR because of the changes we need to revert in the GraphSON 2.0 work you had done a while back on properties (those changes will be for GraphSON 3.0). That was a breaking change and the tests are failing so I think it's doing its job.

> Does this verify that GraphSON and Gryo (and anything else down the line) have all the same serializers? (that is, are capable of handling all the same data -- e.g. we won't miss that Pick.none was not registered in 3.2.1 for GraphSON but it was for Gryo?)

Well - I suppose it helps with this issue, but it's not as though it is actively monitoring our code to do that obviously. In other words, it will help if by convention we do all serialization testing through `gremlin-io-test`. If we do that, then we register a type with the `Model` and it will immediately force testing of both Gryo and GraphSON. If you didn't implement a serializer for one or the other then its going to fail.  If you just decide to add a new class registration to Gryo and don't bother to write a test for it in `gremlin-io-test` then obviously it's not going to have any knowledge of the new registration. 

I think that does lead to some issues which I'd brought up on the dev mailing list in the last couple of weeks. Right now there is some disparity between the types supported by GraphSON and those supported by Gryo. Right now Gryo has a lot more supported classes (many of which are not specifically tested under `gremlin-io-test) - I only covered those classes supported by GraphSON so far. I think that we will want Gryo and GraphSON to support the same types (for 3.0 of both formats) which may mean shrinking the types that Gryo supports (I would think we would want the smallest type library possible), but I'm not sure how practical that is. A lot of the Gryo registrations were added to support OLAP stuff and remote traversal stuff (before we had bytecode). I don't know how easy it is to just drop some of that stuff or which ones are even candidates to be dropped.

okram:
VOTE +1

dkuppitz:

VOTE: +1

PR #521, opened 2017-01-05

spmallette:

Made it so that the `CloseableIterator` is closed by `GraphStep` if it is provided by the iterator supplier. Furthermore, any steps in a `Traversal` implement `AutoCloseable` then its `close()` method will be called. In this way `Traversal.close()`, the common API a user will work with, has a way to release resources in `Graph` implementations that require it.

Builds with `docker/build.sh -t - i -n`.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Traversal.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Traversal.java
@@ -259,8 +259,9 @@ public interface Traversal<S, E> extends Iterator<E>, Serializable, Cloneable, A
      */
     @Override
     public default void close() throws Exception {
-        for (AutoCloseable closeableStep : TraversalHelper.getStepsOfAssignableClassRecursively(AutoCloseable.class, asAdmin())) {
-            closeableStep.close();
+        for(final Step<?,?> step : this.asAdmin().getSteps()) {
+            if(step instanceof AutoCloseable)
+                ((AutoCloseable) step).close();
         }
     }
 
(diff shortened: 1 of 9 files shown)


okram:
Nice. Clean backwards compatible `GraphStep.close()` method which uses `instanceof`.

Are you looking for a VOTE now or are you still building on this PR? -- e.g. integrating it with `Traversal.close()`/etc.

spmallette:
I think I'm done - It's already implemented on `Traversal.close()`. I think it's good for vote. Don't think i'm missing anything.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Traversal.java
```diff
@@ -254,9 +254,14 @@ public default void forEachRemaining(final Consumer<? super E> action) {
         }
     }
 
+    /**
+     * Releases resources opened in any steps that implement {@link AutoCloseable}.
+     */
     @Override
     public default void close() throws Exception {
```
okram: This is bad. `Steps` that are `TraversalParents` are responsible for propagating operations on them to those below. I would do the following:

```
for(final Step<?,?> step : this.getSteps()) {
  if(step instanceof AutoCloseable)
    step.close();
}
```

Next, I would have `TraversalParent`  extend `AutoCloseable` with its default `close()` method being:

```
for(final Traversal.Admin<?,?> traversal : this.getLocalChildren()) {
  traversal.close()
}

for(final Traversal.Admin<?,? traversal: this.getGlobalChildren()) {
  traversal.close();
}
```

spmallette: changes are made now - thanks - didn't know that stuff.

okram:
VOTE +1

lvca:
VOTE +1

twilmes:
VOTE: +1

PR #523, opened 2017-01-05
Add ResponseMessage deserializer for GraphSON2.

newkek:
Motivation:

`ResponseMessage` used to be deserialized manually via a method on GraphSONMessageSerializerV2d0 (`deserializeResponse()`). Which prevented from using the `Mapper` directly to deserialize a response.

Modification:

Deserialization of `ResponseMessage` is done in a deserializer registered by the `GremlinServerModule` and the GraphSONMessageSerializerV2d0 has been changed accordingly. Also added tests of ser/de for `RequestMessage` and `ResponseMessage` from the `ObjectMapper`.

Result:

Calls like `objectMapper.readValue(jsonString, ResponseMessage.class)` work now.


spmallette:
Thanks for this. I tested this little change against tp32 and master and it worked nicely. It also enabled me to remove testing restrictions on gremlin-io-test around `ResponseMessage`

https://github.com/apache/tinkerpop/commit/0819a686a62e2f9ca413d7339460b3b267ea067d

Merged via CTR.

PR #524, opened 2017-01-06
TINKERPOP-1585 & TINKERPOP-1590: DedupGlobalStep and Distributed TinkerMemory

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1585
https://issues.apache.org/jira/browse/TINKERPOP-1590

This was a week long project with @dkuppitz. In short, we made `DedupGlobalStep` a bit better. It had a bug where it was deduping (in OLAP) twice at the master traversals. That is no longer and issue. Moreover, I added an explicit barrier data structure to reduce the amount of data movement between the master barrier and the master step. When working on this, I originally thought the bug was in `TinkerMemory` (`TinkerGraphComputer`) where there is lock contention for barrier data. There is! and I was able to seep up a toy traversal we had from 7 seconds to 2.9 seconds cause of it by having each `TinkerWorker` maintain its own memory data structure that is then merged into the master `TinkerMemory` when the iteration is complete. Standard, we just never thought to do this. Lots of other nick nack goodies...

```
* Fixed an optimization bug in OLAP-based `DedupGlobalStep` where deduping occurred twice.
* `MemoryComputeKey` now implements `Cloneable` which is useful for `BiOperator` reducers that maintain thread-unsafe state.
* `TinkerGraphComputer` now supports distributed `Memory` with lock-free partition aggregation.
```

VOTE +1.


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/DedupGlobalStep.java
```diff
@@ -89,6 +92,17 @@ public ElementRequirement getMaxRequirement() {
 
     @Override
     protected Traverser.Admin<S> processNextStart() {
+        if (null != this.barrier) {
+            this.barrierIterator = this.barrier.entrySet().iterator();
+            this.barrier = null;
+        }
+        while (this.barrierIterator != null && this.barrierIterator.hasNext()) {
+            if (null == this.barrierIterator)
```
dkuppitz: `this.barrierIterator` can never be null within within the `while()` loop. Unless I overlooked something fundamental, `processNextStart` can be simplified to:

```
protected Traverser.Admin<S> processNextStart() {
    if (null != this.barrier) {
        this.barrierIterator = this.barrier.entrySet().iterator();
        this.barrier = null;
        while (this.barrierIterator.hasNext()) {
            final Map.Entry<Object, Traverser.Admin<S>> entry = this.barrierIterator.next();
            if (this.duplicateSet.add(entry.getKey()))
                return PathProcessor.processTraverserPathLabels(entry.getValue(), this.keepLabels);
        }
    }
    return PathProcessor.processTraverserPathLabels(super.processNextStart(), this.keepLabels);
}
```

spark-gremlin/src/test/java/org/apache/tinkerpop/gremlin/spark/process/computer/traversal/strategy/optimization/SparkInterceptorStrategyTest.java
```diff
@@ -142,7 +142,7 @@ public void shouldSuccessfullyEvaluateInterceptedTraversals() throws Exception {
         test(6l, g.V().out().values("name").count());
         test(2l, g.V().out("knows").values("name").count());
         test(3l, g.V().in().has("name", "marko").count());
-        test(6l, g.V().repeat(__.dedup()).times(2).count());
+        test(0l, g.V().repeat(__.dedup()).times(2).count());
```
dkuppitz: Why did that change? `0` as the expected result looks kinda wrong.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/DedupGlobalStep.java
okram: Your code fails. Did you run it? ... The problem is that `this.barrier` is null'd but the `barrierIterator` still exists and you still need to fetch results from it.

spark-gremlin/src/test/java/org/apache/tinkerpop/gremlin/spark/process/computer/traversal/strategy/optimization/SparkInterceptorStrategyTest.java
okram: This is crazy. The right answer is `0`. Think about it. It goes through once, fine, all 6 vertices. The second time -- already seem them! Filter. Thus, 0.

However, I just checked OLTP and it assumes 6. Wondering what is "right" ?. .... 

```
gremlin> g.V().repeat(dedup()).times(2)
==>v[1]
==>v[2]
==>v[3]
==>v[4]
==>v[5]
==>v[6]
gremlin> g.withComputer().V().repeat(dedup()).times(2)
gremlin>
```

Whatever, is decided as the correct answer, we should definitely put this into `RepeatTest`. I just randomly had this query in a Spark test.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/DedupGlobalStep.java
dkuppitz: I didn't run it. Just thought we can get rid of the multiple `null` checks.

spark-gremlin/src/test/java/org/apache/tinkerpop/gremlin/spark/process/computer/traversal/strategy/optimization/SparkInterceptorStrategyTest.java
okram: As discussed in IM. This is actually a bug in `RepeatUnrollStrategy` that was introduced in 3.2, but doesn't exist in 3.1. Added a test case to `DedupTest` and both OLTP and OLAP now behave as expected.

dkuppitz:

VOTE: +1

twilmes:
Nice set of updates.

`docker/build.sh -t -i` success

VOTE: +1

PR #528, opened 2017-01-09
TINKERPOP-1116: Some anonymous traversal steps can be hard typed.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1116

Tweaked the typing of various `__` anonymous traversal steps as various do not need generic typing. This is in master/ as its breaking because user that do `<Vertex>out()` will simply need to do `out()` (and thus, remove the explicit typing).

VOTE +1.


dkuppitz:
VOTE: +1

spmallette:
VOTE +1

PR #529, opened 2017-01-10
TINKERPOP-1549 Implement skip()

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1549

Added `skip(long)` and `skip((Scope,long)` which call the `range(low,high)` equivalents with -1 as the high.

Also added test cases and fixed the OLAP implementation (which probably never worked before).

Note, that I implemented `equals()` for `RangeGlobalStep`, because hash code collisions are very likely. For the same reason I also removed the hash code inequality check from `StepTest`.

VOTE: +1

### DIFF
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -1916,6 +1916,35 @@ g.V(1).both().both().simplePath()
 g.V(1).both().both().simplePath().path()
 ----
 
+[[skip-step]]
+Skip Step
+~~~~~~~~~
+
+The `skip()`-step is analogous to <<range-step,`range()`-step>> save that the higher end range is set to -1.
+
+[gremlin-groovy,modern]
+----
+g.V().skip(2)
+g.V().range(2, -1)
+g.V().skip(2).toString()
+----
+
+The `skip()`-step can also be applied with `Scope.local`, in which case it operates on the incoming collection.
+
+[gremlin-groovy,modern]
+----
+g.V().hasLabel('person').filter(outE('created')).as('p'). <1>
+  map(out('created').values('name').fold()).
+  project('person','primary','other').
+    by(select('p').by('name')).
+    by(limit(local, 1)). <2>
+    by(skip(local, 1)) <3>
+----
+
+<1> For each person who created something...
+<2> ...select the first project (random order) as `primary` and...
+<3> ...select all other projects as `other`.
+
 [[store-step]]
 Store Step
 ~~~~~~~~~~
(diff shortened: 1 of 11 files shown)


gremlin-groovy-test/src/main/groovy/org/apache/tinkerpop/gremlin/process/traversal/step/filter/GroovyRangeTest.groovy
```diff
@@ -108,5 +108,15 @@ public abstract class GroovyRangeTest {
         public Traversal<Vertex, Map<String, String>> get_g_V_asXaX_out_asXbX_out_asXcX_selectXa_b_cX_byXnameX_rangeXlocal_1_2X() {
             new ScriptTraversal<>(g, "gremlin-groovy", "g.V.as('a').out.as('b').out.as('c').select('a','b','c').by('name').range(local,1,2)")
         }
+
+        @Override
+        public Traversal<Vertex, Long> get_g_V_skipX2X_count() {
+            new ScriptTraversal<>(g, "gremlin-groovy", "g.V.skip(2).count()")
```
okram: I would replace this test with one that is an `order().by(...).skip(2)`-style test so you know its actually skipping the "first 2" and not just giving you two results back. Likewise for your `skip(local)`-test.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -1916,6 +1916,35 @@ g.V(1).both().both().simplePath()
...
+Skip Step
+~~~~~~~~~
+
+The `skip()`-step is analogous to <<range-step,`range()`-step>> save that the higher end range is set to -1.
+
+[gremlin-groovy,modern]
+----
+g.V().skip(2)
+g.V().range(2, -1)
+g.V().skip(2).toString()
```
okram: We should not advertise `toString()` as other GLV's don't show compilation information and in fact, down the line, we will most likely change what `toString()` shows in Gremlin-Java/Groovy.

I would add an example that uses `order().by(...).skip(x)` so that people know it skips the first `x` results. For instance, show without `skip(x)`, then show with `skip(x)` after an `order()`.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/__.java
```diff
@@ -810,6 +810,20 @@ protected __() {
     }
 
     /**
+     * @see GraphTraversal#skip(long)
+     */
+    public static <A> GraphTraversal<A, A> skip(final long n) {
```
okram: We call the variable `final long limit` for `limit()`. Perhaps call `n` `skip` so it has the same look and feel. `skip(final long skip)`. Likewise for your `Scope` version.

okram:
VOTE +1.

spmallette:

VOTE +1

PR #531, opened 2017-01-10
Added more recipes

spmallette:
There's no JIRA ticket for this. It's just a random bunch of additions to the recipes documentation. Note that in this update i added an "Appendix" which simply contains some "complex" or otherwise "interesting" traversals that don't seem to fit a pattern/recipe, but remain potentially useful for learning.

I've published the docs here for review:

http://tinkerpop.apache.org/docs/3.2.4-SNAPSHOT/recipes/

VOTE +1


docs/src/recipes/appendix.asciidoc
```diff
@@ -0,0 +1,226 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+////
+Appendix
+========
+
+Many of the recipes are based on questions and answers provided on the gremlin-users mailing list. This section
```
okram: You should `href` to gremlin-users mailing list. https://groups.google.com/forum/#!forum/gremlin-users

docs/src/recipes/appendix.asciidoc
```diff
@@ -0,0 +1,226 @@
...
+----
+g.addV('name','marko').as('marko').
+  addV('name','josh').as('josh').
+  addV('name','daniel').as('daniel').
+  addV('name','matthias').as('matthias').
+  addE('follows').from('josh').to('marko').
+  addE('follows').from('matthias').to('josh').
+  addE('follows').from('daniel').to('josh').
+  addE('follows').from('daniel').to('marko').iterate()
+g.V().as('p').
```
okram: You could also show another version.

```
g.V().group().
        by('name').
        by(project('numFollowers','followers').
             by(__.in('follows').count()).
             by(__.in('follows').values('name').fold())).next()
```

or yet another version....

```
g.V().group().
        by('name').
        by(__.in('follows').values('name').fold().
            project('numFollowers','followers').
              by(count(local)).
              by()).next()
```

docs/src/recipes/appendix.asciidoc
```diff
@@ -0,0 +1,226 @@
...
+         group().
+           by(select(keys)).
+           by(select(values).fold())).
+    unfold().
+    map(union(select(keys).project("a").by(), select(values)).
+    unfold().
+    group().
+      by(select(keys).unfold()).
+      by(select(values).unfold().unfold().fold()))
+----
```
okram: I'm wondering if some of these traversals may "look better" with `match()`. @dkuppitz ?

docs/src/recipes/connected-components.asciidoc
```diff
@@ -0,0 +1,82 @@
...
+is happening at each step. Consider running this example one line at a time (or perhaps even in a step at a time) to
+see the output at each point.
+
+While the above approach returns results nicely, the traversal doesn't appear to work with OLAP. A less efficient
+approach, but one more suited for OLAP execution looks quite similar but does not use `dedup` as heavily (thus
+`GraphComputer` is forced to analyze far more paths):
+
+[gremlin-groovy,existing]
+----
+g.withComputer().V().emit(cyclicPath().or().not(both())).repeat(both()).until(cyclicPath()).
```
okram: I ran your "OLTP" traversal `withComputer()` and it works as expected.. ?

```
gremlin> g.withComputer().V().emit(cyclicPath().or().not(both())).repeat(both()).until(cyclicPath()).
......1>   aggregate("p").by(path()).cap("p").
......2>   unfold().limit(local, 1).dedup().
......3>   map(__.as("v").select("p").unfold().
......4>          filter(unfold().where(eq("v"))).
......5>          unfold().dedup().order().by(id).fold()
......6>   ).dedup()
==>[v[A],v[B],v[C]]
==>[v[D],v[E]]
==>[v[F]]
```

docs/src/recipes/duplicate-edge.asciidoc
```diff
@@ -0,0 +1,142 @@
...
+<3> Unroll the key-value pairs in the `Map` of paths-edges.
+<4> Only the values from the `Map` are needed and as mentioned earlier, those lists with more than one edge would
+containa  duplicate.
+
+This method find the duplicates, but does require more memory than other approaches. A slightly more complex approach
+that uses less memory might look like this:
+
+[gremlin-groovy,existing]
+----
+g.V().as("ov").
```
okram: Here is the `match()`-form.

```
g.V().match(
    __.as("ov").outE().as("e"),
    __.as("e").inV().as("iv"),
    __.as("iv").inE().as("ie"),
    __.as("ie").outV().as("ov")).
      where("ie",neq("e")).
      where("ie",eq("e")).by(label).
    select("ie").
    group().
      by(select("ov","e","iv").by().by(label)).     
    unfold().select(values).
      where(count(local).is(gt(1)))
 ```

File: docs/src/recipes/appendix.asciidoc (hunk 3 above)
Comment (dkuppitz) [inline docs/src/recipes/appendix.asciidoc]: Do we want to look good or perform good? You may have noticed it on the mailing list, I always try to prevent the use of `match()`. There were only a few questions that I ever answered with a match query and that was only for the sake of readability.

File: docs/src/recipes/connected-components.asciidoc (hunk 4 above)
Comment (dkuppitz) [inline docs/src/recipes/connected-components.asciidoc]: I'm pretty sure that's not working in `tp32` (OLAP).

File: docs/src/recipes/duplicate-edge.asciidoc (hunk 5 above)
Comment (dkuppitz) [inline docs/src/recipes/duplicate-edge.asciidoc]: Go home.........

File: docs/src/recipes/appendix.asciidoc (hunk 1 above)
Comment (spmallette) [inline docs/src/recipes/appendix.asciidoc]: I debated on whether to do that or not with recipes in general but given that I usually adapt the traversals to a narrative, alter the data or improve upon them from a performance perspective or usability perspective linking back doesn't always make sense, so rather than do it some of the times, I opted to not do it at all. The "appendix" might be different though....I don't recall if I made any modifications to those or just "copied" them writing my own explanation for what they were doing as a "title". Maybe adding the links for those would be useful in some way....let me think on that.

**Comment by spmallette** [comment]
I'm not so sure how good I am at Gremlin - a good portion of this stuff comes from @dkuppitz answers :smile: - but I'm better at Gremlin for sure after writing these recipes as I'm forced to really analyze and question what's going on in the traversals so that I can write a proper narrative. Of course, it still is a fair bit of work to write all this up, organize it into patterns, sometimes re-write traversals or data to better fit the narrative, etc. so I won't argue about that.

File: docs/src/recipes/appendix.asciidoc (hunk 1 above)
Comment (dkuppitz) [inline docs/src/recipes/appendix.asciidoc]: I believe what @okram meant is just one link, like:

> Many of the recipes are based on questions and answers provided on the [gremlin-users mailing list](https://groups.google.com/forum/#!forum/gremlin-users). This section...

Not a link to every discussion.

File: docs/src/recipes/appendix.asciidoc (hunk 3 above)
Comment (spmallette) [inline docs/src/recipes/appendix.asciidoc]: I wonder if our lack of and extensive set of  `match()` examples continues to create the sense that we don't have a declarative story for Gremlin.

File: docs/src/recipes/appendix.asciidoc (hunk 1 above)
Comment (spmallette) [inline docs/src/recipes/appendix.asciidoc]: i misread - got it

File: docs/src/recipes/appendix.asciidoc (hunk 2 above)
Comment (dkuppitz) [inline docs/src/recipes/appendix.asciidoc]: `group()` is actually not needed here. `p` will be unique in the result, hence it's just a projection:

```
g.V().as('p').map(__.in('follows').values('name').fold()).
  project('person','followers','numFollowers').by(select('p')).by('name')).by().by(count(local))
```

File: docs/src/recipes/appendix.asciidoc [hunk 6]
```diff
@@ -0,0 +1,226 @@
...
+  out("allocated").
+  where(eq("t")).
+  group().
+    by(select("s").by("name")).
+    by(group().by(select("t").by(valueMap("fromTime","toTime"))).
+               by(select("c").dedup().values("name").fold())).next()
+----
+
+[[appendix-d]]
+_In the "modern" graph, with a duplicate edge added, find the vertex pairs that have more than one edge between them._
```
dkuppitz: I think it's important to note that this is only for edge that point into the same direction (`OUT` in the example). The solution for undirected edges is pretty complicated (if we take into account that ids are not necessarily sortable).

```
g.V().as("a").both().as("b").
  groupCount().by(select("a","b")).unfold().
  filter(select(values).is(gt(1))).
  select(keys).not(__.as("x").select("done").unfold().as("y").
                   select("x","y").by(select("a")).by(select("b")).
                   where("x", eq("y"))).store("done")
```

dkuppitz: Another way for undirected edges:

```
gremlin> g.V().where(without("x")).as("a").
......1>   outE().as("e").inV().as("b").
......2>   filter(bothE().where(neq("e")).otherV().where(eq("a"))).store("x").
......3>   select("a","b").dedup()
==>[a:v[1],b:v[3]]
```

@okram I thought the following query should have the same result, but it doesn't (result is empty):

```
g.V().where(without("x")).as("a").
  outE().as("e").inV().as("b").
  where(bothE().where(neq("e")).otherV().as("a")).store("x").
  select("a","b").dedup()
```

The difference is `filter(...where(eq("a")))` (works) vs. `where(....as("a"))` (doesn't work).

spmallette:
Thanks for all the feedback. I've pushed some changes on this branch given the comments I've seen. I think I got all of it. Good to vote now?

dkuppitz:
I went through all the code snippets again, optimized a few math examples and I think it's all looking good now.

VOTE: +1

okram:
VOTE +1.

PR #533, opened 2017-01-16
TINKERPOP-1600 Added base64 encoded string to sasl challenge

spmallette:
This is a small change but I was hoping for some tests by driver providers before merging as this change messes with the authentication scheme a bit. It should be a backward compatible change, but I just wanted to be sure I didn't break anything before merging this.

VOTE +1

PR #534, opened 2017-01-16
TINKERPOP-1566 Kerberos authentication for gremlin-server

vtslab:
This PR includes three items (as stated in the changelog):
  1 Added Kerberos authentication to `gremlin-server` for websockets and nio transport.
  2 Added audit logging of authenticated users and of gremlin queries to `gremlin-server`.
  3 Fixed `gremlin-driver`'s support for string results regarding returned byte arrays 
    from `Sasl` authentication.

Regarding item 1, I did not attempt to provide Kerberos authentication for http 
transport, as I assumed http will not be very popular anymore, now that the GLV's
are available for accessing graphs via gremlin-server.

Item 2, audit logging, naturally belongs to Kerberos authentication. Kerberos is
important in providing access to confidential data, that is, being sure of
someone's identity without having him logging in for each service. Some 
confidential data, like personal data, often have legal obligations regarding 
logging of their access: that is what item 2 provides.

Item 3 is just a minor issue that surfaced during test development of Kerberos 
authentication.

An ample number of integration tests is provided. In addition, I did manual tests
in a working freeIPA Kerberos environment to verify the proper working. 

Reviewers wanting a short reminder of Kerberos authentication are referred to:
http://www.roguelynn.com/words/explain-like-im-5-kerberos/
[It learnt me a lot, I am not trying to be arrogant :-)]

The main design choices I made are:

i) Krb5Authenticator does not refer to policy servers or storage backends for 
authorization, but rather assumes that any user who can be authenticated using
Kerberos, is also authorized to access the service. Others could extend on this.

ii) The JAAS entry for Krb5Authenticator was not made configurable, apart from 
the principal name and keytab location to be provided in the yaml file. Using 
a separate JAAS config file would primarily introduce more flexibility in getting 
the config wrong. This choice is in line with the current situation with all
authenticator configuration in gremlin-server's yaml file. But the choice is 
not consistent with other Apache projects like Hadoop and HBase.

iii) Audit logging was made into a general feature that also works for other 
authenticators. It has to be explicitly enabled, though, with a property in 
gremlin-server's yaml file, because the audit logs can contain confidential data.

iv) Audit logging was given a separate logger apart from the org.apache.tinkerpop
naming tree, so that its level can be set to INFO without influencing level
settings of the normal logging. The logger name, "audit.org.apache.tinkerpop.gremlin.server",
was defined in GremlinServer, for lack of a better location.

v) Apache Kerby was used as the Kerberos Key Distribution Center (KDC) for the 
Kerberos integration tests, because it also belongs to Apache and proved easy to 
use. The project is still in RC2 status, though, but is only a test dependency.

Finally, running integration test on gremlin-server still results in two errors,
probably due to the presence of the gremlinjaas.conf file in the test resources.
I did not correct these, because I was not sure whether it could be due to my
test environment.

Failed tests: 
  GremlinServerAuthIntegrateTest.shouldFailAuthenticateWithPlainTextNoCredentials:130 
expected: class org.apache.tinkerpop.gremlin.driver.exception.ResponseException
but was: class org.ietf.jgss.GSSException
  GremlinServerAuthOldIntegrateTest.shouldFailAuthenticateWithPlainTextNoCredentials:133 
expected: class org.apache.tinkerpop.gremlin.driver.exception.ResponseException
but was: class org.ietf.jgss.GSSException


spmallette:
Thanks for this - Could you please update the name of the pull request to be: "TINKERPOP-1566 Kerberos" - basically we just need the JIRA identifier in the prefix of the name so that it hooks up into the JIRA automation.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ser/AbstractGryoMessageSerializerV1d0.java
```diff
@@ -301,9 +301,12 @@ private Object serializeResultToString(final ResponseMessage msg) {
         if (msg.getResult().getData() == null) return "null";
 
         // the IteratorHandler should return a collection so keep it as such
+        // Sasl authentication needs byte[] to pass unchanged
         final Object o = msg.getResult().getData();
         if (o instanceof Collection) {
             return ((Collection) o).stream().map(d -> null == d ? "null" : d.toString()).collect(Collectors.toList());
+        } else if (o instanceof byte[]) {
```
spmallette: I think you should undo this special handling - based on my explanation on #533

spmallette:
I see that you renamed the title, but it's not exactly what I had. If you don't prefix the title with the actual JIRA identifier (which is all caps and has the dash) the automation won't work.

spmallette:
In my testing of this PR with the latest on master (with #533 merged) that there was a test failure in `GremlinServerAuthOldIntegrateTest` -  It didn't occur without your changes so I'm not sure what might have gone wrong. It had something to do with an expected exception being different from the actual exception. When you rebase this I can take another look at what's going on.

vtslab:
Yes, I mentioned these failing tests in the PR text above. I suspect that the java security libs just pick up the krb5.conf file from the test resources "without asking". It means the test fails faster and tries to find credentials elsewhere when not provided. To resolve it, we either have to adapt the test or we have to try to configure Kerberos without a krb5.conf file (if that is the cause). Either way, it does not file nice if test outcomes depend on unnecessary components being on the classpath or not, so maybe we should look for a third way.

PR title should be OK now.

spmallette:
Sorry - I'd forgotten you'd mentioned the broken tests. Ok - we'll figure out what to do there once you rebase. Thanks for renaming the PR.

spmallette:
I'm not sure if TINKERPOP-1600 fixed it or if you did something in rebase, but the two failing tests dont' seem to be a problem anymore when I run the tests in intellij. As a result, I'm running the full integration tests now to make sure things are solid. After that I'll give a more detailed review of the code and hopefully we can get this merged soon.

spmallette:
Well - looks like they are still broken. My favorite situation - they fail with maven but pass in intellij. I suppose that has something to do with your explanation regarding the classpath.

vtslab:
Hi, I am working on the two failing integration tests. It can be fixed in three ways:
 1. just hide the symptoms and also allow a GSSException for a test that should fail anyway:   ugly!
 2. configure <useForks>false for the failsafe plugin: fine, but may impact test performance
 3. adapt the gremlin-driver handler code, which now chooses GSSAPI if no credentials are supplied. This should be a test on an available JaasEntry.

Which of the ones do you like to see committed on the PR?

spmallette:
it sounds like the third option comes without downsides (you had detracting statements on the other two) so I guess I would go with that. While i haven't thought it all through, this behavior:

> now chooses GSSAPI if no credentials are supplied

sounds like it might not be a sensible default either, so changing that sounds like the right thing to do.

vtslab:
OK, thanks, I'll see to it.  The downside of the third option was implicit: it changes the driver code which is in production everywhere. But that´s why we do this work in the 3.3.x line and I think it will be an imrpovement.

mike-tr-adamson:
I'm concerned about changing the `getMechanism` logic based on the failing tests. The tests can legitimately fail both ways. It would be valid to fix those tests so that they can fail on either a `ResponseException` or a `GSSException`. As has already been said the difference in error being whether the test finds any kerberos configuration that it can use.

The tests were failing on my machine for exactly the same reason - I had an `/etc/krb5.conf` file.

I'm not that hung up on the change because you have to provide a `JAAS_ENTRY` in order to use kerberos. It's more of a principal thing. The tests were failing on my machine prior to this PR so the failure was not related to any of these changes. As such they could be fixed outside of this PR in which case the `getMechanism` change wouldn't be needed.

vtslab:
Hi @mike-tr-adamson, I am glad you entered the discussion. I think your main point is valid, namely that there are circumstances, pointed out by you, when gremlin-driver should select the GSSAPI mechanism even though no JAAS_ENTRY is specified (ToDo: make a test for this to safeguard the desired behavior).
Having said this, the old behavior (select GSSAPI out of the blue if no username/password is supplied) also has its risks and problems given the multitude of SASL mechanisms that people could want to use, see [http://www.iana.org/assignments/sasl-mechanisms/sasl-mechanisms.xhtml](http://www.iana.org/assignments/sasl-mechanisms/sasl-mechanisms.xhtml). Ideally, one would want gremlin-server to provide a token with the mechanism(s) it supports, so that gremlin-driver can use this to instantiate the SaslClient properly. 
In your case, with `javax.security.auth.useSubjectCredsOnly=false` configured, you would have a Gremlin-Server with a Krb5Authenticator configured, the server would provide the GSSAPI token in its authentication request and gremlin-driver would know to select the GSSAPI mechanism. 
However, this ideal situation requires more changes to the gremlin-driver and gremlin-server code. 
I could live now with adding the GSSException as an option to the tests with your explanation how it could be a valid option. This solves the current challenge and we can add this discussion as comments to the code for future reference, when requirements for other SASL mechanisms pop up.

mike-tr-adamson:
Hi @vtslab, the majority of SASL mechanisms that are suitable for this form of authentication require some form of credential, be it token or certificate, at the client end. I agree that these are best approached separately.

> In your case, with javax.security.auth.useSubjectCredsOnly=false configured, you would have a Gremlin-> Server with a Krb5Authenticator configured, the server would provide the GSSAPI token in its
> authentication request and gremlin-driver would know to select the GSSAPI mechanism.

The ideal solution, in my mind anyway, would be for the server to announce which mechanism it wants the client to use. This would allow for a quick fail on the client side if any credentials required by the mechanism weren't available. It ought to be quite straightforward to add this to the authenticate response from the server. It would certainly take away some of the guessing.

I'm not suggesting that this change is done here but may be something for a future enhancement to the authentication.

vtslab:
I interpreted almosta week of silence as consensus on the gremlin-driver behavior, so I made the following changes:
- restored gremlin-driver's Handler (checkout from master)
- added a ToDo to gremlin-driver's Handler regarding receiving allowed mechanisms from gremlin-server
- added GSSException fail options to three tests that required so
- corrected the changelog

spmallette:
sorry - i was good with whatever @mike-tr-adamson was down for. he originally brought in the sasl stuff for tinkerpop, so i was glad he could take a moment to review and comment on this. if he's good with what's here now, we can have committers start to review things.

mike-tr-adamson:
I'm good with what's now there.

spmallette:
I sense that there could be some changes to LICENSE/NOTICE as a result of this work. I will make changes to those files as necessary on merge. In the mean time, the rest of this looks pretty nice.  Good tests, nice documentation updates.

VOTE +1

vtslab:
Thanks, Stephen, for guiding me through this so far. Once this gets merged, I'd like to take a look at authentication with the python driver.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpBasicAuthenticationHandler.java
```diff
@@ -92,6 +102,13 @@ public void channelRead(final ChannelHandlerContext ctx, final Object msg) {
             try {
                 authenticator.authenticate(credentials);
                 ctx.fireChannelRead(request);
+
+                // User name logged with the remote socket address and authenticator classname for audit logging
+                if (authenticationSettings.enableAuditLog) {
+                    String[] authClassParts = authenticator.getClass().toString().split("[.]");
+                    auditLogger.info("User {} with address {} authenticated by {}", credentials.get(PROPERTY_USERNAME),
+                            ctx.channel().remoteAddress().toString().substring(1), authClassParts[authClassParts.length - 1]);
```
robertdale: substring(1) assumes that the remoteAddress always has an unresolved (reverse lookup) hostname. I don't know if this is always the case.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpGremlinEndpointHandler.java
```diff
@@ -188,6 +189,10 @@ public void channelRead(final ChannelHandlerContext ctx, final Object msg) {
             try {
                 logger.debug("Processing request containing script [{}] and bindings of [{}] on {}",
                         requestArguments.getValue0(), requestArguments.getValue1(), Thread.currentThread().getName());
+                if (settings.authentication.enableAuditLog) {
+                    final String address = ctx.channel().remoteAddress().toString().substring(1);
```
robertdale: substring(1) again

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/SaslAuthenticationHandler.java
```diff
@@ -94,13 +99,17 @@ public void channelRead(final ChannelHandlerContext ctx, final Object msg) throw
...
                         final byte[] saslMessage = negotiator.get().evaluateResponse(saslResponse);
                         if (negotiator.get().isComplete()) {
-                            // todo: do something with this user
                             final AuthenticatedUser user = negotiator.get().getAuthenticatedUser();
-
+                            // User name logged with the remote socket address and authenticator classname for audit logging
+                            if (authenticationSettings.enableAuditLog) {
+                                String[] authClassParts = authenticator.getClass().toString().split("[.]");
+                                auditLogger.info("User {} with address {} authenticated by {}", user.getName(),
+                                        ctx.channel().remoteAddress().toString().substring(1), authClassParts[authClassParts.length - 1]);
```
robertdale: substring(1) again

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractEvalOpProcessor.java
```diff
@@ -243,6 +245,10 @@ protected void evalOpInternal(final Context context, final Supplier<GremlinExecu
                     final Iterator itty = IteratorUtils.asIterator(o);
 
                     logger.debug("Preparing to iterate results from - {} - in thread [{}]", msg, Thread.currentThread().getName());
+                    if (settings.authentication.enableAuditLog) {
+                        final String address = context.getChannelHandlerContext().channel().remoteAddress().toString().substring(1);
```
robertdale: substring(1) again

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpBasicAuthenticationHandler.java
vtslab: It looks suspect, indeed, thanks for the remark. remoteAddres is a SocketAddress though, so it would not fail on address resolution. To keep remoteAddress use consistent with gremlin-driver and be on the conservative side, I do not mind putting the string operation sequence in a try{} block (also for the other occurrences below).

docs/src/reference/gremlin-applications.asciidoc
```diff
@@ -1035,6 +1035,7 @@ The following table describes the various YAML configuration options that Gremli
 |=========================================================
 |Key |Description |Default
 |authentication.className |The fully qualified classname of an `Authenticator` implementation to use.  If this setting is not present, then authentication is effectively disabled. |`AllowAllAuthenticator`
+|authentication.enableAuditLog |The available authenticators can issue audit logging messages, binding the authenticated user to his remote socket address and binding requests with a gremlin query to the remote socket address. For privacy reasons, the default value of this setting is false. The audit logging messages are logged at the INFO level via the `audit.org.apache.tinkerpop.gremlin.server` logger, which can be configured using the log4j.properties file. |false
```
vtslab: Answer to this riddle (needed some thought on my part): also other config items, like scriptEngines.<name>.config, are at the bottom of their section. Just because it looks more orderly in the config file.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpBasicAuthenticationHandler.java
robertdale: Let me elaborate.  substring(1) assumes the toString() always starts with '/'.  However, if the hostname were resolved, then it would be in the format of "hostname/IP address:port".   substring(1) would result in "ostname/IP address:port".   It might not ever happen, but wanted to point it out for awareness.

docs/src/reference/gremlin-applications.asciidoc
robertdale: What I meant is that this one was not in alphabetical order and perhaps it should be.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpBasicAuthenticationHandler.java
vtslab: Thanks for clarifying, I will correct it. Tests are still running.

robertdale:
Tested -secure and non-secure configurations with http/console and server logs audit trail.  Note that the string-based requests are logged but bytecode is not.  Possibly a future enhancement.

VOTE +1

spmallette:
ooo - well noted @robertdale - i'm fine with getting that added to `TraversalOpProcessor` at a later date, but it shouldn't be forgotten. maybe i'll just add that in when i go to merge this. thanks for the review. just one more vote needed!

vtslab:
A good catch by @robertdale indeed, about the bytecode requests. I was not even aware those were human readable. A proposal: include the audit logging of bytecode requests in a new Jira ticket that addresses authentication using gremlin-python. This, because the integrate test setup will look similar.

spmallette:
ok @vtslab i won't do the work on merge then. i'm fine with letting it happen with the gremlin-python work.

dkuppitz:
VOTE: +1

PR #535, opened 2017-01-17
TINKERPOP-1601 LazyBarrierStrategy should not check is.testing

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1601

Only a minor improvement (strategies shouldn't behave differently depending on whether we're in testing mode or not). If a test requires that certain strategies don't do anything, then it's the test's job to exclude those strategies.

VOTE: +1

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Bytecode.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Bytecode.java
@@ -61,13 +61,10 @@ public final class Bytecode implements Cloneable, Serializable {
      */
     public void addSource(final String sourceName, final Object... arguments) {
         if (sourceName.equals(TraversalSource.Symbols.withoutStrategies)) {
-            final Class<TraversalStrategy>[] classes = new Class[arguments.length];
-            for (int i = 0; i < arguments.length; i++) {
-                classes[i] = arguments[i] instanceof TraversalStrategyProxy ?
-                        ((TraversalStrategyProxy) arguments[i]).getStrategyClass() :
-                        (Class) arguments[i];
+            for (final Object argument : arguments) {
+                this.sourceInstructions.add(new Instruction(sourceName, argument instanceof TraversalStrategyProxy
+                        ? ((TraversalStrategyProxy) argument).getStrategyClass() : (Class) argument));
             }
-            this.sourceInstructions.add(new Instruction(sourceName, classes));
         } else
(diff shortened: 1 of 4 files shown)


okram:
There are numerous areas where `is.testing` is used for strategies. We shouldn't just do this for one of them and push a PR. We should overhaul the entire system so we don't have some parts of the codebase using the old model and some parts using the new model. In fact, you should also do this for everywhere `is.testing` is used -- e.g. in `Translators` as well.

dkuppitz:
That was my plan. kinda. I was going to open a new ticket to get rid of `is.testing` altogether.

okram:
I was thinking more about this specific problem. The real problem is `ProfileTest`, not `LazyBarrierStrategy`. We need to generalize all the `ProfileTest` test cases such that they not be concerned with particular strategies -- as you are only solving for `LazyBarrierStrategy` (what about provider specific strategies we have not (can not) consider). For instance, you have the traversal prior to strategy application and post strategy application. The tests should use the pre-strategy application to know the steps that should have `ProfileSteps`.

dkuppitz:
The setup method now removes `LazyBarrierStrategy` and all provider specific strategies. We can't remove all strategies, since certain test suites use a different set of extra strategies that are absolutely necessary to make the tests pass (e.g. `TranslationStrategy`).

okram:
This is not the way that this should be done. We shouldn't remove any strategies. What we should do is generalize the `ProfileTests` so they are not so specific about step indexes and the like. For "step indexes and the life" we should create a `ProfileStrategyTest` that does that sort of stuff as in there, we can only load those strategies we care about.

dkuppitz:
So do you suggest a) to remove the assertions that check the metrics for specific steps or b) to find the indexes of all steps that were present before strategies were applied?

okram:
Either works, but I'm leaning more towards (a). I would do something like:

Assume the traversal:

`g.V().out().values('name').count()`

The test algorithm would:

1. If the first step is a `GraphStep` (instanceof), then make sure their is a `ProfileStep` after it and that it has a traverser count of 6.
2. Find the next step that is an instanceof `VertexStep`. (if it doesn't exist, oh well).  Check subsequent `ProfileStep`.
3. Find the next step that is an instanceof `PropertiesStep`./ (if it doesn't exist, oh well).  Check subsequent `ProfileStep`.
4. etc.

In short, try and generalize as much as possible assuming that provider strategies can wildly change the traversal.

Then in a `ProfileStrategyTest` you can do more what @rjbriody did in `ProfileTest` as there you can isolate to single strategy application.

spmallette:
closing - stale

PR #538, opened 2017-01-18
TINKERPOP-1248: OrderGlobalStep should use local star graph to compute sorts, prior to reduction.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1248

We now support having the projected data of both `OrderGlobalStep` and `SampleGlobalStep` being co-located with the `Traverser` via a traverser wrapper called `ProjectedTraverser`. This allows us to order traversers based on any projects of the local star graph -- NOT just `PathProcessor.ID`. This also reduces the complexity of ordering with `by()`-modulation as we don't have to constantly apply a projection traversal. Finally, this reduces the chances of problems with serializing the order/sample traversals -- e.g. what if side-effects are accessed and the traversals in the bi-operators are detached from the parent traversal.

--- before I VOTE +1, I'm going to see if I can use this same model for `GroupStep`. If so, huge. If its complicated, will push off for the next tp32/ release.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -27,7 +27,7 @@ TinkerPop 3.2.4 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 * Added `ProjectedTraverser` which wraps a traverser with a `List<Object>` of projected data.
-* `OrderGlobalStep` and `SampleGlobalStep` now emit traversers with their `by()`-projections and thus, can move beyond the local star graph.
+* `OrderGlobalStep` and `SampleGlobalStep` use `ProjectedTraverser` and now can work up to the local star graph in OLAP.
 * SASL negotiation supports both a byte array and Base64 encoded bytes as a string for authentication to Gremlin Server.
 * Deprecated `TinkerIoRegistry` replacing it with the more consistently named `TinkerIoRegistryV1d0`.
 * Made error messaging more consistent during result iteration timeouts in Gremlin Server.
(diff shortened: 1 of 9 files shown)


okram:
Decided to not touch `GroupStep` till the next release as it is more complicated to handle than `OrderGlobalStep` and `SampleGlobalStep`. Thus, for this PR:

VOTE +1.

okram:
```
[INFO] BUILD SUCCESS
[INFO] Total time: 03:06 h
[INFO] Finished at: 2017-01-18T19:11:37-07:00
[INFO] Final Memory: 136M/1242M
```

spmallette:
I see that you added new Gryo registrations. I assume that GraphSON can serialize `ProjectedTraverser` as there is a general serializer for that. What about `MultiComparator` - is that a concern? To extend on this issue given my recent work in the testing of IO, I'm wondering why certain registrations in Gryo don't require equal "registrations" in GraphSON. Are our tests coming up short in ensuring they have parity? Are we unnecessarily registering in Gryo sometimes? 

I noticed there's no additional tests. I assume that this is because existing tests cover this functionality and no new breaks equals no problems.

okram:
So `MultiComparator` is internal to a `Traversal`. The way in which GraphSON serializes traversals is via their Bytecode. Thus, when the bytecode gets recompiled on the deserializing end, the `MultiComparator` is recreated. The reason this is needed in Gryo is because we use Gryo as the binary format internal to Java-based systems. Giraph and Spark register the `MultiComparator` in their Aggregator framework and Gryo is the serializer for it.

Regarding `ProjectedTraverser`. Again, a concept internal to Gremlin-Java VM. Note that `Traverser` serialization in GraphSON is not complete regardless (only what is required for practical communication):

```java
@Override
        public void serialize(final Traverser traverserInstance, final JsonGenerator jsonGenerator, final SerializerProvider serializerProvider)
                throws IOException {
            jsonGenerator.writeStartObject();
            jsonGenerator.writeObjectField(GraphSONTokens.BULK, traverserInstance.bulk());
            jsonGenerator.writeObjectField(GraphSONTokens.VALUE, traverserInstance.get());
            jsonGenerator.writeEndObject();
        }
```

`ProjectedTraverser` is used by `CollectingBarrierSteps` in OLAP during reductions. For Gremlin-Java VM applications, this is all Gryo serialization stuff with aggregator frameworks.

Finally, regarding testing --- yes, I swapped out `ChainedComparator` for `MultiComparator + ProjectedTraverser` and thus, its tested.

spmallette:
Merged #539 and #538 together locally and all tests pass with `docker/build.sh -t -n -i`

VOTE +1

dkuppitz:
Code looks good. Minor improvements were discussed in Slack.

VOTE: +1

PR #540, opened 2017-01-19
TINKERPOP-1565 GraphSON 3.0 Setup

spmallette:

This pull request started as one thing but then simplified (if that's the word to use) to just setting up the framework (tests, implementation classes, etc) for GraphSON 3.0 so that we can come at building it out with other smaller pull requests in the future. This thing was just getting too big. Note that at this point there is no difference between GraphSON 2.0 and 3.0 save the version numbers. Also 3.0 is now the default version used in TinkerGraph IO and Gremlin Server.

VOTE+1


okram:
As you note, this work will evolve. Lets just get the foundation in (that you have done) and now we can go to town on designing GraphSON 3.0. 

VOTE +1.

davebshow:
I ran my driver tests against this branch using GraphSON3 and all message serialization and deserialization is working fine. I also skimmed through the changes and everything looks good.

VOTE +1

PR #541, opened 2017-01-20
TINKERPOP-1560 Used ManagedConcurrentValueMap in GremlinGroovyClassLoader

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1560

By configuring the `ManagedConcurrentValueMap` to have weak references, the `GremlinGroovyClassLoader`, and therefore the `GremlinGroovyScriptEngine`, are now able to "forget" classes that are no longer used. It was determined that the cache of these classes would grow indefinitely for each script passed to the `GremlinGroovyScriptEngine`, thus allowing the metaspace to continue to grow. It isn't really possible to write tests to verify that this change works, but I did test manually by watching memory usage in a profiler and could see that metaspace memory stayed stable and that classes were unloading from the classloader over time.

@BrynCooke could you please give a look at this one when you get a chance?

VOTE +1


BrynCooke:
+1 but do you need the synchronised blocks?

spmallette:
For some reason I thought it was necessary still, but as I think about it this morning it doesn't seem to be the case - removed.

dkuppitz:
VOTE: +1

robertdale:
VOTE +1

PR #542, opened 2017-01-20
TINKERPOP-1610 Deprecated Groovy test suites

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1610

Graph Providers no longer need to implement these going forward. Note that the new model for testing groovy will be similar to the approach taken for gremlin-python (which obviously doesn't have all this tests required of the provider). By making this change, we reduce the testing burden on graph providers as this is a whole body of tests they can just ignore now. This is a non-breaking change and will setup for removal and migration of the `gremlin-groovy-test` classes to `gremlin-groovy` (or potentally `gremlin-test` if the test is general enough to aid in testing the `GremlinScriptEngine` environment).

This is a pretty simple PR at this point, but since it was a major deprecation I wanted to go through the vote process rather than CTR.

VOTE +1


dkuppitz:
VOTE: +1

okram:
VOTE +1.

PR #547, opened 2017-01-26

pauljackson:
Add support for the closing of `Iterators` returned from
`Vertex.vertices()` and `Vertex.edges()`.
`Iterators` are closed once they are read to completion.
Make `FlatMapStep` implement `AutoCloseable` in case iterator is not
read to completion, should get closed when `Traversal` is closed.
Remove unnecessary null check (null is never an instanceof).
OLTP mode support only. More extensive changes required for OLAP.
NOTE: Rethrowing checked Exception from `close()` as unchecked
RuntimeException in order to retain `Step.reset()` and
`AbstractStep.processNextStart()` signatures.


pauljackson:
Resubmitting this pull request.

PR #548, opened 2017-01-26

pauljackson:
Add support for the closing of Iterators returned from
`Vertex.vertices()` and `Vertex.edges()`.
Make `FlatMapStep` close iterator after read to completion.
Make `EdgeVertexStep`, `PropertiesStep`, `VertexStep`
`AutoCloseable` in case iterator is not read to completion
to ensure closed when Traversal is closed.
Add unchecked `close()` helper method to `CloseableIterator`.
OLTP mode support only. More extensive changes required for OLAP.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/FlatMapStep.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/FlatMapStep.java
@@ -21,7 +21,6 @@ package org.apache.tinkerpop.gremlin.process.traversal.step.map;
 import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
 import org.apache.tinkerpop.gremlin.process.traversal.Traverser;
 import org.apache.tinkerpop.gremlin.process.traversal.step.util.AbstractStep;
-import org.apache.tinkerpop.gremlin.structure.util.CloseableIterator;
 import org.apache.tinkerpop.gremlin.util.iterator.EmptyIterator;
 
 import java.util.Iterator;
@@ -29,10 +28,10 @@ import java.util.Iterator;
 /**
  * @author Marko A. Rodriguez (http://markorodriguez.com)
  */
-public abstract class FlatMapStep<S, E> extends AbstractStep<S, E> {
+public abstract class FlatMapStep<S, E> extends AbstractStep<S, E> implements AutoCloseable {
 
     private Traverser.Admin<S> head = null;
-    protected Iterator<E> iterator = EmptyIterator.instance();
+    private Iterator<E> iterator = EmptyIterator.instance();
 
     public FlatMapStep(final Traversal.Admin traversal) {
         super(traversal);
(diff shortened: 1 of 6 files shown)


pauljackson:
Hadoop build is failing for me, but this happens under normal conditions.

spmallette:
tests all pass with docker so that's good. 

@okram would you prefer to not see this change made to `FlatMapStep` and instead see it implemented specifically in steps that interact with the Structure API, like: `VertexStep`? or is this ok from your perspective?

okram:
Yes. That makes sense to me.

spmallette:
ha - "makes sense" that the way it is now is ok or "makes sense" as with the change i proposed? :smile:

spmallette:
I'm pretty sure @okram was affirming that `VertexStep`, `EdgeVertexStep` and `PropertiesStep` should implement `AutoCloseable` (I don't think that there are others) rather than a blanket change of just applying it to `FlatMapStep`. As the logic is re-used, perhaps you could do an `ElementStep` that extends the `FlatMapStep` and implements `AutoCloseable`. Then have those three steps extend `ElementStep` which would contain the close logic that you have.

okram:
I would avoid the `ElementStep` concept right now. In 3.3.0 we will be creating a really elegant interface system that will give us this sort of flexibitily. For now, just list out the class names.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/FlatMapStep.java
```diff
@@ -44,6 +45,7 @@ public FlatMapStep(final Traversal.Admin traversal) {
                 return this.head.split(this.iterator.next(), this);
             } else {
                 this.head = this.starts.next();
+                closeIterator();
```
okram: This should go before `this.starts.next()` as if `this.starts.next()` is empty, it will throw `FastNoSuchElementException` and the previous iterator would not have been closed.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/GraphStep.java
```diff
@@ -164,13 +164,13 @@ public int hashCode() {
...
 
     /**
-     * Attemps to close an underlying iterator if it is of type {@link CloseableIterator}. Graph providers may choose
+     * Attempts to close an underlying iterator if it is of type {@link CloseableIterator}. Graph providers may choose
      * to return this interface containing their vertices and edges if there are expensive resources that might need to
      * be released at some point.
      */
     @Override
     public void close() throws Exception {
-        if (iterator != null && iterator instanceof CloseableIterator) {
```
okram: Why doesn't `GraphStep` implement `AutoCloseable` like the others?

pauljackson: `GraphStep` doesn't extend `FlatMapStep`, so it doesn't have access to `closeIterator()`, but it can call `CloseableIterator.closeIterator(iterator)`

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/FlatMapStep.java
pauljackson: I'll update the PR.

okram:
VOTE +1.

spmallette:

VOTE +1

dkuppitz:
VOTE: +1

PR #549, opened 2017-01-26
TINKERPOP-1617: Create a SingleIterationStrategy which will do its best to rewrite OLAP traversals to not message pass.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1617

There are various traversals that can be rewritten using `local()` that will enable the `GraphComputer` to avoid a message pass and thus, can accomplish the computation in a single scan of the graph. Benefiting traversal examples include:

```
g.V().out().id() --> g.V().local(out().id())
g.V().out().id().count() --> g.V().local(out().id()).count()
g.V().out().id().dedup().count()
g.V().inE().values("weight") // realize that in-edges are hosted by the out-vertex
g.V().inE().values("weight").sum()
 g.V().as("a").outE().inV().as("b").id().dedup("a", "b").by(T.id).count()
```

Finally, the traversal that sparked this PR:

```
g.V().in().id().select("articleNumber").dedup().count()    // requires one message pass

==translatesTo==>

g.V().local(in().id().select("articleNumber")).dedup().count() // requires no message passing
```

`SingleIterationStrategy` plays well with `SparkSingleIterationStrategy` which determines whether it is necessary to `cache()` and/or `partition()` the graph. If the traversal can be accomplished without a message pass (i.e. a single iteration), then performance is greatly improved as RDD partitions can be dropped as they are processed sequentially.

VOTE +1.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,7 +26,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.2.4 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-* Added `SingleIterationStrategy` as a default `GraphComputer` strategy that can re-write some traversals to not require message passing.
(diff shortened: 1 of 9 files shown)


twilmes:
`mvn clean install` was successful.  Good set of test cases.

VOTE: +1

dkuppitz:
`docker/build.sh -t -i -n` succeeded. Manual tests were showing epic performance results. Best PR in a long time.

VOTE: +1

PR #550, opened 2017-01-27
Fix the serializers lookup to handle the type first, then the class inheritance

aboudreault:
A good example of the need for this fix is with these types:
* datetime.date
* datetime.datetime

Since a datetime is a subclass of a date and that a dict is unordered... it might happen that the wrong serializer is selected. (since isinstance(my_datetime_instance, datetime.date) == True)


spmallette:
Thanks for this. Could you please edit your pull request at the tp32 branch so that we can apply this changed to 3.2.4?

aboudreault:
closed in favor of: https://github.com/apache/tinkerpop/pull/551

PR #551, opened 2017-01-27
Fix the serializers lookup to handle the type first, then the class inheritance

aboudreault:
A good example of the need for this fix is with these types:

    datetime.date
    datetime.datetime

Since a datetime is a subclass of a date and that a dict is unordered... it might happen that the wrong serializer is selected. (since isinstance(my_datetime_instance, datetime.date) == True)

### DIFF
--- a/gremlin-python/src/main/jython/gremlin_python/structure/io/graphson.py
+++ b/gremlin-python/src/main/jython/gremlin_python/structure/io/graphson.py
@@ -293,11 +293,7 @@ class Int32IO(_NumberIO):
     def dictify(cls, n, writer):
         if isinstance(n, bool):
             return n
-        if isinstance(n, LongType):
-            graphson_base_type = Int64IO.graphson_base_type
-        else:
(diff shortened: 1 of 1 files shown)


aboudreault:
@spmallette The PR for tp32.

spmallette:
excellent - thank you

davebshow:
This makes a lot of sense. You can probably remove the instance check for the custom long type after making this change: https://github.com/apache/tinkerpop/blob/master/gremlin-python/src/main/jython/gremlin_python/structure/io/graphson.py#L346-L348

aboudreault:
@davebshow yes, just removed it.

spmallette:
did a quick build with `mvn clean install` and everything passes. if @davebshow is cool with this at this point after the his request change, then VOTE +1

davebshow:
VOTE +1

spmallette:

VOTE +1

okram:
VOTE +1.

spmallette:
I'll go ahead and get this merged.

PR #552, opened 2017-01-27
Remove tests from packages

GregBestland:
Installing your tests package as part of the setup.py is probably not a good idea.

Most every python project has a tests module for unit/integration tests, this is not generally distributed/installed as part of the package. When we install a tests module as part of the gremlinpython package we end up creating a scenario where a namespace collision can occur with projects that wish to leverage the gremlinpython api. The gremlinpython tests module will end up superseding their own local tests directory in the sys.path, this creates problems for them when they go to try to run their own unit/integration tests.  If you do need to install it, i'd recommend giving it a unique name like "gremlinpythontests", so as to prevent a namespace collision with the more commonly used "tests" module.

### DIFF
--- a/gremlin-python/src/main/jython/setup.py
+++ b/gremlin-python/src/main/jython/setup.py
@@ -46,7 +46,7 @@ version = __version__.version
 setup(
     name='gremlinpython',
     version=version,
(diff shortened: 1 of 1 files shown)


spmallette:
I don't know why that was like that. Don't think we had any reason to  include tests that i can think of.  Could you please re-target the pull request at the tp32 branch?

davebshow:
I agree. This causes a separate `tests` module to be installed in `site-packages`. This was probably based on a bad setup file I had. Will upvote after re-target.

PR #554, opened 2017-01-30
TINKERPOP-1599 implement real gremlin-python driver

davebshow:
https://issues.apache.org/jira/browse/TINKERPOP-1599

This PR adds a better driver implementation for gremlin-python:

- Uses a multi-threaded solution to provide asynchronous I/O
- Decouples the underlying websocket client implementation from driver code, making it easy to plug in a different client
- Makes it easy to plug in different protocols, like the Gremlin Server HTTP protocol.
- Adds simple connection pooling used in concurrent requests, which increases driver performance with slow traversals
- Improves driver tests by adding `pytest` fixtures and removing unneeded `unittest` code.

This driver still isn't full featured compared to the Java driver, for example, it doesn't implement `Cluster` to use multiple hosts. But, it is considerably better than the old implementation, and as this was becoming a big PR, I figure we can add more features with subsequent PRs. 

Also note, I know we are in code freeze right now, so I don't expect this to be reviewed/merged for 3.2.4, but I'm busy and I wanted to get this done so I can move on.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,11 @@ TinkerPop 3.2.5 (Release Date: NOT OFFICIALLY RELEASED YET)
 * Refactor `SparkContext` handler to support external kill and stop operations.
 * Fixed an optimization bug in `LazyBarrierStrategy` around appending barriers to the end of a `Traversal`.
 * `TraverserIterator` in GremlinServer is smart to try and bulk traversers prior to network I/O.
+* Improved Gremlin-Python Driver implementation by adding a threaded client with basic connection pooling and support for pluggable websocket clients.
+
+Improvements
+^^^^^^^^^^^^
+TINKERPOP-1599 implement real gremlin-python driver
 
 [[release-3-2-4]]
 TinkerPop 3.2.4 (Release Date: February 8, 2017)
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -31,10 +31,6 @@ TinkerPop 3.2.5 (Release Date: NOT OFFICIALLY RELEASED YET)
 * `TraverserIterator` in GremlinServer is smart to try and bulk traversers prior to network I/O.
 * Improved Gremlin-Python Driver implementation by adding a threaded client with basic connection pooling and support for pluggable websocket clients.
 
-Improvements
-^^^^^^^^^^^^
-TINKERPOP-1599 implement real gremlin-python driver
-
 [[release-3-2-4]]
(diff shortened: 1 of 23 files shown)


spmallette:
@davebshow now that code freeze is lifted i think you should probably rebase this PR. Also, two documentation related items:

1. This works needs a changelog entry - it was probably better you didn't add that yet since you would have had to move it anyway to 3.2.5.
2. This change seems sufficiently large and important that it should have some documentation on usage in the reference docs and should likely have an entry in the upgrade docs as well so that users know that there as an actual driver that can be used for python now.

davebshow:
Yeah I figured we would need to add docs. Where should I add to the reference docs? Maybe in [Gremlin Applications](https://github.com/apache/tinkerpop/blob/master/docs/src/reference/gremlin-applications.asciidoc) after the [Connecting via Java](https://github.com/apache/tinkerpop/blob/master/docs/src/reference/gremlin-applications.asciidoc#connecting-via-java) section?

spmallette:
I think that make sense - just add a "Connecting via Python" section perhaps.

davebshow:
Ok, I have now rebased ad added docs to both reference and upgrading. I also reviewed the old implementation and made a couple small fixes for consistency. I also added standard op processor message serialization. I think this one is getting close to being ready to go.

CHANGELOG.asciidoc
```diff
@@ -29,6 +29,11 @@ TinkerPop 3.2.5 (Release Date: NOT OFFICIALLY RELEASED YET)
 * Refactor `SparkContext` handler to support external kill and stop operations.
 * Fixed an optimization bug in `LazyBarrierStrategy` around appending barriers to the end of a `Traversal`.
 * `TraverserIterator` in GremlinServer is smart to try and bulk traversers prior to network I/O.
+* Improved Gremlin-Python Driver implementation by adding a threaded client with basic connection pooling and support for pluggable websocket clients.
+
+Improvements
+^^^^^^^^^^^^
+TINKERPOP-1599 implement real gremlin-python driver
```
spmallette: you don't need to add those manually - they get added on release (we generate a report out of jira for it).

davebshow:
Well, I've made my last review of this code and pushed a bit of cleanup and a small fix. IMHO, this is ready to be merged. I'll wait to see if there are any more comments or feedback before upvoting.

spmallette:

VOTE +1

davebshow:
I think this is ready. Anyone want to be a third reviewer for this?

VOTE +1

okram:
VOTE +1.

davebshow:
Merged.

PR #555, opened 2017-02-02
TINKERPOP-1271: Refactor SparkContext creation and handling of sc.stop()

artem-aliev:
org.apache.tinkerpop.gremlin.spark.structure.Spark is a SparkContext holder for SparkGraphComputer.
    It was refactored to detect external stop calls and recreate SparkContext in that case.
    Context creation process was reordered to make all configuration options take effect.
    Spark.create() methods return created context now.
    The external stop also requires SPARK-18751 fix, that was integrated into Spark 2.1.
    By the way the refactoring and configuration loading part gives effect on all versions.


okram:
VOTE +1.

spmallette:

VOTE +1

dkuppitz:
VOTE: +1

PR #559, opened 2017-02-15
TINKERPOP-1627: LazyBarrierStrategy should not append an end barrier.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1627

You don't want `LazyBarrierStrategy` appending `NoOpBarriers` to the root traversal. It was doing that because I was trying to be clever and have it consider `RemoteTraversal` I/O (via bulking). Dumb. `RemoteStrategy` should be responsible for that concern.  Fixed.

VOTE +1.


dkuppitz:

VOTE: +1

spmallette:

VOTE +1

PR #560, opened 2017-02-16
TINKERPOP-1554: has(propertyKey) should have a corresponding step in Gremlin-Java.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1554

Currently `has('name')` compiles to `filter(values('name'))`. Its a used enough "step" that it should really have its own direct step compilation for efficiency, understability, and ease of strategy manipulation.

VOTE +1.

### DIFF
--- a/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/FilterRankingStrategyTest.java
+++ b/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/FilterRankingStrategyTest.java
@@ -106,7 +106,7 @@ public class FilterRankingStrategyTest {
                 {__.dedup().has("value", 0).or(not(has("age")), has("age", 10)).has("value", 1), __.has("value", 0).has("value", 1).or(not(has("age")), has("age", 10)).dedup(), Collections.singletonList(InlineFilterStrategy.instance())},
                 {__.dedup().filter(out()).has("value", 0), has("value", 0).filter(out()).dedup(), Collections.emptyList()},
                 {filter(out()).dedup().has("value", 0), has("value", 0).filter(out()).dedup(), Collections.emptyList()},
-                {__.as("a").out().has("age").where(P.eq("a")), __.as("a").out().where(P.eq("a")).has("age"), Collections.emptyList()},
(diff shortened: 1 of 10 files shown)


twilmes:
Looks good.  VOTE: +1

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/FilterRankingStrategyTest.java
```diff
@@ -106,7 +106,7 @@ public void doTest() {
                 {__.dedup().has("value", 0).or(not(has("age")), has("age", 10)).has("value", 1), __.has("value", 0).has("value", 1).or(not(has("age")), has("age", 10)).dedup(), Collections.singletonList(InlineFilterStrategy.instance())},
                 {__.dedup().filter(out()).has("value", 0), has("value", 0).filter(out()).dedup(), Collections.emptyList()},
                 {filter(out()).dedup().has("value", 0), has("value", 0).filter(out()).dedup(), Collections.emptyList()},
-                {__.as("a").out().has("age").where(P.eq("a")), __.as("a").out().where(P.eq("a")).has("age"), Collections.emptyList()},
+                {__.as("a").out().has("age").where(P.eq("a")), __.as("a").out().has("age").where(P.eq("a")), Collections.emptyList()},
```
dkuppitz: Should `has("age")` really be processed prior `where(eq("a"))`? I would assume that the latter is generally faster (and significantly faster if you're dealing with a `ReferenceVertex` that - afaik - fetches properties only on request).

okram: `where(eq('a')` might be faster, but probably not. But we don't know what the `P` predicate is. If its something more complex, than `==` then it will be slower. And no, `ReferenceVertex` doesn't come into play here as those are resolved at "attachment" and not dynamically in mid-traversal.

dkuppitz:
VOTE: +1

okram:
Realizing that this should really go into `master/` as I just walked through a bunch of the code and notes that we use `PropertiesStep` a lot through out various traversal strategies. Going to change my VOTE to -1 as this problem is more complicated that I originally assumed.

PR #561, opened 2017-02-18
TINKERPOP-1635 Fix duplicate serialization of element property in PropertySerializer

FlorianHockmann:
This fixes the duplicate serialization issue in gremlin-python's PropertySerializer:
[https://issues.apache.org/jira/browse/TINKERPOP-1635](https://issues.apache.org/jira/browse/TINKERPOP-1635)

This pull requests removes the second call to `writer.toDict()` and changes a property value in the respective unit test to an int. That makes the test fail with the duplicate serialization and pass with the fixed version.

### DIFF
--- a/gremlin-python/src/main/jython/gremlin_python/structure/io/graphson.py
+++ b/gremlin-python/src/main/jython/gremlin_python/structure/io/graphson.py
@@ -225,7 +225,7 @@ class PropertySerializer(_GraphSONTypeIO):
                 del valueDict["value"]
         return GraphSONUtil.typedValue("Property", {"key": writer.toDict(property.key),
                                                     "value": writer.toDict(property.value),
-                                                    "element": writer.toDict(elementDict)})
+                                                    "element": elementDict})
 
 
 class TraversalStrategySerializer(_GraphSONTypeIO):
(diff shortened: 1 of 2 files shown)


okram:
Should this be directed to `tp32/` and not `master/`? Or is this a breaking change?

FlorianHockmann:
The PropertySerializer doesn't exist in the `tp32` branch, it was only added to the master. So the problem also only exists in `master`.

okram:
Note that we will be heavily refactoring GraphSON for TinkerPop 3.3.0 (`master/`) so most of this will be gutted. I will leave it up to @spmallette / @newkek / @aholmberg to decide whether this is necessary to merge.

FlorianHockmann:
Sure, I just noticed it while working on GraphSON for Gremlin.CSharp and it seemed strange that integer properties were serialized twice.

spmallette:
I'm not sure at this point what will be changed and what won't for 3.3.0 so I think it makes sense to fix stuff that isn't right even if it might change later. @davebshow does this change look ok to you?

davebshow:
Yeah it looks fine to me. It seems that properties are serialized before the element dict is "cleaned up". I don't see why you would serialize it again. However, I haven't really familiarized myself with the new GraphSON work on master yet. Still, I see no reason why we can't merge this. If the code is refactored and it is no longer relevant, so be it.

spmallette:
merged via CTR as this was a small change and there seemed to be general consensus to merge given comments.

PR #563, opened 2017-02-23
TINKERPOP-1638 count() is optimized away in where()

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1638

Fixed a bug in `RangeByIsCountStrategy` that changed the meaning of inner traversals.

VOTE: +1


okram:
VOTE +1.

robertdale:
VOTE +1

PR #564, opened 2017-02-24
TINKERPOP-1639 Support simple String operations

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1639

Added support for simple String operations. `CountLocalStep`, `RangeLocalStep` and `TailLocalStep` can now imitate `String::length()` and `String::substring()`.

A few examples:

```
gremlin> g.V().values("name").count(local)
==>5
==>5
==>3
==>4
==>6
==>5
gremlin> g.V().values("name").limit(local, 3)
==>mar
==>vad
==>lop
==>jos
==>rip
==>pet
gremlin> g.V().values("name").tail(local, 4)
==>arko
==>adas
==>lop
==>josh
==>pple
==>eter
gremlin> g.V().values("name").group().by(range(local, 1, 2)).next()
==>a=[marko, vadas]
==>e=[peter]
==>i=[ripple]
==>o=[lop, josh]
```

VOTE: +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.2.5 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Added support for simple String operations. `CountLocalStep`, `RangeLocalStep` and `TailLocalStep` can now imitate `String::length()` and `String::substring()`.
 * Removed `HasTest.g_V_hasId_compilationEquality` from process test suite as it makes too many assumptions about provider compilation.
 * Deprecated `CustomizerProvider` infrastructure.
 * Improved consistency of the application of bindings to `GremlinScriptEngine` implementations in the `BindingsGremlinPlugin`.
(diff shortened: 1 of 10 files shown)


okram:
I do not think we should support this feature. This is starting to mix semantics between what is a "data structure" and what is a "primitive."

okram:
@dkuppitz asked me to say more of my thoughts on the matter.

I think we need a general theory for how we are going to handle primitive manipulations. That is, math operators, string operations, character operations, type casting operations, etc.

If we just append this feature for `Strings` then we might end up backtracking come a more general theory. It just seems that this PR is a "this would be cool" and implemented. Not: "So, primitives all fit this style of manipulation in a flow-based system..."

Questions that arise:

* What about `unfold()`? Now we are in character manipulation or just single character strings?
* What about `fold()`? Does character folding become strings?
* What about `range()` on doubles/floats? Is that a way to clip precision?

I don't think we should just add this PR without a fully thought through model for all primitives and how they will be effected by the current or "need to be built" steps.

okram:
Also, what other string operations do we need? I'm thinking reg-ex.

So now we have a `filter()` steps for regex, but then what about pattern matching selection. Well, thats not range, but something new. Now you have all your string manipulation features spread out across numerous steps. Is this good? I don't know. Seems that all primitive operations should be grouped via predicate and operators else we will just be building a new data flow language (steps steps steps steps) for primitives and all the operations you want to do on them.

dkuppitz:
It's true that this was just "would be cool" PR, that would have helped me in the short term. The general solution to this problem is proposed in https://issues.apache.org/jira/browse/TINKERPOP-1632.

Closing.

PR #567, opened 2017-03-06
TINKERPOP-1644 Improve script compilation syncronisation

BrynCooke:
Script compilation is synchronised.
Script compilation times are placed in to logs.
Failed scripts will not be recompiled.
Scripts that take over 5 seconds to compile are logged as a warning.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,10 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.2.5 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Script compilation is synchronised.
+* Script compilation times are placed in to logs.
+* Failed scripts will not be recompiled.
+* Scripts that take over 5 seconds to compile are logged as a warning.
 * Split `ComputerVerificationStrategy` into two strategies: `ComputerVerificationStrategy` and `ComputerFinalizationStrategy`.
 * Removed `HasTest.g_V_hasId_compilationEquality` from process test suite as it makes too many assumptions about provider compilation.
 * Deprecated `CustomizerProvider` infrastructure.
--- a/gremlin-groovy/pom.xml
+++ b/gremlin-groovy/pom.xml
@@ -70,6 +70,11 @@ limitations under the License.
             <artifactId>jbcrypt</artifactId>
             <version>0.4</version>
         </dependency>
+        <dependency>
+            <groupId>com.github.ben-manes.caffeine</groupId>
+            <artifactId>caffeine</artifactId>
+            <version>2.3.1</version>
+        </dependency>
         <!-- TEST -->
         <dependency>
             <groupId>org.slf4j</groupId>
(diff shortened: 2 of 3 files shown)


spmallette:
I will clean up a few code formatting things on merge, but this looks good to me. All tests pass with `docker/build.sh -t -n -i`

VOTE +1

dkuppitz:
Yep, missing some spaces here and there, but the code in general looks good.

VOTE: +1

robertdale:
I have a concern on overall performance.  I agree that "Script compilation is synchronised" helps in the rare case when two or more identical scripts are submitted concurrently.  However, in the more likely case of new, unique scripts being submitted concurrently, it would seem that this effectively serializes script compilation creating a bottleneck.  Unless I'm missing something, parallel compilation seems to be the better choice.

dkuppitz:
Hmm, I haven't done any benchmarks, but this seem to be valid points. Just a thought: Would it help to have the class map be like `ManagedConcurrentValueMap<String, Future<Class>> classMap`?

robertdale:
That sounds like a good alternative.

BrynCooke:
I like the idea of using a Future.

robertdale:
VOTE +1

dkuppitz:
This part:

```
if (clazz != null) {
    return clazz.get();
}

clazz = CompletableFuture.supplyAsync(() -> loader.parseClass(script, generateScriptName()));
classMap.put(script, clazz);
return clazz.get();
```

should be more like:

```
classMap.computeIfAbsent(script, s ->
         CompletableFuture.supplyAsync(() -> loader.parseClass(s, generateScriptName())))
return classMap.get(script).get();
```

No?

okram:
Yes. `computeIfAbsent()` is much better than synchronized `get()`, check, `put()`.

spmallette:
I like this way better than the synchronization. I have some reservations about:

```java
CompletableFuture.supplyAsync(() -> loader.parseClass(script, generateScriptName()));
```

That's going to use the `ForkJoinPool` to execute the compilation, which is a shared thread pool for the JVM that isn't configurable and basically outside of our control. Of course, getting rid of it means adding a new `ExecutorService` that we can configure and monitor in `GremlinGroovyScriptEngine`.  Making that kind of change goes pretty deep into various aspects of TinkerPop. If @BrynCooke is ok with the changes I can merge what he has to a branch since it seems like we have general consensus on the approach and then make some additional changes and re-issue the PR for a final VOTE.  If there's no objections I'll start work on that first thing tomorrow.

BrynCooke:
1. ManagedConcurrentValueMap is not a map. I don't really know why we use this. but computeIfAbsent is not present on this class.
2. Guava LoadingCache and MoreExecutors.directExecutor() is the correct way of solving this problem. It allows soft values and does all the syncronisation stuff for us. Is there a reason why guava is not used by Tinkerpop?

If we want avoid guava we need to have a striped lock on script, and to create our own direct executor.

spmallette:
`ManagedConcurrentValueMap` allows you to configure a `Map` with weak/soft/phantom/hard references. Groovy has "Managed" classes for lots of different types of Lists/Maps.

spmallette:
I've merged these changes to a branch so that I can continue with the development needed on this one: 

https://github.com/apache/tinkerpop/tree/TINKERPOP-1644

@BrynCooke you can feel free to close this PR now. I'll issue a new one from the TinkerPop branch when I'm done.

PR #568, opened 2017-03-06
TINKERPOP-1645 Bump to groovy 2.4.9

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1645

Just a version bump, but as it is a version bump an explicit vote would be good.  I've only ran tests for 'tp31' at this point, but will run tests on the other branches prior to merge. There really shouldn't be any problems as this release of groovy was pretty small.

VOTE +1


okram:
VOTE +1.

robertdale:
VOTE +1

PR #569, opened 2017-03-06
TINKERPOP-1438: GraphManager becomes a customizable interface

dpitera:


spmallette:
thanks for submitting this. i didn't realize you'd get something out so quickly. before i dig too deeply into this have you run the full integration test suite to success on this?

pluradj:
No, it is mentioned here http://tinkerpop.apache.org/docs/3.2.4/dev/developer/#_contributing_code_changes

Devs often use Docker for to run it `docker/build.sh -t -i -n`

dpitera:
Thanks Jason! Reading/running now

spmallette:
we can rely on travis as a good basic health check for a PR, but ultimately a full integration test is run on just about everything. at this stage of your PR, i'm mostly interested in the workings of the gremlin-server and gremlin-console integration tests, but making sure your system can do a run of everything isn't a bad idea.

dpitera:
Integration tests finished successfully:

```
[INFO] Reactor Summary:
[INFO]
[INFO] Apache TinkerPop .................................. SUCCESS [8:43.526s]
[INFO] Apache TinkerPop :: Gremlin Shaded ................ SUCCESS [21.090s]
[INFO] Apache TinkerPop :: Gremlin Core .................. SUCCESS [1:14.775s]
[INFO] Apache TinkerPop :: Gremlin Test .................. SUCCESS [12.833s]
[INFO] Apache TinkerPop :: Gremlin Groovy ................ SUCCESS [3:02.384s]
[INFO] Apache TinkerPop :: Gremlin Groovy Test ........... SUCCESS [6.146s]
[INFO] Apache TinkerPop :: TinkerGraph Gremlin ........... SUCCESS [4:27.224s]
[INFO] Apache TinkerPop :: Gremlin Benchmark ............. SUCCESS [6.937s]
[INFO] Apache TinkerPop :: Gremlin Driver ................ SUCCESS [12.051s]
[INFO] Apache TinkerPop :: Neo4j Gremlin ................. SUCCESS [23:50.255s]
[INFO] Apache TinkerPop :: Gremlin Server ................ SUCCESS [17:12.591s]
[INFO] Apache TinkerPop :: Gremlin Python ................ SUCCESS [49.683s]
[INFO] Apache TinkerPop :: Hadoop Gremlin ................ SUCCESS [9:23.200s]
[INFO] Apache TinkerPop :: Spark Gremlin ................. SUCCESS [21:27.122s]
[INFO] Apache TinkerPop :: Giraph Gremlin ................ SUCCESS [2:21:10.655s]
[INFO] Apache TinkerPop :: Gremlin Console ............... SUCCESS [3:01.607s]
[INFO] Apache TinkerPop :: Gremlin Archetype ............. SUCCESS [0.134s]
[INFO] Apache TinkerPop :: Archetype - TinkerGraph ....... SUCCESS [1:12.427s]
[INFO] Apache TinkerPop :: Archetype - Server ............ SUCCESS [12.415s]
[INFO] BUILD SUCCESS
[INFO] Total time: 3:56:48.949s
[INFO] Finished at: Thu Mar 09 19:35:00 UTC 2017
[INFO] Final Memory: 109M/701M
```

okram:
Congrats on doing your first full integration test! Hope you stick around to help out on TinkerPop.

spmallette:
Nice to see that all the tests are passing. Thanks for doing that. 

I've given this a quick review and have a few top-level comments:

* Note that our code style is to explicitly use `final` for all variable declarations that require it - i think we make exceptions on exceptions in catch blocks and the var declaration of a `for()` loop, but that's about it. I think you've missed a few of those.
* Could you please rename `BasicGraphManager` to `DefaultGraphManager`? We only use the term "Basic" in one other class naming, so it's a bit of an anomaly compared to "Default".
* Does this change introduce any compile-time breaking changes for anyone who has depended on the  `GraphManager` class (or other things you've modified)? I'm wondering if this is a change better suited for the master branch and 3.3.0. If it stays on `tp32` I may have to ask you to provide a second pull request to master that can also be evaluated as part of a second review/vote.
* The addition of `addGraph()` is nice.
* Can you elaborate on the usage of `openGraph()` - I think I get it, but it's not clear to me based on the javadoc/default implementation
* Please add a entry (or more) to CHANGELOG for these updates.
* Please update the reference documentation to include your new Gremlin Server configuration option here: http://tinkerpop.apache.org/docs/current/reference/#_configuring_2 - not sure how much you need to write here. Users of `GraphManager` have likely dug pretty deep into the code and will see how it works. Perhaps include a link to the javadoc for `GraphManager` and beef up the class javadoc there so that implementers know what to do.

Just a bit of warning - from my perspective, this is a fairly involved pull request you've started (and thanks!) so it will likely take a some time and iterations to get through before we can get it merged.

dpitera:
> Does this change introduce any compile-time breaking changes for anyone who has depended on the GraphManager class (or other things you've modified)?

I believe the answer is no (in the non-mutated use of pure tinkerpop) because:

1. The previous implementation's graphManager instantiation could only be retrieved by two places:
  1. The [ServerGremlinExecutor](https://github.com/apache/tinkerpop/blob/ad51fb24eb4d621763f6b3c02029daa54e7dc20d/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/util/ServerGremlinExecutor.java#L202) which is only accessible through the [GremlinServer](https://github.com/apache/tinkerpop/blob/c5b9680e70d696d000b945abbd72a3b8a3f97256/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GremlinServer.java#L328) which users do not have access to the instantiation unless they have modified their implementor's logic to define its own "GremlinServer" and not use Tinkerpop's
  2. The [Context](https://github.com/apache/tinkerpop/blob/4293eb333dfbf3aea19cd326f9f3d13619ac0b54/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Context.java#L80) which also is instantiated by the same graphManager pulled from the servergremlinexecutor, and the Context class is final (and thus cannot be extended) and there is no way to retrieve that `Context` in the code.

Therefore, i am not sure who would have been able to make use of the graphManager themselves without using a modified version of tinkerpop.

> Can you elaborate on the usage of openGraph()

The `addGraph()` and `getGraph()` are to act as modifiers to the object holding the graph references, while `opengraph()` allows for a mechanism by which to instantiate a graph dynamically through the graphManager (and thus allow for reference tracking of it) that is not currently in said graphManager's reference tracking object. The use of a `Supplier` here allows for the implementor to define his/her own custom function by which to instantiate the graph

>  If it stays on tp32 I may have to ask you to provide a second pull request to master that can also be evaluated as part of a second review/vote.

I would say once we get this one merged, then we can open the PR on master, but I would like to get it into the next release as well

> it will likely take a some time and iterations to get through before we can get it merged

Sounds good. I usually prefer quality to expediency. When I get around to fixing up some of your requests, I will leave a comment here saying I have pushed new code.

Thanks for your review. Let me know if you have more questions!

spmallette:
> Therefore, i am not sure who would have been able to make use of the graphManager themselves without using a modified version of tinkerpop.

You would be surprised by what folks manage to make use of in their code. If it's public, people will use it. :smile:  anyway, i mostly wanted to be sure we don't introduce any compile time breaks when folks go to use 3.2.5.

pluradj:
It would break if somebody was calling on `new GraphManager(Settings)`. You could change the interface to `IGraphManager` and have `GraphManager` implement it.

dpitera:
> It would break if somebody was calling on new GraphManager(Settings)

If someone were doing this, it would mean they have had to reconstruct a `Settings` object, which would be different than that instantiated inside the `GremlinServer` and I would argue this is _outside the scope_ of something we need to be compatible with, i.e. using a modified version of Tinkerpop, or using it in an unexpected way. What are your thoughts here?

spmallette:
yeah - that's an obvious truth @pluradj . Normally we wouldn't allow such a change on `tp32` because it does immediately make people's code not compile if they used that class in that way but this class is fairly internal. If someone is using it they likely know what they are doing. @dpitera I tend to agree that this is outside the scope of what would need to remain compatible. That said, I think you will have to add an entry to the upgrade docs (i guess the user section???) about this breaking change and just mention how they go about fixing it (i.e. use the `DefaultGraphManager`). We'll probably also want to send a quick email to the dev list to just alert everyone to a break to make sure no one is buggin' about that. Finally, the JIRA issue will need the "breaking" label added to it.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
```diff
@@ -21,139 +21,82 @@
...
-    }
+    public Graph getGraph(String gName);
 
     /**
-     * Get a list of the {@link TraversalSource} instances and their binding names as defined by Gremlin Server
-     * initialization scripts.
+     * Add {@link Graph} g with name {@link String} gName to 
+     * {@link Map<String, Graph>} returned by call to getGraphs()
+     */
+    public void addGraph(String gName, Graph g);
```
pluradj: seems like `removeGraph(String)` or `removeGraph(Graph)` would be useful

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/util/BasicGraphManager.java
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * Selectively commit transactions on the specified graphs or the graphs of traversal sources.
+     */
+    public void commit(final Set<String> graphSourceNamesToCloseTxOn) {
+        closeTx(graphSourceNamesToCloseTxOn, Transaction.Status.COMMIT);
+    }
+
+    /**
+     * Basic graphManager has basic openGraph function.
```
pluradj: javadoc should mention that the `Supplier` is not actually used in this default implementation class.

dpitera:
I have updated my PR with both of your guys' comments, and I have mailed the dev list.

dpitera:
> and I have mailed the dev list.

scratch that-- failed to send lol

CHANGELOG.asciidoc
```diff
@@ -47,6 +47,9 @@ TinkerPop 3.2.5 (Release Date: NOT OFFICIALLY RELEASED YET)
 * Improved error handling of compilation failures for very large or highly parameterized script sent to Gremlin Server.
 * Fixed a bug in `RangeByIsCountStrategy` that changed the meaning of inner traversals.
 * Improved Gremlin-Python Driver implementation by adding a threaded client with basic connection pooling and support for pluggable websocket clients.
+* Changed GraphManager from a final class implementation to an interface.
```
spmallette: minor formatting thing - make sure that class/interface/code names have the backtick around them so that they format properly when the asciidoc is generated. you might want to look at the other docs you wrote as you missed a few instances there as well.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
```diff
@@ -21,139 +21,87 @@
...
+    
     /**
-     * Get a list of the {@link Graph} instances and their binding names as defined in the Gremlin Server
-     * configuration file.
+     * Get {@link Graph} instance whose name matches {@link gName}
      *
-     * @return a {@link Map} where the key is the name of the {@link Graph} and the value is the {@link Graph} itself
+     * @return {@link Graph} if exists, else null 
+     */
+    public Graph getGraph(String gName);
```
spmallette: Looks like you're still missing a bunch of "final" declarations. I would normally just handle them for you at merge, but you have a pretty large PR going here, so it would be helpful if you got them all for us.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/util/ServerGremlinExecutor.java
```diff
@@ -100,9 +101,28 @@ public ServerGremlinExecutor(final Settings settings, final ExecutorService grem
      */
     public ServerGremlinExecutor(final Settings settings, final ExecutorService gremlinExecutorService,
                                  final T scheduledExecutorService, final Class<T> scheduleExecutorServiceClass,
-                                 final GraphManager graphManager) {
+                                 GraphManager graphManager) {
         this.settings = settings;
 
+        if (null == graphManager) {
```
spmallette: looks like there's some indentation problems in the following lines of code. should be four spaces between the curly brackets.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/util/DefaultGraphManager.java
```diff
@@ -0,0 +1,201 @@
...
+        }
+        return supplier.get();
+    }
+
+    /**
+     *  Close graph
+     */
+    public void closeGraph(Graph graph) {
+        try {
+            graph.close();
```
spmallette: What does it mean to actually call `closeGraph()`? If "closed" does that mean it should still be in the available set of graphs? If so it would still be accessible to requests but it would be in a closed state. Different graphs will have different semantics with respect to what `close()` will do. Anyway, just posing the issue for thought at this point....not sure i have a strong opinion yet on what the behavior should be.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
```diff
@@ -21,139 +21,87 @@
...
-                    graph.tx().rollback();
-            }
-        });
-    }
-}
\ No newline at end of file
+    /**
+     * Implementation that allows for custom graph-closing implementations.
+     */
+    public void closeGraph(Graph graph);
```
spmallette: I wonder if the signature for this method should be:

```java
public void closeGraph(final String graphName);
```

maybe i'm missing the point of the method...i'm just basing this on my interpretation of the implementation in `DefaultGraphManager`.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/util/DefaultGraphManager.java
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * Basic graphManager has basic openGraph function.
+     */
+    public Graph openGraph(String graphName, Supplier<Graph> supplier) {
+        final Graph graph = graphs.get(graphName);
+        if (null != graph) {
+            return graph;
+        }
+        return supplier.get();
```
spmallette: Given your explanation of `openGraph()` in your previous comments:

> opengraph() allows for a mechanism by which to instantiate a graph dynamically through the graphManager (and thus allow for reference tracking of it) that is not currently in said graphManager's reference tracking object.

I'd expected to see the result of `supplier.get()` assigned into `graphs` - shouldn't that happen? or no? btw, as i'd mentioned in my earlier comment from a few days ago it would be good to see more javadoc around the `GraphManager` interface methods, as this will now be a public interface that will be promoted as a feature. The explanation you gave made sense - I would just format it to be part of the javadoc.

spmallette:
I think I'm done with this round of comments. I suspect I might have a few more yet about the `GraphManager` interface itself - still rolling some ideas around in my mind on it.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/util/DefaultGraphManager.java
dpitera: > I'd expected to see the result of supplier.get() assigned into graphs

Asbolutely

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
dpitera: I think I will add both methods and allow the implementor to decide which to use. Personally, I think it makes sense to define a close() method on the object to be closed itself. However, this means the implementor needs to have a `name` parameter associated with their graph, which I don't want to force them to do.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/util/DefaultGraphManager.java
dpitera: >  "closed" does that mean it should still be in the available set of graphs? 

I would imagine not for sure. I didn't have access to the graphName for reasons described [here](https://github.com/apache/tinkerpop/pull/569#discussion_r106424327). 

I think going forward the best thing to do is to make the `DefaultGraphManager` implement an actual `closeGraph(String graphName)` and `noop` `closeGraph(Graph graph)`. I would like to leave the interface `closeGraph(Graph graph)` though

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
spmallette: oh, i see what you're saying. i think that if you add both, you should make a default implementation on the one with the String argument to use `getGraph()` and then pass that to the `closeGraph(Graph graph)` method.

was there a reason you went with `closeGraph()` as opposed to `removeGraph()`. it seems like the `GraphManager` is about managing `Graph` instances that it holds and "remove" is a better term for the behavior `GraphManager` is supposed to be doing there.  @pluradj i think you'd suggested this addition - perhaps you could chime in on this.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/util/DefaultGraphManager.java
```diff
@@ -0,0 +1,216 @@
...
+        final Graph newGraph = supplier.get();
+        addGraph(graphName, newGraph);
+        return newGraph;
+    }
+
+    /**
+     * Use {@link String} graphName to retrieve {@link Graph} from the {@link Map} tracking
+     * {@link Graph} references, close the {@link Graph}, and remove it from the {@link Map}.
+     */
+    public final void closeGraph(final String graphName) {
```
pluradj: @spmallette I had suggested `removeGraph(String)` for this as the reciprocal for `addGraph(String, Graph)`. I wouldn't expect remove graph to close the graph since add graph didn't open it.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/util/DefaultGraphManager.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    /**
+     * Close {@link Graph} object.
+     */
+    public final void closeGraph(final Graph graph) {
+        try {
+            graph.close();
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
pluradj: Throwing a `RuntimeException` here would bomb out the server? I don't see why I'd call this method on `DefaultGraphManager` if it doesn't do anything else besides call `graph.close()`.

dpitera:
@spmallette Updated with your suggestions; only thing i have not addressed is `closeGraph()` versus `removeGraph()`.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/util/DefaultGraphManager.java
dpitera: I agree-- this implementation of a graphManager makes sense for users t call `closeGraph(String)`. Suggestions on what to do in the catch block?

spmallette: I'd suggest you change the signature:

```java
public Graph removeGraph(String graphName)
```

Then the caller can make it's own choice as to what to do with the removed instance.

dpitera: And make the `DefaultGraphManager` not call `graph.close()`?

dpitera: OK the more I think about the nuances here, I think you are right-- there should exist a `removeGraph(String)` and `closeGraph(Graph)`. The former is the antithesis to `addGraph()` while the latter is the antithesis to `openGraph()`. 

Even though the `DefaultGraphManager.closeGraph()` will most likely not be called, I think it still makes sense as an example implementation to have it call `graph.close()`. Having said that, how shall I handle the `Exception`?

pluradj: I'd think it would look like `void closeGraph(final Graph graph) throws Exception`, basically repeating the method on [Graph.close()](https://github.com/apache/tinkerpop/blob/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/Graph.java#L303).

spmallette: The more I think about it the less I really understand the use case for `close/openGraph()` when I think about how the `GraphManager` would likely be used. Need to stop thinking about this for today - perhaps i'll think more clearly on it tomorrow and I can formulate a reasonable question or two about it.

dpitera: If it will help make more sense to you, I can show you my WIP implementation for JanusGraph.  I hadn't brought it up to avoid clouding decisions made here based on the needs of a specific implementation.

dpitera: At the expense of making things more confusing, imagine an implementor of Tinkerpop was building a graph database, called `G` for ease.

G promises horizontal scaling and guarantees linearizablity. 

Therefore, G's graph objects must be consistent across multiple nodes. 

Therefore, G will need to be able to instantiate its graph objects, and "close" its graph object's in a custom way-- that is-- one that guarantees linearizable consistency across nodes.

dpitera:
@spmallette No rush, just checking in. Have you thought about this anymore? I had addressed some of your questions [here](https://github.com/apache/tinkerpop/pull/569#discussion_r106475154) and I have pushed code addressing feedback.

spmallette:
I've been really focused on some other things the last few days. Expected to focus on pull requests today (which are stacking up). Of course, the one I started with has me all hung up. Hoping to make my rounds to all of them today, including this one.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
```diff
@@ -21,139 +21,98 @@
...
-                graphs.put(e.getKey(), newGraph);
-                logger.info("Graph [{}] was successfully configured via [{}].", e.getKey(), e.getValue());
-            } catch (RuntimeException re) {
-                logger.warn(String.format("Graph [%s] configured at [%s] could not be instantiated and will not be available in Gremlin Server.  GraphFactory message: %s",
-                        e.getKey(), e.getValue(), re.getMessage()), re);
-                if (re.getCause() != null) logger.debug("GraphFactory exception", re.getCause());
-            }
-        });
-    }
+    public Map<String, Graph> getGraphs();
```
spmallette: So, now we have `getGraphs()` which returns the `Map` of graph names and `Graph` instances but then we also expose get/put/remove operations that just manipulate that same `Map`. That seems like it needs more thought.  In other words, why do:

```java
manager.addGraph("graph", new TinkerGraph());
manager.getGraph("graph");
```

if I can also do:

```java
manager.getGraphs().put("graph",new TinkerGraph());
manager.getGraphs().get("graph");
```

By offering two approaches to do the same thing we open the possibility to error. If a `GraphManager` implementation adds logic to `addGraph()` (or related method) then that logic can simply be bypassed by directly manipulating the `Map` via `getGraphs()`.  

Seems like we would want to use `GraphManager` to enforce the logic of the implementation, thus `getGraphs()` should go away? Perhaps to lessen the breaking change for 3.2.x it is simply deprecated with some additional javadoc that states that the method is expected to return an umodifiable `Map`? replace that method with `Set<String> getGraphNames()` so you could iterate over that to get all the `Graph` instances or maybe have `GraphManager` implement `Iterable` so you could `foreach` and get an `iterator()`? and then finally replace all internal usage of `getGraphs()` so that it can safely be removed on the master branch.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/util/DefaultGraphManager.java
```diff
@@ -0,0 +1,210 @@
...
+        final Graph newGraph = supplier.get();
+        addGraph(graphName, newGraph);
+        return newGraph;
+    }
+
+    /**
+     * Remove {@link Graph} corresponding to {@link String} graphName from {@link Map}
+     * tracking graph references.
+     */
+    public final void removeGraph(final String graphName) {
```
spmallette: Looking at this signature here, I think it would be better to mimic `Map.remove(k)` and return `Graph` as in:

```java
public final Graph removeGraph(final String graphName) 
```

Then a caller would then just do:

```java
manager.removeGraph("graph").close();
```

or whatever they would do with their dynamically added graph.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
```diff
@@ -21,139 +21,98 @@
...
-    }
+    public void commit(final Set<String> graphSourceNamesToCloseTxOn);
+
+    /**
+     * Implementation that allows for custom graph-opening implementations; if the {@link Map}
+     * tracking graph references has a {@link Graph} object corresponding to the {@link String} graphName,
+     * then we return that {@link Graph}-- otherwise, we use the custom {@link Supplier} to instantiate a
+     * a new {@link Graph}, add it to the {@link Map} tracking graph references, and return said {@link Graph}.
+     */
+    public Graph openGraph(final String graphName, final Supplier<Graph> supplier);
```
spmallette: I'm feeling like the second argument here could offer a little more flexibility if it were `BiSupplier<String,Graph>` where the first argument was the `graphName` passed back in on itself (or ultimately whatever name the `GraphManager` implementation used to key the created `Graph` instance. That could be a useful configuration parameter to something like Neo4j where it could use the graph name to uniquely create a data directory for example.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
```diff
@@ -21,139 +21,98 @@
...
+
+    /**
+     * Implementation that allows for custom graph-closing implementations; it is up to the implementor
+     * to decide if this method should also remove the {@link Graph} graph from the {@link Map}
+     * tracking {@link Graph} references (and how it would do so).
+     */
+    public void closeGraph(final Graph graph) throws Exception;
```
spmallette: I've made a number of comments about `closeGraph()` so another one shouldn't be surprising :smile: - I can sorta understand the requirement here as the symmetrical method to `openGraph()` and the idea that the component interacting with the manager won't keep track of the graph names and may only have the `Graph` instance. So, I'm that far along....

So these would be the current thoughts/concerns:

1. The javadoc makes this implementation have a pretty wide open scope leaving it up to the implementation to choose whether or not the `Graph` is removed from the "registry" (i.e. control of the `GraphManager`) - you mention `Map` specifically but the registry might not use that data structure specifically I suppose.  Under what circumstance would an interacting component call `closeGraph()` but not want the `Graph` removed from the registry? To leave it there would allow callers to Gremlin Server to get graphs that were in a "closed" state which doesn't seem right. 
2. Piggy-backing on that last point, if you go down to `DefaultGraphManager`, I think that it should look to remove the instance from the registry. Again, I'm lost for a scenario where you would want a closed `Graph` instance available to Gremlin Server requests.
3. I still think this method is really just an overload of `removeGraph(String)` and should be named as such: `Graph removeGraph(Graph)`. A bit weird as you already have the `Graph` instance and are just returning it back, but it would let the API be more fluent. I think @pluradj implied the same thing in a different comment.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
dpitera: Sounds good to me.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/util/DefaultGraphManager.java
dpitera: Sounds good to me

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
dpitera: >  Under what circumstance would an interacting component call closeGraph() but not want the Graph removed from the registry? 

I mean sometimes things that _exist but are marked as "not usable"_ mean something other than _this doesn't exist. If this concept were important to a specific implementation I am sure they  would have an `graph.unclose()` and then their `openGraph(String graphName)` would `unclose()` the graph that already exists in the reference tracker

> Piggy-backing on that last point, if you go down to DefaultGraphManager, I think that it should look to remove the instance from the registry.

I agree, but if we do not have the String representation of the graphName here then how can we do that?

> I still think this method is really just an overload of removeGraph(String) and should be named as such

I can live with that

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
dpitera: > Seems like we would want to use GraphManager to enforce the logic of the implementation, thus getGraphs() should go away?

Agreed.

I am confused by your suggestion to "lessen the breaking change", but I'll give a stab at what I think you meant, and we can reiterate

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
spmallette: > I agree, but if we do not have the String representation of the graphName here then how can we do that?

I guess you'd have to iterate the `Graph` values in `GraphManager` and compare for equality. I suppose you would be passing the same instance in to `closeGraph()` as is held there? 

Sorry I'm struggling to understand the need for this method. Perhaps you will need to link me to your wip for JanusGraph so that I can better see where you are coming from.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
spmallette: ok - thanks for being patient on this. i knew this was going to be a gnarly pull request. it's one of the reasons i've sorta left this issue alone.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
dpitera: > Sorry I'm struggling to understand the need for this method.

Honestly I guess it's not necessary as you can do the same thing with `removeGraph(String)` by first retrieving the `Graph`. I just thought it made sense from a design perspective to `create` using a `String` and `destroy` the object itself.

dpitera: This is my current WIP implementation of [closeGraph](https://github.com/dpitera/janusgraph/blob/testing/janusgraph-core/src/main/java/org/janusgraph/graphdb/management/JanusGraphManager.java#L190) and [openGraph](https://github.com/dpitera/janusgraph/blob/testing/janusgraph-core/src/main/java/org/janusgraph/graphdb/management/JanusGraphManager.java#L145)

dpitera:
Pushed changes according to previous round of comments. Please re-review whenever you get a chance.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
```diff
@@ -21,139 +21,101 @@
...
- * initialization scripts construct them.
- */
-public final class GraphManager {
-    private static final Logger logger = LoggerFactory.getLogger(GremlinServer.class);
-
-    private final Map<String, Graph> graphs = new ConcurrentHashMap<>();
-    private final Map<String, TraversalSource> traversalSources = new ConcurrentHashMap<>();
+import java.util.function.Function;
 
+public interface GraphManager {
```
spmallette: Since we don't really have any dev docs for how to implement a `GraphManager` could you please add some javadoc here to describe what this interface is for and why someone would implement it?

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/util/DefaultGraphManagerTest.java
```diff
@@ -0,0 +1,109 @@
...
+import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * @author Stephen Mallette (http://stephen.genoprime.com)
+ */
+public class DefaultGraphManagerTest {
```
spmallette: Would you mind adding appropriate tests for the code in `openGraph()`? Not sure if there's other methods/code branches that require testing or not, but perhaps give that a review now that we've better settled the API.

spmallette:
I'm going to do some additional testing with my own `GraphManager` implementation just to see how it all feels. Anyway, I think this is firming up pretty nicely - getting pretty close to being able to call for a final review.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
```diff
@@ -21,139 +21,101 @@
...
-     * initialization scripts.
+     * Get a list of the {@link TraversalSource} instances and their binding names
      *
      * @return a {@link Map} where the key is the name of the {@link TraversalSource} and the value is the
      *         {@link TraversalSource} itself
      */
-    public Map<String, TraversalSource> getTraversalSources() {
-        return traversalSources;
-    }
+    public Map<String, TraversalSource> getTraversalSources();
```
spmallette: It occurred to me while implementing `GraphManager` that `TraversalSource` objects should also be managed by the interface. At a minimum, that would mean similar methods to what we have for `Graph`.

dpitera: ha. I haven't quite wrapped my head around the expected use of `traversalSources` so I honestly can speak nothing to this... If you expect that to be the case, I prefer to open a separate issue and make that a separate PR as I can only speak to the graph managing cause right now

spmallette: I think you just need to deprecate `getTraversalSources()` and include `removeTraversalSource(String tsName)` and `getTraversalSourceNames()`. That should do it. I could do it after merge, but this body of work would be more complete if it were included as part of this PR.

spmallette:
I've finished playing around with my dummy implementation. Just one weird thing that wasn't easy to see without actually working with the API - `addTraversalSource()` and `addGraph()` aren't really "adds" - they are "puts" as  calling them more than once will act as a "replace" operation if the key already exists. I think we want the behavior to generally be a "replace" and thus "putTraversalSource()" and "putGraph()" seem like better names. I suppose if an implementation wanted to limit that ability to only "adds" then the naming with "put" still makes sense as we can see with the use of `Map.put()` in unmodifiable collections.

I see some other odds/ends that could use a tweak or two, but I'll just handle those things after merging as I think you have the meat of this PR in place now once you address those other comments I've made.  Please confirm that you've done a final build with integration tests after all these changes to make sure it's all solid. At that point, I expect we can get committers to review/vote.

This has developed into a really nice pull request. Your effort on it is appreciated.

dpitera:
>  I think we want the behavior to generally be a "replace" and thus "putTraversalSource()" and "putGraph()" seem like better names. 

+1

> This has developed into a really nice pull request. Your effort on it is appreciated.

Thank you sir. Your outstanding review is also very much appreciated.


I just pushed new code with all requested changes; once the TravisCI and Docker IT tests pass, I will post again for hopefully the final review.

dpitera:
IT tests passed! 
```
[INFO] Reactor Summary:
[INFO]
[INFO] Apache TinkerPop .................................. SUCCESS [1:31.220s]
[INFO] Apache TinkerPop :: Gremlin Shaded ................ SUCCESS [12.183s]
[INFO] Apache TinkerPop :: Gremlin Core .................. SUCCESS [1:15.039s]
[INFO] Apache TinkerPop :: Gremlin Test .................. SUCCESS [9.752s]
[INFO] Apache TinkerPop :: Gremlin Groovy ................ SUCCESS [2:50.074s]
[INFO] Apache TinkerPop :: Gremlin Groovy Test ........... SUCCESS [5.952s]
[INFO] Apache TinkerPop :: TinkerGraph Gremlin ........... SUCCESS [4:44.315s]
[INFO] Apache TinkerPop :: Gremlin Benchmark ............. SUCCESS [10.434s]
[INFO] Apache TinkerPop :: Gremlin Driver ................ SUCCESS [15.460s]
[INFO] Apache TinkerPop :: Neo4j Gremlin ................. SUCCESS [4.899s]
[INFO] Apache TinkerPop :: Gremlin Server ................ SUCCESS [16:15.318s]
[INFO] Apache TinkerPop :: Gremlin Python ................ SUCCESS [19.874s]
[INFO] Apache TinkerPop :: Hadoop Gremlin ................ SUCCESS [11:05.683s]
[INFO] Apache TinkerPop :: Spark Gremlin ................. SUCCESS [22:49.909s]
[INFO] Apache TinkerPop :: Giraph Gremlin ................ SUCCESS [2:21:03.271s]
[INFO] Apache TinkerPop :: Gremlin Console ............... SUCCESS [2:56.000s]
[INFO] Apache TinkerPop :: Gremlin Archetype ............. SUCCESS [0.175s]
[INFO] Apache TinkerPop :: Archetype - TinkerGraph ....... SUCCESS [28.255s]
[INFO] Apache TinkerPop :: Archetype - Server ............ SUCCESS [13.480s]
[INFO] BUILD SUCCESS
[INFO] Total time: 3:26:32.403s
[INFO] Finished at: Thu Mar 30 22:40:41 UTC 2017
[INFO] Final Memory: 103M/771M
```

spmallette:
Ran Gremlin Server integration tests locally to success - VOTE +1

pluradj:
Using the GitHub review seems to have made my vote disappear...
Successful `docker/build.sh -t -i -n`. Verified docs and javadocs. Solid work on this @dpitera.

VOTE: +1

dkuppitz:
Well done.

VOTE: +1

dpitera:
Has this been missed by the ASF bot? I see 3 `VOTE: +1`s

pluradj:
@spmallette isn't a bot ;)

pluradj:
deferring the merge to you @spmallette since you had taken the lead on the review, but i'm happy to handle it

dkuppitz:
Feel free to merge this PR @pluradj. The @spmallette bot is currently being rebooted / is on vacation ;).

spmallette:
I'll work on this one today as soon assuming I get caught up on all missed emails and such. I sense the merge to master won't be trivial, but we'll see. Anyway, this one should close out soon.

spmallette:
This is merged - no problems getting it to master.

PR #583, opened 2017-03-24
TINKERPOP-1657 Provide abstraction to easily allow different HttpAuth schemes

krlohnes:
Abstracting over the http authentication allows for easy extensibility
for users/implementors to provide their own classes for http auth beyond
basic auth. The general issue is that there is a fixed overhead to
hashing passwords securely. This change allows for implementing things
like HMAC token auth and plugging them in easily to the gremlin server.

### DIFF
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java
@@ -384,7 +384,23 @@ public class Settings {
          * used to load the implementation from the classpath. Defaults to {@link AllowAllAuthenticator} when
          * not specified.
          */
+        public String authenticator = AllowAllAuthenticator.class.getName();
+
+        /**
+         * The fully qualified class name of the {@link Authenticator} implementation. This class name will be
+         * used to load the implementation from the classpath. Defaults to {@link AllowAllAuthenticator} when
+         * not specified.
+         * @deprecated As of release 3.2.5, replaced by {@link authenticator}.
+         */
+        @Deprecated
         public String className = AllowAllAuthenticator.class.getName();
+
+        /**
+         * The fully qualified class name of the {@link HttpAuthenticationHandler} implementation.
+         * This class name will be used to load the implementation from the classpath.
+         * Defaults to null when not specified.
+         */
+        public String authenticationHandler = null;
 
         /**
          * A {@link Map} containing {@link Authenticator} specific configurations. Consult the
@@ -424,7 +440,7 @@ public class Settings {
          * contain an X.509 certificate chain in PEM format. {@code null} uses the system default.
          */
         public String trustCertChainFile = null;
-        
+
         /**
          * Require client certificate authentication
          */
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java
@@ -384,7 +384,7 @@ public class Settings {
          * used to load the implementation from the classpath. Defaults to {@link AllowAllAuthenticator} when
          * not specified.
          */
-        public String authenticator = AllowAllAuthenticator.class.getName();
+        public String authenticator = null;
 
         /**
          * The fully qualified class name of the {@link Authenticator} implementation. This class name will be
@@ -396,7 +396,7 @@ public class Settings {
         public String className = AllowAllAuthenticator.class.getName();
 
         /**
-         * The fully qualified class name of the {@link HttpAuthenticationHandler} implementation.
+         * The fully qualified class name of the {@link AbstractAuthenticationHandler} implementation.
          * This class name will be used to load the implementation from the classpath.
          * Defaults to null when not specified.
          */
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,10 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.2.5 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+
+* Deprecated `authentication.className` setting in favor of using `authentication.authenticator`
+* Added `authentication.authenticationHandler` setting
+* Added abstraction to authorization to allow users to plug in their own `AbstractAuthorizationHandler` implementations
 * Fixed a `NullPointerException` bug in `B_LP_O_S_SE_SL_Traverser`.
 * `PathRetractionStrategy` now uses the marker-model to reduce recursive lookups of invalidating steps.
 * `ProfileStrategy` now uses the marker-model to reduce recursive lookups of `ProfileSideEffectStep`.
--- a/docs/src/reference/gremlin-applications.asciidoc
+++ b/docs/src/reference/gremlin-applications.asciidoc
@@ -1072,7 +1072,8 @@ The following table describes the various configuration options that Gremlin Ser
 [width="100%",cols="3,10,^2",options="header"]
 |=========================================================
 |Key |Description |Default
-|authentication.className |The fully qualified classname of an `Authenticator` implementation to use.  If this setting is not present, then authentication is effectively disabled. |`AllowAllAuthenticator`
+|authentication.authenticator |The fully qualified classname of an `Authenticator` implementation to use.  If this setting is not present, then authentication is effectively disabled. |`AllowAllAuthenticator`
+|authentication.authenticationHandler | The fully qualified classname of an `AbstractAuthenticationHandler` implementation to use. If this setting is not present, but the `authentication.authenticator` is, it will use that authenticator with the default `AbstractAuthenticationHandler` implementation for the specified `Channelizer` |_none_
 |authentication.config |A `Map` of configuration settings to be passes to the `Authenticator` when it is constructed.  The settings available are dependent on the implementation. |_none_
 |channelizer |The fully qualified classname of the `Channelizer` implementation to use.  A `Channelizer` is a "channel initializer" which Gremlin Server uses to define the type of processing pipeline to use.  By allowing different `Channelizer` implementations, Gremlin Server can support different communication protocols (e.g. Websockets, Java NIO, etc.). |`WebSocketChannelizer`
 |graphs |A `Map` of `Graph` configuration files where the key of the `Map` becomes the name to which the `Graph` will be bound and the value is the file name of a `Graph` configuration file. |_none_
@@ -1194,7 +1195,7 @@ graph database, which must be provided to it as part of the configuration.
 
 [source,yaml]
 authentication: {
-  className: org.apache.tinkerpop.gremlin.server.auth.SimpleAuthenticator,
+  authenticator: org.apache.tinkerpop.gremlin.server.auth.SimpleAuthenticator,
   config: {
     credentialsDb: conf/tinkergraph-credentials.properties}}
 
--- a/docs/src/upgrade/release-3.2.x-incubating.asciidoc
+++ b/docs/src/upgrade/release-3.2.x-incubating.asciidoc
@@ -32,6 +32,16 @@ Please see the link:https://github.com/apache/tinkerpop/blob/3.2.5/CHANGELOG.asc
 Upgrading for Users
 ~~~~~~~~~~~~~~~~~~~
 
+Authentication Configuration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The server settings previously used `authentication.className` to set an authenticator for the the two provided
+authentication handler and channelizer classes to use. This has been deprecated in favor of `authentication.authenticator`.
+A class that extends `AbstractAuthenticationHandler` may also now be provided as `authentication.authenticationHandler`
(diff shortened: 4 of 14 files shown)


spmallette:
Hi and thanks for contributing. In general, I like the concept of this pull request, but I have some thoughts to share to get us both on the same page. I'd thought about doing this when we first decided to get authentication in place. I opted not to do it for at least a couple reasons that I can remember:

1. The `Channelizer` abstraction might be enough of an abstraction. It's not complex by any means - the core of `HttpChannelizer` is 20 lines of easily readable code. If someone doesn't want basic auth, then just write a new `Channelizer` and done. Personally, I'm not sure I want to see us expand different kinds of HTTP  auth in TinkerPop so `HttpChannelizer` with basic auth seems good enough.
2. The configuration system required further pollution of the root of the configuration file. It bugs me that we don't have `Channelizer` specific configurations - e.g. you had to add `httpAuthHandlerClassName` to `Settings` and that value has nothing to do with anything other than `HttpChannelizer`. If you were configured for web sockets it would just be ignored. Not saying that we don't have that issue at hand now (as I think we do), but I didn't want to continue to proliferate it.

Anyway, I'd had it in my mind that solving 2 would make 1 more palatable as something to do, but 2 hasn't come around yet and unless there is a neat way to do it, it will come in as a breaking change to the Gremlin Server configuration file. 

So that's the basic history on this direction for auth configuration - thoughts?

krlohnes:
You make some very good points.

> The Channelizer abstraction might be enough of an abstraction.

I'd thought about that a bit, but in general, I'd rather not have to rewrite (or worse, essentially copy and paste) a piece of OS code that works almost exactly like I need it to, but simply doesn't have an extension point where there could easily be one. 

>Personally, I'm not sure I want to see us expand different kinds of HTTP auth in TinkerPop

I'm not suggesting adding additional "out of the box" authorization schemes, which is why I didn't commit one. I'm really only suggesting making Tinkerpop more easily extensible so users can code their own that fits with their orgs current auth schemes/security requirements. 

>  If you were configured for web sockets it would just be ignored

I think this abstraction could be applied fairly easily to the `WebSocketsChannelizer`. There's already the `SaslAuthenticationHandler` there. I think that with a similar change to the ternary operator in the `SaslAuthenicationHandler.init` as in the HttpChannelizer.init, you could change the abstract `HttpAuthenticationHandler` class to an `AuthenticationHandler` and have the `SaslAuthenticationHandler` extend that. I could then change the config name from `httpAuthHandlerClassName` to `authHandlerClassName`. 

> The configuration system required further pollution of the root of the configuration file

Well, it's not the _root_ of the configuration file. It's under `authentication : { ... }`. 

I do agree that Channelizer config would be nice to have and may follow naturally from the idea of this PR, but I think this PR is a small, iterative change that _doesn't_ break anything. 

I think with the change around the `WebSocketsChannelizer` (included in a new commit), the story around this becomes better. Let me know what you think.

spmallette:
> I think this abstraction could be applied fairly easily to the `WebSocketsChannelizer`

You're right.

> Well, it's not the root of the configuration file. It's under `authentication : { ... }`.

sorry - i missed that and all the better that it can be generally applied to the `WebSocketChannelizer` as i agreed above. of course, given both of the above points, something doesn't feel right. Let me try to explain. It seems like we have a pluggable authentication schema already in SASL and now this pull request suggest that we make SASL pluggable. I think we just need to make the `HttpChannelizer` use SASL. Then if someone wants to do HMAC (or whatever custom security theme) they implement through SASL.  Does that make sense?

> I think this PR is a small, iterative change that doesn't break anything.

agreed - this PR is well scoped as it is. i didn't mean to suggest expanding it, especially now that i understand that the change isn't at the root of the yaml.

spmallette:
@krlohnes I think I'm going to retract what I said about using SASL for this case.  SASL mechanisms might not suit every use case in which case it might not be the best way to go about implementing the authentication scheme you set out to have. So, I think you're basically on the right path with this PR. I'll review based on that.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java
```diff
@@ -387,6 +387,18 @@ public SerializerSettings() {}
         public String className = AllowAllAuthenticator.class.getName();
 
         /**
+         * Enable audit logging of authenticated users and gremlin evaluation requests.
+         */
+        public boolean enableAuditLog = false;
```
spmallette: What is the purpose of this configuration option? It doesn't appear to be used in any way.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java
```diff
@@ -387,6 +387,18 @@ public SerializerSettings() {}
...
+
+        /**
+         * The fully qualified class name of the {@link AuthenticationHandler} implementation.
+         * This class name will be used to load the implementation from the classpath.
+         * Defaults to null when not specified.
+         */
+        public String handlerClassName = null;
```
spmallette: I tried to stay consistent in naming for classes in the config by going with `className` where the key was representative of an object in a `Map` or `List`, but now it kinda bites us as `authentication.className` is used for the `Authenticator` when it now really should probably refer to the handler since it is now configurable. 

Here's an idea:

1. Rename `handlerClassName` to `authenticationHandler`
2. Overload `className` with `authenticator` - if `authenticator` is present then ignore any value in `className`. Update all the yaml files to use `authenticator`. [Deprecate](http://tinkerpop.apache.org/docs/current/dev/developer/#_deprecation) `AuthenticationSettings.className`.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/AuthenticationHandler.java
```diff
@@ -0,0 +1,52 @@
...
+import static io.netty.handler.codec.http.HttpVersion.HTTP_1_1;
+import static org.apache.tinkerpop.gremlin.groovy.plugin.dsl.credential.CredentialGraphTokens.PROPERTY_PASSWORD;
+import static org.apache.tinkerpop.gremlin.groovy.plugin.dsl.credential.CredentialGraphTokens.PROPERTY_USERNAME;
+
+/**
+ * Provides an abstraction point to allow for http auth schemes beyond basic auth.
+ */
+import io.netty.channel.ChannelInboundHandlerAdapter;
+
+public abstract class AuthenticationHandler extends ChannelInboundHandlerAdapter {
```
spmallette: We tend to prefix abstract` classes with "Abstract" as a convention. Also, please move the javadoc down below the `import` statement. You might also want to clean up the list of unused imports a bit.

spmallette:
In addition to the other individual comments I had on the code itself, please also amend this PR to handle documentation requirements:

1. Reference documentation should reflect these new configuration options and extension points.
2. Upgrade docs could use an entry in the  "users" section.
3. Update CHANGELOG as needed for 3.2.x

Finally, did all tests pass with integration tests? Note that using docker can make this process easier if you don't have your environment setup to do the full build - `docker/build.sh -t -i -n`

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java
krlohnes: That sounds good to me.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java
krlohnes: Ah, I'll delete that.

robertdale:
One thing I haven't had much time to think about here is how to propagate ssl client authentication up (into an authentication handler). Right now it's at the channel level only.  That means it is independent of other authentication mechanisms for better or worse.  Permutations to consider:

- needClientAuth REQUIRE, no auth handler: ssl client auth required but no logging of the client
- needClientAuth REQUIRE, +auth handler: both are always required
- needClientAuth OPTIONAL, no auth handler: same as NO SECURITY!
- needClientAuth OPTIONAL, +auth handler: auth handler always required

So one option is to create ssl client auth's own Channel and Auth Handlers thus disallowing mixing.  One of the limitations of the current system is what information is available to the Auth handler. So another option is to add whatever information is required to pass along the fact that ssl client auth worked then the auth handler could make its own determination as to if further authentication is required.  A more complex option would be to introduce authentication chaining thus allowing multiple mechanisms and some sort of requirement scheme - required, sufficient, etc.

I don't know if this PR does help or can help with any of that.   Just some food for thought...

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java
krlohnes: Fixed

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java
krlohnes: Fixed

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/AuthenticationHandler.java
krlohnes: Fixed

spmallette:
@robertdale i'm glad you're thinking at that level wrt to security aspects of Gremlin Server. My understanding of SSL beyond what I've implemented so far is pretty weak, so if you have additional ideas there that would make things better that would be great. I can't say if this PR helps with what you described. My understanding of your comment makes me feel like it doesn't help directly and that the current way to do that would be a fresh `Channelizer` implementation, but I could be missing something.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java
```diff
@@ -384,9 +384,25 @@ public SerializerSettings() {}
          * used to load the implementation from the classpath. Defaults to {@link AllowAllAuthenticator} when
          * not specified.
          */
+        public String authenticator = null;
+
+        /**
+         * The fully qualified class name of the {@link Authenticator} implementation. This class name will be
+         * used to load the implementation from the classpath. Defaults to {@link AllowAllAuthenticator} when
+         * not specified.
+         * Deprecated in favor of {@link authenticator}.
```
spmallette: Note that our deprecation form in javadoc typically looks like this:

```java
@deprecated As of release 3.2.5, replaced by {@link authenticator}.
```

krlohnes:
I got the integration tests passing, I'm hoping to finish up the documentation changes today. 

```
[INFO] Reactor Summary:
[INFO]
[INFO] Apache TinkerPop .................................. SUCCESS [59.115s]
[INFO] Apache TinkerPop :: Gremlin Shaded ................ SUCCESS [9.632s]
[INFO] Apache TinkerPop :: Gremlin Core .................. SUCCESS [1:13.945s]
[INFO] Apache TinkerPop :: Gremlin Test .................. SUCCESS [7.673s]
[INFO] Apache TinkerPop :: Gremlin Groovy ................ SUCCESS [2:20.814s]
[INFO] Apache TinkerPop :: Gremlin Groovy Test ........... SUCCESS [6.424s]
[INFO] Apache TinkerPop :: TinkerGraph Gremlin ........... SUCCESS [4:42.573s]
[INFO] Apache TinkerPop :: Gremlin Benchmark ............. SUCCESS [11.786s]
[INFO] Apache TinkerPop :: Gremlin Driver ................ SUCCESS [14.719s]
[INFO] Apache TinkerPop :: Neo4j Gremlin ................. SUCCESS [23:11.498s]
[INFO] Apache TinkerPop :: Gremlin Server ................ SUCCESS [16:58.561s]
[INFO] Apache TinkerPop :: Gremlin Python ................ SUCCESS [15.162s]
[INFO] Apache TinkerPop :: Hadoop Gremlin ................ SUCCESS [9:23.457s]
[INFO] Apache TinkerPop :: Spark Gremlin ................. SUCCESS [20:19.311s]
[INFO] Apache TinkerPop :: Giraph Gremlin ................ SUCCESS [2:20:57.434s]
[INFO] Apache TinkerPop :: Gremlin Console ............... SUCCESS [2:45.687s]
[INFO] Apache TinkerPop :: Gremlin Archetype ............. SUCCESS [0.115s]
[INFO] Apache TinkerPop :: Archetype - TinkerGraph ....... SUCCESS [13.418s]
[INFO] Apache TinkerPop :: Archetype - Server ............ SUCCESS [11.647s]
[INFO] BUILD SUCCESS
[INFO] Total time: 3:44:24.099s
[INFO] Finished at: Fri Mar 31 22:24:11 UTC 2017
[INFO] Final Memory: 107M/692M
```

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java
krlohnes: Fixed

krlohnes:
@spmallette I've updated the docs to include the changes made in this PR. Let me know if I've missed anything / if you'd like me to reword anything.

spmallette:
Can you please search for references to `authentication.className` and replace those? I don't see any of the actual yaml files updated in the PR.

krlohnes:
Sorry about that. Fixed.

spmallette:

VOTE +1

pluradj:
Did a successful `docker/build.sh -t -i -n`. Reviewed the code, docs, and test case. Solid PR @krlohnes 

VOTE: +1

dkuppitz:
VOTE: +1

PR #584, opened 2017-03-24
TINKERPOP-1095 Gremlin Server performance improvement with SimpleScriptContext

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1095

The issue is about creating a custom `ScriptContext`  for improved usability purposes but it's since morphed to an important performance improvement for the `GremlinScriptEngine` and thus also Gremlin Server.  The change uses the custom `ScriptContext`, named `GremlinScriptContext`, to  prevent usage of the `SimpleScriptContext` which creates `ByteBuffer` instances on every call to `eval()` and those instances were never used making them immediately available to GC. 

For Gremlin Server it seems that it has been cycling GC way more than it should have been. Here's GC activity before my changes:
 
GC Before

and GC activity after my changes:

GC After

Obviously the amount of memory allocated was greatly reduced in the process. The number of `byte[]` instances was allocated held 2.91 GB prior to the changes:

Allocation Before

and for the same workload after the changes just 638 MB:

Allocation After

I can't think of anything wrong with replacing `SimpleScriptContext` and all the tests seem fine as a result with Docker so - VOTE +1


okram:
Solid work. VOTE +1.

twilmes:
Great improvements, tests were successful and code looks good.

VOTE: +1

PR #579, opened 2017-03-23
TINKERPOP-1654 Bumped to Jackson 2.8.7

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1654

Did a build with `mvn clean install && mvn verify -pl gremlin-server -DskipIntegrationTests=false` but expect to do a full build with docker for all branches this merges too after vote. 

VOTE +1

PR #585, opened 2017-03-27
TINEKRPOP-1654 tp32 Bumped to Jackson 2.8.7

spmallette:
This is the follow on to #579 - i don't think it needs an specific vote. There were a compilation problem that that needed resolution as we used a deprecated method that jackson has since dropped.  @newkek could you please have a look at this change - does it look ok to you?


newkek:
@spmallette I actually had a fix slightly different in mind, I just tested and it works and the tests pass, mind if I use that different solution?

spmallette:
sure - i just went with what i thought was happening already. if there's a better way, great! it's why i asked for you to take a look :smile: will you issue a PR to this branch?

newkek:
Cool, thanks - #586

PR #588, opened 2017-03-28
TINKERPOP-1659 Docker uses custom maven settings.xml

robertdale:
This lets docker use a local maven proxy for faster download.  Speeds up build time for those of us out in the boonies with slow internet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ doc/out
 docs/*.asciidoc
 ext/
 .glv
+settings.xml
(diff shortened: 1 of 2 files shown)


spmallette:
Did a basic build with docker without a `settings.xml` included and it works ok.  It's up to you but it might be worth a note in the dev docs that this is a "feature" available to folks:

http://tinkerpop.apache.org/docs/current/dev/developer/#docker-integration

VOTE +1

dkuppitz:
I recently got faster internet, so -1 on this.

Just kidding :). Looks good to me and actually doesn't change anything in the "normal" workflow.

VOTE: +1

robertdale:
They've started laying fiber in the county. Hopefully they'll bring it all the way down my road. But I don't know if that's 1 year or 5 years from now. Until then, I'm at the outer limits of DSL.

VOTE +1

PR #592, opened 2017-04-04
TINKERPOP-1010 Remove deprecated methods around Gremlin Server

spmallette:
While this PR is tied to TINKERPOP-1010 it actually covers a number of issues to include: TINKERPOP-1010, TINKERPOP-1046, TINKERPOP-1049, TINKERPOP-1296, TINKERPOP-1465, and TINKERPOP-1526.

While these changes do drop some public methods/fields/classes, the changes should really only affect a fairly small group of users - those who embed Gremlin Server or who might use it in their testing. Even so, I decided to just aim this PR at `master` and not `tp32`.

Builds to success with `docker/build.sh -t -i -n`.

VOTE +1


pluradj:
Successful build with `docker/build.sh -t -i -n`. Doc updates look good.

VOTE: +1

twilmes:
VOTE: +1

PR #594, opened 2017-04-06
TINKERPOP-1577 Building gremlin-python with python 2/3

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1577

I'd forgotten that I'd long ago made this change to the python build. It basically sets up two separate python environments with virtualenv - one for python 2.x and one for 3.x - so that we can test both versions against our code. 

VOTE +1

### DIFF
--- a/gremlin-python/pom.xml
+++ b/gremlin-python/pom.xml
@@ -359,17 +359,28 @@
                                              specifically). note that the commands to install wheel are largely for
                                              safety in case someone is using an older version of virtualenv (which
                                              doesn't install wheel by default) -->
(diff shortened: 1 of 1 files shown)


davebshow:
Should this include documentation on how to run against the different interpreters?

spmallette:
I think the existing documentation covers it, because it already discusses virtualenv which is responsible for setting up the different python environments to test against.

davebshow:
Tested using mvn clean install with .glv file. Gremlin tests ran successfully against Python 2.7 and Python 3.5 interpreter.

VOTE +1

okram:
VOTE +1.

PR #595, opened 2017-04-06
TINKERPOP-1608 XSLT for converting GraphSON

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1608

Finished up the PR originally started on #501 - this basically just takes the work @joshsh started over there, adds a test and documentation. 

Builds well with `mvn clean install -DincludeNeo4j`

VOTE +1


gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/structure/io/IoTest.java
```diff
@@ -286,6 +289,34 @@ public void shouldReadWriteSelfLoopingEdges() throws Exception {
...
+
+        @Test
+        @FeatureRequirement(featureClass = Graph.Features.EdgeFeatures.class, feature = Graph.Features.EdgeFeatures.FEATURE_ADD_EDGES)
+        @FeatureRequirement(featureClass = Graph.Features.VertexFeatures.class, feature = Graph.Features.VertexFeatures.FEATURE_ADD_VERTICES)
+        @FeatureRequirement(featureClass = VertexPropertyFeatures.class, feature = FEATURE_STRING_VALUES)
+        @FeatureRequirement(featureClass = VertexPropertyFeatures.class, feature = FEATURE_INTEGER_VALUES)
+        @FeatureRequirement(featureClass = VertexPropertyFeatures.class, feature = FEATURE_FLOAT_VALUES)
+        @FeatureRequirement(featureClass = VertexPropertyFeatures.class, feature = FEATURE_BOOLEAN_VALUES)
+        @FeatureRequirement(featureClass = VertexPropertyFeatures.class, feature = FEATURE_LONG_VALUES)
+        @FeatureRequirement(featureClass = VertexPropertyFeatures.class, feature = FEATURE_DOUBLE_VALUES)
```
pluradj: only need `String` and `Integer` vertex properties.

docs/src/reference/the-graph.asciidoc
```diff
@@ -444,6 +444,32 @@ try (final InputStream stream = new FileInputStream("tinkerpop-modern.xml")) {
 }
 ----
 
+GraphML was a supported format in TinkerPop 2.x, but there were several issues that made it inconsistent with the
+specification that were corrected for 3.x. As a result, attempting to read a GraphML file generated by 2.x with the
+3.x `GraphMLReader` will result in error. To help with this problem, an XSLT file is provided as a resource in
+`gremlin-core` which will transform 2.x GraphML to 3.x GraphML. It can be used as follows:
+
+[source,java]
```
pluradj: If this doc is Java rather than Groovy, the syntax needs to be tidied up a bit with semicolons and double quoted strings.

spmallette:
Thanks @pluradj - cleaned up the couple odds/ends you mentioned.

okram:
VOTE +1

dkuppitz:
Would be nice if the transformation would work with standard tools like `xsltproc`.`

```
daniel@cube /projects/apache/tinkerpop (TINKERPOP-1608) TP2-to-TP3 GraphML XSLT $ xsltproc ./gremlin-core/src/main/resources/tp2-to-tp3-graphml.xslt ./gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/structure/io/graphml/tinkerpop-classic-tp2.xml
runtime error: file ./gremlin-core/src/main/resources/tp2-to-tp3-graphml.xslt line 30 element copy
Cannot add an attribute node to a non-element node.
runtime error: file ./gremlin-core/src/main/resources/tp2-to-tp3-graphml.xslt line 30 element copy
Attribute nodes must be added before any child nodes to an element.
```

I tried to fix it by changing the order of some sections (more or less randomly), but that didn't fix anything. If you know how to make that work, great! If not, still

VOTE: +1

gremlin-core/src/main/resources/tp2-to-tp3-graphml.xslt
```diff
@@ -0,0 +1,59 @@
...
+            <data key="labelV">vertex</data>
+        </node>
+    </xsl:template>
+
+    <xsl:template match="graphml:edge">
+        <edge id="{@id}" source="{@source}" target="{@target}">
+            <data key="labelE">
+                <xsl:value-of select="@label"/>
+            </data>
+            <xsl:apply-templates select="node()|@*"/>
```
dkuppitz: Adding to my previous comment; If this line uses `select="node()"` only, then `xsltproc` is able to produce a "valid" output. I'm not sure if there're corner cases where something could be missing in the output, but at a first sight it looks good to me.

spmallette: I don't remember my XSLT syntax too well - what's the difference? the pattern of `"node()|@*` is used everywhere else - do you know why it's fine to use in those other spots but not here?

dkuppitz: I used XSLT many years ago and probably forgot everything I ever knew about it. Thus I have absolutely no idea why only this line is causing trouble. I was just trying out random things and commented out certain lines to find the root of the error.

spmallette: @pluradj how's your XSLT? any thoughts on this thread?

robertdale: xsltproc is complaining because the apply-templates is not immediately after a new node when trying to apply selected attributes. From what I can tell, the only attributes that should be in the <edge> element are already explicitly defined. It is not necessary to copy attributes (@*) back into the <edge> element.  That appears to be the main difference between TP2 and TP3 - that the label attribute is now a <data> element with key attribute.
Thus, the attribute selection can be removed.

`<xsl:apply-templates select="node()"/>`

PR #596, opened 2017-04-06
TINKERPOP-1443 API Checker

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1443

This is basically the continuation to #494 - didn't really change anything though @dkuppitz did get docker working. Travis seems ok with the build as well. No one seemed to have a problem with requiring maven 3.2.5 or better for our build (see the linked DISCUSS thread in JIRA), so keep in mind that if you run an earlier version it will cause trouble (only on the master branch obviously). 

I guess we will need a release of 3.3.0 before we will have any real use out of this as breaking changes are allowed between our last release on 3.2.x and our future 3.3.0. 

VOTE +1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,18 @@ jdk:
   - oraclejdk8
 sudo: required
 dist: trusty
+cache:
+  directories:
+    - $HOME/mvn-home
+
+install:
+  # install maven 3.3.9 if it is not in cache already
+  - bash .travis.install-maven.sh "3.3.9" "${HOME}/mvn-home"
+  - export M2_HOME=${HOME}/mvn-home
+  - export PATH=${HOME}/mvn-home/bin:${PATH}
+  # visual check that we have the correct maven version installed
+  - mvn -version
+
 addons:
   apt:
     packages:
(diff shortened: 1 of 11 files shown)


dkuppitz:
VOTE: +1

.travis.yml
```diff
@@ -5,6 +5,18 @@ jdk:
   - oraclejdk8
 sudo: required
 dist: trusty
+cache:
+  directories:
+    - $HOME/mvn-home
+
+install:
+  # install maven 3.3.9 if it is not in cache already
```
spmallette: Pushed fixes for both of those. Note that 3.2.5 is the actual minimum maven version required.

pluradj: then shouldn't Travis use 3.2.5?

spmallette: I'm not sure I see why it matters. The requirement is at least 3.2.5 and 3.3.9 meets that requirement. I suppose we could pin travis to the minimum version but is there any particular reason to do so?

pluradj: right, consistency and built-in testing for the stated minimum Maven level. probably no big deal either way.

pluradj:
VOTE: +1

spmallette:
@pluradj i changed it to 3.2.5 for travis - travis looks happy with the build so i'll prepare to merge today.

PR #600, opened 2017-04-09
TINKERPOP-1552 Add Gremlin-CSharp and Gremlin-DotNet

FlorianHockmann:

Organization of the C# project:
- `Gremlin.CSharp` is the C# GLV, it is nearly completely auto generated (except for the `Graph` class)
- `Gremlin.Net`: Contains the driver that is basically an extension of the [existing Gremlin.Net driver](https://github.com/FlorianHockmann/Gremlin.Net) and the structure API, including GraphSON writers and readers.
- `Gremlin.Net.Process`: Driver independent implementation + interfaces of common functionality for .NET GLVs. Allows Gremlin-CSharp to be used together with another driver.

So this organization makes it possible to use Gremlin-CSharp together with another driver and it also allows to reuse a big part of the functionality for another .NET GLV.

One design decision I took: The method names comply with .NET conventions and therefore use pascal case, so `g.V().has('name','marko').next()` becomes `g.V().Has('name','marko').Next()`. This should make it feel very natural for any C# programmer to work with Gremlin-CSharp. An added benefit is that it avoids reversed keywords in C# like `in` or `as` that would require a special treatment otherwise.

In general, Gremlin-CSharp is very close to Gremlin-Python. This also means that it is not type-safe, all steps simply expect a `params object[]`. It was very easy to generate Gremlin-CSharp this way but this should be improved in the future.

In order to support building everything from Maven, I decided to use a Maven plugin that allows building of .NET Core projects: https://github.com/kaspersorensen/dotnet-maven-plugin
This made it relatively simple to integrate the .NET projects into the regular build process.

The documentation is hugely borrowed from that of Gremlin-Python. However, I couldn't really test the generation of the documentation as I already got errors with an unmodified version using the docker image. So it would be nice if someone could check whether the documentation is generated correctly with my changes.

When you want to try out Gremlin-CSharp: You can build it on most modern operating systems, you only need a recent version of the [.NET Core SDK](https://www.microsoft.com/net/core). Alternatively, you can also use docker. I added the necessary steps to install .NET Core to the `Dockerfile`.


spmallette:
I saw your development branch in the works and was wondering how close you were to getting it ready to actually submit it. I guess it was closer than I thought. :) Just to make sure expectations are set right, it may take some time to get this reviewed and some iterations of reviews/fixes before we can really vote on it. 

One thing I'm worried about is that we don't have our GLV test framework really worked out well yet. I was hoping to get to that early in this year, but other priorities have drawn me away from that. 

In any case, thanks for doing this. Very nice to see support for a language that really doesn't work on the JVM at all, i.e. no ScriptEngine support (that I know of). I sense that will produce a whole host of new concerns to GLVs that we will need to sort out.   I'll have to dust off my .NET knowledge - it's been a while since I've done anything with that. :smile:

FlorianHockmann:
I developed the C# project before in another repository and only used the development branch in the end to merge it into TinkerPop. That's why it seemed to go so quickly 😄

Do you want to figure out the test framework before merging in new GLVs or can we merge it in and then adopt the framework later on? Maybe it's even an advantage to have a GLV ready that is completely independent of Java for developing the testing framework as it should be generally possible to use the framework with other languages when it works with a .NET based GLV.
I also tried to to reach a good code coverage, but a common testing framework for all GLVs would be really nice of course.

If you have any .NET related problems, don't hesitate to ask. But in general I think that it is much easier to work with .NET now thanks to .NET Core which allows to do basically everything from the command line like building and testing and it's OS-independent. [Rider](https://www.jetbrains.com/rider/) might also be interesting for you, especially when you work on Linux and are used to IntelliJ.

spmallette:
> Do you want to figure out the test framework before merging in new GLVs or can we merge it in and then adopt the framework later on? 

My inclination is to not be too hasty - note that the JS glv has been sitting for a while at #450 - and to use this opportunity to figure out what we need to do to properly bring both of these in.  One thing that may happen is that we merge both the PRs to a feature branch of this repo that TinkerPop can maintain/rebase/etc. You (and others) could in turn submit PRs against that until it's ready to merge to the release branches. This approach has the advantage that we can rig up the platform independent test suite around the existing project and the new GLVs.

spmallette:
@FlorianHockmann don't want you to think that this pr was forgotten. there didn't seem to be any objection to merging this pr to a local tinkerpop branch for further collaboration - so that much will likely happen in the more immediate term. longer term, i'm still thinking through options for the overall test framework and for serialization issues glvs are facing. there will be discussion on the dev list around that topic i think (though i i long ago brought up cucumber tests on the dev list as the approach for generalized testing and there were no other suggestions presented, so in a sense we have some consensus there on that avenue of thinking).

FlorianHockmann:
No problem @spmallette, I followed the thread on the dev list and I think that it is a good intermediate solution to merge this (and other GLVs like Gremlin-JavaScript) to a separate TinkerPop branch. This would make it easier to collaborate on those GLVs and on the testing framework. It may then also be a good idea to publish prerelease versions of the GLVs to get first feedback by interested users.

Regarding the testing framework itself: I never used Cucumber before but I played a bit with SpecFlow which is basically the .NET version of Cucumber and I think that it looks very promising for our use case of generating tests for the different languages based on common specifications. Unfortunately, SpecFlow [isn't yet ready for .NET Core](https://github.com/techtalk/SpecFlow/projects/2) but that shouldn't influence the decision on a testing framework for the GLVs as it is only an intermediate problem.

spmallette:
I've cherry picked this PR to this branch:

https://github.com/apache/tinkerpop/tree/tp32-glv

and made a commit or two to get it to compile against the tp32 branch. It's in there with #450 as well. This will allow others to more easily collaborate on the GLVs so that we can actually get these ready for merge to the release branches. We'll keep the JIRA issue open until tp32-glv is solid. In the mean time I think we can close this PR.

FlorianHockmann:
Thanks!

What's your stance on publishing prerelease packages of the GLVs? That would allow users to test them out so we can get some early feedback and additionally graph vendors can use them to create their own packages with custom serializers for their types and to support their own predicates. But I can of course understand if you prefer to wait with that for the testing framework.

spmallette:
does nuget have a SNAPSHOT system like java? i might not be opposed to SNAPSHOT release as we do that with java all the time and it comes with a certain kind of understanding that things will change (sometimes harshly). I'd be against other forms of "release".

spmallette:
sorry - changed this up. i created TINKERPOP-1489 branch for js work and TINKERPOP-1552 for c# work. both branches extend from tp32-glv. in this way, the GLVs are not coupled together to be forced to be merged together. they can be developed at their own pace. note that tp32-glv can house changes useful to both of these GLVs (stuff like the revised testing framework).

FlorianHockmann:
I don't know how it works with Java, but NuGet uses semantic versioning and treats packages with a suffix as pre-release packages. So something like `3.2.5-prerelease1` or `3.2.5-beta01` would work. xUnit for example uses `alpha`, `beta` and `rc` followed by a build tag: https://www.nuget.org/packages/xunit
It's only important to increment the suffix when something changed.
(For more information, see: https://docs.microsoft.com/en-us/nuget/create-packages/prerelease-packages)

spmallette:
I suppose this answers it:

https://github.com/NuGet/Home/issues/1891

It sounds like they have the feature, but I don't completely follow it. Perhaps it will make sense to you and you could clarify.

FlorianHockmann:
Ok, so for Java you can simply publish new `SNAPSHOT` packages and clients will always take the latest version?
That won't work for NuGet as it will first search for the package with the specified version in its local cache. So a user will always use the first _version_ of the package with a version as `3.2.5-SNAPSHOT` that was retrieved from nuget.org. He won't get any updates after that until we publish a package with a higher version such as `3.2.5-Z`, because he already has the package with the version `3.2.5-SNAPSHOT` in his cache.
The solution mentioned in that comment suggests that users don't explicitly specify the version number in their projects. Instead they can specify the version as `3.2.5-*` (instead of `3.2.5-beta01` for example) which means that they always get the latest pre-release version, but that still requires later packages to have a higher version number. (That solution was also specific to the now obsolete `project.json` project format and I am not sure if it still works exactly like that.)
So we definitely have to increase the version number somehow when we want to publish updates. We basically have the following options for that:

1. A fixed tag such as `SNAPSHOT` or `beta` plus a number that has to be manually incremented each time we change something (to Gremlin-DotNet or Gremlin-CSharp): `3.2.5-beta01`, `3.2.5-beta02`, ...
2. A fixed tag plus a build tag that is set during the build process (for example by Travis): `3.2.5-beta-build1234`, `3.2.5-beta-build1235`, ...

xUnit uses a combination of both and also goes through `alpha`, `beta` and `rc`.

But since there probably won't be that many changes to Gremlin-CSharp or Gremlin-DotNet my personal preference is that we simply increment a number in the suffix manually each time a change is merged into the C# GLV branch. So we could start for example with `3.2.5-beta01`.

spmallette:
i guess we have precedence for this with python and pypi. I'd forgotten that we do a form of "SNAPSHOT" that dynamically generates a timestamp based version:

https://github.com/apache/tinkerpop/blob/master/gremlin-python/src/main/jython/setup.py#L39

As a result, I'd feel better about doing that, though we'd still have to discuss it on dev. I still think there's some work to be done though before we can get to that. Hopefully I can start to sort out what those things are this week.

PR #601, opened 2017-04-13
TINKERPOP-1668: RepeatUnrollStrategy should not execute if there is a LoopStep used.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1668

If the child traversal of `RepeatStep` has a `LoopsStep`, then `RepeatUnrollStrategy` should not be allowed as it the loops will have been linearized and thus, never incremented. Added test case to verify expected working behavior.

VOTE +1.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,9 +26,10 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.2.5 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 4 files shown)


dkuppitz:
Lambda steps could do something with `it.loops()` and thus should have the same effect.

okram:
Yes, and I decided not to include that given that its rare and the only way to test would be to look for a `LambdaHolder` which would then make the strategy not apply to lots of other situations. I can add that break out of if you think its worth it.

dkuppitz:
Very rare indeed. Maybe we should just add a tiny note (and/or an example) in the repeat steps docs.

```
g.V().repeat(out().sideEffect {println it.loops()}).iterate() // don't try this at home
g.withoutStrategies(RepeatUnrollStrategy).V().repeat(out().sideEffect {println it.loops()}).iterate() // cool
```

dkuppitz:
Cool.

VOTE: +1

twilmes:
VOTE: +1

PR #605, opened 2017-04-28
TINKERPOP-1671 Propagate exception to Future instead of calling thread

tolbertam:
For [TINKERPOP-1671](https://issues.apache.org/jira/browse/TINKERPOP-1671)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/remote/RemoteConnection.java
```diff
@@ -67,7 +67,11 @@
         // default implementation for backward compatibility to 3.2.4 - this method will probably just become
         // the new submit() in 3.3.x when the deprecation is removed
         final CompletableFuture<RemoteTraversal<?, E>> promise = new CompletableFuture<>();
-        promise.complete(submit(bytecode));
+        try {
+            promise.complete(submit(bytecode));
+        } catch (Exception t) {
```
tolbertam: Chose to limit to `Exception` types instead of `Throwable` since `RuntimeException`/`Error` are probably acceptable to be thrown on calling thread since they are unexpected and likely not caused by the traversal failing.

mpollmeier:
VOTE +1

robertdale:
`docker/build.sh -i -t -n`  success
VOTE +1

spmallette:

VOTE +1

PR #606, opened 2017-05-03
Corrected copy-paste error in until step for predicates.

ml86:
Correct me if I am wrong, but I think the use of `addEmitToTraversal` in `until` step causes unintended traverser splits.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
@@ -2388,7 +2388,7 @@ public interface GraphTraversal<S, E> extends Traversal<S, E> {
      */
     public default GraphTraversal<S, E> until(final Predicate<Traverser<E>> untilPredicate) {
         this.asAdmin().getBytecode().addStep(Symbols.until, untilPredicate);
-        return RepeatStep.addEmitToTraversal(this, (Traversal.Admin<E, ?>) __.filter(untilPredicate));
+        return RepeatStep.addUntilToTraversal(this, (Traversal.Admin<E, ?>) __.filter(untilPredicate));
     }
 
     /**


okram:
Whoa!? That is a huge problem. Can you do three things for this please?

1. Can you write a test case in `RepeatTest` that shows it breaking, but with your fix, it working.
2. Can you direct your PR to `tp32/` branch, not `master/`.
3. Can you update our `upgrade.asciidoc` if need be to say that there was a "breaking" change due to a break?

spmallette:
Did this problem exist in 3.1.x? If so, perhaps it should be fixed back in tp31....

ml86:
I am not able to get the test setup running for developing the requested tests. I tried : 
```
export GREMLIN_TESTS='org.apache.tinkerpop.gremlin.process.traversal.step.branch.RepeatTest$Traversals'
mvn -Dmaven.javadoc.skip=true --projects gremlin-test test
```
But that does not executes org.apache.tinkerpop.gremlin.structure.ExceptionCoverageTest. So I am stuck and cannot proceed.

spmallette:
You don't run `gremlin-test` module. That just contains the tests themselves. You need to run a `Graph` implementation that uses `gremlin-test`, like `tinkergraph-gremlin`. I think that's your problem.

ml86:
I continued to read the code and it turned auto that my initial assumption that the traverser split was caused by the use of addEmitToTraversal was wrong. The split would occur anyway since RepeatStep::doUntil calls TraversalUtil.test() which always splits the traverser. 
This brings me to the question: Are the splits made in TraversalUtil::test, TraversalUtil::apply and TraversalUtil::applyAll really necessary? As far as I understand the code the traversal given to those functions would anyway split if necessary (e.g. a map step). But by always splitting in the 3 mentioned methods you also do it for e.g. the LambdaFilterStep which does seem unnecessary and, more importantly, prevents user from accessing/modifying the original traverser and the corresponding sack.

okram:
You need splits so the traverser remains "local" to the child branch. I think your fix for this particular problem is correct, it would be great if you could also add a test case to `RepeatTest` that uses the updated method (and show to yourself that it fails without your change). From there we can VOTE and merge your fix.

mpollmeier:
I spoke to @ml86 (we are colleagues) and we think that this error doesn't actually change the functionality. It only leads to computational overhead. But that makes it hard to write a test for. Since it's a very obvious copy/pasta error, I'd say let's just merge it as is.

okram:
It must effect semantics as `until` is different from `emit`. ?

mpollmeier:
In theory, yes. But it seems to behave fine as it is. Here's a scenario for the gremlin-groovy shell:

```
g = TinkerGraph.open()
v0 = g.addVertex("x", 1)
v1 = g.addVertex("x", 2)
v2 = g.addVertex("x", 2)
v0.addEdge("edge", v1)
v1.addEdge("edge", v2)
g.traversal().V(v0.id).repeat(out()).until(has("x", 2)).path()
```

If `until` would incorrectly behave like `emit`, I would assume that we'd see all three nodes in the path. Instead we only see v0 and v1, which is the correct behavious for repeat/until. Is our scenario missing something?

okram:
Please see https://github.com/apache/tinkerpop/pull/615 . Here is what I did:

1. Created a JIRA first.
2. Fixed the problem.
3. Created a test case that fails with the problem, but succeeds with the fix.

Note that the problem was in `until(Predicate)` which is a lambda and thus your example using `has("x",2)` isn't calling the broken method -- its calling `until(Traversal)`. Please see the test case in my PR (#615) and note the lambda usage. Without the fix, `emit()` would be called instead an an infinite loop occurs.

mpollmeier:
Thanks Marko!

PR #608, opened 2017-05-20
TINKERPOP-1618 Removed dependence on groovy for gremlin-python

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1618

Executed code generation with the gmavenplus plugin and the Groovy templating library. This approach also gets the files generated in a much earlier (and more correct) phase of the maven life cycle which may be important to other GLVs using this as a pattern. It may yet mean better build workflow options for gremlin-python.

Builds with `docker/build.sh -t -i -n` 

It occurred to me this morning that some docs likely need updating based on this pattern. I'll do that work on commit.

VOTE +1


okram:
Cool! -- VOTE +1.

dkuppitz:
VOTE: +1

PR #609, opened 2017-05-22
TINKERPOP-1675 Throw underlying unchecked exception in processNextStart

tolbertam:
For consistency with previous behavior, if an unchecked exception is set on the returned future throw that instead of the wrapped CompletionException.

For [TINKERPOP-1675](https://issues.apache.org/jira/browse/TINKERPOP-1675).


spmallette:

VOTE +1

robertdale:
VOTE +1

okram:
VOTE +1.

PR #615, opened 2017-05-26
TINKERPOP-1678: until(Predicate) is actually calling emit(Predicate)

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1678

Fixed a typo-bug in `GraphTraversal` where `until(Predicate)` was actually calling `emit(Predicate)`. A very low probability traversal and thus, it was never uncovered until now. However, fixed and added a test case that verifies that the previous behavior was wrong and the new behavior is write.

VOTE +1.

NOTE: This is the work done here https://github.com/apache/tinkerpop/pull/606, but now there is a test case.


spmallette:

VOTE +1

twilmes:
VOTE: +1

mpollmeier:
VOTE +1

PR #618, opened 2017-06-06
TINKERPOP-915 Add combined handler for Http and Websockets

krlohnes:
[TINKERPOP-915](https://issues.apache.org/jira/browse/TINKERPOP-915)

Most of this is tests. I added an integration test that goes over the
functionality of both the http and websocket channelizers using the new
`WsAndHttpChannelizer`. I added an additional test on top of that to
switch between using WebSockets and http.

The change itself leverages the existing `WebSocketChannelizer` to
provide the base pipeline setup. It has everything needed handler-wise
to service both http and ws connections. The
`WsAndHttpChannelizerHandler` then detects whether the incoming request
is a plain http connection or a WebSockets connection. If it's an http
connection, the channelizer handler swaps out the request handler
appropriately for whether or not authentication has been enabled.

### DIFF
--- a/docs/src/upgrade/release-3.2.x-incubating.asciidoc
+++ b/docs/src/upgrade/release-3.2.x-incubating.asciidoc
@@ -39,6 +39,15 @@ it has not been promoted as the primary way to add `IoRegistry` instances to ser
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-1694[TINKERPOP-1694]
 
+WsAndHttpChannelizer
+^^^^^^^^^^^^^^^^^^^^
+
+The `WsAndHttpChannelizer` has been added to allow for processing both WebSocket and Http requests on the same
+port and gremlin server. The `SaslAndHttpBasicAuthenticationHandler` has also been added to service
+authentication for both protocols in conjunction with the `SimpleAuthenticator`.
+
+See: link:https://issues.apache.org/jira/browse/TINKERPOP-915[TINKERPOP-915]
+
 Upgrading for Providers
 ~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/WebSocketHandlerUtil.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/WebSocketHandlerUtil.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.tinkerpop.gremlin.server.handler;
+
+import io.netty.handler.codec.http.HttpMessage;
+
+import static io.netty.handler.codec.http.HttpHeaders.Names.UPGRADE;
+import static io.netty.handler.codec.http.HttpHeaders.Names.CONNECTION;
+
+/**
+ * A class to handle common WebSocket operations
+ * @author Keith Lohnes lohnesk@gmail.com
+ */
+final class WebSocketHandlerUtil {
+
+    static boolean isWebSocket(final HttpMessage msg) {
+        final String connectionHeader = msg.headers().get(CONNECTION);
+        final String upgradeHeader = msg.headers().get(UPGRADE);
+        return (null != connectionHeader && connectionHeader.equalsIgnoreCase("Upgrade")) ||
+               (null != upgradeHeader && upgradeHeader.equalsIgnoreCase("WebSocket"));
+    }
+}
--- a/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerIntegrateTest.java
+++ b/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerIntegrateTest.java
@@ -56,7 +56,6 @@ import org.apache.tinkerpop.gremlin.server.op.AbstractEvalOpProcessor;
 import org.apache.tinkerpop.gremlin.server.op.standard.StandardOpProcessor;
 import org.apache.tinkerpop.gremlin.structure.Graph;
 import org.apache.tinkerpop.gremlin.structure.T;
-import org.apache.tinkerpop.gremlin.server.channel.NioChannelizer;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.apache.tinkerpop.gremlin.structure.util.detached.DetachedVertex;
 import org.apache.tinkerpop.gremlin.structure.util.empty.EmptyGraph;
@@ -169,9 +168,6 @@ public class GremlinServerIntegrateTest extends AbstractGremlinServerIntegration
             case "shouldBatchResultsByTwos":
                 settings.resultIterationBatchSize = 2;
                 break;
-            case "shouldWorkOverNioTransport":
-                settings.channelizer = NioChannelizer.class.getName();
-                break;
             case "shouldEnableSsl":
             case "shouldEnableSslButFailIfClientConnectsWithoutIt":
                 settings.ssl = new Settings.SslSettings();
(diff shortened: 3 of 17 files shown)


spmallette:
This is really great stuff. I tried doing this a few times and for whatever reason I couldn't get it to work right. Looking forward to reviewing this one.

spmallette:
I probably should have mentioned this on my last post, but I probably won't review this until next week as we are currently preparing for release and the release branches are frozen during testing.

krlohnes:
@spmallette Thanks for the heads up.

spmallette:
Sorry @krlohnes - if you're not following along we had a snag in 3.2.5 release so the release branches still aren't open. Probably won't see this PR handled until next week some time at this rate.

krlohnes:
@spmallette Yeah, I'm on the mailing list, I've been following along. No worries.

spmallette:
`GremlinServerWsAndHttpIntegrateTest.java` is a nice validation that all the REST tests work and a reasonable portion websockets tests work, but is there a nicer way to do that? Under this model, every time we add a new test we have to also implement that test method in `GremlinServerWsAndHttpIntegrateTest.java`. We're also not really testing all functionality of websockets as it only does `GremlinServerIntegrateTest`. 

anyway - any thoughts on a better testing model?

krlohnes:
I can probably figure something out with JUnit `Parameters` for the `GremlinServerIntegrateTest` and `GremlinServerHttpIntegrateTest`. I'll have a go at it.

spmallette:
Not sure how I feel about it but just throwing stuff out there to think about....Another option might be to just have a smoke test for the combined channelizer? or maybe all tests should run over your combined channelizer and we just need smoke tests for the specific ones? do you feel like smoke tests are too risky given your changes? Do we need to re-test all functionality over all the channelizers?

krlohnes:
My change can be summed up by 

"Setup the WebSocket stuff and insert the `GremlinEndpointHandler` into the `ChannelPipeline` to serve Http requests when necessary"

Changes to the `WebSocketChannelizer` and `SaslAuthenticationHandler` would likely be picked up easily enough by simply using the combined channelizer. If there was an issue with the `HttpChannelizer`, it may not be picked up by those tests.

Running all of the tests from both was mostly for my own sanity. I wanted to make sure I wasn't missing something.

If you're thinking about minimizing impact from extra tests running in the TP suite, I'd probably suggest running the combined channelizer on WS tests and both on `GremlinServerHttpIntegrateTest`.

spmallette:
ok - let me think about it a bit before you write too much code in any particular direction. we'll also give others some time to chime in in case they have an opinion.

spmallette:
The more i think about this the more I wonder if testing lots of server configurations with the same scenarios is useful.  Generally speaking the `Channelizer` isn't responsible for many of the features under test in `GremlinServerIntegrateTest` and there are other tests that are probably testing `Channelizer` capabilities that aren't represented there (maybe more true on master than tp32 - you might need a second PR when we sort this one out).

I think it would be worth identifying the tests that validate that a `Channelizer` is doing its job right and make a common set of tests that are executed over different server configurations. This would probably mean cherry picking tests from existing ones and moving them to this new test class.  

For example, `shouldBatchResultsByTwos` has nothing to do with the `Channelizer` selection really. That's an integration test for the `OpProcessor` really. On a similar note, `shouldReceiveFailureTimeOutOnScriptEval` is a integration test of the `GremlinExecutor` - such a test would not need to be executed across multiple configurations.

On the other hand, `shouldStartWithDefaultSettings` is quite generic and would be a good general smoke test. That one could be cherry picked to the new common test class and renamed to "shouldReturnListOfNumbers".  We'd probably also want to know if SSL/authentication can be enabled across all `Channelizer` implementations - that's a common feature and one controlled by the `Channelizer`.

Does that make sense to you at all?

krlohnes:
Yeah, I definitely think that makes sense, and it's likely the simplest solution to the tests here. 

I'll make sure the new test hits `shouldStartWithDefaultSettings`, ssl and auth, and keep the test that toggles between the two protocols to make sure they can both be connected to. I think those would be a reasonable check on whether or not this functionality works. I'll ping you once I pick out those tests. If anything else comes to mind, let me know, and I'm happy to add it.

spmallette:
@krlohnes have you done any performance testing to determine if the combined channelizer shows any difference to the independent channelizers? i think that would be worthwhile to know in evaluating this PR. As an added note which I didn't mention before I don't think - this PR will need reference docs, upgrade docs and changelog entries.

krlohnes:
@spmallette I did some testing between the http channelizer and the combined one.

With Basic Auth, the mean latency with 10000 requests for the combined channelizer was 148.3 ms and the standalone channelizer was 148.9 ms.

Without Basic Auth the mean latency with 10000 request for the combined channelizer was 4.8ms and the standalone channelizer was 4.9 ms

Given the results for http and that the ChannelPipeline is only manipulated for http, I didn't run any performance tests for WebSockets.

I'll write up the docs once I'm done refactoring the Channelizer tests.

spmallette:
interesting - thanks for that information.

krlohnes:
@spmallete I refactored the channelizer tests and removed some elsewhere when I thought they were appropriately replaced.  I'm ultimately not all too familiar with Java NIO. I added it as another channelizer test, but I can delete that test class and the associated code in the abstract test class as it really only runs a single test. I wasn't quite sure what the best way to go with that was. I also added in the doc changes.

docs/src/upgrade/release-3.2.x-incubating.asciidoc
```diff
@@ -39,6 +39,13 @@ it has not been promoted as the primary way to add `IoRegistry` instances to ser
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-1694[TINKERPOP-1694]
 
+WsAndHttpChannelizer
+^^^^^^^^^^^^^^^^^^^^
+
+The `WsAndHttpChannelizer` has been added to allow for processing both WebSocket and Http requests on the same
+port and gremlin server. The `SaslAndHttpBasicAuthenticationHandler` has also been added to service
+authentication for both protocols in conjunction with the `SimpleAuthenticator`.
+
```
spmallette: please include a link to the JIRA here. see other entries for the example.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerIntegrateTest.java
```diff
@@ -212,11 +208,6 @@ public Settings overrideSettings(final Settings settings) {
                 // Trust ONLY the server cert
                 settings.ssl.trustCertChainFile = SERVER_CRT;
             	break;
-            case "shouldStartWithDefaultSettings":
```
spmallette: you removed this test configuration option, but left the test itself? did you mean to do that?

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/channel/AbstractGremlinChannelizerIntegrateTest.java
```diff
@@ -0,0 +1,326 @@
...
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+import java.util.Base64;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+import static org.apache.tinkerpop.gremlin.driver.AuthProperties.Property;
+
+abstract class AbstractGremlinServerChannelizerIntegrateTest extends AbstractGremlinServerIntegrationTest {
```
spmallette: The file name and the class name are different

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/WebSocketHandlerUtil.java
```diff
@@ -0,0 +1,38 @@
...
+import io.netty.handler.codec.http.HttpMessage;
+
+import static io.netty.handler.codec.http.HttpHeaders.Names.UPGRADE;
+import static io.netty.handler.codec.http.HttpHeaders.Names.CONNECTION;
+
+/**
+ * A class to handle common WebSocket operations
+ * @author Keith Lohnes lohnesk@gmail.com
+ */
+public class WebSocketHandlerUtil {
```
spmallette: As this is a utility class, please make it final and give it a private constructor. Maybe it doesn't even need to be public?

spmallette:
I think NIO should have the same capabilities as the WS channelizer. I think you could setup NIO to test the same things WS does. 

Hate to nitpick but could you do a quick pass through your changes and `final` your variables as it's our style.

I like the "channel" tests. We really needed something unified like that.

krlohnes:
@spmallette I fixed the review comments and I think I hit all the places I missed the `final` variables.

krlohnes:
Okay. I got the NIOChannelizer tests passing now as well, things are `final`ed where appropriate, and I've addressed your PR comments.

spmallette:

VOTE +1

spmallette:
I fixed the broken test on tp32.  I guess this needs to be rebased again. Not sure how that failing test snuck through review....kinda odd

spmallette:

VOTE +1

robertdale:
VOTE +1

okram:
VOTE +1

spmallette:
@krlohnes looks like you have all the votes together for this now. it went into tp32 quite well of course, but when i tried to go to master, i hit some problems (tests failing). 

do you mind submitting a second separate pull request to master with these changes? if you do that i can get them both merged at the same time (no need more voting imo).

krlohnes:
@spmallette done.

krlohnes:
@spmallette From the other PR on master I'd had open

>Also, your PR looks a little odd - the commit id is not the same as the one on your other PR.

I'm not quite sure how to go about not mucking up the history here. I'd cherry-picked and fixed the merge conflicts/test failures but that changes the commit sha. 

I tried merging this branch with a new branch. That fixes the the commit id issue, but adds a merge commit. 

I'm a _little_ stumped on how to approach it from here.

spmallette:
the merge commit is acceptable in this case. i'd even expect a follow on commit to fix test problems and such. `git merge` should give us as clean a history as we can get i this situation. thanks for asking on how to proceed.

PR #620, opened 2017-06-08
TINKERPOP-1552 C# GLV: Generics and enum generation

jorgebay:
Added type safety to traversal interface. For example:

```csharp
Vertex vertex = g.V().Next();
// Compile time error
Edge edge = g.V().Next();
```

As C# generics is more strict than in Java, some methods require the user to specify the expected type:
```csharp
// Specify the type of the value of the property "age"
IList<int> userAgeList = g.V().HasLabel("user").Values<int>("age").ToList();
```

Additionally, I've changed the letter case of the enums generated in C# to match the C# naming convention, ie: `g.V().Values<int>("age").Order().By(Order.Decr)`


jorgebay:
cc @FlorianHockmann

PR #623, opened 2017-06-09
TINKERPOP-1552: Minor code cleanup for Gremlin.Net

FlorianHockmann:

This mainly removes explicit type castings that are now obsolete thanks to changes from the pull request #620 from @jorgebay. It also makes the Dictionary CSharpToJavaEnums in NamingConversions private to avoid direct access to it as GetEnumJavaName should be used instead.


jorgebay:
lgtm! +1

PR #626, opened 2017-06-13
TINKERPOP-1489: Update Javascript GLV

jorgebay:
Address feedback and provide maven integration:
- Reorganize gremlin-javascript into node.js project
- Simplify javascript code generators
- Generate package.json to match project version
- Introduce Promise factory for third-party promise library integration (ie: bluebird)
- Use toList() and next() traversal methods
- Include Remote connection implementation
- Fix enum casing
- Use Maven nodejs plugin
- .gitignore and .npmignore at gremlin-javascript level
- Run integration tests using a gremlin-server instance
- Add gremlin-javascript doc in gremlin-variants.asciidoc

Only supports Node.js, as it's basically the only server side js runtime. Browser support can be added in the future. I've encapsulated third party (node.js only for now) dependencies in the module `utils.js`. I've dropped support for Nashorn runtime as it has no immediate practical use. 

It includes versioning and packaging with nodejs toolchain (ie: publish new versions on the Node.js package manager repo) and uses com.github.eirslett frontend-maven-plugin for maven integration (used `skipIntegrationTests` flag).

As it isn't possible to provide blocking IO on Node.js / JavaScript, `next()` and `toList()` provide the same functionality as the Java equivalents but wrapped in a `Promise`.

`next()` traversal method is part of the [async iterator protocol](https://github.com/tc39/proposal-async-iteration) implementation, as proposed by @jbmusso.

A traversal can be iterated using `for await` on modern runtimes: 

```javascript
for await (const vertex of g.V()) {
  console.log(vertex.label);
}
```

`toList()` returns a `Promise` that can be awaited upon:

```javascript
const vertices = await g.V().toList();
```

Promises are created using built-in Promise constructor by default. The user can provide a promise factory to integrate a third-party promise libraries (like bluebird, Q, ...).


jorgebay:
You can run the Node.js tests using `mvn clean install -DskipIntegrationTests=false`

spmallette:
Merged - very cool. Tests work wonderfully.

PR #627, opened 2017-06-15
TINKERPOP-1552: Improve comments in Gremlin-DotNet

FlorianHockmann:

This change adds many comments to Gremlin-DotNet, especially for the GLV part. Some comments where taken from the respective Java classes. It also adds comments to most of the auto-generated methods to tell the user what the method does.

Example for the anonymous class `__`:
```cs
/// <summary>
///     Spawns a <see cref="GraphTraversal{SType, EType}" /> and adds the out step to that traversal.
/// </summary>
public static GraphTraversal<object, Vertex> Out(params object[] args)
{
    return new GraphTraversal<object, object>().Out(args);
}
```
And for a `GraphTraversal` step:
```cs
/// <summary>
///     Adds the aggregate step to this <see cref="GraphTraversal{SType, EType}" />.
/// </summary>
public GraphTraversal< S , E > Aggregate (params object[] args)
{
    Bytecode.AddStep("aggregate", args);
    return Wrap< S , E >(this);
}
```
Apart from the comment improvements, this also marks `GraphTraversalSource.WithBindings` as obsolete as it is also marked as deprecated in Gremlin-Java.

### DIFF
--- a/gremlin-dotnet/glv/AnonymousTraversal.template
+++ b/gremlin-dotnet/glv/AnonymousTraversal.template
@@ -24,18 +24,25 @@
 using System.Collections.Generic;
 using Gremlin.Net.Structure;
 
+// THIS IS A GENERATED FILE - DO NOT MODIFY THIS FILE DIRECTLY - see pom.xml
 namespace Gremlin.Net.Process.Traversal
 {
     /// <summary>
-    ///     THIS IS A GENERATED CLASS - DO NOT MODIFY THIS CLASS DIRECTLY - see pom.xml
+    ///     An anonymous <see cref="GraphTraversal{SType, EType}" />.
     /// </summary>
     public static class __
     {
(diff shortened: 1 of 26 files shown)


spmallette:
Nice additions. I wonder how hard it would be to figure out how to add a link to the official javadoc:

http://tinkerpop.apache.org/javadocs/x.y.z/full/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.html#addV--

I think the x.y.z could be dynamically driven based on the pom.xml without too much trouble. wdyt?

FlorianHockmann:
It shouldn't be too hard to create the link, but I don't really know where to include it. The `summary` element should just include a short description and the `remarks` element is only visible in the generated API docs I think. I also never saw a link included in the document comments. So I don't know how that would work.

PR #628, opened 2017-06-15
TINKERPOP-1550 Made ganglia/graphite optional dependencies

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1550

These dependencies must now be installed manually by users. Updated docs and the LICENSE for Gremlin Server. Gets rid of the weird one-off exclusion of an LGPL license (couldn't support Ganglia out of the box anyway) and lessens transitive dependency burdens.

Builds with `mvn clean install && mvn verify -pl gremlin-server -DskipIntegrationTests=false && mvn verify -pl gremlin-console -DskipIntegrationTests=false`

VOTE +1


gremlin-server/pom.xml
```diff
@@ -63,10 +63,12 @@ limitations under the License.
         <dependency>
             <groupId>com.codahale.metrics</groupId>
             <artifactId>metrics-graphite</artifactId>
+            <optional>true</optional>
```
dkuppitz: Shouldn't this include the version? Asking because you've explicitly mentioned 3.0.2 before.

pom.xml
```diff
@@ -597,6 +597,7 @@ limitations under the License.
                 <groupId>com.codahale.metrics</groupId>
                 <artifactId>metrics-graphite</artifactId>
                 <version>${metrics.version}</version>
+                <optional>true</optional>
```
dkuppitz: Same here. It's probably better to have the version tag included.

gremlin-server/pom.xml
spmallette: All of metrics is under `<dependencyManagement>` in the root pom.xml so the version is defined there:

https://github.com/apache/tinkerpop/blob/268333b4778c87ce6c2429eda97e3e1adc7757e3/pom.xml#L602-L642

I don't think it needs to be specified again.

pom.xml
spmallette: I don't follow - the `<version>` is there. It's presented as a pom variable, but it's defined.

dkuppitz: Doh, I probably only focused on the addition of the optional tag and assumed it's the same dependency block as above. All good here.

dkuppitz:
VOTE: +1

okram:
VOTE +1

PR #629, opened 2017-06-15
TINKERPOP-1552: Clean-up Gremlin-DotNet project files

FlorianHockmann:

This removes some obsolete configuration options and improves the package meta information. Especially the description was extended to reflect the current state of Gremlin-DotNet. This explanation can be removed as soon as the old Gremlin.Net driver is obsolete (probably when a first release version of Gremlin-DotNet is released).
The version is now 3.2.5-beta1.

I also enabled the generation of an XML document containing the documentation comments which will be displayed to the user with IntelliSense. The warning about missing comments had to be disabled for some files as we currently just don't have comments for those. I think that's ok for now, but we should try to find a way to create meaningful comments where they are missing right now for the first official (non-beta) release.


gremlin-dotnet/glv/AnonymousTraversal.template
```diff
@@ -27,6 +27,8 @@ using Gremlin.Net.Structure;
 // THIS IS A GENERATED FILE - DO NOT MODIFY THIS FILE DIRECTLY - see pom.xml
 namespace Gremlin.Net.Process.Traversal
 {
+#pragma warning disable 1591
```
jorgebay: Is this needed for this source file, all methods are being generated with docs.

jorgebay:
Apart from my previous comment, lgtm.

gremlin-dotnet/glv/AnonymousTraversal.template
FlorianHockmann: Looks like you're right. I thought that it would be necessary to also document all function arguments and return values, but it seems to be enough to have a `summary` for each element. This also applies to the `GraphTraversal` class. I'll update the pull request to remove the `pragmas` for both classes.

FlorianHockmann: Done.

spmallette:
I'm going to merge this then tp32-glv and this branch on the latest from tp32 now that the release is done. Should be in shortly.

PR #630, opened 2017-06-17
Restore 5 second cutoff time in TraversalInterruptionComputerTest

sjudeng:
The change in https://github.com/apache/tinkerpop/commit/9983fd210872d57750d49191c391ba8e380e6023 included decreasing the maximum total test time threshold from 5 seconds to 1 second in TraversalInterruptionComputerTest. This PR proposes to revert the timeout back to the original 5 seconds to resolve test failures observed in some custom graph computer implementations.


spmallette:
Thanks - hopefully 5 seconds will suffice. I suppose this time limit is more of a failsafe to prevent the test from hanging indefinitely so it could be bigger if a graph needed longer iteration time for some reason.

PR #631, opened 2017-06-18
TINKERPOP-1445 Add ellipses for long property values

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1445

This was an easy change but I decided to do this as a PR so that the change would be noted as it does affect something that is user facing. I did this change for 3.3.0 only in case someone is depending on the `toString()` value of `Property` for some reason.

```text
gremlin> graph = TinkerGraph.open()
gremlin> g = graph.traversal()
gremlin> g.addV().property('short','pretty short').property('long','this one is much longer so it is going to get truncated a bit').as('a').
......1>   addE('self').property('short','pretty short').property('long', 'this one is much longer so it is going to get truncated a bit').from('a').to('a')
==>e[3][0-self->0]
gremlin> g.V().properties()
==>vp[short->pretty short]
==>vp[long->this one is much ...]
gremlin> g.E().properties()
==>p[short->pretty short]
==>p[long->this one is much ...]
```

VOTE +1

### DIFF
--- a/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/structure/PropertyTest.java
+++ b/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/structure/PropertyTest.java
@@ -59,9 +59,35 @@ public class PropertyTest {
     public static class BasicPropertyTest extends AbstractGremlinTest {
         @Test
         @FeatureRequirementSet(FeatureRequirementSet.Package.VERTICES_ONLY)
-        public void shouldHaveStandardStringRepresentation() {
+        public void shouldHaveStandardStringRepresentationForVertexProperty() {
             final Vertex v = graph.addVertex("name", "marko");
             final Property p = v.property("name");
+            assertEquals(StringFactory.propertyString(p), p.toString());
+        }
+
+        @Test
+        @FeatureRequirementSet(FeatureRequirementSet.Package.VERTICES_ONLY)
+        public void shouldHaveTruncatedStringRepresentationForVertexProperty() {
+            final Vertex v = graph.addVertex("name", "maria de la santa cruz rosalina agnelia rodriguez cuellar rene");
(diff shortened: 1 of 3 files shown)


dkuppitz:
VOTE: +1 (and an extra +1 for finding a super long name that actually exists :))

okram:
VOTE +1. (However, I wouldn't add the <space> between the last line and the `...` but whateves)

spmallette:
hmm - i didn't add that. It's the `StringUtils` utility. I guess it is not smart about the final space.

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/structure/PropertyTest.java
```diff
@@ -59,14 +59,40 @@
...
             final Property p = v.property("name");
             assertEquals(StringFactory.propertyString(p), p.toString());
         }
 
         @Test
         @FeatureRequirementSet(FeatureRequirementSet.Package.VERTICES_ONLY)
+        public void shouldHaveTruncatedStringRepresentationForVertexProperty() {
+            final Vertex v = graph.addVertex("name", "maria de la santa cruz rosalina agnelia rodriguez cuellar rene");
+            final Property p = v.property("name");
+            assertEquals(StringFactory.propertyString(p), p.toString());
```
robertdale: The only thing that this test validates is that the toString has the same output as the StringFactory which may be all well and good.  However, if StringFactory changes, the would still pass, but there could be an end-user display impact that isn't caught.  I'm not against it as long as we all understand the implications. I prefer hardcoded expectations to catch these things.

spmallette: That test is more to validate that providers are using the `StringFactory` (or at least make their stuff align to `StringFactory` output).

PR #632, opened 2017-06-19
TINKERPOP-1519: tinker graph computer does not handle multiple scopes

sheldonkhall:
https://issues.apache.org/jira/browse/TINKERPOP-1519

This change modifies the tinker graph computer so that each message sent in a vertex program remembers its scope. Previously when the receiveMessages method on TinkerMessenger was called it would loop through ALL message scopes and then ALL messages, which is incorrect. Now the method loops over each scope, and then each message within that scope.

I have added the regression test suggested in the JIRA ticket and run mvn clean install locally to confirm everything passes OK.

I had a quick look at the latest master, and this bugfix may need to be merged there too.


spmallette:
@sheldonkhall  I just wanted to quickly poke in here and thank you for submitting this - @okram is probably the best person to review this and he hasn't been feeling well. He will probably get to it next week.

okram:
The problem is real, the fix works given the specific test case.

VOTE +1.

dkuppitz:
VOTE: +1

spmallette:
VOTE: +1

PR #638, opened 2017-06-20
TINKERPOP-999 Removed generics from ServerGremlinExecutor

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-999

Pretty simple - I probably could have CTR'd this but since it was a breaking change on 3.3.0 I wanted folks to be more aware of it in case someone was using this class in some way I'm not imagining.

Builds with `mvn clean install -DskipTests && mvn verify -pl gremlin-server -DskipIntegrationTests=false`

VOTE +1

### DIFF
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GremlinServer.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GremlinServer.java
@@ -69,12 +69,12 @@ public class GremlinServer {
     private Channel ch;
 
     private CompletableFuture<Void> serverStopped = null;
-    private CompletableFuture<ServerGremlinExecutor<EventLoopGroup>> serverStarted = null;
+    private CompletableFuture<ServerGremlinExecutor> serverStarted = null;
 
     private final EventLoopGroup bossGroup;
     private final EventLoopGroup workerGroup;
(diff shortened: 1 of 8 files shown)


gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GremlinServer.java
```diff
@@ -108,54 +108,24 @@ public GremlinServer(final Settings settings) {
             workerGroup = new NioEventLoopGroup(settings.threadPoolWorker, threadFactoryWorker);
         }
 
-        serverGremlinExecutor = new ServerGremlinExecutor<>(settings, null, workerGroup, EventLoopGroup.class);
+        serverGremlinExecutor = new ServerGremlinExecutor(settings, null, workerGroup, null);
```
dkuppitz: Looks like the 4th parameter is no longer needed. Since the PR targets `master/`, it would be fine to remove it altogether, no?

spmallette: i think you're right. I sorta wonder if the second parameter is of value. Gremlin Server is the only thing that really creates that ExecutorService.

okram:
VOTE +1

PR #641, opened 2017-06-28
TINKERPOP-1703: Make EdgeOtherVertexStep non-final

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1703

The ticket really is about making {{EdgeOtherVertexStep}} non-final so providers, like DSEGraph, can extend it accordingly.

Simple fix.

**SIDENOTE**: @dkuppitz -- I wonder if there is a more efficient way to implement this step. I looked into using `Path.get(i)`, but it seems while it makes less data, it is more expensive to execute (at least for `ImmutablePath`).

VOTE +1.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,7 @@ TinkerPop 3.2.6 (Release Date: NOT OFFICIALLY RELEASED YET)
 This release also includes changes from <<release-3-1-8, 3.1.8>>.
 
 * Added annotations to the traversal metrics pretty print.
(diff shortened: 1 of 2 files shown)


spmallette:
VOTE +1 - I think providers will generally appreciate being able to extend the existing steps. Seems easier to allow this now that we have a good feeling for the stability of the steps themselves.

dkuppitz:
VOTE: +1

BrynCooke:
Can we rename the PR?

okram:
Done.

okram:
Merged. I think that its not being auto-closed cause I changed the PR title.

PR #642, opened 2017-06-28
TINKERPOP-1704: XXXTranslators are not being respective of BulkSet and Tree.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1704

`JavaTranslator` would convert `Tree` to `HashMap` and `BulkSet` to `HashSet`. While passing in such objects through `Bytecode` is low probability, it is still good to cover this corner case (especially for `Tree`, where the semantics are a bit different than `Map`).

VOTE +1


spmallette:

VOTE +1

dkuppitz:
VOTE: +1

PR #645, opened 2017-06-29
TINKERPOP-1698 Gryo 3.0

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1698

Gryo 3.0 is in place. A good bunch of the changes just had to do with adding more tests to cover 3.0 and 1.0.  For 3.3.0 Gryo 3.0 will be the default. Users can easily switch back to 1.0 through various configuration options. Gryo 3.0 isn't really compatible with 1.0 - there may be partial compatibility here and there but it wouldn't be recommended to try to mix them in any way.

Full builds with docker work nicely.

VOTE +1


okram:
VOTE +1

dkuppitz:

VOTE: +1

PR #648, opened 2017-06-30
TINKERPOP-1710: Add a note on tree() by-modulation and uniqueness of tree branches.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1710

There was confusion as to what should happen what `tree()` sees duplicate objects in the same depth of the tree. The behavior is as expected and not a "bug." Tree is constructed using path objects that are "horizontally" linked. If two objects are the same at the same depth, then they are merged into one none in the tree data structure.

VOTE +1

### DIFF
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -2283,6 +2283,20 @@ tree['marko']['josh']
 tree.getObjectsAtDepth(3)
 ----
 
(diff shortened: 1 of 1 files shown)


twilmes:
Travis sure is finicky today with time based tests, doc update looks good.

VOTE: +1

spmallette:
VOTE +1

PR #658, opened 2017-07-10
TINKERPOP-1714 GremlinExecutor checks for timeout from time script submitted

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1714

Prior to this change timeouts were considered from the time the script started evaluation which probably isn't in line with user expectations but also made it hard to clear the job queue on an overloaded server. This change also has the benefit of simplifying the code as well as reducing the number of jobs submitted to the various thread pools.

VOTE +1


dkuppitz:
VOTE: +1

okram:
VOTE +1

spmallette:
This is merged but hasn't auto-closed yet - seems like the mirror/events are behind. Will keep an eye on it to see that it does indeed close itself.

PR #663, opened 2017-07-10
[TINKERPOP-915]Add combined handler for Http and Websockets

krlohnes:
[TINKERPOP-915](https://issues.apache.org/jira/browse/TINKERPOP-915)

Most of this is tests. I added an integration test that goes over the
functionality of both the http and websocket channelizers using the new
`WsAndHttpChannelizer`. I added an additional test on top of that to
switch between using WebSockets and http.

The change itself leverages the existing `WebSocketChannelizer` to
provide the base pipeline setup. It has everything needed handler-wise
to service both http and ws connections. The
`WsAndHttpChannelizerHandler` then detects whether the incoming request
is a plain http connection or a WebSockets connection. If it's an http
connection, the channelizer handler swaps out the request handler
appropriately for whether or not authentication has been enabled.


spmallette:
I'm not sure if you're done with this, but as it sits right now you didn't resolve the compilation problems (and even when you resolve those there are test failures - or maybe i didn't resolve the merges/compilation problems correctly). please run the build with integration tests to see the problems. 

Also, your PR looks a little odd - the commit id is not the same as the one on your other PR.

PR #669, opened 2017-07-11
TINKERPOP-1686 Thread safe Traversal Metrics

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1686

Improved the thread safety of  traversal metrics. Some graph providers may write to these metrics from multiple threads and it may thus cause problems. These changes had to occur on 3.3.0 when a break was allowable as the internals of the class itself had to change and that bound the related classes to Gryo 1.0 compatibility along the 3.2.x line.

VOTE +1


twilmes:
VOTE: +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversalMetrics.java
```diff
@@ -23,52 +23,64 @@
...
-        this.totalStepDuration = totalStepDurationNs;
-        this.computedMetrics = new LinkedHashMap<>(this.metrics.size());
-        final AtomicInteger counter = new AtomicInteger(0);
-        metricsMap.forEach(metric -> {
-            this.computedMetrics.put(metric.getId(), metric.getImmutableClone());
-            this.indexToLabelMap.put(counter.getAndIncrement(), metric.getId());
-        });
+    public DefaultTraversalMetrics(final long totalStepDurationNs, final List<MutableMetrics> orderedMetrics) {
+        totalStepDuration = totalStepDurationNs;
+        for (int ix = 0; ix < orderedMetrics.size(); ix++) {
```
dkuppitz: Good that `orderedMetrics.forEach()` was replaced with a simple `for` loop, but I'm not sure if calling get(ix) 3 times really makes it more efficient. Maybe:

```
int ix = 0;
for (final MutableMetrics metric : orderedMetrics) {
    stepIndexedMetrics.put(metric.getId(), metric.getImmutableClone());
    positionIndexedMetrics.put(ix++, metric.getImmutableClone());
}
```

PR #644, opened 2017-06-29
TINKERPOP-1552 Gremlin.Net

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1552

Presenting the Gremlin.Net GLV. It follows patterns for GLV generation established by gremlin-python and has build/deployment automation in maven. It does not have an extensive test suite given the JVM-centric nature of gremlin-test, but it does contain both unit and integration tests that should provide some confidence. 

Upgrade documentation has not been updated yet since it isn't clear if we will officially release this with 3.2.6 or not. We may just do SNAPSHOT style releases until we gather more feedback. I suspect that we will have more changes to Gremlin.Net before 3.2.6 is ready for release, but the work seems stable enough to be in a release branch at this point.

Builds with docker nicely - @dkuppitz could you please take a look at the docker changes for this? not sure if anything else needs to be done (or done differently).

VOTE +1

PR #670, opened 2017-07-12
TINKERPOP-1552 (master) Gremlin .NET

spmallette:
This is the follow-on pull request to #644 that merges those changes to master.  It required some additional work unfortunately so I formed a separate pull request. Note that there are still failing .NET integration tests.

I also don't think we should merge this until TINKERPOP-1427 ( #672 )  is merged to master which will mean another rebase for this pull request. 

This pull request does not require a VOTE at this time - it's just here as a placeholder for collaboration until everything is ready.


spmallette:
just rebased on master and force pushed

jorgebay:
I'm looking into the `g_V_HasXname_markoX_ValueMap_Next` failure on gremlin-dotnet, it looks like the generation process was producing:

```
public GraphTraversal<S, IDictionary<string, E2>> ValueMap<E2>(params object[] args)
```

and now:

```
public GraphTraversal< S , IDictionary<object, E2> > ValueMap<E2> (params object[] args)
```

for java, both signatures exists (with different parameters). The safest route would be to use a second type parameter:
```
public GraphTraversal< S , IDictionary<K, V> > ValueMap<K, V> (params object[] args)
```
I'll try to find a solution. Afterwards, we should look into ways to do .NET generation as deterministic as possible.

spmallette:
I was just trying to get it to compile/run when I did that change. The generation seems a bit brittle. Would be nice to see that stuff "just work" regardless of what API changes gremlin sees. 

> Afterwards, we should look into ways to do .NET generation as deterministic as possible.

agreed - let's save that for future work. i'd just like to get this code into mainline with as few additional changes as possible at this point.

jorgebay:
I've reused the `methodsWithSpecificTypes` to generate `ValueMap()` with 2 type parameters.

spmallette:
Note that the #672 is out for review now. Once that's merged we can get this .NET branch rebased, get Gremlin.NET configurable for both GraphSON 2.0 and 3.0 and this body of work can be merged.

spmallette:
Just rebased and force pushed now that #672 has been merged. This means that we now just need to get Gremlin.Net working with GraphSON 3.0 which added `g:Map`, `g:List` and `g:Set`.  The behavior should be similar to java and python where you can select the serializer you want to use.  That way Gremlin.Net will work with GraphSON 2.0 or 3.0 depending on user selection.

jorgebay:
It looks like my last commit c7a798c9e2f18936227d0fc423edfc2f71d7981f was lost in the force push, I'll reinstate it.
About getting it to work with GraphSON3, I'll try to do that today.

spmallette:
sorry @jorgebay  - that's the first time that's happened. i thought i had pulled your change but i must not have......

jorgebay:
np really!

jorgebay:
I've added support for GraphSON3 on the Gremlin .NET:
- Added `GraphSON3Writer` and `GraphSON3Reader`.
- Modified unit tests to run using GraphSON2 and GraphSON2.

I'm getting a failures on the integration tests of the .NET GLV, that start with:
```
[xUnit.net 00:00:01.2357664]   Starting:    Gremlin.Net.IntegrationTest
[WARNING] Exception processing a script on request [RequestMessage{, requestId=bad279b8-6f42-4e47-ad08-a0343239bc1d, op='eval', processor='', args={gremlin=1, language=unknown}}].
```
I don't understand what changed at protocol level (besides the mime type) to cause this issue.

@FlorianHockmann @spmallette I will not able to dedicate time the following weeks to this. Feel free to adapt / override my changes.

spmallette:
I reverted back to 2.0 for the default for testing. Only two tests failing under than model now that i have the server configured right. Not sure why there's two failures though on 2.0. Still trying to figure it out.

spmallette:
Just pushed a fix for one of the two failing tests on this branch (still running on graphson 2.0 at this time). It was an authentication configuration problem with Gremlin Server. No clue yet as to what the issue is with the other tests just yet.  Hoping that once the GraphSON 2.0 stuff all works we can try graphson 3.0. i guess i might even consider merging it all with just 2.0 support. we won't have an official release of gremlin-dotnet anyway, so full support of graphson 3.0 isn't super critical.

PR #673, opened 2017-07-12
TINKERPOP-1709: Add a list of all the steps that support by()/from()/to()/as()/option()

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1709

Listed and referenced (linked to) those steps that support the various "modulations." The only one not added was `as()`, given that its not really a modulator and works for all steps.

VOTE +1


dkuppitz:
VOTE: +1

PR #678, opened 2017-07-17
TINKERPOP-1715: update spark version to 2.2

artem-aliev:
That required:
  - more spark conflicting dependency exclusion in pom.xml
  - more spark and scala classes registration in tinkerPop gryo

### DIFF
--- a/gremlin-groovy/pom.xml
+++ b/gremlin-groovy/pom.xml
@@ -93,6 +93,10 @@ limitations under the License.
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>junit</groupId>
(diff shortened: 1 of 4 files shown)


gremlin-groovy/pom.xml
```diff
@@ -93,6 +93,10 @@ limitations under the License.
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
artem-aliev: Hm, It was needed for build.....
Now I can not reproduce the failure, so rollback it

spmallette:

VOTE +1

okram:
VOTE +1

davebshow:
VOTE +1

PR #682, opened 2017-07-25
TINKERPOP-1736 Sack step evaluated by groovy interprets numbers in an unexpected way

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1736

Fixed a bug in `BigDecimal` divisions in `NumberHelper` that potentially threw an `ArithmeticException`.

VOTE: +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -27,6 +27,12 @@ TinkerPop 3.1.8 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 * Fixed a `MessageScope` bug in `TinkerGraphComputer`.
+* Fixed a bug in `BigDecimal` divisions in `NumberHelper` that potentially threw an `ArithmeticException`.
+
+Bugs
+^^^^
+
+* TINKERPOP-1736 Sack step evaluated by Groovy interprets numbers in an unexpected way
 
 [[release-3-1-7]]
 TinkerPop 3.1.7 (Release Date: June 12, 2017)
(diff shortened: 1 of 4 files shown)


robertdale:
VOTE +1

okram:
VOTE +1

spmallette:
I didn't get a chance to review/vote on this before it was merged, but I dont' think this is what I expected to happen when I wrote the JIRA. I didn't think we would produce direct support for `BigDecimal` so much as coerce `BigDecimal` into double so that when Groovy did its magic to:

```groovy
g.withSack(2).V().sack(Operator.div).by(constant(3.0)).sack()
```

we'd just get back a double - not a `BigDecimal`. Directly supporting `BigDecimal` means we need to support it as a GraphSON type. We already do, but only in the extended module so if you wanted to use `BigDecimal` you'd have to include that configuration. No GLV at this point supports the extended module (though @davebshow is working on that). Do we really need to provide direct support for `BigDecimal` when the problem we were trying to solve was related to Groovy taking a literal and auto-assigning it to `BigDecimal` (sorta unexpected)? or do users really need the ability to do `constant(BigDecimal.valueOf(3))`?

dkuppitz:
IMO it makes sense to support `BigDecimal`. If not for big numbers, then at least for the better precision.

okram:
Ah......... I didn't realize that either. BigDecimal is not something we want to support.

spmallette:
I've dug into supporting `BigDecimal` directly and it's turning into a mess a bit. Python does have `Decimal` class which is like `BigDecimal` but it doesn't serialize through the standard JSON processor:

https://stackoverflow.com/questions/1960516/python-json-serialize-a-decimal-object

so I coerced to float to match our extended module format:

http://tinkerpop.apache.org/docs/current/dev/io/#_bigdecimal

for a number in the `@value`. That sorta worked, but it then causes a loss in precision anyway. I think it's just better to go with what I said before. Just coerce the `BigDecimal` groovy pushes on us to double internally and then it all just works. The test should not explicitly test `BigDecimal`. It should test:

```groovy
g.withSack(2).V().sack(Operator.div).by(constant(3.0)).sack()
```

which in Java will just test double stuff, but in Groovy it will produce a `BigDecimal` and our code gets tested. Does that make sense?

dkuppitz:
I created 2 new PR's, one for `tp31/`, the other for `master/`.

`tp31`: #683 
`master`: #684

PR #690, opened 2017-08-09
TINKERPOP-1744 Unwrap AggregateException for sync io operations

jorgebay:
https://issues.apache.org/jira/browse/TINKERPOP-1744

Includes a test expecting the underlying exception (`ResponseException`) to be thrown.

Builds with `mvn clean install -P gremlin-dotnet`

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Structure/Utils.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Structure/Utils.cs
@@ -1,0 +1,55 @@
+﻿#region License
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#endregion
+
+using System;
+using System.Runtime.ExceptionServices;
+using System.Threading.Tasks;
+
+namespace Gremlin.Net.Structure
+{
+    /// <summary>
+    /// Contains useful methods that can be reused across the project. 
+    /// </summary>
+    internal static class Utils
+    {
+        /// <summary>
+        /// Waits the completion of the provided Task.
+        /// When an AggregateException is thrown, the inner exception is thrown.
+        /// </summary>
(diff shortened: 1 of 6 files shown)


jorgebay:
Sorry, I've messed up with the name of the pull request, affecting JIRA ticket 1445 logs...

gremlin-dotnet/src/Gremlin.Net/Structure/Utils.cs
```diff
@@ -0,0 +1,55 @@
...
+ * under the License.
+ */
+
+#endregion
+
+using System;
+using System.Runtime.ExceptionServices;
+using System.Threading.Tasks;
+
+namespace Gremlin.Net.Structure
```
FlorianHockmann: Wouldn't it make more sense to put this in `Gremlin.Net.Process` as all users of this are in that namespace?

gremlin-dotnet/src/Gremlin.Net/Structure/Utils.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// </summary>
+        public static void WaitUnwrap(this Task task)
+        {
+            try
+            {
+                task.Wait();
+            }
+            catch (AggregateException ex)
+            {
+                if (ex.InnerExceptions.Count == 1)
```
FlorianHockmann: Why not iterating over the `InnerExceptions` and then rethrowing all of them? That's done for example in the first listing here: https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/exception-handling-task-parallel-library

jorgebay: Its not possible to throw more than 1 exception, once you throw the following code is unreachable.

The `if` block is to be future proof, but its not required for now: all `AggregateException` instances thrown have just 1 inner exception, as those are caused by using `await` (not multiple parallel ops).

gremlin-dotnet/src/Gremlin.Net/Structure/Utils.cs
jorgebay: It's being used in both `Gremlin.Net.Driver` and `Gremlin.Net.Process` namespaces.

gremlin-dotnet/src/Gremlin.Net/Structure/Utils.cs
FlorianHockmann: Of course, you're right, the iteration only makes sense in the documentation because they don't rethrow their `CustomException`. I didn't really think that through when writing the comment...

gremlin-dotnet/src/Gremlin.Net/Structure/Utils.cs
FlorianHockmann: I still think that it fits better in `Process` as `Structure` only contains the Graph elements and the `IO` related stuff, but that's basically a question of personal taste. So I'll give this my +1 vote independently of this.

jorgebay: I've changed to `Gremlin.Net.Process` :)

jorgebay:
VOTE: +1.

jorgebay:
Does any other contributor has time to look into this PR?
I would like to include it before the code freeze as it involves an API change (different exceptions thrown).

spmallette:
VOTE +1

PR #694, opened 2017-08-16
TINKERPOP-1748 Callout comments break code snippets

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1748

Made callouts in docs prettier and copy+paste friendly.

Tested with `docker/build.sh -d`. Looks great and all snippets can now be pasted into the console w/o producing any syntax errors due to callout comments.

VOTE: +1

### DIFF
--- a/docs/postprocessor/processor.awk
+++ b/docs/postprocessor/processor.awk
@@ -19,13 +19,33 @@
 
 BEGIN {
   firstMatch=1
+  styled=0
 }
 
 /Licensed to the Apache Software Foundation/ {
   isHeader=1
 }
 
-firstMatch || !isHeader
+/<\/style>/ {
+  if (!styled) {
+    print ".invisible {position: absolute; left: 9999px}"
+    styled=1
+  }
+}
+
+!/<span class="comment">/ {
+  if (firstMatch || !isHeader) {
+    print gensub(/(<b class="conum">)\(([0-9]+)\)(<\/b>)/,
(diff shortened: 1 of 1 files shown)


robertdale:
Looks like it works for some but not all.
The very first one on the reference page didn't work:
Graph graph = TinkerGraph.open(); (1)

Looks like there are ones with circles and have the round brackets which don't work.
Then there are the ones with circles without round brackets which do work.

Then there's still the issue where comments inline of a multi-line command doesn't work in groovy console.

dkuppitz:
Huh, interesting, these are static / non-evaluated code blocks. I will check why they were not affected.

> Then there's still the issue where comments inline of a multi-line command doesn't work in groovy console.

This is only true for the static blocks, right? I've tried a few snippets from evaluated code blocks (single- and multi-line) and they all worked.

robertdale:
Ah, I see now. You've added line continuations `\` on the mid-traversals.  Excellent.

dkuppitz:
Static code blocks are now clean too.

robertdale:
VOTE +1

spmallette:
This looks good - does this basically allow us to close: 

https://issues.apache.org/jira/browse/TINKERPOP-1447

It doesn't do everything that ticket says, but that's about as good as we're going to get I would imagine, right?

VOTE +1

dkuppitz:
Well, this PR only tackles the callout comments. Any other comment would still break copy+paste. However, the cool thing about this solution is, that it doesn't involve JavaScript, only a minor CSS tweak.

PR #696, opened 2017-08-18
TINKERPOP-1754: Do not add non deserialisable exception to ScriptReco…

artem-aliev:
…rdReader IOExceptions


spmallette:
I plan to merge this via CTR - I don't feel like there's a need to vote on this one. It's pretty straightforward.

PR #697, opened 2017-08-18
TINKERPOP-1753 OrderStep not able to order by non-integer numbers

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1753

Fixed ordering of collections / streams containing mixed number types.

I decided to go with `try / catch`, since this should really be an exceptional case. Usually number types shouldn't be mixed up, hence I didn't want this PR to have a negative impact on the overall order performance by always checking the types upfront.

VOTE: +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.2.7 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Fixed a bug that prevented Gremlin from ordering lists and streams made of mixed number types.
 
 
 [[release-3-2-6]]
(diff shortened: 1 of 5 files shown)


dkuppitz:
Btw, I will update `CHANGELOG` after the code freeze.

okram:
Eek. Yea. That `try/catch` sorta sucks. Can you do some performance testing with and without they `try/catch`? Would you mind?

dkuppitz:
Alright, turned out it doesn't have a measurable performance impact.

```
if (first instanceof Number && second instanceof Number) {
    return NumberHelper.compare((Number) second, (Number) first);
}
return Comparator.<~>naturalOrder().compare((Comparable) first, (Comparable) second);
```

is as fast as

```
try {
    return Comparator.<~>naturalOrder().compare((Comparable) first, (Comparable) second);
} catch (final ClassCastException ex) {
    if (first instanceof Number && second instanceof Number) {
        return NumberHelper.compare((Number) first, (Number) second);
    } else throw ex;
}
```

Tested using the Grateful Dead graph and the following query:

```
g.V().hasLabel("song").
  project("a","b").
    by("name").
    by(outE("followedBy").values("weight").sum()).
  order().
    by(select("b"), decr)
```

Thus I will go with the first variant, not using `try / catch`.

okram:
Cleaner `if/else` clause.

```
return first instanceof Number && second instanceof Number ?
    NumberHelper.compare((Number) second, (Number) first) : 
    Comparator.<~>naturalOrder().compare((Comparable) first, (Comparable) second);
```

okram:
VOTE +1

spmallette:

VOTE +1

PR #698, opened 2017-08-23
added repeat simple path test

xytxytxyt:


okram:
What is the intention of the test? What are you trying to prove?

xytxytxyt:
the short story is that there was a defect in dse graph exposed by this query, that luckily was caught because one of our tests happened to have it, but could be added to tinkerpop so that at least it can be caught earlier next time

so it's basically to plug a hole

more information: https://datastax.jira.com/browse/DSP-14204

okram:
Is the `limit(1)` necessary for the DSEGraph testing because against MODERN its the same result?There is only 1 outgoing path between marko and ripple.

```
gremlin> g.V().has('name', 'marko').repeat(outE().inV().simplePath()).until(has('name', 'ripple')).limit(1).path().by(values('name')).by(T.label)
==>[marko,knows,josh,created,ripple]
gremlin> g.V().has('name', 'marko').repeat(outE().inV().simplePath()).until(has('name', 'ripple')).path().by(values('name')).by(T.label)
==>[marko,knows,josh,created,ripple]
```

xytxytxyt:
not sure, since i don't know what in dse graph failed... what's the difference between having it there and not?

okram:
The result is the same. However, I would prefer to have "the simplest query that fails" and thus, no `limit(1)`, but if the DSEGraph traversal strategies are failing cause of `limit(1)`, then we can include it. Perhaps you could test to see if you get the right result in DSEGraph without `limit(1)`.

```
graph.io(gryo()).readGraph('data/tinkerpop-modern.kryo')
g = graph.traversal()
```

xytxytxyt:
i removed the `limit(1)`

okram:
VOTE +1. (I can do the final merge when all the votes are tallied)

gremlin-groovy-test/src/main/groovy/org/apache/tinkerpop/gremlin/process/traversal/step/branch/GroovyRepeatTest.groovy
```diff
@@ -97,5 +97,10 @@ public abstract class GroovyRepeatTest {
         public Traversal<Vertex, Map<String, Long>> get_g_V_repeatXbothX_untilXname_eq_marko_or_loops_gt_1X_groupCount_byXnameX() {
             new ScriptTraversal<>(g, "gremlin-groovy", "g.V.repeat(both()).until{it.get().value('name').equals('lop') || it.loops() > 1}.groupCount.by('name')")
         }
+
+        @Override
+        public Traversal<Vertex, Path> get_g_V_hasXname_markoX_repeatXoutE_inV_simplePathX_untilXhasXname_rippleXX_path_byXnameX_byXlabelX() {
+            new ScriptTraversal<>(g, "gremlin-groovy", "g.V().has('name', 'marko').repeat(outE().inV().simplePath()).until(has('name', 'ripple')).limit(1).path().by(values('name')).by(T.label)")
```
dkuppitz: Not a big deal, but Groovy tests should use Gremlin sugar notation.

okram: Good find. I'll tweak that accordingly on merge.

spmallette:

VOTE +1

okram:
Merged. I had to tweak both tests a bit. `limit(1)` still in the groovy test. The test used `values('name')` instead of just `name`. Groovy test was not in sugar syntax.

okram:
@xytxytxyt -- please close this ticket if everything looks good to you.

xytxytxyt:
thanks!

PR #699, opened 2017-08-28
TINKERPOP-1759 Improve hashcode and equals for Traverser implementations

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1759

Improved `hashcode()` and `equals()` for all Traverser implementations and added a performance test to ensure that things don't get reverted when someone thinks that the hashcode implementations could be rewritten.

VOTE: +1


tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/process/traversal/TinkerGraphPerformanceTest.java
```diff
@@ -0,0 +1,53 @@
...
+ * @author Daniel Kuppitz (http://gremlin.guru)
+ */
+public class TinkerGraphPerformanceTest {
+
+    /**
+     * Check the traverser's hashcode() implementation if this test fails. hashcode() should
+     * not generate too many hash collisions.
+     */
+    @Test
+    public void sacksWithoutMergerShouldBeFast() {
```
okram: This should be in a class called `SackStrategyTest` as `SackStrategy` is what is being analyzed. This would be analogous to how we do timing tests in `RepeatUnrollStrategyTest`.

dkuppitz: Hmm, I'm not sure how to rewrite the test using the `inject()` approach. The following snippet shows no performance issues at all:

```
strategies = new DefaultTraversalStrategies()
strategies.addStrategies(SackStrategy.build().initialValue({1}).create())

t = __.inject((1..8192).toArray()).barrier(); []
t.asAdmin().setStrategies(strategies)
t.asAdmin().applyStrategies()
t.iterate()
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/traverser/B_O_Traverser.java
```diff
@@ -62,10 +62,12 @@ public void setStepId(final String stepId) {
         this.future = stepId;
     }
 
+    protected final boolean equals(final B_O_Traverser other) {
+        return super.equals(other) && other.future.equals(this.future);
+    }
+
     @Override
```
okram: This class does not have a `hashCode()` method. It should since it implements `equals()`. Its easy, its just `this.t.hashCode()`.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/traverser/LP_O_OB_P_S_SE_SL_Traverser.java
```diff
@@ -106,17 +106,18 @@ public void dropPath() {
 
     @Override
     public int hashCode() {
-        return super.hashCode() + this.path.hashCode();
+        if (carriesUnmergeableSack()) {
```
okram: Can you use inline `returns carriesUnmergableSack() ? ... : ...` please?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/traverser/O_Traverser.java
```diff
@@ -23,6 +23,7 @@
 import org.apache.tinkerpop.gremlin.process.traversal.traverser.util.AbstractTraverser;
```
okram: There is no `hashCode()` method for this class either like `B_O_...`. Again, its just `return t.hashCode()`

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/traverser/B_O_S_SE_SL_Traverser.java
```diff
@@ -110,17 +112,22 @@ public void merge(final Traverser.Admin<?> other) {
...
+        return null != this.sack && (null == this.sideEffects || null == this.sideEffects.getSackMerger());
+    }
+
     @Override
     public int hashCode() {
-        return this.t.hashCode() + this.future.hashCode() + this.loops;
+        return carriesUnmergeableSack() ? System.identityHashCode(this) : (super.hashCode() ^ this.loops);
+    }
+
+    protected final boolean equals(final B_O_S_SE_SL_Traverser other) {
```
okram: I'm a little scared of this "double equals()" definition. @spmallette -- do you have any thoughts on this matter? Note that there is an `equals(Object)` method like always, but then there is this added `protected final equals(SpecificClass)`. I have a feeling that this could get screwy for languages like Groovy. Perhaps there is a better way to do what is needed without overloading `equals()`?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/traverser/B_O_Traverser.java
dkuppitz: That's intentional. B_O_Traverser's hashcode is the same as O_Traverser's. Same story below. In both cases equals() will do the "deep" check, but hashcode() ignores fields that are unlikely different.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/traverser/LP_O_OB_P_S_SE_SL_Traverser.java
dkuppitz: Be glad that your agreeability score is 1. Mine is 3, thus I won't start a discussion about these 2 variants ending up having the same bytecode. Will change that tomorrow.

dkuppitz:
I can rename them or just make them package private. I'd prefer the latter. In fact that's what I've at first... I changed it to prevent your "please make that final protected" comment ;)

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/decoration/SackStrategyTest.java
```diff
@@ -0,0 +1,66 @@
...
+public class SackStrategyTest {
+
+    public static class PerformanceTest {
+
+        /**
+         * Check the traverser's hashcode() implementation if this test fails. hashcode() should
+         * not generate too many hash collisions.
+         */
+        @Test
+        public void sacksWithoutMergerShouldBeFast() {
```
spmallette: I don't think we should introduce tests that do this type of assertion. There are no filters for "PerformanceTest" in master anymore so this test will always run. I think that if we are concerned about system performance and want to test for it, then we should add something to gremlin-benchmarks.

dkuppitz:
`docker/build.sh -t -n -i` still succeeds after recent changes.

Still VOTE: +1

okram:
I'm scared -- VOTE +1.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/traverser/B_O_Traverser.java
robertdale: I agree with @okram. There should be an explicit `hashCode()` so it doesn't look like it is mistakenly forgotten.

robertdale:
VOTE +1

PR #702, opened 2017-08-30
TINKERPOP-1764: Generalize MatchStep to localize all barriers, not just reducing barriers.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1764

Prior to this moment, only reducing barrier steps in a `match()`-clause forced the clause to be computed locally (via a `flatMap()`-wrap). However, it has been deemed important to generalize this behavior to all barriers (e.g. `order()`, `limit()`, etc.) as identified by @doanduyhai (Gremlin power user). As such, the generalization has been done, but at the expense of a breaking change. However, realize that this breaking change would only be breaking behavior that is "random" (and most users should not be doing anyways -- but you never know).

*PREVIOUSLY*

```
gremlin> g.V().match(
......1>   __.as('a').outE('created').order().by('weight',decr).limit(1).inV().as('b'),
......2>   __.as('b').has('lang','java')
......3> ).select('a','b').by('name')
==>[a:marko,b:lop]
```

*CURRENTLY*

```
gremlin> g.V().match(
......1>   __.as('a').outE('created').order().by('weight',decr).limit(1).inV().as('b'),
......2>   __.as('b').has('lang','java')
......3> ).select('a','b').by('name')
==>[a:marko,b:lop]
==>[a:josh,b:ripple]
==>[a:peter,b:lop]
```

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.2.7 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Generalized `MatchStep` to locally compute all clauses with barriers (not just reducing barriers).
 * Ensured that plugins were applied in the order they were configured.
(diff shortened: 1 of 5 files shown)


doanduyhai:
Please remark that the result of _PREVIOUSLY_ is already wrong if we assume that the ordering is global. Indeed the global ordering on `outE('created').values('weight').inV()` should yield {label=software,id=5,lang=java,name=ripple}

image

robertdale:
`docker/build.sh -i -t -n` SUCCESS
VOTE +1

spmallette:
Nice upgrade docs - VOTE +1

PR #703, opened 2017-09-02
TINKERPOP-1756 Added an EmbeddedRemoteConnection.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1756

This enables a "remote" to be mocked so as to be run locally in the same JVM. Not sure if there are other applications for this.

VOTE +1


dkuppitz:
VOTE: +1

robertdale:
VOTE +1

PR #704, opened 2017-09-02
TINKERPOP-1766 Gremlin.Net: Add handling for closed connections

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-1766

This avoids `Connections` from being added back to the `ConnectionPool` when they were closed which happens when the server gets unavailable. In that case all existing `Connections` are removed from the `ConnectionPool` and closed as they are most likely also not `Open` anymore. This is in line with the handling for closed connections in `gremlin-driver`.

Unfortunately, I don't know of a good way to test something like this. We probably would have to make internal classes like `Connection` public which isn't that nice as users shouldn't see them in my opinion.

VOTE +1

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
@@ -101,18 +101,15 @@ namespace Gremlin.Net.Driver
 
         private void RemoveAllConnections()
         {
-            while (!_connections.IsEmpty)
+            while (_connections.TryTake(out var connection))
             {
-                _connections.TryTake(out var connection);
                 connection.Dispose();
             }
         }
 
         private async Task TeardownAsync()
         {
-            var closeTasks = new List<Task>(_connections.Count);
-            closeTasks.AddRange(_connections.Select(conn => conn.CloseAsync()));
-            await Task.WhenAll(closeTasks).ConfigureAwait(false);
+            await Task.WhenAll(_connections.Select(c => c.CloseAsync())).ConfigureAwait(false);
         }
 
         #region IDisposable Support
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
@@ -55,7 +55,7 @@ namespace Gremlin.Net.Driver
             if (connection == null)
                 connection = await CreateNewConnectionAsync().ConfigureAwait(false);
 
-            return new ProxyConnection(connection, AddConnection);
+            return new ProxyConnection(connection, AddConnectionIfOpen);
         }
 
         private async Task<Connection> CreateNewConnectionAsync()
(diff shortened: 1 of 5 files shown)


gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
```diff
@@ -74,6 +85,36 @@ private void AddConnection(Connection connection)
...
+            lock (_connectionsLock)
+            {
+                TeardownAsync().WaitUnwrap();
+                RemoveAllConnections();
+            }
+        }
+
+        private void RemoveAllConnections()
+        {
+            while (!_connections.IsEmpty)
```
jorgebay: Using `TryTake()` and `IsEmpty` can create race conditions.
Use `while (bag.TryTake(out c))` instead.

gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
```diff
@@ -74,6 +85,36 @@ private void AddConnection(Connection connection)
...
+            {
+                _connections.TryTake(out var connection);
+                connection.Dispose();
+            }
+        }
+
+        private async Task TeardownAsync()
+        {
+            var closeTasks = new List<Task>(_connections.Count);
```
jorgebay: The list isn't necessary, you can use the `Task.WhenAll(IEnumerable<Task>)` overload instead: `Task.WhenAll(_connections.Select(c => c.CloseAsync()))`

jorgebay:
I've left some line comments on the pr.

Apart from the issue this patch is addressing, we should create separate tickets for issues related to the .NET driver pool:
- Blocking (use of `lock`).
- An unbounded number of connections can be created under pressure.
- Lack of request pipelining on the same connection.

I wasn't able to focus on the .NET driver itself, my initial priority was getting a stable `GraphTraversal` API and bytecode serialization, but we can tackle these issues progressively in the next versions.

FlorianHockmann:
Thanks for your comments so far, I'll implement the changes you suggested.

I definitely agree that the driver and especially the connection pool needs to be improved and I would really appreciate any effort you (and of course als anyone else) put into this as I'm not exactly an expert in writing database drivers.

Could you clarify what you mean with _request pipelining on the same connection_? If you mean sending multiple requests in parallel over the same websocket connection, then I think that it's not possible as [the documentation](https://docs.microsoft.com/en-us/dotnet/api/system.net.websockets.clientwebsocket.sendasync?view=netframework-4.7#Remarks) states that

> Exactly one send and one receive is supported on each ClientWebSocket object in parallel.

The upside of this is of course that the implementation of the `Connection` and the `ConnectionPool` is much simpler when every `Connection` can just send one single request at a time as we don't have to match incoming responses with the correct requests they belong to.

FlorianHockmann:
I just pushed a commit that addresses your comments, @jorgebay.

jorgebay:
About request pipelining, you can send in a queue one by one regardless of whether or not it was received. On the read side, once a message is received you issue a following call to `ReceiveAsync()` (regardless whether a message has been sent). That way the client can send x messages without waiting for a response.

Should we file a different ticket for the pool improvements?

About the changes in this pr: VOTE +1

FlorianHockmann:
> Should we file a different ticket for the pool improvements?

Yes, I think a dedicated ticket for those improvements is a good idea  and it makes probably sense if you create it as you already made some good suggestions.

jorgebay:
I've filed the tickets 1774 and 1775 for the pool enhancements.

spmallette:
VOTE +1

spmallette:
@FlorianHockmann  When you merge a PR to tp32, please make sure you also merge tp32 to master. Technically it is best to do that locally first, make sure both tp32 and master are building properly, then push master and then tp32. 

Also, the JIRA for this PR still needs to closed. Thanks.

FlorianHockmann:
Thanks for the explanation @spmallette. I was about to ask about the approach for merging pull requests as the documentation only talks about the case where separate pull requests are used for the different release branches. (Maybe we should extend that for simpler cases where only a pull request to tp32 is enough?)

I just tried to merge tp32 to master, but ran into a merge conflict as apparently the changes from TINKERPOP-1744 (PR #690) are missing in master although GitHub shows the merge commit also for the master branch: https://github.com/apache/tinkerpop/commit/6a3fc39a912bec98707dd69d461955330627df10

Do I now have to cherry-pick them to get them into master before I merge tp32 or what is the best approach in cases like this?

I'll close the JIRA ticket as soon as the changes are also in master.

spmallette:
....what the heck?! I wonder what got messed up there. i don't see a merge from jorge from tp32 to master after that, so i guess it never got in there, yet how could it have been ignored? we did tons of merges from tp32 to master after his merge to tp32.

i guess you should cherry-pick them into master and then do your merge. that's really strange. i'm not sure what to make of that.

jorgebay:
It's weird because 6a3fc39a912bec98707dd69d461955330627df10 was merged by @spmallette on master in 75502eef222b0dfc7aa83c2456ab319a9f9dd1a6 (he beat me to it because I was taking too long to do it) and this branch started in 98e4e1b8201f50e1f34458a82c20f2737aa6ae53, which is down the line.

spmallette:
yeah - i remember that now! i wonder how i messed that up.........

FlorianHockmann:
I just cherry-picked the two commits from @jorgebay's PR and merged tp32 into master afterwards. So master should be in a good state now.

PR #705, opened 2017-09-04
make TinkerGraph cloneable

mpollmeier:
most people use tinkergraph for testing, and if there are traversals that manipulate the graph, it's useful to be able to clone it up front, especially for larger graphs.


robertdale:
This doesn't look like a deep copy e.g. modifying a property would be reflected in both graphs.  Which may be fine. I think any limitations should be noted in the clone method comment.

mpollmeier:
That's very true. I just experimented with deep clones and it doesn't look straightforward
* kryo can do it directly, but all referenced classes need a no-arg-constructor, and since there are some coming in from apache commons, it doesn't look like it's feasible. https://github.com/EsotericSoftware/kryo#copyingcloning
* java's builtin ByteArrayOutputStream has a similar problem: all referenced classes need to be marked as `Serializable`
* implementing `clone` for TinkerVertex, TinkerEdge, TinkerProperty etc. is hard because they all reference each other, i.e. you're in a chicken and egg situation. 

I guess we have two options: 
1) accept that it's a shallow copy and add a comment, as you suggested
2) disregard this PR

Thoughts? Any other ideas for making a deep copy? There's always the option to serialise it into graphml/graphson etc., but that's slow for large graphs.

spmallette:
You could deep clone (brand new vertices/edges/properties) of tinkergraph by using detachment. Do something like this:

```text
gremlin> graph1 = TinkerFactory.createModern()
==>tinkergraph[vertices:6 edges:6]
gremlin> graph2 = TinkerGraph.open()
gremlin> graph1.vertices().forEachRemaining { v -> DetachedFactory.detach(v, true).attach(Attachable.Method.create(graph2)) }
gremlin> graph2
==>tinkergraph[vertices:6 edges:0]
gremlin> graph1.edges().forEachRemaining { e -> DetachedFactory.detach(e, true).attach(Attachable.Method.create(graph2)) }
gremlin> graph2
==>tinkergraph[vertices:6 edges:6]
```

I think that gets it, right?

mpollmeier:
Nice! And it does preserve the IDs as well. I'll update the PR. Thanks @spmallette

mpollmeier:
there we go @spmallette @robertdale

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerGraph.java
```diff
@@ -340,6 +342,14 @@ private void validateHomogenousIds(final List<Object> ids) {
         }
     }
 
+    @Override
+    public TinkerGraph clone() {
+        TinkerGraph cloned = new TinkerGraph(configuration);
```
spmallette: please `final` variables as we tend to do that as a style throughout the code base

tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerGraphTest.java
```diff
@@ -592,6 +592,31 @@ public void shouldSerializeWithColorClassResolverToTinkerGraphUsingDeprecatedTin
         }
     }
 
+    @Test
+    public void shouldClone() {
+        final TinkerGraph g = TinkerGraph.open();
+
+        Vertex marko = g.addVertex("name", "marko", "age", 29);
```
spmallette: more need for `final` keyword

tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerGraphTest.java
```diff
@@ -592,6 +592,31 @@ public void shouldSerializeWithColorClassResolverToTinkerGraphUsingDeprecatedTin
...
+        marko.addEdge("knows", stephen);
+
+        final TinkerGraph g2 = g.clone();
+        Vertex michael = g2.addVertex("name", "michael");
+        michael.addEdge("likes", marko);
+        michael.addEdge("likes", stephen);
+        g2.traversal().V().property("newProperty", "someValue").toList();
+        g2.traversal().E().property("newProperty", "someValue").toList();
+
+        assertEquals("original graph should be unchanged", new Long(2), g.traversal().V().count().next());
```
spmallette: perhaps it's a good idea to add an assertion for vertices/edges in the clone showing they are not the same object since it is a deep copy?

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerGraph.java
```diff
@@ -340,6 +342,14 @@ private void validateHomogenousIds(final List<Object> ids) {
         }
     }
 
+    @Override
```
spmallette: A touch of javadoc here to explain that this provide a deep clone of the graph/vertices/edges and preserves ids.

spmallette:
@mpollmeier a few things:

1. could you please retarget this at tp32 - seems like it would be ok to include this feature there
2. update CHANGELOG to mention this change
3. add a brief entry to the upgrade docs to bring attention to this new feature
4. looks like you need to rebase given some conflicts and i made some other comments on the code itself 

generally speaking though, it looks good. nice

okram:
I don't think it is a good idea to implement `TinkerGraph.clone()`. I think, instead, that there could be a "clone utility" that works for all graphs, not just TinkerGraph. E.g.

```
GraphHelper.cloneGraph(original, new)
```

The reasoning of implementing `clone()` just for test cases is weak and doesn't warrant the semantic confusion that this would entail (as no other graphs are cloneable).

spmallette:
`clone()` won't work well for all graphs though. Making it a generalized utility for `clone()` just seems to invite more trouble than having it just available to TinkerGraph where it at least makes some sense. Cloning a JanusGraph or  DseGraph instance, - weird - but `clone()` a small in-memory TinkerGraph - makes more sense. 

Any reason you think that `clone()` on TinkerGraph is different from a TinkerGraph specific method like `clear()`?

okram:
`TinkerGraph.clear()` is bad too. It is `g.V().drop()`.

spmallette:
hmm - and then a TinkerGraph strategy that converts `g.V().drop()`  to a `clear()` operation somehow (to avoid overhead of iterating all the vertices? 

by that logic is `clone()` really `g.E().subgraph('sg').cap('sg')`?

okram:
Yes. Exactly.

Here is the thing is -- I don't really care for TinkerPop3 anymore so it can go as people want it too, I won't fight. But if you want to start aligning concepts so TinkerPop4 is done "right," we need to start getting away from methods on `Graph` as users will not have access to `Graph` only `Connections`. Thus, every action/process is a `Traversal` and thus, all notions of cloning, clearing, etc. should be via compilation/strategy_introspection of the `Traversal`.

"But Marko, what about schema stuff like `createIndex`? How will that work in TinkerPop4" ...  That is a vendor specific schema DSL. `schema.index("name")`, etc.

I think instead of just going "willy nilly" with TinkerPop3, we should start to make decisions on its design that will align best with TinkerPop4 so we start to get the right thinking in place now for the big effort to come.

mpollmeier:
Ok gents, next iteration is ready for review:
* addressed stephen's comments, rebased and changed target to tp32
* moved the impl to GraphHelper which lives next to ElementHelper in org.apache.tinkerpop.gremlin.structure.util. Note: I added a dependency with scope `test` to gremlin-core
* updated CHANGELOG and upgrade docs

mpollmeier:
hmm, since I introduced a dependency from gremlin-core:test to tinkergraph-gremlin the build complains about a cyclic dependency. I thought that would have been fine, given that it's just a test dependency. Any suggestions how to best resolve that without scrapping the test?

robertdale:
That dependency cycle is bad.  It should probably be put in `gremlin-test`. Maybe even make it a [Graph Feature](http://tinkerpop.apache.org/docs/current/reference/#_features) - Cloning.

mpollmeier:
@spmallette addressed your comments. 
VOTE +1

docs/src/upgrade/release-3.2.x-incubating.asciidoc
```diff
@@ -88,6 +88,11 @@ to the list of locally computed clauses.
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-1764[TINKERPOP-1764]
 
+Clone a graph
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
robertdale: These `^` need to be of the same length as the preceding title line otherwise it won't render correctly.

mpollmeier: fixed

robertdale:
VOTE +1

docs/src/upgrade/release-3.2.x-incubating.asciidoc
```diff
@@ -88,6 +88,11 @@ to the list of locally computed clauses.
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-1764[TINKERPOP-1764]
 
+Clone a graph
+^^^^^^^^^^^^^
+If you want to quickly clone a graph (e.g. because you have set it up in a test and your traversals manipulate the graph) you can now use `GraphHelper.cloneElements(Graph original, Graph clone)`.
```
spmallette: Please mention that this new class is in `gremlin-test` - "...you can now use `GraphHelper.cloneElements(Graph original, Graph clone)` which is part of the `gremlin-test` module."

CHANGELOG.asciidoc
```diff
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 TinkerPop 3.2.7 (Release Date: NOT OFFICIALLY RELEASED YET)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* added `GraphHelper.cloneElements(Graph original, Graph clone)` to quickly clone a graph
```
spmallette: minor - upper case "A" in "added" and end with a period.

spmallette:
Added a couple of minor nits as comments, but otherwise:

VOTE +1

@mpollmeier if you vote yourself, you'll have your 3 +1s and can merge. please be sure to merge tp32 to master as well.

mpollmeier:
Thanks @spmallette. I already voted +1 above. Made those changes, merged (--no-ff) to tp32 and then merged tp32 to master (again --no-ff).

PR #707, opened 2017-09-07
TINKERPOP-1768 Bump to Jackson 2.8.10

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1768

Adds some key bug fixes.

All tests pass with `docker/build.sh -t -n -i` for both tp32 and master

VOTE +1


dkuppitz:
VOTE: +1

robertdale:
Did it fix https://issues.apache.org/jira/browse/TINKERPOP-1509 ?

spmallette:
I don't think so, but I think I didn't go too deep into it. I just removed the ignore annotation on the test and it still failed. I think we needed this fix anyway though for something else, though I can't recall what immediately. I'll have to dig into it when I look at TINKERPOP-1509 directly.

robertdale:
VOTE +1

PR #709, opened 2017-09-11
TINKERPOP-1761: Cancel script evaluation timeout

ko-mueller:
https://issues.apache.org/jira/browse/TINKERPOP-1761

I basically added the code again which was wrongly removed in TINKERPOP-1714, plus added a test. All unit tests passed and I confirmed manually that this fixes the issue I experienced.


dkuppitz:
VOTE: +1

spmallette:
@ko-mueller thanks for submitting this. I just wanted to let you know that it may take some time to get this merged. Every time this area of code gets changed there is suffering, so I'd like to be sure adding this code back doesn't create some new subtle issue that we solved with TINKERPOP-1714.

ko-mueller:
Alright, it just would be nice to have this fixed in the next release. I didn't upgrade from 3.2.5 to 3.3.0 because of this. It's not a big issue for me right now, just that I'd like to stay current (without implementing a work-around for this).

spmallette:
@BrynCooke i've been looking at this on and off since it was submitted and i can't come up with any concrete reason not to vote to merge this. i think we still get to release the {{evaluationFuture}} for GC with this added blob of code that handles cancellation of the cancellation. any last words?

robertdale:
Looks good.
VOTE +1

spmallette:

VOTE +1

PR #712, opened 2017-09-12
TINKERPOP-1752: Gremlin.Net: Generate completely type-safe methods

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-1752

All steps are now (basically) type-safe and the original argument names from Gremlin-Java are used. However, we currently don't support some Java types like `Comparator`. Those were simply replaced by `object` until we find a better solution. A problem with this workaround is that some overloads from Gremlin-Java are not supported in Gremlin.Net as they would result in the same method signature.

The implementation of `Bindings` is now basically the same as in Gremlin-Java. This change was necessary as `Bindings.Of()` can no longer return a `Binding` object

These changes also revealed a bug in the tests where the `WithoutStrategies` source step was called with objects of strategies instead of just with their types. However, `WithoutStrategies` still can't work right now as a GraphSON serializer is missing for `Type`.

The `Group` step for example looks now like this:
```cs
/// <summary>
///     Adds the group step to this <see cref="GraphTraversal{SType, EType}" />.
/// </summary>
public GraphTraversal< S , IDictionary<K, V> > Group<K, V> ()
{
    Bytecode.AddStep("group");
    return Wrap< S , IDictionary<K, V> >(this);
}

/// <summary>
///     Adds the group step to this <see cref="GraphTraversal{SType, EType}" />.
/// </summary>
public GraphTraversal< S , E > Group (string sideEffectKey)
{
    Bytecode.AddStep("group", sideEffectKey);
    return Wrap< S , E >(this);
}
```
So this should also solve [TINKERPOP-1751](https://issues.apache.org/jira/browse/TINKERPOP-1751) that mentioned the missing overload for `Group` that returns `IDictionary<K, V>`.

### DIFF
--- a/gremlin-dotnet/glv/GraphTraversal.template
+++ b/gremlin-dotnet/glv/GraphTraversal.template
@@ -65,9 +65,17 @@ namespace Gremlin.Net.Process.Traversal
         /// <summary>
         ///     Adds the <%= method.methodName %> step to this <see cref="GraphTraversal{SType, EType}" />.
         /// </summary>
-        public GraphTraversal< <%= method.t1 %> , <%= method.t2 %> > <%= toCSharpMethodName.call(method.methodName) %><%= method.tParam %> (params object[] args)
+        public GraphTraversal< <%= method.t1 %> , <%= method.t2 %> > <%= toCSharpMethodName.call(method.methodName) %><%= method.tParam %> (<%= method.parameters %>)
         {
-            Bytecode.AddStep("<%= method.methodName %>", args);
+        <%  if (method.parameters.contains("params ")) {
+      %>    var args = new List<object> {<%= method.paramNames.init().join(", ") %>};
+            args.AddRange(<%= method.paramNames.last() %>);
+            Bytecode.AddStep("<%= method.methodName %>", args.ToArray());<%
+            }
+            else {
+      %>    Bytecode.AddStep("<%= method.methodName %>"<% if (method.parameters) out << ', '+ method.paramNames.join(", ") %>);<%
+            }
+        %>
             return Wrap< <%= method.t1 %> , <%= method.t2 %> >(this);
         }
 <% } %>
--- a/gremlin-dotnet/glv/GraphTraversal.template
+++ b/gremlin-dotnet/glv/GraphTraversal.template
@@ -68,7 +68,7 @@ namespace Gremlin.Net.Process.Traversal
         public GraphTraversal< <%= method.t1 %> , <%= method.t2 %> > <%= toCSharpMethodName.call(method.methodName) %><%= method.tParam %> (<%= method.parameters %>)
         {
         <%  if (method.parameters.contains("params ")) {
-      %>    var args = new List<object> {<%= method.paramNames.init().join(", ") %>};
+      %>    var args = new List<object>(<%= method.paramNames.init().size() %> + <%= method.paramNames.last() %>.Length) {<%= method.paramNames.init().join(", ") %>};
             args.AddRange(<%= method.paramNames.last() %>);
             Bytecode.AddStep("<%= method.methodName %>", args.ToArray());<%
             }
--- a/gremlin-dotnet/src/Gremlin.Net/Process/Traversal/Bindings.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Process/Traversal/Bindings.cs
@@ -21,6 +21,9 @@
 
 #endregion
 
+using System.Collections.Generic;
+using System.Threading;
+
 namespace Gremlin.Net.Process.Traversal
 {
     /// <summary>
@@ -29,14 +32,42 @@ namespace Gremlin.Net.Process.Traversal
     public class Bindings
     {
         /// <summary>
+        ///     Gets an instance of the <see cref="Bindings" /> class.
+        /// </summary>
+        public static Bindings Instance { get; } = new Bindings();
+
+        private static readonly ThreadLocal<Dictionary<object, string>> BoundVariableByValue =
+            new ThreadLocal<Dictionary<object, string>>();
+
+        /// <summary>
         ///     Binds the variable to the specified value.
         /// </summary>
         /// <param name="variable">The variable to bind.</param>
         /// <param name="value">The value to which the variable should be bound.</param>
         /// <returns>The bound value.</returns>
-        public Binding Of(string variable, object value)
+        public TV Of<TV>(string variable, TV value)
         {
-            return new Binding(variable, value);
+            var dict = BoundVariableByValue.Value;
+            if (dict == null)
+            {
+                dict = new Dictionary<object, string>();
+                BoundVariableByValue.Value = dict;
+            }
+            dict[value] = variable;
+            return value;
+        }
+
+        internal static string GetBoundVariable<TV>(TV value)
+        {
+            var dict = BoundVariableByValue.Value;
+            if (dict == null)
+                return null;
+            return !dict.ContainsKey(value) ? null : dict[value];
+        }
+
+        internal static void Clear()
+        {
+            BoundVariableByValue.Value?.Clear();
         }
     }
 }
--- a/gremlin-dotnet/src/Gremlin.Net/Process/Traversal/Bindings.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Process/Traversal/Bindings.cs
@@ -62,7 +62,8 @@ namespace Gremlin.Net.Process.Traversal
             var dict = BoundVariableByValue.Value;
             if (dict == null)
                 return null;
-            return !dict.ContainsKey(value) ? null : dict[value];
+            dict.TryGetValue(value, out var variable);
+            return variable;
         }
 
         internal static void Clear()
--- a/gremlin-dotnet/src/Gremlin.Net/Process/Traversal/Bytecode.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Process/Traversal/Bytecode.cs
@@ -21,7 +21,10 @@
 
 #endregion
 
+using System;
+using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Gremlin.Net.Process.Traversal
 {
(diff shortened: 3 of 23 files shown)


spmallette:
@FlorianHockmann sorry but it looks like you have to rebase this one as I just merged #713 which moved the groovy scripts out of the pom.xml.  i didn't realize that change was going to affect this PR. I'll try to review as soon as you do that so that this PR doesn't go stale again. thanks.

FlorianHockmann:
@spmallette No problem, I could also have seen that by myself. To be honest, I just looked at the title of that PR and didn't recognize that it was about simplifying the GLV generation. In general, I really like the changes you introduced there as writing that much Groovy code inside of a pom.xml wasn't really nice.

FlorianHockmann:
Rebased on tp32, but now one of the two CI builds failed due to a failing test in `spark-gremlin`: `SparkContextStorageCheck.shouldSupportDirectoryFileDistinction`. From what I can see, the failing assert checks whether 10 files were actually created and here `ls` only returns 8 files.
I have honestly no idea how my changes could cause this test to fail as I didn't modify anything outside of `gremlin-dotnet`.

spmallette:
Travis is a flakey - I tend to think of a failure as a warning especially when one passes and one fails.  I wouldn't worry about it. If you did your own build and all is good then no problem.

FlorianHockmann:
Ok, good to know.
In addition to the Gremlin.Net tests, all tests pass with `docker/build.sh -t -i`.

jorgebay:
I wont be able to review it today, I'll probably have some time to look into this early next week.

spmallette:
Changes look generally ok to me, but I'll let @jorgebay do a more technical review of the code itself as i'm still trying to get my head back into .NET. Getting rid of the `params object[] args` seems like a helpful thing to do for users. Based on that: 

VOTE +1

gremlin-dotnet/glv/GraphTraversal.template
```diff
@@ -65,9 +65,17 @@ namespace Gremlin.Net.Process.Traversal
         /// <summary>
         ///     Adds the <%= method.methodName %> step to this <see cref="GraphTraversal{SType, EType}" />.
         /// </summary>
-        public GraphTraversal< <%= method.t1 %> , <%= method.t2 %> > <%= toCSharpMethodName.call(method.methodName) %><%= method.tParam %> (params object[] args)
+        public GraphTraversal< <%= method.t1 %> , <%= method.t2 %> > <%= toCSharpMethodName.call(method.methodName) %><%= method.tParam %> (<%= method.parameters %>)
         {
-            Bytecode.AddStep("<%= method.methodName %>", args);
+        <%  if (method.parameters.contains("params ")) {
+      %>    var args = new List<object> {<%= method.paramNames.init().join(", ") %>};
```
jorgebay: We should set the initial capacity of the list to avoid unnecessary resizing of the underlying array.
It should be the sum of the length of `method.paramNames.init()` and `method.paramNames.last()`.

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/Bytecode.cs
```diff
@@ -79,7 +85,77 @@ public void AddSource(string sourceName, params object[] args)
         /// <param name="args">The traversal method arguments.</param>
         public void AddStep(string stepName, params object[] args)
         {
-            StepInstructions.Add(new Instruction(stepName, args));
+            StepInstructions.Add(new Instruction(stepName, FlattenArguments(args)));
+            Bindings.Clear();
```
jorgebay: Even though `Bindings` is based on a `ThreadLocal<T>` instance, I think the implementation is not thread-safe.

For example: Multiple tasks executed serially on the same thread, modifying the same bindings dictionary.
Besides not being thread-safe, it doesn't support defining a binding on 1 thread and adding the step on another, sample:

```csharp
// thread 1
// define bindings
Bindings.Instance.Of(/* */);
await SomeUnRelatedTask();
// thread 2
// Do something with that binding
g.V()/**/;
```

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/Bindings.cs
```diff
@@ -29,14 +32,42 @@ namespace Gremlin.Net.Process.Traversal
...
+            dict[value] = variable;
+            return value;
+        }
+
+        internal static string GetBoundVariable<TV>(TV value)
+        {
+            var dict = BoundVariableByValue.Value;
+            if (dict == null)
+                return null;
+            return !dict.ContainsKey(value) ? null : dict[value];
```
jorgebay: Use `Dictionary{K, V}.TryGetValue()` to save one operation.

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/Bytecode.cs
FlorianHockmann: This is definitely a problem, but do you see a way to fix it? Unfortunately, I think we have to use `static` here unless we want to create overloads for each parameter in the Traversal API that takes a `Binding` instead of the unbound variable which would result in a lot of overloads.

Since [`Bindings` don't seem to improve the performance with Bytecode](https://lists.apache.org/thread.html/c10baff72e67a8b7728c42b5c4e983a83e82b8f6964b3aa465f0e341@<dev.tinkerpop.apache.org>) when no lambdas are used which we don't support anyway, we may also remove the support for `Bindings` altogether from Gremlin.Net. Then we also don't need the `FlattenArguments` and `ConvertArgument` functions in `Bytecode` which are not exactly intuitive.

Or if we want to continue supporting them, we could also add a note to the documentation to make users aware of these concurrency problems with `Bindings`.

jorgebay: I agree that a solution could be an overkill on the API...

I'm +1 from not supporting bindings in the .NET GLV (and document it), I don't see much purpose.

What do you think about not supporting Bindings on the .NET GLV @okram @spmallette @robertdale ?

spmallette: Any reason we don't support lambdas? Even if .NET can't support them natively for some reason wouldn't we minimally support the ability to pass a python/groovy/etc lambda? it's kinda weird that way, but i think back to the point that kuppitz made on the dev list the other day where he stated that he doesn't always find a way out of using lambdas in production systems he works on - so ultimately users will need that kind of capability i think.

FlorianHockmann: That's a good point. So we should probably support lambdas. To be honest, I haven't really looked into what would be necessary to support lambdas. If I'm not mistaken then we can easily support Python and Groovy lambdas when we add a dedicated GraphSON serializer for them and probably a class that wraps the lambda together with its language. Then the steps that take a lambda would simply accept an object of that class in Gremlin.Net. I can create a separate ticket for this when we agree that we want to support them.

When we want to support lambdas then we should also support `Bindings`. That leaves us with the question of whether we find a better implementation that doesn't suffer from the concurrency problems @jorgebay mentioned or if the current implementation in this pull request is acceptable.

Apart from that, shouldn't we clarify the section in the documentation about bindings? It currently states for example [for gremlin-python](http://tinkerpop.apache.org/docs/current/reference/#_bindings):

> If a traversal is going to be executed repeatedly, but with different parameters, then bindings should be used.

because:
> translation and compilation can be skipped as the previously compiled version should be cached

which is only true when a lambda is used if I understood it correctly.

FlorianHockmann:
I finally found some time to work on this again and fixed the issues mentioned by @jorgebay. However, I left the `Bindings` implementation unchanged despite the problems with concurrent access as it seems to be still the best solution. (I'm of course open for suggestions on how this can be improved.)

I also noticed that my changes broke the `WithoutStrategies` source step as that now correctly expects to get the `Types` of the Strategies to exclude for which Gremlin.Net had no serializer. So I added a serializer that works basically like the respective one in gremlin-python as it also simply creates an object of the `Type` and then this object will be serialized as before. A unit test ensures that all Strategies have a parameterless constructor as we can't serialize their `Type` otherwise.
Honestly, I was a bit surprised that I had to serialize `Types` by serializing an empty object of that `Type` although the IO docs show that the GraphSON type `g:Class` [can be serialized like this](http://tinkerpop.apache.org/docs/3.3.0/dev/io/#_class):
```json
{
  "@type" : "g:Class",
  "@value" : "java.io.File"
}
```
but the Gremlin Server couldn't deserialize the Strategy class when I serialized it like this. So is the documentation wrong here? @spmallette: Could you clarify my confusion here?
Also the IO docs don't mention how `TraversalStrategies` are serialized in general. Should we add that?

The build is currently failing, but that seems to be caused by travis-ci/travis-ci#8607. I built Gremlin.Net locally and executed the tests without any problems.

BTW: Would it make sense to create a separate pull request for `master` or can we simply execute `generate.groovy` later when this is merged from `tp32` into `master`?

spmallette:
Looks like traversal strategy classes have their own serialization forms....................

```json
{
	"@type": "g:Bytecode",
	"@value": {
		"source": [
			["withStrategies", {
				"@type": "g:SubgraphStrategy",
				"@value": {
					"vertices": {
						"@type": "g:Bytecode",
						"@value": {
							"step": [
								["has", "name", "marko"]
							]
						}
					}
				}
			}]
		]
	}
}
```

I didn't know that. I guess the GraphSON spec in the IO docs (and testing framework) is incomplete. :rage1:

jorgebay:
VOTE +1

jorgebay:
All .NET tests passes with: 
```
mvn clean install -DskipTests
cd gremlin-dotnet && mvn clean install -P gremlin-dotnet
```

I'm running a `docker/build.sh -t -n -i`

spmallette:
VOTE +1

spmallette:
@jorgebay did the docker build ever finish successfully?

jorgebay:
Yes it builds, this PR is good to go on my end!

Sorry I forgot to confirm on Friday :/

jorgebay:
@FlorianHockmann Can we merge this?

FlorianHockmann:
> @FlorianHockmann Can we merge this?

Sure, I was on vacation, hence the delay. I'll merge it today.

FlorianHockmann:
I had to do a few changes during the merge into master which can be seen [in this commit](https://github.com/apache/tinkerpop/commit/ab2fd2cadd64a7af1438977ad9d31e9feff16eec), but now everything looks good and both branches build locally.

PR #713, opened 2017-09-13
TINKERPOP-1779 Bump to GMavenPlus 1.6

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1779

Moved embedded groovy scripts that were in the pom to their own files. They are more manageable that way as you get some syntax highlighting/intellisense in the IDE when they have the right file extension. It also makes the pom a bit easier to follow. Removed a bunch of python test resources that were no longer in use. Centralized the scripts that start/stop Gremlin Server for GLV tests so that we don't have them copy/pasted everywhere.

VOTE +1


okram:
VOTE +1

dkuppitz:
VOTE: +1

PR #714, opened 2017-09-13
TINKERPOP-1782 RangeByIsCountStrategy doesn't handle floating point numbers properly

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1782

Fixed a bug in `RangeByIsCountStrategy` that led to unexpected behaviors when predicates were used with floating point numbers.

VOTE: +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Added `GraphHelper.cloneElements(Graph original, Graph clone)` to the `gremlin-test` module to quickly clone a graph.
 * Bump to GMavenPlus 1.6.
 * Added better error message for illegal use of `repeat()`-step.
+* Fixed a bug in `RangeByIsCountStrategy` that led to unexpected behaviors when predicates were used with floating point numbers.
 * Bump to Jackson 2.8.10.
 * Added an `EmbeddedRemoteConnection` so that it's possible to mimic a remote connection within the same JVM.
 * Supported interruption for remote traversals.


gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/RangeByIsCountStrategyTest.java
```diff
@@ -99,6 +99,9 @@
                 {__.and(__.out().count().is(0), __.in().count().is(1)), __.and(__.not(__.out()), __.in().limit(2).count().is(1))},
                 {__.and(__.out().count().is(1), __.in().count().is(0)), __.and(__.out().limit(2).count().is(1), __.not(__.in()))},
                 {__.or(__.out().count().is(0), __.in().count().is(0)), __.or(__.not(__.out()), __.not(__.in()))},
+                {__.path().filter(__.count().is(gte(0.5))).limit(5), __.path().identity().limit(5)}, // unfortunately we can't just remove the filter step
```
okram: Why can't you remove the `filter()` step? That is, why the `identity()` step replacement?

dkuppitz:
It breaks other parts of the code when you attempt to modify the parent traversal, in particular when you remove one or more steps. I don't remember where the exception was thrown, but it was an IndexOutOfBoundsException. So apparently there's a for-loop with an upper bound that is only initialized once (which is good, I wouldn't like to recount the steps in each iteration).

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/RangeByIsCountStrategyTest.java
dkuppitz: It breaks other parts of the code when you attempt to modify the parent traversal, in particular when you remove one or more steps. I don't remember where the exception was thrown, but it was an `IndexOutOfBoundsException`. So apparently there's a `for` loop with an upper bound that is only initialized once (which is good, I wouldn't like to recount the steps in each iteration).

robertdale:
Tests in the jira ticket work as expected.
VOTE +1

okram:
VOTE +1

PR #716, opened 2017-09-18
TINKERPOP-1785 Added strong name signing for .NET components

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1785

I guess we need strong naming for legacy support of this Microsoft feature. 

VOTE +1


jorgebay:
The strong name for Gremlin.Net.dll assembly (generated using Gremlin.Net.csproj) is missing, we should add it in the template as well.

spmallette:
ohhhhhh - smart. no wonder the PR looks "wrong".

FlorianHockmann:
Thanks for doing this, @spmallette!

LGTM, VOTE +1

jorgebay:
We forgot to consider that, as we are strong named assembly, we must avoid changing the assembly version for patch versions to minimize the need of user-defined binding redirects.

In our case, it would mean to set a fix assembly version within the same major and minor. The generated Gremlin.Net.csproj should contain the 3 following xml tags:

```xml
<!--  Version: 3.2.7   -->
<Version>$projectVersion</Version>

<!--  FileVersion: 3.2.7.0   -->
<FileVersion>$projectVersion.0</FileVersion>

<!--  AssemblyVersion: 3.2.0.0  -->
<AssemblyVersion>$projectMajorMinorVersion.0.0</AssemblyVersion>
```

spmallette:
Thanks @jorgebay - i wasn't aware of that issue. How does it look now?

gremlin-dotnet/glv/Gremlin.Net.csproj.template
```diff
@@ -25,7 +25,9 @@ limitations under the License.
   </PropertyGroup>
 
   <PropertyGroup Label="Package">
-    <Version>${projectVersion}</Version>
+    <Version>$projectVersion</Version>
+    <FileVersion>$projectVersion.0</FileVersion>
```
jorgebay: `FileVersion` only accepts `major.minor.patch.build` format, we need to strip any other character like `-SNAPSHOT`.

FlorianHockmann: A good option to simplify this might be with the `VersionPrefix` and `VersionSuffix` elements:
```
<VersionPrefix>3.2.7</VersionPrefix>
<VersionSuffix>rc1</VersionSuffix>
```
Then we can use the `VersionPrefix` directly for the `FileVersion`.

spmallette: ok - fixed `<FileVersion>` - i just did some regex. i didn't quite understand the suggestion that @FlorianHockmann made, but I can adjust if you guys determine it is right to do so. at least it builds now.

FlorianHockmann: My suggestion was to replace this:
```xml
<Version>3.2.7-SNAPSHOT</Version>
<FileVersion>3.2.7.0</FileVersion>
<AssemblyVersion>3.2.0.0</AssemblyVersion>
```
by this:
```xml
<VersionPrefix>3.2.7</VersionPrefix>
<VersionSuffix>SNAPSHOT</VersionSuffix>
<FileVersion>3.2.7.0</FileVersion>
<AssemblyVersion>3.2.0.0</AssemblyVersion>
```
as the second option seems to be easier to generate, but both are completely valid and we can stay with the first option when you already managed to adapt the generation for that.

spmallette: ok - thanks. as it was pretty straightforward to do it as i did it, i'll just leave it as is with the regex solution.

jorgebay:
`dotnet pack gremlin-dotnet/src/Gremlin.Net/Gremlin.Net.csproj -c  Release` generates the package containing with assembly with the intended attributes.

VOTE +1

PR #717, opened 2017-09-18
TINKERPOP-1783: PageRank gives incorrect results for graphs with sinks

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1783

There were a couple of problems with `PageRankVertexProgram`. In order to describe the problems, its necessary to explain PageRank.

***

Initially, every vertex gets `1.0/vertexCount` amount of energy. Thus, the total energy in the system is 1.0.

At every iteration, 0.85 of that energy is distributed amongst adjacent neighbors and 0.15 is distributed amongst all vertices (this is the `alpha` parameter w/ the default being 0.85). If a vertex does not have outgoing edges and it can not distribute its 0.85 energy amongst its neighbors, then all that energy is teleported amongst all vertices.

The algorithm above ensures strong connectivity because the 0.15 energy distribution simulates a strongly connected graph via the concept of "teleportation." According to a Markov chain proof on ergodicity in strongly connected graphs, PageRank yields a positive eigenvector. That eigenvector is your rank vector and that rank vector's sum total is always the initial energy the system at iteration 0.

***

I fixed three things in this PR:

1. Vertex count is now determined in iteration 0 and no longer needs users to provide a vertex count.
2. Teleportation energy (global energy) is defined via a `Memory` variable.
3. Isolated vertices send their energy to the teleportation variable.

Here are the results compared with iGraph's implementation (note that iGraph does a convergence check while TinkerPop does not. It would be extremely expensive to test convergence in a distributed system -- thus, slightly different results).

```
VERTEX  iGRAPH    TINKERPOP
marko   0.1119788 0.11375485828040575
vadas   0.1370267 0.14598540145985406
lop     0.2665600 0.30472082661863686
josh    0.1620746 0.14598540145985406
ripple  0.2103812 0.1757986539008437
peter   0.1119788 0.11375485828040575
```

Finally, a proof of energy conservation:

```
gremlin> 0.11375485828040575 + 0.14598540145985406 + 0.30472082661863686 + 0.14598540145985406 + 0.1757986539008437 + 0.11375485828040575
==>1.00000000000000018
```

VOTE +1

### DIFF
--- a/docs/src/upgrade/release-3.3.x.asciidoc
+++ b/docs/src/upgrade/release-3.3.x.asciidoc
@@ -32,6 +32,39 @@ Please see the link:https://github.com/apache/tinkerpop/blob/3.3.1/CHANGELOG.asc
 Upgrading for Users
 ~~~~~~~~~~~~~~~~~~~
 
+PageRankVertexProgram
+^^^^^^^^^^^^^^^^^^^^^
+
+There were two major bugs in the way in which PageRank was being calculated in `PageRankVertexProgram`. First, teleportation
+energy was not being distributed correctly amongst the vertices at each round. Second, terminal vertices (i.e. vertices
+with no outgoing edges) did not have their full gathered energy distributed via teleportation. As a secondary benefit,
+`PageRankVertexProgram` also always calculates the vertex count and no longer requires the user to specify the vertex count.
+
+For users upgrading, note that while the relative rank orders will remain "the same," the actual PageRank values will differ
+from prior TinkerPop versions.
+
+```
+VERTEX  iGRAPH    TINKERPOP
+marko   0.1119788 0.11375485828040575
+vadas   0.1370267 0.14598540145985406
+lop     0.2665600 0.30472082661863686
+josh    0.1620746 0.14598540145985406
+ripple  0.2103812 0.1757986539008437
+peter   0.1119788 0.11375485828040575
+```
+
+Normalization preserved through computation:
+
+```
+0.11375485828040575 +
+0.14598540145985406 +
+0.30472082661863686 +
+0.14598540145985406 +
+0.1757986539008437 +
+0.11375485828040575
+==>1.00000000000000018
+```
+
 IO Defaults
 ^^^^^^^^^^^
 
--- a/docs/src/upgrade/release-3.3.x.asciidoc
+++ b/docs/src/upgrade/release-3.3.x.asciidoc
@@ -69,6 +69,8 @@ Two other additions to `PageRankVertexProgram` were provided as well.
 1. It now calculates the vertex count and thus, no longer requires the user to specify the vertex count.
 2. It now allows the user to leverage an epsilon-based convergence instead of having to specify the number of iterations to execute.
 
+See: link:https://issues.apache.org/jira/browse/TINKERPOP-1783[TINKERPOP-1783]
+
 IO Defaults
 ^^^^^^^^^^^
 
--- a/docs/src/upgrade/release-3.3.x.asciidoc
+++ b/docs/src/upgrade/release-3.3.x.asciidoc
@@ -37,8 +37,7 @@ PageRankVertexProgram
 
 There were two major bugs in the way in which PageRank was being calculated in `PageRankVertexProgram`. First, teleportation
 energy was not being distributed correctly amongst the vertices at each round. Second, terminal vertices (i.e. vertices
-with no outgoing edges) did not have their full gathered energy distributed via teleportation. As a secondary benefit,
-`PageRankVertexProgram` also always calculates the vertex count and no longer requires the user to specify the vertex count.
+with no outgoing edges) did not have their full gathered energy distributed via teleportation.
 
 For users upgrading, note that while the relative rank orders will remain "the same," the actual PageRank values will differ
 from prior TinkerPop versions.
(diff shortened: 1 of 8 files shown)


dkuppitz:
VOTE: +1

spmallette:
does this fix only go to master for some reason? or can it also go on tp32?

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/computer/ranking/pagerank/PageRankVertexProgramTest.java
```diff
@@ -38,7 +38,7 @@
     @LoadGraphWith(MODERN)
     public void shouldExecutePageRank() throws Exception {
         if (graphProvider.getGraphComputer(graph).features().supportsResultGraphPersistCombination(GraphComputer.ResultGraph.NEW, GraphComputer.Persist.VERTEX_PROPERTIES)) {
-            final ComputerResult result = graph.compute(graphProvider.getGraphComputer(graph).getClass()).program(PageRankVertexProgram.build().create(graph)).submit().get();
+            final ComputerResult result = graph.compute(graphProvider.getGraphComputer(graph).getClass()).program(PageRankVertexProgram.build().iterations(30).create(graph)).submit().get();
```
dkuppitz: What about a test for convergence-based execution?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/traversal/step/map/PageRankVertexProgramStep.java
```diff
@@ -47,7 +47,7 @@
 
     private PureTraversal<Vertex, Edge> edgeTraversal;
     private String pageRankProperty = PageRankVertexProgram.PAGE_RANK;
-    private int times = 30;
+    private int times = -2;
```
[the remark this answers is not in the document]
okram: No, cause of how iterations in `GraphComputer` differ from iterations in PageRank. We immediately get a `+1` and I need it to be `-1`, not `0`.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/ranking/pagerank/PageRankVertexProgram.java
```diff
@@ -155,61 +161,61 @@ public PageRankVertexProgram clone() {
...
+        memory.set(CONVERGENCE_ERROR, 0.0d);
+        return -1 == this.totalIterations ?
+                memory.<Double>get(CONVERGENCE_ERROR) < this.epsilon :
+                memory.getIteration() >= this.totalIterations;
     }
 
     @Override
     public String toString() {
-        return StringFactory.vertexProgramString(this, "alpha=" + this.alpha + ", iterations=" + this.totalIterations);
+        return StringFactory.vertexProgramString(this, "alpha=" + this.alpha + ", iterations=" + this.totalIterations + ", epsilon=" + this.epsilon);
```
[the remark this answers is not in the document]
okram: Done.

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/computer/ranking/pagerank/PageRankVertexProgramTest.java
okram: There are plenty of `pageRank()`-tests with some testing convergence and some testing `times()`. See `PageRankTest`. What other test do you have in mind?

okram:
@spmallette -- I'm putting this into `TP 3.3.x` because it is a breaking change. I believe it is too late into `3.2.x` to introduce such changes.

okram:
You know what we should do. It should be `maxIterations`. This way, if it converges before hand, great. Done. If it doesn't, well, it will go for `maxIterations` and then stop. This way `epsilon` and `maxIterations` are not mutually exclusive settings.

okram:
Pushed `maxIterations` concept.

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/computer/ranking/pagerank/PageRankVertexProgramTest.java
dkuppitz: I was thinking more about a test, that verifies, that there were X iterations (where X is a pre-determined number). Now, with the `maxIterations` concept, it would make even more sense to verify that the PR VP terminates early.

docs/src/upgrade/release-3.3.x.asciidoc
```diff
@@ -32,6 +32,43 @@ Please see the link:https://github.com/apache/tinkerpop/blob/3.3.1/CHANGELOG.asc
...
+0.1757986539008437 +
+0.11375485828040575
+==>1.00000000000000018
+```
+
+Two other additions to `PageRankVertexProgram` were provided as well.
+
+1. It now calculates the vertex count and thus, no longer requires the user to specify the vertex count.
+2. It now allows the user to leverage an epsilon-based convergence instead of having to specify the number of iterations to execute.
+
```
spmallette: Please include the "See: ...." link to the JIRA issue.

okram: Done.

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/computer/ranking/pagerank/PageRankVertexProgramTest.java
okram: Pushed. Added a test for iterations breaking, epsilon breaking, and energy conservation.

dkuppitz:
VOTE: +1

spmallette:

VOTE +1

PR #723, opened 2017-09-26
TINKERPOP-1792 Fixed GremlinScriptEngine bug in lambda processing

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1792

GremlinScriptEngine instances should now properly select the appropriate TraversalSource from bindings rather than select a random one. The TraversalSource is locally aliased to a "hidden" variable at the time of evaluation to avoid naming clashes with local variables defined in the bytecode itself. Added a number of validations to be sure that users get appropriate errors if they try to use the eval() method the wrong way. Included new tests to ensure that GremlinScriptEngine have the appropriate behavior.

VOTE +1


okram:
VOTE +1

dkuppitz:
Nice set of tests.

VOTE: +1

PR #726, opened 2017-10-02
TINKERPOP-1795: Getting Lambda comparator message for .profile() step

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1795

Some objects are hard to check for "lambdaness" and require a `toString()` analysis. We checked for `@` but that doesn't work for named steps within `order()` (when using `profile()`). That is, its error prone. I came up with a new way to check for lambdas that works for Java, Groovy, and Python/Jython. 

@spmallette -- can you please verify that a "toString()" of a lambda in C# will hold as well for the pattern.

VOTE +1


spmallette:

VOTE +1

dkuppitz:
VOTE: +1

PR #728, opened 2017-10-03
Do not strong-freeze dependencies

aboudreault:
In setup.py, strongly freezing deps is not recommended since it can easily create a conflict with other packages in a user environment. 

Also, If you know that gremlinpython works well with a lower package version (e.g six 1.6), I recommend to just change it to the minimum required to be more flexible.


spmallette:

VOTE +1

davebshow:
VOTE +1

jorgebay:
VOTE +1

PR #729, opened 2017-10-04
TINKERPOP-1632: Create a set of default functions

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1632

This is an implementation of lambda-less math capabilities in Gremlin leveraging the String-based calculator model proposed by @twilmes. This model deviates from Gremlin's standard function composition and nesting model to provide a easy to read "math processor" that leverages Gremlin scopes for data access (e.g. path data, map keys, and side-effects). The feature is encapsulated in `MathStep` which implements `PathProcessor` and `Scoping`. Furthermore, `by()`-modulation is supported. `by()`-modulators are applied in the order in which the variable is first used in the equation. Thus:

```
g.V().as("a").out("knows").by("b").
  math("b + a").
    by(out().count()).
    by("age")
```

In the above: `a` -> `age` and `b` -> `out().count()`. More complex examples are provided below:

```
gremlin> g.V().as('a').out('knows').as('b').math('a + b').by('age')
==>56.0
==>61.0
gremlin> g.V().as('a').out('created').as('b').
......1>   math('b + a').
......2>     by(both().count().math('_ + 100')).
......3>     by('age')
==>132.0
==>133.0
==>135.0
==>138.0
gremlin> g.withSideEffect('x',10).V().values('age').math('_ / x')
==>2.9
==>2.7
==>3.2
==>3.5
gremlin> g.withSack(1).V(1).repeat(sack(sum).by(constant(1))).times(10).emit().sack().math('sin _')
==>0.9092974268256817
==>0.1411200080598672
==>-0.7568024953079282
==>-0.9589242746631385
==>-0.27941549819892586
==>0.6569865987187891
==>0.9893582466233818
==>0.4121184852417566
==>-0.5440211108893698
==>-0.9999902065507035
```

### DIFF
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -1207,6 +1207,50 @@ system to leverage the filter for index lookups.
 IMPORTANT: A `where()`-step is a filter and thus, variables within a `where()` clause are not globally bound to the
 path of the traverser in `match()`. As such, `where()`-steps in `match()` are used for filtering, not binding.
 
+[[math-step]]
+=== Math Step
+
+The `math()`-step (*math*) enables scientific calculator functionality within Gremlin. This step deviates from the common
+function composition and nesting formalisms to provide an easy to read string-based math processor. Variables within the
+equation map to scopes in Gremlin -- e.g. path labels, side-effects, or incoming map keys. This step supports
+`by()`-modulation where the `by()`-modulators are applied in the order in which the variables are first referenced
+within the equation. Note that the reserved variable `_` refers to the current numeric traverser object incoming to the
+`math()`-step.
+
+[gremlin-groovy,modern]
+----
+g.V().as('a').out('knows').as('b').math('a + b').by('age')
+g.V().as('a').out('created').as('b').
+  math('b + a').
+    by(both().count().math('_ + 100')).
+    by('age')
+g.withSideEffect('x',10).V().values('age').math('_ / x')
+g.withSack(1).V(1).repeat(sack(sum).by(constant(1))).times(10).emit().sack().math('sin _')
+----
+
+The operators supported by the calculator include: `*`, `+`, `\`, `^`, and `%`.
+Furthermore, the following built in functions are provided:
+
+* `abs`: absolute value
+* `acos`: arc cosine
+* `asin`: arc sine
+* `atan`: arc tangent
+* `cbrt`: cubic root
+* `ceil`: nearest upper integer
+* `cos`: cosine
+* `cosh`: hyperbolic cosine
+* `exp`: euler's number raised to the power (`e^x`)
+* `floor`: nearest lower integer
+* `log`: logarithmus naturalis (base e)
+* `log10`: logarithm (base 10)
+* `log2`: logarithm (base 2)
+* `sin`: sine
+* `sinh`: hyperbolic sine
+* `sqrt`: square root
+* `tan`: tangent
+* `tanh`: hyperbolic tangent
+* `signum`: signum function
+
 [[max-step]]
 === Max Step
 
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MathStep.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MathStep.java
@@ -21,7 +21,6 @@ package org.apache.tinkerpop.gremlin.process.traversal.step.map;
 
 import net.objecthunter.exp4j.Expression;
 import net.objecthunter.exp4j.ExpressionBuilder;
-import org.apache.commons.lang.StringUtils;
 import org.apache.tinkerpop.gremlin.process.traversal.Pop;
 import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
 import org.apache.tinkerpop.gremlin.process.traversal.Traverser;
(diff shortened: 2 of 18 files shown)


dkuppitz:
You can add a static compiled regex pattern:

```
private static final Pattern EQUATION_PATTERN = Pattern.compile("\\b(?!abs|acos|asin|atan|cbrt|ceil|cos|cosh|exp|floor|log|log10|log2|sin|sinh|sqrt|tan|tanh|signum)(_|([A-Za-z][A-Za-z0-9]*))\\b");
```

...then `getVariables()` can be as simple as:

```
protected static final Set<String> getVariables(final String equation) {
    final Matcher matcher = EQUATION_PATTERN.matcher(equation);
    final Set<String> variables = new LinkedHashSet<>();
    while (matcher.find()) {
        variables.add(matcher.group());
    }
    return variables;
}
```

And perhaps to increase the readability and maintainability, we should do something like this:

```
private static final String[] FUNCTIONS = new String[] {
        "abs", "acos", "asin", "atan",
        "cbrt", "ceil", "cos", "cosh",
        "exp",
        "floor",
        "log", "log10", "log2",
        "signum", "sin", "sinh", "sqrt",
        "tan", "tanh"
    };

private static final Pattern EQUATION_PATTERN = Pattern.compile("\\b(?!" +
        String.join("|", FUNCTIONS) + ")(_|([A-Za-z][A-Za-z0-9]*))\\b");
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MathStep.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    @Override
+    protected Traverser.Admin<Double> processNextStart() {
+        return PathProcessor.processTraverserPathLabels(super.processNextStart(), this.keepLabels);
+    }
+
+    @Override
+    protected Double map(final Traverser.Admin<S> traverser) {
+        final Expression expression = new ExpressionBuilder(this.equation)
```
twilmes: Looking at the exp4j source, I think an `Expression`  can be reused. If so, this could be moved up into the constructor to save the repeated initialization cost.

okram: Not with threading. Thus, OLAP and any other execution engine that uses threads will get inconsistent results.

okram:
```
[INFO] BUILD SUCCESS
[INFO] Total time: 02:34 h
[INFO] Finished at: 2017-10-04T16:21:53-06:00
[INFO] Final Memory: 165M/1543M
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MathStep.java
twilmes: Yes, good point. I was trying to think of a way to do it with a `ThreadLocal` but can't think of one. I think I found a way around the threading issue that at least circumvents the repeated expression parsing. We can't get away from the extra object creation but the `Expression` constructor can take an `Expression` as input. That creates a copy with the tokens already pulled out so no parsing is needed. A golden copy of the expression could be created in the constructor and then used to seed the method-level expressions.

I ran this quick little test against grateful-dead.kryo. It's a nonsensical query, but I wanted to see how much of a difference parsing made.
`clockWithResult(20000){g.V().hasLabel("song").math("_ * 100 + 200 / 300").by('performances').limit(1000).toList()}`

original: 0.87 ms
`new Expression(expression)`: 0.38 ms

twilmes:
This is going to come in super handy.

A hearty VOTE: +1

docs/src/reference/the-traversal.asciidoc
```diff
@@ -1207,6 +1207,50 @@ system to leverage the filter for index lookups.
...
+* `log`: logarithmus naturalis (base e)
+* `log10`: logarithm (base 10)
+* `log2`: logarithm (base 2)
+* `sin`: sine
+* `sinh`: hyperbolic sine
+* `sqrt`: square root
+* `tan`: tangent
+* `tanh`: hyperbolic tangent
+* `signum`: signum function
+
```
spmallette: please reference the ticket number here.

spmallette:
Does exp4j contain a NOTICE file in its jar (usually in  `/META-INF`)?

okram:
I just unzipped the jar and the META-INF/ does not contain it. Just a pom.xml in there.

spmallette:
The .net build is failing:

```text
Process/Traversal/GraphTraversal.cs(500,36): error CS0246: The type or namespace name 'Double' could not be found (are you missing a using directive or an assembly reference?) [/usr/src/tinkerpop/gremlin-dotnet/src/Gremlin.Net/Gremlin.Net.csproj]
Process/Traversal/__.cs(414,46): error CS0246: The type or namespace name 'Double' could not be found (are you missing a using directive or an assembly reference?) [/usr/src/tinkerpop/gremlin-dotnet/src/Gremlin.Net/Gremlin.Net.csproj]
```

FlorianHockmann:
> The .net build is failing:

`Double` seems to be a type that we didn't have in the traversal API until now. It should be simply mapped to `double` in C#. This should only require an addition to the `toCSharpTypeMap` in Gremlin.Net's [`generate.groovy`](https://github.com/apache/tinkerpop/blob/master/gremlin-dotnet/glv/generate.groovy#L29).

spmallette:
just pushed a fix for the double issue....it was easy - pretty much as @FlorianHockmann deduced. re-running docker build now.

spmallette:

VOTE +1

PR #730, opened 2017-10-09
TINKERPOP-1797: LambdaRestrictionStrategy and LambdaMapStep in by()-modulation.

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1797

We have had too many problems with `LambdaRestrictionStrategy` because it is difficult to know what is a true lambda. I have now simply hardcoded the lambda determination as a `String` analysis of the lambda object for Java, Groovy, and Python. As we add more GLVs, we can add more string mappers. Every other solution thus far has either been too lenient or too restrictive.

VOTE +1


dkuppitz:
VOTE: +1

spmallette:

VOTE +1

PR #732, opened 2017-10-13
TINKERPOP-1799 Fixed serialization of Path objects for python.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1799

This issue only related to GraphSON 3.0 and was fairly straightforward to fix. Tests did not pick up the failure because they continued to use GraphSON 2.0 format in the test itself.

VOTE +1


robertdale:
VOTE +1

jorgebay:
VOTE +1

PR #733, opened 2017-10-17
TINKERPOP-1801: fix profile() timing in OLAP by adding worker iterati…

artem-aliev:
…on timings to step metrics

this is a simple fix that do not change any API


artem-aliev:
that should be against tp32 branch

artem-aliev:
See #734

PR #734, opened 2017-10-17
TINKERPOP-1801: fix profile() timing in OLAP by adding worker iterati…

artem-aliev:
…on timings to step metrics

this is a simple fix that do not change any API

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -23,6 +23,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 [[release-3-2-7]]
 === TinkerPop 3.2.7 (Release Date: NOT OFFICIALLY RELEASED YET)
 
+* profile() timing is fixed for OLAP by adding worker iteration timings to step metrics
 * Better respected permissions on the `plugins.txt` file and prevented writing if marked as read-only.
 * Added getters for the lambdas held by `LambdaCollectingBarrierStep`, `LambdaFlatMapStep` and `LambdaSideEffectStep`.
 * Fixed an old hack in `GroovyTranslator` and `PythonTranslator` where `Elements` were being mapped to their id only.
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/traversal/MemoryTraversalSideEffects.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/traversal/MemoryTraversalSideEffects.java
@@ -169,6 +169,11 @@ public final class MemoryTraversalSideEffects implements TraversalSideEffects {
         memoryTraversalSideEffects.phase = phase;
     }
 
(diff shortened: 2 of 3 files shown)


okram:
This is a nice update @artem-aliev because it doesn't change API and it is general for all `GraphComputer` implementations. Great! A couple things please for a solid VOTE.

1. Please update the `CHANGELOG.asciidoc` with the change you made.
2. In this PR discussion, please provide a `CUT/PASTE` of what the new metrics `toString()` looks like so people can judge its merits.

Thank you.

artem-aliev:
The fix add iteration time to the appropriate step. Iteration time is a time between workerIterationStart() and workerIterationEnd() callbacks.
So before fix the timing looks like 
```
gremlin> g.V().out().out().count().profile()
==>Traversal Metrics
Step                                                               Count  Traversers       Time (ms)    % Dur
GraphStep(vertex,[])                                                9962        9962          70.873    48.95
VertexStep(OUT,vertex)                                           1012657        3745          37.132    25.65
VertexStep(OUT,edge)                                             2101815        6192          36.751    25.39
CountGlobalStep                                                        1           1           0.018     0.01
                                            >TOTAL                     -           -         144.775        -
```
While query runs 10s seconds.
After the fix:
```
gremlin> g.V().out().out().count().profile()
==>Traversal Metrics
Step                                                               Count  Traversers       Time (ms)    % Dur
GraphStep(vertex,[])                                                9962        9962       14186.809    97.43
VertexStep(OUT,vertex)                                           1012657        3745         340.051     2.34
VertexStep(OUT,edge)                                             2101815        6192          33.684     0.23
CountGlobalStep                                                        1           1           0.004     0.00
                                            >TOTAL                     -           -       14560.549        -
```
That shows that most of the time for this OLAP query was spend in the initial iteration (actually star graph creation).
There still could be some inconsistencies because
1. No computer specific setup time is measured
2. Spark has a lot of lazy staff, so most of Spark RDD setup are counted as a first step.
3. The algorithm could fail to assign timing to the right step in bas of sophisticated queries.

By the way new timing is pretty close to the wall clock time.

okram:
VOTE +1

dkuppitz:
VOTE: +1

spmallette:
This doesn't build for me with just regular `mvn clean install`

artem-aliev:
Something strange happened, during rebases. I'll fix.

artem-aliev:
I have fixed test failures.
    TinkerPopComputer does not call ComputerPorgram.execute methods if spit has no vertices.
    For example: modern graph has 6 vertices but computer has 8 cores, there will be two empty splits.
    TraversalVertexProgram use execute step to setup next profiling step, so it is not setup side effects properly for empty splits.
    So tests did not filed in docker but failed on computer with more then 6 cores.
    The fix add check that profile side effects were regester properly before using

spmallette:
@okram if all looks good to you after artem's changes you should be free to merge this now:

VOTE +1

PR #737, opened 2017-10-21
TINKERPOP-1798 Fix signature of MutationListener.vertexPropertyChanged()

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1798

This method was referencing a Property when it should have been referencing a VertexProperty. Implemented by way of deprecation of the old method so that this ends up being a non-breaking change.

VOTE +1


dkuppitz:
VOTE: +1

robertdale:
VOTE +1

PR #738, opened 2017-10-30
TINKERPOP-1820 Include Python and .NET GLVs on TravisCI

jorgebay:
https://issues.apache.org/jira/browse/TINKERPOP-1820

Run Python and .NET GLV tests on travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,4 +22,4 @@ before_install:
   - sudo apt-get update
   - sudo apt-get install dotnet-dev-1.0.4
 script: 
-  - "mvn clean install -Dorg.slf4j.simpleLogger.defaultLogLevel=WARN -Dci"
+  - "mvn clean install -q -Dci"
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,6 @@ jdk:
 sudo: required
 dist: trusty
 
-env:
-  - MAVEN_OPTS=-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
-
 install:
   # include glvs
   - touch gremlin-python/.glv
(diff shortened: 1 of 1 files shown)


spmallette:
I see that you used the `-q` option. What happens if there is a failure in native python tests? currently i don't see any output for those, i'm wondering how that will fail when it does. 

I think it's good to include the GLVs on Travis, but it won't eliminate local builds completely because travis doesn't run integration tests - nor do we want it to probably. This build already takes 20 minutes on Travis which is a reasonably fast feedback from the build server.  I don't think we want to lose that.

jorgebay:
> I see that you used the -q option. What happens if there is a failure in native python tests? currently i don't see any output for those, i'm wondering how that will fail when it does.

I was looking for ways to limit the output, TravisCI only accepts up to 4Mb which is a lot... On previous commits I've tried to set either: 

`-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn`
or 
`-Dorg.slf4j.simpleLogger.defaultLogLevel=WARN`

But it was not being considered by `mvn`. Maybe including any of [those options on the root pom](https://github.com/apache/tinkerpop/blob/2e4f8bf7c3a1c9e39729ab5aab4886f1f469d641/pom.xml#L1335-L1336) would work, that way I could remove the quiet flag, wdyt?

> I think it's good to include the GLVs on Travis, but it won't eliminate local builds completely because travis doesn't run integration tests - nor do we want it to probably.

hmm... Looking at the output, I see that both unit and integration tests are run for Gremlin.Net which is unexpected (would be the only module doing both).

Now that I think of it, maybe it would be best to run different jobs on the same build, using [a build matrix](https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix) or even [build steps](https://docs.travis-ci.com/user/build-stages/define-steps/). That way we could run individual profiles (glvs, integration tests for different modules, ...) in different jobs for the same build.

We should find a way to continuously run most of the test suite, to ease the burden of having to run it locally.

jorgebay:
I'm closing this one. I'll open a new one to only add Gremlin.Net tests support.

PR #740, opened 2017-11-01
TINKERPOP-1814: Some process tests require the graph API

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1814

Fixed assertions in the test cases outlined by @BrynCooke that use the structure API when they should stick with the process API. I found more areas than specified in the ticket. I also suspect that other process test cases will have similar problems. As these are exposed, please just make a new ticket accordingly - besides being monotonously boring, these are easy to fix.

VOTE +1.


spmallette:

VOTE +1

dkuppitz:
VOTE: +1

PR #741, opened 2017-11-01
TINKERPOP-1807 Gremlin-Python doesn't support GraphSON types g:Date, g:Timestamp and g:UUID

davebshow:
https://issues.apache.org/jira/browse/TINKERPOP-1807

This PR add support for GraphSON types `UUID`, `Date`, and `Timestamp` as well as associated tests. `UUID` and `Date` are straightforward, as the Python standard library includes `Date` (`datetime.datetime`) and `UUID` (`uuid.UUID`) objects. However, in Python, a timestamp is just a float, so in order to support serialization to a GraphSON type Timestamp, I added a dummy class `timestamp` to the `statics` module, similar to the `long` class for Python 3.

Note, another PR will be required for master with implementations for both GraphSON2 and 3. I will open upon approval of this one.

### DIFF
--- a/gremlin-python/src/main/jython/tests/structure/io/test_graphson.py
+++ b/gremlin-python/src/main/jython/tests/structure/io/test_graphson.py
@@ -18,16 +18,20 @@ under the License.
 '''
 
 __author__ = 'Marko A. Rodriguez (http://markorodriguez.com)'
-
+import datetime
 import json
+import uuid
+
 from mock import Mock
 
 import six
 
 from gremlin_python.statics import *
-from gremlin_python.structure.graph import Vertex, Edge, VertexProperty, Property
+from gremlin_python.structure.graph import (
+    Vertex, Edge, VertexProperty, Property, Graph)
 from gremlin_python.structure.graph import Path
-from gremlin_python.structure.io.graphson import GraphSONWriter, GraphSONReader, GraphSONUtil
+from gremlin_python.structure.io.graphson import (
+    GraphSONWriter, GraphSONReader, GraphSONUtil)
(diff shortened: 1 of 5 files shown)


spmallette:
Please include a CHANGELOG entry as well as user upgrade documentation.

spmallette:

VOTE +1

gremlin-python/src/main/jython/tests/structure/io/test_graphson.py
```diff
@@ -121,6 +125,19 @@ class X(object):
         serdes.objectify.assert_called_once_with(value, reader)
         assert o is serdes.objectify()
 
+    def test_datetime(self):
+        dt = self.graphson_reader.readObject(json.dumps({"@type": "g:Date", "@value": 1481750076295}))
+        assert isinstance(dt, datetime.datetime)
```
FlorianHockmann: Wouldn't it make sense to also assert that the value is correct here? Then the test would validate the GraphSON reader completely and it stays consistent with the other tests in this class.

davebshow: Yeah I guess. TBH, I don't find any of these tests too meaningful, which is why I added the functional IO tests at the end that actually send and receive messages from the server and verify that they are equal. I can add the asserts though if you want.

davebshow: Done

FlorianHockmann:
LGTM, VOTE: +1

robertdale:
VOTE +1

PR #743, opened 2017-11-02
TINKERPOP-1813: Subgraph step requires the graph API

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1813

This ticket wanted `SubgraphStep` to not use the structure API. This is not something we will support in TinkerPop3. The structure API is a fundamental layer to our step API. It was advised to the provider to simply write their own custom `ProviderSubgraphStep` which avoids the structure API as they see fit. However, what was accomplished in this ticket was to change `SubgraphTest` to only use traversals (process API) in its assertions. This theme is good -- our process test suite should never rely on the structure API (beyond `ReferenceXXX` data).

VOTE +1


spmallette:

VOTE +1

dkuppitz:
VOTE: +1

PR #745, opened 2017-11-10
TINKERPOP-1830: fix race condition in TinkerIndex

mpollmeier:
My colleage @fabsx00 discovered a race condition in tinkergraph's index creation. He fixed it by simply replacing `parallelStream` with `stream`. Quoting his analysis:

> So, reading the code, you see that this.put is called in parallel, but that method seems to contain a race as get is called on the index, checked for null, and a subsequent write is performed. It still seems like using stream here fixes the problem we've been seeing, and the performance hit is not significant.

Ticket: https://issues.apache.org/jira/browse/TINKERPOP-1830

After initial feedback I can backport this onto tp32.

### DIFF
--- a/tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerIndex.java
+++ b/tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerIndex.java
@@ -48,17 +48,16 @@ final class TinkerIndex<T extends Element> {
 
     protected void put(final String key, final Object value, final T element) {
         Map<Object, Set<T>> keyMap = this.index.get(key);
-        if (keyMap == null) {
-            keyMap = new ConcurrentHashMap<>();
-            this.index.put(key, keyMap);
+        if (null == keyMap) {
+            this.index.putIfAbsent(key, new ConcurrentHashMap<Object, Set<T>>());
+            keyMap = this.index.get(key);
         }
         Set<T> objects = keyMap.get(value);
         if (null == objects) {
-            objects = new HashSet<>();
-            keyMap.put(value, objects);
+            keyMap.putIfAbsent(value, ConcurrentHashMap.newKeySet());
+            objects = keyMap.get(value);
         }
         objects.add(element);
-
     }
 
     public List<T> get(final String key, final Object value) {
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -23,6 +23,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 [[release-3-2-7]]
 === TinkerPop 3.2.7 (Release Date: NOT OFFICIALLY RELEASED YET)
 
+* Fixed a race condition in `TinkerIndex`.
 * Fixed an `ArrayOutOfBoundsException` in `hasId()` for the rare situation when the provided collection is empty.
 * Bump to Netty 4.0.52
 * `TraversalVertexProgram` `profile()` now accounts for worker iteration in `GraphComputer` OLAP.


robertdale:
The race condition is in `put()` and it's used elsewhere. It would make more sense to fix `put()` to be thread-safe.  Then we can keep `parallelStream()`

```
        Map<Object, Set<T>> keyMap = this.index.get(key);
        if (keyMap == null) {
            keyMap = this.index.putIfAbsent(key, new ConcurrentHashMap<>());
        }
        Set<T> objects = keyMap.get(value);
        if (null == objects) {
            objects = keyMap.putIfAbsent(value, ConcurrentHashMap.newKeySet());
        }
        objects.add(element);
```

mpollmeier:
Agreed, that's a better solution @robertdale. Note that `putIfAbsent` is not generally thread safe, but since we use `ConcurrentHashMap` it should be. I'll amend the commit.

robertdale:
@mpollmeier Sorry, I thought that `putIfAbsent()` returned the current map but it returns old mapping or `null` hence the `NPEs`.  Should be:
```java
    protected void put(final String key, final Object value, final T element) {
        Map<Object, Set<T>> keyMap = this.index.get(key);
        if (null == keyMap) {
            Map<Object, Set<T>> tmpMap = new ConcurrentHashMap<>();
            this.index.putIfAbsent(key, tmpMap);
            keyMap = this.index.get(key);
        }
        Set<T> objects = keyMap.get(value);
        if (null == objects) {
            Set<T> tmpObjects = ConcurrentHashMap.newKeySet();
            keyMap.putIfAbsent(value, tmpObjects);
            objects = keyMap.get(value);
        }
        objects.add(element);
    }
```

mpollmeier:
I made the same incorrect assumption. Fixed with your latest snippet and tested with our large testbase. No NPEs, no race conditions so far, looks good. 
VOTE +1

robertdale:
`mvn -pl tinkergraph-gremlin -DskipIntegrationTests=false -am clean install` build SUCCESS
VOTE +1

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerIndex.java
```diff
@@ -48,17 +48,18 @@ public TinkerIndex(final TinkerGraph graph, final Class<T> indexClass) {
 
     protected void put(final String key, final Object value, final T element) {
         Map<Object, Set<T>> keyMap = this.index.get(key);
-        if (keyMap == null) {
-            keyMap = new ConcurrentHashMap<>();
-            this.index.put(key, keyMap);
+        if (null == keyMap) {
+            Map<Object, Set<T>> tmpMap = new ConcurrentHashMap<>();
```
spmallette: can you add a `final` here please?

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerIndex.java
```diff
@@ -48,17 +48,18 @@ public TinkerIndex(final TinkerGraph graph, final Class<T> indexClass) {
...
+            this.index.putIfAbsent(key, tmpMap);
+            keyMap = this.index.get(key);
         }
         Set<T> objects = keyMap.get(value);
         if (null == objects) {
-            objects = new HashSet<>();
-            keyMap.put(value, objects);
+            Set<T> tmpObjects = ConcurrentHashMap.newKeySet();
```
spmallette: can you add a `final` here please?

spmallette:
Please also include an update to CHANGELOG for this fix.

mpollmeier:
Instead of marking them as final I simply dropped the assignment. 
Re tp32: yes, I also mentioned this in the description of the ticket. I've just squashed all commits and rebased onto tp32

spmallette:
sorry - i missed the `tp32` comment in your description. your changes look fine now. 

VOTE +1

mpollmeier:
ok that's done. added a changelog entry, merged this branch into tp32 and merged tp32 into master.

PR #746, opened 2017-11-10
TINKERPOP-1829 Made detachment configurable in EventStrategy

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1831

The default behavior is representative of how EventStrategy currently works so there is no breaking change here. In the interest of avoiding breaking changes, I didn't want to shift the API to make the change cleaner and thus created TINKERPOP-1831 for future consideration.

VOTE +1


okram:
VOTE +1

dkuppitz:
VOTE: +1

PR #748, opened 2017-11-15
TINKERPOP-1834: Consider iterate() as a first class step

okram:
https://issues.apache.org/jira/browse/TINKERPOP-1834

When a user triggers a "terminal method" (e.g. `toList()`, `toSet()`, `iterate()`), that method is now appended to the traversal's `Bytecode`. This allows providers to know what the user's intention is for the results of the traversal. This can allow them to make internal optimizations about how to execute the traversal. The drawback of this is that `Translators` need to know about which steps are "terminal" and to avoid applying them during translation. To make this easy, `Translator.TERMINAL_STEPS` is a `public static final` field to allow easy introspection.

VOTE +1.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,6 +25,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 This release also includes changes from <<release-3-2-7, 3.2.7>>.
 
+* Added terminal steps to `Bytecode` so that bytecode consumers know the terminal method called on the traversal.
 * Fixed bug in serialization of `Path` for GraphSON 3.0 in `gremlin-python`.
 * Added support for GraphSON 3.0 in Gremlin.Net.
 * Added `math()`-step which supports scientific calculator capabilities for numbers within a traversal.
--- a/docs/src/upgrade/release-3.3.x.asciidoc
+++ b/docs/src/upgrade/release-3.3.x.asciidoc
@@ -169,6 +169,19 @@ changes there may prove important for the provider's implementation.
 
 ==== Graph Database Providers
 
+===== Terminal Method Added to Traversal Bytecode
+
+There are a set of "terminal methods" associated with a `Traversal`.
+These include: `toList()`, `toSet()`, `promise()`, `iterate()`, etc.
+For some providers, it is important to know which terminal method the user called when executing a traversal. This can allow
+them to make certain assumptions about what data to return. Every finalizing terminal method is
+appended to the traversal's `Bytecode`.
+
+```
+gremlin> g.V().out().iterate().getBytecode()
+==>[[], [V(), out(), iterate()]]
+```
+
 ===== IO Version Check
 
 In the `Graph.io()` method, providers are to bootstrap the `Io` instance returned with their own custom serializers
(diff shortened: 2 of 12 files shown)


okram:
```
[INFO] BUILD SUCCESS
[INFO] Total time: 02:30 h
[INFO] Finished at: 2017-11-15T12:13:52-07:00
[INFO] Final Memory: 164M/1526M
```

spmallette:
In reviewing this, I think Gremlin Server needs a change. When we submit a remote traversal. Gremlin Server needs to check for `iterate()` in the bytecode and actually call `iterate()` rather than `next()` that way the graph instance gets the right method called on it for any additional optimization that could occur there. Plus, if the client did call `iterate()` it's better that Gremlin Server simply return no results - that's a big optimization. I can work on that change and just commit it to this branch as part of this PR.

okram:
Makes sense.

okram:
I think we over engineered this ticket. I believe @dkuppitz has the best idea.

```
public Traversal<S,E> iterate() {
  this.filter(false);
  while(hasNext()) {
    next();
  } 
  return this;
}
```

spmallette:
Unless I'm not thinking of something, this approach won't work for remote traversals. When you call iterate on the client, it will serialize to bytecode. The bytecode ends up on the server and then becomes deserialized back into a traversal. This server traversal will not know that the client called `iterate()` and then won't be able to also call `iterate()` on the server side - which goes back to the optimization I mentioned in my previous comment.

okram:
There is nothing for "the server" to know. Thus, gutting the introspection. By appending `filter(false)` to the bytecode, you have a traversal that returns nothing. Which is exactly what you want. No need for the receiving/executing engine to reason about anything. A 3-line change to `iterate()` would do the trick.

spmallette:
oh....`filter(false)` referred to adding a `FilterStep` to the end of the traversal basically that didn't allow any traversers through. Of course, a provider writing a `TraversalStrategy` wouldn't be able to detect "false" if you use the a lambda in `filter()`, so the provider couldn't optimize properly. Are you thinking something else here?

dkuppitz:
`filter(false)` == `not(identity())`

okram:
Yes, `filter(false)` is just "the concept." The step would be:

```
filter(FalseTraversal.instance())
```

Or as @dkuppitz says -- `not(identity())`.

BrynCooke:
Would a cap step with no side effect keys be a better fit here rather than filter?
It would remove the need for a 'false' traversal.

spmallette:
I think I like @BrynCooke  suggestion. an empty `cap()` currently generates an error so this might be a good use for it. being equivalent to `iterate()` seems logical to me.

okram:
A `cap()` step is a `SupplyBarrierStep` and thus, requires an emission -- a single emission, but an emission nonetheless. The ultimate solution step has to a `FilterStep` by nature so it can truly return "nothing." However, we are now getting into particulars that don't need arguing. The concept is what matters. Now that I get the problem, you simply don't want to have to "reattach" and send data back if you don't have to, then the solution is, filter everything as the last step. That is, have `iterate()` append to the bytecode a "filter all." Easy. How that is done, just needs some fiddlin' around.

okram:
Closing. Going to provide a simpler solution.

PR #751, opened 2017-11-19
TINKERPOP-1832: TraversalHelper.replaceStep sets previousStep to the wrong step

pietermartin:
https://issues.apache.org/jira/browse/TINKERPOP-1832

`TraversalHelper.replaceStep` needs to first remove the step before adding in the new step. Else it gets its `previousStep` pointer wrong.

Added `TraversalHelperTest.shouldSetPreviousStepToEmptyStep` which was the scenario that originally made me aware of the issue.

Ran `mvn clean install` and tested Sqlg's tests with the fix.


pietermartin:
VOTE +1

spmallette:
This PR is targetted at master (3.3.x) - any reason to not also target tp32 (3.2.x) or is this just a problem on the 3.3.x line?

pietermartin:
I'd have to check but pretty sure its the same on 3.2.x

dkuppitz:
It is the same in 3.2.x and was actually testing the fix in `tp32` when the discussion first came up.

VOTE: +1 (after retargeting)

pietermartin:
Closing to retarget tp32

PR #747, opened 2017-11-12
TINKERPOP-1784 GLV Test Framework

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1784

This PR is for the GLV Test Framework. It contains an implementation for gremlin-python. It is not a complete porting of all the process test suite, but does provide coverage for almost all steps. I think I've built enough here to provide enough for evaluation of the framework itself - my intention is to backfill tests after this merges. I do think I have enough tests here to yield confidence in GLVs that implement it, thus allowing baseline for official release.

All tests pass with `docker/build.sh -t -n -i`

VOTE +1

PR #754, opened 2017-11-21
TINKERPOP-1827 Gremlin .NET: Test Suite Runner

jorgebay:
https://issues.apache.org/jira/browse/TINKERPOP-1827

Adds support for Gherkin features for Gremlin.NET.

Changes are made on top of #747, I'll rebase it once the GLV test suite is merged, in order to speed up reviewing (selecting the range of commits).

Of a total of 302 scenarios, 28 are ignored mostly due to lambdas not being supported on C# (274 succeeds).

I've included a mechanism to ignore scenarios manually, but a reason must be provided. Currently, there are 2 valid reasons for a scenario to be ignored:
- `Scenario ignored because lambdas are not supported in Gremlin.NET`
- `Scenario ignored because there is conversion issue for IDictionary<K, V> instaces in Gremlin.NET (TINKERPOP-1837)`

As there isn't a usable cucumber/gherkin runner for .NET Core, gherkin features evaluation are launched from a single xunit test (`GherkinTestRunner`), to avoid having another entry point to maintain.

Currently, the output shows all scenarios executed and a number to identify failure, which is shown at the bottom.


jorgebay:
I rebased it against `tp32`, all tests pass.

VOTE +1

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/TraversalEvaluation/TraversalParser.cs
```diff
@@ -0,0 +1,472 @@
...
+            }
+            var numericText = match.Value.ToUpper();
+            i += match.Value.Length - 1;
+            if (numericText.EndsWith("L"))
+            {
+                return LiteralParameter.Create(Convert.ToInt64(match.Value.Substring(0, match.Value.Length - 1)));
+            }
+            if (numericText.EndsWith("F"))
+            {
+                return LiteralParameter.Create(Convert.ToSingle(match.Value.Substring(0, match.Value.Length-1)));
```
FlorianHockmann: Add `CultureInfo.InvariantCulture` for the `Convert.ToSingle` call to avoid problems when the local culture uses a different decimal mark.

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/TraversalEvaluation/TraversalParser.cs
```diff
@@ -0,0 +1,472 @@
...
+            }
+            if (match.Groups[1].Value != "")
+            {
+                // Captured text with the decimal separator
+                return LiteralParameter.Create(Convert.ToDecimal(match.Value));
```
FlorianHockmann: Same as above.

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/TraversalEvaluation/TraversalParser.cs
jorgebay: Nice catch! I'll push a fix for it.

FlorianHockmann:
With your latest commit all scenarios work for me.

I try to review the changes by the end of the week.

jorgebay:
I've rebased after #757 was merged. Included all previously skipped scenarios related to `IDictionary<K, V>` coercion.

I included two type-related fixes to the Match and Group features.

```
Total scenarios: 328. Passed: 306. Failed: 0. Skipped: 22.
```

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/GherkinTestRunner.cs
```diff
@@ -0,0 +1,381 @@
...
+        private string GetRootPath()
+        {
+            var codeBaseUrl = new Uri(GetType().GetTypeInfo().Assembly.CodeBase);
+            var codeBasePath = Uri.UnescapeDataString(codeBaseUrl.AbsolutePath);
+            DirectoryInfo rootDir = null;
+            for (var dir = Directory.GetParent(Path.GetDirectoryName(codeBasePath));
+                dir.Parent != null;
+                dir = dir.Parent)
+            {
+                if (dir.Name == "gremlin-dotnet" && dir.Parent?.Name == "tinkerpop")
```
FlorianHockmann: I would prefer if we could make this more flexible so that other names than _tinkerpop_ for the directory are allowed. Since the directory isn't part of the repo it can have an arbitrary name.
(The test runner failed for me because the directory name _TinkerPop_.)

gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/DecimalConverter.cs
```diff
@@ -0,0 +1,34 @@
...
+ * under the License.
+ */
+
+#endregion
+
+using System;
+
+namespace Gremlin.Net.Structure.IO.GraphSON
+{
+    internal class DecimalConverter : NumberConverter
```
FlorianHockmann: Could you please add two unit tests that read and write a `decimal`? We have those tests for all other writers and readers, including the numerical ones, so we should stay consistent and also add them here.

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/GherkinTestRunner.cs
jorgebay: Agree, looking for `tinkerpop` directory is incorrect.
I've changed it to look for `gremlin-dotnet` directory which has a `pom.xml` file. There are different approaches we can use (like using an env var) but each has its own downsides.

FlorianHockmann:
VOTE +1

spmallette:
VOTE +1

PR #757, opened 2017-11-22
TINKERPOP-1837 Gremlin .NET: Provide type coercion between IDictionary<K, V> instances

jorgebay:
https://issues.apache.org/jira/browse/TINKERPOP-1837

(The new Travis job comes in handy to validate features like this one :) )

VOTE +1

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Process/Traversal/DefaultTraversal.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Process/Traversal/DefaultTraversal.cs
@@ -24,6 +24,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Reflection;
 using System.Threading.Tasks;
 
 namespace Gremlin.Net.Process.Traversal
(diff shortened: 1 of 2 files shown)


gremlin-dotnet/src/Gremlin.Net/Process/Traversal/DefaultTraversal.cs
```diff
@@ -86,9 +87,61 @@ public void Reset()
...
+                var valueType = type.GenericTypeArguments[1];
+                var mapType = typeof(Dictionary<,>).MakeGenericType(keyType, valueType);
+                var result = (IDictionary) Activator.CreateInstance(mapType);
+                foreach (DictionaryEntry kv in dictValue)
+                {
+                    result.Add(GetValue(keyType, kv.Key), GetValue(valueType, kv.Value));
+                }
+                return result;
+            }
+            if (value is IEnumerable enumerableValue && genericType == typeof(IList<>))
```
FlorianHockmann: Can't we change this to also allow `List<>`  as the `genericType` instead of `IList<>`?

jorgebay: I think continue using the interfaces as the return data type allow us to hide implementation details and its an accurate representation of the Java types (`Map<,>` and `List<>` are interfaces in java).

FlorianHockmann: I don't want to change the return types for the traversal API, but a user can specify the type for some steps directly like you did in the added tests: `ValueMap<IList<int>>`. When `ValueMap<List<int>>` is used instead then it won't work right now.

jorgebay: Its always easier to go from an strict API to a wider range of representations in the future than the other way around.

FlorianHockmann: Ok, then lets leave it as it is in this PR. We can still change it in case it leads to problems for users.

FlorianHockmann:
The changes look good to me.

VOTE +1

spmallette:
VOTE +1

PR #758, opened 2017-11-29
TINKERPOP-1734 DSL for Gremlin .NET

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1734

Provided tests and documentation for implementing DSLs in .Net. No changes were needed to the code base for this - just used C# extension methods as suggested by @jorgebay . Only downside was anonymous traversals are a little weird because C# doesn't let you extend static classes and extensions methods can't be used on them either. The only recourse is to wrap static methods in the extending class which is a lot of extra boilerplate. I opted to just recommend that users create their own class for spawning anonymous traversals that just works independently of `__`. 

VOTE +1

### DIFF
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -3218,3 +3218,93 @@ social.persons("marko").youngestFriendsAge()
 social.persons().filter(__.createdAtLeast(2)).count()
 ----
 
+=== Gremlin.Net
+
+Developing DSLs for .Net is most easily implemented using link:https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods[Extension Methods]
+as they don't require direct extension of classes in the TinkerPop hierarchy. Extension Method classes simply need to
+be constructed for the `GraphTraversal` and the `GraphTraversalSource`. Unfortunately, anonymous traversals (spawned
+from `__`) can't use the Extension Method approach as they do not work for static classes and static classes can't be
+extended. The only option is to re-implement the methods of `__` as a wrapper in the anonymous traversal for the DSL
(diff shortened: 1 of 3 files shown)


FlorianHockmann:
Really nice to see how easy it is to implement a DSL in Gremlin.Net and I like your suggestion to simply create another version of `__` to deal with the fact that we can't extend it in C#.

My only suggestion is to change the name of the classes in the examples that contain extension methods to end on `Extensions` as that is a convention in C#. So `SocialTraversal` would become `SocialTraversalExtensions` and `SocialTraversalSource` becomes `SocialTraversalSourceExtensions`. That also communicates the fact that those classes don't inherit from `GraphTraversal` and `GraphTraversalSource`.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -3218,3 +3218,93 @@ social.persons("marko").youngestFriendsAge()
...
+}
+----
+
+Note the creation of `__Social` as the Social DSL's "extension" to the available ways in which to spawn anonymous
+traversals. The use of the double underscore prefix in the name is just a convention to consider using and is not a
+requirement. To use the DSL, bring it into scope with the `using` directive:
+
+[source,csharp]
+----
+using Dsl;
```
jorgebay: I would include a [`using static` directive](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-static) in the code sample:

```csharp
using static Dsl.__Social;
```

which allows using `__Social` methods directly:

```csharp
social.Persons().Filter(CreatedAtLeast(2)).Count();
```

spmallette:
Addressed all the feedback with the latest changed and rebased to get rid of the changelog conflict.

jorgebay:
VOTE +1

FlorianHockmann:
VOTE +1

PR #759, opened 2017-11-30
TINKERPOP-1846 Fixed bug in LambdaRestrictionStrategy

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1846

Traversals using Lambda scripts were not triggering this strategy. The strategy expected the word "lambda" in the toString() of the step that held it. Lambda objects weren't including that word. This is ready for review/feedback. I'll withhold my vote while tests are running and others have a chance to chime in as there are other ways this could be resolved.


okram:
VOTE +1

spmallette:

VOTE +1

dkuppitz:
VOTE: +1

PR #760, opened 2017-11-30
TINKERPOP-1806 Consistent use of Gremlin.Net instead of Gremlin-DotNet

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1806

This change really just applies to documentation. Directory structures, links and other identifiers have been left as "dotnet". I will do a search for similar changes to master and update accordingly when I merge this.

VOTE +1


jorgebay:
lgtm! VOTE +1

FlorianHockmann:
I think you got all occurrences of Gremlin-DotNet, so: VOTE +1

PR #768, opened 2018-01-03
TINKERPOP-1861 Modify VertexProgram.Builder.create() to take Graph as varargs

PBGraff:
VertexProgram.Builder.create() now takes varargs Graph instead of just Graph. This change has been propagated through all methods affected.

WIP since build fails due to API changes. Javadocs might need additional small modifications.


robertdale:
I don't think we should break the interface.  I think these should be new methods and `@deprecate` the old ones. The old methods can call the new methods.

spmallette:
Please rename the pull request to be prefixed by the JIRA number: TINKERPOP-1861 (that way it will tie into the JIRA integration)

Also, is there any reason to not deprecate this back along the tp32 line?

PBGraff:
@spmallette We can apply this to the TP 3.2.x line as well. Should I modify the PR to apply to the tp32 branch instead?

spmallette:
Yes - @PBGraff please re-target the PR to tp32.

PR #769, opened 2018-01-04
TINKERPOP-1844 Default GraphSON 3.0 for GLV tests in gremlin-python

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1844

This is a breaking change because `g:Set` no longer returns a `Set` in python. It is coerced to `List` to account for numeric equality semantics of Java vs Python.

VOTE +1


dkuppitz:
VOTE: +1

davebshow:
VOTE +1

PR #770, opened 2018-01-04
TINKERPOP-1862 TinkerMessenger proper handling of Direction.BOTH

PBGraff:
This PR presents a fix for `TinkerMessenger` that handled `Direction.BOTH` by identifying the `Vertex` that is not the one receiving the message. In the case of self-loops, the same `Vertex` is still returned so these are handled properly.

Note that self-loops cause messages to be sent/received twice since the `Edge` is both `IN` and `OUT`.


spmallette:
I assume this problem also exists in the tp32 branch (3.2.x) - if so, could you please re-target this PR to that branch? also, can you please include tests to validate that self-loops are being handled properly?

PR #778, opened 2018-01-15
TINKERPOP-1870: Extends TraverserSet to have Vertex index for remote …

artem-aliev:
…traversers

That replaces linear search in reversal traversal interator with hashtable lookup.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoVersion.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoVersion.java
@@ -52,6 +52,7 @@ import org.apache.tinkerpop.gremlin.process.traversal.traverser.LP_O_OB_S_SE_SL_
 import org.apache.tinkerpop.gremlin.process.traversal.traverser.O_OB_S_SE_SL_Traverser;
 import org.apache.tinkerpop.gremlin.process.traversal.traverser.O_Traverser;
 import org.apache.tinkerpop.gremlin.process.traversal.traverser.ProjectedTraverser;
+import org.apache.tinkerpop.gremlin.process.traversal.traverser.util.VertexTraverserSet;
 import org.apache.tinkerpop.gremlin.process.traversal.traverser.util.TraverserSet;
 import org.apache.tinkerpop.gremlin.process.traversal.util.DefaultTraversalMetrics;
 import org.apache.tinkerpop.gremlin.process.traversal.util.ImmutableMetrics;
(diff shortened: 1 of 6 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoVersion.java
```diff
@@ -305,6 +307,7 @@ public String getVersion() {
             add(GryoTypeReg.of(RangeGlobalStep.RangeBiOperator.class, 114));
             add(GryoTypeReg.of(OrderGlobalStep.OrderBiOperator.class, 118, new JavaSerializer()));
             add(GryoTypeReg.of(ProfileStep.ProfileBiOperator.class, 119));
+            add(GryoTypeReg.of(VertexTraverserSet.class, 173));
```
spmallette: please move the "last id" comment down from:

https://github.com/apache/tinkerpop/pull/778/files#diff-f1eb3822f6d5659dd2bdc913ae475241R208

you might need to create a separate PR against master for us to review because of this change because Gryo 3.0 will need a similar entry on master (i suspect the kryo id will be different) and we'd need to be sure this all works on master as well.

spmallette:
I was about to ask for more details on how much of an improvement this was in terms of performance, but it's on the JIRA issue - for others reviewing, please see:

https://issues.apache.org/jira/browse/TINKERPOP-1870

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoVersion.java
artem-aliev: master PR: https://github.com/apache/tinkerpop/pull/781
I have make both tp32 and tp33 ids the same. so I skipped 171,172 in  tp32

spmallette: Ok - we'd not skipped before, but it doesn't hurt anything. thanks for the second PR. building it now with docker

spmallette:
I'm posting this to all open PRs of current relevance - this PR needs to be rebased against the branch it is targeted against given a broken python dependency that was published to pypi a day or so ago. I've pushed a fix on tp32 and master at this point and David Brown is working on getting an issue raised with the project that initiated the problem.

artem-aliev:
rebased

spmallette:

VOTE +1

dkuppitz:
VOTE: +1

robertdale:
VOTE +1

PR #779, opened 2018-01-16

Stargazer4869:
It seem the edgeFunction defined in Local MessageScope will not be applied anywhere when you are using SparkGraphComputer, I wonder why that is. It will cause a difference in result with what you will have when you are using TinkerGraphComputer.
In this pull request I send out a message after it is processed by the edgeFunction, so the SparkGraphComputer will behave the same as the TinkerGraphComputer does.


dkuppitz:
Can you please a) create a Jira ticket for this issue (then rename the PR accordingly) and b) add a test case, that would fail without this change?

spmallette:
I handled (a) for you @Stargazer4869 - https://issues.apache.org/jira/browse/TINKERPOP-1872 could you please just edit the pull request name to include that JIRA issue id of "TINKERPOP-1872" as a prefix so it links up with our JIRA integration. 

btw, does this issue occur on the tp32 branch as well? perhaps this PR should be retargeted to that branch. 

I'd agree with @dkuppitz that a test demonstrating the problem would be very helpful. Thanks for your effort on this.

Stargazer4869:
Still working on the test case. A little question here, although maven can generate a jar file for the module gremlin-shade, my IDE is not ok with it. It can find dependencies which should provided by this module, such as:
import org.apache.tinkerpop.shaded.kryo.Kryo;
Any way I can deal with this? It will be great help. My IDE is IntelliJ, if there is a solutions for Eclipse, it will be fine too.

Stargazer4869:
As for which branch should I apply this code changes to.
This issue does occur on the tp32 branch too. So, should I make this change only on tp32 branch, and leave the master branch to you guys?

spmallette:
> my IDE is not ok with it. 

please see: http://tinkerpop.apache.org/docs/current/dev/developer/#_ide_setup_with_intellij

> So, should I make this change only on tp32 branch, and leave the master branch to you guys?

as this does not look like a complicated change, you can just issue the single PR to tp32 and we can handle the merged to master. thanks

spmallette:
I'm posting this to all open PRs of current relevance - this PR needs to be rebased against the branch it is targeted against given a broken python dependency that was published to pypi a day or so ago. I've pushed a fix on tp32 and master at this point and David Brown is working on getting an issue raised with the project that initiated the problem.

Stargazer4869:
Added a test case for this fix.
Retargeted this PR to tp32 branch.

dkuppitz:
This PR is busted. You're trying to merge your `master/` into TinkerPop's `tp32/` branch.
You should make your changes in your local `tp32/` and then submit a PR.

PR #784, opened 2018-01-19
TINKERPOP-1874 P does not appear to be serialized consistently in GraphSON

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1874

Added  bunch of tests for this and found that .NET was still working, but that was just because of how the gherkin processor was generating the traversals (it was always converting `P` args to lists basically). I modified the `P` template for .NET and made it so that `within` and `without` are handled differently from the other enums - `within` and `without` will always pass a collection rather than an individual object if there is only one item present in the arg array.

I also updated the IO docs to include `within` and `without` so that it was clear to implementers what was expected for them. I may have to come back and do some cleanup here, but I already went deep on my PR to master for this so I'd rather just CTR in the doc fixes later.

VOTE +1

### DIFF
--- a/docs/src/dev/io/graphson.asciidoc
+++ b/docs/src/dev/io/graphson.asciidoc
@@ -27,8 +27,7 @@ g = graph.traversal()
 
 toJson = { o, type, comment = "" ->
   println "Writing ${type}"
-  "${type}\n" +
-  "^".multiply(type.length()) + "\n\n" +
+  "==== ${type}\n\n" +
   (comment.isEmpty() ? "" : comment + "\n\n") +
   "[source,json]\n" +
   "----\n" +
@@ -40,8 +39,7 @@ toJson = { o, type, comment = "" ->
 file = new File("out-graphson-1d0.txt")
 file.withWriter { writer ->
 
-  writer.write("Graph Structure\n")
-  writer.write("~~~~~~~~~~~~~~~\n\n")
+  writer.write("=== Graph Structure\n\n")
   writer.write(toJson(graph.edges().next(), "Edge"))
   writer.write(toJson(g.V().out().out().path().next(), "Path"))
   writer.write(toJson(graph.edges().next().properties().next(), "Property"))
@@ -52,8 +50,7 @@ file.withWriter { writer ->
   writer.write(toJson(graph.vertices().next().properties().next(), "VertexProperty"))
 
   writer.write("\n")
-  writer.write("RequestMessage\n")
-  writer.write("~~~~~~~~~~~~~~\n\n")
+  writer.write("=== RequestMessage\n\n")
   def msg = null
   msg = RequestMessage.build("authentication").
                        add("saslMechanism", "PLAIN", "sasl", "AHN0ZXBocGhlbgBwYXNzd29yZA==").create()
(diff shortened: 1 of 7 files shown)


docs/src/dev/io/graphson.asciidoc
```diff
@@ -3580,6 +3572,41 @@ The following `Bytecode` example represents the traversal of `g.V().hasLabel('pe
 }
 ----
 
+==== P within
```
jorgebay: Good idea to add `within` and `without` examples. Can we also add a little introductory text saying that `P` expects a single value or a list of values, except in the case of `within` and `without` where its always a list of values?

jorgebay:
I've added a small suggestion.
Looks great, VOTE +1.

docs/src/dev/io/graphson.asciidoc
```diff
@@ -1279,7 +1270,7 @@ The following `RequestMessage` is an example of a sessionless request for a scri
 
 ==== Authentication Challenge
 
-When authentication is enabled, an initial request to the server will result in an authentication challenge. The typical response message will appear as follows, but handling it could be different dependending on the SASL implementation (e.g. multiple challenges maybe requested in some cases, but not in the default provided by Gremlin Server).
+When authentication is enabled, an initial request to the server will result in an authentication challenge. The typical response message will appear as follows, but handling it could be different dependending on the SASL implementation (e.g. multiple challenges maybe requested in some cases, but no in the default provided by Gremlin Server).
```
robertdale: Was this meant to be changed?

docs/src/dev/io/graphson.asciidoc
```diff
@@ -4066,11 +4075,12 @@ The following `RequestMessage` is an example of a sessionless request for a scri
 }
 ----
 
+
 === ResponseMessage
 
 ==== Authentication Challenge
 
-When authentication is enabled, an initial request to the server will result in an authentication challenge. The typical response message will appear as follows, but handling it could be different dependending on the SASL implementation (e.g. multiple challenges maybe requested in some cases, but not in the default provided by Gremlin Server).
+When authentication is enabled, an initial request to the server will result in an authentication challenge. The typical response message will appear as follows, but handling it could be different dependending on the SASL implementation (e.g. multiple challenges maybe requested in some cases, but no in the default provided by Gremlin Server).
```
robertdale: And this one?
/no/not/
/maybe/may be/

docs/src/dev/io/graphson.asciidoc
```diff
@@ -75,8 +72,7 @@ file.withWriter { writer ->
   writer.write(toJson(msg, "Sessionless Eval", "The following `RequestMessage` is an example of a sessionless request for a script evaluation with an alias that binds the `TraversalSource` of \"g\" to \"social\"."))
 
   writer.write("\n")
-  writer.write("ResponseMessage\n")
-  writer.write("~~~~~~~~~~~~~~~\n\n")
+  writer.write("=== ResponseMessage\n\n")
   msg = ResponseMessage.build(UUID.fromString("41d2e28a-20a4-4ab0-b379-d810dede3786")).
                         code(org.apache.tinkerpop.gremlin.driver.message.ResponseStatusCode.AUTHENTICATE).create()
   writer.write(toJson(msg, "Authentication Challenge", "When authentication is enabled, an initial request to the server will result in an authentication challenge. The typical response message will appear as follows, but handling it could be different dependending on the SASL implementation (e.g. multiple challenges maybe requested in some cases, but no in the default provided by Gremlin Server)."))
```
robertdale: Maybe it was this one that meant to be /no/not/ ?
And /maybe/may be/ ?

docs/src/dev/io/graphson.asciidoc
```diff
@@ -187,8 +180,7 @@ file.withWriter { writer ->
...
```
robertdale: /no/not/
/maybe/may be/

docs/src/dev/io/graphson.asciidoc
spmallette: that's a good idea. i will make a note to do it after i merge if that's ok. i'd rather not have to re-work the PR on master as a result of that little change. i have a bunch of cleanup to do anyway after these merge.

docs/src/dev/io/graphson.asciidoc
spmallette: that's so weird....i will fix it. i thought i already had.

spmallette:
I will clean up all of @robertdale "no/not" stuff when i merge.

robertdale:
VOTE +1

spmallette:
Here's the follow up fixes given the comments: https://github.com/apache/tinkerpop/commit/d5b6ebb08229be07761d11c3a4065f21a8dcddd3

PR #788, opened 2018-01-22

Stargazer4869:
Re-create the pull request to replace the one on master branch.
The edgeFunction defined in Local MessageScope will not be applied anywhere when you are using SparkGraphComputer. It will cause a difference in result with what you will have when you are using TinkerGraphComputer.


spmallette:
sorry, it too me a while to get to this. It looks like you have a good unit tests for `SparkMessenger` but is there an integration test that would also work so that it could be added to the general test suite. In this way we could see it executed against `TinkerMessenger` and potentially other places where an edgeFunction might be in use?

Stargazer4869:
Hi, @spmallette 
Is there an integration test case which involved TinkerMessenger or TinkerGraphComputer? I could write a similar one for SparkGraphComputer. It will make it easier for me to create an integration test case for SparkMessenger

spmallette:
Please see #793 which contains your changes here plus an integration test i was able to write to produce the problem. I wasn't sure how to advise you to do it, without doing it myself, so I hope you don't mind my finishing this one up for you. Please close this pull request when you get a moment...thanks.

PR #790, opened 2018-01-24
TINKERPOP-1875 Gremlin-Python only aggregates to list when using GraphSON3

davebshow:
https://issues.apache.org/jira/browse/TINKERPOP-1875

This PR fixes server response message deserialization in Gremlin-Python. Improper deserialization was leading to `aggregate_to` always resulting in a list as well as API comparability issues between GraphSON2 and GraphSON3 message serializers.

This PR also improves testing--all Gremlin-Python tests using a remote connection are now run with both GraphSON 2 and 3 serializers.

### DIFF
--- a/gremlin-python/src/main/jython/tests/conftest.py
+++ b/gremlin-python/src/main/jython/tests/conftest.py
@@ -1,4 +1,4 @@
-'''
+"""
 Licensed to the Apache Software Foundation (ASF) under one
 or more contributor license agreements.  See the NOTICE file
 distributed with this work for additional information
@@ -15,7 +15,7 @@ software distributed under the License is distributed on an
 KIND, either express or implied.  See the License for the
 specific language governing permissions and limitations
 under the License.
-'''
+"""
 import concurrent.futures
 import pytest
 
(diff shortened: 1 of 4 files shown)


gremlin-python/src/main/jython/tests/conftest.py
```diff
@@ -65,10 +65,14 @@ def fin():
         request.addfinalizer(fin)
         return client
 
-@pytest.fixture
+@pytest.fixture(params=['v2', 'v3'])
```
spmallette: this is good. doesn't help us with radish test problems though. we still need to do https://issues.apache.org/jira/browse/TINKERPOP-1864 - not sure what's going on there

gremlin-python/src/main/jython/tests/conftest.py
```diff
@@ -1,4 +1,4 @@
-'''
+"""
```
spmallette: any reason you switched to double quotes here? seems like we used singles everywhere else.

davebshow: When possible, I try to follow the style guidelines laid out in [PEP 8](https://www.python.org/dev/peps/pep-0008/#string-quotes). Docstrings specifically relate to  [PEP 257](https://www.python.org/dev/peps/pep-0257/). Gremlin Python wasn't originally written with these guidelines in mind, so sometimes I make little style fixes in code that I am updating. Maybe not the best idea, since it doesn't really relate to the PR.

spmallette: i'm fine with adhering to the style guides. perhaps just do it consistently? perhaps CTR in fixes across the board for all the license headers, so that it's not just in this one file.

spmallette:

VOTE +1

gremlin-python/src/main/jython/tests/conftest.py
davebshow: Ok willdo.

davebshow:
VOTE +1

jorgebay:
lgtm, VOTE +1

PR #792, opened 2018-01-27
TINKERPOP-1854 Add Lambdas for Gremlin.Net

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-1854

Since we don't have a way to parse C# lambdas on server side, this PR simply allows to send lambdas in other languages with Gremlin.Net:

- For a Groovy lambda: `Lambda.Groovy("it.get().value('name').length()")`
- and for a Python lambda: `Lambda.Python("lambda x: len(x.get().value('name'))")`

All features are passing now - no features are ignored any more. Therefore I also removed the functionality to ignore features in Gremlin.Net. (I hope that we won't need to ignore any features in Gremlin.Net in the future, but if we have to then it shouldn't be an issue to get this functionality back.)

Unfortunately, it's not possible to use this new `Lambda` type for arguments in the traversal API due to the fact that in Java the type `Function` would correspond to this type, but Java also allows to pass for example the enum `T` as a `Function` because `T` implements `Function<Element, Object>`. So we cannot simply replace all `Function` types from Gremlin-Java with `Lambda` in Gremlin.Net as that would break for example this step:
```
public default <U> GraphTraversal<S, E> by(final Function<U, Object> function,
          final Comparator comparator)
```
that can be called like this: `by(T.value, Order.incr)` which wouldn't be possible when the first argument would only take a `Lambda` in Gremlin.Net. (Luckily, I found this thanks to a Gherkin scenario that makes exactly this call.)
So, currently all arguments where a lambda is possible simply expect an `object`.

VOTE +1

### DIFF
--- a/docs/src/reference/gremlin-variants.asciidoc
+++ b/docs/src/reference/gremlin-variants.asciidoc
@@ -46,7 +46,7 @@ implementation of Gremlin and serves as the foundation by which all other Gremli
 === The Lambda Solution
 
 Supporting link:https://en.wikipedia.org/wiki/Anonymous_function[anonymous functions] across languages is difficult as
-most language do not support lambda introspection and thus, code analysis. In Gremlin-Java, Java8 lambdas can be leveraged.
+most languages do not support lambda introspection and thus, code analysis. In Gremlin-Java, Java8 lambdas can be leveraged.
 
 [source,java]
 g.V().out("knows").map(t -> t.get().value("name") + " is the friend name") <1>
@@ -281,7 +281,7 @@ re-construction machine-side.
 === The Lambda Solution
 
 Supporting link:https://en.wikipedia.org/wiki/Anonymous_function[anonymous functions] across languages is difficult as
-most language do not support lambda introspection and thus, code analysis. In Gremlin-Python,
+most languages do not support lambda introspection and thus, code analysis. In Gremlin-Python,
 a link:https://docs.python.org/2/reference/expressions.html#lambda[Python lambda] should be represented as a zero-arg callable
 that returns a string representation of a lambda. The default lambda language is `gremlin-python` and can be changed via
 `gremlin_python.statics.default_lambda_language`. When the lambda is represented in `Bytecode` its language is encoded
@@ -343,8 +343,10 @@ var g = graph.Traversal().WithRemote(new DriverRemoteConnection(new GremlinClien
 
 When a traversal from the `GraphTraversalSource` is iterated, the traversal’s `Bytecode` is sent over the wire via the registered
 `IRemoteConnection`. The bytecode is used to construct the equivalent traversal at the remote traversal source.
-Since Gremlin.Net currently doesn't support lambda expressions, all traversals can be translated to Gremlin-Java on the remote
-location (e.g. Gremlin Server).
+Moreover, typically the bytecode is analyzed to determine which language the bytecode should be translated to. If the traversal
+does not contain lambdas, the remote location (e.g. Gremlin Server) will typically
+use Gremlin-Java. If it has lambdas written in Groovy, it will use Gremlin-Groovy (e.g. `GremlinGroovyScriptEngine`).
+Likewise, if it has lambdas represented in Python, it will use Gremlin-Python (e.g. `GremlinJythonScriptEngine`).
 
 IMPORTANT: Gremlin.Net’s `ITraversal` interface supports the standard Gremlin methods such as `Next()`, `NextTraverser()`, `ToSet()`,
 `ToList()`, etc. Such "terminal" methods trigger the evaluation of the traversal.
@@ -432,6 +434,25 @@ edgeValueMaps = g.V().OutE().ValueMap(true).ToList();
 NOTE: Many of the TraversalStrategy classes in Gremlin.Net are proxies to the respective strategy on Apache TinkerPop’s
 JVM-based Gremlin traversal machine. As such, their `Apply(ITraversal)` method does nothing. However, the strategy is
 encoded in the Gremlin.Net bytecode and transmitted to the Gremlin traversal machine for re-construction machine-side.
+
+=== The Lambda Solution
+
+Supporting link:https://en.wikipedia.org/wiki/Anonymous_function[anonymous functions] across languages is difficult as
+most languages do not support lambda introspection and thus, code analysis. While Gremlin.Net doesn't support C# lambdas, it
+is still able to represent lambdas in other languages. When the lambda is represented in `Bytecode` its language is encoded
+such that the remote connection host can infer which translator and ultimate execution engine to use.
+
+[source,csharp]
+----
+g.V().Out().Map<int>(Lambda.Groovy("it.get().value('name').length()")).Sum<int>().ToList();      <1>
+g.V().Out().Map<int>(Lambda.Python("lambda x: len(x.get().value('name'))")).Sum<int>().ToList(); <2>
+----
+
+<1> `Lambda.Groovy()` can be used to create a Groovy lambda. 
+<2> `Lambda.Python()` can be used to create a Python lambda.
+
+The `Lambda` class implements interfaces like `IFunction` and `IPredicate` that mirror their Java counterparts which makes it possible
+to use lambdas with Gremlin.Net for the same steps as in Gremlin-Java.
 
 [[gremlin-javascript]]
 == Gremlin-JavaScript
--- a/docs/src/reference/gremlin-variants.asciidoc
+++ b/docs/src/reference/gremlin-variants.asciidoc
@@ -451,8 +451,8 @@ g.V().Out().Map<int>(Lambda.Python("lambda x: len(x.get().value('name'))")).Sum<
 <1> `Lambda.Groovy()` can be used to create a Groovy lambda. 
 <2> `Lambda.Python()` can be used to create a Python lambda.
 
-The `Lambda` class implements interfaces like `IFunction` and `IPredicate` that mirror their Java counterparts which makes it possible
-to use lambdas with Gremlin.Net for the same steps as in Gremlin-Java.
+The `ILambda` interface returned by these two methods inherits interfaces like `IFunction` and `IPredicate` that mirror
+their Java counterparts which makes it possible to use lambdas with Gremlin.Net for the same steps as in Gremlin-Java.
 
 [[gremlin-javascript]]
 == Gremlin-JavaScript
--- a/gremlin-dotnet/glv/generate.groovy
+++ b/gremlin-dotnet/glv/generate.groovy
@@ -54,10 +54,10 @@ def toCSharpTypeMap = ["Long": "long",
                        "TraversalStrategy[]": "ITraversalStrategy[]",
                        "Function": "IFunction",
                        "BiFunction": "IBiFunction",
-                       "UnaryOperator": "object",
+                       "UnaryOperator": "IUnaryOperator",
                        "BinaryOperator": "IBinaryOperator",
                        "Consumer": "IConsumer",
-                       "Supplier": "object",
+                       "Supplier": "ISupplier",
                        "Comparator": "IComparator",
                        "VertexProgram": "object"]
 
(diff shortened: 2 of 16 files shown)


jorgebay:
I didn't have time to review this one, I hope I will be able to look into it by the end this week.

FlorianHockmann:
The discussion on [TINKERPOP-1857](https://issues.apache.org/jira/browse/TINKERPOP-1857) shows that we actually still need the functionality to ignore tests. So I'll have to update this PR to keep that functionality in.

FlorianHockmann:
> The discussion on TINKERPOP-1857 shows that we actually still need the functionality to ignore tests. So I'll have to update this PR to keep that functionality in.

I just did this update. So the PR is ready for review again.

spmallette:
I'll let @jorgebay dig into the c# stuff here, but from my view this looks pretty good and it's nice to see this feature fully supported now. Great to know that all those feature tests pass.  It's too bad that this doesn't match java too well:

> Unfortunately, it's not possible to use this new Lambda type for arguments in the traversal API

but i guess there's not much we can do about that.  Perhaps, this limitations should be explained in the documentation for "The Lambda Solution"? From a documentation perspective, I think that (1) this is a big enough improvement that you should make note of it in the upgrade documentation and (2) this needs a changelog entry before you merge.

Otherwise, all tests pass with `docker/build.sh -t -n -i`

VOTE +1

gremlin-dotnet/glv/generate.groovy
```diff
@@ -48,7 +48,7 @@ def toCSharpTypeMap = ["Long": "long",
                        "TraversalMetrics": "E2",
                        "Traversal": "ITraversal",
                        "Traversal[]": "ITraversal[]",
-                       "Predicate": "TraversalPredicate",
+                       "Predicate": "object",
```
jorgebay: I think we should use an interface `ILambda` similar to java's counterpart: https://github.com/apache/tinkerpop/blob/3.2.7/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/function/Lambda.java#L33

`Lambda.Groovy()` and `Lambda.Python()` could return `ILambda` instances. That way we can have a compile type check for methods like `Filter()`, `Until()`, etc... to accept only `ILambda` instances.

gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/LambdaSerializer.cs
```diff
@@ -0,0 +1,45 @@
...
+        private const int DefaultArgument = -1;
+
+        public Dictionary<string, dynamic> Dictify(dynamic objectData, GraphSONWriter writer)
+        {
+            Lambda lambda = objectData;
+            var valueDict = new Dictionary<string, dynamic>
+            {
+                {"script", lambda.LambdaExpression},
+                {"language", lambda.Language},
+                {"arguments", DefaultArgument}
```
jorgebay: Use `lambda.Arguments` (which for the existent implementations would return `-1`).

gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/LambdaSerializer.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Gremlin.Net.Structure.IO.GraphSON
+{
+    internal class LambdaSerializer : IGraphSONSerializer
+    {
+        private const int DefaultArgument = -1;
+
+        public Dictionary<string, dynamic> Dictify(dynamic objectData, GraphSONWriter writer)
+        {
+            Lambda lambda = objectData;
```
jorgebay: The same, rely on an interface instead of an implementation, use `ILambda`.

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/IgnoreException.cs
```diff
@@ -40,9 +40,6 @@ private static string GetMessage(IgnoreReason reason)
             string reasonSuffix = null;
             switch (reason)
             {
-                case IgnoreReason.LambdaNotSupported:
```
jorgebay: Can we replace it with a comment, like `// Include detailed reason for each enum here`?

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -436,6 +438,22 @@ NOTE: Many of the TraversalStrategy classes in Gremlin.Net are proxies to the re
...
+=== The Lambda Solution
+
+Supporting link:https://en.wikipedia.org/wiki/Anonymous_function[anonymous functions] across languages is difficult as
+most languages do not support lambda introspection and thus, code analysis. While Gremlin.Net doesn't support C# lambdas, it
+is still able to represent lambdas in other languages. When the lambda is represented in `Bytecode` its language is encoded
+such that the remote connection host can infer which translator and ultimate execution engine to use.
+
+[source,csharp]
+----
+g.V().Out().Map<int>(Lambda.Groovy("it.get().value('name').length()")).Sum<int>().ToList();      <1>
```
jorgebay: Should this be `Lambda.Groovy("{ it.get().value('name').length() }")`, with curly brackets?

gremlin-dotnet/glv/generate.groovy
FlorianHockmann: I don't know if you saw what I wrote in the initial post of this PR, but the problem is that those steps expect a `Function` in Java which is also implemented by enums like `T`. So, when we restrict them to something like `ILambda` in .NET then they can't take enum values anymore as enums can't implement interfaces in .NET.

docs/src/reference/gremlin-variants.asciidoc
FlorianHockmann: I don't think that we need curly brackets here. It also looks like this for Gremlin-Python in the docs: `map(lambda: ("it.get().value('name').length()", "gremlin-groovy"))`

gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/LambdaSerializer.cs
FlorianHockmann: Would you prefer the interface even if we don't use it in the traversal API (see my response to your other comment)? I usually don't introduce interfaces for classes that only hold data without any functionality, but I can of course add such an interface if you prefer it that way.

gremlin-dotnet/glv/generate.groovy
FlorianHockmann: You're right, we should solve this problem, but I think that the changes for that are big enough to validate its own issue and it's also somewhat unrelated to this PR. So I created a new issue for that: [TINKERPOP-1901](https://issues.apache.org/jira/browse/TINKERPOP-1901). I suggest that we resolve that one first and then I simply rebase this PR afterwards and let the `Lambda` class implement whichever interface we introduce there for those cases.

jorgebay: Makes sense 👍

FlorianHockmann:
I just rebased on `tp32` and hopefully also addressed your suggestions. `Lambda` now implements a bunch of interfaces that mirror the Java functional interfaces from the traversal API:
```cs
public class Lambda : IFunction, IBiFunction, IPredicate, IUnaryOperator, IBinaryOperator,
        IComparator, IConsumer, ISupplier
```
This should make it easier for users to see where lambdas can be used. I also added a short note to the documentation about these interfaces for lambdas.

With all these new interfaces there is now just one Java type left for which we don't have a .NET equivalent: `VertexProgram`. So the next release will again increase type-safety of the traversal API a lot.

We are now down to just 5 ignored scenarios for `tp32` 🎉

@jorgebay, you linked the Java `Lambda` interface in one of your comments. The same file also contains classes like `TwoArgLambda` that are very similar to the `Lambda` class I implemented in this PR, but they use [different methods for the different interfaces](https://github.com/apache/tinkerpop/blob/3.2.7/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/function/Lambda.java#L158) like:
```java
public static <A> Consumer<A> consumer(final String lambdaSource,
        final String lambdaLanguage)
```
Do we also want to have similar methods in Gremlin.Net? It would be very easy to implement and I can image that just having a `Lambda` object could lead to some confusion as to where it can be used.

jorgebay:
The implementation looks great and the final `GraphTraversal` and `GraphTraversalSource` APIs make a lot of sense, like:

```csharp
public GraphTraversal<S, E2> Branch<E2> (IFunction function) {}
```

This enables us that if we wanted to in the future, we could implement a lambda in C# that takes a `Func<T, TResult>`, like:

```csharp
IFunction fn = Lambda.GetFunction<string, int>(text => text.length)
int traversal.Branch(fn).Next();
```

Just 1 suggestion:
Can we make `Lambda` class `static` and make the `Lambda.Groovy()` and `Lambda.Python()` return an `ILambda` interface?

```csharp
public interface ILambda : IFunction, IBiFunction, IPredicate,
                           IUnaryOperator, IBinaryOperator, IComparator, IConsumer, ISupplier
```

```csharp
// Make the implementation internal
internal class StringBasedLambda: ILambda
{
  public string LambdaExpression { get; }
  public string Language { get; }
  public object Arguments { get; }
  public Lambda(string expression, string language)
  {
    // ...
  }
}
```

Then, the serializers could take an `ILambda`: `{typeof(ILambda), new LambdaSerializer()}`

FlorianHockmann:
> Can we make Lambda class static and make the Lambda.Groovy() and Lambda.Python() return an ILambda interface?

Sure, making the implementation internal makes perfect sense to me.

But what do you think about the question of whether we want to return a general `ILambda` interface vs specific `IFunction` interfaces and so on?

So first approach (basically the current one in this PR):
```cs
public static class Lambda
{
    public static ILambda Groovy(string expression)
    {
        return new StringBasedLambda(expression, "gremlin-groovy");
    }

    public static ILambda Python(string expression)
    {
        return new StringBasedLambda(expression, "gremlin-python");
    }
}
```

alternative approach:

```cs
public static class Lambda
{
    public static string DefaultLanguage { get; set; } = "gremlin-groovy";

    public static IFunction Function(string expression, string language)
    {
        return new StringBasedLambda(expression, language);
    }

    public static IPredicate Predicate(string expression, string language)
    {
        return new StringBasedLambda(expression, language);
    }

    // ...
    
    public static IPredicate Predicate(string expression)
    {
        return Predicate(expression, DefaultLanguage);
    }

    // ...
}
```

jorgebay:
I would prefer a generic `ILambda` for string based lambdas (groovy and python) but I don't have an strong opinion about it.

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/Lambda.cs
```diff
@@ -0,0 +1,59 @@
...
+#endregion
+
+namespace Gremlin.Net.Process.Traversal
+{
+    /// <summary>
+    ///     Represents a lambda.
+    /// </summary>
+    public interface ILambda : IFunction, IBiFunction, IPredicate, IUnaryOperator, IBinaryOperator, IComparator,
+        IConsumer, ISupplier
+    {
```
jorgebay: We could add the properties to the `ILambda` interface, in the same way it's done in java: https://github.com/apache/tinkerpop/blob/f5f4a046f211851e901d303c6593975314171e55/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/function/Lambda.java#L33-L39

The risk of that interface ever changing is shared with Java and a single serializer (for any `ILambda`) will only be needed.

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/TraversalEvaluation/TraversalParser.cs
```diff
@@ -398,7 +398,7 @@ private static ITokenParameter ParseParameter(string text, ref int i)
             {
                 return ParseNumber(text, ref i);
             }
-            if (text.Substring(i, 3).StartsWith("__."))
+            if (text.Length >= i + 3 && text.Substring(i, 3).StartsWith("__."))
```
dkuppitz: Not a big deal, but with the length check included, it can be an equality check, rather than `StartsWith()`.

FlorianHockmann: Done

PR #793, opened 2018-01-30

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1872

This pull request basically adds an integration test for the fix presented in #788 and indeed this test fails for Spark (while passing for TinkerGraph and Giraph). After merging the change from that branch here, Spark passes, so this all looks good now. 

Thanks to @Stargazer4869 for the original pull request - #788 can be closed at this point.

VOTE +1


dkuppitz:
VOTE: +1

robertdale:
VOTE +1

PR #797, opened 2018-02-12
TINKERPOP-1887 Allow to use bindings in predicates

eonpatapon:
Currently the python GLV does not allow to have bindings in predicates,
eg:

    g.V().has('foo', lt(('bar', 10)))
    g.V().has('foo', inside(('a', 10), ('b', 20)))

This patch fixes this

### DIFF
--- a/gremlin-python/src/main/jython/gremlin_python/process/traversal.py
+++ b/gremlin-python/src/main/jython/gremlin_python/process/traversal.py
@@ -403,6 +403,19 @@ class Bytecode(object):
         if isinstance(arg, Traversal):
             self.bindings.update(arg.bytecode.bindings)
             return arg.bytecode
+        elif isinstance(arg, P):
+            binding = self.__convertArgument(arg.value)
+            if isinstance(binding, Binding):
+                arg.value = binding.value
+            else:
+                arg.value = binding
+            if arg.other is not None:
+                binding = self.__convertArgument(arg.other)
+                if isinstance(binding, Binding):
+                    arg.other = binding.value
+                else:
+                    arg.other = binding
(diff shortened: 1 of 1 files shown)


davebshow:
This needs a test to demonstrate that it works. I haven't run the code yet, but what happens to the key value of the binding? Does this generate GraphSON with bindings, or just the values of the bindings?

eonpatapon:
Yes, can you point me to the test suite ? I wasn't able to find it in the repo

davebshow:
https://github.com/apache/tinkerpop/tree/master/gremlin-python/src/main/jython/tests

eonpatapon:
After playing with the code and tests I'm realizing this does not work at all :)

Closing for now.

davebshow:
There is a bit of a learning curve working with the Gremlin-Python source. Do you plan on giving this another try? If so, I will help you review. If not, let me know and I can give it a go.

eonpatapon:
After playing a bit, I'm not sure if what I want to do is supported by
the gremlin-server. Actually I don't know how the bytecode should look.

Given the following query in python GLV: 

    `g.V().has('foo', P.eq(('bar', 5)))`

I translate it to the following bytecode:

    `[['V'], ['has', 'foo', ['eq', binding[bar=5]]]]`

But this doesn't yield any results from the gremlin server.

In tests I found some other form of query: `g.V().has('foo', ('bar', P.gt(5)))`, ie:

    `[['V'], ['has', 'foo', binding[bar=['eq', 5]]]]`

But it doesn't yield any result either.

Is there some bytecode reference somewhere to read more about it ? I
didn't find any.

PR #798, opened 2018-02-14
TINKERPOP-1857 - Consistency in GLV tests

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1857

This PR largely brings the GLV test suite into consistency with the java test suite. It enables a check that i've long had that will fail the build if someone adds a new test to the java test suite but not to the GLV suite so things should stay in sync going forward. 

That said, this body of work isn't quite perfect in that:

1. There's some failing tests for .NET and JS for which separate JIRA issues have been opened (or will be opened)- it's a small handful of may 30 tests total out of hundreds.
2. There were a bunch of tests that didn't translate well to the GLV suite - i had to "ignore" those for now. New issues will be created once the issues can be classified properly.

Even with those detractors, this remains a pretty large body of work that validates a far greater percentage of our GLV code than before so I think it's ready to be merged as it stands.

VOTE +1

### DIFF
--- a/gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/GherkinTestRunner.cs
+++ b/gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/GherkinTestRunner.cs
@@ -37,8 +37,28 @@ namespace Gremlin.Net.IntegrationTest.Gherkin
 {
     public class GherkinTestRunner
     {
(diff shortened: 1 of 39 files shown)


jorgebay:
The increased coverage is great!

I'm still puzzled by ticket 1891 (`P.not()` serialization) but that's a different issue.

VOTE +1

FlorianHockmann:
LGTM, VOTE +1

PR #800, opened 2018-02-17
TINKERPOP-1726 Added idleReadLimit and idleWriteLimit for Gremlin Server

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1726

This enables Gremlin Server to periodically ping clients and auto-close zombie connections when a client disappears unexpectedly. I did a fair bit of manual testing as this wasn't easy to write unit tests for and it all seems to behave as expected. I did write a decent integration test so that's good. I'm not sure how useful this feature is for the more advanced drivers like the Java Driver because it constantly pings from its end to keep the connection alive even if the client isn't sending requests. I've also found that somehow the server was already pretty good at realizing when a client using the Java Driver goes dead completely (maybe something happening at a network level? i couldn't isolate it) and cleaning up. So...anyway, this seems like more of a precautionary feature for users who find they have a problem like the one described in the JIRA issue or perhaps if they are using a less featured driver. Note that the feature is disabled by default.

VOTE +1


gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/AbstractChannelizer.java
```diff
@@ -147,7 +149,12 @@ public void init(final ServerGremlinExecutor<EventLoopGroup> serverGremlinExecut
     public void initChannel(final SocketChannel ch) throws Exception {
         final ChannelPipeline pipeline = ch.pipeline();
 
-        if (sslContext.isPresent()) pipeline.addLast(PIPELINE_SSL, sslContext.get().newHandler(ch.alloc()));
+        sslContext.ifPresent(sslContext -> pipeline.addLast(PIPELINE_SSL, sslContext.newHandler(ch.alloc())));
+
+        // checks for no activity on a channel and triggers an event that is consumed by the OpSelectorHandler
+        // and either closes the connection or sends a ping to see if the client is still alive
+        if (supportsIdleMonitor())
+            pipeline.addLast(new IdleStateHandler((int) (settings.idleReadLimit / 1000), (int) (settings.idleWriteLimit / 1000),0));
```
jorgebay: Is this setting in millis or seconds?

spmallette: The constructor for `IdleStateHandler` is seconds but all Gremlin Server time settings are milliseconds so i had to convert: https://netty.io/4.0/api/io/netty/handler/timeout/IdleStateHandler.html#IdleStateHandler-int-int-int-

jorgebay: Then I think we should use seconds as a time unit for this setting :)

CHANGELOG.asciidoc
```diff
@@ -25,6 +25,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 * Modified `GremlinDslProcessor` so that it generated the `getAnonymousTraversalClass()` method to return the DSL version of `__`.
 * Added the "Kitchen Sink" test data set.
+* Added `idleReadLimit` and `idleWriteLimit` to Gremlin Server that enables a "ping" and auto-close for seemingly dead clients.
```
robertdale: I think these should be called something more meaningful.
idleReadLimit:  idleConnectionTimeout
idleWriteLimit: keepAliveInterval

docs/src/reference/gremlin-applications.asciidoc
```diff
@@ -1073,7 +1073,8 @@ The following table describes the various configuration options that Gremlin Ser
...
 |gremlinPool |The number of "Gremlin" threads available to execute actual scripts in a `ScriptEngine`. This pool represents the workers available to handle blocking operations in Gremlin Server. When set to `0`, Gremlin Server will use the value provided by `Runtime.availableProcessors()`. |0
 |host |The name of the host to bind the server to. |localhost
-|useEpollEventLoop |try to use epoll event loops (works only on Linux os) instead of netty NIO. |false
+|idleReadLimit |Time in milliseconds that the server will allow a channel to not receive requests from a client before it automatically closes. If enabled, the value provided should typically exceed the amount of time given to `idleWriteLimit`. Set this value to `0` to disable this feature. |0
+|idleWriteLimit |Time in milliseconds that the server will allow a channel to not send responses to a client before it sends a "ping" to see if it is still present. If it is present, the client should respond with a "pong" which will thus reset the {@link #idleReadLimit} and keep the channel open. If enabled, this number should be smaller than the value provided to the `idleReadLimit`. Set this value to `0` to disable this feature. |0
```
robertdale: `{@link #idleReadLimit}` looks like a javadoc comment and doesn't render here.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/AbstractChannelizer.java
robertdale: I think we should keep it millis for consistency across configuration options.  There could be a note stating it has only second-resolution.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerIntegrateTest.java
```diff
@@ -293,6 +312,21 @@ private static SslContext createServerSslContext() {
         return scriptEngineConf;
     }
 
+    @Test
+    public void shouldPingChannelIfClientDies() throws Exception {
```
robertdale: You should be able test for a dead client by opening a raw WebSocket, or even TCP, connection and simply not responding to anything.

Or separately, just disable the idleWriteLimit (keepalive) and connect. Then it should timeout the client.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/AbstractChannelizer.java
spmallette: yeah - i have that same feeling @robertdale - i'm considering some validation or something there in addition to documentation fixes.

spmallette: Adding some general logging that explains the conversion:

```text
[INFO] GremlinServer - idleConnectionTimeout was set to 500 which resolves to 0 seconds when configuring this value - this feature will be disabled
[INFO] GremlinServer - keepAliveInterval was set to 0 which resolves to 0 seconds when configuring this value - this feature will be disabled
```

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerIntegrateTest.java
spmallette: good idea to just disable keepalive - that made it easier to write a test there.

spmallette:
I think I've incorporated all the suggested changes. Please take a look around and see if I missed anything. I will re-run tests on my side.

docs/src/reference/gremlin-applications.asciidoc
```diff
@@ -1073,7 +1073,8 @@ The following table describes the various configuration options that Gremlin Ser
...
 |host |The name of the host to bind the server to. |localhost
-|useEpollEventLoop |try to use epoll event loops (works only on Linux os) instead of netty NIO. |false
+|idleConnectionTimeout |Time in milliseconds that the server will allow a channel to not receive requests from a client before it automatically closes. If enabled, the value provided should typically exceed the amount of time given to `keepAliveInterval`. Note that while this value is to be provided as milliseconds it will resolve to second precision. Set this value to `0` to disable this feature. |0
+|keepAliveInterval |Time in milliseconds that the server will allow a channel to not send responses to a client before it sends a "ping" to see if it is still present. If it is present, the client should respond with a "pong" which will thus reset the `#idleConnectionTimeout` and keep the channel open. If enabled, this number should be smaller than the value provided to the `idleConnectionTimeout`. Note that while this value is to be provided as milliseconds it will resolve to second precision. Set this value to `0` to disable this feature. |0
```
robertdale: I think you want to remove the `#` on 
```
`#idleConnectionTimeout`
```

spmallette: damn

robertdale:
The only thing I don't like is that the client keep-alive doesn't seem to reset the 'write idle time' so the server still sends ping requests to the client even though the client is communicating.

robertdale:
Yes, manually. I was surprised too. In practice, I don't think it's a big deal. The only reason I could see even using server-side keep-alive is because client-side keep-alive was sufficient. So in that case, you probably wouldn't even have client-side keep-alive on.

VOTE +1

jorgebay:
VOTE +1

PR #802, opened 2018-02-25
Add docker images for console and server TINKERPOP-1897

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-1897

The Docker images are very similar to [those of @jbmusso](https://github.com/jbmusso/docker-tinkerpop). Especially the `docker-entrypoint.sh` scripts are basically the same.

### Base image: Alpine Linux
I opted for an OpenJDK base image with Alpine Linux to get the images as small as possible. The Console image now only has a size of 120 MB, and the Server image of 155 MB - compared to 466 MB and 469 MB (uncompressed sizes, hub.docker.com only shows compressed sizes).

### Maven integration
The Maven integration is done with a [plugin from Spotify](https://github.com/spotify/dockerfile-maven). It works similar to how we build the GLVs in that the Docker build is disabled by default and can be enabled either by explicitly selecting the profile `docker-images` or by creating a `.docker` file in the modules directory. (This also documented in the developer docs.)

### To Do: More tags
Right now the images will just get the `project.version` as their tag. We probably want to add `latest` on master for the 3.3.z line and we might want to add tags like `3.3` to allow users to just select the minor version to use and then always get the newest image for that release branch. But we have to make sure that we don't add tags like `3.3.2` to `SNAPSHOT` images.
I'm afraid that my Maven skills are too limited to get those tags without hard coding them and especially to avoid `SNAPSHOT` images from getting normal release tags. My only idea so far is to add a dedicated `execution` element for the plugin that only adds the release tags and that we have to execute manually. But maybe someone can come up with a more automatic way?

### DIFF
--- a/docs/src/dev/developer/development-environment.asciidoc
+++ b/docs/src/dev/developer/development-environment.asciidoc
@@ -197,6 +197,11 @@ authentication information on the ~/.npmrc file. The easiest way to do that is t
 must be done only once, as the auth token doesn't have an expiration date and it's stored on your file system. If
 this account is newly created then request that a PMC member add your account to the "gremlin" package on npm.
 
+Deploying Docker images to link:https://hub.docker.com/[Docker Hub] requires authentication information in the
+`~/.docker/config.json` file. This information can simply be added with the `docker login` command which will ask for
+credentials. This must be done only once. Afterwards, `docker push` can be used to push images to Docker Hub which will
+be done automatically on `mvn deploy` or it can be triggered manually with `mvn dockerfile:push`.
+
 [[building-testing]]
 == Building and Testing
 
--- a/docs/src/dev/developer/development-environment.asciidoc
+++ b/docs/src/dev/developer/development-environment.asciidoc
@@ -197,9 +197,11 @@ authentication information on the ~/.npmrc file. The easiest way to do that is t
 must be done only once, as the auth token doesn't have an expiration date and it's stored on your file system. If
 this account is newly created then request that a PMC member add your account to the "gremlin" package on npm.
 
-Deploying Docker images to link:https://hub.docker.com/[Docker Hub] requires authentication information in the
-`~/.docker/config.json` file. This information can simply be added with the `docker login` command which will ask for
-credentials. This must be done only once. Afterwards, `docker push` can be used to push images to Docker Hub which will
+Deploying Docker images to link:https://hub.docker.com/[Docker Hub] requires an account that is a member of the TinkerPop
+organization. So if you don't already have an account on Docker Hub then create one and request that
+a PMC member adds your account to the TinkerPop organization. Afterwards, authentication information needs to be added to
+the `~/.docker/config.json` file. This information can simply be added with the `docker login` command which will ask for
+credentials. This must be done only once. Finally, `docker push` can be used to push images to Docker Hub which will
 be done automatically on `mvn deploy` or it can be triggered manually with `mvn dockerfile:push`.
 
 [[building-testing]]
--- a/docs/src/dev/developer/development-environment.asciidoc
+++ b/docs/src/dev/developer/development-environment.asciidoc
@@ -106,7 +106,7 @@ The build optionally requires link:https://www.microsoft.com/net/core[.NET Core 
 `gremlin-dotnet` module. If .NET Core SDK is not installed, TinkerPop will still build with Maven, but .NET projects
 will be skipped.
 
-`gremlin-dotnet` can be build and tested from the command line with:
+`gremlin-dotnet` can be built and tested from the command line with:
 
 [source,text]
 mvn clean install -Pgremlin-dotnet
@@ -126,6 +126,20 @@ using `com.github.eirslett:frontend-maven-plugin` plugin. This copy of the Node.
 other existing Node.js runtime instances in your machine.
 
 See the <<release-environment,Release Environment>> section for more information on release manager configurations.
+
+[[docker-environment]]
+=== Docker Environment
+The build optionally requires Docker to build Docker images of Gremlin Server and Gremlin Console.
+
+The Docker images can be built from the command line with:
+[source,text]
+----
+mvn clean install -Pdocker-images
+----
+which enables the "docker-images" Maven profile or in a more automated fashion simply add a `.docker` file to the
+directories of the `gremlin-server` and/or `gremlin-console` modules which will signify to Maven that Docker is present
+in the environment.
+
 
 [[release-environment]]
 === Release Environment
--- a/docs/src/dev/developer/development-environment.asciidoc
+++ b/docs/src/dev/developer/development-environment.asciidoc
@@ -136,9 +136,7 @@ The Docker images can be built from the command line with:
 ----
 mvn clean install -Pdocker-images
 ----
-which enables the "docker-images" Maven profile or in a more automated fashion simply add a `.docker` file to the
-directories of the `gremlin-server` and/or `gremlin-console` modules which will signify to Maven that Docker is present
-in the environment.
+which enables the "docker-images" Maven profile.
 
 
 [[release-environment]]
--- a/docs/src/dev/developer/development-environment.asciidoc
+++ b/docs/src/dev/developer/development-environment.asciidoc
@@ -129,15 +129,16 @@ See the <<release-environment,Release Environment>> section for more information
 
 [[docker-environment]]
 === Docker Environment
-The build optionally requires Docker to build Docker images of Gremlin Server and Gremlin Console.
 
-The Docker images can be built from the command line with:
+The build optionally requires Docker to build Docker images of Gremlin Server and Gremlin Console. The Docker images
+can be built from the command line with:
+
 [source,text]
 ----
-mvn clean install -Pdocker-images
+mvn clean install -pl gremlin-server,gremlin-console -DdockerImages
 ----
-which enables the "docker-images" Maven profile.
 
+which enables the "docker-images" Maven profile.
 
 [[release-environment]]
 === Release Environment
--- a/docs/src/dev/developer/release.asciidoc
+++ b/docs/src/dev/developer/release.asciidoc
@@ -111,7 +111,7 @@ under release is protected. Tweaks to documentation and other odds and ends rela
 during this period.
 . At some point during the week:
 .. Run the full integration test suite: `mvn clean install -DskipIntegrationTests=false -DincludeNeo4j`
-.. Build and test the Docker images: `mvn clean install -pl gremlin-server,gremlin-console -Pdocker-images`
+.. Build and test the Docker images: `mvn clean install -pl gremlin-server,gremlin-console -DdockerImages`
 .. Deploy a final SNAPSHOT to the snapshot repository as well as GLV and Docker image pre-releases.
 .. Review LICENSE and NOTICE files to make sure that no <<dependencies,changes are needed>>.
 .. Review javadoc filters on the "Core API" docs to be sure nothing needs to change.
@@ -235,8 +235,8 @@ for generating javadoc and without that the binary distributions won't contain t
 . Review the GLV releases at link:https://pypi.org/project/gremlinpython/[PyPi],
 link:https://www.nuget.org/packages/Gremlin.Net/[nuget] and link:https://www.npmjs.com/package/gremlin[npm]
 . Deploy the Docker images
-.. `mvn deploy -pl gremlin-console -DskipTests -Pdocker-images`
-.. `mvn deploy -pl gremlin-server -DskipTests -Pdocker-images`
+.. `mvn deploy -pl gremlin-console -DskipTests -DdockerImages`
+.. `mvn deploy -pl gremlin-server -DskipTests -DdockerImages`
 . Review the deployed Docker images at link:https://hub.docker.com/r/tinkerpop/gremlin-console/[Console]
 and link:https://hub.docker.com/r/tinkerpop/gremlin-server/[Server]
 . `svn co --depth empty https://dist.apache.org/repos/dist/dev/tinkerpop dev; svn up dev/xx.yy.zz`
--- a/docs/src/dev/developer/release.asciidoc
+++ b/docs/src/dev/developer/release.asciidoc
@@ -111,7 +111,8 @@ under release is protected. Tweaks to documentation and other odds and ends rela
 during this period.
 . At some point during the week:
 .. Run the full integration test suite: `mvn clean install -DskipIntegrationTests=false -DincludeNeo4j`
-.. Deploy a final SNAPSHOT to the snapshot repository as well as GLV pre-releases.
+.. Build and test the Docker images: `mvn clean install -pl gremlin-server,gremlin-console -Pdocker-images`
+.. Deploy a final SNAPSHOT to the snapshot repository as well as GLV and Docker image pre-releases.
 .. Review LICENSE and NOTICE files to make sure that no <<dependencies,changes are needed>>.
 .. Review javadoc filters on the "Core API" docs to be sure nothing needs to change.
 .. Review JIRA tickets in the release and ensure that:
--- a/docs/src/dev/developer/release.asciidoc
+++ b/docs/src/dev/developer/release.asciidoc
@@ -233,6 +233,11 @@ for generating javadoc and without that the binary distributions won't contain t
 .. `mvn deploy -pl gremlin-javascript -DskipTests -Dnpm`
 . Review the GLV releases at link:https://pypi.org/project/gremlinpython/[PyPi],
 link:https://www.nuget.org/packages/Gremlin.Net/[nuget] and link:https://www.npmjs.com/package/gremlin[npm]
+. Deploy the Docker images
+.. `mvn deploy -pl gremlin-console -DskipTests -Pdocker-images`
+.. `mvn deploy -pl gremlin-server -DskipTests -Pdocker-images`
+. Review the deployed Docker images at link:https://hub.docker.com/r/tinkerpop/gremlin-console/[Console]
+and link:https://hub.docker.com/r/tinkerpop/gremlin-server/[Server]
 . `svn co --depth empty https://dist.apache.org/repos/dist/dev/tinkerpop dev; svn up dev/xx.yy.zz`
 . `svn co --depth empty https://dist.apache.org/repos/dist/release/tinkerpop release; mkdir release/xx.yy.zz`
 . Copy release files from `dev/xx.yy.zz` to `release/xx.yy.zz`.
--- a/docs/src/reference/gremlin-applications.asciidoc
+++ b/docs/src/reference/gremlin-applications.asciidoc
@@ -1946,31 +1946,7 @@ $ docker run tinkerpop/gremlin-server:x.y.z
 -----oOOo-(3)-oOOo-----
 
 [INFO] GremlinServer - Configuring Gremlin Server from conf/gremlin-server.yaml
-[INFO] MetricManager - Configured Metrics ConsoleReporter configured with report interval=180000ms
-[INFO] MetricManager - Configured Metrics CsvReporter configured with report interval=180000ms to fileName=/tmp/gremlin-server-metrics.csv
-[INFO] MetricManager - Configured Metrics JmxReporter configured with domain= and agentId=
-[INFO] MetricManager - Configured Metrics Slf4jReporter configured with interval=180000ms and loggerName=org.apache.tinkerpop.gremlin.server.Settings$Slf4jReporterMetrics
-[INFO] DefaultGraphManager - Graph [graph] was successfully configured via [conf/tinkergraph-empty.properties].
-[INFO] ServerGremlinExecutor - Initialized Gremlin thread pool.  Threads in pool named with pattern gremlin-*
-[INFO] ScriptEngines - Loaded gremlin-groovy ScriptEngine
-[INFO] GremlinExecutor - Initialized gremlin-groovy ScriptEngine with scripts/empty-sample.groovy
-[INFO] ServerGremlinExecutor - Initialized GremlinExecutor and preparing GremlinScriptEngines instances.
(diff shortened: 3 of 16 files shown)


docs/src/reference/gremlin-applications.asciidoc
```diff
@@ -1830,6 +1862,63 @@ $ curl -X POST -d "{\"gremlin\":\"divideIt(8, 2)\"}" "http://localhost:8182"
...
+[source,text]
+----
+$ docker run tinkerpop/gremlin-server:x.y.z
+[INFO] GremlinServer - 
+         \,,,/
+         (o o)
+-----oOOo-(3)-oOOo-----
+
+[INFO] GremlinServer - Configuring Gremlin Server from conf/gremlin-server.yaml
+[INFO] MetricManager - Configured Metrics ConsoleReporter configured with report interval=180000ms
```
spmallette: I suggest you elide some of the log output here....less to maintain should the output ever change. maybe "..." everything starting here down to (but not including):

```text
[INFO] GremlinServer$1 - Gremlin Server configured with worker thread pool of 1, gremlin pool of 4 and boss thread pool of 1.
```

spmallette:
Thanks for doing this @FlorianHockmann - i'm just dumping my brain of questions and thoughts on this one.

So `mvn deploy` will deploy to where? dockerhub?  If so, could you include something in the dev docs on how you configure your system to allow for that? 

i assume we would want it to deploy here: https://hub.docker.com/u/tinkerpop/ but i can't remember  who controls that...i think it's one of us......

Also, what happens when we do a `mvn deploy` on a SNAPSHOT? that would be bad right? I think we need to prevent such things from happening somehow.

what does adding this extra step do to the the overall build time (when building docker is enabled, of course)? I like that you made it a separate `<profile>` and used the `.docker` pattern we have with `.glv`. Part of this PR will likely need to include updates for the release process as well.

spmallette:
I just reset my docker account password and it turns out that i'm one of the owners on the TinkerPop organization - so that's good.

FlorianHockmann:
I just pushed two commits, one removes the log output as you suggested and the other one was unfortunately necessary on my system to get all tests running with `./docker/build.sh -t -n -i`. It simply increases the timeout for one test in `gremlin-driver` as that was failing consistently for me.
As already discussed on the mailing list, I can't build the documentation right now, so I can't test if my changes broke anything there.


> So mvn deploy will deploy to where? dockerhub?

My understanding of the `dockerfile-maven-plugin` is that it will simply wrap `docker push` for deployment which should default to dockerhub.


>  If so, could you include something in the dev docs on how you configure your system to allow for that?

It should be enough to do [`docker login`](https://docs.docker.com/engine/reference/commandline/login/) once on your system which should ask for the credentials and then store them in the config. After that, deployment through maven should just work.

How about we deploy release candidate images once this is merged and then write the release docs when we've actually done one successful deployment?
Otherwise I could also test deployment with this plugin with some test image to a private repo and then write the dev docs.

> Also, what happens when we do a mvn deploy on a SNAPSHOT? that would be bad right? I think we need to prevent such things from happening somehow.

I'm not sure yet about that one yet. It would just result in an image with a tag like `3.2.8-SNAPSHOT` which shouldn't be a problem.
The ideal case would probably be that we use unique tags like `3.2.8-SNAPSHOT-abcde` and then tag the latest of these SNAPSHOT images additionally with `3.2.8-SNAPSHOT`, but just having the `3.2.8-SNAPSHOT` tag should also be ok.
Now, doing a `mvn deploy` and therefore `docker push` for an image that was already pushed shouldn't do anything when the artifacts didn't change that go in that image.

Alternatively, we could also try to disable the deployment phase for this plugin for `SNAPSHOT` versions, but I don't know how we could do that with Maven.

> what does adding this extra step do to the the overall build time (when building docker is enabled, of course)?

I just checked with my Ubuntu VM that has just 4 cores and 6 GB RAM and got the following times for `mvn install -pl gremlin-server -DskipTests`:
* Without docker build enabled: 33s
* With docker build enabled and no images present: 2:10 min
* With docker build enabled, but the image already present: 51s

When no images are present locally, then it has to download and extract the OpenJDK image which is 82 MB big, that took a big part of that time.

spmallette:
> Otherwise I could also test deployment with this plugin with some test image to a private repo and then write the dev docs.

That would be great if you could test things out in private first since that is an option. It would be nice to know that we've worked out all the kinks in a throwaway repo before we start putting things out in public.

> It should be enough to do docker login

that's good to know - a few words on that in the dev docs should be sufficient so that no one has to go look it up in google.

> Now, doing a mvn deploy and therefore docker push for an image that was already pushed shouldn't do anything when the artifacts didn't change that go in that image.

if i understand you correctly, you're saying that if you push the image to 3.3.2-SNAPSHOT and then make changes and push again, the 3.3.2-SNAPSHOT would update...is that right? if that is the case, then that behavior seems to match the nature of a java "SNAPSHOT" (except that there is no archival of the older SNAPSHOT).  Either way though, I don't think we need to pollute the TinkerPop dockerhub with SNAPSHOT deployments. I think we should look for ways to disable the docker plugin on  `deploy` if it is a SNAPSHOT version. I'm not sure that there is a way to do that with maven directly, but I think we can come up with a way to have the effect of not sending SNAPSHOT images to dockerhub.

FlorianHockmann:
I just verified my assumption that `docker login` should be enough to get deployment of Docker images working through Maven and added a short description of that to the dev docs.

> if i understand you correctly, you're saying that if you push the image to 3.3.2-SNAPSHOT and then make changes and push again, the 3.3.2-SNAPSHOT would update...is that right? if that is the case, then that behavior seems to match the nature of a java "SNAPSHOT" (except that there is no archival of the older SNAPSHOT).

Yes, exactly. Changing something about the image means that the ID will change and the tag will be added to the new image. So the old one won't be available any more through Docker Hub as it isn't tagged any more. That way, we only have one image per tag available and it will always be the most recent one.

>  Either way though, I don't think we need to pollute the TinkerPop dockerhub with SNAPSHOT deployments. I think we should look for ways to disable the docker plugin on deploy if it is a SNAPSHOT version.

I agree, I just don't know how this could be done. The plugin [contains `skip` flags](https://github.com/spotify/dockerfile-maven#skip-docker-goals-bound-to-maven-phases), but we have to activate them somehow for SNAPSHOT versions.

spmallette:
i saw a few solutions to the issue of skipping when SNAPSHOT - not sure which is best, but here's one:

https://stackoverflow.com/a/39139979/1831717

we might actually find usage for this elsewhere in tinkerpop once the process is established. i seem to remember some hacky things we do around this problem, but i don't recall exactly what......maybe i'm making it up :)

FlorianHockmann:
Thanks @spmallette, that actually looks like a good solution.

I just pushed a commit that uses this solution. This not only allows skipping of `docker push` for SNAPSHOT versions, but I could also enable tagging of the minor version when the version is not SNAPSHOT, e.g., `3.2` for the `tp32` branch. I can also add the `latest` tag on `tp33` either with another PR or while merging from `tp32`. So the _To Do_ from this PR's initial post is now also done.

To sum up what happens for the different versions:

* `3.y.z-SNAPSHOT`:
  * Images are just built.
  * Tag: `3.y.z-SNAPSHOT`
* `3.2.z`:
  * Images are built and pushed.
  * Tags: `3.2.z` and `3.2`
* `3.3.z` (not directly part of this PR):
  * Images are built and pushed.
  * Tags: `3.3.z`, `3.3`, and `latest`

I also verified the generated docs with `./docker/build.sh -d`.

VOTE +1

spmallette:
I haven't looked at this in detail yet, but is your intent to see this merged before we release 3.2.8/3.3.2? Personally, I'd like to see it wait until a following release so that we don't need to do any release instruction changes. I think we might even want to try to do a release candidate to docker first to make sure we get the release process right for when we go official. thoughts?

FlorianHockmann:
I'm ok with merging this after the next release. Starting with a release candidate first like we usually do for GLVs makes sense in my opinion. Then we can also deal with things like the description for Docker Hub and more people can try out the images before we do an official release.

spmallette:
Since the `build-helper-maven-plugin` configuration is identical in both the gremlin-server/gremlin-console pom.xml files and the "SNAPSHOT detection" would be generally useful (i think it has value to the GLVs) i think that you should be able to move that entire configuration to the root pom.xml (where the `build-helper-maven-plugin` is already being used). I'm pretty sure that doing that will make that `only.when.is.snapshot.used` to all sub-modules. I think you can then remove the definition for `build-helper-maven-plugin` from the `<pluginManagement>` element.

FlorianHockmann:
> move that entire configuration to the root pom.xml

Good point @spmallette, I just pushed a commit for that and also rebased the PR.

gremlin-console/pom.xml
```diff
@@ -336,5 +336,59 @@ limitations under the License.
...
+                                    </buildArgs>
+                                </configuration>
+                            </execution>
+                            <execution>
+                                <id>docker-image-tag-minor-version</id>
+                                <goals>
+                                    <goal>tag</goal>
+                                </goals>
+                                <configuration>                                                     
+                                    <tag>3.2</tag>
```
spmallette: I guess the `<tag>` must be manually updated, but perhaps that not a problem because it stays bound to the branch from patch version to patch version. Though, I do feel like that might be an easy thing to forget on minor version upgrades - perhaps the dev docs should have a note about that wherever we bump version back to SNAPSHOT. Or perhaps this could be made more dynamic somehow?

spmallette:
@FlorianHockmann could you rebase this please? maybe we can try to get this polished up now.

FlorianHockmann:
Rebased and removed the hard-coded version. The major and minor version are now parsed with the `build-helper-maven-plugin` and added as tags to the image when it's not a `SNAPSHOT` version.

docs/src/dev/developer/development-environment.asciidoc
```diff
@@ -183,6 +197,11 @@ authentication information on the ~/.npmrc file. The easiest way to do that is t
 must be done only once, as the auth token doesn't have an expiration date and it's stored on your file system. If
 this account is newly created then request that a PMC member add your account to the "gremlin" package on npm.
 
+Deploying Docker images to link:https://hub.docker.com/[Docker Hub] requires authentication information in the
```
spmallette: Please add a list item here in the "Release & Promote" section so that we remember to publish docker images:

http://tinkerpop.apache.org/docs/current/dev/developer/#_release_promote

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/ResultQueueTest.java
```diff
@@ -281,7 +281,7 @@ public void shouldDrainAsItemsArrive() throws Exception {
                 latch.countDown();
             });
 
-            assertThat(latch.await(3000, TimeUnit.MILLISECONDS), is(true));
+            assertThat(latch.await(10000, TimeUnit.MILLISECONDS), is(true));
```
spmallette: any specific reason for the increase...were you just getting some timeouts here?

spmallette:
what is the advantage to building these images if a `.docker` file is present? don't we really just need to build those images on demand (i.e. manually enable the profile with ` -Pdocker-images`)? like, if someone was testing them specifically? or if someone was deploying them?

docs/src/dev/developer/development-environment.asciidoc
spmallette: "Release Environment" should also make mention that the person doing the deployment needs a dockerhub account and needs to request addition from a PMC member to the TinkerPop organization.

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/ResultQueueTest.java
spmallette: not need to revert if it was causing hassles. just curious.

docs/src/reference/gremlin-applications.asciidoc
```diff
@@ -353,6 +353,38 @@ variable initialization code into the console.
 Like, execution mode, it is also possible to pass multiple scripts by specifying multiple `-i` options. See the
 <<execution-mode, Execution Mode Section>> for more information on the specifics of that capability.
 
+[[gremlin-console-docker-image]]
+=== Docker Image
+The Gremlin Console can also be started as a link:https://hub.docker.com/r/tinkerpop/gremlin-console/[Docker image]:
```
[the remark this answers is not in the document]
FlorianHockmann: Definitely a good idea! It's easy to verify the Docker images themselves as the Dockerfiles are pretty clear and you can build and then test the images locally, but the deployment is definitely harder to test. I tried it with a small example Maven project but the integration in the larger TinkerPop project makes it of course more complicated and we could also use a SNAPSHOT deployment to check whether something is still missing in the documentation regarding deployment.

FlorianHockmann:
> what is the advantage to building these images if a .docker file is present?

A reason to keep this option is that it makes it easier to detect when changes break the Docker images, e.g., when the naming of the zip archives or their location changes. For those cases, it would be good to build the images with every Maven build (assuming that Docker is installed locally).
The build should also basically be a no-op when nothing changed about the Docker images, although the build time slightly increases from 33s to 51s for `gremlin-server` on my system without and with activated `docker-images` profile and the image already present (from my comment on 1 Mar).

I expanded the dev docs for the two aspects you mentioned (Docker Hub account and release process for the Docker images).

spmallette:
> detect when changes break the Docker images, e.g., when the naming of the zip archives or their location changes

That's fair, though that specific example seems really unlikely to occur to me at this point and if it did happen the fix is relatively quick (i would guess) so even if it happened during code freeze week we wouldn't have a big problem.  We could even add a point to the pre-flight check to make sure we generate images during code freeze - http://tinkerpop.apache.org/docs/current/dev/developer/#_pre_flight_check 

Can you think of any other reasons it would fail to build based on more "day-to-day" changes?

FlorianHockmann:
No, you're right. That is something that is very unlikely to occur and I can't think of other (more likely) reasons where the build of the docker images could fail as they really only copy the packaged artefacts together with the `docker-entrypoint.sh` scripts into the image.
It is more likely that the built images themselves stop working for some reason, but we couldn't find that during the maven build anyway. So making it a habit to check the Docker images before a release seems to be the better option to me.

I'll push a commit to remove the `.docker` file option and add a note to the docs section you linked.

docs/src/dev/developer/release.asciidoc
```diff
@@ -111,7 +111,8 @@ under release is protected. Tweaks to documentation and other odds and ends rela
 during this period.
 . At some point during the week:
 .. Run the full integration test suite: `mvn clean install -DskipIntegrationTests=false -DincludeNeo4j`
-.. Deploy a final SNAPSHOT to the snapshot repository as well as GLV pre-releases.
+.. Build and test the Docker images.
```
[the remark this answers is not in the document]
FlorianHockmann: Right, that should be included. I just amended my last commit accordingly.

FlorianHockmann:
@spmallette you now voted +1 on this but didn't you want to push release candidates of the Docker images during the review?

jorgebay:
Nice!

VOTE +1

spmallette:
@FlorianHockmann I think you can merge this. I will do the deploy afterwards when I can tag the rc1 properly. If something goes wrong we can just CTR in some fixes. I'm feeling pretty confident about how this is going to work now after playing with things locally a bit. I also just pushed a minor update to make the command more like our others with `-DdockerImages` rather than explicitly using the `-P` directly which we typically dont' do. Also disabled the deployment of java artifacts since docker deployments will occur out of band and we won't want those re-deployed when we go to deploy to docker.

When I see this merged, I will do the rc1 deployment. This was a nice PR and I think people will really appreciate this. Thanks!

FlorianHockmann:
> I also just pushed a minor update to make the command more like our others with -DdockerImages rather than explicitly using the -P directly which we typically dont' do. Also disabled the deployment of java artifacts since docker deployments will occur out of band and we won't want those re-deployed when we go to deploy to docker.

Makes sense.

I just rebased this on `tp32` as there was a small merge conflict in the CHANGELOG and will merge it now.

Really excited to see TinkerPop's first own Docker images 😃

spmallette:
I've published but look at what is up there:

https://hub.docker.com/u/tinkerpop/

I only see the 3.2 tag - where the 3.2.10-rc1? shouldn't that be up there too?

FlorianHockmann:
> I only see the 3.2 tag - where the 3.2.10-rc1? shouldn't that be up there too?

Yes and the `3.2` tag should only exist for non-SNAPSHOT versions, but I guess that we didn't think about release candidate versions when we added that check:
```xml
<name>only.when.is.snapshot.used</name>
<value>${project.version}</value>
<regex>.*-SNAPSHOT</regex>
```
https://github.com/apache/tinkerpop/blob/tp32/pom.xml#L283

Regarding the full version (3.2.10-rc1), my assumption was that this is already handled by:
```xml
<configuration>
    <tag>${project.version}</tag>
```
https://github.com/apache/tinkerpop/blob/tp32/gremlin-console/pom.xml#L369
and I'm pretty sure that it worked like that in my tests.

But maybe we can add the tagging as a separate `execution` just for this tag?

FlorianHockmann:
> Separately, do I need to edit the repository short/long description manually? can that not be automatically done on deploy using content from our github repo?

I'm afraid that only works for automated builds which we can't use as it would have to work with just a Dockerfile and all other files already being present. Since we need to build the artifacts first with Maven this seems to be no option for us.
So, yes, I guess we have to edit the descriptions manually in the web interface.

I just [CTRed a commit](https://github.com/apache/tinkerpop/commit/1c3f34f52b2651897aa2fe89527cae8d2357dcc6) to tp32 that made the two `docker-entrypoint.sh` scripts executable as the Docker images contained versions of them that were actually not executable making the images basically unusable.
I assumed that git would preserve the executable bit as I set it locally, but apparently it was necessary to set it manually again :-/

spmallette:
> Yes and the 3.2 tag should only exist for non-SNAPSHOT versions, 

I believe that aspect of this works fine. Nothing deploys when the version is a SNAPSHOT.

> But maybe we can add the tagging as a separate execution just for this tag?

it appears you need multiple `push` executions:

https://github.com/spotify/dockerfile-maven/issues/10

i'm trying it now.

> So, yes, I guess we have to edit the descriptions manually in the web interface.

ok - not a big deal. i can fix that.

FlorianHockmann:
> I believe that aspect of this works fine. Nothing deploys when the version is a SNAPSHOT.

But do we want to add the `3.2` tag for a `3.2.10-rc1` image? That means that users who specify `3.2` and were previously on an official release `3.2.9` that they now get a release candidate version.
IMO release candidate images should only have their full version as a Docker tag and not something like `3.2` as that usually implies a stable version.

spmallette:
so - that changes works. we can tweak it now if necessary now that we know it works:

https://github.com/apache/tinkerpop/commit/144c69801de8135ac04663da4063260378c71336

> But do we want to add the 3.2 tag for a 3.2.10-rc1 image? ... IMO release candidate images should only have their full version as a Docker tag and not something like 3.2 as that usually implies a stable version.

As i don't really use docker, I don't know what people tend to expect when they use it. Can you tweak the poms accordingly please with some CTR commits and I can try to deploy again?

spmallette:
Updated the DockerHub descriptions a bit - looks prettier now:

https://hub.docker.com/r/tinkerpop/gremlin-server/
https://hub.docker.com/r/tinkerpop/gremlin-console/

FlorianHockmann:
Just pushed two CTR commits to hopefully get the tags right:

[aac534e11](https://github.com/apache/tinkerpop/commit/aac534e11fa406bedc61dbc060d06740b898e170) adds a check to only tag Docker images with `3.2` when the version is not a prerelease version, so not `3.2.10-SOMETHING`.

[d4faa66](https://github.com/apache/tinkerpop/commit/d4faa6605495e072d65578682d2ed5fd12bdb5ad) adds the `latest` tag on `tp33` branch and on `master`. Once we release 3.4.0 we need to remove the `latest` tagging from the `tp33` branch so that `3.4.0` will be latest. (I don't see a way to automate this so we really have to remember to do it manually before the release of 3.4.0.)

This should result in the following tags:
* tp32 release candidates: 3.2.10-rc1
* tp32 releases: 3.2.10, 3.2
* tp33 release candidates: 3.3.4-rc1
* tp33 releases: 3.3.4, 3.3, latest

To give a short explanation of why this tagging makes sense in my opinion:
* Users who always want to have the newest version that is still stable can use the `latest` tag. This is also implicitly used when you do `docker run tinkerpop/gremlin-console`. So `latest` should always be available and always point to something stable.
* Users who to stay on `tp32` for example because their graph provider doesn't support 3.3.0 yet can use the `3.2` tag and will always get the latest stable version of that branch. (Same applies of course to the `3.3` tag.)
* All other users will select a version explicitly. This could either be a stable version like `3.2.10` or a release candidate when they want to check out new features earlier, but they should always get a stable version when they choose a tag like `3.2.10`.

spmallette:
Here's what I just did:

* Confirmed that SNAPSHOT does not deploy images.
* Confirmed that 3.2.10-rc1 deploys 3.2.10-rc1 only - don't see 3.2 up there any more

I guess that's as much as I can test for now until we release. I will say that so far these tweaks haven't been painful so if we have to sort something out the next time we do a release it shouldn't be too challenging to get it done.

> Once we release 3.4.0 we need to remove the latest tagging from the tp33 branch so that 3.4.0 will be latest. 

Please make a ticket in JIRA as a reminder for that and assign it a fix version of 3.4.0 so we don't forget.

Other than that, I think this is all good now. Feel free to close the TINKERPOP-1897 in JIRA now. Stellar job on this @FlorianHockmann - thanks for sticking with it to get it to the end.

FlorianHockmann:
> Please make a ticket in JIRA as a reminder for that and assign it a fix version of 3.4.0 so we don't forget.

Done: [TINKERPOP-1980](https://issues.apache.org/jira/browse/TINKERPOP-1980)

And thank you, @spmallette, for the thorough review and for doing the first deployment with the fixes you had to do to get it actually working.

PR #803, opened 2018-02-27
TINKERPOP-1522 Order of select() scopes

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1522

This PR changes the order of `select()` scope from 

* side-effects, maps, paths

to

* maps, side-effects, paths

Note that I discovered a bug in OLAP while writing the test case, that's why there is an extra `barrier()` step. I'll create another ticket to address this issue.

### DIFF
--- a/docs/src/upgrade/release-3.4.x.asciidoc
+++ b/docs/src/upgrade/release-3.4.x.asciidoc
@@ -64,7 +64,6 @@ gremlin> g.V().group().
 ......1>   by(label).
 ......2>   by(outE().values("weight").sum())
 ==>[person:3.5]
-gremlin> 
 gremlin> g.V().group().
 ......1>   by(label).
 ......2>   by(coalesce(outE().values("weight"), constant(0)).sum())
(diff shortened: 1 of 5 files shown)


robertdale:
VOTE +1

spmallette:
As this is has a subtle "breaking" aspect to it, I think you need to write up some upgrade docs that explain the change in Gremlin semantics.

docs/src/upgrade/release-3.4.x.asciidoc
```diff
@@ -28,3 +28,8 @@ NEED AN IMAGE
 Please see the link:https://github.com/apache/tinkerpop/blob/3.4.0/CHANGELOG.asciidoc#release-3-4-0[changelog] for a complete list of all the modifications that are part of this release.
 
 === Upgrading for Users
+
+==== Change in order of select() scopes
+
+The order of select scopes has been changed to: maps, side-effects, paths
+Previously the order was: side-effects, maps, paths - which made it almost impossible to select a specific map entry if a side-effect with the same name existed.
```
spmallette: two things: 

1. Could you go the extra step and add an example of the behavior before/after the change. I think people would understand it better if they could see the example (remember the source in upgrade is static so you should just copy/paste the example from your console)

2. note that we typically add a link to the relevant JIRA ticket: https://github.com/apache/tinkerpop/blame/master/docs/src/upgrade/release-3.3.x.asciidoc#L53

spmallette:
VOTE +1 - pending my comment on the upgrade docs

PR #805, opened 2018-03-02
TINKERPOP-1777 Gremlin .max step returns -2147483648 for empty result sets

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1777

Fixed the behavior of `min()`, `max()`, `mean()` and `sum()`.
If no input is given, these steps will now throw a `FastNoElementException`.

Also adjusted existing test cases to be in line with the new behavior.

VOTE: +1


CHANGELOG.asciidoc
```diff
@@ -25,6 +25,8 @@ NEED AND IMAGE
 
 This release also includes changes from <<release-3-3-2, 3.3.2>>.
 
+* Fixed a bug in `ReducingBarrierStep`, that returned the provided seed value despite no elements being available.
```
spmallette: Should this be a changelog entry for 3.4.0 as this is pointed at the master branch?

docs/src/upgrade/release-3.4.x.asciidoc
```diff
@@ -28,3 +28,45 @@ NEED AN IMAGE
...
+gremlin> g.V().group().
+......1>   by(label).
+......2>   by(outE().values("weight").sum())
+==>[person:3.5]
+gremlin> 
+gremlin> g.V().group().
+......1>   by(label).
+......2>   by(coalesce(outE().values("weight"), constant(0)).sum())
+==>[software:0,person:3.5]
+----
```
spmallette: please add the "See: {jira_issue}"

gremlin-test/features/branch/Union.feature
```diff
@@ -135,7 +135,6 @@ Feature: Step - union()
       | d[3].l   |
       | d[0].l   |
       | d[1.9].d |
-      | d[0].i   |
```
spmallette: just so i'm clear, this value is no longer in the output because vadas (aka `v[2]`) has no out edges to sum a weight on...is that right? if so, do you happen to know what catches the thrown `FastNoSuchElementException` and prevents it from bubbling up just kill the whole traversal? (i'm just trying to follow how the code works here for my own edification)

dkuppitz: > just so i'm clear, this value is no longer in the output because vadas (aka v[2]) has no out edges to sum a weight on...is that right?

Exactly.

It's the same thing as doing just `g.V(2).outE()` - the traverser would simply die as there are no outgoing edges. Likewise with `outE().values('weight').sum()`, there simply is no sum, hence the traverser dies.

PR #807, opened 2018-03-03
TINKERPOP-1738 Bump to jackson 2.9.4

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1738

This fixes a problem where Jackson was missing an error in GraphSON type information depending on the order of the JSON elements. It required some internal changes to the GraphSONTypeSerializer given API shifts in jackson - should not really affect TinkerPop  users. Modified the original test that demonstrated this problem to be a bit more robust.

VOTE +1


dkuppitz:
VOTE: +1

Joengenduvel:
VOTE: +1
We cannot proceed using Jackson 2.9.3 and below: https://nvd.nist.gov/vuln/detail/CVE-2017-17485

TomCools:
VOTE: +1 same issue as Joengenduvel. OWASP scan is required for our project and this is listed as a Major level issue.

jorgebay:
VOTE: +1

PR #810, opened 2018-03-03
TINKERPOP-1446 Added standard string representation for Path objects

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1446

Only applying this to 3.4.x as people might be relying on the old string representations of `Path` objects for some reason. 

VOTE +1


dkuppitz:
VOTE: +1

FlorianHockmann:
VOTE +1

PR #812, opened 2018-03-06
TINKERPOP-1911 Refactored JavaTranslator

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1911

Cached methods that were being reflected on every translation which improved performance a bit. Here's how it runs before the change:

```text
Benchmark                                             Mode  Cnt        Score        Error  Units
JavaTranslatorBenchmark.testTranslationLong          thrpt   20    13267.284 ±   1166.704  ops/s
JavaTranslatorBenchmark.testTranslationMedium        thrpt   20   334756.147 ±  23429.373  ops/s
JavaTranslatorBenchmark.testTranslationShort         thrpt   20  1381906.511 ± 141710.801  ops/s
JavaTranslatorBenchmark.testTranslationWithStrategy  thrpt   20    36350.729 ±  2518.351   ops/s
```

and here's how it runs after:

```text
Benchmark                                             Mode  Cnt        Score       Error  Units
JavaTranslatorBenchmark.testTranslationLong          thrpt   20    15801.483 ±   546.613  ops/s
JavaTranslatorBenchmark.testTranslationMedium        thrpt   20   365673.060 ±  6101.193  ops/s
JavaTranslatorBenchmark.testTranslationShort         thrpt   20  1560609.216 ± 37680.931  ops/s
JavaTranslatorBenchmark.testTranslationWithStrategy  thrpt   20    39533.823 ±  1317.264  ops/s
```

Not a huge difference but seemed to be worth doing. 

All tests pass with `mvn clean install && mvn verify -pl gremlin-server -DskipIntegrationTests=false`

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -27,6 +27,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Modified `GremlinDslProcessor` so that it generated the `getAnonymousTraversalClass()` method to return the DSL version of `__`.
 * Added the "Kitchen Sink" test data set.
 * Fixed deserialization of `P.not()` for GraphSON.
+* Improved performance of `JavaTranslator` by caching reflected methods required for traversal construction.
 * Added `idleConnectionTimeout` and `keepAliveInterval` to Gremlin Server that enables a "ping" and auto-close for seemingly dead clients.
 * Fixed a bug in `NumberHelper` that led to wrong min/max results if numbers exceeded the Integer limits.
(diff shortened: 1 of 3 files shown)


BrynCooke:
+1 (Non binding)

robertdale:
What's the difference between the global method cache and the local method cache?

spmallette:
The local cache is specific to the `GraphTraversalSource` that's passed into the `of()` construction method. As the `GraphTraversalSource` is specific to the instance and is not known until construction I figured it was better if it were managed local to the `JavaTranslator` instance

jorgebay:
lgtm, VOTE +1

PR #813, opened 2018-03-08
TINKERPOP-1914 Support construct a GremlinServer instance from gremlin executor service

javeme:
https://issues.apache.org/jira/browse/TINKERPOP-1914

This constructor is useful when Gremlin Server is being used in an embedded
style and there is a need to share thread pools with the hosting application.


spmallette:
i think this is ok to add back. i obviously never factored out the underlying ability to pass that service to the `ServerGremlinExecutor` so I think it might be ok to allow it to be passed in on the `GremlinServer` class. Currently, this PR is targeted at the master branch which means it is aimed at not being available until 3.4.0. If you would like to see this change in the 3.3.2 release that will happen in a few weeks, then you should re-target to the tp33 branch.

javeme:
@spmallette Thanks for reminding, I already re-target to the `tp33` branch: https://github.com/apache/tinkerpop/pull/821

spmallette:
ok - thanks. would you mind closing this PR please? i will merge forward the other one to master.

PR #814, opened 2018-03-09
TINKERPOP-1896 Fixed bug in lambda processing for python

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1896

When a lambda is in a traversal, the bytecode gets pushed to a scriptengine for evaluation. For python that means, using the JythonTranslator. Prior to this change the JythonTranslator largely relied on the PythonTranslator to convert bytecode to Jython-syntaxed Gremlin. The problem there is that in the JythonScriptEngine, the Traversal bindings are Java objects and so if translated to pure Python syntax, traversal steps like as() become as_() and thus not part of the Java API that is expected - and then errors. Not sure why this didn't come up sooner, but it really only ends up being an issue if you are using bytecode, use gremlin server, use jython in gremlin server and then submit a traversal with a lambda - so perhaps that hasn't been a common combination.

Had to roll back declaring completely victor on TINKERPOP-1898/1895 but didn't want to kill the commit work done thus far as it still holds some value. I intend to close 1895 as duplicate of 1898 and leave 1898 open for now - that fix seems somewhat out of our hands. Will document my findings in JIRA.

VOTE +1


jorgebay:
Makes sense, VOTE +1

dkuppitz:
VOTE: +1

PR #815, opened 2018-03-11
TINKERPOP-1901 Transform enums into classes in Gremlin.Net

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-1901

The various Gremlin tokens like `T`, `Order`, and so on are now classes in Gremlin.Net so they can implement interfaces which makes them usable in more steps.

Using the example from the issue description, the Java step
```java
public default <V> GraphTraversal<S, E> by(final Traversal<?, ?> traversal,
        final Comparator<V> comparator)
```
now has this equivalent in Gremlin.Net:
```cs
public GraphTraversal<S, E> By (ITraversal traversal, IComparator comparator)
```
and `Order` implements `IComparator`.

This example already shows that the Gremlin.Net versions of those **interfaces aren't generic** which means in practice that they can still be used in some places where it shouldn't be possible. Especially `T` implements `Function<Element, Object>` in Java, but simply `IFunction` in Gremlin.Net. So users can still use `T` for steps that don't take a lambda with an `Element` argument. However, this was also possible before as we simply had the `object` type in those places before this PR. So I'd say that we tackle that one later.

All tests pass with `docker/build.sh -t -i -d`

VOTE +1


gremlin-dotnet/src/Gremlin.Net/Process/Traversal/Cardinality.cs
```diff
@@ -26,11 +26,16 @@ namespace Gremlin.Net.Process.Traversal
...
-        List,
-		Set,
-		Single
+        private Cardinality(string enumValue)
+            : base("Cardinality", enumValue)
+        {            
+        }
+
+        public static Cardinality List => new Cardinality("list");
+		public static Cardinality Set => new Cardinality("set");
```
jorgebay: `"\n\t\t"` creates this weird layout as we use 4 spaces for indentation instead of tabs.
We can use either `"\n", followed by 8 spaces` or avoid invoking `join()` and call `.each()` on the template itself.

FlorianHockmann: I decided to go with the latter so that the template can handle the formatting itself.

spmallette:
VOTE +1

PR #817, opened 2018-03-15
TINKERPOP-1920 Fixed P.within/without() handling for collections

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1920

Made the logic the same as what we have in Java currently for `P.within()` and `P.without()` for handling single collection values. I pretty much just made this work so open to input from C# experts for a nicer handling of `P`, but for now in terms of fixing the problem:

Builds with `mvn clean install -pl :gremlin-dotnet,:gremlin-dotnet-source,:gremlin-dotnet-tests -DskipIntegrationTests=false`

VOTE+1

### DIFF
--- a/gremlin-dotnet/glv/P.template
+++ b/gremlin-dotnet/glv/P.template
@@ -22,6 +22,12 @@
 #endregion
 
 // THIS IS A GENERATED FILE - DO NOT MODIFY THIS FILE DIRECTLY - see pom.xml
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
 namespace Gremlin.Net.Process.Traversal
 {
 #pragma warning disable 1591
@@ -88,9 +94,18 @@ namespace Gremlin.Net.Process.Traversal
 <% } %><% pmethods.findAll{ it in ["within", "without"] }.each { method -> %>
         public static P <%= toCSharpMethodName.call(method) %>(params object[] args)
         {
-            return new P("<%= method %>", args);
+            if (args.Length == 1 && args[0] is ICollection<object> collection)
+                return new P("<%= method %>", ToGenericArray(collection));
+            else
+                return new P("<%= method %>", args);
         }
 <% } %>
+
+        private static T[] ToGenericArray<T>(ICollection<T> collection)
+        {
+            return collection?.ToArray() ?? new T[0];;
+        }
+
         /// <inheritdoc />
         public override string ToString()
         {
--- a/gremlin-dotnet/src/Gremlin.Net/Process/Traversal/P.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Process/Traversal/P.cs
@@ -22,6 +22,12 @@
 #endregion
 
 // THIS IS A GENERATED FILE - DO NOT MODIFY THIS FILE DIRECTLY - see pom.xml
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
 namespace Gremlin.Net.Process.Traversal
 {
 #pragma warning disable 1591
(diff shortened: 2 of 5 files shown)


gremlin-dotnet/src/Gremlin.Net/Process/Traversal/P.cs
```diff
@@ -148,12 +154,29 @@ public static P Test(params object[] args)
 
         public static P Within(params object[] args)
         {
-            return new P("within", args);
+            if (args.Length == 1 && args[0].GetType().GetInterfaces().Contains(typeof(ICollection)))
```
jorgebay: Can be simplified to `args.Length == 1 && args[0] is ICollection<object>`

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/P.cs
```diff
@@ -148,12 +154,29 @@ public static P Test(params object[] args)
...
+                return new P("without", ToGenericArray((ICollection<object>) args[0]));
+            else
+                return new P("without", args);
+        }
+
+
+        private static T[] ToGenericArray<T>(ICollection<T> collection)
+        {
+            if (collection == null)
```
jorgebay: We can express the logic in this method using null-conditional and null-coalescing operators:

```csharp
private static T[] ToGenericArray<T>(ICollection<T> collection)
{
    return collection?.ToArray() ?? new T[0];
}
```

spmallette:
@jorgebay i implemented your suggestions - thanks

jorgebay:
VOTE +1

dkuppitz:
VOTE +1 (pending Florian's 2 comments)

spmallette:
@FlorianHockmann I rebased and now the tests are failing (with and without your suggested changes) - i assume this has something to do with the lambda work that just merged - any ideas?

```text
Failures:
1) g_V_hasIdXwithinXemptyXX_count: Failed
Xunit.Sdk.ContainsException: Assert.Contains() Failure
Not found: 6
In value:  Object[] [0]
   at Xunit.Assert.Contains[T](T expected, IEnumerable`1 collection, IEqualityComparer`1 comparer)
   at Gremlin.Net.IntegrationTest.Gherkin.CommonSteps.AssertResult(String characterizedAs, DataTable table) in /home/smallette/git/apache/incubator-tinkerpop/gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/CommonSteps.cs:line 188
2) g_V_notXhasIdXwithinXemptyXXX_count: Failed
Xunit.Sdk.ContainsException: Assert.Contains() Failure
Not found: 0
In value:  Object[] [6]
   at Xunit.Assert.Contains[T](T expected, IEnumerable`1 collection, IEqualityComparer`1 comparer)
   at Gremlin.Net.IntegrationTest.Gherkin.CommonSteps.AssertResult(String characterizedAs, DataTable table) in /home/smallette/git/apache/incubator-tinkerpop/gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/CommonSteps.cs:line 188
```

gremlin-dotnet/glv/P.template
```diff
@@ -94,8 +94,8 @@ namespace Gremlin.Net.Process.Traversal
 <% } %><% pmethods.findAll{ it in ["within", "without"] }.each { method -> %>
         public static P <%= toCSharpMethodName.call(method) %>(params object[] args)
         {
-            if (args.Length == 1 && args[0] is ICollection<object>)
-                return new P("<%= method %>", ToGenericArray((ICollection<object>) args[0]));
+            if (args.Length == 1 && args[0] is ICollection<object> collection)
+                return new P("without", ToGenericArray(collection));
```
FlorianHockmann: I don't think the method should be hard-coded as `without` here 😄

I just debugged one of the failing tests and was really confused when I saw `P.without()` in the Bytecode although the scenario only uses `P.within()`, but good that we now have such a good test coverage with the Gherkin features that we immediately find slips like this one 😅

jorgebay: Nice catch! I've looked at it and missed it completely :)

spmallette: hahaha....oops - copy/paste. thanks-that was dumb on my part

PR #819, opened 2018-03-17
Use error-prone compiler

leventov:
See http://errorprone.info/. Fixes a number of bugs.

This is just first step; maintainers of this project are encouraged to enable more checks eventually, e. g. "Experimental ERROR" patterns from here: http://errorprone.info/bugpatterns

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/DefaultScriptCustomizer.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/DefaultScriptCustomizer.java
@@ -25,6 +25,7 @@ import java.nio.file.Files;
 import java.util.Collection;
 import java.util.List;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * Default implementation of the {@link ScriptCustomizer} that can create the script list from a list of files or
@@ -39,7 +40,9 @@ public class DefaultScriptCustomizer implements ScriptCustomizer {
     public DefaultScriptCustomizer(final List<File> files) {
         this(files.stream().map(f -> {
             try {
-                return Files.lines(f.toPath(), StandardCharsets.UTF_8).collect(Collectors.toList());
+                try (Stream<String> lines = Files.lines(f.toPath(), StandardCharsets.UTF_8)) {
+                    return lines.collect(Collectors.toList());
+                }
             } catch (IOException ioe) {
                 throw new IllegalStateException(ioe);
             }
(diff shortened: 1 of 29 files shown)


spmallette:
Thanks for taking a look at our build. Unfortunately, this kind of change can't really be considered as part of just a pull requeswt. we'd need significant community discussion about (1) whether or not we want our build altered with more tooling at this stage of its development (2) if we did want more tooling, we'd need to discuss the right tool for the job (it could be "errorprone", it might not be - i'm sure people have lots of opinion on this topic) (3) if we did actually choose one or more, then we'd need to decide which checks we would enable. Anyway, if you'd like to lead that discussion to get a change like this in, please begin a DISCUSS thread on our dev mailing list. Thanks again.

spmallette:
Thanks again for taking a look at our build and participating in the community discussion around this change, but given the disposition of the thread, it doesn't look like we'll be taking this approach:

https://lists.apache.org/thread.html/b8abc4d43028b24e862b893078ec012803dc47db464f40ae0e7dbee8@%3Cdev.tinkerpop.apache.org%3E

leventov:
Could you reopen this PR? I'm going to leave out the build changes and keep only the code changes.

leventov:
@spmallette is it considerable?

spmallette:
sorry - i didn't see your reply earlier i the week. i think this PR even without the build changes had problems. the PR branch wasn't building I don't think (i seem to remember @robertdale bringing that up at various points). was all that fixed? 

thus far we've only talked about the build changes as I thought that was the primary point of the PR so i never bothered to comment much on the changes themselves. since you ask about that now, i'd say that they are hard to evaluate as they have been submitted. Because it is one giant PR, I don't understand the reasoning for these changes. You do have some detailed comments about the changes to `TraversalStrategy` which is nice, but that body of change is terrifying to be honest. It wouldn't surprise me if that was the source of the build problems.  That sort of change to that important area of code would require its own pull request and could require a discussion on the dev list. 

So, reopen this PR? - I'd say "no". I'd say that if you feel really strongly about these changes for some reason, then they should be broken apart into new individual PRs so that they can each be evaluated on their own merit. As a reviewer (and you need 3 committers to vote +1 for us to merge), personally I'd be much happier to see PRs that do more than add generics or reformat imports because some tool said to do it. Like, if you truly have a much better way to sort traversal strategies - wow...that would be awesome! Or if traversal strategy application was bug prone and that tool your using uncovered that and it fixed a subtle problem no one ever noticed - excellent! But, please frame your PR in that fashion so that we understand what we're exactly evaluating. Organizing your work that way will help immensely on our side.

Hope that helps frame my opinion on this. I'm sure someone else will chime in if they think something different. Thanks!

PR #821, opened 2018-03-19
TINKERPOP-1914 Support construct a GremlinServer instance from gremlin executor service

javeme:
https://issues.apache.org/jira/browse/TINKERPOP-1914

This constructor is useful when Gremlin Server is being used in an embedded
style and there is a need to share thread pools with the hosting application.


dkuppitz:
VOTE: +1

PR #823, opened 2018-03-22
TINKERPOP-1866 Support g:T for .NET

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1866

Issuing this PR for @FlorianHockmann now that TINKERPOP-1865 has merged. 

Builds with ` mvn clean install -DskipTests && mvn verify -pl :gremlin-dotnet-tests -DskipIntegrationTests=false`

VOTE +1

### DIFF
--- a/gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/CommonSteps.cs
+++ b/gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/CommonSteps.cs
@@ -240,8 +240,7 @@ namespace Gremlin.Net.IntegrationTest.Gherkin
 
         private static object ToT(string enumName, string graphName)
         {
-            return typeof(T).GetProperties()
-                .First(p => string.Equals(p.Name, enumName, StringComparison.OrdinalIgnoreCase)).GetValue(null);
+            return T.GetByValue(enumName);
         }
 
         private static object ToNumber(string stringNumber, string graphName)
--- a/gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/CommonSteps.cs
+++ b/gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/CommonSteps.cs
@@ -26,6 +26,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Reflection;
 using System.Text.RegularExpressions;
 using Gherkin.Ast;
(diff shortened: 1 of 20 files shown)


gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/CommonSteps.cs
```diff
@@ -239,7 +240,8 @@ private static object ToLambda(string stringLambda, string graphName)
 
         private static object ToT(string enumName, string graphName)
```
dkuppitz: What's the point of the `graphName` argument?

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/ScenarioData.cs
```diff
@@ -30,6 +30,7 @@
 using Gremlin.Net.Process.Remote;
 using Gremlin.Net.Process.Traversal;
 using Gremlin.Net.Structure;
+using static Gremlin.Net.Process.Traversal.__;
```
dkuppitz: Seems like an accident, as it wasn't needed before and nothing else has changed in this file.

spmallette: i saw that, but didn't kill it b/c it was Florian's changes. i guess i should have questioned it. i will try to remove that change locally and see what happens.

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/CommonSteps.cs
spmallette: It's not really useful to this conversion, but it's the expected signature for the lambda where this method is used. See further up in the definition of `Parsers` where it requires a `Func<string,string,object>` a two arg function basically. maybe we could make that nicer, but this is all plumbing code for tests, so perhaps not a big issue?

dkuppitz: No, not an issue, I was just curious.

dkuppitz:
VOTE: +1

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/ScenarioData.cs
spmallette: that commit was irrelevant - killed it out and force pushed

PR #826, opened 2018-03-26
Pinned netty to 4.0.x for spark/console

spmallette:
Spark and Gremlin Console use netty 4.0.x. They have to be the same version because spark is incompatible with 4.1.x so if the console has that version present then spark jobs won't submit and docs won't generate. Luckily, the driver is fine with 4.0.x in terms of connecting to 4.1.x from the testing i've tried.

seems really hacky - especially with gremlin server/spark deployment problems.......second thoughts.....

### DIFF
--- a/gremlin-console/pom.xml
+++ b/gremlin-console/pom.xml
@@ -50,6 +50,20 @@ limitations under the License.
             <groupId>org.apache.tinkerpop</groupId>
             <artifactId>gremlin-driver</artifactId>
             <version>${project.version}</version>
+            <!-- force the console to use netty 4.0.56.Final to comply with spark 2.2.0 which is incompatible with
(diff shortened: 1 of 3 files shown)


dkuppitz:
Now `docker/build.sh -d` failed for me on this branch in `implementations-giraph.asciidoc`.

spmallette:
i didn't have that problem - i just did a docker build with `-d`. maybe try again? 

i just started `-t -i -n` btw

dkuppitz:
Yep, the second attempt worked. The first one probably just timed out cause I had multiple containers running in parallel. Perfecto.

VOTE: +1

robertdale:
Looks like netty is upgraded in Spark 2.3.0 only. https://issues.apache.org/jira/browse/SPARK-19552

PR #828, opened 2018-03-28
updating the reference documentation

harsh9t:
added the following:
- fix the logo
- added prefixes
- added limitations
- added examples

todo:
- cover all limitations and special cases with examples


spmallette:
@harsh9t  i've merged your PR and rebased the TINKERPOP-1878 branch on tp33. please make sure that you update your fork.  i'm beginning to think we should rebase to master so that sparql-gremlin releases with 3.4.0 but that's a different discussion. 

note that i reformatted some of your docs a bit and that the docs were failing to generate in some cases due to errors in the sparql syntax. i've also noticed that a few of your examples return no results, which isn't ideal. if you could give a look to those (ones marked with `[source,groovy]` were returning errors and i guess you'd either have to generate docs or run the sparql manually to see which aren't returning results) and offer a follow-on PR to fix that up, that would be most helpful. 

Thanks for this PR btw - nice to see the documentation rounding out.

PR #830, opened 2018-04-05
TINKERPOP-1143 Moved TraversalEngine to gremlin-test

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1143

`TraversalEngine` and related infrastructure was deprecated way back on 3.2.0 but it was kept around because it was used in some infrastructure for testing. Graph providers should in no way still be using this interface and should have long ago gone to `Computer` as the `gremlin-core` infrastructure long ago quit using `TraversalEngine` in any capacity.

The only weird part was that `MatchStep` used `TraversalEngine` as a switch - I just converted that to use a boolean for checks that determine if the step is being executed `onComputer` or not.

VOTE +1


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MatchStep.java
```diff
@@ -362,7 +360,7 @@ private boolean hasPathLabel(final Path path, final Set<String> labels) {
         while (true) {
             if (this.first) {
                 this.first = false;
-                this.initializeMatchAlgorithm(TraversalEngine.Type.STANDARD);
+                this.initializeMatchAlgorithm(true);
```
dkuppitz: Shouldn't this be `false`?

spmallette: oh wow - good catch. i guess the tests pass either way though...don't know if that's bad.???

dkuppitz: It's not bad. I don't fully understand the code in `CountMatchAlgorithm` (the only place where the boolean flag makes a difference), but it seems like it only reorders the match traversals, possibly to optimize memory consumption and/or processing time.

spmallette: that makes sense - fixed and pushed

dkuppitz:
VOTE: +1

robertdale:
VOTE +1

PR #831, opened 2018-04-06
TINKERPOP-1410 Removed test artifact of hadoop-gremlin

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1410

The hadoop-gremlin test artifact allowed us to re-use some classes in spark and giraph testing, but (1) we probably shouldn't have done that in the first place as it's not a great practice and (2) we no longer have giraph and there really aren't any providers that rely on these classes that I'm aware of. I moved around a few classes to get this to all work which is described in more detail in upgrade docs. All the test still execute as they did before.

VOTE +1


dkuppitz:
VOTE: +1

robertdale:
VOTE +1

PR #832, opened 2018-04-07
TINKERPOP-1934 Bumped httpclient to 4.5.5

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1934

Removed httpcore NOTICE entry as this more recent version does not contain that designation in its NOTICE.

VOTE +1


dkuppitz:
VOTE: +1

robertdale:
VOTE +1

PR #834, opened 2018-04-09
TINKERPOP-1927 Coerced BulkSet to g:List in GraphSON 3.0

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1927

Since GLVs don't have `BulkSet` infrastructure and the need to include it doesn't seem high the most direct fix here is to coerce to `g:List` as the behavior for `BulkSet` iteration is basically that same as a List.

VOTE +1


FlorianHockmann:
Nice, no ignored scenarios anymore in Gremlin.Net!

VOTE +1

jorgebay:
Great!
VOTE +1

PR #835, opened 2018-04-09
TINKERPOP-1705 Removed rebindings API from java driver

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1705

The "rebindings" API methods were long ago deprecated in 3.1.0-incubating but were not removed in 3.2.x or 3.3.x for extended support of that feature - especially for third-party drivers. By now, most users should have moved on to use the "aliases" API instead and most of the drivers should be supporting that as well, especially since most of the drivers are now officially TinkerPop maintained.

I didn't add any upgrade docs because I started a "Deprecation Removal" section in #830 and will want to add an entry to that. I'd like to get that one merged first at which point I can rebase and add a "rebindings" entry there.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -23,11 +23,12 @@ NEED AND IMAGE
 [[release-3-4-0]]
 === TinkerPop 3.4.0 (Release Date: NOT OFFICIALLY RELEASED YET)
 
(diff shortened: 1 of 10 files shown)


dkuppitz:
VOTE: +1

robertdale:
VOTE +1

PR #836, opened 2018-04-11
TINKERPOP-1912 Remove MD5 checksums

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1912

This PR basically just adds another step to the release process, that removes the auto-generated MD5 checksums.

See: http://www.apache.org/dev/release-publishing.html#distribution_maven

VOTE: +1


spmallette:
I think this looks right. We only care about dropping md5 from the dev/release directories for official apache releases: 

VOTE +1

robertdale:
VOTE +1

PR #840, opened 2018-04-18
Gremlin.net: Added a TimeSpanConverter for GraphSON.

danielcweber:
I can add more converters for extended types if appreciated and open the appropriate ticket in JIRA.

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/TimeSpanConverter.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/TimeSpanConverter.cs
@@ -1,0 +1,45 @@
+﻿#region License
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#endregion
+
+using System;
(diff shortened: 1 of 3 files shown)


spmallette:
Hello and thanks for contributing. I guess Gremlin.NET doesn't support all of extended IO yet:

danielcweber:
Alright, I will try to follow the suggested process.

FlorianHockmann:
> Not sure if @FlorianHockmann or @jorgebay have any further comments on how to proceed with this.

Not really, I think you already mentioned everything.

> I guess you would have to write your own body of tests for these serializers since the process suite doesn't cover them.

Having unit tests that don't require a Gremlin Server instance for most functionality is a good idea in general in my opinion as it makes it easier to find bugs. That's why we also have unit tests in Gremlin.Net for the serialization of all the types that are probably also covered by some of the process suite tests.

And also thanks from me for contributing to TinkerPop!

gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/TimeSpanConverter.cs
```diff
@@ -0,0 +1,45 @@
...
+using Newtonsoft.Json.Linq;
+
+namespace Gremlin.Net.Structure.IO.GraphSON
+{
+    internal class TimeSpanConverter : IGraphSONSerializer, IGraphSONDeserializer
+    {
+        public Dictionary<string, dynamic> Dictify(dynamic objectData, GraphSONWriter writer)
+        {
+            TimeSpan value = objectData;
+            return GraphSONUtil.ToTypedValue("Duration", XmlConvert.ToString(value), "gx");
```
jorgebay: `XmlConvert.ToString()` uses ISO-8601 representation (ie `PT8H6M12.345S`)?

I see no info on the MSDN: https://msdn.microsoft.com/en-us/library/z1955e3x

We are using `PTnHnMnS` as defined by java Duration: https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#toString--

danielcweber: Seems to be implicit by how XML treats durations: https://www.w3.org/TR/xmlschema-2/#duration. The output of XmlConvert.ToString() follows the PTnHnMnS format.

jorgebay: Oh, nice TIL!

danielcweber:
-> https://github.com/apache/tinkerpop/pull/842

PR #841, opened 2018-04-18
TINKERPOP-1941 Removed deprecated structure api exception

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1941

These were related to elements not being found and were deprecated a long long time ago. Pretty easy one, but since it's a breaking change I figured it should go through review.

Builds with `mvn clean install`.

VOTE +1


jorgebay:
lgtm, VOTE +1

PR #842, opened 2018-04-19
TINKERPOP-1945

danielcweber:
https://issues.apache.org/jira/browse/TINKERPOP-1945

@FlorianHockmann @jorgebay This needs review and further discussion. Some types are left out because it is unclear how they should be represented on .net platforms.


danielcweber:
Ok, this needs a rebase I guess...

danielcweber:
Rebase done.

spmallette:
> gx:InetAddress - I'm not sure what we can expect to get in @value here as the example shows localhost

There's not much more to it than that. We use the standard Jackson serializer to handle that and it pretty much does this:

```text
gremlin> i = InetAddress.getByName("java.sun.com")
==>java.sun.com/156.151.59.19
gremlin> mapper = GraphSONMapper.build().create().createMapper()
==>org.apache.tinkerpop.shaded.jackson.databind.ObjectMapper@76b224cd
gremlin> mapper.writeValueAsString(i)
==>"java.sun.com"
```

danielcweber:
So on the .net side, that would not be more than a string?

danielcweber:
I can't really explain the build failure, the failing test runs just fine locally with an updated Newtonsoft.Json, is the package somehow force-downgraded?

danielcweber:
There we go, a force downgrade of Newtonsoft.Json:
https://travis-ci.org/apache/tinkerpop/jobs/379192651#L4033

@spmallette Could you have a look at it?

FlorianHockmann:
Probably one of the dependencies of the project has in turn a dependency on Newtonsoft.Json and wants explicitly version 9 of it and not some higher version. [The docs](https://docs.microsoft.com/en-us/nuget/consume-packages/dependency-resolution) say that when one package depends on >= 9.0.0 and one on >= 11.0.0, then 11.0.0 should win as it satisfies both requirements. This assumes that both dependencies have the same distance to the base package in the dependency graph and that they only specify the versions as minimum requirements which is the default.
However, when one dependency wants explicitly version 9.0.0 and not a higher version, then there isn't much we can do I guess.

When you find a solution to always get version 11.0.0 (e.g. without using a local `Nuget.config` file as that would require all developers to also adapt their local config), then great, but otherwise I suggest that we just not support those really big integers for now and postpone that as an additional feature for later.

danielcweber:
I might force a dependency on 11.0.0 from the unit test project itself, but I'm puzzled why it runs locally (same resolution logic should apply here), and the project explicitly depending on 9.0.0 is yet to find...

FlorianHockmann:
Yeah, I also did a quick search but only found dependencies like >= 9.0.0.

> but I'm puzzled why it runs locally 

Did you modify your Nuget.config? It's possible to specify there that Nuget should always take the highest version.

danielcweber:
>Did you modify your Nuget.config? It's possible to specify there that Nuget should always take the highest version.

No, haven't bothered with nuget.configs since .net core came around.

danielcweber:
An explicit reference from the unit test project seems to do the trick. Still unclear what depends on 9.0.0 though...

danielcweber:
@spmallette @FlorianHockmann Is this good to go now? It's definitely an improvmenet even without those whackier data types that probably aren't emitted by database providers anyway, if this got merged and (beta-)released at some point, we could resume working here...

FlorianHockmann:
Good to see that you resolved the problem with the Newtonsoft.Json version.

I think it makes sense to only support GraphSON types for which we find good equivalents in .NET and leave others out for now, like `g:MonthDay` for example. When users actually need them, then we can still think about how we can implement them in .NET.

Regarding your latest commits where you merged serializers and deserializers together: We usually have serializers and deserializers separated in different classes. The only exception I see is the `NumberConverter` with its child classes which we put together in just one base class so we don't have to create two subclasses for every numerical datatype.
I think we should stick with separated serializers and deserializers, not because I like it better that way, but because other TinkerPop GLVs also do it like that and we try to stay consistent across GLVs.

Apart from that, the PR is good to go from my side.

danielcweber:
I thought it was exactly vice-versa, there are quite a few "Converter" classes that combine serialization and deserialization... I can undo this of course.

FlorianHockmann:
Yeah, but they all extend `NumberConverter`, don't they? I remember that there was a decision for gremlin-python a while back to separate serializers and deserializers in different classes to stay consistent with gremlin-java. So, yes, it would be good if you reverted that.

gremlin-dotnet/test/Gremlin.Net.UnitTest/Structure/IO/GraphSON/GraphSONReaderTests.cs
```diff
@@ -39,7 +39,13 @@ public class GraphSONReaderTests
         private GraphSONReader CreateStandardGraphSONReader()
         {
             return new GraphSONReader();
-            }
+        }
+
+        [Fact]
+        public void Assert_loaded_Newtonsoft_Json_version()
+        {
+            Assert.Equal(new Version(11, 0, 0, 0), typeof(JToken).Assembly.GetName().Version);
```
FlorianHockmann: Sorry but I completely missed this test in my last review. I thought that you just added the test to find out what was causing the other test to fail, but I don't think that we should add a test like that as we would have to change it every time we update Newtonsoft.Json.

danielcweber: I can remove it, but since it is still unclear what forces that 9.0.0 reference, we might actually just leave it in and update it on a Newtonsoft.Json-update, it's easily done. It might be helpful in looking for the root cause for that 9.0.0 dependency and for future mishaps where the wrong version loads.

FlorianHockmann: Ok, good point. But then I would suggest that you add a comment that explains why version 11 is needed and maybe also change the test name to be more in line with the other tests. How about something like _NewtonsoftJsonVersionShouldSupportReallyBigIntegers_?

danielcweber:
Done and rebased.

FlorianHockmann:
Great and thanks for your contribution @danielcweber!

VOTE +1

jorgebay:
Sorry I'm late to the party, I haven't got much time during the past weeks: nice contribution @danielcweber !

I'm a little concerned about the difference between serializers and deserializers supported.

Can we limit the scope of this PR to types for which we support serialization and deserialization?
Some of the extended types are not very popular but once we add support for them, the user base for those could grow. From the user perspective, retrieving and storing are expected to be symmetric: if one operation is supported the other should also be.
Mapping 1 type (like `DateTime`) to multiple TinkerPop types could lead to obscure serialization errors that the user might not be able to identify.

danielcweber:
I don't know how to get total symmetry into that. As long as Tinkerpop I/O-docs don't explain the important difference between e.g. a Date and an Instant, and as long as there are vastly more time-related structures on the Java side than there are on the .net side, we can't do better than this I guess.

spmallette:
We probably should never have started support for those Java specific duration types. dah - stupid exotic types.  anyway.......

There is no precedence for these extended types in GLVs i don't think so we should probably be careful about this choice from now. If we can't get symmetry in a particular language, perhaps we should avoid support in the first place, at least for now. I say "at least for now" because it's easy to make it asymmetric later if we want, but the reverse of that is not so easy. Thoughts?

spmallette:
> Could we then somehow mark types as deprecated that we identify as problematic for non-Java languages? 

Not sure, but feel free to start fire up a DISCUSS thread on the mailing list about that if you like. Starting to talk about deprecation of types is an interesting idea to start polishing up for a core set of types for @jorgebay and https://issues.apache.org/jira/browse/TINKERPOP-1942

FlorianHockmann:
After thinking a bit more about this, my conclusion was that deprecating those types doesn't really make sense unless we plan on GraphSON 4.0 as we couldn't remove the types before that. Since we already have those types since GraphSON 2.0 and they can only be used when the extended module is explicitly selected I guess that we don't have to take immediate action at the moment.
Usage of the extended types is probably relatively low in general anyway.

But for the new serialization format @jorgebay suggested we should really check which types we want to support at all. I added a comment to the ticket so we don't forget that aspect.

So @danielcweber, looks like it would be best if you could remove all deserializers again for which no serializer exists and vice versa. Sorry for the unnecessary work you put into this.

danielcweber:
>remove all deserializers again for which no serializer exists and vice versa

Why would we do that? If we can construct (i.e. deserialize) e.g. a DateTime of whatever a server gives us, fine. Doesn't mean we have to send (i.e. serialize) obscure types. As it currently is, it's more of a 'it just works'-approach.

FlorianHockmann:
> @FlorianHockmann Why would we do that? If we can construct (i.e. deserialize) e.g. a DateTime of whatever a server gives us, fine. Doesn't mean we have to send (i.e. serialize) obscure types. As it currently is, it's more of an 'it just works'-approach.

I think the problem Jorge sees with that is the following:

1. You receive e.g. an `gx:Instant`.
2. It gets deserialized to `DateTime`.
3. You work with this `DateTime` and then send it back to the server.
4. It will be serialized to `g:Date`. <- Can't be used where the original `gx:Instant` came from.

but right, lets wait for him to come back on this.


> I think the point is to establish the intent to users that we no longer intend to support a type at some point in the future. I'm not sure what deprecation means in the IO sense.

Ok, since I'm also not sure about how we would want to handle the "deprecation" of types but since we seem to agree on the fact that communicating to users somehow which types they can expect to be supported across GLVs and which ones not, I'll start a DISCUSS thread on the mailing list. Maybe others have good ideas for that.

FlorianHockmann:
In case someone is following here but is not on the dev mailing list:
https://lists.apache.org/thread.html/32c47e8682c2848a9cc90eaff0a0da1f9bf4bd5a94b746a3e143f341@%3Cdev.tinkerpop.apache.org%3E

spmallette:
@danielcweber thanks for your patience on this pull request. Sometimes the seemingly simple tasks turn into larger more complex discussions unfortunately.  

I'm not sure if you've been following along as I don't recall seeing your name pop up in the previously referenced dev mailing list discussion where we tried to hash this all out, but it appears that the consensus is to maintain symmetry of serializers for .NET and to not allow one way serialization for the more richly/strongly typed GLVs. Hopefully you still are available to make the appropriate changes to help us get this merged.

danielcweber:
@spmallette I have been following the discussion. To be honest, I don't care too much about the symmetry. Symmetry doesn't help me much if my GLV decides to return some obscure type that I can't process because I also can't send it and it would break symmetry. I am just interested in Gremlin.net being able to send Java-Durations for .net-TimeSpans to the database.

I am available to make the appropriate changes, which would probably just be to retrire support for the non-symmetric types. I just hope in the end there will be enough support retained for us to work with CosmosDB graph (they aren't even running on Tinkerpop).

spmallette:
> Symmetry doesn't help me much if my GLV decides to return some obscure type that I can't process 

I sense the chance of that is a bit low, especially if in the future we look to discourage use of these Java specific types. Most graphs don't natively support these Java Time types, so getting one you don't expect would be odd. I think the original intent for having these at all was to better support the old script API where you might do some calculations in your Gremlin script and return one of these types, so in that case, you would be in control of whether or not you used that type in the first place.

> I am available to make the appropriate changes,... 

Thanks, that's appreciated. 

> I just hope in the end there will be enough support retained for us to work with CosmosDB graph

Please let us know how that pans out.

spmallette:
@danielcweber should we expect any change here or is this ok to close at this point?

danielcweber:
Sorry, I lost track of this issue since I worked around the TimeSpan issue with simple longs (which can be summed up nicely which is impossible for durations). I can delete the unneeded converters though, takes a minute.

danielcweber:
@spmallette Should be symmetrical now!

FlorianHockmann:
@danielcweber the build is now failing because some of the added files miss the required license header. Could you please add that?

gremlin-dotnet/test/Gremlin.Net.UnitTest/Structure/IO/GraphSON/GraphSONWriterTests.cs
```diff
@@ -346,6 +347,70 @@ public void ShouldSerializeLambda()
...
+        }
+
+        [Fact]
+        public void ShouldSerializeInt16()
+        {
+            var writer = CreateStandardGraphSONWriter();
+
+            var graphSon = writer.WriteObject((short)100);
+
+            Assert.Equal("{\"@type\":\"g:Int16\",\"@value\":100}", graphSon);
```
FlorianHockmann: `short` is also an extended type. So, the type should be `gx:Int16` here. (You did it right for the deserialization part though.)

gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/TimeSpanDeserializer.cs
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+#endregion
+
+using System.Xml;
+using Newtonsoft.Json.Linq;
+
+namespace Gremlin.Net.Structure.IO.GraphSON
+{
+    internal class TimeSpanDeserializer : IGraphSONDeserializer
```
FlorianHockmann: How about `DurationDeserializer` as a name for this class and `DurationSerializer` for the serializer?
I think that we always used the GraphSON type name for the names of the serializers and deserializers instead of the .NET type names (e.g., `UuidSerializer` instead of `GuidSerializer` or `DateSerializer` instead of `DateTimeOffsetSerializer`).

danielcweber:
Revised based on the comments, rebased onto the current tp32 branch.

FlorianHockmann:
Great, thanks for making those changes.

My VOTE +1 from nearly 3 months ago still holds with these recent changes.

gremlin-dotnet/src/Gremlin.Net/Gremlin.Net.csproj
```diff
@@ -48,7 +48,7 @@ Please see the reference documentation at Apache TinkerPop for more information
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
```
jorgebay: I don't understand the need to bump the json dependency of the actual Gremlin.Net library.

FlorianHockmann: The reason was that this version of Newtonsoft.Json couldn't deserialize _TooBigInteger_ according to @danielcweber and I didn't see a good reason why we shouldn't update the dependency.
Are you against the version bump?

FlorianHockmann: That's why I wanted to upgrade Newtonsoft.Json for the next release anyway.

spmallette:
I'm working on merging this now.....

FlorianHockmann:
I'm looking into this. At a first glance it seems to be based on the combination of the upgraded Newtonsoft.Json version and the addition of the .NET Standard 2.0 target.

spmallette:
I don't know if this matters but I noticed that the upgrade of that JSON thing didn't seem to matter to the tests. they passed on the old version (i sorta tested that version by accident because the PR updated the proj file directory and not the template so maven overwrote the change back to 9) - so do we need 9? or do we need a test case that shows we need 9?

FlorianHockmann:
I think I found the cause of the problem: We're using `dynamic` in Gremlin.Net which requires the `Microsoft.CSharp` package. We always got that package because Newtonsoft.Json also had a dependency on it. However, Newtonsoft.Json seems to not use it anymore for .NET Standard 2.0:


.NETStandard 1.3

    Microsoft.CSharp (>= 4.3.0)
    NETStandard.Library (>= 1.6.1)
    System.ComponentModel.TypeConverter (>= 4.3.0)
    System.Runtime.Serialization.Formatters (>= 4.3.0)
    System.Runtime.Serialization.Primitives (>= 4.3.0)
    System.Xml.XmlDocument (>= 4.3.0)

.NETStandard 2.0

    No dependencies.

So, we're now getting errors for every use of `dynamic` with .NET Standard 2.0 as the target framework. I'm now adding a direct dependency on Microsoft.CSharp to Gremlin.Net. So, this can't happen again, even if Newtonsoft.Json would get completely rid of that dependency.
We already had that dependency before, it was only kind of _hidden_ behind our dependency on Newtonsoft.Json.

danielcweber:
Thanks for accepting the contribution!

For what it's worth, I am developing a [type-safe ORM](https://github.com/ExRam/ExRam.Gremlinq) (abusing the 'R' in ORM a bit ;-) on top of Gremlin.net, (among others). If this is of any interest for you wrt. to the Tinkerpop repo, let me know!

spmallette:
neat - I'd be happy to suggest adding it to our provider index on the tinkerpop home page assuming it meets the listing policy:

http://tinkerpop.apache.org/policy.html

please let me know if that's something you'd like to see and if so, whether or not you think you meet the policy.

FlorianHockmann:
Very interesting project! I will definitely keep an eye on it.

I just merged the TINKERPOP-1945 feature branch together with a [CTR commit](https://github.com/apache/tinkerpop/commit/ea18963e543de90668f401938217f5aa21f2ff29) that adds the Microsoft.CSharp dependency as described above. So, we should be finally done now with this ticket 😅

danielcweber:
I guess it could qualify for the listing with some minor tweaks. It doesn't have a release yet (things get deliberately broken by me all the time, so it's at most in some version 0.x, if there was a release), and, well, there is __some__ documentation... I will update the docs to reflect what it can already do today, and probably even package a preview nuget. Thanks for the offer!

spmallette:
nice @FlorianHockmann - i will manually close this PR now

@danielcweber hope we get to see some more pull requests from you in the future. really nicely done - i bragged about your effort on twitter: https://twitter.com/spmallette/status/1029033230228447232

danielcweber:
Thanks a lot for the twitter mention! In a completely unrelated issue, is anybody of your team in contact with the graph implementors over at CosmosDB? I have no luck in getting [this issue](https://feedback.azure.com/forums/263030-azure-cosmos-db/suggestions/33998623-cosmosdb-s-implementation-of-the-tinkerpop-dsl-has) across and I wonder if there's a workaround I haven't thought of yet...

spmallette:
I'm not personally in touch with anyone over there really. I usually just ping folks over there on twitter if I need to convey/ask something - try Luis Bosquez:

https://twitter.com/_lbosq

danielcweber:
Alright, thanks!

jorgebay:
Great work @danielcweber ! Thanks for your patience during the process.

danielcweber:
@spmallette Hi. Regarding the possible listing of ExRam.Gremlinq, I am in the process of revising the Readme.md (you know, "_some_ documentary" ;-) ) and there's a preview release on nuget of [ExRam.Gremlinq](https://www.nuget.org/packages/ExRam.Gremlinq/3.1.2-preview-0002), [ExRam.Gremlinq.Providers.WebSocket](https://www.nuget.org/packages/ExRam.Gremlinq.Providers.WebSocket/3.1.2-preview-0002) and [ExRam.Gremlinq.CosmosDb](https://www.nuget.org/packages/ExRam.Gremlinq.CosmosDb/3.1.2-preview-0002).

spmallette:
I proposed the listing on the dev list...the preview release is nice, but could you reference it in your README so that it's clear. Also, please include mention of "Apache TinkerPop™" in reference to "Gremlin" - thanks.

danielcweber:
Done!

PR #843, opened 2018-04-19
TINKERPOP-1947 Fixed path history problem with mutations

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1947

When a key of a mutation was a traversal it was not being integrated to the parent traversal and tracked in the Parameters. Works now with:

```text
gremlin> g.withSideEffect('k','name').addV('person').property(select('k'),'stephen')
==>v[3]
```

VOTE +1


dkuppitz:
VOTE: +1

PR #847, opened 2018-04-20
JavaScript GLV: Document lack of GraphSON3 support

jorgebay:
Document that GraphSON3 is not supported and how to add GraphSON2 serializers to the Gremlin Server 3.3+.

No need to vote, we can just CTR.


jorgebay:
Ah, makes sense to just do a manual update for now. Remember the info in #846 for 3.2.8 and this one + #846 for 3.3.2.

I'll close these two and submit a new one with the doc updates without the bugs information.

PR #854, opened 2018-04-23
TINKERPOP-1933 gremlin-python maximum recursion depth exceeded on large responses

davebshow:
https://issues.apache.org/jira/browse/TINKERPOP-1933

Recursive calls to the `data_received` were causing max recursion depth error with large streaming responses. This PR remove the recursion in favor of a simple loop. I didn't add any new tests. Locally, I tested like this:

```python
from gremlin_python.driver import client


groovy = """
def num_nodes = 1..100000
for (n in num_nodes) {
    g.addV().next()
}
"""
c = client.Client('ws://localhost:8182/gremlin', 'g')
results = c.submit(groovy)
results.all().result()

results = c.submit("g.V().count()")
print("Added %d nodes to graph" % results.all().result()[0])

# This was always ok
print("Ok up to len: %d" % len(c.submit("g.V().limit(61888)").all().result()))

try:
    # This used to throw error
    print("Retrieved %d nodes, no more recursion issue" % len(c.submit("g.V().limit(100000)").all().result()))
except:
    pass
```

### DIFF
--- a/gremlin-python/src/main/jython/gremlin_python/driver/connection.py
+++ b/gremlin-python/src/main/jython/gremlin_python/driver/connection.py
@@ -73,6 +73,9 @@ class Connection:
         return future
 
     def _receive(self):
-        data = self._transport.read()
-        self._protocol.data_received(data, self._results)
+        while True:
+            data = self._transport.read()
+            status_code = self._protocol.data_received(data, self._results)
+            if status_code != 206:
+                break
         self._pool.put_nowait(self)
--- a/gremlin-python/src/main/jython/gremlin_python/driver/protocol.py
+++ b/gremlin-python/src/main/jython/gremlin_python/driver/protocol.py
@@ -80,21 +80,21 @@ class GremlinServerWSProtocol(AbstractBaseProtocol):
                 {'sasl': base64.b64encode(auth).decode()})
             self.write(request_id, request_message)
             data = self._transport.read()
+            # Allow recursive call for auth
             self.data_received(data, results_dict)
         elif status_code == 204:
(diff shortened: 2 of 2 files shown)


spmallette:
Builds with `mvn clean install -DskipTests && mvn verify -pl gremlin-python -DskipIntegrationTests=false`  - I don't think you need a test for this necessarily. 

VOTE +1

davebshow:
VOTE +1

dkuppitz:
VOTE: +1

But I do think a simple test would be nice. The code screams "make me recursive" and a future contributor, not aware of this issue, might try to change it back and not realize that it introduces a bug.

spmallette:
i dunno - that's sorta why i left it alone.  if you can't think of any way to test it well, a nice long explanatory comment with the JIRA referenced where you made the change might be enough to scare off another dev from coming along and trying to change that.

VivienGiraud:
Hi,
We hit the same problem, so we rose the recursion limit :  
> from sys import setrecursionlimit
> setrecursionlimit(20000)

But to be honest we are afraid it could crash as we use it in AWS Lambda (Limited environnement).
Any idea on ETA? Next release date?
Thanks everyone

spmallette:
I would have liked to have seen it for 3.2.9/3.3.3 but it didn't merge in time (in the process of releasing those now). We don't have a scheduled release date yet for 3.2.10/3.3.4.

VivienGiraud:
OK, thanks @spmallette for the quick answer. We will try @davebshow patch on our side ;)
BTW thank you @davebshow  for this useful patch!

PR #855, opened 2018-04-24
TINKERPOP-1950 Cached global strategy lookups during traversal construction

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1950

This change leads to a 1.5x to 2x speed improvement in traversal construction. It is especially effective when processing traversals that have many child traversals within them as this method is called for not only the parent traversal but all the children as well.

VOTE +1

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalStrategies.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalStrategies.java
@@ -63,6 +63,8 @@ import java.util.stream.Collectors;
  */
 public interface TraversalStrategies extends Serializable, Cloneable {
 
+    static Set<Class> LOADED = new HashSet<>();
+
     static List<Class<? extends TraversalStrategy>> STRATEGY_CATEGORIES = Collections.unmodifiableList(Arrays.asList(TraversalStrategy.DecorationStrategy.class, TraversalStrategy.OptimizationStrategy.class, TraversalStrategy.ProviderOptimizationStrategy.class, TraversalStrategy.FinalizationStrategy.class, TraversalStrategy.VerificationStrategy.class));
 
     /**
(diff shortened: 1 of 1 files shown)


dkuppitz:
VOTE: +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalStrategies.java
```diff
@@ -63,6 +63,8 @@
  */
 public interface TraversalStrategies extends Serializable, Cloneable {
 
+    static Set<Class> LOADED = new HashSet<>();
```
jorgebay: Shouldn't we use a collection that supports concurrent modification? Different threads might be adding items to the collection.

Also, we could move the declaration to `GlobalCache`?.

spmallette: ha - i was so excited that this changed worked i didn't bother to think too hard about what i was doing. good suggestions in both cases - i've made those changes now. thanks

jorgebay:
VOTE +1

spmallette:
That method isn't available on `ConcurrentHashMap.newKeySet()` - is there some other way to get to that?

robertdale:
Well, you would do use ConcurrentHashMap ;-)
```
static Map<Class, Boolean> LOADED = new ConcurrentHashMap<>();
        try {
            LOADED.computeIfAbsent(graphOrGraphComputerClass, (unused) -> {

                final String graphComputerClassName = null != graphOrGraphComputerClass.getDeclaringClass()
                        ? graphOrGraphComputerClass.getCanonicalName().replace(
                                "." + graphOrGraphComputerClass.getSimpleName(),
                                "$" + graphOrGraphComputerClass.getSimpleName())
                        : graphOrGraphComputerClass.getCanonicalName();
                try {
                    Class.forName(graphComputerClassName);
                } catch (ClassNotFoundException e) {
                    throw new RuntimeException(e);
                }

                // keep track of stuff we already loaded once - stuff in this if/statement isn't
                // cheap and this
                // method gets called a lot, basically every time a new traversal gets spun up
                // (that includes
                // child traversals. perhaps it is possible to just check the cache keys for
                // this information, but
                // it's not clear if this method will be called with something not in the cache
                // and if it is and
                // it results in error, then we'd probably not want to deal with this block
                // again anyway
                return Boolean.TRUE;
            });
        } catch (Exception e) {
            throw new IllegalStateException(e.getCause().getMessage(), e.getCause());
        }

```

spmallette:
ah - ok, just use the Map rather than the Set....fair enough

spmallette:
done - https://github.com/apache/tinkerpop/commit/1d9e6dc6d30c5c7d56e4007527365793eb1f223e

PR #857, opened 2018-04-25
TINKERPOP-1869 Allowed iterate() after profile()

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1869

The `iterate()` method adds a `NoneStep` that was preventing traversals like `g.V().profile().iterate()`. Not the typical type of traversal you would do, but it seems like it should be allowed for consistency sake.

VOTE +1

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/verification/StandardVerificationStrategy.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/verification/StandardVerificationStrategy.java
@@ -24,6 +24,7 @@ import org.apache.tinkerpop.gremlin.process.traversal.Step;
 import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
 import org.apache.tinkerpop.gremlin.process.traversal.TraversalStrategy;
 import org.apache.tinkerpop.gremlin.process.traversal.step.branch.RepeatStep;
+import org.apache.tinkerpop.gremlin.process.traversal.step.filter.NoneStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.sideEffect.ProfileSideEffectStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.sideEffect.SideEffectCapStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.util.ReducingBarrierStep;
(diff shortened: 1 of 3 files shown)


robertdale:
VOTE +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/verification/StandardVerificationStrategy.java
```diff
@@ -63,15 +64,17 @@ public void apply(final Traversal.Admin<?, ?> traversal) {
...
+        // (2) 2nd last step when accompanied by the cap step or none step (i.e. iterate() to nothing)
+        // (3) 3rd to last when the traversal ends with a RequirementsStep.
         final Step<?, ?> endStep = traversal.asAdmin().getEndStep();
         if (TraversalHelper.hasStepOfClass(ProfileSideEffectStep.class, traversal) &&
                 !(endStep instanceof ProfileSideEffectStep ||
                         (endStep instanceof SideEffectCapStep && endStep.getPreviousStep() instanceof ProfileSideEffectStep) ||
-                        (endStep instanceof RequirementsStep && (
-                                endStep.getPreviousStep() instanceof SideEffectCapStep ||
-                                        endStep.getPreviousStep() instanceof ProfileSideEffectStep)))) {
+                        (endStep instanceof NoneStep && endStep.getPreviousStep() instanceof SideEffectCapStep && endStep.getPreviousStep().getPreviousStep() instanceof ProfileSideEffectStep) ||
```
jorgebay: I think we can save some `instanceof` checks if we introduce a method similar to `TraversalHelper.hasStepOfClass(ProfileSideEffectStep.class, traversal)` but instead of returning a boolean, it returns the step instance. Then, from the `ProfileSideEffectStep` instance itself, it's easier to do the checks.

spmallette: yeah - @dkuppitz is thinking about this kind of stuff. i had ideas too, but figured it was something for a different issue. Generally speaking, I think that if we want to do something about the complexity of strategy writing, we really need to carefully analyze the strategies that are out there and find good general functions to add that can be best used across all of them rather than looking at just a bad problem in one and build a function that only fits that case.

PR #860, opened 2018-04-27
TINKERPOP-1595 Optimize TraversalVertexProgram

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1595

I did as the issue requested and optimized `TraversalVertexProgram`. Definitely some hotspots in there that have been removed. I'd say that the main fixes involved the direct lookup of `clone()` in `MemoryComputeKey` rather than cycling all methods and the use of Base64 encoding instead of CSV style string/regex derser for traversals in OLAP.

Feel free to VOTE but this change will not be merged until we open development for 3.2.10.

VOTE +1


dkuppitz:
VOTE: +1

robertdale:
VOTE +1

PR #862, opened 2018-05-09
TINKERPOP-1961 Removed duplication of images in binaries.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1961

The asciidoc plugin copies all files in the source directories of its execution. The asciidoc plugin was upgraded between 3.2.5 and 3.2.6 which introduced this new behavior.

VOTE +1

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -811,9 +811,14 @@ limitations under the License.
                                     <goal>process-asciidoc</goal>
                                 </goals>
                                 <configuration>
-                                    <sourceDirectory>${asciidoc.input.dir}/</sourceDirectory>
+                                    <resources>
+                                        <resource>
+                                            <directory>${asciidoc.input.dir}/</directory>
+                                            <targetPath>${htmlsingle.output.dir}/</targetPath>
+                                            <excludes>${asciidoc.input.dir}/static/*.*</excludes>
+                                        </resource>
+                                    </resources>
                                     <sourceDocumentName>index.asciidoc</sourceDocumentName>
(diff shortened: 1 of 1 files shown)


robertdale:
After running `bin/process-docs.sh` the resulting console/server zip files no longer contain `docs/` let alone `images`. Is there a different step that needs to be run?

spmallette:
Confirmed that once you do `bin/process-docs.sh` it generates just the `/images` directory to the right place without `/static/images`. To get that into the zip distributions you need to go an extra step and do `mvn install`  (during release this occurs during `mvn deploy`).  The zip distributions are much smaller now.

robertdale:
I did:
1. `bin/process-docs.sh`
2. `mvn install`
3. `find . -name '*.zip'`
4. `unzip -t ./gremlin-server/target/apache-tinkerpop-gremlin-server-3.2.9-distribution.zip | grep images`
I don't see any images
3. `mvn deploy`
I can see the maven-assembly-plugin rebuild the zip file. Still don't see any images in any zips

spmallette:
I dunno man - here's a fast way to test which I just went through:

1. `bin/process-docs.sh --dryRun` - check `target/docs/htmlsingle` - should just have `images/` directory there.
1. `mvn package -pl gremlin-server -DskipTests`
1. look at `gremlin-server/target/apache-tinkerpop-gremlin-server-x.y.z-standalone` as that is used to build the zip - should just have `docs/images/` 
1. the zip is in there too to check and looks ok to me

robertdale:
I walked through the actual release process and this time I do see the correct packaging.
VOTE +1

PR #865, opened 2018-05-13
TINKERPOP-1963 Fixed branch() problems with reducing steps as options

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1963

Not sure I fixed this in the best way possible, but I added an `IdentityStep` to the front of each option:

https://github.com/apache/tinkerpop/compare/tp32...TINKERPOP-1963?expand=1#diff-546a91c51e7b731ccea30f0e4aff5d83R68

and then later checked that step to see if anything had been injected into it to determine if the choice had evaluated to `true` or not (since that was the existing way in which that was done).  I'm open to better approaches to resolving this problem if they can be suggested. Also ready to hear that "this way won't work because..." In the mean time.....

VOTE +1


dkuppitz:
VOTE: +1

davebshow:
VOTE +1

PR #871, opened 2018-06-04
TINKERPOP-1975 Introduced with() step and Configuring interface

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1975

This PR is about `with()` step which introduces a way to provide a way to supply configurations to steps. Especially useful for algorithm type steps like `pageRank()` which had previously relied on `by()` which was insufficient for this purposes and a bit forced to try to prevent additional expansion of the core API.

As a result of `with()` we now have the `Configuring` interface to apply to steps that can accept `with()` options. Moved `Parameterizing` to the package containing the other step interfaces. Changed interface hierarchy with `Parameterizing` and `Mutating` now that `Configuring` is present. Both of those changes make this body of work a breaking change to existing APIs though it should really only affect graph providers and even then the fix is quite simple as described in the upgrade documentation.

VOTE +1


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/CoreImports.java
```diff
@@ -260,6 +261,7 @@
         CLASS_IMPORTS.add(PeerPressureVertexProgram.class);
         CLASS_IMPORTS.add(PageRankMapReduce.class);
         CLASS_IMPORTS.add(PageRankVertexProgram.class);
+        CLASS_IMPORTS.add(PageRankVertexProgramStep.PageRank.class);
```
dkuppitz: I think `PeerPressureVertexProgramStep.PeerPressure` should be a core import too.

dkuppitz:
VOTE: +1

robertdale:
VOTE +1

PR #873, opened 2018-06-08
TINKERPOP-1518 Cache Graph.Features for tests

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1518

For graphs that have an expensive startup time but a static set of `Graph.Features` the new `GraphProvider.getStaticFeatures()` method should help speed their test times as it will prevent creation of a new `Graph` instance for the test if the graph does not support the requisite features. For testing purposes, this capability was implemented for Neo4j, but it didn't do much to speed the execution time of the test suite - saved a couple minutes or so. It's generally pretty cheap to create Neo4j instances themselves, so I guess that is expected.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -33,6 +33,7 @@ This release also includes changes from <<release-3-3-3, 3.3.3>>.
 * `min()` and `max()` now support all types implementing `Comparable`.
 * Change the `toString()` of `Path` to be standardized as other graph elements are.
 * `hadoop-gremlin` no longer generates a test artifact.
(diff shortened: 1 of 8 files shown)


dkuppitz:
VOTE: +1

Though I would change the Jira ticket title before it makes it into the CHANGELOG, as the implementation doesn't really cache anything.

spmallette:
Good point - updated to "Provide a way for providers to expose static Graph.Features to tests"

jorgebay:
Nice doc on the new method, VOTE +1

PR #875, opened 2018-06-10
    TINKERPOP-1967  Add a connectedComponent() step - vtslab contribution

vtslab:
This merges the new OLAP step into a corrected version of the old recipe. I did not adapt the release-update file, which is no longer consistent with the sutiation (leave that to you). I made a comment in the TINKERPOP-1967 branch proper about the gremlin-console import section.


spmallette:
@vtslab is there something amiss with this PR? any idea why are there so many conflicted files?

vtslab:
PR is malformed due to problems in master. I close this one and prepare a new PR with the same commits.

PR #878, opened 2018-06-15
Updated transpilers.asciidoc

harsh9t:
Changes:
- fixed non-working queries (queries which returned no results)
- updated documentation with more examples in the limitations section
- updated execution tags for queries which are for reference only
- commented out the pattern filter examples for now

To Do:
- fix OPTIONAL queries 
- fix Pattern filter queries
- inspect and patch special cases in code


spmallette:
merged - also note that i've now rebased TINKERPOP-1878 on master

PR #881, opened 2018-06-21
TINKERPOP-1365 Refactored use of Random in tests

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1365

Each test suite now uses the same instance of Random which prints the seed given to it so that if there is a failure, we can easily try to recreate it by taking the seed and passing it in as a system property with -DtestSeed. The output logs as:

```text
[INFO] org.apache.tinkerpop.gremlin.TestHelper - *** THE RANDOM TEST SEED IS 1529585621517 ***
```

I didn't replace all of the old use of `Random` - there were a couple places where I felt like a true `Random` was still necessary and just left it alone despite the horror it might cause. Going forward we'll want to use the `TestHelper.RANDOM` whenever possible (or, better, just not use a `Random` at all).

VOTE +1

### DIFF
--- a/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/TestHelper.java
+++ b/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/TestHelper.java
@@ -18,7 +18,6 @@
  */
 package org.apache.tinkerpop.gremlin;
 
-import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
 import org.apache.tinkerpop.gremlin.structure.Direction;
 import org.apache.tinkerpop.gremlin.structure.Edge;
 import org.apache.tinkerpop.gremlin.structure.Graph;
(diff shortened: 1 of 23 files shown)


gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/TestHelper.java
```diff
@@ -51,11 +52,19 @@
...
 
+    public static final Random RANDOM;
     private static final String SEP = File.separator;
     private static final char URL_SEP = '/';
     public static final String TEST_DATA_RELATIVE_DIR = "test-case-data";
 
+    static {
+        final long seed = Long.parseLong(System.getProperty("testSeed", String.valueOf(System.currentTimeMillis())));
+        logger.info("*** THE RANDOM TEST SEED IS {} ***", seed);
+        RANDOM = new Random();
```
dkuppitz: This should be `RANDOM = new Random(seed);`.

spmallette: wow - the whole point of the PR and i didn't even do it right :disappointed:  :rage4:

PR #882, opened 2018-06-21
TINKERPOP-1990 Add a shortestPath() step

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1990

Implemented `ShortestPathVertexProgram` and `ShortestPathVertexProgramStep`.

`docker/build.sh -t -i -n -d` passed.

VOTE: +1

I can wait for TINKERPOP-1991 before I merge it or merge it into `master/` now and then go from there with TINKERPOP-1991, I don't care.

### DIFF
--- a/docs/src/recipes/shortest-path.asciidoc
+++ b/docs/src/recipes/shortest-path.asciidoc
@@ -49,6 +49,16 @@ length three), but this example is not considering that.
 <2> It might be interesting to know the path lengths for all paths between vertex "1" and "5".
 <3> Alternatively, one might wish to do a path length distribution over all the paths.
 
+The following code block demonstrates how the shortest path from `v[1]` to `v[5]` can be queried in OLAP, using the `shortestPath()` step.
+
+[gremlin-groovy,existing]
+----
+g = g.withComputer()
+g.V(1).shortestPath().
+  with(ShortestPath.edges, Direction.OUT).
+  with(ShortestPath.target, hasId(5))
+----
+
 The previous example defines the length of the path by the number of vertices in the path, but the "path" might also
 be measured by data within the graph itself. The following example use the same graph structure as the previous example,
 but includes a "weight" on the edges, that will be used to help determine the "cost" of a particular path:
@@ -95,6 +105,17 @@ calculated cost. With some slight modifications given the use of `select` it bec
 the output. Note that the path with the lowest "cost" actually has a longer path length as determined by the graph
 structure.
 
+The next code block demonstrates how the `shortestPath()` step can be used in OLAP to determine the shortest weighted path.
+
+[gremlin-groovy,existing]
+----
+g = g.withComputer()
+g.V(1).shortestPath().
+  with(ShortestPath.edges, Direction.OUT).
+  with(ShortestPath.distance, 'weight').
+  with(ShortestPath.target, hasId(5))
+----
+
 The following query illustrates how `select(<traversal>)` can be used to find all shortest weighted undirected paths
 in the modern toy graph.
 
@@ -136,3 +157,44 @@ g.withSack(0.0).V().as("from").       <1>
 <7> Order the output by the start vertex id and then the end vertex id (for better readability).
 <8> Deduplicate vertex pairs (the shortest path from `v[1]` to `v[6]` is the same as the path from `v[6]` to `v[1]`).
 
+Again, this can be translated into an OLAP query using the `shortestPath()` step.
+
+[gremlin-groovy,existing]
+----
+result = g.withComputer().V().
+  shortestPath().
+    with(ShortestPath.distance, 'weight').
+    with(ShortestPath.includeEdges, true).
+  filter(count(local).is(gt(1))).
+  group().
+    by(project('from','to').
+         by(limit(local, 1)).
+         by(tail(local, 1))).
+  unfold().
+  order().
+    by(select(keys).select('from').id()).
+    by(select(keys).select('to').id()).toList()
+----
+
+The obvious difference in the result is the absence of property values in the OLAP result. Since OLAP traversers are not
+allowed to leave the local star graph, it's not possible to have the exact same result in an OLAP query. However, the determined
+shortest paths can be passed back into the OLTP `GraphTraversalSource`, which can then be used to query the values.
+
+[gremlin-groovy,existing]
+----
+g.withSideEffect('v', []).                            <1>
+  inject(result.toArray()).as('kv').select(values).
+  unfold().
+  map(unfold().as('v_or_e').
+      coalesce(V().where(eq('v_or_e')).store('v'),
+               select('v').tail(local, 1).bothE().where(eq('v_or_e'))).
+      values('name','weight').
+      fold()).
+  group().
+    by(select('kv').select(keys)).unfold().
+  order().
+    by(select(keys).select('from').id()).
+    by(select(keys).select('to').id()).toList()
+----
+
+<1> The side-effect `v` is used to keep track of the last processed vertex, hence it needs to be an order-preserving list. Without this explicit definition `v` would become a `BulkSet` which doesn't preserve the insert order.
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -2698,6 +2698,67 @@ link:++http://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/grem
 link:++http://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gremlin/structure/Column.html++[`Column`],
 link:++http://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gremlin/process/traversal/Pop.html++[`Pop`]
 
+[[shortestpath-step]]
+=== ShortestPath step
+
+The `shortestPath()`-step provides an easy way to find shortest non-cyclic paths in a graph. It is configurable
+using the `with()`-modulator with the options given below.
+
+[width="100%",cols="3,3,15,5",options="header"]
+|=========================================================
+| Key | Type | Description | Default
+| `target` | `Traversal` | Sets a filter traversal for the end vertices (e.g. `__.has('name','marko')`). | all vertices (`__.identity()`)
+| `edges` | `Traversal` or `Direction` | Sets a `Traversal` that emits the edges to traverse from the current vertex or the `Direction` to traverse during the shortest path discovery. | `Direction.BOTH`
+| `distance` | `Traversal` or `String` | Sets the `Traversal` that calculates the distance for the current edge or the name of an edge property to use for the distance calculations. | `__.constant(1)`
+| `maxDistance` | `Number` | Sets the distance limit for all shortest paths. | none
+| `includeEdges` | `Boolean` | Whether to include edges in the result or not. | `false`
+|=========================================================
+
+[gremlin-groovy,modern]
+----
+g = g.withComputer()
+g.V().shortestPath() <1>
+g.V().has('person','name','marko').shortestPath() <2>
+g.V().shortestPath().with(ShortestPath.target, __.has('name','peter')) <3>
+g.V().shortestPath().
+        with(ShortestPath.edges, Direction.IN).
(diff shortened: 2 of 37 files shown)


spmallette:
I will try to find some time to do TINKERPOP-1991 and see what is involved. If it's a mess then maybe we don't try to shove  TINKERPOP-1991 into 3.4.0 and await 3.5.0 then just merge `shortestPath()` and `connectedComponent()` for 3.4.0. I'd say that if we get stuck without TINKERPOP-1991 we probably shouldn't add anymore algorithms until 3.5.0 so as to not further pollute the core Gremlin space.  thinky think....

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/search/path/ShortestPathVertexProgram.java
```diff
@@ -0,0 +1,557 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.tinkerpop.gremlin.process.computer.search.path;
+
+import org.apache.commons.configuration.Configuration;
+import org.apache.tinkerpop.gremlin.process.computer.*;
```
twilmes: Looks like the IDE wild-carded a few of these imports

spmallette:
I just issued #897 - note my comments there about TINKERPOP-1991. I think we should just leave that alone for 3.4.0 and come back to it. I also think we should disregard my comments about pushing these off to 3.5.0...we will save a lot of headache with these steps so i think there are better here now than later. 

@dkuppitz could you please rebase. i can review this next week in more detail.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/traversal/step/map/ShortestPathVertexProgramStep.java
```diff
@@ -0,0 +1,174 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.tinkerpop.gremlin.process.computer.traversal.step.map;
+
+import org.apache.tinkerpop.gremlin.process.computer.*;
```
spmallette: nit: You have some wildcard imports here....

docs/src/reference/the-traversal.asciidoc
```diff
@@ -2489,6 +2489,62 @@ link:++http://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/grem
...
+| `target` | `Traversal` | Sets a filter traversal for the end vertices (e.g. `__.has('name','marko')`). | all vertices (`__.identity()`)
+| `edges` | `Traversal` or `Direction` | Sets a `Traversal` that emits the edges to traverse from the current vertex or the `Direction` to traverse during the shortest path discovery. | `Direction.BOTH`
+| `distance` | `Traversal` or `String` | Sets the `Traversal` that calculates the distance for the current edge or the name of an edge property to use for the distance calculations. | `__.constant(1)`
+| `maxDistance` | `Number` | Sets the distance limit for all shortest paths. | none
+| `includeEdges` | `Boolean` | Whether to include edges in the result or not. | `false`
+|=========================================================
+
+[gremlin-groovy,modern]
+----
+a = g.withComputer()
```
spmallette: I don't know that we should start doing "a" here...we don't reference the `GraphTraversalSource` any other way than "g" anywhere else in the docs.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/search/path/ShortestPathVertexProgram.java
```diff
@@ -0,0 +1,581 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+
+/**
+ * @author Daniel Kuppitz (http://gremlin.guru)
```
spmallette: It would be nice to see more javadoc and inline comments in here. I'd like to know more about "why" in a lot of the stuff i see here, especially in `execute()`, `storeState()` and `loadState()`. Similar to the inline comments i have in `connectedComponent()` - https://github.com/apache/tinkerpop/pull/897/files#diff-b0b5f3644f5f656c492f3f6ddae2c7feR117 it would be also nice to know why halted traversers are setup the way that they are if you figured that out somehow because it's still mysterious to me.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/remote/RemoteGraph.java
```diff
@@ -60,6 +60,10 @@
         test = "org.apache.tinkerpop.gremlin.process.traversal.step.map.PageRankTest",
         method = "*",
         reason = "hmmmm")
+@Graph.OptOut(
+        test = "org.apache.tinkerpop.gremlin.process.traversal.step.map.ShortestPathTest",
+        method = "*",
+        reason = "hmmmm")
```
spmallette: I'm going to update the "hmmm" to https://issues.apache.org/jira/browse/TINKERPOP-1976 - please do the same in your PR. :smile:

docs/src/reference/the-traversal.asciidoc
dkuppitz: I know, but we also never mixed two traversal sources in a single code block. Better suggestions?

spmallette: oh - sorry - i didn't see that. maybe just don't put it all in one block? reserve that last one as its own example with some introductory text to separate and explain what its doing?

FlorianHockmann:
Just wondered: Does the _Shortest Path_ recipe still make sense with the new `shortestPath()` step? And should the section in the reference docs about shortest paths and the recipe link to each other?

spmallette:
@dkuppitz could you rebase this one more time please?

dkuppitz:
Done.

docs/src/recipes/shortest-path.asciidoc
```diff
@@ -48,6 +48,17 @@ course, it is possible for there to be more than one path in the graph of the sa
 length three), but this example is not considering that.
 <2> It might be interesting to know the path lengths for all paths between vertex "1" and "5".
 <3> Alternatively, one might wish to do a path length distribution over all the paths.
+<4> The preferred way to get a shortest path in OLAP is the `shortestPath()` step.
```
spmallette: hmm - `<4>` doesn't refer to anything in the code above. did you mean to add that callout?

docs/src/recipes/shortest-path.asciidoc
```diff
@@ -136,3 +158,44 @@ g.withSack(0.0).V().as("from").       <1>
...
+      values('name','weight').
+      fold()).
+  group().
+    by(select('kv').select(keys)).unfold().
+  order().
+    by(select(keys).select('from').id()).
+    by(select(keys).select('to').id()).toList()
+----
+
+<1> The side-effect `v` is used to keep track of the last processed vertex, hence it needs to be an order-preserving list. Without this explicit definition `v` would become a `BulkSet` which doesn't preserve the insert order.
```
spmallette: interesting

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Traversal.java
```diff
@@ -496,15 +496,17 @@ public default void reset() {
         public void setGraph(final Graph graph);
 
         public default boolean equals(final Traversal.Admin<S, E> other) {
-            final List<Step> steps = this.getSteps();
-            final List<Step> otherSteps = other.getSteps();
-            if (steps.size() == otherSteps.size()) {
-                for (int i = 0; i < steps.size(); i++) {
-                    if (!steps.get(i).equals(otherSteps.get(i))) {
-                        return false;
+            if (this.getClass().equals(other.getClass())) {
```
spmallette: I guess this changes makes sense. Are you sure there was no reason why we didn't have it that way to begin with? Just playing devil's advocate, but should `Traversal` equality in some way be based on the steps the traversal contains rather than the class it has?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -2451,6 +2452,24 @@ else if (value instanceof Traversal)
...
     }
 
+    /**
+     * Executes a Shortest Path algorithm over the graph.
+     *
+     * @return the traversal with the appended {@link ShortestPathVertexProgramStep}
+     * @see <a href="http://tinkerpop.apache.org/docs/${project.version}/reference/#shortestpath-step" target="_blank">Reference Documentation - ShortestPath Step</a>
+     */
+    public default GraphTraversal<S, Path> shortestPath() {
+        if (this.asAdmin().getEndStep() instanceof GraphStep) {
```
spmallette: So, what does this mean exactly? Like, if you did `g.V().has('name','marko')` it wouldn't filter "marko" vertex without you throwing an `identity()` in there?

docs/src/recipes/shortest-path.asciidoc
dkuppitz: Ah, forgot to remove it, that's from when OLTP and OLAP were still mixed in one block.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Traversal.java
dkuppitz: I don't think so. The main reason why I had to change it was that traversals without steps were considered equal using the old method. That means all `TrueTraversal`, `ConstantTraversal`, `ValueTraversal`, etc. instances were equal compared to each other.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
dkuppitz: Oh, totally forgot about this corner case (that's something you should test in `ConnectedComponentVP` too). And right, if you only have `g.V().has(...)` followed by a VP step, your VP won't see any halted traversers and thus behave like it was just `g.V()`. I spent quite some time trying to fix that properly, but couldn't come up with anything.

docs/src/recipes/shortest-path.asciidoc
dkuppitz: I was surprised that the elements didn't get dereferenced automatically (`result` contains `ReferencePath` objects holding `ReferenceVertex` and `ReferenceEdge` instances). If they were automatically reattached, the `map` step would be as simple as this:

```
  map(unfold().values('name','weight').fold())
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
spmallette: `connectedComponent()` seems to work without adding that `identity()` step:

```text
gremlin> graph = TinkerFactory.createModern()
==>tinkergraph[vertices:6 edges:6]
gremlin> g = graph.traversal().withComputer()
==>graphtraversalsource[tinkergraph[vertices:6 edges:6], graphcomputer]
gremlin> g.V().has('name',within('marko','vadas','lop','josh')).connectedComponent()
==>v[1]
==>v[3]
==>v[2]
==>v[4]
```

Not sure what's different....but now i'm questioning what i have in `connectedComponent()` again on something related.

spmallette:
Let's get #909 merged, rebase this, add GLV tests and then merge this one (doing the same for #897 )

PR #885, opened 2018-06-23
Add check for connection state before returning it TINKERPOP-1978

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-1978

This PR simply adds the requested check of whether a connection is actually open that was returned from the pool before returning it to the driver. Before it was possible that a connection would be used that was already closed while it was in the connection pool.

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
@@ -23,7 +23,6 @@
 
 using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Gremlin.Net.Process;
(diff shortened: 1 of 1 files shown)


gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
```diff
@@ -45,19 +44,28 @@ public ConnectionPool(ConnectionFactory connectionFactory)
...
+            {
+                connection = null;
+                lock (_connectionsLock)
+                {
+                    if (_connections.IsEmpty) return false;
+                    _connections.TryTake(out connection);
+                }
+
+                if (connection.IsOpen) return true;
+                connection.Dispose();
```
dkuppitz: Won't that lead to a never-ending loop? I'm assuming that `connection.Dispose()` won't remove it from the `_connections` collection, but I might be wrong.

FlorianHockmann: `_connections.TryTake(out connection)` removes the connection from the collection. So, here it is already no longer part of `_connections`. That may be a bit confusing as `_connections` doesn't contain all connections in the pool but only those that are currently unused. That made this class a lot easier to implement as no checks are needed of whether a connection is already in use or not.
For reference: The [`ConcurrentBag<T>.TryTake`](https://msdn.microsoft.com/en-us/library/dd381774(v=vs.110).aspx) docs.

dkuppitz:
VOTE: +1

jorgebay:
VOTE +1

PR #886, opened 2018-06-29
TINKERPOP-1987 Bump to Spark 2.3.1 and Netty 4.1.25.Final

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1987
https://issues.apache.org/jira/browse/TINKERPOP-1993

These really needed to be bumped together because going to netty 4.1.x doesn't work nicely with Spark 2.2.x - Spark only upgraded to that version on 2.3.x. 

That odd change to `gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs` is incidental - that modification is already on master and rebase doesn't seem to to want to get rid of it. Hopefully no one is too turned off by that little stain in change history - doesn't seem like a huge problem.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,6 +25,8 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 This release also includes changes from <<release-3-3-3, 3.3.3>>.
 
+* Bumped to Netty 4.1.25.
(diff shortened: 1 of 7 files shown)


dkuppitz:
VOTE: +1

robertdale:
@spmallette any idea why the spark test is hanging in travis?

PR #887, opened 2018-07-02
fix(DriverRemoteConnection): return the promise on close

elliotttf:
The jsdoc indicates that the `DriverRemoteConnection.close()` method returns a `Promise`; however, the first time the method is called it returns `undefined`.


jorgebay:
Nice catch! looks good: VOTE: +1

jorgebay:
I've filed a ticket to track this bug fix: https://issues.apache.org/jira/browse/TINKERPOP-1995

When creating the ticket, I realized we should we fix this issue in all the active branches... @elliotttf  do you mind basing your patch on `tp32` branch?

elliotttf:
@jorgebay don't mind at all. Would you like me to keep it open against master, or reopen this against tp32?

jorgebay:
Thanks!

Please reopen this one against `tp32`, we will merge `tp32` to `tp33` to `master` afterwards.

PR #891, opened 2018-07-15
TINKERPOP-2006 - Fix for valid GraphML export when graph properties o…

svanteschubert:
…f a vertex and edge have similar names

I have created an JIRA issue for this patch:
https://issues.apache.org/jira/browse/TINKERPOP-2006

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphml/GraphMLWriter.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphml/GraphMLWriter.java
@@ -18,6 +18,21 @@
  */
 package org.apache.tinkerpop.gremlin.structure.io.graphml;
 
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import javax.xml.XMLConstants;
+import javax.xml.stream.XMLOutputFactory;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamWriter;
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.tinkerpop.gremlin.structure.Direction;
 import org.apache.tinkerpop.gremlin.structure.Edge;
 import org.apache.tinkerpop.gremlin.structure.Element;
@@ -29,21 +44,6 @@ import org.apache.tinkerpop.gremlin.structure.io.GraphWriter;
 import org.apache.tinkerpop.gremlin.structure.io.Io;
 import org.apache.tinkerpop.gremlin.structure.util.Comparators;
 import org.apache.tinkerpop.gremlin.util.iterator.IteratorUtils;
-
-import javax.xml.XMLConstants;
-import javax.xml.stream.XMLOutputFactory;
-import javax.xml.stream.XMLStreamException;
-import javax.xml.stream.XMLStreamWriter;
-import java.io.IOException;
-import java.io.OutputStream;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
 
 /**
  * GraphMLWriter writes a Graph to a GraphML OutputStream. Note that this format is lossy, in the sense that data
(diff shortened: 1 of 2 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphml/GraphMLWriter.java
```diff
@@ -218,20 +220,33 @@ private void writeTypes(final Map<String, String> identifiedVertexKeyTypes,
                             final Map<String, String> identifiedEdgeKeyTypes,
                             final XMLStreamWriter writer) throws XMLStreamException {
         // <key id="weight" for="edge" attr.name="weight" attr.type="float"/>
-        final Collection<String> vertexKeySet = getVertexKeysAndNormalizeIfRequired(identifiedVertexKeyTypes);
+        Collection<String> vertexKeySet = getVertexKeysAndNormalizeIfRequired(identifiedVertexKeyTypes);
```
twilmes: Could these still be made `final`?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphml/GraphMLWriter.java
```diff
@@ -218,20 +220,33 @@ private void writeTypes(final Map<String, String> identifiedVertexKeyTypes,
...
+        Collection<String> edgeKeySet = getEdgeKeysAndNormalizeIfRequired(identifiedEdgeKeyTypes);
+        // in case vertex and edge may have the same attribute name, the key id in graphml have to be different
+        intersection = CollectionUtils.intersection(vertexKeySet, edgeKeySet);
+        // speeding-up later checks
+        if(intersection.isEmpty()){
```
twilmes: I can see what you're doing for the `null` check but I'm wondering if it's worth it versus just initializing the intersection to an empty set at the start to make the code a bit simpler. Did you do some performance comparison and identify that the null check approach was a lot faster?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphml/GraphMLWriter.java
```diff
@@ -218,20 +220,33 @@ private void writeTypes(final Map<String, String> identifiedVertexKeyTypes,
...
+            intersection = null;
+        }
         for (String key : vertexKeySet) {
             writer.writeStartElement(GraphMLTokens.KEY);
-            writer.writeAttribute(GraphMLTokens.ID, key);
+            if(intersection != null && intersection.contains(key)){
```
twilmes: Maybe here and below you could set a key variable once, something like:
```
String keyVal = key;
if (!intersection.isEmpty() && intersection.contains(key)) {
   keyVal = key.concat(GraphMLTokens.VERTEX_SUFFIX)
}
writer.writeAttribute(GraphMLTokens.ID, key);
```

spmallette:
@svanteschubert do you plan to come back to this one and #892 ? I think that if you address @twilmes issues here we could push forward at least this change and get it merged. Also, if you are coming back to this, I think we'd want to see some tests for this change somehow. None of the toy graphs have shared edge/vertex property keys that I can think of, so you might need special test graph data for this.

svanteschubert:
1) Yes, "Collection<String> vertexKeySet.." and same for the edge can still be made final, good catch.

2) No, I did not make any performance test, do you have one for Tinkerpop, what would you suggest? 
In any case, I leave it to your taste to adopt it to your desire {#emotions_dlg.wink}

3) There is a key declared in the header of the for loop, what is your suggestion? I do not fully understand your suggestion. Again, I have no strong feelings, in case you like to adopt the source. I only want to remove the reported problem.

4) The scenario comes from creating a graph with Tinkerpop using same-named property on edge and vertex and serializing it to GraphML. A regression test might be the extending an existing Graph example with the color property for an edge and vertex, which would be the correct location/class to create in Tinkerpop such a regression test?

svanteschubert:
What is the correct GitHub / Tinkerpop procedure to update a patch? Do you have a pointer?

I decided to start from scratch and cloned a fresh Tinkerpop repro, to start a fresh git branch from the latest sources. 

But the master has build problems on my machine with the latest patches on Windows (JDK 181).
Do you build as well on Windows? Going to reboot here.. ;-)

spmallette:
you should have been able to just keep pushing commits to the same branch you had and they would have reflected here in this pull request.

if you destroyed your repo and re-cloned then i think you will have to close this pull request and open a fresh one. if you want this particular commit as a starting point you should be able to still:

`git fetch upstream pull/891/head:pr-891`

where `upstream` is this repo and then cherry-pick the commit.  i assume that will work.

spmallette:
oh - and if your build is failing on the `gremlin-python` module then that is a known issue at this point. i only just fixed my local build and i'm still not sure how i did that. feel free to ignore failures in that module.

PR #894, opened 2018-07-20
TINKERPOP-2011 Use NumberHelper on choose()

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-2011

Treat numerical options in `ChooseStep` as in any other numerical comparison (value matters, data type does not). It was a little bit trickier than I thought as option values are only used for `HashMap` lookups and not for direct comparisons. Anyway, I think this solution is pretty solid (and easy to extend, should we run into any similar issues in the future).

VOTE: +1

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/branch/BranchStep.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/branch/BranchStep.java
@@ -260,7 +260,7 @@ public class BranchStep<S, E, M> extends ComputerAwareStep<S, E> implements Trav
 
         @Override
         public String toString() {
-            return "PickTokenKey(" + number + ")";
+            return number.toString();
         }
     }
 }


dkuppitz:
By the way, the `Dockerfile`, that's removed by this PR, has nothing to do with the PR, but the file is an autogenerated file that shouldn't be in the repository.

FlorianHockmann:
Looks like I accidentally added that `Dockerfile`. I should have checked more carefully what my commit added but shouldn't we prevent that from happening again with an entry in the `.gitignore` file?

dkuppitz:
Yep, done. I added an entry in the root `.gitignore` file and added another `.gitignore` file in `docker/hadoop/`. The two auto-generated `Dockerfile`s will be ignored from now on.

twilmes:
Code & tests LGTM.

VOTE: +1

FlorianHockmann:
Thanks for handling the `.gitignore`. I also reviewed the code and it looks good to me.

So: VOTE +1

PR #897, opened 2018-07-27
TINKERPOP-1967 connectedComponent()

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1967

Adds a `connectedComponent()` step. Was previously planning to hold this and #882 until i'd explored TINKERPOP-1991 in more detail, but I don't think adding that for 3.4.0 is reasonable. I think we cover a lot of the common problems/questions that we deal with day-to-day with these two algorithms so while we sorta extend the number of steps in the "core" space, Given that a separate "algorithm" space was determined after both of these steps were done, I think we can just include these for now and then work toward the "algorithm" space later. In the mean time we can look to avoid adding new "algorithm" steps to "core".

VOTE +1


FlorianHockmann:
Shouldn't we also add the `ConnectedComponent` class to the GLVs so users can easily do something like this in other languages:
```java
g.V().hasLabel('person').
  connectedComponent().
    with(ConnectedComponent.propertyName, 'component').
    with(ConnectedComponent.edges, outE('knows'))
```
?

docs/src/recipes/connected-components.asciidoc
```diff
@@ -35,46 +54,92 @@ g.addV().property(id, "A").as("a").
...
 ----
-g.V().emit(cyclicPath().or().not(both())).repeat(both()).until(cyclicPath()).  <1>
-  path().aggregate("p").                                                       <2>
-  unfold().dedup().                                                            <3>
-  map(__.as("v").select("p").unfold().                                         <4>
-         filter(unfold().where(eq("v"))).
-         unfold().dedup().order().by(id).fold()).
-  dedup()                                                                      <5>
+g.withComputer().V().connectedComponent().
+    group().by('gremlin.connectedComponentVertexProgram.component').
```
dkuppitz: I think you should use the constant `ConnectedComponentVertexProgram.COMPONENT`.

spmallette: This comment combined with the one from @FlorianHockmann is troublesome because GLVs don't have `ConnectedComponentVertexProgram` so the Gremlin won't work.  I think that's why I wrote the docs this way. Maybe I need to move `COMPONENT` to `ConnectedComponent` and then allow GLVs to have access. I'll do it that way. 

@dkuppitz do you have a similar issue with this?

https://github.com/apache/tinkerpop/pull/882/files#diff-36e52ac0c49a08a7f3e6ee54b60a3745R73

spmallette: Ended up dong this: https://github.com/apache/tinkerpop/commit/627b6e95b24130dea5582903add52c06b7b64a41

spmallette:
@FlorianHockmann regarding your point - yes, those tokens should be added to GLVs. I think we should wait for #893 to merge because it started the pattern for adding such tokens. I think that approach can be generalized better and then we can implement it for this PR and for the @dkuppitz at #882. It may even be best to simply merge all three PRs and then implement that improvement once they are all in a single branch (i'd be in favor of that approach).

spmallette:
@dkuppitz i just pushed a fix for halted traverser stuff - does that look right to you now? or am i still missing something?

dkuppitz:
Mucho mejor.

```
gremlin> g.withComputer().V().dedup().connectedComponent().valueMap()
==>[gremlin.connectedComponentVertexProgram.component:[1],name:[ripple],lang:[java]]
==>[gremlin.connectedComponentVertexProgram.component:[1],name:[marko],age:[29]]
==>[gremlin.connectedComponentVertexProgram.component:[1],name:[lop],lang:[java]]
==>[gremlin.connectedComponentVertexProgram.component:[1],name:[vadas],age:[27]]
==>[gremlin.connectedComponentVertexProgram.component:[1],name:[josh],age:[32]]
==>[gremlin.connectedComponentVertexProgram.component:[1],name:[peter],age:[35]]
```

Perhaps there should be a test case for this query, just in case somebody's gonna try to optimize the extra halted traverser code away.

spmallette:
cool - pushed a test

PR #898, opened 2018-07-29
gremlin-javascript: Typescript typings definition

daem0ndev:
Provides an index.d.ts typings definition file for TypeScript consumers of `gremin-javascript`.

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/index.d.ts
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/index.d.ts
@@ -1,0 +1,1243 @@
+/**
+  * ./lib/process/traversal
+  */
+namespace t {
+  class EnumValue {
+    constructor(typeName: string, elementName: string);
+    typeName: string;
+    elementName: string;
+    /**
+      * Returns elementName
+      */
+    toString(): string;
+  }
+
+  /**
+    * Represents an operation.
+    */
+  class P {
+    constructor(operator: string, value: string, other: string);
+    operator: string;
+    value: string;
+    other: string;
+    /**
+      * Returns the string representation of the instance.
+      * @returns {string}
+      */
+    toString(): string;
+    and(arg: string): string;
+    or(arg: string): string;
+    static between(...args): P;
+    static eq(...args): P;
+    static gt(...args): P;
+    static gte(...args): P;
+    static inside(...args): P;
+    static lt(...args): P;
+    static lte(...args): P;
+    static neq(...args): P;
+    static not(...args): P;
+    static outside(...args): P;
+    static test(...args): P;
+    static within(...args): P;
+    static without(...args): P;
+  }
+
+  class Traversal {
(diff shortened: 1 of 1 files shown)


daem0ndev:
I'm working on a NestJS module for gremlin client that will essentially be a clean typescript wrapper over this gremlin-javascript library. If there is any interest, I'll gladly publish it.

spmallette:
Hi, thanks for the contribution. I'm sure @jorgebay will look in on this when he's back from vacation.  In the mean time, for my own information, could you comment on how this file is to be maintained? do changes to gremlin-java that then mean changes to gremlin-javascript imply manual changes to this file?

daem0ndev:
yes unfortunately its manually maintained right now, although I did my best to organize the file in such a way that its easy to map it to the default exports. The only way to get this for free is if we converted the whole project to true typescript :)

jorgebay:
Nice @daem0ndev! I think its a good idea to provide typing definitions inside the package.

[Gremlin Variant APIs](http://tinkerpop.apache.org/docs/current/tutorials/gremlin-language-variants/) like gremlin-javascript are generated from the Java API using a set of groovy templates and scripts, see https://github.com/apache/tinkerpop/blob/master/gremlin-javascript/glv/generate.groovy for example. We should do the same in this case, do you want to adapt your pull request to generate the `.ts` file?

Also, could you target `tp32` branch instead?

spmallette:
That was going to be my suggestion too - I think that it can be auto-generated then it should be.

mattallenuk:
I'm also using Gremlin Javascript in a Typescript project. I'd be happy to give the typings file generation a crack if @daem0ndev is not able to.

daem0ndev:
hey folks! sorry i went missing :) so @mattallenuk please feel free to go ahead, we actually ran into some hiccups when trying to integrate with gremlin-javascript and remote calls when doing things like serializing a subgraph and using lambda functions so we went to a java based api layer instead so I have not needed the javascript / typescript package. Would be great if it supported some of those things (maybe it does and I just dont know how to use it?) lol

spmallette:
`subgraph()` serialization is a major problem without a clear solution for GLVs. there is nothing for anyone to deserialize a returned subgraph into for GLVs and so it has no where to go. all you can do with a GLV is return a form of edge list i guess and then reconstruct your graph on the client in "something else" (whatever that is - perhaps an in-memory graph written in that native language??).  Not sure where to go there until TinkerPop gets in the business of building Gremlin Virtual Machines in every language at which point you'd have a `tinkergraph-js` and a `tinkergraph-python` and so on....

jorgebay:
> I'd be happy to give the typings file generation 

@mattallenuk I saw you've [created the ticket](https://issues.apache.org/jira/browse/TINKERPOP-2027): Nice! we can discuss on the tickets if there are any doubts.

spmallette:
I've closed this pull request given the recent discussion here to deal with this on TINKERPOP-2027.

PR #899, opened 2018-07-30
TINKERPOP-2005: Intermittent NullPointerException in response handling

dimas-b:
Address [TINKERPOP-2005](https://issues.apache.org/jira/browse/TINKERPOP-2005) by ignoring more than one "final" response messages from `AbstractEvalOpProcessor`.

Add `isFinalResponse()` getter to `ResponseStatusCode`

Introduce `ResponseHandlerContext` to allow tracking the final response
status per request message.

Update `AbstractOpProcessor`, `AbstractEvalOpProcessor` and related classes
to write response messages through `ResponseHandlerContext` methods as
opposed to `ChannelHandlerContext` methods.

----

In addition to the failure mode described in [TINKERPOP-2005](https://issues.apache.org/jira/browse/TINKERPOP-2005) non-timeout script evaluation exceptions in AbstractEvalOpProcessor may occur after the script has started executing. In this situation it it critical to prevent potentially writing multiple final (e.g. error vs. success) responses back to the client. This failure mode is hard to replicate in a pure Gremlin server, however the new `AbstractEvalOpProcessorTest` class provides coverage for it.

Because of this failure mode ResponseHandlerContext.writeAndFlush(...) methods should not throw exceptions for attempts to send multiple final messages. This is to avoid multiple error response messages sent from OpExecutorHandler (which has a `catch(Exception)` clause).

----

Testing performed:

* Full suite of unit tests (i.e. `mvn clean install`)
* Full suite of `gremlin-server` integration tests (`mvn verify -pl gremlin-server -DskipIntegrationTests=false`)
* Manual testing according to comment under [TINKERPOP-2005](https://issues.apache.org/jira/browse/TINKERPOP-2005).

### DIFF
--- a/gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/message/ResponseStatusCodeTest.java
+++ b/gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/message/ResponseStatusCodeTest.java
@@ -21,7 +21,8 @@ package org.apache.tinkerpop.gremlin.driver.message;
 
 import org.junit.Test;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 
 public class ResponseStatusCodeTest {
 
--- a/gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/message/ResponseStatusCodeTest.java
+++ b/gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/message/ResponseStatusCodeTest.java
@@ -1,0 +1,35 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.tinkerpop.gremlin.driver.message;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class ResponseStatusCodeTest {
+
+    @Test
+    public void shouldIndicateFinalMessagesStatusCodes() {
+        assertFalse(ResponseStatusCode.AUTHENTICATE.isFinalResponse());
+        assertFalse(ResponseStatusCode.PARTIAL_CONTENT.isFinalResponse());
+        assertTrue(ResponseStatusCode.SUCCESS.isFinalResponse());
+        assertTrue(ResponseStatusCode.SERVER_ERROR_TIMEOUT.isFinalResponse());
+    }
+}
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/ResponseHandlerContext.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/ResponseHandlerContext.java
@@ -1,0 +1,80 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.tinkerpop.gremlin.server;
+
+import org.apache.tinkerpop.gremlin.driver.message.ResponseMessage;
+import org.apache.tinkerpop.gremlin.driver.message.ResponseStatusCode;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * A context for asynchronously writing response messages related to a particular request.
+ * <p>The "write" methods of this class ensure that at most one {@link ResponseStatusCode#isFinalResponse() final}
+ * response message is written to the underlying channel. Attempts to write more than one final response message will
+ * result in an {@link IllegalStateException}.</p>
+ * <p>Note: an object of this class should be used instead of writing to the channel directly when multiple threads
+ * are expected to produce final response messages concurrently. Callers must ensure that the same
+ * {@link ResponseHandlerContext} is used by all threads writing response messages for the same request.</p>
+ *
+ * @author Dmitri Bourlatchkov
+ */
+public class ResponseHandlerContext {
+
+    private final Context context;
+    private final AtomicBoolean finalResponseWritten = new AtomicBoolean();
+
+    public ResponseHandlerContext(Context context) {
+        this.context = context;
+    }
+
+    public Context getContext() {
+        return context;
+    }
+
+    /**
+     * Writes a response message to the underlying channel while ensuring that at most one
+     * {@link ResponseStatusCode#isFinalResponse() final} response is written.
+     * <p>Note: this method should be used instead of writing to the channel directly when multiple threads
+     * are expected to produce response messages concurrently.</p>
+     * <p>Attempts to write more than one final response message will result in an {@link IllegalStateException}.</p>
+     * @see #writeAndFlush(ResponseStatusCode, Object)
+     */
+    public void writeAndFlush(ResponseMessage message) {
+        writeAndFlush(message.getStatus().getCode(), message);
+    }
+
+    /**
+     * Writes a response message to the underlying channel while ensuring that at most one
+     * {@link ResponseStatusCode#isFinalResponse() final} response is written.
+     * <p>The caller must make sure that the provided response status code matches the content of the message.</p>
+     * <p>Note: this method should be used instead of writing to the channel directly when multiple threads
+     * are expected to produce response messages concurrently.</p>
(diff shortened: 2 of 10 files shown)


gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/message/ResponseStatusCodeTest.java
```diff
@@ -0,0 +1,35 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.tinkerpop.gremlin.driver.message;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
```
spmallette: Please make a pass through your PR to remove wildcard imports as our style is to explicitly list imports individually.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/ResponseHandlerContext.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @author Dmitri Bourlatchkov
+ */
+public class ResponseHandlerContext {
+    private static final Logger logger = LoggerFactory.getLogger(ResponseHandlerContext.class);
+
+    private final Context context;
+    private final AtomicBoolean finalResponseWritten = new AtomicBoolean();
+
+    public ResponseHandlerContext(Context context) {
```
spmallette: Please make a pass through your PR to look for additional places where you can `final` variables - it's our code style to be explicit in that way.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractEvalOpProcessor.java
```diff
@@ -222,9 +223,33 @@ protected AbstractEvalOpProcessor(final boolean manageTransactions) {
...
+            rhc.writeAndFlush(ResponseMessage.build(context.getRequestMessage())
+                    .code(ResponseStatusCode.SERVER_ERROR)
+                    .statusAttributeException(ex)
+                    .statusMessage(ex.getMessage()).create());
+        }
+    }
+
+    /**
+     * A variant of {@link #evalOpInternal(Context, Supplier, BindingSupplier)} that is suitable for use in situations
+     * when multiple threads may produce {@link ResponseStatusCode#isFinalResponse() final} response messages
```
spmallette: seems like the "final" should be moved out of the `{@link }` bracket?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractEvalOpProcessor.java
```diff
@@ -222,9 +223,33 @@ protected AbstractEvalOpProcessor(final boolean manageTransactions) {
      *                                script evaluation.
      * @param bindingsSupplier A function that returns the {@link Bindings} to provide to the
      *                         {@link GremlinExecutor#eval} method.
+     * @see #evalOpInternal(ResponseHandlerContext, Supplier, BindingSupplier)
      */
     protected void evalOpInternal(final Context context, final Supplier<GremlinExecutor> gremlinExecutorSupplier,
```
spmallette: it was good that you kept the old signature as other people might depend on this. i was thinking we might deprecate, this version of the method, but perhaps that is not necessary or perhaps even undesired. I guess, the question is whether or not the new method would ever be called directly or would it always be called through this method? Perhaps the other method should be `private`?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractOpProcessor.java
```diff
@@ -72,8 +73,19 @@ protected AbstractOpProcessor(final boolean manageTransactions) {
...
     protected void handleIterator(final Context context, final Iterator itty) throws TimeoutException, InterruptedException {
+        handleIterator(new ResponseHandlerContext(context), itty);
+    }
+
+    /**
+     * A variant of {@link #handleIterator(Context, Iterator)} that is suitable for use in situations when multiple
+     * threads may produce {@link ResponseStatusCode#isFinalResponse() final} response messages concurrently.
+     * @see #handleIterator(Context, Iterator)
+     */
+    protected void handleIterator(final ResponseHandlerContext rhc, final Iterator itty) throws TimeoutException, InterruptedException {
```
spmallette: same question here as on `evalOpInternal()` - is it reasonable for someone to directly call this method? or is it more likely that they would continue to call the original method which would let us control creation of the `ResponseHandlerContext` ourselves?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractOpProcessor.java
```diff
@@ -72,8 +73,19 @@ protected AbstractOpProcessor(final boolean manageTransactions) {
...
      * @throws TimeoutException if the time taken to serialize the entire result set exceeds the allowable time.
+     * @see #handleIterator(ResponseHandlerContext, Iterator)
      */
     protected void handleIterator(final Context context, final Iterator itty) throws TimeoutException, InterruptedException {
+        handleIterator(new ResponseHandlerContext(context), itty);
+    }
+
+    /**
+     * A variant of {@link #handleIterator(Context, Iterator)} that is suitable for use in situations when multiple
+     * threads may produce {@link ResponseStatusCode#isFinalResponse() final} response messages concurrently.
```
spmallette: should `final` be moved out of the brackets?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractOpProcessor.java
```diff
@@ -252,15 +264,29 @@ protected boolean isForceFlushed(final ChannelHandlerContext ctx, final RequestM
...
                                      final MessageSerializer serializer, final boolean useBinary, final List<Object> aggregate,
                                      final ResponseStatusCode code) throws Exception {
         return makeFrame(ctx, msg, serializer, useBinary, aggregate, code, Collections.emptyMap());
     }
 
+    /**
+     * Caution: {@link #makeFrame(ResponseHandlerContext, RequestMessage, MessageSerializer, boolean, List, ResponseStatusCode, Map)}
+     * should be used instead of this method whenever a {@link ResponseHandlerContext} is available.
+     */
     protected static Frame makeFrame(final ChannelHandlerContext ctx, final RequestMessage msg,
```
spmallette: I would just deprecate this I think...I mean, if proper operation requires that people call the new method then maybe we should push people to that method more directly.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractEvalOpProcessor.java
dimas-b: I actually intended it to be this way to create a hyperlink for the word "final" pointing to the "isFinalResponse()" method. Would you prefer to have the link named the same as the target method?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractEvalOpProcessor.java
dimas-b: Yeah, I was thinking about that but could not really reach any decision myself, so I kept it more accessible. In current code no other classes need to call the new method, so I'm ok to make it `private`.

However, if we keep the new `evalOpInternal` method `private` by the same logic it might be reasonable to keep the new `makeFrame` method `private` as well. If other classes ever need to use it directly it would be under similar conditions as calling the new `evalOpInternal` method... and right now other classes do not need to do that.

What do you think?

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/message/ResponseStatusCodeTest.java
dimas-b: Sorry, missed that. Fixed in 63ae13c24.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/ResponseHandlerContext.java
dimas-b: Sorry, missed that. Fixed in 63ae13c24.

dimas-b:
I see a test failure related to log message interception in the automated build:

```
[ERROR] shouldAllowAtMostOneFinalResponse[writeAndFlush(ResponseStatusCode, Object)](org.apache.tinkerpop.gremlin.server.ResponseHandlerContextTest)  Time elapsed: 0.027 s  <<< FAILURE!
java.lang.AssertionError
	at org.apache.tinkerpop.gremlin.server.ResponseHandlerContextTest.shouldAllowAtMostOneFinalResponse(ResponseHandlerContextTest.java:118)
```

I'll look into that presently (works in my env., of course :) )

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/ResponseHandlerContext.java
dimas-b: Fixed in 0e0d26cda.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractEvalOpProcessor.java
spmallette: ha - i didn't know that was a possible javadoc syntax. if that's valid for a link then it's fine.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractEvalOpProcessor.java
spmallette: well - i guess, keep it open as `protected` - i guess if someone is writing one of these they will need to know how all of it works  i guess. perhaps this is where other PRs comes in for tp33 and master. maybe we deprecate in 3.2.x and then remove them for 3.4.0.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractOpProcessor.java
spmallette: ok - i' mostly just thinking out loud. hard to predict exactly how folks are using this or will use this.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractOpProcessor.java
dimas-b: keeping current link syntax (as discussed in comments under `AbstractEvalOpProcessor.java`)

spmallette:

VOTE +1

PR #903, opened 2018-08-01
Gremlin.Net: Add ConnectionPool min and max sizes TINKERPOP-1774

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-1774

The Gremlin.Net `ConnectionPool` now has min and max sizes. The pool will be initialized with the configured minimum number of connections on creation. It will also no longer create an unlimited number of connections. Instead, a `TimeoutException` will be thrown when the max limit is reached and no connection became available within a configurable time.

The implementation is very similar to that of the `ConnectionPool` in the Java driver. This also means that it doesn't contain a lock any more which was already suggested in the ticket by @jorgebay.

I also optimized a bit of async code in the driver by removing `await`s that weren't necessary because they were immediately returned. In those cases, the `Task` can also be just returned without being awaited which avoids the need for the compiler to generate a state machine just for the `await`.

Tests passed with `./docker/build.sh -t -i` a few days ago but I made some minor changes afterwards and also rebased on `master` and now I can't build any more due to the current problem with gremlin-python mentioned by @spmallette on the mailing list. 

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,7 +25,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 This release also includes changes from <<release-3-3-3, 3.3.3>>.
 
-* Added min and max connection pool sizes for Gremlin.Net which are configurable through optional ConnectionPoolSettings.
+* Added min and max connection pool sizes for Gremlin.Net which are configurable through optional `ConnectionPoolSettings`.
 * Implemented `ShortestPathVertexProgram` and the `shortestPath()` step.
 * `AbstractGraphProvider` uses `g.io()` for loading test data.
 * Added the `io()` start step and `read()` and `write()` termination steps to the Gremlin language.
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,6 +25,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 This release also includes changes from <<release-3-3-3, 3.3.3>>.
 
+* Added min and max connection pool sizes for Gremlin.Net which are configurable through optional ConnectionPoolSettings.
 * Implemented `ShortestPathVertexProgram` and the `shortestPath()` step.
 * `AbstractGraphProvider` uses `g.io()` for loading test data.
 * Added the `io()` start step and `read()` and `write()` termination steps to the Gremlin language.
--- a/docs/src/upgrade/release-3.4.x.asciidoc
+++ b/docs/src/upgrade/release-3.4.x.asciidoc
@@ -242,9 +242,9 @@ when dealing with that event. To make this easier, the event now raises with a `
 
 link:https://issues.apache.org/jira/browse/TINKERPOP-1831[TINKERPOP-1831]
 
-==== Gremlin.Net: Configurable Max and Min ConnectionPool Sizes
+==== Gremlin.Net Connection Pool
 
-Gremlin.Net's `ConnectionPool` now has a minimum and a maximum size. These sizes are configurable through added
+Gremlin.Net's `ConnectionPool` now has a minimum and a maximum size. These sizes are configurable through the newly added
 `ConnectionPoolSettings`. The minimum size determines how many connections are initially created. The maximum size
 is an upper limit of connections that can be created. When this limit is reached and another connection is needed,
 then the connection pool waits for a connection to become available again. The time to be waited is limited by the
--- a/docs/src/upgrade/release-3.4.x.asciidoc
+++ b/docs/src/upgrade/release-3.4.x.asciidoc
@@ -242,6 +242,17 @@ when dealing with that event. To make this easier, the event now raises with a `
 
 link:https://issues.apache.org/jira/browse/TINKERPOP-1831[TINKERPOP-1831]
 
+==== Gremlin.Net: Configurable Max and Min ConnectionPool Sizes
+
+Gremlin.Net's `ConnectionPool` now has a minimum and a maximum size. These sizes are configurable through added
+`ConnectionPoolSettings`. The minimum size determines how many connections are initially created. The maximum size
+is an upper limit of connections that can be created. When this limit is reached and another connection is needed,
+then the connection pool waits for a connection to become available again. The time to be waited is limited by the
+newly introduced option `ConnectionPoolSettings.WaitForConnectionTimeout`. A `TimeoutException` is thrown when
+no connection becomes available until this timeout is reached.
+
+See: link:https://issues.apache.org/jira/browse/TINKERPOP-1774[TINKERPOP-1774]
+
 ==== Deprecation Removal
 
 The following deprecated classes, methods or fields have been removed in this version:
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/AsyncAutoResetEvent.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/AsyncAutoResetEvent.cs
@@ -25,7 +25,7 @@ using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 
-// The implementation is based on this blog post by Stephen Toub:
+// The implementation of this class is inspired by this blog post from Stephen Toub:
 // https://blogs.msdn.microsoft.com/pfxteam/2012/02/11/building-async-coordination-primitives-part-2-asyncautoresetevent/
 
 namespace Gremlin.Net.Driver
@@ -42,25 +42,19 @@ namespace Gremlin.Net.Driver
         /// <summary>
         ///     Asynchronously waits for this event to be set or until a timeout occurs.
         /// </summary>
-        /// <param name="timeout">A <see cref="TimeSpan"/> that that represents the number of milliseconds to wait.</param>
+        /// <param name="timeout">A <see cref="TimeSpan"/> that represents the number of milliseconds to wait.</param>
         /// <returns>true if the current instance received a signal before timing out; otherwise, false.</returns>
         public async Task<bool> WaitOneAsync(TimeSpan timeout)
         {
             var tcs = new TaskCompletionSource<bool>();
             var waitTask = WaitForSignalAsync(tcs);
-            if (waitTask.IsCompleted) return true;
+            if (waitTask.IsCompleted) return waitTask.Result;
             
             await Task.WhenAny(waitTask, Task.Delay(timeout)).ConfigureAwait(false);
-            lock (_waitingTasks)
-            {
-                if (!waitTask.IsCompleted)
-                {
-                    // The wait timed out, so we need to remove the waiting task.
-                    _waitingTasks.Remove(tcs);
-                    tcs.SetResult(false);
-                }
-            }
+            if (waitTask.IsCompleted) return waitTask.Result;
             
+            StopWaiting(tcs);
+
             return waitTask.Result;
         }
 
@@ -76,6 +70,15 @@ namespace Gremlin.Net.Driver
                 _waitingTasks.Add(tcs);
             }
             return tcs.Task;
+        }
+
+        private void StopWaiting(TaskCompletionSource<bool> tcs)
+        {
+            lock (_waitingTasks)
+            {
+                _waitingTasks.Remove(tcs);
+                tcs.SetResult(false);
+            }
         }
         
         /// <summary>
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/AsyncAutoResetEvent.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/AsyncAutoResetEvent.cs
@@ -1,0 +1,103 @@
+﻿#region License
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+// The implementation is based on this blog post by Stephen Toub:
+// https://blogs.msdn.microsoft.com/pfxteam/2012/02/11/building-async-coordination-primitives-part-2-asyncautoresetevent/
+
+namespace Gremlin.Net.Driver
+{
+    /// <summary>
+    ///     An async version of the AutoResetEvent.
+    /// </summary>
+    public class AsyncAutoResetEvent
+    {
+        private static readonly Task<bool> CompletedTask = Task.FromResult(true);
+        private readonly List<TaskCompletionSource<bool>> _waitingTasks = new List<TaskCompletionSource<bool>>();
+        private bool _isSignaled;
+        
+        /// <summary>
+        ///     Asynchronously waits for this event to be set or until a timeout occurs.
+        /// </summary>
+        /// <param name="timeout">A <see cref="TimeSpan"/> that that represents the number of milliseconds to wait.</param>
+        /// <returns>true if the current instance received a signal before timing out; otherwise, false.</returns>
+        public async Task<bool> WaitOneAsync(TimeSpan timeout)
+        {
+            var tcs = new TaskCompletionSource<bool>();
+            var waitTask = WaitForSignalAsync(tcs);
+            if (waitTask.IsCompleted) return true;
+            
+            await Task.WhenAny(waitTask, Task.Delay(timeout)).ConfigureAwait(false);
+            lock (_waitingTasks)
+            {
+                if (!waitTask.IsCompleted)
+                {
+                    // The wait timed out, so we need to remove the waiting task.
+                    _waitingTasks.Remove(tcs);
+                    tcs.SetResult(false);
+                }
+            }
+            
+            return waitTask.Result;
+        }
+
+        private Task<bool> WaitForSignalAsync(TaskCompletionSource<bool> tcs)
+        {
+            lock (_waitingTasks)
+            {
+                if (_isSignaled)
+                {
+                    _isSignaled = false;
+                    return CompletedTask;
+                }
+                _waitingTasks.Add(tcs);
+            }
+            return tcs.Task;
+        }
+        
+        /// <summary>
+        ///     Sets the event.
+        /// </summary>
+        public void Set()
+        {
+            TaskCompletionSource<bool> toRelease = null;
+            lock (_waitingTasks)
+            {
+                if (_waitingTasks.Count == 0)
+                {
+                    _isSignaled = true;
+                }
+                else
+                {
+                    toRelease = _waitingTasks[0];
+                    _waitingTasks.RemoveAt(0);
+                }
+            }
+
+            toRelease?.SetResult(true);
+        }
+    }
+}
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
@@ -23,7 +23,8 @@
 
 using System;
 using System.Collections.Concurrent;
-using System.Linq;
+using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
 using Gremlin.Net.Process;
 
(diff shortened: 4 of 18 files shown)


CHANGELOG.asciidoc
```diff
@@ -25,6 +25,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 This release also includes changes from <<release-3-3-3, 3.3.3>>.
 
+* Added min and max connection pool sizes for Gremlin.Net which are configurable through optional ConnectionPoolSettings.
```
spmallette: nit: could you place the single ticks around the `ConnectionPoolSettings` please?

docs/src/upgrade/release-3.4.x.asciidoc
```diff
@@ -177,6 +177,17 @@ when dealing with that event. To make this easier, the event now raises with a `
 
 link:https://issues.apache.org/jira/browse/TINKERPOP-1831[TINKERPOP-1831]
 
+==== Gremlin.Net: Configurable Max and Min ConnectionPool Sizes
```
spmallette: Please shorten the header a bit because it turns into a table of contents item that ends up crowding the frame. Perhaps just go with "Gremlin.Net Connection Pool" - it just needs to be long enough to catch someone's eye who is looking.

docs/src/upgrade/release-3.4.x.asciidoc
```diff
@@ -177,6 +177,17 @@ when dealing with that event. To make this easier, the event now raises with a `
...
+
+Gremlin.Net's `ConnectionPool` now has a minimum and a maximum size. These sizes are configurable through added
```
spmallette: Minor grammatical fix by adding the phrase "the newly" -  "These sizes are configurable through the newly added `ConnectionPoolSettings`".

docs/src/upgrade/release-3.4.x.asciidoc
spmallette: Also, while this is a nice explanation of the this feature, perhaps it's time to add a "Configuration" section to the reference documentation itself? If no one else agrees, that's fine - just a thought I had given this new feature. No new users are going to find this nice explanation here in the upgrade docs i don't think.

gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPoolSettings.cs
```diff
@@ -0,0 +1,55 @@
...
+    ///     Holds settings for the <see cref="ConnectionPool"/>.
+    /// </summary>
+    public class ConnectionPoolSettings
+    {
+        private const int DefaultMinPoolSize = 8;
+        private const int DefaultMaxPoolSize = 128;
+        private static readonly TimeSpan DefaultWaitForConnectionTimeout = TimeSpan.FromSeconds(3);
+
+        /// <summary>
+        ///     Gets or sets the minimum size of a connection pool.
```
spmallette: Once we have a pattern for this:

https://issues.apache.org/jira/browse/TINKERPOP-2010

We should do the same for .NET so this nice structured documentation gets published.

spmallette:
If you rebase at this point (and rebuild docker images) you should get the build good again.

gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPoolSettings.cs
FlorianHockmann: Yes, that's a good idea. The API docs should be published somewhere. I already did this for the early pre-TinkerPop version of Gremlin.Net: https://florianhockmann.github.io/Gremlin.Net/api/Gremlin.Net.html
We could probably use a similar approach again.

docs/src/upgrade/release-3.4.x.asciidoc
FlorianHockmann: I like the idea of adding docs for the configuration of Gremlin.Net to the docs, especially since the configuration is getting more and more complicated. I'll try to come up with some docs for this.

gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPoolSettings.cs
FlorianHockmann: I created [TINKERPOP-2018](https://issues.apache.org/jira/browse/TINKERPOP-2018) to track this.

FlorianHockmann:
I rebased and added a commit that should address your comments. The reference docs now contain a _Configuration_ section for Gremlin.Net.

I was also able to run all tests with `./docker/build.sh -t -i` (thanks @dkuppitz for fixing the Docker image).

spmallette:
VOTE +1

gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
```diff
@@ -33,91 +34,134 @@ internal class ConnectionPool : IDisposable
...
+            var start = DateTimeOffset.Now;
+            var remaining = _waitForConnectionTimeout;
+            do
             {
-                ConsiderUnavailable();
-                connection.Dispose();
-                return;
-            }
-            AddConnection(connection);
+                if (_newConnectionAvailable.WaitOne(remaining))
```
jorgebay: This loop blocks the calling thread, which is a problem as we would be mixing async and blocking code.

We should implemented in a way that it awaits for a connection to be available. Here are some good starting points (I'm not aware of a built-in mechanism in .NET): 
- https://github.com/StephenCleary/AsyncEx/wiki/AsyncAutoResetEvent
- https://blogs.msdn.microsoft.com/pfxteam/2012/02/11/building-async-coordination-primitives-part-2-asyncautoresetevent/

FlorianHockmann: Thanks for pointing this out @jorgebay 

I actually experimented with something like an `AutoResetEvent` for the very first version of Gremlin.Net, but didn't get it to work and therefore ended up with the plain `lock`. I guess what I was missing back then was simply the knowledge about the [CAS operation](https://en.wikipedia.org/wiki/Compare-and-swap) and how it can be used to avoid synchronous locks.
I'll give this a try again to come up with a version that hopefully completely removes synchronous blocking. I guess this will result in our own version of an `AsyncAutoResetEvent` that allows us to wait with a timeout as that seems to be missing in both existing implementations (from the blog and the library you linked). (I'm also not sure whether it's worth it to add a dependency on `AsyncEx` just to get one class.)

jorgebay: I would prefer getting the `AsyncAutoResetEvent` source code (If the license permits it) or develop our own version, than adding a dependency.

FlorianHockmann: I finally found some time to work on this again and implemented an `AsyncAutoResetEvent` that is based on the blog post by Stephen Toub. The main difference to the version in the blog post is that it only accepts waits that specify a timeout and it uses a `List` as the underlying data structure instead of a `Queue` as we need to remove wait tasks from it that timed out.
Would be good if you could give this a thorough review as it's probably easy to make small mistakes in a primitive like this that will be hard to find later on.

gremlin-dotnet/src/Gremlin.Net/Driver/AsyncAutoResetEvent.cs
```diff
@@ -0,0 +1,103 @@
...
+ * under the License.
+ */
+
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+// The implementation is based on this blog post by Stephen Toub:
```
jorgebay: As we are licensing this source code, please note that the source code is original, even if it was inspired by Stephen Toub concepts / ideas.

gremlin-dotnet/src/Gremlin.Net/Driver/AsyncAutoResetEvent.cs
```diff
@@ -0,0 +1,103 @@
...
+        /// <param name="timeout">A <see cref="TimeSpan"/> that that represents the number of milliseconds to wait.</param>
+        /// <returns>true if the current instance received a signal before timing out; otherwise, false.</returns>
+        public async Task<bool> WaitOneAsync(TimeSpan timeout)
+        {
+            var tcs = new TaskCompletionSource<bool>();
+            var waitTask = WaitForSignalAsync(tcs);
+            if (waitTask.IsCompleted) return true;
+            
+            await Task.WhenAny(waitTask, Task.Delay(timeout)).ConfigureAwait(false);
+            lock (_waitingTasks)
```
jorgebay: This `lock` statement can be moved within `if (!waitTask.IsCompleted)` block. That way, if the task is completed no lock is involved.

gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
```diff
@@ -33,91 +34,135 @@ internal class ConnectionPool : IDisposable
...
             return newConnection;
         }
 
-        private void AddConnectionIfOpen(Connection connection)
+        private async Task<Connection> WaitForConnectionAsync()
         {
-            if (!connection.IsOpen)
+            var start = DateTimeOffset.Now;
+            var remaining = _waitForConnectionTimeout;
+            do
```
jorgebay: I think that the loop doesn't make much sense now. We only have to `await` for a connection available once and return it or return or throw an `TimeoutException`.

gremlin-dotnet/src/Gremlin.Net/Driver/AsyncAutoResetEvent.cs
```diff
@@ -0,0 +1,103 @@
...
+        /// </summary>
+        /// <param name="timeout">A <see cref="TimeSpan"/> that that represents the number of milliseconds to wait.</param>
+        /// <returns>true if the current instance received a signal before timing out; otherwise, false.</returns>
+        public async Task<bool> WaitOneAsync(TimeSpan timeout)
+        {
+            var tcs = new TaskCompletionSource<bool>();
+            var waitTask = WaitForSignalAsync(tcs);
+            if (waitTask.IsCompleted) return true;
+            
+            await Task.WhenAny(waitTask, Task.Delay(timeout)).ConfigureAwait(false);
```
jorgebay: Each `Task.Delay()` uses an individual timer that are [a limited resource](https://blogs.msdn.microsoft.com/oldnewthing/20090827-00/?p=16963/).

A workaround would be to have a single timer that ticks at certain interval (e.g: 100ms) and checks the queue of tasks.

jorgebay:
Also, I'm not sure I understand how this will come into place with TINKERPOP-1775.

Currently a connection is being used to one send and receive one request/response pair at a time. We should allow sending multiple requests without waiting for a response and support unordered responses (based on the request id). You can look at the gremlin-javascript connection for example:
- When the request is sent, a [callback handler is registered](https://github.com/apache/tinkerpop/blob/3.3.3/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/driver-remote-connection.js#L106-L109)
- When the response is obtained, [the callback is obtained by request id and invoked](https://github.com/apache/tinkerpop/blob/3.3.3/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/driver-remote-connection.js#L146).

We can use the pool to balance the requests between connections. If we want to throttle requests, we should introduce a new setting like "max inflight per connection".

gremlin-dotnet/src/Gremlin.Net/Driver/AsyncAutoResetEvent.cs
FlorianHockmann: Are you sure that that is still true? I just did a quick test and executed 1 Mio. times `Task.Delay` in parallel with a 10 second delay and the whole test finished in 11 seconds. Unfortunately I couldn't find any more up-to-date information about this.

The full test for reference:
```cs
[Fact]
public async Task TimersTest()
{
    const int nrTasks = 1000000;
    var tasks = new List<Task>(nrTasks);
    for (var i = 0; i < nrTasks; i++)
    {
        tasks.Add(Task.Delay(TimeSpan.FromSeconds(10)));
    }

    await Task.WhenAll(tasks);
}
```

gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
FlorianHockmann: The problem is that another thread can call `TryGetConnectionFromPool` without waiting for a signal on `_newConnectionAvailable` as that's the first thing happening in `GetAvailableConnectionAsync`.
So, even if we get a signal that a new connection is available, we still can't be sure that we are actually getting this connection. Therefore, the thread can wait again for another signal until it actually reaches its timeout. (The Java driver does it the same way.)

gremlin-dotnet/src/Gremlin.Net/Driver/AsyncAutoResetEvent.cs
jorgebay: hmm... indeed, it looks like in .NET Core it uses a single native timer for all timer instances: https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/src/System/Threading/Timer.cs

I wouldn't know about other runtimes.

FlorianHockmann: I just tried the same test with .NET Framework 4.6 and got the same result. This is also in line with [the reference source for .NET Framework 4.7.2](https://referencesource.microsoft.com/#mscorlib/system/threading/timer.cs). Mono seems to use this reference source as a basis [for their implementation](https://github.com/mono/mono/blob/c5b88ec4f323f2bdb7c7d0a595ece28dae66579c/mcs/class/referencesource/mscorlib/system/threading/timer.cs).

Using `Task.Delay()` together with `Task.WhenAny()` to time out tasks is also a pattern recommended in the two resources you linked to earlier. So, it would be really strange if it worked only for a small number of tasks.

Are you still in favour of a more complicated implementation like such a timer queue you mentioned?

FlorianHockmann:
> We should allow sending multiple requests without waiting for a response and support unordered responses (based on the request id)

Wouldn't that mean that the server sends responses as soon as they are ready on the connection that was used to send the request? And couldn't that result in multiple responses being sent over the same connection in parallel? In that case I'm afraid that it wouldn't work with the `ClientWebSocket` we're currently relying on due to this restriction:

> Exactly one send and one receive is supported on each ClientWebSocket object in parallel.

[Source](https://docs.microsoft.com/dotnet/api/system.net.websockets.clientwebsocket.sendasync?view=netcore-2.1#remarks)

Or did I misunderstand this somehow?

gremlin-dotnet/src/Gremlin.Net/Driver/AsyncAutoResetEvent.cs
FlorianHockmann: I'm not sure how to express that without stating me explicitly as the author which I would like to avoid. I changed the formulation a bit to make it more clear that the source code was not copied. Let me know if you have any other suggestions.

jorgebay:
Its one send at a time, that means that we have to send it one at a time but not necessary awaiting to receive after each send.

```c#
// Send in a loop until the queue is exhausted
while (writeQueue.TryDequeue(out var item))
{
    await ws.SendAsync(item);
}
```

```c#
// Always have an outstanding receiveasync call
// We should not await on the ReceiveAsync()
ws.ReceiveAsync().ContinueWith(
  item => 
  {
    ParseItem(item);
    ReceiveNext();
  }
  , ExecuteSynchronously);
```

FlorianHockmann:
Thanks for the clarification. I think I understand now how that would look like. It is of course a nice improvement. However, I wonder what would be the concrete advantages for the user? One advantage I can see is that it should reduce the number of connections necessary to deal with the same amount of requests as we wouldn't block connections while waiting on a response from the server any more.
Do you see other advantages?

More importantly for this PR: To me, connection pipelining ([TINKERPOP-1775](https://issues.apache.org/jira/browse/TINKERPOP-1775)) seems to be mostly unrelated to this PR. We would still need a connection pool and that pool should still have configurable min and max sizes. Adding a setting like _max inflight per connection_ would make the pool maybe a bit more complicated but I think that the changes in this PR still make sense. What do you say, @jorgebay?

jorgebay:
> Do you see other advantages?
It allows the client to prime the server with enough requests in order to support maximum throughput. Considering the multithreaded asynchronous nature of the server, there is far more advantages in request pipelining than with multiple connections.
Considering that the server is multithreaded, you can think multiple connections w/o request pipelining as a system using multiple threads and blocking on each thread; being a single connection with request pipelining the equivalent of async non-blocking code.

Additionally, supporting the same (or more) throughput with fewer connections has the advantage of having less memory consumption and being able to support a greater number of clients.

Regarding the relation of this ticket with TINKERPOP-1775, I think we have 2 options:
a) We see this change as an incremental improvement (which definitely is) without considering TINKERPOP-1775.
b) We try to think now how TINKERPOP-1774 and TINKERPOP-1775 will come into place together, regardless of which we implement first. I think the ultimate solution would be for the pool to balance request between connections, and without taking them out of the pool each time a request/response is issued.

I think we should take the approach in B, otherwise implementing it without thinking of TINKERPOP-1775 might introduce unnecessary complexity (i.e: a queue of pools that get enqueued and dequeued all the time).

FlorianHockmann:
>I think the ultimate solution would be for the pool to balance request between connections, and without taking them out of the pool each time a request/response is issued.

Why do you want to keep the connections in the pool when a read/write operation is executed on them? They can't be used at that time anyway and I think that the current design is a lot easier where a connection is taken from the pool for the request and returned simply by disposing its `ProxyConnection`.
I think it should be possible to implement such a request pipelining without changing the connection pool completely. The connections would only be returned faster. Balancing requests between connections should also not be necessary as the FIFO nature of the pool should already provide that without any changes.

Another aspect we should consider here is that this PR includes breaking changes as it adds a limit to the number of connections that can be used. That is why I would really like this to land in 3.4.0. TINKERPOP-1775 on the other hand can be implemented any time. So, there is no rush to implement it right now and I'm not sure whether I will have the time to also implement that in time for 3.4.0 as I will be OOO for ~2 weeks starting Wednesday and the 3.4.0 release date shouldn't be that far away.
Long story short, I don't think that this PR should be blocked by TINKERPOP-1775 if that means that this PR will have to wait for TinkerPop 3.5.0 or even 4.0.0. (If you - or someone else - find the time to tackle TINKERPOP-1775, then that's of course a different matter.)

jorgebay:
> Why do you want to keep the connections in the pool when a read/write operation is executed on them? They can't be used at that time anyway and I think that the current design is a lot easier where a connection is taken from the pool for the request and returned simply by disposing its `ProxyConnection`.

That's exactly why, because the connection **can** be used concurrently for requests.

Imagine the user wants to submit 10 queries, the first of them will take the server 100ms to send the response and the other nine will take 5ms. If you allow the connection to send the 10 requests without waiting for the response, the latency for the nine requests will be ~5ms. With the current approach, its ~105ms.

> I think it should be possible to implement such a request pipelining without changing the connection pool completely.

I disagree. There is no point in having a `ProxyConnection` and a pool that dequeues when a connection can be used concurrently.

jorgebay:
> What I meant here is that the connection cannot be used again until `SendAsync` completed. So, my suggestion for request pipelining is simply that the connection is taken out of the pool and then returned back as soon as `SendAsync` was successfully awaited.

From an API perspective, having a `Connection` method `SendAsync()` that yields the response once its received makes perfect sense. From the internals, I think we must look it in a different way:

- The sending process is just putting it into the write queue.
- For receiving, there should always be an outstanding call to `ws.ReceiveMessageAsync()`, once the response is parsed, find the callback in the in-flight requests (by request id) and invoke it.

So, on the implementation side, there is no need to await for the sending process to be finished, it just a new item in the queue. In the case the sending failed (e.g., the connection gets closed), the mechanism is the same: find the callback and invoke it.

> I just don't see what the advantage is to keep the connections in the pool while they are in use

There shouldn't be "in use" / "idle" states of a connection. The write queue is always sending (while there are items in the queue) and there is always an outstanding call to receive from the ws.

FlorianHockmann:
>>What I meant here is that the connection cannot be used again until SendAsync completed. So, my suggestion for request pipelining is simply that the connection is taken out of the pool and then returned back as soon as SendAsync was successfully awaited.

> From an API perspective, having a Connection method SendAsync() that yields the response once its received makes perfect sense. From the internals, I think we must look it in a different way

I think we are talking about different things here. I meant that the `ClientWebSocketConnection` should be taken out of the pool when `SendAsync` is called on it until it's awaited. Not that it should be taken out until the response was also received. Although this might require further changes as we currently only manage `Connections` with the `ConnectionPool` which already abstracts sending and receiving away into the common method `SubmitAsync`.

With the implementation I currently have in mind, it would look something like this:

1. A request comes in
2. A connection is taken out of the pool
3. `SendAsync()` is called on that connection and awaited
4. The connection is returned into the pool (and can therefore be used for the next `SendAsync()`)
5. The request finishes as soon as the response is received (like you described with continuous calls to `ReceiveMessageAsync()` and the invoke of the right callback).

That way, we don't need a dedicated write queue and most of the connection pool can stay as (relatively) simple as it currently is.

> There shouldn't be "in use" / "idle" states of a connection.

How do you want to avoid an _in use_ state when the connection cannot be used while it is _in use_, meaning that it is currently sending a message? Your supposed write queue would also need connections to actually send the requests. So, it would also need a similar connection pool.

Should we maybe move this discussion somewhere else, like to [our HipChat room](https://apache.hipchat.com/chat/room/2013155)?

jorgebay:
I think we should abandon the idea about removing connections from a pool and the `ProxyConnection`, there is no point and it's not a good pattern.

All connections in the pool are available and it's up to the pool to balance the load (incoming send requests) between the connections not restrict concurrent operations, at least for now.

> the connection cannot be used while it is in use

That is incorrect, the underlying transport (websocket) supports just 1 sending at a time (not 1 sending and 1 receive at a time). So we limit concurrency **on the connection**, by sending serially all incoming send requests. **We should never use 1 send followed by 1 receive call**, we should send items serially and always be receiving in the background.

FlorianHockmann:
Ok, I think I finally got your proposition. Took me way longer than it should have to be honest 😅 

The changes would basically look like this if I got everything right:
* `Connection`:
  * Gets its own write queue
  * New requests are simply appended to this queue.
  * Sends messages from its queue as fast as possible, one `SendAsync()` after another
  * Maintains a lookup table: `requestId` --> response callback
  * Continuously `awaits` `ReceiveAsync()` and invokes the right callback for every received response based on the `requestId`
* `ConnectionPool`:
  * Keeps all created `Connections` in its pool (they aren't taken out of the pool completely while in use any more)
  * Either returns a random `Connection` or returns the least-used `Connection` based on an in-flight counter on the `Connection` (the latter would be the better solution but is a bit more complex).

Is that the design you are proposing or did I still get something wrong?

I guess what kept me the entire time from understanding this is that I for some reason assumed that messages could be received on a different WebSocket connection than where the send was performed. That would have allowed for an easier approach without the write queue.

This brings us now back to your initial question of whether we want to merge the changes from this PR first or whether we want to implement it together with the request pipelining we now discussed in great length.

@spmallette: Could you say something about the planned release date for TinkerPop 3.4.0? The last statement I remember was something like _end of summer_.

I could probably try to implement the request pipelining in October, but when code freeze is before that or in the beginning of October, then I'd say that we should definitely merge this PR first. Otherwise, it would have to wait for TinkerPop 3.5.0.

Your (@jorgebay) argument for implementing both tickets together was:

> I think we should take the approach in B, otherwise implementing it without thinking of TINKERPOP-1775 might introduce unnecessary complexity (i.e: a queue of connections that get enqueued and dequeued for each request).

but this enqueueing and dequeueing of connections is already present in the current implementation.
This PR basically adds these points:
* Min and max sizes 
* [CAS pattern](https://en.wikipedia.org/wiki/Compare-and-swap) to ensure that the max size is respected
* An `AsyncAutoResetEvent` to replace the synchronous lock

I think those three aspects would still be needed with request pipelining. So, in my opinion, this PR is really an incremental improvement that doesn't add much complexity that we might need to remove again in order to implement request pipelining.

jorgebay:
Yes, that's the design I was proposing, a minor note:

> Either returns a random `Connection` or returns the least-used `Connection` based on an in-flight counter on the `Connection` (the latter would be the better solution but is a bit more complex).

The easiest solution is to round robin through connections, we can use a "modulo counter":

```csharp
// Its important to get a reference of the connection array that doesn't change
// Otherwise the connection length might be modified (see copy-on-write)
Connection[] connections = GetConnections();
var connectionIndex = Interlocked.Increment(ref _indexCounter) % connections.Length;
var connectionToReturn = connections[connectionIndex];
// TODO: Overflow protection for `_indexCounter`
```

Regarding whether to implement both approaches, I think implementing TINKERPOP-1774 with the current implementation of the pool (dequeuing) comes at a cost of introducing complexity (in particular `AsyncAutoResetEvent`).

The implementation for this patch is really good, but we can be introducing corner cases and I think it's not worth the risk. I'd say let's wait for both solutions to be implemented.

FlorianHockmann:
> I think implementing TINKERPOP-1774 with the current implementation of the pool (dequeuing) comes at a cost of introducing complexity (in particular AsyncAutoResetEvent).

I guess that depends on whether we want to implement TINKERPOP-1775 with our without a `max in-flight` requests per connection as that would result in some connections having to wait and therefore an `AsyncAutoResetEvent` would be helpful again.
Without such a limit, I wonder when we would create new connections at all. When we just round-robin incoming requests to connections in the pool, then when should a new connection be created?

spmallette:
@FlorianHockmann i don't personally feel like calling for a release this month as there are things i think i'd still like to see completed, so i think we're at a point of evaluating month to month at this point as to when a release date should be proposed. 3.4.0 is the hold up for me....if we wanted to release some fixes we could probably do 3.2.10 and 3.3.4 and hold 3.4.0 for later. we do have a lot of bug fixes that could go up.  either way, 3.4.0 isn't ready yet imo.

jorgebay:
Max in-flight should be a threshold that, when reached, more connections are created.

When max amount of connections is reached, we can either:
a) Continue queueing at connection level (adding it to the queue).
b) Throttle requests by throwing an error, like `MaxLimitExceededException` or something like that.

From my experience, I found that b is far more clear.

In any case, a) and b) don't require additional queueing at pool level (ie: `AsyncAutoResetEvent`).

FlorianHockmann:
Thanks for the insight @spmallette. That means for this PR that we don't have to hurry at the moment.

> Max in-flight should be a threshold that, when reached, more connections are created.

So, max in-flight would be a global threshold? I think it would be hard to set a meaningful limit for that as it depends on the maximum pool pool size.
The Java driver has a `maxSimultaneousUsagePerConnection` and creates a new connection when the least used connection has reached that limit (until the max pool size is reached).

> When max amount of connections is reached, we can either:
a) Continue queueing at connection level (adding it to the queue).
b) Throttle requests by throwing an error, like MaxLimitExceededException or something like that.

The Java driver waits for an available connection in that case which needs to have less than the `maxSimultaneousUsagePerConnection` requests in progress.
That approach makes sense to me and I don't really like throwing an exception only because the limit is reached because of a high number of requests that might only be a temporary problem.

If we want to follow a similar approach like that of the Java driver, then we still need to wait for an available connection which would also require such an `AsyncAutoResetEvent`.

FlorianHockmann:
> looks like you are focused on getting this one merged

No, I don't really mind waiting for TINKERPOP-1775 and implementing both together. I just had two motivations for merging this PR first independently of TINKERPOP-1775:

1. Adding limits to the `ConnectionPool` is a breaking change and therefore can only go into version 3.4.0 or 3.5.0.
2. From my point of view, the changes from this PR would also be necessary to implement request pipelining.

Point 1 introduced some pressure of time for this PR as I don't want to wait for version 3.5.0. However, that isn't valid any more when we don't plan to release 3.4.0 in the next few weeks.
Point 2 was mainly motivated by the fact that the Java `ConnectionPool` has basically the same implementation as introduced by this PR and I saw that as something like the reference implementation.

I know that you have way more experience with writing drivers and you made some good points which convince me more and more that the .NET driver could look different than the Java driver when we implement both tickets together.

So, I agree and let's wait until we have a PR for both tickets. It should also be easier to continue the discussion on the concrete implementation for the request pipelining once we have a first implementation.

papa-pep:
Is this slated for 3.4.0?  Periodically I'm seeing requests being held open for unusually long times until an exception is thrown: 

System.Net.Sockets.SocketException      An internal WebSocket error occurred. Please see the innerException, if present, for more details. Unable to read data from the transport connection:   An existing connection was forcibly closed by the remote host. An existing  connection was forcibly closed by the remote host

Do you think this would help alleviate that issue?  I'm currently using 3.4.0rc2 in order to return necessary request attributes from CosmosDb to monitor performance.

FlorianHockmann:
Yes, this is planned for 3.4.0. It just takes a bit longer as we decided to implement this together with request pipelining ([TINKERPOP-1775](https://issues.apache.org/jira/browse/TINKERPOP-1775)).

However, I'm not sure if it resolves the issue you're describing. Your problem sounds like a connection was closed because it was idle for some time and then the driver tried to use it again.
These changes should reduce the number of connections that are opened and especially the number of connections that are idle for a long time, but there will still be situations where connections are idle. I assumed that pings should prevent this from happening. Do you know whether CosmosDB supports pings for its Gremlin API?

Could you also please provide a full stack trace? It would especially be interesting to see where in Gremlin.Net the exception originated.

FlorianHockmann:
The problem was also just reported by two other users of Cosmos DB in this issue:
https://issues.apache.org/jira/browse/TINKERPOP-2090
So, I suppose that we move the discussion over to that ticket.

papa-pep:
I'll check out that ticket, thanks.

FYI - Here is the error information and portion of the stack trace beginning with Gremlin.Net.  What is generally happening is that a request is made to cosmosDb and 1-10 minutes later this error is returned, even for very simple traversals such a g.V(id).

Message:
An internal WebSocket error occurred. Please see the innerException, if present, for more details. Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host. An existing connection was forcibly closed by the remote host 

Exception Type:
System.Net.Sockets.SocketException

Failed Method:
Gremlin.Net.Driver.WebSocketConnection+<ReceiveMessageAsync>d__8.MoveNext

Callstack:
System.Net.WebSockets.WebSocketException: at System.Net.WebSockets.WebSocketBase.ThrowIfConvertibleException (System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089) at System.Net.WebSockets.WebSocketBase+<ReceiveAsyncCore>d__45.MoveNext (System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089) at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089) at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089) at Gremlin.Net.Driver.WebSocketConnection+<ReceiveMessageAsync>d__8.MoveNext (Gremlin.Net, Version=3.4.0.0, Culture=neutral, PublicKeyToken=d2035e9aa387a711) at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089) at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089) at Gremlin.Net.Driver.Connection+<ReceiveAsync>d__14`1.MoveNext (Gremlin.Net, Version=3.4.0.0, Culture=neutral, PublicKeyToken=d2035e9aa387a711) at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089) at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089) at Gremlin.Net.Driver.Connection+<SubmitAsync>d__8`1.MoveNext (Gremlin.Net, Version=3.4.0.0, Culture=neutral, PublicKeyToken=d2035e9aa387a711) at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089) at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089) at Gremlin.Net.Driver.ProxyConnection+<SubmitAsync>d__3`1.MoveNext (Gremlin.Net, Version=3.4.0.0, Culture=neutral, PublicKeyToken=d2035e9aa387a711) at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089) at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089) at Gremlin.Net.Driver.GremlinClient+<SubmitAsync>d__6`1.MoveNext (Gremlin.Net, Version=3.4.0.0, Culture=neutral, PublicKeyToken=d2035e9aa387a711) at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089) at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089) at Gremlin.Net.Driver.GremlinClientExtensions+<SubmitAsync>d__4`1.MoveNext (Gremlin.Net, Version=3.4.0.0, Culture=neutral, PublicKeyToken=d2035e9aa387a711) at

FlorianHockmann:
Closed in favour of #1016 which includes the changes from this PR.

PR #904, opened 2018-08-02
TINKERPOP-2016 Bumped to Jackson 2.9.6

robertdale:
https://issues.apache.org/jira/browse/TINKERPOP-2016

Bumped to Jackson 2.9.6 for CVE.

All tests pass with `docker/build.sh -i -t -n`

The CI build errors are Grapes related. I built several times locally and through docker with no issues.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -44,6 +44,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Added concrete configuration methods to `SparkGraphComputer` to make a more clear API for configuring it.
 * Fixed a bug in `TinkerGraphCountStrategy`, which didn't consider that certain map steps may not emit an element.
 * Fixed a bug in JavaScript GLV where DriverRemoteConnection close() method didn't returned a Promise instance.
+* Bumped to Jackson 2.9.6.
 
 [[release-3-2-9]]
 === TinkerPop 3.2.9 (Release Date: May 8, 2018)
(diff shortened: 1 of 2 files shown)


spmallette:
Shouldn't this point at tp32?

robertdale:
@spmallette Thanks.

spmallette:

VOTE +1

FlorianHockmann:
I just searched whether the library would also be used somewhere else and found it [in `spark-gremlin`](https://github.com/apache/tinkerpop/blob/tp32/spark-gremlin/pom.xml#L211):
```
<dependency>
    <groupId>com.fasterxml.jackson.core</groupId>
    <artifactId>jackson-databind</artifactId>
    <version>2.4.4</version>
</dependency>
```
Shouldn't that also be updated or is there a reason why we have to use that specific version there?

spmallette:
Not sure what we can do about that. That's not our dependency. Someone would have to raise that issue in Spark/Hadoop or whatever is bringing that in. All we did was make their version consistent. I don't think we can arbitrarily force upgrade on our end.

FlorianHockmann:
Ah ok, I didn't see that. I guess I'm still not really used to the way Maven works 😅 

But in that case: VOTE +1

spmallette:
To be clear, I don't think we can force upgrade from 2.4.x to 2.9.x as it creates breaking changes in spark. Sadly, we don't even seem to have a route to resolve that for 3.4.0 because even with the Spark 2.3.x bump we still only get to jackson 2.6.x. Not sure what spark is doing with that.

FlorianHockmann:
At least they are working on it for a future release:
Ticket: https://issues.apache.org/jira/browse/SPARK-24601
PR: https://github.com/apache/spark/pull/21596

Although it sounds more as if it would only land in Spark 3:

> Users can (perhaps should) be shading Jackson but I bet most won't. This is why it's safe for Spark 3, risky for Spark 2.4.

[Comment](https://github.com/apache/spark/pull/21596#issuecomment-402521806)

PR #907, opened 2018-08-05
Fix Sack Split.

RainMark:
Why sack spliter use the initial sack?

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/traverser/B_O_S_SE_SL_Traverser.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/traverser/B_O_S_SE_SL_Traverser.java
@@ -103,14 +103,14 @@ public class B_O_S_SE_SL_Traverser<T> extends B_O_Traverser<T> {
     @Override
     public <R> Traverser.Admin<R> split(final R r, final Step<T, R> step) {
         final B_O_S_SE_SL_Traverser<R> clone = (B_O_S_SE_SL_Traverser<R>) super.split(r, step);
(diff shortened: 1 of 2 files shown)


robertdale:
Is there a test that covers this?

RainMark:
Nop,there is no test for sack split operator. 
If I pass in a special split operator, this operator can't obtain source traverser's sack value with apply() method.

robertdale:
@RainMark Could you create an issue in Jira and describe the problem?   See also [Contributing code changes](https://tinkerpop.apache.org/docs/current/dev/developer/#_contributing_code_changes). It would be good to see a test case that fails without the change and passes with the change but also satisfies the examples in the documentation.

I also want to make sure you're not running into this
- http://tinkerpop.apache.org/docs/current/reference/#sack-step
```
The first example does not use a split operator and as such, the same map is propagated to
all traversers (a global data structure). The second example, demonstrates how Map.clone()
ensures that each traverser’s sack contains a unique, local sack.
```

spmallette:
@RainMark did you have a chance to consider the previous comment from @robertdale ?  While I appreciate that you have a solution to a problem you are seeing, I share his concern about understanding the nature of the problem itself that is being fixed.

spmallette:
hi @RainMark do you plan to come back to address the comments here?

RainMark:
@spmallette yes,I will come back to address the comments. just a little busy recently

RainMark:
@spmallette @robertdale 

The sack of clone traverser is correctly.
Sorry, I'm new for java, I never noticed Java's object clone method.
Recently, We develop a gremlin query engine based on c++, So I ask this question.
Anyway,Thank you very much!

spmallette:
> We develop a gremlin query engine based on c++,

interesting....Is that code available as open source?

PR #908, opened 2018-08-06
TINKERPOP-2017 Add check for Column in `by(Function)`

yura-ata:
https://issues.apache.org/jira/browse/TINKERPOP-2017

Minor change to check for `Column` in `by(Function)`
Excuse me if I've missed something: test (it was difficult to come up with a test for this because only internal representation of traversal was changed), changelog update, etc.) - would appreciate any guidance :)

Nevertheless, `mvn clean install` passes


spmallette:
I think i'm ok with how you've done this. There are process suite tests that hit both the original and the new code path and they pass, so i think you're covered by what was already there. 

However, do you mind editing this pull request to target the tp32 branch? this change can be applied there can't it?

spmallette:
VOTE +1

twilmes:
Looks good @yura-ata. Did you re-target `master` again for a particular reason?

VOTE: +1

spmallette:
Cherry picked this commit to tp32 - a0a243b256dbfdf66bb5ce24958fc91c7e6eed03

PR #911, opened 2018-08-13
TINKERPOP-2024 Make server archetype use remote traversal

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2024

Since we're promoting remote traversals over scripts it would be better for the archetype to use them.

VOTE +1


FlorianHockmann:
Thanks for implementing this so quickly, @spmallette 

LGTM, VOTE +1

dkuppitz:
VOTE +1

PR #912, opened 2018-08-13
TINKERPOP-2023 SSL Enhancements

robertdale:
https://issues.apache.org/jira/browse/TINKERPOP-2023

Gremlin Server no longer supports automatically creating self-signed certificates.

Cluster client no longer trusts all certs by default as this is an insecure configuration. (TINKERPOP-2022)
If no trust store is configured, Cluster will use the default CA certs.
To revert to the previous behavior and accept all certs, it must be explicitly configured.

Introduces JKS and PKCS12 support. JKS is the legacy Java Key Store. PKCS12 has better cross-platform support and is gaining in adoption. Be aware that JKS is the default on Java 8.  Java 9 and higher use PKCS12 as the default. Both Java keytool and OpenSSL tools can create, read, update PKCS12 files.

Other new features include specifying SSL protocols and cipher suites.

The packaged `*-secure.yaml` files now restrict the protocol to `TLSv1.2` by default.

The implication of all of the above changes means that the packaged `*-secure.yaml` files no longer "just work". Minimally, the server files must be configured with a key/cert.

PEM-based configurations are deprecated, to be removed in a future release.

`mvn clean install -DskipIntegrationTests=false  -pl :gremlin-server` passes all tests

VOTE +1


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
```diff
@@ -579,7 +657,9 @@ public Builder sslContext(final SslContext sslContext) {
          * File location for a SSL Certificate Chain to use when SSL is enabled. If this value is not provided and
          * SSL is enabled, the {@link TrustManager} will be established with a self-signed certificate which is NOT
          * suitable for production purposes.
+         * @deprecated
```
spmallette: Please include the full standard "deprecation message" wherever `@Deprecated` is used: 

`@deprecated As of release 3.2.10, replaced by blah blah blah"

docs/src/reference/gremlin-applications.asciidoc
```diff
@@ -732,7 +732,10 @@ The following table describes the various configuration options for the Gremlin
 |connectionPool.keepAliveInterval |Length of time in milliseconds to wait on an idle connection before sending a keep-alive request. Set to zero to disable this feature. |1800000
 |connectionPool.keyCertChainFile |The X.509 certificate chain file in PEM format. |_none_
```
spmallette: since some of these settings are deprecated, I think that we should just remove them from the documentation. we usually don't document deprecated concepts.

spmallette:
Nice PR description and good upgrade docs. I was reading through the current docs on "Security" and they seem insufficient:

http://tinkerpop.apache.org/docs/current/reference/#security

since you seem to have a good handle on some many of these SSL details, could you perhaps fix that section up a bit? Since we can no longer do `enableSsl` and get that self-signed cert, i think it would be helpful to users to have some additional/explicit instruction for getting SSL working. Thoughts?

FlorianHockmann:
Since this is a breaking change, wouldn't it be better to target `master` so that this can go into 3.4.0?

spmallette:
I wondered that too....i figured the reasoning for tp32 was that this was a "security" issue and thus the "breaking" aspect was justified. is that a fair way to characterize this the branch targetting @robertdale ?

robertdale:
Right, when it's security-related, you're given a pass to break anything! :rage1: :rage2:  :rage3: :rage4: 

Seriously, the breaking change is only if you relied on the server generating self-signed certs or if the client accepted all certs. In the server case, you have to configure SSL correctly.  In the client case, you can revert to previous behavior with a one line change.  In both cases, I think they are trivial enough configuration changes that warrant going back to tp32 to set things right.

spmallette:
>  if you relied on the server generating self-signed certs

i think it's fine to break those folks because they shouldn't have been doing that in the first place.

FlorianHockmann:
> They settings are @Deprecated not removed - for example:

Ah, thanks for the clarification. I indeed only looked at the changes to the docs. Deprecating those settings is of course a different thing.
Now I understand Robert's conclusion that only settings are broken that relied on the server to generate a self-signed cert.

k4rthikr:
@robertdale Should we also provide a config entry for disabling certain cipher suites? For example TLS_RSA_WITH_3DES_EDE_CBC_SHA is supported. This cipher suite has a 64-bit block size, and the security properties of this construction do not hold when more than 32 GB of data are sent over a single TLS connection. Thus, this cipher suite is not suitable for long-lived TLS connections, such as those opened to a database. In fact, it wouldn't hurt to just programmatically remove these ciphers.

robertdale:
@k4rthikr That's a little deeper than I want to get into. `sslCipherSuites` provides a whitelist. Whitelists are always preferred over blacklists. If this one were blacklisted, there are probably 10 others that were missed. I also don't want to imply that these are managed and updated by us. There are just too many and changes more often than say TLS protocol version. IMO, the scope of this work was to set reasonable defaults that made Gremlin Server+Client more secure and raise awareness to the user. It will be up to the user to use good ciphers, good key lengths, etc.

spmallette:
VOTE +1 - server integration tests are good - nice job with docs and everything. thanks for doing this nerd work

k4rthikr:
I agree. Thanks for the change!

dkuppitz:
Very nice work.

VOTE: +1

PR #868, opened 2018-05-14
Tinkerpop 1913

ashwini-ms:
Adding support to de-serialize status attribute for gremlin dotnet. 

More details : https://issues.apache.org/jira/browse/TINKERPOP-1913

PR #913, opened 2018-08-14
Add support for rate-limited request retries when utilizing cosmosDb

papa-pep:
In order to support better application stability in high stability scenarios, it is necessary to implement rate-limited retries as defined by the cosmosDb documentation. 
(https://docs.microsoft.com/en-us/azure/cosmos-db/request-units#RequestRateTooLarge)
It would be a nice enhancement in the future if additional performance related telemetry could be exposed on a per-traversal basis, but those changes would have a larger impact on the framework.

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/AttributeKeys.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/AttributeKeys.cs
@@ -1,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Gremlin.Net.Driver
+{
+    internal static class AttributeKeys
+    {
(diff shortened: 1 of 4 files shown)


spmallette:
Thanks for taking the time to submit this pull request, but I don't think we can accept it. TinkerPop is vendor agnostic and this change seems highly specific to CosmosDB and its operations. While I can see why this is a useful change for CosmosDB users, I'm not sure that it belongs here. Perhaps it is an extension to CosmosDB libraries or maybe there is something that needs to change here in TinkerPop to allow such capability in a CosmosDB library?

I thought that #868 was related to this "rate limiting" issue - perhaps from a TinkerPop perspective the solution is already underway there? thoughts?

PR #914, opened 2018-08-14
Do not format and reparse eval timeout arg when evaluating request.

newkek:


newkek:
I've pushed a fix for tests that used to send the Eval_timeout as an int instead of a long, but also made the server robust in case an int is still sent. Not sure if we want to do that or to enforce the `long` by failing.

spmallette:
Integration tests are working good - code changes look fine: VOTE +1

robertdale:
VOTE +1

PR #919, opened 2018-08-22
String loop to String builder [tp32]

otaviojava:
String concatenation in loops. As every String concatenation copies the whole String, usually it is preferable to replace it with explicit calls to StringBuilder.append() or StringBuffer.append().

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversalMetrics.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversalMetrics.java
@@ -109,14 +109,14 @@ public final class DefaultTraversalMetrics implements TraversalMetrics, Serializ
     private void appendMetrics(final Collection<? extends Metrics> metrics, final StringBuilder sb, final int indent) {
         // Append each StepMetric's row. indexToLabelMap values are ordered by index.
         for (Metrics m : metrics) {
-            String rowName = m.getName();
+            StringBuilder rowName = new StringBuilder(m.getName());
 
             // Handle indentation
             for (int ii = 0; ii < indent; ii++) {
-                rowName = "  " + rowName;
+                rowName.insert(0, "  ");
             }
             // Abbreviate if necessary
-            rowName = StringUtils.abbreviate(rowName, 50);
+            rowName = new StringBuilder(StringUtils.abbreviate(rowName.toString(), 50));
 
             // Grab the values
             final Long itemCount = m.getCount(ELEMENT_COUNT_ID);
@@ -125,7 +125,7 @@ public final class DefaultTraversalMetrics implements TraversalMetrics, Serializ
 
             // Build the row string
 
-            sb.append(String.format("%n%-50s", rowName));
+            sb.append(String.format("%n%-50s", rowName.toString()));
 
             if (itemCount != null) {
                 sb.append(String.format(" %21d", itemCount));
@@ -187,11 +187,11 @@ public final class DefaultTraversalMetrics implements TraversalMetrics, Serializ
     private static String padLeft(final String text, final int amountToPad) {
         // not sure why this method needed to exist. stupid string format stuff and commons utilities wouldn't
         // work for some reason in the context this method was used above.
-        String newText = text;
+        StringBuilder newText = new StringBuilder(text);
         for (int ix = 0; ix < amountToPad; ix++) {
-            newText = " " + newText;
+            newText.insert(0, " ");
         }
-        return newText;
+        return newText.toString();
     }
 
     private void computeTotals() {
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversalMetrics.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversalMetrics.java
@@ -109,12 +109,13 @@ public final class DefaultTraversalMetrics implements TraversalMetrics, Serializ
     private void appendMetrics(final Collection<? extends Metrics> metrics, final StringBuilder sb, final int indent) {
         // Append each StepMetric's row. indexToLabelMap values are ordered by index.
         for (Metrics m : metrics) {
-            final StringBuilder metricName = new StringBuilder(m.getName());
+            final StringBuilder metricName = new StringBuilder();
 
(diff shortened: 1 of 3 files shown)


robertdale:
@otaviojava did you decompile the bytecode to make sure the compiler didn't do that automagically?   Otherwise, StringBuilder vars should be final.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversalMetrics.java
```diff
@@ -208,14 +208,14 @@ private void handleNestedTraversals(final Traversal.Admin traversal, final Mutab
...
         // Append each StepMetric's row. indexToLabelMap values are ordered by index.
         for (Metrics m : metrics) {
-            String rowName = m.getName();
+            final StringBuilder metricName = new StringBuilder(m.getName());
 
             // Handle indentation
             for (int ii = 0; ii < indent; ii++) {
-                rowName = "  " + rowName;
+                metricName.insert(0, "  ");
             }
```
robertdale: Would it make sense to put this block before the name then it won't have to do inserts?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversalMetrics.java
```diff
@@ -286,10 +286,10 @@ private void appendMetrics(final Collection<? extends Metrics> metrics, final St
     private static String padLeft(final String text, final int amountToPad) {
         // not sure why this method needed to exist. stupid string format stuff and commons utilities wouldn't
         // work for some reason in the context this method was used above.
-        String newText = text;
+        final StringBuilder newText = new StringBuilder(text);
         for (int ix = 0; ix < amountToPad; ix++) {
-            newText = " " + newText;
+            newText.insert(0, " ");
         }
```
robertdale: Would it make sense to put this block before `text` then it won't have to do inserts?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversalMetrics.java
otaviojava: That is the same previous behavior, but using StringBuilder this time.
https://docs.oracle.com/javase/8/docs/api/java/lang/StringBuilder.html#insert-int-java.lang.String-

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversalMetrics.java
otaviojava: That is the same previous behavior, but using StringBuilder this time.
https://docs.oracle.com/javase/8/docs/api/java/lang/StringBuilder.html#insert-int-java.lang.String-

otaviojava:
@robertdale 
I think so, after the merge in the master, I can create a second one to bot tp32 and tp33.
But, I don't know. That is my first contribution to the project.

spmallette:
thanks for the contribution @otaviojava - please edit the PR to target tp32. we handle the rest of the merge flow to tp33 and master from there unless there are really ugly merges or something between release branches.

otaviojava:
@spmallette done.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversalMetrics.java
robertdale: Yes, I get that. What I'm suggesting is that instead of having to `insert` data later on thus incurring the cost of array copies, why not put the indentation loop first, then everything after can be `append` only.

otaviojava: Do you mean this way?
```java
  private static String padLeft(final String text, final int amountToPad) {
        // not sure why this method needed to exist. stupid string format stuff and commons utilities wouldn't
        // work for some reason in the context this method was used above.
        final StringBuilder newText = new StringBuilder();
        for (int ix = 0; ix < amountToPad; ix++) {
            newText.append(" ");
        }
        newText.append(text);
        return newText.toString();
    }

```

robertdale: Right.

otaviojava: Done in both, thank you.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversalMetrics.java
otaviojava: Done, thank you.

otaviojava:
@spmallette @robertdale When I changed the branch to tp32. When execute this code:
`return new ASTTransformationCustomizer(InterpreterMode.class);` It returns an NPE, I checked the code.

```java
private static CompilePhase findPhase(ASTTransformation transformation) {
        if (transformation==null) throw new IllegalArgumentException("Provided transformation must not be null")
        final Class<?> clazz = transformation.class//this line returns a NPE
        final GroovyASTTransformation annotation = clazz.getAnnotation(GroovyASTTransformation)
        if (annotation==null) throw new IllegalArgumentException("Provided ast transformation is not annotated with "+GroovyASTTransformation.name)

        annotation.phase()
    }
```
The file: ASTTransformationCustomizer.groovy

I did not change the code at this point.
Suggestions?

robertdale:
@otaviojava was that locally?  the travis build failed due to grapes download.

robertdale:
Hmm.. it happens on tp32.

robertdale:
Nevermind, still just grapes.

dkuppitz:
In the future, can we please let those (potential) CTR's sit for a bit longer? I haven't had a chance to really look at it and I guess I would have voted against it (it's not bad, but also not perfect in my opinion).

So I'm not asking to revert it, just for a little more patience (at least for outside PRs).

robertdale:
@dkuppitz Will do.

PR #920, opened 2018-08-22
optmizes collection copy with Collections addAll

otaviojava:
Copying of array contents to a collection where each element is added individually using a for a loop. We can be replaced by a call to Collections.addAll().
As the documentation says:

> Adds all of the specified elements to the specified collection. Elements to be added may be specified individually or as an array. The behavior of this convenience method is identical to that of c.addAll(Arrays.asList(elements)), but this method is likely to run significantly faster under most implementations.
When elements are specified individually, this method provides a convenient way to add a few elements to an existing collection:

https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#addAll-java.util.Collection-T...-

In my experience using either Collections.addll or Collections.addAll used to be 30% faster then add from loop.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
@@ -1480,22 +1480,39 @@ public interface GraphTraversal<S, E> extends Traversal<S, E> {
      * @since 3.2.2
      */
     public default GraphTraversal<S, E> hasId(final Object id, final Object... otherIds) {
-        if (id instanceof P)
+        if (id instanceof P) {
             return this.hasId((P) id);
+        }
         else {
-            final List<Object> ids = new ArrayList<>();
+            Object[] ids;
             if (id instanceof Object[]) {
-                Collections.addAll(ids, (Object[]) id);
-            } else
-                ids.add(id);
+                ids = (Object[]) id;
+            } else {
+                ids = new Object[] {id};
+            }
+            int size = ids.length;
+            int capacity = size;
             for (final Object i : otherIds) {
                 if (i.getClass().isArray()) {
-                    Collections.addAll(ids, (Object[]) i);
-                } else
-                    ids.add(i);
+                    final Object[] tmp = (Object[]) i;
+                    int newLength = size + tmp.length;
+                    if (capacity < newLength) {
+                        ids = Arrays.copyOf(ids, capacity = size + tmp.length);
+                    }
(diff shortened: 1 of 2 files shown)


dkuppitz:
I can see why `Collections.addAll(...)` would be faster than `c.addAll(Arrays.asList(...))`, but not why it would be faster than a simple loop. I think if you really want to squeeze out as much as possible, you should prevent the use of lists altogether (ultimately the list is converted back to an array, so there's really no point of dealing with lists there, other than making the code more readable).

Only having arrays could look like this:

```java
default GraphTraversal<S, E> hasId(final Object id, final Object... otherIds) {
    if (id instanceof P)
        return this.hasId((P) id);
    else {
        Object[] ids;
        if (id instanceof Object[]) {
            ids = (Object[]) id;
        } else {
            ids = new Object[] {id};
        }
        int size = ids.length;
        int capacity = size;
        for (final Object i : otherIds) {
            if (i.getClass().isArray()) {
                final Object[] tmp = (Object[]) i;
                int newLength = size + tmp.length;
                if (capacity < newLength) {
                    ids = Arrays.copyOf(ids, capacity = size + tmp.length);
                }
                System.arraycopy(tmp, 0, ids, size, tmp.length);
                size = newLength;
            } else {
                if (capacity == size) {
                    ids = Arrays.copyOf(ids, capacity = size * 2);
                }
                ids[size++] = i;
            }
        }
        if (capacity > size) {
            ids = Arrays.copyOf(ids, size);
        }
        this.asAdmin().getBytecode().addStep(Symbols.hasId, ids);
        return TraversalHelper.addHasContainer(this.asAdmin(), new HasContainer(T.id.getAccessor(), ids.length == 1 ? P.eq(ids[0]) : P.within(ids)));
    }
}
```

This is untested and thus not benchmarked, I'm just guessing that it should perform better than the current implementation.

spmallette:
hi @otaviojava - do you plan to come back to this pull request to consider the comment that @dkuppitz has made here?

otaviojava:
hey @spmallette yes, but just next week

spmallette:
@otaviojava just wanted to point out that we are preparing for release (3.2.10 and 3.3.4) in the next couple of weeks. as this change target tp32 and thus those releases. I just wanted to make you aware of the release dates in case you wanted to see this merged for those releases.

otaviojava:
@spmallette done, sorry by the delay.
@dkuppitz thank you for the support

robertdale:
VOTE +1

dkuppitz:
I haven't done any extensive performance benchmarks, but the overall build time went down by 1 minute -- wohoo :)

Anyway, `docker/build.sh -t -i` passed.

VOTE +1

spmallette:
VOTE +1

PR #921, opened 2018-08-24
Add self loop example to Cycle Detection recipe.

newkek:
Not sure if it is worth it but I figured out that self loops weren't detected in the given example.

### DIFF
--- a/docs/src/recipes/cycle-detection.asciidoc
+++ b/docs/src/recipes/cycle-detection.asciidoc
@@ -48,9 +48,36 @@ the length of the cycle is known to be three and there is no need to exceed that
 cycle. It returned three, because there was one for each vertex that started the cycle (i.e. one for `A`, one for `B`
 and one for `C`). This next line introduce deduplication to only return unique cycles.
 
-The above case assumed that the need was to only detect cycles over a path length of three. It also respected the
-directionality of the edges by only considering outgoing ones. What would need to change to detect cycles of
-arbitrary length over both incoming and outgoing edges in the modern graph?
+The above case assumed that the need was to only detect cycles over a path length of three.
+It also respected the directionality of the edges by only considering outgoing ones.
+
+Also note that the traversal above won't detect self-loops (vertices directly connected to
+themselves). To do so, you would need to `.emit()` a Traverser before the repeat()-loop.
+
+[gremlin-groovy]
+----
+g.addV().property(id,'a').as('a').
+  addV().property(id,'b').as('b').
+  addV().property(id,'c').as('c').
+  addV().property(id,'d').as('d').
+  addE('knows').from('a').to('b').
+  addE('knows').from('b').to('c').
+  addE('knows').from('c').to('a').
+  addE('knows').from('a').to('d').
+  addE('knows').from('c').to('d').
+  addE('self').from('a').to('a').iterate()
+g.V().as('a').
+  emit().
+    repeat(outE().inV().simplePath()).
+    times(2).
+  outE().inV().where(eq('a')).
+  path().
+    by(id).
+    by(label)
+----
+
+What would need to change to detect cycles of arbitrary length over both incoming and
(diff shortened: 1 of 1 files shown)


docs/src/recipes/cycle-detection.asciidoc
```diff
@@ -48,6 +48,25 @@ the length of the cycle is known to be three and there is no need to exceed that
...
+  addV().property(id,'c').as('c').
+  addV().property(id,'d').as('d').
+  addE('knows').from('a').to('b').
+  addE('knows').from('b').to('c').
+  addE('knows').from('c').to('a').
+  addE('knows').from('a').to('d').
+  addE('knows').from('c').to('d').
+  addE('self').from('a').to('a').iterate()
+g.V().as('a').emit().repeat(out().simplePath()).times(2).
+  where(out().as('a')).path()
```
dkuppitz: I would tweaks the traversal a little bit to make the output easier to grasp:
```
gremlin> g.V().as('a').
......1>   emit().
......2>     repeat(outE().inV().simplePath()).
......3>     times(2).
......4>   outE().inV().where(eq('a')).
......5>   path().
......6>     by(id).
......7>     by(label)
==>[a,self,a]
==>[a,knows,b,knows,c,knows,a]
==>[b,knows,c,knows,a,knows,b]
==>[c,knows,a,knows,b,knows,c]
```

newkek: Good point!

newkek: @dkuppitz what is the added value of `outE().inV()` compared to `out()`?

dkuppitz: Really just readability. To me `[a,self,a]` looks better than just `[a,a]`.

docs/src/recipes/cycle-detection.asciidoc
```diff
@@ -48,6 +48,31 @@ the length of the cycle is known to be three and there is no need to exceed that
...
+g.V().as('a').
+    emit().
+        repeat(outE().inV().simplePath()).
+        times(2).
+    outE().inV().where(eq('a')).
+    path().
+      by(id).
+      by(label)
+----
+
```
dkuppitz: Cool, much better. What I didn't notice in the first place, was that the paragraph below really only refers to the previous code snippets. I think you should move it above you changes.

newkek: Hm the issue with the Traversal below is that it improves things further by formatting results more and doing unidirectional traversing on an unrestricted path length. Since the examples get more complicated as you progress through the document, if I put this self loop traversal under the more complicated one it feels like going backwards. Additionally the traversal below refers to the modern graph, whereas the first one and this self-loop one are similar, wdyt?

dkuppitz: I mean just put the paragraph that starts with
> The above case assumed that ...

above your new block that starts with
> Note that these ...

That would make more sense, no?

newkek:
```
The above case assumed that the need was to only detect cycles over a path length of three.
It also respected the directionality of the edges by only considering outgoing ones.

Also note that this traversal won't detect self-loops [....]
--
[self-loop code example]
--

What would need to change to detect cycles of arbitrary length over both incoming and
outgoing edges, in the modern graph?

--
[bi-directional code example]
--

[...]
```

Is that what you meant?

dkuppitz: Yep.

newkek: Okay. well no strong opinion here I can change it for that

newkek: Done

dkuppitz:
VOTE +1

PR #922, opened 2018-08-25
TINKERPOP-1959: Gremlin Javascript ability to send a script to the Gremlin Server.

mattallenuk:
Hiya,

Submitting a PR for consideration re sending a script to a Gremlin server. 

I've added an eval() function to GraphTraversal and TraveralSource classes. This function takes an optional script string and bindings map. If a script is provided it will be sent directly to the server, discarding any previous glv steps, if no script is provided one will be built from the previously set glv steps.

RemoteStrategy class has been changed to handle submission of eval operations to the DriverRemoteConnection.

Bytecode class has had a toScript() function added that takes the glv instructions and produces a script with bindings.

Traversal class has been changed to handle responses from eval op execution.

I wanted to have the option to build from glv instructions as I'll be using the library to access an Azure CosmoDB which currently doesn't support bytecode but when it does it will make it much easier to update projects to using bytecode by removing the eval function and replacing with one of the other terminal actions.

I do have a couple of queries regarding the proposed changes:

1) Is it safe to assume that a returned value not containing a bulk property was executed by an eval op?
When parsing the returned data from the server in the TraversalSource class it appears that any data returned from a bytecode op contains a bulk property. Can it be assumed then that any return value that doesn't contain a bulk property is a valid return value for eval op, or should there be some sort of checking in place?

```
_getNext() {
    while (this.traversers && this._traversersIteratorIndex < this.traversers.length) {
      let traverser = this.traversers[this._traversersIteratorIndex];
      if (traverser.bulk && traverser.bulk > 0) {
        traverser.bulk--;
        return { value: traverser.object, done: false };
      } else if (traverser.bulk === undefined) {
        return { value: traverser, done: true }
      }
      this._traversersIteratorIndex++;
    }
```

2) Should every parameter be bound when building a dynamic script?
The Bytecode::toScript() function converts any glv parameters into bindings to be sent to the server. Should this make consideration for parameters that are likely not set by users. For example g.addV('flowers', 'Lilac'), can it be assumed that flowers was set by the implementer and Lilac by a user? If so should the function be changed to only bind parameters we can assume are set by users?

3) Does the unit test need to be more thorough and check more varied instructions and more complex glv queries/instructions?


gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/remote-connection.js
```diff
@@ -66,7 +67,27 @@ class RemoteStrategy extends TraversalStrategy {
...
     }
-    return this.connection.submit(traversal.getBytecode()).then(function (remoteTraversal) {
+
+    let instructions = traversal.getBytecode();
+    let op = 'bytecode';
+    let processor = 'traversal';
+    let args = null;
+
+    // check if the last instruction is an eval statement
+    if (instructions.stepInstructions.length && instructions.stepInstructions[instructions.stepInstructions.length-1][0] === 'eval') {
```
bterlson: Aliasing `instructions.stepInstructions` will make this line easier to read.

Could also consider relying on the falsiness of empty arrays and simplify to `if (sis && sis[sis.length-1][0] === 'eval') {`

mattallenuk: Yep valid point I'll certainly make those changes.

spmallette:
Excellent - thanks for doing this. @jorgebay is the better person to handle this review in full, but I'd like to offer a few points:

> Is it safe to assume that a returned value not containing a bulk property was executed by an eval op?

I'm not sure I would call it "safe", but that is typically what happens. I would say that it is not completely "safe" to do so because an `eval()` could return anything and thus return a `BulkSet` and the user would expect that and I guess it would unroll to individual items when you iterated which the user wouldn't expect. 

> Should every parameter be bound when building a dynamic script?

We refer to what you've done there as a `Translator` (i.e. something that converts bytecode to the string representation of the Gremlin) and that's neat that you have that. We don't implement that as a `toString()` on `Bytecode` but perhaps that's ok though. FYI, here's what a `Translator` looks like in Java:

```groovy
String gremlin = GroovyTranslator.of("g").translate(bytecode)
```

I don't believe that our translators convert arguments to bindings. It assumes them to be literals. We only construct bindings if the user constructs them as first order `Bindings` in the Gremlin itself, but I don't know if we have that concept here in javascript. For consistency sake, we might want to. You can see the approach taken by Python to handle that as an example:

http://tinkerpop.apache.org/docs/current/reference/#_bindings

> Does the unit test need to be more thorough and check more varied instructions and more complex glv queries/instructions?

On the JVM side we test the translators through the GLV suite (essentially we use the JVM process test suite which is basically the same thing).  Maybe the same should be done here.

Separate from your questions, I'd offer that `g.eval()` isn't really in line with how other GLV drivers submit scripts.  While I'd agree it to be convenient to have it work that way, I think it would be better to see that working in a way more consistent with other GLVs.

jorgebay:
Glad to see you working on this @mattallenuk !

In line with what @spmallette was mentioning, I think we need expose a similar API as the rest of GLVs. Python GLV is a good example (because its a scripting language as JavaScript) but if you have some understanding of C#, you can check out gremlin-dotnet as well.

Besides the `RemoteConnection` the other GLVs expose a `Client` (one abstraction above), that is able to submit traversals in bytecode or in script to evaluate, you can reading here:
- https://github.com/apache/tinkerpop/blob/tp32/gremlin-python/src/main/jython/gremlin_python/driver/client.py
- https://github.com/apache/tinkerpop/blob/tp32/gremlin-python/src/main/jython/tests/driver/test_client.py#L39-L48

mattallenuk:
Hiya guys,

I have moved the code out of the existing classes and have created a Client class that can handle script sending and a Translator class that can convert bytecode into a string. 

The Translator class accepts bindings in the form of a javascript object passed into the step function and converts it into Gremlin Groovy Tuple in the translated script. I haven't considered the impact of this upon Bytecode so this may need to be changed or addressed on the Bytecode side, or Bytecode may already handle parameter maps?

In terms of consumption is there a way we could make the Translator more transparent to the user, I did have an eval function added to Traversal which did the conversion and sending, I removed this but would that be an acceptable solution? It's not inline with other GLV's unfortunately.

jorgebay:
> I haven't considered the impact of this upon Bytecode so this may need to be changed or addressed on the Bytecode side, or Bytecode may already handle parameter maps?

The Bytecode generation already has a way to denote parameters embedded in the query, so it's ok to implement this independently.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/process/translator.js
```diff
@@ -0,0 +1,93 @@
...
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+'use strict';
+
+/**
+ * Class to translate glv bytecode steps into executable Gremlin-Groovy script
+ */
+class Translator {
```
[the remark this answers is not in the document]
mattallenuk: Hi @jorgebay,

The purpose of the translator is to allow building of a script from bytecode for users that don't yet have bytecode implementation on their Gremlin Server but want to use GLV instructions rather than writing scripts in text. 

Case in point, I'll be using the project on Azure CosmoDB but I cannot send bytecode as it's not yet supported, they're working on it. So I don't want to have to write all my queries as strings and then have to go back at a later date and convert to glv steps. I'd rather use the glv steps and go back later and implement terminal commands. This might just be my own personal preferences, but I'm sure others would appreciate that flexibility?

This is how it would be used:

```javascript
const g = new graph.Graph().traversal();
g.V().order().by('age', t.order.decr);
const script = new Translator('g').translate(g.getBytecode());
```

That script could then be submitted via the client.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/client.js
```diff
@@ -0,0 +1,65 @@
...
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+'use strict';
+
+const DriverRemoteConnection = require('./driver-remote-connection');
+const Bytecode = require('../process/bytecode');
+
+class Client extends DriverRemoteConnection {
```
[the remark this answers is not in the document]
mattallenuk: np, I'll get that implemented :)

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/process/translator.js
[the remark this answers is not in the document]
mattallenuk: And that is why I was querying bindings in the bytecode as you'll use the glv steps to build a query before converting it to a script. If you are happy for the translator class to stay then I'll need to know how parameters are implemented in bytecode to ensure the translation is done correctly.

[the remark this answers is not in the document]
bterlson: Just want to say, as another CosmosDB user, that I very much would love to be able to use the fluent step API even though cosmos doesn't yet support bytecode submissions. I hadn't considered it was possible to convert the glv steps to a script, but now that I see it working, I want it badly 😊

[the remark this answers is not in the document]
mattallenuk: @jorgebay I'm more than happy to move the Translator functionality into a support project if it doesn't fit within the gremlin-javascript project. I realise it might be too specific a case and hope I haven't overstepped a mark by including it with this PR, it just made total sense to me.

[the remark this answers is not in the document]
spmallette: I think translators are an established part of TinkerPop at this point. We have them for Java, Groovy, and Python and I've personally found them to be insanely useful on so many internal odds/ends. We don't over-publicize them because we don't want to promote "scripts" but I understand the current need for CosmosDB. 

Anyway, including a Javascript translator is an acceptable addition imo.

gremlin-server/src/test/resources/org/apache/tinkerpop/gremlin/server/gremlin-server-integration.yaml
```diff
@@ -41,6 +41,7 @@ serializers:
   - { className: org.apache.tinkerpop.gremlin.driver.ser.GraphSONMessageSerializerV1d0, config: { ioRegistries: [org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerIoRegistryV1d0] }}
 processors:
   - { className: org.apache.tinkerpop.gremlin.server.op.session.SessionOpProcessor, config: { sessionTimeout: 28800000 }}
+  - { className: org.apache.tinkerpop.gremlin.server.op.standard.StandardOpProcessor, config: { maxParameters: 64 }}
```
spmallette: Do we have to test for more than the default parameters (which is 16)?

mattallenuk: No we don't I'd just read somewhere that it was the recommended. I'll remove that option so the default is used.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/client.js
```diff
@@ -0,0 +1,87 @@
...
+   * Send a request to the Gremlin Server, can send a script or bytecode steps.
+   * @param {Bytecode|string} message The bytecode or script to send
+   * @param {Object} bindings The script bindings, if any.
+   * @returns {Promise}
+   */
+  submit(message, bindings) {
+    if (typeof message === 'string' || message instanceof String) {
+      const args = {
+        'gremlin': message,
+        'bindings': bindings,
```
jorgebay: I think bindings need to be serialized.

Currently, I think it will only work with json supported types (boolean, number, strings), but it should work with any type supported by GraphSON2.

jorgebay: oh, I see that `args` are adapted by the `Connection` afterwards, never mind my previous comment.

@mattallenuk: Can we include a test containing a binding like a `Long` or `Set` or any type that is not a native js type?

mattallenuk: @jorgebay yes, I'll get that done.

spmallette:
hi @mattallenuk - Judging from the comments it seems like the only thing missing here code-wise is an added test. Administratively, there's some documentation (reference docs, upgrade docs, changelog entry) to add for this feature as well. I don't mean to pressure or rush you but do you happen to have any idea when you might come back to this to finish up the final bits? 

I mainly ask because we are in the midst of planning the release of 3.2.10 and 3.3.4 and we'd like to see this work of yours included. If your schedule doesn't allow for you to come back to this right now, perhaps I could assist in polishing up the final odds/ends for you?

mattallenuk:
@spmallette, my apologies I've been unwell the last week. I have done the test but it showed that I need to parse the bindings so that the standard Op processor can read them, it doesn't appear to like bytecode bindings. I'm hoping to have that done and tested today.

On the documentation side I noticed that there doesn't appear to be any documentation in the other glv sections about submitting scripts, am I okay to add this documentation to the javascript glv section?

spmallette:
sorry to hear about the sickness. i hope your doing well now. you're right about the documentation. i think we wanted to focus folks on bytecode based traversals. we do have a section for gremlin-python for submitting scripts, but i think i discouraged the one for .NET. I'd forgotten that.  I'd had the idea that i was going to massively rework the documentation on 3.4.0 to better organize all this GLV stuff and that hasn't fully happened yet.

ok, so, i guess you just need the test/serialization fixups then, a changelog entry and a note in upgrade docs to say that this is now there...perhaps just add some example usage there.  thanks

mattallenuk:
@jorgebay @spmallette So I've written the following test, however I'm getting a Provided key/value array length must be a multiple of two when submitting. I am guessing this is due to submitting the cardinality as a string value. How should non-native JS types be submitted via script?

`connection.submit('g.V(vert).property(card, nm, val)', { vert: 1, card: t.cardinality.set, nm: 'test', val: 12 } )` 

this produces:

[RequestMessage{, requestId='reqid', op='eval', processor='', args={gremlin=g.V(1).property(card, nm, val), aliases={g=gmodern}, bindings={vert=1, card=set, nm=test, val=12}, language=gremlin-groovy, accept=application/json}}]

mattallenuk:
@spmallette Doing better now thank you!

mattallenuk:
I've just tried it not parsing the bindings to strings and get the same issue. So I'm wondering now if I do need to parse the parameters or if I'm just calling the addProperty function wrong?

`[RequestMessage{, requestId=27ad0188-8b5a-40e0-bf52-b158a1e92507, op='eval', processor='', args={gremlin=g.V(1).property(card, nm, val), aliases={g=gmodern}, bindings={card={typeName=Cardinality, elementName=set}, nm=test, val=12}, language=gremlin-groovy, accept=application/json}}].`

spmallette:
`addProperty()` looks right.  On `tp32` I guess the default is GraphSON 2.0 which has this form for `Cardinality` values:

http://tinkerpop.apache.org/docs/3.2.9/dev/io/#_cardinality

I see that the `accept` is set to "application/json". On `tp32` I think that defaults to GraphSON 1.0 which wouldn't be able to deal with complex objects. You can only pass primitives that are compatible with JSON data types. Does it work if you don't use `Cardinality`? maybe, just do something more simple like:

````js
connection.submit('x+y', { x: 1, y: 1 } );
```

If that works then change to GraphSON 2.0 - "application/vnd.gremlin-v2.0+json" - and see what happens for complex stuff like:

````js
connection.submit(card.toString()', { card: t.cardinality.set } );
```

spmallette:
It works with Java and GraphSON 2.0:

```text
gremlin> cluster = Cluster.build().serializer("GRAPHSON_V2D0").create()
==>localhost/127.0.0.1:8182
gremlin> client = cluster.connect()
==>org.apache.tinkerpop.gremlin.driver.Client$ClusteredClient@24a298a6
gremlin> client.submit("g.addV().property(c,k,v)", [c:set,k:'name',v:123]).all().get()
==>result{object=v[4] class=org.apache.tinkerpop.gremlin.structure.util.detached.DetachedVertex}
gremlin> client.submit("g.V().has('name',123).count()").all().get()
==>result{object=1 class=java.lang.Long}
```

So, i'm not sure what's wrong. If you want to commit/push what you have i can try to test on my own to see what I (or maybe someone else) can figure out.

mattallenuk:
@spmallette Ok, have pushed the changes along with the documentation changes we'd discussed. Hopefully it's an easy fix.

mattallenuk:
@spmallette I can confirm that this test works for bytecode submission so possibly an issue in the way the bindings are being submitted? Should the language parameter be gremlin-groovy or something else?

spmallette:
"gremlin-groovy" is fine. The server should default to that if the argument isn't present.

spmallette:
@mattallenuk  I figured out what was going on. The `EnumValue` was being interpreted as a `Map` when serializing to GraphSON. The change is pretty trivial, but I have a bunch of other little changes here/there. I'm going to merge this PR to a TINKERPOP-1959 branch for review (hope that's ok with you) and hopefully we can get this merged up today.

spmallette:
PR for this is now on #952

PR #923, opened 2018-08-28
TINKERPOP-2028: Register GremlinServerModule to GraphSON message serializer

newkek:


newkek:
Left the current behavior unchanged but deprecated the current method, in favor of another `GraphSONMessageSerializer` constructor that gives the opportunity to child classes of `AbstractGraphSON2MessageSerializer` to register their own modules when a `GraphSONMapper.Builder` is provided.

spmallette:
This seems to build now along with gremlin server integration tests. Any reason to not go to `tp33`? if not, please squash and retarget and i can go ahead an move to merge.

Please also take a look at adding `final` to variables where needed.  A CHANGELOG entry might be good too. 

Thanks for this.

VOTE +1

newkek:
The reason this is targeting master is because although I did not break backward compatibility and put previous outdated methods as Deprecated I wasn't sure what the strategy is here for retiring things in patch versions. Also there would be a slight change of behavior if anybody were to have also extended `AbstractGraphSON2MessageSerializer`. Let me know if that's acceptable to put in a patch or not, if so I don't any other reason not to do it in tp33.

spmallette:
ok - i understand. it can stay here for master only.

PR #924, opened 2018-08-31
Add self loop example to Cycle Detection recipe.

newkek:


newkek:
sorry - I messed up - shouldn't have opened that

PR #925, opened 2018-09-01
TINKERPOP-2026 Make closing of connections more robust in Gremlin.Net.Driver 

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2026

This should solve the problem described in the ticket where apparently the server already initiated closing of the connection when the driver tries to close it which led to an exception.

VOTE +1

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/WebSocketConnection.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/WebSocketConnection.cs
@@ -43,10 +43,23 @@ namespace Gremlin.Net.Driver
 
         public async Task CloseAsync()
         {
-            await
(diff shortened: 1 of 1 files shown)


FlorianHockmann:
No, there is no reason for not merging this. I'm just currently on vacation. So, I can't do it myself right now.
But if you or anyone else wants to merge this, then feel free to do that.

SamiMajed:
Thank you, Florian. Is this fix scheduled to be part of the 3.4 release? 
If so, what would be the target release date?

spmallette:
looks like this is targetted for all the open release branches: 3.2.10, 3.3.4 and 3.4.0. we just agreed on the dev list to push of 3.4.0 for a few months as it doesn't feel "done". we did however agree to do a 3.4.0-rc2 release as early as next week.  Since florian gave the ok, i will likely merge this one for him so that it lands in time for the release candidate.

PR #928, opened 2018-09-09
TINKERPOP-2015 Expose WebSocket configuration in Gremlin.Net driver

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2015

Users can now provide a delegate to the Gremlin.Net driver that will be used to configure WebSocket connections. Using a delegate for this has the advantage that it's extensible as the options available via [`ClientWebSocketOptions`](https://docs.microsoft.com/dotnet/api/system.net.websockets.clientwebsocketoptions) change between different .NET implementations, e.g., as mentioned in the ticket, a `RemoteCertificateValidationCallback` was added with .NET Core 2.1.

When #903 is merged first, then we need to add this new option also to the docs when merging this into `master` as #903 adds descriptions for the config options to the docs.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,6 +25,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 * Removed conflicting non-indy groovy core dependency
 * Bumped jython-standalone 2.7.1
+* Added a delegate to the Gremlin.Net driver that can be used to configure the WebSocket connection.
 * SSL security enhancements
 * Added Gremlin version to Gremlin Server startup logging output.
 * Fixed problem with Gremlin Server sometimes returning an additional message after a failure.
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
@@ -23,6 +23,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Net.WebSockets;
 using System.Text;
 using System.Threading.Tasks;
 using Gremlin.Net.Driver.Messages;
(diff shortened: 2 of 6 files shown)


FlorianHockmann:
Yes, that should be possible. With this PR, you can provide a delegate to configure the ClientWebSocketOptions which should also include custom headers.

dkuppitz:
IMO it would be better to expose the options as a property unless there's a good reason to not allow any changes in options after the client has been initialized. But since the underlying `ClientWebSocket` allows a change in options at any point in time, `GremlinClient` should preferably do the same.

FlorianHockmann:
I also wanted to implement it like that as first, but unfortunately the [`ClientWebSocket.Options` property](https://docs.microsoft.com/dotnet/api/system.net.websockets.clientwebsocket.options) only has a `get`, but no `set`. So, we have no way to provide our own `ClientWebSocketOptions` object. I also don't see how we could forward changes to the `ClientWebSocketOptions` of every already created connection.

dkuppitz:
Hm, alright. I don't have an environment to test it myself, but apparently I'm missing something; I thought it would be a simple property delegation.

VOTE +1

deejvince:
@FlorianHockmann

Could you please post a working example on how to use the delegate with the driver connection with GremlinClient?

FlorianHockmann:
@deejvince: Did you see [the integration test](https://github.com/apache/tinkerpop/blob/master/gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Driver/GremlinClientTests.cs#L233) added with this PR?

The relevant part is basically this:
```cs
var webSocketConfiguration =
                new Action<ClientWebSocketOptions>(options =>
                {
                    options.UseDefaultCredentials = false;
                    options.KeepAliveInterval = TimeSpan.FromMilliseconds(11);
                });
var gremlinClient = new GremlinClient(gremlinServer,
                webSocketConfiguration: webSocketConfiguration);
```

The `Action` you provide will be called for every `ClientWebSocket` created by the driver of Gremlin.Net.

I haven't tested that specifically, but it should also be possible to set a request header like this:
```cs
var webSocketConfiguration =
                new Action<ClientWebSocketOptions>(options =>
                {
                    options.SetRequestHeader("Authorization", "Bearer " + token);
                });
```

Please report back if that doesn't work.

jantzeca:
My latest Gremlin.net NuGet package is showing as 3.3.3. Is the package for these new changes not ready yet?

spmallette:
@jantzeca this merged a day or so ago.  it has not been released yet to nuget. we have releases scheduled for 3.2.10 and 3.3.4 in the next couple of weeks. it will be available then.

jantzeca:
@spmallette Okay, thank you!

PR #929, opened 2018-09-10
TINKERPOP-2035 Pass custom headers to the websocket connection

deejvince:
In order to authorize to the gremlin server (in this case AWS Neptune) i need to pass custom headers for authorization to the service. the PR will allow to send any headers (like an API key or signature) and pass them to the Websocket function.
While there can be many use cases for this, in my case i pass the Authorization & x-amz-date required for Signature V4 (IAM) authentication.

VOTE: +1

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/driver-remote-connection.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/driver-remote-connection.js
@@ -49,7 +49,7 @@ class DriverRemoteConnection extends RemoteConnection {
    * @param {String} [options.traversalSource] The traversal source. Defaults to: 'g'.
    * @param {GraphSONWriter} [options.writer] The writer to use.
    * @param {Authenticator} [options.authenticator] The authentication handler to use.
-   * @param {String} [options.headers] Add custom headers to the request.
+   * @param {Object} [options.headers] An associative array containing the additional header key/values for the initial request.
    * @constructor
    */
   constructor(url, options) {
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/driver-remote-connection.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/driver-remote-connection.js
@@ -49,12 +49,14 @@ class DriverRemoteConnection extends RemoteConnection {
    * @param {String} [options.traversalSource] The traversal source. Defaults to: 'g'.
    * @param {GraphSONWriter} [options.writer] The writer to use.
    * @param {Authenticator} [options.authenticator] The authentication handler to use.
+   * @param {String} [options.headers] Add custom headers to the request.
    * @constructor
    */
   constructor(url, options) {
     super(url);
     options = options || {};
     this._ws = new WebSocket(url, {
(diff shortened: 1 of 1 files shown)


deejvince:
Hi @JavierGarciaGarcia,
Yes I made it work by using url ws://neptune-endpoint:8182/gremlin
Signing host;x-amz-date

And passing the signature in the Authorization header as well as the required x-amz-date header to the WebSocket via option.headers.

beebs-systap:
Created [TINKERPOP-2035](https://issues.apache.org/jira/browse/TINKERPOP-2035) to track in JIRA.

JavierGarciaGarcia:
Hi  @deejvince, sorry to bother you again, but i'm still having the same issue using this url. Just to be sure, you've used "ws://neptune-endpoint:8182/gremlin\nSigning host;x-amz-date" as url, correct?

If so, how are you generating the authorization header?. I'm using aws4 library

jorgebay:
Thanks for the patch @deejvince and thanks for filing TINKERPOP-2035 @beebs-systap !

I think its a good idea to expose ways to set the request header, as it's part of the [protocol](https://tools.ietf.org/html/rfc6455). You will need the option to set the header on all GLVs (i.e., Python, C#, Java), right?

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/driver-remote-connection.js
```diff
@@ -49,12 +49,14 @@ class DriverRemoteConnection extends RemoteConnection {
    * @param {String} [options.traversalSource] The traversal source. Defaults to: 'g'.
    * @param {GraphSONWriter} [options.writer] The writer to use.
    * @param {Authenticator} [options.authenticator] The authentication handler to use.
+   * @param {String} [options.headers] Add custom headers to the request.
```
jorgebay: The correct type of the `@param` is `{Object}`.

What do you think about providing a little more detail, like: "An associative array containing the additional header key/values for the initial request"

deejvince: Hi, You're correct, i'm fixing this right away.
Thanks!

deejvince: Regarding other GLVs, yes i believe its important to have that.

jorgebay:
lgtm, VOTE: +1

Once it gets approved, I'll create the tickets for the rest of the GLVs.

spmallette:
sorry, i just noticed this, but @jorgebay should this be targetted at tp32?

jorgebay:
Yes, as this is a small patch I'll apply it directly to `tp32`, instead of asking the OO to rebase it.

deejvince:
@jorgebay @spmallette, Could you merge this PR, or it requires some more approvals ?

jorgebay:
I was waiting for +1 for another committer, as it's a new feature.

jorgebay:
Landed in 54ad5f45e73bae05b957b89539cb6a61d557f445, thanks @deejvince and @beebs-systap !

jorgebay:
I've created the tickets for the other languages on JIRA.

PR #931, opened 2018-09-11
TINKERPOP-2029 ConcurrentModificationException for InlineFilterStrategy

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-2029

To fix the original issue, I basically rewrote `ConnectiveStrategy`. The infix notation of `and()` and `or()` now supports an arbitrary number of traversals and `ConnectiveStrategy` produces a traversal with correct `AND` and `OR` semantics.

The reason for targeting `master/` is that there were no strict semantics before and thus this needs to be considered a breaking change.

```
# BEHAVIOR

Input: a.or.b.and.c.or.d.and.e.or.f.and.g.and.h.or.i

## BEFORE
Output: or(a, or(and(b, c), or(and(d, e), or(and(and(f, g), h), i))))

## NOW
Output: or(a, and(b, c), and(d, e), and(f, g, h), i)
```

VOTE +1

### DIFF
--- a/docs/src/upgrade/release-3.4.x.asciidoc
+++ b/docs/src/upgrade/release-3.4.x.asciidoc
@@ -31,8 +31,19 @@ Please see the link:https://github.com/apache/tinkerpop/blob/3.4.0/CHANGELOG.asc
 
 ==== Changed infix behavior
 
-The infix notation of `and()` and `or()` now supports an arbitrary number of traversals and `ConnectiveStrategy` produces a traversal with the correct AND and OR semantics. Furthermore,
-previous versions failed to apply 3 or more `and()` steps in an infix notation, this is now fixed.
+The infix notation of `and()` and `or()` now supports an arbitrary number of traversals and `ConnectiveStrategy` produces a traversal with proper AND and OR semantics.
+
+```
+Input: a.or.b.and.c.or.d.and.e.or.f.and.g.and.h.or.i
+
+## BEFORE
+Output: or(a, or(and(b, c), or(and(d, e), or(and(and(f, g), h), i))))
+
+## NOW
+Output: or(a, and(b, c), and(d, e), and(f, g, h), i)
+```
+
+Furthermore, previous versions failed to apply 3 or more `and()` steps using the infix notation, this is now fixed.
 
 [source,groovy]
 ----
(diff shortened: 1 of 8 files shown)


docs/src/upgrade/release-3.4.x.asciidoc
```diff
@@ -29,6 +29,29 @@ Please see the link:https://github.com/apache/tinkerpop/blob/3.4.0/CHANGELOG.asc
 
 === Upgrading for Users
 
+==== Changed infix behavior
+
+The infix notation of `and()` and `or()` now supports an arbitrary number of traversals and `ConnectiveStrategy` produces a traversal with the correct AND and OR semantics. Furthermore,
+previous versions failed to apply 3 or more `and()` steps in an infix notation, this is now fixed.
+
```
robertdale: The graph example is good for those familiar with the graph but it's not immediately clear what the difference in behavior is. I really like the example from the comments. I think it would be good to add here (not necessarily verbatim) for clarity. 
```
# BEHAVIOR

Input: a.or.b.and.c.or.d.and.e.or.f.and.g.and.h.or.i

## BEFORE
Output: or(a, or(and(b, c), or(and(d, e), or(and(and(f, g), h), i))))

## NOW
Output: or(a, and(b, c), and(d, e), and(f, g, h), i)
```

docs/src/upgrade/release-3.4.x.asciidoc
```diff
@@ -29,6 +29,29 @@ Please see the link:https://github.com/apache/tinkerpop/blob/3.4.0/CHANGELOG.asc
...
+In previous versions the above traversal 
+[source,groovy]
+----
+gremlin> g.V().repeat(__.in('traverses').repeat(__.in('develops')).emit()).emit().values('name')
+==>stephen
+==>matthias
+==>marko
+----
+
+See: link:https://issues.apache.org/jira/browse/TINKERPOP-967[TINKERPOP-967]
```
robertdale: There should be a link to this issue as well - https://issues.apache.org/jira/browse/TINKERPOP-2029

docs/src/upgrade/release-3.4.x.asciidoc
```diff
@@ -29,6 +29,29 @@ Please see the link:https://github.com/apache/tinkerpop/blob/3.4.0/CHANGELOG.asc
...
+previous versions failed to apply 3 or more `and()` steps in an infix notation, this is now fixed.
+
+[source,groovy]
+----
+gremlin> g.V().has("name","marko").and().has("age", lt(30)).or().has("name","josh").and().has("age", gt(30)).and().out("created")
+==>v[1]
+==>v[4]
+----
+
+In previous versions the above traversal
```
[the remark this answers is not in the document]
dkuppitz: Ugh, I was going to remove this piece as I mentioned previous versions in the comments above.

robertdale:
VOTE +1

spmallette:
VOTE +1

PR #932, opened 2018-09-13
TINKERPOP-2033 Maintain order in profile() annotations

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2033

Used a synchronized map around a `LinkedHashMap` rather than `ConcurrentHashMap`. Not expecting a performance issue with this as its for `profile()` step which doesn't have to be "fast" necessarily as it is a debugging step. 

Tested manually for backward compatibility across a number of different versions. That was quite time consuming because I'd not realized that the we'd introduced a perhaps unintended break between 3.3.0 and 3.2.x when it comes to "metrics" serialization in Gryo 1.0. Ultimately, I should have just accepted the output i was seeing in the IO tests as being correct in terms of what the compatibility was.

Note that while the Gryo 1.0 test file changed for 3.4.0 it still remains compatible with previous versions - there is just a different representation of a {{Map}} in the binary. All of this seems somehow a  non-issue because Gryo 3.0 has been the default since 3.3.0 released.

VOTE +1


nastra:
+1

dkuppitz:
VOTE +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoSerializersV3d0.java
```diff
@@ -427,7 +428,12 @@ else if (1 == arguments)
...
             kryo.writeObject(output, object.getCounts());
-            kryo.writeObject(output, object.getAnnotations());
+
+            // annotations is a synchronized LinkedHashMap - get rid of the "synch" for serialization as gryo
+            // doesn't know how to deserialize that well and LinkedHashMap should work with 3.3.x and previous
+            final Map<String, Object> annotations = new LinkedHashMap<>();
+            object.getAnnotations().forEach(annotations::put);
+            kryo.writeObject(output, annotations);
 
             // kryo might have a problem with LinkedHashMap value collections. can't recreate it independently but
```
robertdale: Do we not need to worry about this comment?

spmallette: i wonder what that was in reference to.................we serder `LinkedHashMap` all over the place

spmallette: @robertdale i thought about it some more. the problem (whatever it was) wasn't related to `LinkedHashMap` itself which is what i'm serializing in the new code. the comment is in relation to the collection returned from `LinkedHashMap.values()` which is why I converted that collection to an `ArrayList`. better explanation?

robertdale:
VOTE +1

PR #934, opened 2018-09-21
Apply StringEscapeUtils in GroovyTranslator#convertToString(String) [tp32]

justinchuch:
Preface:
There's character escaping issue when translating.

What's changed:
- `org.apache.tinkerpop.gremlin.groovy.jsr223.GroovyTranslator#convertToString`
Apply `StringEscapeUtils.escapeJava(String)` for String object.

- `org.apache.tinkerpop.gremlin.groovy.jsr223.GroovyTranslatorTest`
Add a few test cases to verify the changes.

### DIFF
--- a/gremlin-groovy-test/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GroovyTranslatorTest.java
+++ b/gremlin-groovy-test/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GroovyTranslatorTest.java
@@ -30,7 +30,10 @@ import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.__;
 import org.apache.tinkerpop.gremlin.process.traversal.strategy.decoration.SubgraphStrategy;
 import org.apache.tinkerpop.gremlin.process.traversal.strategy.decoration.TranslationStrategy;
 import org.apache.tinkerpop.gremlin.process.traversal.strategy.verification.ReadOnlyStrategy;
+import org.apache.tinkerpop.gremlin.structure.Edge;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
+import org.apache.tinkerpop.gremlin.structure.util.detached.DetachedEdge;
+import org.apache.tinkerpop.gremlin.structure.util.detached.DetachedVertex;
 import org.apache.tinkerpop.gremlin.util.function.Lambda;
 import org.junit.Test;
 
@@ -116,6 +119,17 @@ public class GroovyTranslatorTest extends AbstractGremlinTest {
         assertEquals(7, sacks.get(5).intValue());
         //
         assertEquals(24, t.getSideEffects().<Number>get("lengthSum").intValue());
+
+        final String script = GroovyTranslator.of("g").translate(t.getBytecode());
+        assertEquals("g.withSideEffect(\"lengthSum\",(int) 0).withSack((int) 1)" +
+                        ".V()" +
+                        ".filter({it.get().label().equals('person')})" +
+                        ".flatMap({it.get().vertices(Direction.OUT)})" +
+                        ".map({it.get().value('name').length()})" +
+                        ".sideEffect({ x -> x.sideEffects(\"lengthSum\", x.<Integer>sideEffects('lengthSum') + x.get()) })" +
+                        ".order().by({a,b -> a <=> b})" +
+                        ".sack({ a,b -> a + b })",
+                script);
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


gremlin-groovy-test/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GroovyTranslatorTest.java
```diff
@@ -141,4 +144,52 @@ public void shouldHandleEmptyMaps() {
...
+        final Object id2 = "name:20:foo\\u0020bar#50";
+
+        final Vertex vertex2 = DetachedVertex.build().setLabel(label2).setId(id2).create();
+        final String script2 = GroovyTranslator.of("g").translate(g.V().inject(vertex2).asAdmin().getBytecode());
+        assertEquals("g.V().inject(new org.apache.tinkerpop.gremlin.structure.util.detached.DetachedVertex(\"name:20:foo\\\\u0020bar#50\",\"user\", Collections.emptyMap()))", script2);
+    }
+
+    @Test
+    @LoadGraphWith(LoadGraphWith.GraphData.MODERN)
+    public void shouldHandleLambda() {
```
spmallette: is this doing anything discernibly different than  `shouldSupportStringSupplierLambdas()` test above? if there is a specific you are trying to assert here, perhaps it could be just appended into that test?

gremlin-groovy-test/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GroovyTranslatorTest.java
```diff
@@ -141,4 +144,52 @@ public void shouldHandleEmptyMaps() {
     public void shouldHaveValidToString() {
         assertEquals("translator[h:gremlin-groovy]", GroovyTranslator.of("h").toString());
     }
+
+    @Test
+    @LoadGraphWith(LoadGraphWith.GraphData.MODERN)
```
spmallette: It doesn't look like you need the "modern" graph data here - you can remove this annotation.

gremlin-groovy-test/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GroovyTranslatorTest.java
```diff
@@ -141,4 +144,52 @@ public void shouldHandleEmptyMaps() {
...
+    public void shouldHandleTraversalAddV() {
```
spmallette: the names of our tests are meant to convey information about what is being tested (when possible). to that end, you're not really testing "addV()" here but you're testing "string escaping", so something more like `shouldEscapeStrings()` would probably be a better name. I would probably combine this test with `shouldHandleVertex()` as they are both validating the "escaping" but if you have a better name for that test as it test something more specific than that then feel free to just rename the other test and keep them separate.

gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GroovyTranslator.java
```diff
@@ -115,7 +116,8 @@ else if (object instanceof Bytecode)
         else if (object instanceof Traversal)
             return convertToString(((Traversal) object).asAdmin().getBytecode());
         else if (object instanceof String) {
-            return (((String) object).contains("\"") ? "\"\"\"" + object + "\"\"\"" : "\"" + object + "\"").replace("$", "\\$");
+            return (((String) object).contains("\"") ? "\"\"\"" + StringEscapeUtils.escapeJava((String) object) + "\"\"\"" : "\"" + StringEscapeUtils.escapeJava((String) object) + "\"")
+                    .replace("$", "\\$");
```
spmallette: since you used the "groovy" package for this, i was wondering if it still handles the "$" properly or if we still need the final `replace("$", "\\$")`? either way, could you please modify your tests to include a "$" character so that we can be sure that gets asserted?

gremlin-groovy-test/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GroovyTranslatorTest.java
justinchuch: I have appended the test `shouldSupportStringSupplierLambdas` to include assert the result of `translate`.

gremlin-groovy-test/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GroovyTranslatorTest.java
justinchuch: Thanks for pointing that out. I have combined the tests.

gremlin-groovy-test/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GroovyTranslatorTest.java
justinchuch: I have renamed the test method and separate into two:
1. `shouldEscapeStrings`
2. `shouldHandleVertexAndEdge`

gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GroovyTranslator.java
justinchuch: The final `replace("$", "\\$")` has to be preserved or else some other original tests will be failed.

dkuppitz:
Changes and tests look good to me.

VOTE +1

spmallette:
VOTE +1

robertdale:
VOTE +1

PR #935, opened 2018-09-24
TINKERPOP-2025 Change to SHA-256/512 and drop SHA-1 for releases

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-2025

Unfortunately, upgrading the Apache parent pom didn't solve the problem completely. With the upgrade we get the sha512 checksum file for the source release, but not for the other artifacts.

This PR upgrades the parent pom, but also adds a step in the release process to replace any remaining sha1 checksum file with its sha512 counterpart.

The release validation script was adjusted to reflect those changes; the validation will fail if there exists

* a md5 file
* a sha1 file
* no asc file
* no sha512 file

... for any of the release artifacts.

VOTE +1


robertdale:
VOTE +1

spmallette:
I suppose that it's not surprising that they only did the source release. Technically, Apache only releases source code - binaries are for user convenience only.

VOTE +1

PR #938, opened 2018-09-24
Add link to Elixir language driver

samhavens:
We have an Elixir language driver for Gremlin and would like the community to be aware of it.

### DIFF
--- a/docs/site/home/index.html
+++ b/docs/site/home/index.html
@@ -270,6 +270,7 @@ limitations under the License.
             <li><a href="https://github.com/marcelocf/gremlin_client">gremlin_client</a> (ruby) - A Gremlin Server driver for Ruby.</li>
             <li><a href="http://tinkerpop.apache.org/docs/current/reference/#connecting-via-java">gremlin-driver</a> (java) - A Gremlin Server driver for Java.</li>
             <li><a href="https://github.com/qasaur/gremgo">gremgo</a> (go) - A Gremlin Server driver for Go.</li>
+            <li><a href="https://github.com/Revmaker/gremlex">gremlex</a> (elixir) - A Gremlin Server driver for Elixir.</li>
             <li><a href="http://tinkerpop.apache.org/docs/current/reference/#gremlin-DotNet">Gremlin.Net</a> (.NET - C#) - Gremlin Server driver for .NET.</li>
             <li><a href="https://github.com/PommeVerte/gremlin-php">gremlin-php</a> (php) - A Gremlin Server driver for PHP.</li>
             <li><a href="http://tinkerpop.apache.org/docs/current/reference/#gremlin-python">gremlin-python</a> (python) - Gremlin Server driver for Python.</li>


spmallette:
@samhavens  First, thanks for making us aware of this. Second, I've started a thread on the dev mailing list to call attention to this (all web site changes go through community discussion over there) - you can see the thread here:

https://lists.apache.org/thread.html/6172f72fea030a0c46f983eda42321b7c4a958bed6038fe676990fc2@%3Cdev.tinkerpop.apache.org%3E

Third, could you please edit your README (and perhaps even the project description) to mention "Apache TinkerPop™".  

Finally, do you have a more complex example of Gremlin in Elixer? I'd love to see more of the syntax - I looked at the tests, but I didn't see a nice sized query - preferably using the "modern" graph for context.  Even if it's not all working yet, could you demonstrate what this would look like:

```
g.V().has("name","marko").        
  out("knows").                     
  out("knows").
  values("name") 
```

and maybe:

```
 g.V().match(
      as("a").out("knows").as("b"),
      as("a").out("created").as("c"),
      as("b").out("created").as("c"),
      as("c").in("created").count().is(2)).
        select("c").by("name")
```

samhavens:
@spmallette Thanks for the feedback. Once we pull in [this PR](https://github.com/Revmaker/gremlex/pull/43), I think that should cover it.

Regarding the last example, we aren't sure Gremlex would ever support that directly in Elixir, so we show it as a raw query. Is that ok?

spmallette:
hehe - feel free to use that in your README if you like it :smile: 

gremlin-bottle

on twitter: https://twitter.com/spmallette/status/1044634316813357056

barakyo:
@spmallette there seems to be much debate over the logo, this is one we came up with internally:

image

(I'll follow up your question about byte code in another response)

spmallette:
hahaha - that's awesome!

barakyo:
@spmallette the Elixir driver was modeled closely after the gremlin-python driver. A call to [Gremlex.Graph.g()](https://github.com/Revmaker/gremlex/blob/master/lib/gremlex/graph.ex#L30-L31) create an [Erlang Queue](http://erlang.org/doc/man/queue.html). This basically starts the queue and from there we build queries using the rest of the functions. Each function accepts the queue as its first argument and updates the queue by appending a tuple where the first value in the tuple is the function name and the second is its arguments. We then return the queue back to the user to allow them to continue to build the query. In case you aren't familiar with Elixir, the pipe operator (`|>`) allows the nice syntax of being able to pass the output from one function to the next. This allowed us to easily build queries and unit test our [query builder](https://github.com/Revmaker/gremlex/blob/master/test/graph_test.exs#L16-L18).

An Erlang queue is represented by a tuple with two lists (front and back), so an empty queue would look like:

```Elixir
iex(1)> Graph.g()
{[], []}
```
Calling a function simply appends a tuple to the queue with its arguments:

```
iex(2)> Graph.g() |> Graph.v()
{[{"V", []}], []}
```

Once you've built you're query, we pass the structure to our Client module which in turn compiles the data structure to a query using [Graph.encode/1](https://github.com/Revmaker/gremlex/blob/master/lib/gremlex/graph.ex#L335-L366). Once the query is compiled, we construct a [request](https://github.com/Revmaker/gremlex/blob/master/lib/gremlex/request.ex#L15) to make to the server.

You can see how we compile a query in our [tests](https://github.com/Revmaker/gremlex/blob/master/test/graph_test.exs#L393-L402). Here is a simple example:

```Elixir
iex(4)> Graph.g() |> Graph.v() |> Graph.encode
"g.V()"
```

spmallette:
thanks for the explanation. it looks like it's script based and not bytecode based. it would be cool to see bytecode support in the future. you might also consider implementing our test suite to ensure compliance with the Gremlin language. they are gherkin based tests and thus designed to be cross-language compatible: 

https://github.com/apache/tinkerpop/tree/master/gremlin-test/features

You can read a bit about their structure here:

http://tinkerpop.apache.org/docs/3.4.0-SNAPSHOT/dev/developer/#_gremlin_language_test_cases

Anyway - very cool. Please keep us up to date on how things go with your project. Please feel free to use the gremlin-users mailing list to promote/support Gremlex:

https://groups.google.com/forum/#!forum/gremlin-users

spmallette:
Merged - web site is published though it may take some time to propagate the changes so as to be visible

PR #939, opened 2018-09-25
TINKERPOP-2045 removed duplicate non-indy groovy core dep

robertdale:
https://issues.apache.org/jira/browse/TINKERPOP-2045

Exclude non-indy transitive dep on groovy core.

`mvn clean install` success

VOTE +1


spmallette:
VOTE +1

dkuppitz:
VOTE +1

PR #942, opened 2018-09-27
TINKERPOP-2049 Added with(k) overload

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2049

Makes flag-like uses of with() a bit more succinct. Basically, the shorthand of `with(k,true)` is simply `with(k)`.

VOTE +1


dkuppitz:
VOTE +1

robertdale:
VOTE +1

robertdale:
VOTE +1

PR #944, opened 2018-10-01
TINKERPOP-2041 Text Predicates

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-2041

This PR adds a few text predicates (`TP`):

* `startsWith`
* `endsWith`
* `contains`

By the contract definition, every predicate needs an opposite predicate, hence I also added the following:

* `startsNotWith`
* `endsNotWith`
* `absent`

However, I don't like the naming too much. If anybody has better suggestions, I'm all ears.

VOTE +1


robertdale:
I would like to see `TP` renamed to `TextP`.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -3356,24 +3356,32 @@ interface. Steps that allow for this type of modulation will explicitly state so
...
+| `P.gte(number)` | Is the incoming number greater than or equal to the provided number?
+| `P.inside(number,number)` | Is the incoming number greater than the first provided number and less than the second?
+| `P.outside(number,number)` | Is the incoming number less than the first provided number or greater than the second?
+| `P.between(number,number)` | Is the incoming number greater than or equal to the first provided number and less than the second?
+| `P.within(objects...)` | Is the incoming object in the array of provided objects?
+| `P.without(objects...)` | Is the incoming object not in the array of the provided objects?
+| `TextP.startsWith(string)` | Does the incoming `String` start with the provided `String`?
+| `TextP.endsWith(string)` | Does the incoming `String` end with the provided `String`?
+| `TextP.contains(string)` | Does the incoming `String` contain the provided `String`?
+| `TextP.startsNotWith(string)` | TODO: find a better name
```
spmallette: maybe we should put all the "nots" together so that:

* `notStartingWith()`
* `notEndingWith()`
* `notContaining()`

They seem to read a bit less awkwardly when I say them out loud. if we wanted we could make it all consistent and be like:

* `startingWith()`
* `endingWith()`
* `containing()`
* `notStartingWith()`
* `notEndingWith()`
* `notContaining()`

but maybe not necessary???

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Text.java
```diff
@@ -0,0 +1,123 @@
...
+ * under the License.
+ */
+package org.apache.tinkerpop.gremlin.process.traversal;
+
+import java.util.function.BiPredicate;
+
+/**
+ * @author Daniel Kuppitz (http://gremlin.guru)
+ */
+public enum Text implements BiPredicate<String, String> {
```
spmallette: Could you please complete all the javadoc on this class since it's public facing like `Compare` - I think you need all the `@since` tags on everything in here.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TextP.java
```diff
@@ -0,0 +1,107 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.tinkerpop.gremlin.process.traversal;
+
+import java.util.function.BiPredicate;
+
+/**
+ * @author Daniel Kuppitz (http://gremlin.guru)
```
spmallette: Just needs some class javadoc - looks like you covered the methods pretty well.

docs/src/reference/the-traversal.asciidoc
dkuppitz: Sounds good to me, much better than what it is now.

spmallette:
I think we need to add `TextP` to the IO tests and provide an example in the related IO docs. I can do that tomorrow.

spmallette:

VOTE +1

robertdale:
VOTE +1

PR #945, opened 2018-10-02
TINKERPOP-2053 Support for OptionsStrategy

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2053

`OptionsStrategy` lets users pass arbitrary configurations at a traversal level. At first I thought this might be a first class object, but decided instead to try it out as a `TraversalStrategy`. In that way it was implemented with little overhead. If it becomes useful and popular in the future then we can consider making it more integrated into the `TraversalSource` and `Traversal` itself. I didn't implement this for javascript because I don't think we have all the infrastructure there for supporting `TraversalStrategy` stuff yet. 

VOTE +1


robertdale:
VOTE +1

dkuppitz:
VOTE +1

PR #949, opened 2018-10-04
TINKERPOP-2056 Use NumberHelper in Compare

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-2056

Instead of converting all numbers to `BigDecimal`, `Compare` predicates now use `NumberHelper` for numerical comparisons.

`docker/build.sh -t -i -n` passed.

VOTE +1

PR #950, opened 2018-10-04
TINKERPOP-2058 Contains predicates should rely on Compare predicates

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-2058

If the object to be filtered is a number, `Contains` predicates will now scan the provided collection, comparing each element using `Compare.eq`. For non-numeric values `Contains.within` will still use `collection.contains()` in order to make use of search-optimized data types (e.g. `Set`).

This is a breaking change as the test suite previously ensured that number types had an effect on `Contains` predicates. This, however, was inconsistent with `Compare` predicates which ignore the number type.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,6 +25,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 This release also includes changes from <<release-3-3-3, 3.3.3>>.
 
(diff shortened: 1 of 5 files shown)


dkuppitz:
Reminder: Merge #949 first, then rebase this PRs branch and fix `PTest` (include differing number tests for `Contains` predicates).

https://github.com/apache/tinkerpop/pull/949/files#diff-dc8c8777ab12a44729fc5f2c9c255c66R115

spmallette:
VOTE +1

PR #948, opened 2018-10-03
Optimizes Map with enum using the EnumMap implementation [tp32]

otaviojava:
This PR replaces the HashMap implementation to [EnumMap](https://docs.oracle.com/javase/8/docs/api/java/util/EnumMap.html) that as its documentation says:

> A specialized Map implementation for use with enum type keys. All of the keys in an enum map must come from a single enum type that is specified, explicitly or implicitly, when the map is created. Enum maps are represented internally as arrays. This representation is extremely compact and efficient.

PR #951, opened 2018-10-05
 Optimizes Set with enum using the EnumSet implementation [tp32]

otaviojava:
This PR replaces the HashSet implementation to [EnumSet](https://docs.oracle.com/javase/7/docs/api/java/util/EnumSet.html) that as its documentation says:

> A specialized Set implementation for use with enum types. All of the elements in an enum set must come from a single enum type that is specified, explicitly or implicitly, when the set is created. Enum sets are represented internally as bit vectors. This representation is extremely compact and efficient. The space and time performance of this class should be good enough to allow its use as a high-quality, typesafe alternative to traditional int-based "bit flags." Even bulk operations (such as containsAll and retainAll) should run very quickly if their argument is also an enum set.

That is a continuation of the PR: https://github.com/apache/tinkerpop/pull/948 that has the same benchmark value.


dkuppitz:
VOTE +1

spmallette:
@otaviojava just wanted to let you know that we hadn't forgotten this. we're just in the middle of release so the release branches (like the one you're targeting) are currently frozen while that is happening. should be re-opened soon enough.

otaviojava:
you welcome

PR #953, opened 2018-10-05
TINKERPOP-1972 Fix inject step in Gremlin.Net

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-1972

We had two issues with the `inject` step in Gremlin.Net. The first one prevented the Gherkin test runner from executing scenarios that included an `inject` step. This was fixed by @jorgebay with the first commit in this PR. Thanks for that, Jorge!

The second issue was that the arguments passed to the `inject` step were always wrapped in an array which was caused by the fact that the `inject` step takes a `params` array with a generic member type. We created a `List<S>` for these arguments and provided that to the method `ByteCode.AddStep()` which expects a `params object[]` argument. Since the compiler can't convert `List<S>` into `object[]`, it simply treated the `List<S>` as the first argument which resulted in an `object[]` with just one element, namely the `List<S>`.
Simply using a `List<object>` for the arguments and then converting each argument to `object` fixes the problem.
This could also have affected other steps, but `inject` was the only one with such a generic `params` array which is probably why we didn't notice the problem before. 

All .NET tests, including the GLV scenarios, are now passing again 🎉

VOTE +1

### DIFF
--- a/gremlin-dotnet/glv/GraphTraversal.template
+++ b/gremlin-dotnet/glv/GraphTraversal.template
@@ -22,6 +22,7 @@
 #endregion
 
 using System.Collections.Generic;
+using System.Linq;
 using Gremlin.Net.Structure;
 
 // THIS IS A GENERATED FILE - DO NOT MODIFY THIS FILE DIRECTLY - see pom.xml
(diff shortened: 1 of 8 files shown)


spmallette:

VOTE +1

dkuppitz:
VOTE +1

PR #961, opened 2018-10-10
TINKERPOP-2061 Added with() option for traversal configuration

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2061

Basically just shorthands the direct use of `OptionStrategy` which is much more verbose and less nice to look at. See the upgrade doc changes in this PR for an example of usage.

VOTE +1

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalSource.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalSource.java
@@ -24,6 +24,7 @@ import org.apache.tinkerpop.gremlin.process.computer.Computer;
 import org.apache.tinkerpop.gremlin.process.computer.GraphComputer;
 import org.apache.tinkerpop.gremlin.process.computer.traversal.strategy.decoration.VertexProgramStrategy;
 import org.apache.tinkerpop.gremlin.process.remote.RemoteConnection;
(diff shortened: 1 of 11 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalSource.java
```diff
@@ -93,6 +96,39 @@ private Symbols() {
...
+     * @since 3.4.0
+     */
+    public default TraversalSource with(final String key, final Object value) {
+        final Optional<OptionsStrategy> opt = getStrategies().getStrategy(OptionsStrategy.class);
+        final OptionsStrategy strategy = opt.orElse(OptionsStrategy.EMPTY);
+        final OptionsStrategy.Builder builder = OptionsStrategy.build();
+        strategy.getOptions().forEach(builder::with);
+        builder.with(key, value);
+        return withStrategies(builder.create());
+    }
```
dkuppitz: Perhaps a little bit more elegant:
```java
    public default TraversalSource with(final String key, final Object value) {
        final OptionsStrategy.Builder builder = OptionsStrategy.build();
        getStrategies().getStrategy(OptionsStrategy.class)
                .ifPresent(optionsStrategy -> optionsStrategy.getOptions().forEach(builder::with));
        builder.with(key, value);
        return withStrategies(builder.create());
    }
```

spmallette: done

PR #964, opened 2018-10-17
TINKERPOP-2071: gremlin-python: g:Set graphson deserializer should return a python set

aboudreault:
If there is additional documentation you would like somewhere... please, point me where.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -23,6 +23,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 [[release-3-3-5]]
 === TinkerPop 3.3.5 (Release Date: NOT OFFICIALLY RELEASED YET)
 
+* gremlin-python: the graphson deserializer for g:Set should return a python set
 
 [[release-3-3-4]]
 === TinkerPop 3.3.4 (Release Date: October 15, 2018)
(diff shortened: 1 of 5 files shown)


spmallette:
Thanks for submitting this....Travis seems unhappy on the gremlin-python version of the build....could you please have a look at that? Also, this looks like it needs a rebase now that we have the tp33 branch re-opened for development. That should get CHANGELOG setup for new entries. Please don't include the JIRA issue there - just a bullet point with your representation of what changed will suffice.

spmallette:
I think there needs to be a documentation update too - i just noticed this:

http://tinkerpop.apache.org/docs/3.3.4/reference/#_limitations

aboudreault:
@spmallette I addressed comments and fixed the initial issue we were seeing in travis. However, I am not able to reproduce this one: https://travis-ci.org/apache/tinkerpop/jobs/445261231#L7397 , everything seems to pass locally. Any pointer?

spmallette:
I re-ran the travis build and the error is gone. probably a fluke of travis or something. I guess this is good now.  do you happen to know if there is a test for the alternative scenario somewhere (i.e. where a set becomes a list)?

aboudreault:
@spmallette Let me look at how this scenario stuff works. I will add one that return a list.

spmallette:
I think this one is ok at this point - note the creation of https://issues.apache.org/jira/browse/TINKERPOP-2082 for other GLVs 

VOTE +1

PR #966, opened 2018-10-23
TINKERPOP-2070: Introduce Connection and ResultSet abstractions

jorgebay:
https://issues.apache.org/jira/browse/TINKERPOP-2070

Extract `Connection` implementation out of the `DriverRemoteConnection`.

The `DriverRemoteConnection` now uses a `Client` instance, a `Client` instance may contain one or more `Connection` instances, similar to the Python and .NET glv.

Introduce `ResultSet` that is aligned to other GLVs: 
- `Client` and `Connection` `submit()` methods return a `ResultSet` instance
- `DriverRemoteConnection` submissions returns an instance of `RemoteTraversal`.

This is a large-ish refactor but it doesn't represent a breaking change for the user.

Additionally, I've identified an issue when we introduced script submission support in gremlin-javascript (my bad!): We exposed a property named `traversers` (see previous code examples). I've maintained that property and marked it as deprecated, to be removed in 3.4, and updated all code examples to stop referencing to it.

### DIFF
--- a/docs/src/upgrade/release-3.2.x-incubating.asciidoc
+++ b/docs/src/upgrade/release-3.2.x-incubating.asciidoc
@@ -112,18 +112,19 @@ Gremlin Javascript can now submit script, with optional bindings, using the `Cli
 [source,javascript]
 ----
 const gremlin = require('gremlin');
-const connection = new gremlin.driver.Client('ws://localhost:8182/gremlin', { traversalSource: 'g' });
+const client = new gremlin.driver.Client('ws://localhost:8182/gremlin', { traversalSource: 'g' });
 
-connection.submit('g.V().tail()')
-    .then(response => {
-        console.log(response.traversers.length);
-        console.log(response.traversers[0]);
+client.submit('g.V().tail()')
+    .then(result => {
+        console.log(result.length);
+        console.log(result.toArray()[0]);
     });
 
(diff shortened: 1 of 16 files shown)


jorgebay:
VOTE +1

docs/src/upgrade/release-3.2.x-incubating.asciidoc
```diff
@@ -112,18 +112,19 @@ Gremlin Javascript can now submit script, with optional bindings, using the `Cli
...
 ----
 const gremlin = require('gremlin');
-const connection = new gremlin.driver.Client('ws://localhost:8182/gremlin', { traversalSource: 'g' });
+const client = new gremlin.driver.Client('ws://localhost:8182/gremlin', { traversalSource: 'g' });
 
-connection.submit('g.V().tail()')
-    .then(response => {
-        console.log(response.traversers.length);
-        console.log(response.traversers[0]);
+client.submit('g.V().tail()')
```
spmallette: we usually don't edit past upgrade notes - they are static and bound to that version, but i think i see what you're trying to do here, i.e. hide away `traversers`....i guess that's ok

jorgebay: Yes, the idea is to hide a deprecated property :)

spmallette:
VOTE +1

PR #967, opened 2018-10-23
TINKERPOP-2077: Add default create() method to VertexProgram.Builder

PBGraff:
Adds a default create() method to the VertexProgram.Builder so that users don't need to specify null when no Graph is needed.


spmallette:
i've re-targetted the PR to tp33 because the tp32 branch is only for the worst bugs at this point.  restarted the travis build as well.

spmallette:
VOTE +1

PR #968, opened 2018-10-23
TINKERPOP-2078 Added AnonymousTraversalSource

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2078

Basically replaces `EmptyGraph.instance().withRemote()` for `traversal().withRemote()` which is a bit more intuitive.

```text
gremlin> g = traversal().withGraph(TinkerFactory.createModern())
gremlin> g = traversal().withRemote('conf/remote-graph.properties')
==>graphtraversalsource[emptygraph[empty], standard]
gremlin> g = traversal().withRemote(DriverRemoteConnection.using('localhost',8182))
==>graphtraversalsource[emptygraph[empty], standard]
```

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,8 +26,9 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 This release also includes changes from <<release-3-2-11, 3.2.11>>.
 
 * Fixed `PersistedOutputRDD` to eager persist RDD by adding `count()` action calls.
-* gremlin-python: the graphson deserializer for g:Set should return a python set
+* Deserialized `g:Set` to a Python `Set` in GraphSON in `gremlin-python`.
 * Display the remote stack trace in the Gremlin Console when scripts sent to the server fail.
+* Added `AnonymousTraversalSource` which provides a more unified means of constructing a `TraversalSource`.
 * Changed behavior of GraphSON deserializer in gremlin-python such that `g:Set` returns a Python `Set`.
 * Changed behavior of `iterate()` in Python, Javascript and .NET to send `none()` thus avoiding unnecessary results being returned.
 
--- a/docs/src/reference/gremlin-variants.asciidoc
+++ b/docs/src/reference/gremlin-variants.asciidoc
@@ -100,7 +100,7 @@ Gremlin-Python users will typically make use of the following classes.
 
 [source,python]
 >>> from gremlin_python import statics
->>> from gremlin_python.structure.graph import traversal
+>>> from gremlin_python.process.anonymous_traversal_source import traversal
 >>> from gremlin_python.process.graph_traversal import __
 >>> from gremlin_python.process.strategies import *
 >>> from gremlin_python.driver.driver_remote_connection import DriverRemoteConnection
@@ -485,11 +485,11 @@ in Gremlin-Java. The `GraphTraversalSource` requires a RemoteConnection implemen
 [source,javascript]
 ----
 const gremlin = require('gremlin');
-const traversal = gremlin.structure.Graph.traversal;
+const traversal = gremlin.process.AnonymousTraversalSource.traversal;
 const DriverRemoteConnection = gremlin.driver.DriverRemoteConnection;
 ----
 
-A traversal source can be spawned with `RemoteStrategy` from an empty `Graph`.
+A traversal source can be spawned with `RemoteStrategy` from an `AnonymousTraversalSource`.
 
 [source,javascript]
 ----
--- a/docs/src/reference/gremlin-variants.asciidoc
+++ b/docs/src/reference/gremlin-variants.asciidoc
@@ -100,7 +100,7 @@ Gremlin-Python users will typically make use of the following classes.
 
 [source,python]
 >>> from gremlin_python import statics
->>> from gremlin_python.structure.graph import Graph
+>>> from gremlin_python.structure.graph import traversal
 >>> from gremlin_python.process.graph_traversal import __
 >>> from gremlin_python.process.strategies import *
 >>> from gremlin_python.driver.driver_remote_connection import DriverRemoteConnection
@@ -155,11 +155,8 @@ $ bin/gremlin-server.sh conf/gremlin-server-modern-py.yaml
 NOTE: The command to use `install` need only be executed once to gather `gremlin-python` dependencies into Gremlin Servers'
 path. Future starts of Gremlin Server will not require that command.
 
-Within the CPython console, an empty `Graph` is created and a traversal source is spawned with `RemoteStrategy`.
-
 [source,python]
->>> graph = Graph()
->>> g = graph.traversal().withRemote(DriverRemoteConnection('ws://localhost:8182/gremlin','g'))
+>>> g = traversal().withRemote(DriverRemoteConnection('ws://localhost:8182/gremlin','g'))
 
 When a traversal from the `GraphTraversalSource` is iterated, the traversal's `Bytecode` is sent over the wire
 via the registered `RemoteConnection`. The bytecode is used to construct the equivalent traversal at the remote traversal source.
--- a/docs/src/reference/gremlin-variants.asciidoc
+++ b/docs/src/reference/gremlin-variants.asciidoc
@@ -155,6 +155,9 @@ $ bin/gremlin-server.sh conf/gremlin-server-modern-py.yaml
 NOTE: The command to use `install` need only be executed once to gather `gremlin-python` dependencies into Gremlin Servers'
 path. Future starts of Gremlin Server will not require that command.
 
+Within CPython console, a `GraphTraversalSource` is created from the anonymous `traversal()` method where the "g"
+provided to the `DriverRemoteConnection` corresponds to the name of a `GraphTraversalSource` on the remote end.
+
 [source,python]
 >>> g = traversal().withRemote(DriverRemoteConnection('ws://localhost:8182/gremlin','g'))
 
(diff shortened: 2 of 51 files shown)


CHANGELOG.asciidoc
```diff
@@ -23,6 +23,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 [[release-3-3-5]]
 === TinkerPop 3.3.5 (Release Date: NOT OFFICIALLY RELEASED YET)
 
+* Added `AnonymouseTraversalSource` which provides a more unified means of constructing a `TraversalSource`.
```
newkek: nit: "Anonymouse"

spmallette: i thought i fixed that!

newkek:
Oh actually, I may have gotten confused. Is the goal of this to also deprecate `graph.traversal()`?

spmallette:
I dont' think i want to deprecate that as of yet even though we have a way create `TraversalSource` from local graphs this way.  I will make all of this much more clear in 3.4.0 docs since they will be `/current` docs on next release and they will be the focus of things.

jorgebay:
lgtm, I like the new approach.

I'm +1 on the gremlin-javascript implementation.

On the C# side, I personally don't like the `Traversal_()` workaround much, we could name it differently or look at alternatives (like named `using` directives), but I'll let @FlorianHockmann have a look at it first.

FlorianHockmann:
I'm not sure I'm following what the problem exactly is. 

> C# preventing static import of Traversal in some cases mistaking it for a package name

But all you have to statically import is the class which is `Gremlin.Net.Process.Traversal.AnonymousTraversalSource` (taken from your changes in this PR). So, why doesn't it work if the method is named `Traversal`? I just renamed `Traversal_()` locally to `Traversal()` and didn't run into any problems.

In general, I would also like to avoid a workaround like `Traversal_()` if possible.

spmallette:
> In general, I would also like to avoid a workaround like Traversal_() if possible.

my IDE was complaining - i guess i can try again and remove it since it worked for you.

spmallette:
@FlorianHockmann when i change it i get errors from the compiler like this:

> Process/Traversal/GraphTraversalSourceTests.cs(36,21): error CS0118: 'Gremlin.Net.UnitTest.Process.Traversal' is a namespace but is used like a variable [/home/smallette/git/apache/incubator-tinkerpop/gremlin-dotnet/test/Gremlin.Net.UnitTest/Gremlin.Net.UnitTest.csproj]

did you change something else to make it work? if you have a fix, just let me know and I'll stand back so you can push it directly to the branch....

FlorianHockmann:
I just used the rename refactoring from my IDE and I was able to compile, but let me try again and then push the change if everything works, including the tests.

FlorianHockmann:
Now I see what you mean. My IDE simply fixed the problem during the rename refactoring by adding `AnonymousTraversalSource.` for all calls to `Traversal()` where you're getting the error.

I had only inspected the changes in 2 classes manually and it worked in both. I think the problem is if the namespace of the test also includes the `Traversal` part.

I guess it's ok to write those tests as:
```cs
var g = AnonymousTraversalSource.Traversal().WithRemote(connection);
```
This doesn't really affect users as long as they don't also use the word _Traversal_ in their namespaces.

One other idea in general for the change in this issue: What if we rename `AnonymousTraversalSource` to just `Graph` and then also don't recommend the `static using` here? This would result in:

```cs
var g = Graph.Traversal().WithRemote(connection);
```

which is nearly the same as what we had before but a bit cleaner than:

```cs
var g = new Graph().Traversal().WithRemote(connection);
```

Calling the method `Traversal()` without showing on what it is called looks a bit strange in general in my opinion. Just using a static class `Graph` and showing that `Traversal()` is a member of `Graph` avoids this.

spmallette:
> One other idea in general for the change in this issue: What if we rename AnonymousTraversalSource

well, the idea is to construct the `TraversalSource` in an anonymous fashion (i.e. without a graph or remote) so putting `Graph` back in the mix doesn't seem right to me even if called statically. out of necessity i placed `traversal()` on `Graph` in js/python but with C# being the most analogous to Java in terms of project structure I feel less happy about doing so here. 

I'd be ok with the other idea of just going with `Traversal()` and referencing the full name only as needed in the tests given that the problem only happens when "Traversal" is in the package name.

newkek:
> Calling the method Traversal() without showing on what it is called looks a bit strange in general in my opinion.

I agree, in general implying a static import in a code example can lead to confusion. If you don't know the context you never really know where the method/static member comes from (the proof is with all the people asking questions about missing static imports in the GLVs lately).
It would be less confusing if we were to include the class' name in the example in addition to the static method. Then, `AnonymousTraversalSource.[T/t]raversal()` isn't extremely gorgeous so I was thinking we could have used a more generic name for the class like `TinkerPop` or `Gremlin`, just my 2 cents

FlorianHockmann:
Irrespective of the discussion about the API in general, I just pushed the changes to drop the `_` suffix for `Traversal()`.

spmallette:
thanks @FlorianHockmann - hopefully i can focus on finishing touches and get this ready for actual review for tomorrow.

@newkek I appreciate the point about users messing it up, but static imports and this style we have are the basis for Gremlin. personally, I'd rather not try to tackle that problem with this ticket.  fwiw, i'm making a good effort to try to bring better attention to the whole "import" issue in the 3.4.0 docs. hopefully that will help.

spmallette:
dah - @FlorianHockmann  i messed up. i didn't integrate your change after my own rebase to fix the changelog conflict. can you please rebase and force push your changes for this branch again (don't pull this remote branch).

spmallette:
@FlorianHockmann never mind - i just did the refactor myself - hope you don't mind.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/remote/DriverRemoteConnection.java
```diff
@@ -101,6 +101,24 @@ private DriverRemoteConnection(final Cluster cluster, final boolean tryCloseClus
         this.conf = Optional.of(conf);
     }
 
+    /**
+     * Creates a {@link DriverRemoteConnection} using a new {@link Cluster} instance created from the supplied
```
FlorianHockmann: (nitpick) _from the supplied configuration file_ <- This overload takes no configuration file.
(Same for the other overload a few lines below.)

spmallette:
All good suggestions @FlorianHockmann - I believe I've implemented all of them and pushed now.

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -477,24 +485,23 @@ in Gremlin-Java. The `GraphTraversalSource` requires a RemoteConnection implemen
 [source,javascript]
 ----
 const gremlin = require('gremlin');
-const Graph = gremlin.structure.Graph;
+const traversal = gremlin.structure.Graph.traversal;
```
jorgebay: This should be exposed at:`gremlin.process.traversal`, to match java's `org.apache.tinkerpop.gremlin.process.traversal`

```suggestion
const traversal = gremlin.process.traversal;
```

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -100,7 +100,7 @@ Gremlin-Python users will typically make use of the following classes.
 
 [source,python]
 >>> from gremlin_python import statics
->>> from gremlin_python.structure.graph import Graph
+>>> from gremlin_python.structure.graph import traversal
```
jorgebay: To match java/C#/javascript, I think we should expose it under `process` not `structure`.

gremlin-javascript/src/main/javascript/gremlin-javascript/index.js
```diff
@@ -72,7 +72,8 @@ module.exports = {
     GraphTraversal: gt.GraphTraversal,
     GraphTraversalSource: gt.GraphTraversalSource,
     statics: gt.statics,
-    Translator
+    Translator,
+    traversal: gt.traversal
```
jorgebay: This is `undefined`, its currently exposed under `./lib/structure/graph`.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.js
```diff
@@ -158,11 +159,20 @@ function areEqual(obj1, obj2) {
   return false;
 }
 
+/**
+* Returns an anonymous traversal source.
+* @returns {GraphTraversalSource}
+*/
+function traversal() {
```
jorgebay: This method is being implemented on `./lib/structure/graph.js` when it should be under `lib/process` directory.

spmallette:
@jorgebay i tried to put `traversal()` in process for python and js. in python i ran into some cyclic dependency problems. in js i had other issues (that i'd assumed were similar to python). That's why the `traversal()` method is where it is in those cases.  I just attempted to move it again for js and keep getting weird errors i don't understand. I really shouldn't spend any more time floundering around on this one - could you please take a look at moving it for me (just push the fix right to the branch)?

If you get it working, i'll look at what i can do in python to move it, even if it means creating an `AnonymousTraversalSource` sort of py file to do it.

jorgebay:
I'll try to look at it today or tomorrow.

spmallette:
I think that this is "final" now. I just pushed the changes to python to include a form of `AnonymousTraversalSource` like the javascript approach @jorgebay added. Fixed up docs. Hopefully, everyone is happy with this one now and we can merge. I'd like to do that tomorrow as this issue is holding up other work.

PR #971, opened 2018-10-25
TINKERPOP-2062 added Traversal class to CoreImports

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-2062

Not really worth a PR, could have been a CTR, but just in case someone has any complaints about adding this import.

VOTE +1


spmallette:
VOTE +1

jorgebay:
sgtm, VOTE +1

PR #972, opened 2018-10-26
TINKERPOP-2067 Allow getting raw JToken data from Gremlin.Net.Driver.GremlinClient.

danielcweber:
https://issues.apache.org/jira/browse/TINKERPOP-2067

### DIFF
--- a/gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Driver/GremlinClientTests.cs
+++ b/gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Driver/GremlinClientTests.cs
@@ -29,6 +29,7 @@ using Gremlin.Net.Driver;
 using Gremlin.Net.Driver.Exceptions;
 using Gremlin.Net.Driver.Messages;
 using Gremlin.Net.IntegrationTest.Util;
+using Newtonsoft.Json.Linq;
 using Xunit;
 
 namespace Gremlin.Net.IntegrationTest.Driver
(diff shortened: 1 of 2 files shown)


gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Driver/GremlinClientTests.cs
```diff
@@ -68,6 +69,20 @@ public async Task ShouldHandleBigResponse()
             }
         }
 
+        [Fact]
+        public async Task ShouldHandle_JToken_Request()
```
FlorianHockmann: Another nit, please rename the test to be consistent with the other ones, e.g., no underscores (except when they just describe the tested traversal which isn't the case here) and the test name could a bit more specific. How about something like `ShouldReturnResultWithoutDeserializingItForJTokenType`?

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Driver/GremlinClientTests.cs
```diff
@@ -68,6 +69,20 @@ public async Task ShouldHandleBigResponse()
...
+        {
+            var gremlinServer = new GremlinServer(TestHost, TestPort);
+            using (var gremlinClient = new GremlinClient(gremlinServer))
+            {
+                var gremlinScript = "'someString'";
+                
+                var response = await gremlinClient.SubmitWithSingleResultAsync<JToken>(gremlinScript);
```
FlorianHockmann: I'm not sure but wouldn't this test also succeed without your change as the returned data is just a string here? Maybe it would make more sense to return something that includes a type in its GraphSON form like an `int`?

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Driver/GremlinClientTests.cs
```diff
@@ -68,6 +69,20 @@ public async Task ShouldHandleBigResponse()
...
+
+                Assert.NotNull(response);
```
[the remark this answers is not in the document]
danielcweber: Done, more assertions now!

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Driver/GremlinClientTests.cs
danielcweber: Done.

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Driver/GremlinClientTests.cs
danielcweber: It would not anymore since there's more assertions now about the actual JObject that's returned, which is not just a plain string but

    {
        "@type": "g:List",
        "@value": [
            "someString"
         ]
    }

FlorianHockmann: Hmm, didn't realize that the result would be wrapped in a list in cases like this, but you obviously proved that it's the case with the added asserts.

dkuppitz:
VOTE +1

FlorianHockmann:
Sure, I can take care of the merge, but if you want to do it, then by all means go ahead! It's not like I own the code or anything 😅

PR #981, opened 2018-10-31
TINKERPOP-2084 Display remote stacktrace in gremlin console

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2084

Making the remote stacktrace visible to the user saves a visit to the server to determine what went amiss with failed script when the remote failing message is not enough to debug the problem. Plus, it makes the Gremlin Console in ":remote console" mode behave more like it is in local mode.

Builds with `mvn clean install -pl gremlin-console -DskipIntegrationTests=false`

VOTE +1


jorgebay:
Thats super useful, VOTE +1

PR #982, opened 2018-10-31
TINKERPOP-2072 Refactored TypeTranslator for direction extension

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2072

This approach makes it much easier to extend `TypeTranslator` as the core functionality of the `GroovyTranslator` is itself a `TypeTranslator`. Simply extend that and go. Probably should have been this way from the beginning. This is a breaking change but on a fairly hidden/internal component - should only affect providers.

VOTE +1


dkuppitz:
VOTE +1

PR #983, opened 2018-11-02
TINKERPOP-2069 Document Gremlin.Net driver config

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2069

This simply adds documentation on how to configure the Gremlin.Net driver to the reference docs. It also fixes a small mistake in the provider docs where we forgot to change the classes for GraphSON3 and it adds `localhost` as the default `hostname` to be consistent with the Java driver.

I built the docs with `./docker/build.sh -d`.

VOTE +1

### DIFF
--- a/docs/src/dev/provider/index.asciidoc
+++ b/docs/src/dev/provider/index.asciidoc
@@ -576,9 +576,9 @@ internal class MyTypeReader : IGraphSONDeserializer
     }
 }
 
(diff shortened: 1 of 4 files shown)


spmallette:
VOTE +1

dkuppitz:
VOTE +1

PR #984, opened 2018-11-02
TINKERPOP-2065 iterate() now sends NoneStep

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2065

Fixed this for python, .NET and js. This means that results will no longer be sent back from the server to the client (i.e. no client side iteration).

VOTE +1


FlorianHockmann:
Nice to get this optimization in the GLVs!

VOTE +1

jorgebay:
Nice! VOTE +1

PR #985, opened 2018-11-05
TINKERPOP-2064 Expose response attributes in JavaScript Driver

jorgebay:
https://issues.apache.org/jira/browse/TINKERPOP-2064

Exposed response attributes in the `ResultSet` class of the JavaScript Driver.

Additionally, I've taken the opportunity to remove the deprecated property `traversers` from `ResultSet`.

VOTE +1


spmallette:
VOTE +1

PR #986, opened 2018-11-05
TINKERPOP-2059 Modulation of valueMap()

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-2059

Implemented `by()` and `with()` modulators for `valueMap()` step.

`docker/build.sh -t -i -n` passed and docs generated via `docker/build.sh -d` look good.

VOTE +1

### DIFF
--- a/docs/src/upgrade/release-3.4.x.asciidoc
+++ b/docs/src/upgrade/release-3.4.x.asciidoc
@@ -29,6 +29,76 @@ Please see the link:https://github.com/apache/tinkerpop/blob/3.4.0/CHANGELOG.asc
 
 === Upgrading for Users
 
+==== Modulation of valueMap()
+
+The `valueMap()` step now supports `by` and `with` modulation, which also led to the deprecation of `valueMap(true)` overloads.
+
+===== by() modulation
+
+With the help of the `by()` modulator `valueMap()` result values can now be adjusted, which is particularly useful to turn multi-/list-values into single values.
+
+```
+gremlin> g.V().hasLabel("person").valueMap()
+==>[name:[marko],age:[29]]
+==>[name:[vadas],age:[27]]
+==>[name:[josh],age:[32]]
+==>[name:[peter],age:[35]]
+gremlin> g.V().hasLabel("person").valueMap().by(unfold())
+==>[name:marko,age:29]
+==>[name:vadas,age:27]
+==>[name:josh,age:32]
+==>[name:peter,age:35]
+```
+
+===== with() modulation
+
+The `with()` modulator can be used to include certain tokens (`id`, `label`, `key` and/or `value`).
+
+The old way (still valid, but deprecated):
+
+```
+gremlin> g.V().hasLabel("software").valueMap(true)
+==>[id:10,label:software,name:[gremlin]]
+==>[id:11,label:software,name:[tinkergraph]]
+gremlin> g.V().has("person","name","marko").properties("location").valueMap(true)
+==>[id:6,key:location,value:san diego,startTime:1997,endTime:2001]
+==>[id:7,key:location,value:santa cruz,startTime:2001,endTime:2004]
+==>[id:8,key:location,value:brussels,startTime:2004,endTime:2005]
+==>[id:9,key:location,value:santa fe,startTime:2005]
+```
+
+The new way:
+
+```
+gremlin> g.V().hasLabel("software").valueMap().with(WithOptions.tokens)
+==>[id:10,label:software,name:[gremlin]]
+==>[id:11,label:software,name:[tinkergraph]]
+gremlin> g.V().has("person","name","marko").properties("location").valueMap().with(WithOptions.tokens)
+==>[id:6,key:location,value:san diego,startTime:1997,endTime:2001]
+==>[id:7,key:location,value:santa cruz,startTime:2001,endTime:2004]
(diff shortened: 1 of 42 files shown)


docs/src/upgrade/release-3.4.x.asciidoc
```diff
@@ -29,6 +29,70 @@ Please see the link:https://github.com/apache/tinkerpop/blob/3.4.0/CHANGELOG.asc
...
+gremlin> g.V().hasLabel("software").valueMap().with(WithOptions.tokens, WithOptions.labels)
+==>[label:software,name:[gremlin]]
+==>[label:software,name:[tinkergraph]]
+gremlin> g.V().has("person","name","marko").properties("location").valueMap().with(WithOptions.tokens, WithOptions.values)
+==>[value:san diego,startTime:1997,endTime:2001]
+==>[value:santa cruz,startTime:2001,endTime:2004]
+==>[value:brussels,startTime:2004,endTime:2005]
+==>[value:santa fe,startTime:2005]
+```
+
```
spmallette: needs a "See:" with the JIRA link.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -560,7 +560,7 @@
      * @see <a href="http://tinkerpop.apache.org/docs/${project.version}/reference/#valuemap-step" target="_blank">Reference Documentation - ValueMap Step</a>
      * @since 3.0.0-incubating
      */
-    public default <E2> GraphTraversal<S, Map<String, E2>> valueMap(final String... propertyKeys) {
+    public default <E2> GraphTraversal<S, Map<Object, E2>> valueMap(final String... propertyKeys) {
```
spmallette: I suppose this ticket was marked as "breaking" because of this signature change? I think it would be worth explaining this change a bit in the upgrade docs so that people understand that the generics are different.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -576,6 +576,7 @@
      * @see <a href="http://tinkerpop.apache.org/docs/${project.version}/reference/#valuemap-step" target="_blank">Reference Documentation - ValueMap Step</a>
      * @since 3.0.0-incubating
      */
+    @Deprecated
```
spmallette: needs javadoc "@deprecated As of release 3.4.0, replaced by {@link ...}" also needed below in the `__` class.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/PropertyMapStepTest.java
```diff
@@ -37,10 +38,12 @@
                 __.valueMap("name"),
                 __.valueMap("age"),
                 __.valueMap("name", "age"),
-                __.valueMap(true),
-                __.valueMap(true, "name"),
-                __.valueMap(true, "age"),
-                __.valueMap(true, "name", "age"),
```
spmallette: I don't think we should remove any old tests for the old signature. Those still need to be tested since we've only deprecated the approach.

gremlin-test/features/map/AddVertex.feature
```diff
@@ -336,7 +336,7 @@ Feature: Step - addV()
       | marko |
     And the graph should return 2 for count of "g.V().has(\"name\",\"marko\")"
 
-  Scenario: g_V_asXaX_hasXname_markoX_outXcreatedX_asXbX_addVXselectXaX_labelX_propertyXtest_selectXbX_labelX_valueMapXtrueX
+  Scenario: g_V_asXaX_hasXname_markoX_outXcreatedX_asXbX_addVXselectXaX_labelX_propertyXtest_selectXbX_labelX_valueMap_withXtokensX
```
spmallette: Mentioned this earlier, but let's keep the `valueMap(true)` tests since we have to maintain that deprecated code. There are other places as well obviously where the `valueMap(true)` tests need to be brought back. Just duplicate the test basically - one for the old way and one for the new way.

dkuppitz: I kept the boolean overload tests in `ValueMapTest`. I don't think `AddVertexTest` should be used to test several `valueMap` overloads, but ... I'll add them back.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/PropertyMapStepTest.java
dkuppitz: Can't mix those here, the hashCodes collide.

gremlin-test/features/map/AddVertex.feature
spmallette: ok - as long as we at least exercise the old stuff sufficiently then no need to bring them all back. I just saw these removed and didn't realize that there were others still hanging about.

spmallette:
nice job - VOTE +1

PR #993, opened 2018-11-11
TINKERPOP-2051 Uniqueness of property ids

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-2051

*At this point this  PR is for discussions only.*

`docker/build.sh -t -i -n` almost passed - `GryoCompatibilityTest` seems to be the only test that's broken, but thus far I have no idea how to fix that.

**NOTES**

The change in `ChooseTest` was necessary as it was randomly failing for Neo4j (fails consistently with random seed `1541860031806L`). I'm not sure if this was really related to my changes, but since it looked like it was only the assertion that was bad, I just fixed it instead of trying it on another branch.

The adjustment in `DetachedVertexProperty`'s (de)serialization was necessary to make everything work in Spark. I don't know what this change would break in `tp33/`, hence this PR only targets `master/`.


spmallette:
JIRA for this one is closed - reasoning in the comments.

PR #999, opened 2018-11-19
some improvements

nastra:
tested locally with `mvn clean install`

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversalStrategies.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversalStrategies.java
@@ -25,25 +25,23 @@ import org.apache.tinkerpop.gremlin.structure.util.StringFactory;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 
 /**
  * @author Marko A. Rodriguez (http://markorodriguez.com)
  */
 public class DefaultTraversalStrategies implements TraversalStrategies {
 
-    protected List<TraversalStrategy<?>> traversalStrategies = new ArrayList<>();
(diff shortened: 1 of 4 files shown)


spmallette:
I think these changes make sense presuming that we don't register the same strategy more than once, which I don't think we do - @dkuppitz please think that one through a bit - is there any situation where we allow for that.  Other than that, initializing the `ArrayList` in `flattenArguments()` and going to `LinkedHashSet` to avoid some looping seem to make sense to me.

VOTE +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversalStrategies.java
```diff
@@ -25,25 +25,23 @@
...
     @Override
     @SuppressWarnings({"unchecked", "varargs"})
     public TraversalStrategies addStrategies(final TraversalStrategy<?>... strategies) {
-        final List<TraversalStrategy<?>> concurrent = new ArrayList<>(this.traversalStrategies);
         for (final TraversalStrategy<?> addStrategy : strategies) {
-            for (final TraversalStrategy<?> currentStrategy : concurrent) {
-                if (addStrategy.getClass().equals(currentStrategy.getClass()))
-                    this.traversalStrategies.remove(currentStrategy);
-            }
+            this.traversalStrategies.remove(addStrategy);
```
dkuppitz: Do we really need this loop? In a `Set` we can just keep adding strategies, nothing would happen if a strategy was already added.

nastra: I don't know, I'm just keeping the same functionality but with a different data structure.

dkuppitz: Yea, I just tried to get rid of the loop, but that doesn't work because at some point we rely on a replacement behavior.

dkuppitz:
VOTE +1

PR #1001, opened 2018-11-21
uses the directly field instead of getter method

otaviojava:
Hello everyone, I still studying the code to help more in the project, then I found a small performance improvement. That is access the field directly instead of using the getter setter.

```java
@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
@Measurement(iterations = 20, time = 1, timeUnit = TimeUnit.SECONDS)
@Fork(3)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@State(Scope.Thread)
public class ClientBenchmark {


    private final ObjectMapper mapper = GraphSONMapper.build().version(GraphSONVersion.V3_0).create().createMapper();
    private final Graph graph = TinkerFactory.createModern();
    private final Optional<Vertex> vertex = graph.traversal().V().tryNext();

    @Setup
    public void setup() {
    }

    @Benchmark
    public String write() throws JsonProcessingException {
        return mapper.writeValueAsString(vertex);
    }


}
```

* Benchmark.write  avgt   60  192,218  ns/op
* Benchmark.write  avgt   60  212,345  ns/op (around 10% faster)

```java
@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
@Measurement(iterations = 20, time = 1, timeUnit = TimeUnit.SECONDS)
@Fork(3)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@State(Scope.Thread)
public class ClientBenchmark {


    private final ObjectMapper mapper = GraphSONMapper.build().version(GraphSONVersion.V3_0).create().createMapper();
    private final Graph graph = TinkerFactory.createModern();
    private final Vertex vertex = graph.traversal().V().next();

    @Setup
    public void setup() {
    }

    @Benchmark
    public String write() throws JsonProcessingException {
        return mapper.writeValueAsString(vertex);
    }


}
```

Original code:    avgt   60  1938,602 ± 50,353  ns/op
The code changed: avgt   60  2144,034 ± 51,190  ns/op  (10% faster)

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/AbstractGremlinScriptEngineFactory.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/AbstractGremlinScriptEngineFactory.java
@@ -89,13 +89,13 @@ public abstract class AbstractGremlinScriptEngineFactory implements GremlinScrip
     @Override
     public Object getParameter(final String key) {
         if (key.equals(ScriptEngine.ENGINE)) {
-            return this.getEngineName();
+            return engineName;
         } else if (key.equals(ScriptEngine.ENGINE_VERSION)) {
             return this.getEngineVersion();
         } else if (key.equals(ScriptEngine.NAME)) {
             return engineName;
         } else if (key.equals(ScriptEngine.LANGUAGE)) {
-            return this.getLanguageName();
+            return languageName;
         } else if (key.equals(ScriptEngine.LANGUAGE_VERSION)) {
             return this.getLanguageVersion();
         } else
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/DefaultGremlinScriptEngineManager.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/DefaultGremlinScriptEngineManager.java
@@ -428,7 +428,7 @@ public class DefaultGremlinScriptEngineManager implements GremlinScriptEngineMan
 
                     globalScope.put(kv.getKey(), kv.getValue());
                 });
-        engine.setBindings(getBindings(), ScriptContext.GLOBAL_SCOPE);
+        engine.setBindings(globalScope, ScriptContext.GLOBAL_SCOPE);
 
         // merge in bindings that are marked with engine scope. there typically won't be any of these but it's just
         // here for completeness. bindings will typically apply with global scope only as engine scope will generally
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/GraphFilter.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/GraphFilter.java
@@ -319,9 +319,9 @@ public final class GraphFilter implements Cloneable, Serializable {
     public boolean equals(final Object object) {
         if (!(object instanceof GraphFilter))
             return false;
-        else if (((GraphFilter) object).hasVertexFilter() && !((GraphFilter) object).getVertexFilter().equals(this.vertexFilter))
+        else if (((GraphFilter) object).hasVertexFilter() && !((GraphFilter) object).vertexFilter.equals(this.vertexFilter))
             return false;
-        else if (((GraphFilter) object).hasEdgeFilter() && !((GraphFilter) object).getEdgeFilter().equals(this.edgeFilter))
+        else if (((GraphFilter) object).hasEdgeFilter() && !((GraphFilter) object).edgeFilter.equals(this.edgeFilter))
             return false;
         else
             return true;
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/bulkloading/IncrementalBulkLoader.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/bulkloading/IncrementalBulkLoader.java
@@ -49,12 +49,12 @@ public class IncrementalBulkLoader implements BulkLoader {
     public Vertex getOrCreateVertex(final Vertex vertex, final Graph graph, final GraphTraversalSource g) {
         final Iterator<Vertex> iterator = useUserSuppliedIds()
                 ? g.V().hasId(vertex.id())
-                : g.V().has(vertex.label(), getVertexIdProperty(), vertex.id().toString());
+                : g.V().has(vertex.label(), bulkLoaderVertexId, vertex.id().toString());
         return iterator.hasNext()
                 ? iterator.next()
                 : useUserSuppliedIds()
                 ? g.addV(vertex.label()).property(T.id, vertex.id()).next()
-                : g.addV(vertex.label()).property(getVertexIdProperty(), vertex.id().toString()).next();
+                : g.addV(vertex.label()).property(bulkLoaderVertexId, vertex.id().toString()).next();
     }
 
     /**
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/traversal/MemoryTraversalSideEffects.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/traversal/MemoryTraversalSideEffects.java
@@ -166,7 +166,7 @@ public final class MemoryTraversalSideEffects implements TraversalSideEffects {
         final Set<MemoryComputeKey> keys = new HashSet<>();
         final TraversalSideEffects sideEffects =
                 traversal.getSideEffects() instanceof MemoryTraversalSideEffects ?
-                        ((MemoryTraversalSideEffects) traversal.getSideEffects()).getSideEffects() :
+                        ((MemoryTraversalSideEffects) traversal.getSideEffects()).sideEffects :
                         traversal.getSideEffects();
         sideEffects.keys().
                 stream().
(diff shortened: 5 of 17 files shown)


spmallette:
Of all the changes here, the most compelling one is the change made to `GraphSONTypeIdResolver` (the rest occur in methods that I don't think get called frequently enough to really make a noticeable impact in performance - just guessing from a cursory glance at this PR).  You provided a small general benchmark which is nice, but could you please specifically do a benchmark over the `GraphSONTypeIdResolver` to see what kind of performance improvement we get there as a result of this?

You can try to benchmark the it directly but you could probably also just instantiate a `ObjectMapper` for GraphSON 3.0 and fire a bunch of different GraphSON types at it in succession to do the benchmark.  Just an example with a `Vertex` - the full list of GraphSON types is [here](http://tinkerpop.apache.org/docs/3.3.4/dev/io/)

```text
gremlin> mapper = GraphSONMapper.build().version(GraphSONVersion.V3_0).create().createMapper()
==>org.apache.tinkerpop.shaded.jackson.databind.ObjectMapper@352e612e
gremlin> x = mapper.writeValueAsString(g.V(1).next())
==>{"@type":"g:Vertex","@value":{"id":{"@type":"g:Int32","@value":1},"label":"person","properties":{"name":[{"@type":"g:VertexProperty","@value":{"id":{"@type":"g:Int64","@value":0},"value":"marko","label":"name"}}],"age":[{"@type":"g:VertexProperty","@value":{"id":{"@type":"g:Int64","@value":1},"value":{"@type":"g:Int32","@value":29},"label":"age"}}]}}}
gremlin> v = mapper.readValue(x, Vertex.class)
==>v[1]
```

if you demonstrate a speed improvement there that would be huge :smile:

otaviojava:
Is that right?

```java
@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
@Measurement(iterations = 20, time = 1, timeUnit = TimeUnit.SECONDS)
@Fork(3)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@State(Scope.Thread)
public class ClientBenchmark {


    private final ObjectMapper mapper = GraphSONMapper.build().version(GraphSONVersion.V3_0).create().createMapper();
    private final Graph graph = TinkerFactory.createModern();
    private final Optional<Vertex> vertex = graph.traversal().V().tryNext();

    @Setup
    public void setup() {
    }

    @Benchmark
    public String write() throws JsonProcessingException {
        return mapper.writeValueAsString(vertex);
    }


}
```


* Benchmark.write  avgt   60  192,218  ns/op
* Benchmark.write  avgt   60  212,345  ns/op (around 10% slower)

spmallette:
That's not encouraging. Why are you serializing `Optional` though? Why not just serialize the `Vertex`:

```text
Vertex vertex = graph.traversal().V().next();
```

otaviojava:
Ops, You're right. My mistake sorry for that.
I created a new test class, that looks 10% faster than original code.

```java
@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
@Measurement(iterations = 20, time = 1, timeUnit = TimeUnit.SECONDS)
@Fork(3)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@State(Scope.Thread)
public class ClientBenchmark {


    private final ObjectMapper mapper = GraphSONMapper.build().version(GraphSONVersion.V3_0).create().createMapper();
    private final Graph graph = TinkerFactory.createModern();
    private final Vertex vertex = graph.traversal().V().next();

    @Setup
    public void setup() {
    }

    @Benchmark
    public String write() throws JsonProcessingException {
        return mapper.writeValueAsString(vertex);
    }


}
```

Original code:    avgt   60  1938,602 ± 50,353  ns/op
The code changed: avgt   60  2144,034 ± 51,190  ns/op  (10% faster)

otaviojava:
@spmallette Actually, sorry that was my mistake. Both results show that became faster 10%. The JMH, the framework benchmark, the time is defined as score points and higher means good.
I've updated both comments.
English is not my first language, I did this mistake.

spmallette:
I see you updated that `Optional` benchmark. i didn't look at the numbers themselves very carefully and just paid attention to what you wrote - so there is no discrepancy to consider - nice. 

then, this is a pretty awesome PR for that GraphSON improvement.  specific changes to GraphSON serialization are a huge help and any others that you can come up with would be appreciated.

otaviojava:
You are welcome. I'm studying the code, and I hope to help more. If there is anything that I can help it will my pleasure.

robertdale:
I have a hard time following this thread and what the performance numbers mean.  The analysis in scores seem to contradict one another. In one instance, the lower time is 'faster' then later in another instance lower was 'slower'.  It was stated higher scores are better but it seems to me that with AverageTime, lower scores should be better. That is less time is faster.  Fewer nanoseconds per operation (ns/op) is faster.   Can someone summarize the final before and after scores?

spmallette:
The analysis contradicted itself earlier but from what I understood that was a typo  and it was corrected. where do you still see that to be the case @robertdale ?

robertdale:
The analysis now appears consistent in that higher times are marked as 'faster'. It's still not clear to me why a higher average time is 'faster'.

spmallette:
huh - the benchmark mode is

```text
@BenchmarkMode(Mode.AverageTime) 
```

so, i see where you're coming from now.  i thought i was looking at something like "op/sec" for some reason......... :unamused:  if that's the case, then yeah, it's gotten consistently slower for the change unless there's something i'm not understanding.

@otaviojava can you please comment?

otaviojava:
I'll retest again, I check

spmallette:
@otaviojava thanks for coming back to this. a few things:

1. nice to see that the throughput is 10% better given your revised testing
2. Is there any reason not to target this change at the `tp33` branch so that 3.3.5 can get this performance boost?
3. this branch is now conflicted - could you please rebase it, preferably on `tp33` given item 2 above.
4. I'm still a little curious as to why the  your run with `Mode.AverageTime` did not show a good result, but throughput did. do you have any comment on that?

once you've rebased, i'd like to conduct some of my own tests.

otaviojava:
@spmallette rebase done, I can create a PR for each one once this PR be merged

otaviojava:
retest this please

spmallette:
@otaviojava did you have any comments pertaining to my other questions (items 2 and 4 in my last comment)? If you have no thoughts on 2, then I think you should go ahead and rebase to `tp33` so that we get the benefit of this change on the 3.3.x line. As for 4, is there a simple explanation for that?

otaviojava:
2 - I did in master because of I usually from the master. But I can do a cherry-pick after the merge to other branches :) that is not a big deal.
3 - I checked several times, the honest answer I did not get how the average hows to do different match than throughput.

robertdale:
I wouldn't worry too much about the benchmarks. Micro benchmarks can sometimes be tricky to design and  interpret. It appears that Throughput is calculated over all threads (Fork=3) whereas AverageTime is per op, single thread so it's not the actual inverse of each other. Also for throughput, the actual boost here is (598/580) 3% not 10%.  The variance wasn't published but if it's anything like the 3% for Average Time, then the boost is marginal. I don't think it will be perceptible unless it's over a huge dataset but even then there are so many other things to contend with. The throughput benchmark points to how quickly that performance boost dissipates.

In any case, I think we can all agree that in general not having a method call overhead is going to be slightly faster.   As long as the code changes are sound, let's just take what we can get.

spmallette:
I've merged this as it is, though in the future I'd prefer to see PRs aimed at the earliest release branch for which it is intended and not the other way around where we have to backport changes. could you please issue a PR for `tp33` with the same changes?

otaviojava:
@robertdale @spmallette 

The JMH is a microbenchmark that the OpenJDK uses to performance improvement that is trustful.
The issue was a compiler trick that I did a mistake.
The code below will show the same result on both metrics:


```java
@Warmup(iterations = 5, time = 1)
@Measurement(iterations = 20, time = 1)
@Fork(3)
@BenchmarkMode(Mode.Throughput)
@OutputTimeUnit(TimeUnit.MILLISECONDS)
@State(Scope.Thread)
public class ClientBenchmark {


    private ObjectMapper mapper = GraphSONMapper.build().version(GraphSONVersion.V3_0).create().createMapper();
    private Graph graph = TinkerFactory.createModern();
    private Vertex vertex = graph.traversal().V().next();

    @Setup
    public void setup() {
        this.mapper = GraphSONMapper.build().version(GraphSONVersion.V3_0).create().createMapper();
        this.graph = TinkerFactory.createModern();
        this.vertex = graph.traversal().V().next();
    }

    @Benchmark
    public String write(Blackhole blackhole) throws JsonProcessingException {
        String value = mapper.writeValueAsString(vertex);
        blackhole.consume(value);
        return value;
    }

}

```

spmallette:
I believe that I need to revert/modify some of these changes. For classes that are designed for extension, like `GraphStep` it's possible that a method called in the subclass could be overridden to not use the field (somewhat unlikely in these cases but remotely possible) in which case the behavior we've had would be broken. I intend to go through each change and either fully revert it or in cases where the performance improvement is important (i.e. GraphSON) I will keep the change but add a `final` to the method declaration to ensure that it's not overridden with the expectation that it will alter behavior somewhere else. The use of `final` is technically a breaking change so it should probably only happen on `master`.  I'm not sure what that means for #1014 at this point. Maybe we should just keep this change to `master` only.

spmallette:
I put the proposed changes i alluded to in the previous comment on #1015

PR #1003, opened 2018-11-26
TINKERPOP-2100 coalesce() creating unexpected results when used with order()

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-2100

This PR fixes a bug in `CoalesceStep` which squared the bulk if the step followed a `Barrier` step.

VOTE +1


robertdale:
VOTE +1

PR #1004, opened 2018-11-27
TINKERPOP-2080 Removed withRemote from TraversalSource

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2080

Users should now be using `AnonymousTraversalSource` which is the more unified way to create the `TraversalSource` instance.

VOTE +1


dkuppitz:
VOTE +1

PR #1007, opened 2018-11-27
TINKERPOP-2018 Generate API docs for .NET

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2018

Docs generate on "generate-resources" phase in maven so `mvn clean install -pl :gremlin-dotnet-source` should do everything (i.e. download docfx, patch it for mono, then generate). Docs generate to `gremlin-dotnet/src/_site`. 

VOTE +1


spmallette:
You need a `.glv` file in `gremlin-dotnet/src` to trigger the profile in which case i'd wonder if you see this message printing?

https://github.com/apache/tinkerpop/pull/1007/files#diff-eb2d4cdb0a821b1a1e7166f54bc46d2aR144

Could that be the problem?

FlorianHockmann:
Do you have mono and the correct version of nuget.exe available? I think the step that downloads nuget.exe comes after the doc generation at the moment. So it could also fail because of that.
@spmallette: The step that downloads nuget.exe should be probably moved into the `docfx` generation to avoid this problem.

spmallette:
> The step that downloads nuget.exe should be probably moved into the docfx generation to avoid this problem.

I didn't do that on purpose i guess. I felt like the only people generating these docs would be those who were publishing them and thus those doing deployments who would have the setup. I just didn't want to complicate the regular build for the average user to pull in more stuff.

FlorianHockmann:
Ok, valid point. Then we just need to be aware that we need to run everything at least once on a new machine to download nuget.exe until the docs can be built in another run.

FlorianHockmann:
LGTM and thanks for taking care of this, @spmallette and @dkuppitz

VOTE +1

dkuppitz:
VOTE +1

PR #1008, opened 2018-11-28
TINKERPOP-2104 Allow ImportCustomizer to handle fields

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2104

We can now import `Field` instances to the `ScriptEngine` which allows for import of public fields from classes like `WithOptions`:

```text
gremlin> labels
==>2
```

Note that I only allowed for the ability to do this and didn't bother to add any new imports. For `WithOptions` it would just create a bunch of conflicts with existing tokens (e.g. keys, values, etc).

 VOTE +1


dkuppitz:
LGTM.

VOTE +1

jorgebay:
looks good :shipit: 
VOTE +1

PR #1009, opened 2018-11-29
TINKERPOP-2106 throw TimeoutException instead of TraversalInterruptedException

javeme:
https://issues.apache.org/jira/browse/TINKERPOP-2106

When gremlin executes timeout, it throws an error "TraversalInterruptedException: null".
According to our user feedback(https://github.com/hugegraph/hugegraph/issues/102), the message is not easy to understand. we think it is clearer if it prompts timeout error directly.

This commit converts some InterruptedException to a TimeoutException.


spmallette:
i was a bit hesitant about this one at first, but perhaps this makes sense. looking at the `TraversalOpProcessor`, it seems to have similar behavior converting the `TraversalInterruptionException` to a simple `TimeoutException`.  

VOTE +1

spmallette:
Merge to tp33/master with c923564065854901c62811bb42dd8ad0b938cc81

PR #1010, opened 2018-12-03
TINKERPOP-2075 Added ReferenceElementStrategy

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2075

This strategy converts all graph elements to references. Configured this strategy in Gremlin Server by default for the example graphs. Made a few minor updates to tests that were assuming the return of "detached" elements.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,7 @@ This release also includes changes from <<release-3-3-3, 3.3.3>>.
 * Use `Compare.eq` in `Contains` predicates to ensure the same filter behavior for numeric values.
 * Added `OptionsStrategy` to allow traversals to take arbitrary traversal-wide configurations.
 * Added text predicates.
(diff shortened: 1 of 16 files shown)


dkuppitz:
Do you have a scenario in mind where providers would construct a new `ReferenceElement` step? I don't think it's necessary and I guess if you would have thought of such a scenario, you wouldn't have nested the class inside `ReferenceElementStrategy`; hence I think `ReferenceElement` constructor can be package-private.

Anyway, since this is just a minor thing ...

VOTE +1

spmallette:
yeah.........i'll sleep on that. wasn't sure what to do with it. such a dumb little step. 

oh, i wanted to ask....this should be a finalization strategy right? i tried it as a decoration at first and it didn't work right, so i figured it needed to be applied at the end and that seemed to fix things. :man_shrugging:  should i have done something else?

dkuppitz:
No, `FinalizationStrategy` is correct, nothing else to do here.

PR #1011, opened 2018-12-05
TINKERPOP-1849 Provide a way to fold() with an index

dkuppitz:
https://issues.apache.org/jira/projects/TINKERPOP/issues/TINKERPOP-1849

`docker/build.sh -t -i -n` passed and the docs built with `docker/build.sh -d` look good.

VOTE +1

### DIFF
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -1084,6 +1084,42 @@ g.V().identity()
 
 link:++http://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.html#identity--++[`identity()`]
 
+[[index-step]]
+=== Index Step
+
+The `index()`-step (*map*) indexes each element in the current collection. If the current traverser's value is not a collection, then it's treated as a single-item collection. There two indexers
+available, which can be chosen using the `with()` modulator. The list indexer (default) creates a list for each collection item, with the first item being the original element and the second element
+being the index. The map indexer created a linked hash map in which the index represents the key and the original item is used as the value.
+
+[gremlin-groovy,modern]
+----
+g.V().hasLabel("software").values("name").fold().
+  order(Scope.local).
+  index().
+  unfold().
+  order().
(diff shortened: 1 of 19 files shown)


docs/src/reference/the-traversal.asciidoc
```diff
@@ -1084,6 +1084,42 @@ g.V().identity()
 
 link:++http://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.html#identity--++[`identity()`]
 
+[[index-step]]
+=== Index Step
+
+The `index()`-step (*map*) indexes each element in the current collection. If the current traverser's value is not a collection, then it's treated as a single-item collection. There two indexers
```
spmallette: I assume that "There two indexers" should be "There are two indexers"

docs/src/reference/the-traversal.asciidoc
```diff
@@ -1084,6 +1084,42 @@ g.V().identity()
 
 link:++http://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.html#identity--++[`identity()`]
 
+[[index-step]]
```
spmallette: I think you should have something in upgrade docs to introduce this feature so that folks know it is there.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -1084,6 +1084,42 @@ g.V().identity()
...
+=== Index Step
+
+The `index()`-step (*map*) indexes each element in the current collection. If the current traverser's value is not a collection, then it's treated as a single-item collection. There two indexers
+available, which can be chosen using the `with()` modulator. The list indexer (default) creates a list for each collection item, with the first item being the original element and the second element
+being the index. The map indexer created a linked hash map in which the index represents the key and the original item is used as the value.
+
+[gremlin-groovy,modern]
```
spmallette: You mention that impact of doing `index()` on something that is not a collection in the first sentence of this section but I think that it's worth demonstrating as an example so that people connect the relevance of that statement.

spmallette:
VOTE +1

PR #1013, opened 2018-12-09
connection.py - add finally to return connection to the client pool i…

abrag:
https://issues.apache.org/jira/browse/TINKERPOP-2105

will return the connection to the client pool after an error/exception in read/write op

### DIFF
--- a/gremlin-python/src/main/jython/gremlin_python/driver/connection.py
+++ b/gremlin-python/src/main/jython/gremlin_python/driver/connection.py
@@ -68,6 +68,8 @@ class Connection:
                 done = self._executor.submit(self._receive)
                 result_set.done = done
                 future.set_result(result_set)
+            finally:
+                self._pool.put_nowait(self)
 
         future_write.add_done_callback(cb)
         return future
@@ -78,4 +80,3 @@ class Connection:
             status_code = self._protocol.data_received(data, self._results)
             if status_code != 206:
                 break
-        self._pool.put_nowait(self)


spmallette:
sorry - i just haven't had a chance to really look at this in detail. hopefully i can get at it tomorrow. thanks for submitting it.

abrag:
fixed, 

this commit will solve a critical issue we had on our prod python server, after a few exceptions we left without any available connection to Neptune
but still it feels like a patch as ideally the client (who handle the connection pool) should allocate and free the connections and not the connection itself

spmallette:
Thanks for fixing the existing tests. Two things:

1. I presume this is a bug on 3.3.x as well as the unreleased 3.4.0. If so, could you please re-target this PR at the `tp33` branch
2. You provided a small python script to reproduce that problem. Is there any reason you could not convert that into a test as part of this PR so that we have some insurance against regression?

> ideally the client (who handle the connection pool) should allocate and free the connections and not the connection itself

I think your approach is fine, but I suppose that the `Client` which borrows the `Connection` should be the thing to return it to the pool. If you felt like refactoring that at some point I think that approach would be ok as well. Thanks.

spmallette:
I just wanted to let you know that we intend to freeze the code base at the end of this week as we prepare for release the first week of 2019. No pressure for you to come back to this but I just wanted to let you know in case you wanted to see a fix in the next release.

abrag:
@spmallette 

- added a simple test it should cover the problem
- I will open new PR for tp33 (waiting for the tests to complete), we can wait for 34 if it will be released soon :)

PR #1014, opened 2018-12-11
uses the directly field instead of getter method

otaviojava:


spmallette:
i backported the most important items to tp33 manually here:

https://github.com/apache/tinkerpop/commit/b51b1a7c46b42258cf74180f8c51712970984a45

given comments on #1001 and #1015. we won't merge this directly. thanks.

PR #1016, opened 2018-12-11
Add request pipelining and a fixed ConnectionPool size TINKERPOP-1775 and TINKERPOP-1774

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-1774
https://issues.apache.org/jira/browse/TINKERPOP-1775

This allows connections to be reused for different requests in parallel which should lead to better utilization of connections / lower resource usage. The `ConnectionPool` now also has a fixed size whereas it could previously create an unlimited number of connections. Each `Connection` can handle up to `MaxInProcessPerConnection` requests in parallel.
If this limit is reached for all connections, then a `NoConnectionAvailableException` is thrown which makes this a breaking change.

This change fixes both issues together as discussed in PR #903.

All tests pass with `./docker/build.sh -t -i` and I verified that the changed docs look as expected with `./docker/build.sh -d`.

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
@@ -163,31 +163,30 @@ namespace Gremlin.Net.Driver
         {
             if (Interlocked.CompareExchange(ref _writeInProgress, 1, 0) != 0)
                 return;
-            SendNextMessageFromQueue();
+            SendMessagesFromQueueAsync().Forget();
+        }
+
+        private async Task SendMessagesFromQueueAsync()
+        {
+            while (_writeQueue.TryDequeue(out var msg))
+            {
+                try
+                {
+                    await SendMessageAsync(msg).ConfigureAwait(false);
+                }
+                catch (Exception e)
+                {
+                    await CloseConnectionBecauseOfFailureAsync(e).ConfigureAwait(false);
+                    break;
+                }
+            }
             Interlocked.CompareExchange(ref _writeInProgress, 0, 1);
         }
 
-        private void SendNextMessageFromQueue()
+        private async Task CloseConnectionBecauseOfFailureAsync(Exception exception)
         {
-            if (_writeQueue.TryDequeue(out var msg))
-                SendMessageAsync(msg).ContinueWith(
-                    t =>
-                    {
-                        if (t.IsFaulted)
-                        {
-                            CloseConnectionBecauseOfFailure(t.Exception);
-                            return;
-                        }
-
-                        SendNextMessageFromQueue();
-                    }, TaskContinuationOptions.ExecuteSynchronously);
-        }
-
-        private void CloseConnectionBecauseOfFailure(AggregateException exception)
-        {
-            exception.Handle(_ => true);
             EmptyWriteQueue();
-            CloseAsync().WaitUnwrap();
+            await CloseAsync().ConfigureAwait(false);
             NotifyAboutConnectionFailure(exception);
         }
 
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
@@ -22,17 +22,26 @@
 #endregion
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Net.WebSockets;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using Gremlin.Net.Driver.Messages;
-using Gremlin.Net.Driver.ResultsAggregation;
+using Gremlin.Net.Process;
 using Gremlin.Net.Structure.IO.GraphSON;
 using Newtonsoft.Json.Linq;
 
 namespace Gremlin.Net.Driver
 {
+    internal interface IResponseHandlerForSingleRequestMessage
+    {
+        void HandleReceived(ResponseMessage<JToken> received);
+        void Finalize(Dictionary<string, object> statusAttributes);
+        void HandleFailure(Exception objException);
+    }
+
     internal class Connection : IConnection
     {
         private readonly GraphSONReader _graphSONReader;
@@ -42,6 +51,13 @@ namespace Gremlin.Net.Driver
         private readonly WebSocketConnection _webSocketConnection;
         private readonly string _username;
         private readonly string _password;
+        private readonly ConcurrentQueue<RequestMessage> _writeQueue = new ConcurrentQueue<RequestMessage>();
+
+        private readonly ConcurrentDictionary<Guid, IResponseHandlerForSingleRequestMessage> _callbackByRequestId =
+            new ConcurrentDictionary<Guid, IResponseHandlerForSingleRequestMessage>();
+        private int _connectionState = 0;
+        private int _writeInProgress = 0;
+        private const int Closed = 1;
 
         public Connection(Uri uri, string username, string password, GraphSONReader graphSONReader,
             GraphSONWriter graphSONWriter, string mimeType, Action<ClientWebSocketOptions> webSocketConfiguration)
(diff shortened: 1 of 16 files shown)


FlorianHockmann:
> I would recommend to remove the dynamic pool sizes (min/max), but I leave it up to you. In my experience, the use for dynamic pools is very limited. Usually when there is spike in requests, opening new connections can lead to a latency with no added benefit (there is a limited number of requests that a server can process at any given time).

You mean that the pool should simply have a fixed size instead? Wasn't that what you asked for in TINKERPOP-1774:

> Similar to the java connection pool, we should limit the maximum amount of connections and start with a minimum number.

or what do you mean exactly with dynamic pool sizes?

I don't really have a strong opinion on that. My assumptions was just that it's probably hard for users to set a good value for the size. Dynamic pool sizes basically avoid this as the pool will automatically grow to a size that is big enough to handle the number of parallel requests the application typically uses. Later on, we could also add a mechanism to remove idle connections if they were only created because of a spike in requests as too many idle connection seem to be problematic with certain providers (especially CosmosDB, judging from the tickets we currently have open). Although this mechanism would of course make the latency problem even worse for spikes when the connections were already closed again since the last spike.

jorgebay:
The behaviour that motivated TINKERPOP-1774 was the unbounded number of connections that was being created, but it's true that references the java driver as the reference implementation, sorry about that.

I think having a fixed amount of connections leads to a predictable behaviour (latency / resource provisioning / etc) and a simpler implementation. With request pipelining, there are few cases where the user will need more than 1 connection (in that case, the user can configure a different fixed number).

FlorianHockmann:
Since I don't really have a strong opinion on this and because I see your points about simpler code and predictable latencies, I'll change this to use a fixed size unless someone else chimes in and explains why we should use dynamic sizes.

gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
```diff
@@ -154,6 +156,38 @@ private string SaslArgument()
...
+        {
+            if (Interlocked.CompareExchange(ref _writeInProgress, 1, 0) != 0)
+                return;
+            if (_writeQueue.TryDequeue(out var msg))
+                SendMessageAsync(msg).ContinueWith(
+                    t =>
+                    {
+                        if (t.IsFaulted)
+                        {
+                            _callbackByRequestId[msg.RequestId].HandleFailure(t.Exception);
```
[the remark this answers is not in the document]
FlorianHockmann: > AggregateException.Handle() should be called here, as the faulted task is not going to be yielded to the user.

You mean simply like this (+ the logic from your second remark here)?

```cs
t.Exception.Handle(e =>
    {
        _callbackByRequestId[msg.RequestId].HandleFailure(t.Exception);
        _callbackByRequestId.TryRemove(msg.RequestId, out _);
        return true;
    });
```

> Additionally, I think if the sending fails, the whole connection should be "trashed" (empty queues, respond with the errors back to the user) and close the connection.

That sounds like a good idea. I guess we have to throw the exception then for every request for which we have a callback registered at that time.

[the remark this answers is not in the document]
FlorianHockmann: Ah, didn't know about this. I'll add this line like you suggested.

gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
```diff
@@ -154,6 +156,38 @@ private string SaslArgument()
             return Convert.ToBase64String(authBytes);
         }
 
+        private void SendNextMessage()
+        {
+            if (Interlocked.CompareExchange(ref _writeInProgress, 1, 0) != 0)
+                return;
+            if (_writeQueue.TryDequeue(out var msg))
```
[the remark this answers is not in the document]
FlorianHockmann: Good catch, I think that was actually the cause of the time outs on Travis.

gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
```diff
@@ -154,6 +159,67 @@ private string SaslArgument()
...
 
+        private void BeginSendingMessages()
+        {
+            if (Interlocked.CompareExchange(ref _writeInProgress, 1, 0) != 0)
+                return;
+            SendNextMessageFromQueue();
+            Interlocked.CompareExchange(ref _writeInProgress, 0, 1);
+        }
+
+        private void SendNextMessageFromQueue()
```
jorgebay: Suggestion: to simplify the implementation you can create an async method:

```C#
private async Task SendNextMessageFromQueue() {}
```

And use `await` for your dequeue-send flow. If you catch your exceptions, its safe to start the call for `SendNextMessageFromQueue()` as "fire and forget". The end result will be the same (as `ContinueWith()` does exactly that) but the code will be simpler.
You can look at a `Forget()` implementation here: https://github.com/datastax/csharp-driver/blob/46e8da823512f12cc4f109e6812ff611a9f14e20/src/Cassandra/Tasks/TaskHelper.cs#L380-L387

jorgebay: There should be no concern regarding license, it's Apache licensed, I'm the author and I'm committer here :)

gremlin-dotnet/src/Gremlin.Net/Driver/SingleMessageResultReceiver.cs
```diff
@@ -0,0 +1,88 @@
...
+        public ResponseHandlerForSingleRequestMessage(GraphSONReader graphSonReader)
+        {
+            _graphSONReader = graphSonReader;
+        }
+
+        public void HandleReceived(ResponseMessage<JToken> received)
+        {
+            var receivedData = typeof(T) == typeof(JToken)
+                ? new[] {received.Result.Data}
+                : _graphSONReader.ToObject(received.Result.Data);
```
jorgebay: It would be nice to remove the dynamic call here, i.e., cast to `object`.

That way the failure would be far more simple to understand if its static code.

FlorianHockmann: You mean that `_graphSONReader.ToObject()` shouldn't return `dynamic`?

jorgebay: I mean converting it to whatever you expect (`IEnumerable`?) to avoid making `_aggregator.Add(d)` and `_result.Add(d)` being dynamic calls.

gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
```diff
@@ -55,87 +71,75 @@ internal class Connection : IConnection
...
-                var received = await _webSocketConnection.ReceiveMessageAsync().ConfigureAwait(false);
-                var receivedMsg = _messageSerializer.DeserializeMessage<ResponseMessage<JToken>>(received);
-
-                status = receivedMsg.Status;
-                status.ThrowIfStatusIndicatesError();
-
-                if (status.Code == ResponseStatusCode.Authenticate)
+            var state = Volatile.Read(ref _connectionState);
+            if (state == Closed) return;
+            _webSocketConnection.ReceiveMessageAsync().ContinueWith(
```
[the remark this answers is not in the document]
FlorianHockmann: Good idea! I made a similar change like for the send part. All exceptions are caught (exceptions where the request ID is known are already caught at a lower level to only let that request fail) and the same failure handle function is called like for the send part.

gremlin-dotnet/src/Gremlin.Net/Driver/SingleMessageResultReceiver.cs
FlorianHockmann: OK, then I really misunderstood your suggestion. I tried your suggestion, but unfortunately it wasn't that easy as it would require a change to this line, too:
```cs
_result.Add(d);
```
which doesn't work if `d` is of type `object`. We would need a cast to `T` here, but that would require more handling in case that `d` is not of type `T`.
So, I prefer not to change this now and instead improve type safety of the deserialization in general some time later.

gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
```diff
@@ -55,87 +71,86 @@ internal class Connection : IConnection
...
+        {
+            var receivedMsg = _messageSerializer.DeserializeMessage<ResponseMessage<JToken>>(received);
+            
+            try
+            {
+                TryParseResponseMessage(receivedMsg);
+            }
+            catch (Exception e)
+            {
+                _callbackByRequestId[receivedMsg.RequestId].HandleFailure(e);
```
jorgebay: NIT, we can avoid the second read access to the dictionary by using:

```C#
if (_callbackByRequestId.TryRemove(receivedMsg.RequestId, out var responseHandler))
{
    responseHandler.HandleFailure(e);
}
```

FlorianHockmann: Oh yes, kind of obvious. I just pushed a commit with this change.

spmallette:
@FlorianHockmann now that python issues are resolved I intend to start the release process soon.  are you still expecting to merge this for 3.4.0?

FlorianHockmann:
> now that python issues are resolved I intend to start the release process soon. are you still expecting to merge this for 3.4.0?

Yes, from my side this is ready to be merged. I just wanted to wait for another review and not use lazy consensus as the change isn't exactly trivial and also because of the holidays where probably not so many contributors had the chance to review this. However, if you want to start the release process, then this PR shouldn't stop it in my opinion. So, I can merge this tomorrow morning or later today.

spmallette:
ok - i have to start with 3.2.11 first so it will be a bit before i get to 3.4.0. you have some time to finish up.

PR #1017, opened 2018-12-11
Add isContextNullOrStopped() method to Spark

alexliu68:


spmallette:
Going to merge this via CTR as it's pretty simple - just exposes some state of the `SparkContext` instance. Thanks.

PR #1018, opened 2018-12-13
TINKERPOP-2111 Added BulkSet as a GraphSON Type

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2111

Adding `BulkSet` brings some consistency to how results are processed and sets things up better for GraphBinary. Implemented for all GLVs though they do not yet have `BulkSet` implementations - they simply coerce to `List` during deserialization. This is basically what was happening before but in earlier versions it was happening on the server side.

All tests pass with `docker/build.sh -t  -i`

VOTE +1


jorgebay:
VOTE +1

gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/BulkSetSerializer.cs
```diff
@@ -0,0 +1,54 @@
...
+            // so this query will be trouble. we'd need a legit BulkSet implementation here in C#. this current 
+            // implementation is here to replicate the previous functionality that existed on the server side in 
+            // previous versions.
+            var result = new List<Tuple<object,int>>();
+            for (var i = 0; i < jArray.Count; i += 2)
+            {
+                result.Add(new Tuple<object,int>(reader.ToObject(jArray[i]), (int) reader.ToObject(jArray[i + 1])));
+            }
+
+            return result.SelectMany(x => Enumerable.Repeat(x.Item1, x.Item2)).ToList();
```
dkuppitz: The whole loop and remapping can be simplified to:

```
return Enumerable.Range(0, jArray.Count / 2).SelectMany(i =>
        Enumerable.Repeat(reader.ToObject(jArray[i * 2]), (int) reader.ToObject(jArray[i * 2 + 1]))).
    ToList();
```

spmallette: nicely done. i really wanted a `partition()` function, but what you're doing there is better than that stupid for loop

PR #1019, opened 2018-12-15
TINKERPOP-2114 Document common Gremlin anti-patterns

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-2114

This PR adds a documentation that discusses common anti-patterns.

Docs built with `docker/build.sh -d` look good.

VOTE +1

### DIFF
--- a/docs/src/recipes/anti-patterns.asciidoc
+++ b/docs/src/recipes/anti-patterns.asciidoc
@@ -1,0 +1,291 @@
+////
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+////
+
+[[long-traversals]]
+== Long Traversals
+
+It can be tempting to generate long traversals, e.g. to create a set of vertices and edges based on information that
+resides within an application. For example, let's consider two lists - one that contains information about persons and
+another that contains information about the relationship between these persons. To illustrate the problem we will
+create two list with a few random map entries.
+
+[gremlin-groovy]
+----
+:set max-iteration 10
+rnd = new Random(123) ; x = []
+persons = (1..100).collect {["id": it, "name": "person ${it}", "age": rnd.nextInt(40) + 20]}
+relations = (1..500).collect {[rnd.nextInt(persons.size()), rnd.nextInt(persons.size())]}.
+  unique().grep {it[0] != it[1] && !x.contains(it.reverse())}.collect {
+    x << it
+    minAge = Math.min(persons[it[0]].age, persons[it[1]].age)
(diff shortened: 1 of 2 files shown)


docs/src/recipes/anti-patterns.asciidoc
```diff
@@ -0,0 +1,291 @@
...
+[[unnecessary-steps]]
+== Unnecessary Steps
+
+There are quite a few steps and patterns that can be combined into a much shorter form. TinkerPop is trying to optimize queries, by
+rewriting such patterns automatically using traversal optimization strategies. These strategies, however, do have a few preconditions
+and under certain circumstance they will not attempt to rewrite a traversal. For example, if the traversal has path computations
+enabled (e.g. by using certain steps, such as `path()`, `simplePath()`, `otherV()`, etc.), then the assumption is that all steps are
+required in order to produce the desired path.
+
+An often seen anti-pattern is the one that explicitely traversers to an edge and then to a vertex without using any filters.
```
jbmusso:
```suggestion
An often seen anti-pattern is the one that explicitely traverses to an edge and then to a vertex without using any filters.
```

docs/src/recipes/anti-patterns.asciidoc
```diff
@@ -0,0 +1,291 @@
...
+
+[gremlin-groovy,modern]
+----
+g.V().hasLabel("person").outE("created").inV().dedup()    <1>
+g.V().hasLabel("software").inE("created").outV().count()  <2>
+----
+
+<1> The `created` edge is never really needed as the traversal only asks for all things that were created by all persons in the graph.
+    These "things" are only represented by the adjacent vertices, not the edges.
+<2> This traversals counts the persons in the graph who created software. The interesting thing about this query is that it actually
```
jbmusso:
```suggestion
<2> This traversal counts the persons in the graph who created software. The interesting thing about this query is that it actually 
```

FlorianHockmann:
Thanks for putting this together. I definitely see some of these examples in traversals newer users write and I didn't even know myself about the difference between traversals like `groupCount().by(label)` and `groupCount().by(label())`.

VOTE +1

docs/src/recipes/anti-patterns.asciidoc
robertdale: explicitly

docs/src/recipes/anti-patterns.asciidoc
```diff
@@ -0,0 +1,291 @@
...
+----
+g.V().as('a').
+  both().where(lt('a')).by(id).as('b').
+  both().where(lt('a')).by(id).where(gt('b')).by(id).as('c').
+  not(both().where(eq('a'))).
+  select('a','b','c').
+    by('name')
+----
+
+Ignoring the anti-patterns that were discussed before there's not much wrong with the traversal, but note the two chained `where()`-steps
```
spmallette: I feel like there's maybe two different things that you mean here so I can't say what the correction is but there's something awkward about how this phrase is written:

> Ignoring the anti-patterns that were discussed before there's not much wrong with the traversal,

spmallette:
This is great - VOTE +1

docs/src/recipes/anti-patterns.asciidoc
dkuppitz: "Ignoring the anti-patterns that were discussed before" -- I added this because the queries use `both()` (instead of `both(<label>)`), which was declared to be an anti-pattern in an earlier section. I can change it to `both("knows","created")`, but I wanted the reader to focus on the `where()` steps, hence I kept the other steps as short as possible.

spmallette: oh...now i get it. i think i was pausing my reading in the wrong place. i think a comma after "before" would help:

"Ignoring the anti-patterns that were discussed before, there's not much wrong with the traversal, but note the two chained `where()`-steps..."

PR #1021, opened 2018-12-18
TINKERPOP-2116: Use explicit Bindinds object instead of a 2-tuple

aboudreault:


FlorianHockmann:
Can you please also update [the docs](http://tinkerpop.apache.org/docs/current/reference/#_bindings) accordingly? Otherwise, new users will still try the tuple version.

aboudreault:
@FlorianHockmann Yep, already checked to update it, but this docs is not anymore in master.

spmallette:
i will look into that "bindings" section after this merges.

FlorianHockmann:
> Yep, already checked to update it, but this docs is not anymore in master.

Oh right, I only looked at the currently published version of the docs and didn't think about the restructuring of the docs @spmallette did.

But the changes look good to me, so:
VOTE +1

dkuppitz:
VOTE +1

spmallette:
I realized that i didn't remove the bindings content - i just moved it out from under its own section header to the "lambda" stuff where it has more meaning as bindings only have context in relation to those lambda based traversals:

https://github.com/apache/tinkerpop/commit/4658d8342d4d7b4e92d442bb138fec611c65472e

PR #1023, opened 2018-12-20
TINKERPOP-2105 fix for TP33

abrag:
@spmallette as discussed in https://github.com/apache/tinkerpop/pull/1013


abrag:
python tests are green
I do not understand why the other 2 failed before even starting :/

dkuppitz:
VOTE +1

Also, I think this PR is a replacement for #1013 (meaning the old one can be closed).

PR #1024, opened 2018-12-26
TINKERPOP-2113 Fix P.Within() for list arguments in Gremlin.NET

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2113

This was apparently pretty much the same bug as with the problems we had earlier with the `inject()` step ([TINKERPOP-1972](https://issues.apache.org/jira/browse/TINKERPOP-1972)). So, the fix was also basically the same.

All .NET tests pass.

VOTE +1


spmallette:
VOTE +1

dkuppitz:
VOTE +1

PR #1026, opened 2018-12-26
TINKERPOP-2088 Enable SourceLink for Gremlin.Net

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2088

This requires contributors to update to .NET Core SDK 2.1.300 or newer.
Travis and the Docker build now use the most recent SDK 2.2.

SourceLink should make it easier for users to debug Gremlin.Net within their application as they get better access to the source code of Gremlin.Net during debugging. That is why the .NET open-source library guidance recommends to enable it for libraries.

The generated NuGet package includes a reference to our git repository as well as the commit that belongs to the version included in the package:
tinkerpop-2088-sourcelink

I verified that the Docker build still works with `docker/build.sh`.

VOTE +1

NOTE: We don't need to merge this for the upcoming release, especially since we have have a code freeze (although we don't strictly follow that for GLVs). I just had some spare time to work on this, but we can also merge it after the 3.3.5 release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,11 +18,12 @@ install:
   - mvn -version
 
 before_install:
+  - sudo apt-get install -y dpkg # workaround for travis-ci/travis-ci#9361
   - wget -q https://packages.microsoft.com/config/ubuntu/14.04/packages-microsoft-prod.deb
   - sudo dpkg -i packages-microsoft-prod.deb
   - sudo apt-get install apt-transport-https
   - sudo apt-get update
-  - sudo apt-get install dotnet-sdk-2.1
+  - sudo apt-get install dotnet-sdk-2.2
 
 jobs:
   include:
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -20,17 +20,16 @@ FROM ubuntu:trusty
 MAINTAINER Daniel Kuppitz <me@gremlin.guru>
 
 RUN apt-get update \
-    && apt-get -y install software-properties-common python-software-properties apt-transport-https \
+    && apt-get -y install software-properties-common python-software-properties apt-transport-https curl \
     && echo oracle-java8-installer shared/accepted-oracle-license-v1-1 select true | debconf-set-selections \
     && add-apt-repository -y ppa:webupd8team/java \
-    && sh -c 'echo "deb [arch=amd64] https://packages.microsoft.com/repos/microsoft-ubuntu-trusty-prod trusty main" > /etc/apt/sources.list.d/dotnetdev.list' \
-    && apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 417A0893 \
+    && sh -c 'curl -s https://packages.microsoft.com/config/ubuntu/14.04/packages-microsoft-prod.deb -o packages-microsoft-prod.deb' \
+    && sh -c 'dpkg -i packages-microsoft-prod.deb' \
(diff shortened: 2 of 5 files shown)


dkuppitz:
The Docker builds seem to be broken now. After deleting the existing docker images, I get the following error when `docker.build.sh` tries to rebuild the images:

```
...
Updating certificates in /etc/ssl/certs... 0 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d....done.
Processing triggers for ureadahead (0.100.0-16) ...
Errors were encountered while processing:
 /var/cache/apt/archives/aspnetcore-runtime-2.2_2.2.1-1_amd64.deb
E: Sub-process /usr/bin/dpkg returned an error code (1)
The command '/bin/sh -c apt-get update     && apt-get -y install software-properties-common python-software-properties apt-transport-https curl     && echo oracle-java8-installer shared/accepted-oracle-license-v1-1 select true | debconf-set-selections     && add-apt-repository -y ppa:webupd8team/java     && sh -c 'curl -s https://packages.microsoft.com/config/ubuntu/14.04/packages-microsoft-prod.deb -o packages-microsoft-prod.deb'     && sh -c 'dpkg -i packages-microsoft-prod.deb'     && sh -c 'echo "deb https://download.mono-project.com/repo/ubuntu stable-trusty main" | sudo tee /etc/apt/sources.list.d/mono-official-stable.list'     && apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF     && apt-get update     && apt-get install -y oracle-java8-installer gawk git maven openssh-server subversion zip     && apt-get install -y --force-yes dotnet-sdk-2.2 python python-dev python3-dev python-pip build-essential mono-devel     && pip install virtualenv virtualenvwrapper     && pip install --upgrade pip     && rm -rf /var/lib/apt/lists/* /var/cache/oracle-jdk8-installer' returned a non-zero code: 100
```

FlorianHockmann:
That's strange. It still works fine for me. I used this to delete any existing images and then start the build:

```bash
docker system prune -af
docker/build.sh
```

We could in general add another Travis job to run `docker/build.sh` to ensure that we don't accidentally break the Docker build. I remember that we had issues like this already a few times now, although I'm of course not sure whether this approach would have found every problem in the Docker build.

FlorianHockmann:
```
 has premature member 'control.tar.xz' before 'control.tar.gz', giving up
```

Wait, I got exactly the same error in this PR with Travis and added this line to the `.travis.yml` which fixed it there:
```
- sudo apt-get install -y dpkg # workaround for travis-ci/travis-ci#9361
```
and as the comment indicates, this problem is described here: travis-ci/travis-ci#9361 (where I also got this workaround).

Can you try whether it also fixes the problem with the Docker build if you add that line to the Dockerfile?

dkuppitz:
Yep, nice, that solved it!

dkuppitz:
Well, the images build fine now, but `gremlin-dotnet-source` doesn't.

FlorianHockmann:
Do you get an exception?

dkuppitz:
See: https://gist.github.com/dkuppitz/1b912747180b07ab08d4eb98175c119b

I'm really wondering how you're able to get a clean build using the same docker images, so confusing...

spmallette:
wonder if that's related to nuget....we still have some odd build problems around clean repos that don't have nuget present i think. i have that on my list of things to review.

dkuppitz:
It finally worked! I'll add `git -fxd` in our Docker build script.

PR #1037, opened 2019-01-09
Fix table rendering in Sparql reference docs.

newkek:
The rendering was off due to the `|` characters in the table, this seems to fix it


spmallette:
thanks - we have so many bad formatting errors in the docs on this 3.4.0 release - stinks

PR #1039, opened 2019-01-13
TINKERPOP-2131 Gremlin.Net: Better explain connection pool exceptions

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2131

Two new exception classes are added to better reveal the reason for a `NoConnectionAvailableException`:
* The pool is empty because the server is unavailable
    -> `ServerUnavailableException`
* All connections have reached the max in-flight requests limit
    -> `ConnectionPoolBusyException`
Both extend `NoConnectionAvailableException` so this is a non-breaking change.

VOTE +1


spmallette:
VOTE +1

FlorianHockmann:
> though a CHANGELOG entry would be good

Right, I just added an entry.

dkuppitz:
VOTE +1

PR #1041, opened 2019-01-16
TINKERPOP-2125 Extend release validation script

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-2125

Added source file check in release validation script.

VOTE +/-0

At this point, the validation does not pass. We need to make a decision on what to do with the following files:

```
./docs/static/images/gremlin-characters.pdf
./docs/static/images/gremlintron.pdf
./docs/static/images/gremlin-sacks.pdf
./docs/static/images/gremlin-gym.pdf
```

Do we whitelist them or should they be removed from the repository?

### DIFF
--- a/bin/validate-distribution.sh
+++ b/bin/validate-distribution.sh
@@ -136,6 +136,20 @@ echo "OK"
 
 if [ "${TYPE}" = "SOURCE" ]; then
 cd ${DIR_NAME}
+echo -n "* checking source files ... "
+find . -type f | xargs -n1 -I {} file {} --mime | grep 'charset=binary' | cut -f1 -d: |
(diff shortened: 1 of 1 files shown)


spmallette:
I'm not sure what those are.............i assume they are just high-res versions of the png files of the same name? i guess i would just whitelist them since they've been there forever at this point. i assume we can allow *.pdf (we're really trying to mostly avoid compiled code like `.exe`, `.dll`, `.jar`, etc)

dkuppitz:
PDFs are now whitelisted and the validation passes.

VOTE +1

spmallette:
VOTE +1

PR #1043, opened 2019-01-19
Fix NPE is not thrown when null id is set to AbstractStep

porunov:
This PR removes the redundant `Objects.nonNull(id);` call for `setId` method.


spmallette:
Could you please explain this change a bit further? In what way is this null check redundant? where else is a non-null ensured to make this check unnecessary?

porunov:
@spmallette this check returns either `true` of `false`. The returned value isn't used (i.e. the call is redundant). `Objects.nonNull(id);` doesn't throw an exception if the `id` is null.

dkuppitz:
Perhaps it should be replaced with `Objects.requireNonNull(...)` then.

spmallette:
ah - i was thinking of the exception throwing version when i looked at it. then the @dkuppitz suggestion is probably what we want here.

porunov:
@dkuppitz @spmallette Thank you for the clarification. I've replaced it with  `Objects.requireNonNull`.

PR #1046, opened 2019-01-24
TINKERPOP-2138 Provide way to disable the global function cache

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2138

Added a new `globalFunctionCacheEnabled` configuration to the `GroovyCompilerGremlinPlugin` and an overriding setting for sessions on `SessionOpProcessor` that disables the global function cache.  In cases where you don't need the cache (i.e. don't use global functions) it does remove a bit of reflection based processing from every single script request and obviously saves on memory formerly used by the cache. 

VOTE +1


dkuppitz:
LGTM.

VOTE +1

jorgebay:
Makes sense, VOTE +1

PR #1047, opened 2019-01-24
TINKERPOP-2139 Properly handle client side serialization exceptions

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2139

Prior to this change serialization exceptions were basically just logged because any failure would signal a bad host and mark it as dead, but that's not really what was happening obviously if the client was never even able to try to send the message to the host in the first place. These sorts of client exceptions are now handled better and actually throw an exception that the calling client can see and utilize while not killing the host.

VOTE +1


nastra:
VOTE +1

dkuppitz:
VOTE +1

PR #1049, opened 2019-01-25
TINKERPOP-2144 Better handle Authenticator failures

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2144

Failures occurred on the first round of a SASL request when the negotiator was being created and if an exception was tossed there neither the server nor driver handled the situation well. By explicitly capturing and handling it, we can now properly assert results in kerberos tests without having to try to assert server logs which wasn't always dependable for some reason. This change should prevent random failures around kerberos tests which has long been a problem.

VOTE +1


dkuppitz:
VOTE +1

PR #1050, opened 2019-01-28
Test GraphBinary with nested lists/sets/maps

nastra:


spmallette:
Thanks for this - merged via CTR as the change is for additional tests and Travis passed without issue.

PR #1051, opened 2019-01-29
Make readValue(..)/writeValue(..) protected

nastra:
This is so that simple custom type serializers can reuse the same logic
without having to put their type serializer into the same package
structure because readValue() was package-private.
Also I believe it makes sense to make writeValue() protected as well.


jorgebay:
My intention was to make it `protected` initially (not package-private), thanks @nastra !

I'll merge it via CTR once the build finishes.

PR #1052, opened 2019-01-31
Release dropped response Frames in ResponseHandlerContext

dimas-b:
A Frame may wrap a ReferenceCounted buffer, which is normally released
when it is written back to the client. However, if the Frame is dropped
by ResponseHandlerContext the buffer also has to be released.

---

This is a follow-up fix for #899

----

Testing performed:
* Full suite of unit tests (i.e. `mvn clean install`)
* Full suite of gremlin-server integration tests (`mvn verify -pl gremlin-server -DskipIntegrationTests=false`)
* Manual testing with the Gremlin Server started with `-Dio.netty.leakDetection.level=PARANOID`


spmallette:
thank you - merged this little fix as CTR

PR #1056, opened 2019-02-05
fix: Path#toString() variable reference bug

thefliik:
I'm going through the process of creating typescript definitions for the javascript library, and I noticed a bug in Path#toString(). It incorrectly attempts to call a non-existant local `objects` variable instead of calling the class property `this.objects`.

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.js
@@ -128,7 +128,7 @@ class Path {
   }
 
   toString() {
-    return 'path[' + objects.join(", ") +  ']';
+    return 'path[' + this.objects.join(", ") +  ']';
   }
 
   equals(other) {


thefliik:
Seperately, would you be interested in a PR to port the whole javascript library over to typescript?

spmallette:
Adding typescript definitions is being discussed here:

https://issues.apache.org/jira/browse/TINKERPOP-2027

perhaps you could read through that and include any thoughts you might have. i don't think anyone is working on that issue at this time, but you might want to ask on that in any comments you provide.

jorgebay:
Thanks @thefliik , I've filed a ticket to keep track of this: https://issues.apache.org/jira/browse/TINKERPOP-2152

Can you target branch `tp33` for this to be applied to the 3.3 line?

thefliik:
@jorgebay in the `tp33` branch it looks like the Path#toString() method has been removed entirely, so when/if that branch is merged into master this PR will be unnecessary. Feel free to close.

spmallette:
`tp33` continually merges to `master` - wonder why `toString()` isn't present on that branch....

thefliik:
The code in question on `tp33`:

https://github.com/apache/tinkerpop/blob/tp33/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.js#L118-L139

jorgebay:
`toString()` was introduced as part of TK-1446, it didn't have a string representation on the js side prior to that.

@thefliik sorry for the confusion, nvm.

Fixing it on `master` makes sense, VOTE +1

PR #1057, opened 2019-02-07
Changed loop break check to break out of the connection creation loop

IsaacBoy:
Changed loop break check to break out of the connection creation loop if greater than or equal to the expected amount of connections are created. A situation exists where unless exactly the expected amount of connections are made when checked, an infinite loop of connections may be opened.

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
@@ -70,7 +70,7 @@ namespace Gremlin.Net.Driver
             while (true)
             {
                 var nrOpened = Interlocked.CompareExchange(ref _nrConnections, PoolEmpty, PoolEmpty);
-                if (nrOpened == _poolSize) break;
+                if (nrOpened >= _poolSize) break;
                 if (nrOpened != PoolPopulationInProgress)
                 {
                     await PopulatePoolAsync().ConfigureAwait(false);
@@ -81,7 +81,7 @@ namespace Gremlin.Net.Driver
         private async Task PopulatePoolAsync()
         {
             var nrOpened = Interlocked.CompareExchange(ref _nrConnections, PoolPopulationInProgress, PoolEmpty);
-            if (nrOpened == PoolPopulationInProgress || nrOpened == _poolSize) return;
+            if (nrOpened == PoolPopulationInProgress || nrOpened >= _poolSize) return;
 
             try
             {


IsaacBoy:
Issue: https://github.com/apache/tinkerpop/compare/master...IsaacBoy:TINKERPOP-2155

IsaacBoy:
Issue in JIRA: https://issues.apache.org/jira/browse/TINKERPOP-2155

spmallette:
thanks for submitting this - in principle your change makes sense to me so VOTE +1 but i'll let @FlorianHockmann  provide his thoughts since he probably knows the most about this.

FlorianHockmann:
Thanks for this contribution.
I also noticed problems with the way the connection pool now handles resizing and closing of broken connections ([TINKERPOP-2135](https://issues.apache.org/jira/browse/TINKERPOP-2135)) which are probably related. Improving this situation will most likely require a bigger change, but since you have described a concrete problem in TINKERPOP-2155 and because your change is really minimal, I think that it makes sense to merge this to have your problem fixed. We can still improve the other aspects later.
So: VOTE +1.

IsaacBoy:
Cool. Does somebody manually initiate the merge into master?

spmallette:
yes. it will get merged. typically we need 3 +1s from committers OR a single +1 and a seven day waiting period for objections:

http://tinkerpop.apache.org/docs/current/dev/developer/#rtc

either way, should merge shortly. as it's a small change i could probably just merge it given florian's feedback, but since it's close to both a third +1 and the 7 day period i think i'll just wait it out.

dkuppitz:
VOTE +1

tomek22:
I just spent almost a full day figuring out why our service leaks memory and can't open sockets. I tracked it down to this issue. Managed to reproduce locally by killing the gremlin server after the connection pool has already been created and then restarting gremlin server. I was very happy to see it fixed in master, but sad to see it hasn't been released yet. Any idea when there will be a nuget package available with this fix? We had to workaround the issue by moving the gremlin client to request scope instead of singleton which I expect will result in some performance penalty :(

spmallette:
We have been floating the idea of a release for mid-march - so a couple of weeks or so.

tomek22:
I see tinkerpop 3.4.1 release happened 2 days ago, but the latest nuget on nuget.org is still 3.4.0.
Our workaround for this issue is not bullet proof and we've had leaks with it, so we're really looking forward to this fix :)
Any idea when the nuget will be available?

spmallette:
The repository was tagged and release artifacts were generated, but the decision to release is still undergoing the VOTE process:

https://lists.apache.org/thread.html/07f4e399ebdf1f7aca2a30518eb72a1d642015e614e243286ad70c23@%3Cdev.tinkerpop.apache.org%3E

Given how that thread is going ,you could see the new version in nuget as early as Friday afternoon EST, but it might be as late as Monday.

PR #1058, opened 2019-02-08
TINKERPOP-2154 GraphBinary: Release working buffers on failure

jorgebay:
https://issues.apache.org/jira/browse/TINKERPOP-2154

Ensures that reference counts of allocated buffers are set to 0 when there is a failure during serialization.

Includes a test for memory allocation when introducing failures.

VOTE +1


spmallette:
i guess there isn't a better way to do that kind of stuff since we allocate within the serializers themselves. oh well. 

needs a changelog entry but VOTE +1

dkuppitz:
VOTE +1

PR #1059, opened 2019-02-12
Improve download page

spmallette:
I didn't bother to create a ticket for this as it's more of a administrative thing, but I've reworked the download page to have a dropdown list of the archived versions. In that way, the page doesn't scroll so much and other content lower in the page is more immediately visible. This should also make our release days a bit less error prone in retiring old versions to archives as it's now just a matter of editing the dropdown to include the archived versions - easier that cut/paste.

What I have is functional though perhaps not awesome in terms of jquery/bootstrap/js so if this can be improved please let me know. I feel like I've tested pretty well, but it would be great if a few folks could `bin/generate-home.sh` and try the mechanics out directly.

VOTE +1

### DIFF
--- a/docs/site/home/downloads.html
+++ b/docs/site/home/downloads.html
@@ -21,18 +21,6 @@ limitations under the License.
        <a href="http://tinkerpop.apache.org/docs/current/reference/#gremlin-console">Gremlin Console</a> and <a href="http://tinkerpop.apache.org/docs/current/reference/#gremlin-server">Gremlin Server</a>
        downloads are binary distributions, which contain pre-packaged versions of these important TinkerPop applications that are designed to work out-of-the-box
        when unpackaged. The source distribution is a snapshot of the source code and files used in the building of those  binary distributions.</p>
-    <p>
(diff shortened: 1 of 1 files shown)


robertdale:
The current instructions are:

https://tinkerpop.apache.org/docs/current/dev/developer/#_pmc_vote
> PMC Vote
> 4. d. Update Downloads page, when moving "Current Releases" to "Archived Releases" recall that the hyperlink must change to point to version in the Apache Archives.

With this, we would just add a new `<option>` line and not have to update the archive links, correct?

spmallette:
yes - i will update that before i merge.

dkuppitz:
Looks, good, much cleaner now. I haven't tested all the links, but I picked out some random releases and all links worked as expected.

VOTE +1

robertdale:
Firefox renders the dropdown as black.

image

robertdale:
I pushed a fix. Looks like there is no 'bootstrap-select' class.  I changed it to use the standard bootstrap form/select - https://getbootstrap.com/docs/3.3/css/#forms-controls

PR #1060, opened 2019-02-12
TINKERPOP-1435 GraphSON extended module support for gremlin-python

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1435

Implemented `Duration`, `Char`, `ByteBuffer`, `Byte`, `BigInteger` and `BigDecimal` which were the most critical ones right now. Tried to do `InetAddress` but there didn't seem to be anything in Python suitable to handled that type except for `ip_address` but it wouldn't parse things like "localhost" and other things that Java will so it seemed a bit incompatible. Also ignored all the `java.time.*` things for now...wonder if those shouldn't just be deprecated....

Builds with `mvn clean install && mvn verify -pl gremlin-python`

VOTE +1

### DIFF
--- a/gremlin-python/src/main/jython/gremlin_python/structure/io/graphsonV3d0.py
+++ b/gremlin-python/src/main/jython/gremlin_python/structure/io/graphsonV3d0.py
@@ -29,7 +29,7 @@ import six
 from aenum import Enum
 
 from gremlin_python import statics
-from gremlin_python.statics import FloatType, FunctionType, IntType, LongType, TypeType, DictType, ListType, SetType
+from gremlin_python.statics import FloatType, FunctionType, IntType, LongType, TypeType, DictType, ListType, SetType, SingleByte
 from gremlin_python.process.traversal import Binding, Bytecode, P, Traversal, Traverser, TraversalStrategy, T
 from gremlin_python.structure.graph import Edge, Property, Vertex, VertexProperty, Path
 
@@ -564,6 +564,22 @@ class Int32IO(Int64IO):
     python_type = IntType
     graphson_type = "g:Int32"
     graphson_base_type = "Int32"
+
+
+class ByteIO(_NumberIO):
+    python_type = SingleByte
+    graphson_type = "gx:Byte"
+    graphson_base_type = "Byte"
+
+    @classmethod
+    def dictify(cls, n, writer):
+        if isinstance(n, bool):  # because isinstance(False, int) and isinstance(True, int)
+            return n
+        return GraphSONUtil.typedValue(cls.graphson_base_type, n, "gx")
+
+    @classmethod
+    def objectify(cls, v, _):
+        return int.__new__(SingleByte, v)
 
 
 class VertexDeserializer(_GraphSONTypeIO):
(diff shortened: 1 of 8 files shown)


gremlin-python/src/main/jython/gremlin_python/structure/io/graphsonV3d0.py
```diff
@@ -513,17 +548,83 @@ class Int64IO(_NumberIO):
     graphson_type = "g:Int64"
     graphson_base_type = "Int64"
 
+    @classmethod
+    def dictify(cls, n, writer):
+        # if we exceed Java long range then we need a BigInteger
```
jorgebay: Shouldn't we throw in case we exceed the range of a int64?

jorgebay:
It's looking good, I have a doubt: would addition of `ByteIO` and others represent a breaking change for Python users? I mean the expected type was x and now is y.

cc @aboudreault

aboudreault:
It is. But not sure if we can do otherwise if we want to add this serializer. Stephen, is the server really rejecting the data if we send a gx:Int32 in the range of the Byte possible values?

gremlin-python/src/main/jython/gremlin_python/structure/io/graphsonV3d0.py
spmallette: in python `long` is not limited to 64 bits so i thought it would be a better translation to magically convert to `BigInteger` so that java could consume it.  not good?

spmallette:
The server won't reject `g:Int32` if it's a byte size value but it won't interpret it as a java `Byte`. This change just allows python to interpret/send the type. I'm not sure what is a breaking change here though. Right now, if Gremlin Server sends a `gx;Byte` python can't deserialize it and gremlin-python has no way to send a single `gx:Byte`.  isn't this an additive feature?  

I suppose that maybe BigInteger might produce different behavior that could be considered breaking. Currently, `gx:BigInteger` coming from Gremlin Server won't be deserialized and I imagine results in error and if python has a number outside of 64bit size and sends it as `g:Long` it will fail to deserialize on Gremlin Server. So, behavior of numbers is a little different because they remove error conditions, but isn't that ok? 

or am i not thinking about something correctly? (entirely possibly that i am actually as i might be misunderstanding how python is handling some of this stuff given my middling knowledge of the language)

gremlin-python/src/main/jython/gremlin_python/structure/io/graphsonV3d0.py
aboudreault: That is correct. Note that this serializer function is used by the BigIntegerIO as well. So otherwise.. we would have no way to write a `gx:BigInteger`.

aboudreault:
oh, right.. at the moment the user will get back a non-deserialized `{''@type': 'gx:Byte', '@value': 42}`. I had in mind that the `int` would be returned as is.

I think we are good Stephen.

jorgebay:
Just wanted to double check, thanks @spmallette and @aboudreault .

This is a huge improvement, VOTE +1

PR #1062, opened 2019-02-16
TINKERPOP-2163 Improved performance of JavaTranslator method selection

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2163

Every time I think there's nothing else to squeeze out of this class I find something new. This time I cached results of `Method.getParameters()` as well as varargs checks. The former offered the bigger benefit as calls to that method forced a `clone()` of the parameter array, but the latter also helped.

Prior to the change on `tp33` we were getting:

```text
Benchmark                                             Mode  Cnt        Score       Error  Units
JavaTranslatorBenchmark.testTranslationLong          thrpt   20    25812.057 ±   504.233  ops/s
JavaTranslatorBenchmark.testTranslationMedium        thrpt   20   316019.034 ± 13852.097  ops/s
JavaTranslatorBenchmark.testTranslationShort         thrpt   20  1403109.145 ± 41460.176  ops/s
JavaTranslatorBenchmark.testTranslationWithStrategy  thrpt   20    49316.015 ±  1701.437  ops/s
```

and now, with this change, we're looking at:

```text
Benchmark                                             Mode  Cnt        Score       Error  Units
JavaTranslatorBenchmark.testTranslationLong          thrpt   20    30270.986 ±   553.973  ops/s
JavaTranslatorBenchmark.testTranslationMedium        thrpt   20   385286.380 ±  6572.709  ops/s
JavaTranslatorBenchmark.testTranslationShort         thrpt   20  1724457.005 ± 11173.083  ops/s
JavaTranslatorBenchmark.testTranslationWithStrategy  thrpt   20    60989.899 ±   533.268  ops/s
```

That's roughly 20% improvement. Not bad.

VOTE +1

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/JavaTranslator.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/JavaTranslator.java
@@ -52,7 +52,7 @@ public final class JavaTranslator<S extends TraversalSource, T extends Traversal
 
     private final S traversalSource;
     private final Class<?> anonymousTraversal;
-    private static final Map<Class<?>, Map<String, List<Method>>> GLOBAL_METHOD_CACHE = new ConcurrentHashMap<>();
+    private static final Map<Class<?>, Map<String, List<ReflectedMethod>>> GLOBAL_METHOD_CACHE = new ConcurrentHashMap<>();
     private final Map<Class<?>, Map<String,Method>> localMethodCache = new ConcurrentHashMap<>();
     private final Method anonymousTraversalStart;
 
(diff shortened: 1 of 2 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/JavaTranslator.java
```diff
@@ -273,4 +274,20 @@ private Method getStartMethodFromAnonymousTraversal() {
...
+        private final boolean hasVarArgs;
+
+        public ReflectedMethod(final Method m) {
+            this.method = m;
+
+            // the reflection getParameters() method calls clone() every time to get the Parameter array. caching it
+            // saves a lot of extra processing
+            this.parameters = m.getParameters();
+
+            this.hasVarArgs = parameters.length > 0 && parameters[method.getParameters().length - 1].isVarArgs();
```
robertdale: Replace `method.getParameters()` with `parameters`?

spmallette: thanks. i should not have missed that. i'll change it and re-run the benchmark.

nastra:
@spmallette one tiny suggestion to avoid creation of unnecessary arrays would be:
```
if (arguments.length > 0) {
            final Object[] argumentsCopy = new Object[arguments.length];
            for (int i = 0; i < arguments.length; i++) {
                argumentsCopy[i] = translateObject(arguments[i]);
            }
        }
```

spmallette:
@nastra i assume that this would accomplish what you're saying without having to change all the code that uses `argumentsCopy`:

`final Object[] argumentsCopy = arguments.length > 0 ? new Object[arguments.length] : arguments;`

nastra:
@spmallette yep LGTM

robertdale:
`docker/build.sh -i -t` success
VOTE +1

dkuppitz:
VOTE +1

PR #1064, opened 2019-02-18
TINKERPOP-2159 EventStrategy doesn't handle multi-valued properties

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-2159

Fixed multi-valued property handling in `EventStrategy` / `AddPropertyStep`. The `EventStrategy` output for the traversals added in `TinkerGraphPlayTest.testPlayDK()` is as follows:

```
Vertex [v[1]] added to graph [tinkergraph[vertices:1 edges:0]]
Vertex [v[1]] property [vp[name->null]] change to [name1] in graph [tinkergraph[vertices:1 edges:0]]
Vertex [v[1]] property [vp[name->name1]] change to [name2] in graph [tinkergraph[vertices:1 edges:0]]
Vertex [v[1]] property [vp[name->name2]] change to [name2] in graph [tinkergraph[vertices:1 edges:0]]
Vertex [v[2]] added to graph [tinkergraph[vertices:2 edges:0]]
Vertex [v[2]] property [vp[name->null]] change to [name1] in graph [tinkergraph[vertices:2 edges:0]]
Vertex [v[2]] property [vp[name->null]] change to [name2] in graph [tinkergraph[vertices:2 edges:0]]
Vertex [v[2]] property [vp[name->null]] change to [name2] in graph [tinkergraph[vertices:2 edges:0]]
Vertex [v[3]] added to graph [tinkergraph[vertices:3 edges:0]]
Vertex [v[3]] property [vp[name->null]] change to [name1] in graph [tinkergraph[vertices:3 edges:0]]
Vertex [v[3]] property [vp[name->null]] change to [name2] in graph [tinkergraph[vertices:3 edges:0]]
Vertex [v[3]] property [vp[name->name2]] change to [name2] in graph [tinkergraph[vertices:3 edges:0]]
```

It's not quite right as the old property will always be `vp[name->null]` when `Cardinality.list` is used or when `Cardinality.set` is used and the value doesn't exist yet. However, I think that's the only to make it work without any major breaking changes.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -30,6 +30,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Modified Java driver to use IP address rather than hostname to create connections.
 * Fixed potential for `NullPointerException` with empty identifiers in `GraphStep`.
 * Postpone the timing of transport creation to `connection.write` in Gremlin Python.
+* Made `EventStrategy` compatible with multi-valued properties.
 
 [[release-3-3-8]]
 === TinkerPop 3.3.8 (Release Date: August 5, 2019)
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -3484,6 +3484,12 @@ l = new ConsoleMutationListener(graph)
 strategy = EventStrategy.build().addListener(l).create()
 g = graph.traversal().withStrategies(strategy)
 g.addV().property('name','stephen')
+g.V().has('name','stephen').
+  property(list, 'location', 'centreville', 'startTime', 1990, 'endTime', 2000).
+  property(list, 'location', 'dulles', 'startTime', 2000, 'endTime', 2006).
+  property(list, 'location', 'purcellville', 'startTime', 2006)
+g.V().has('name','stephen').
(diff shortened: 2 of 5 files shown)


spmallette:
hmm - not sure what to do with this one. it doesn't appear to be fixable without a breaking change. also, if we did apply the breaking change given the current model for how we do things, we'd be opening the potential for generating massive events (i.e. realizing a list/set cardinality previous version could be really large). I almost feel like there should be a new type of event for this case and that we shouldn't use `VertexPropertyChangedEvent` for multiproperties. i dunno.....not sure if that's the solution either.

dkuppitz:
Well, the way it works now (in this PR) kinda makes sense. The current behavior has not been changed, new code paths were only introduced for the cases that threw exceptions previously:

`Cardinality.list`: Now triggers a change event, with the old property always being empty (makes sense, because the new property will always be added and not change an existing one)
`Cardinality.set`: The old property will be empty if the value doesn't exist yet, otherwise, it will be the property with the existing value.

Not too bad, IMO. The proper way would be to change `VertexProperty`, so that it combines all values with the same key (e.g. in `The Crew` graph every person would have only one `location` property w/ multiple values), but that really is too much of a breaking change I think.

spmallette:
ok - that's a fair way to rationalize it.  i think if you write upgrade/reference docs that way this approach should be fine.

PR #1067, opened 2019-02-20
TINKERPOP-2164: Reduce chance of hash collisions in Bytecode and its inner classes

nastra:
https://issues.apache.org/jira/browse/TINKERPOP-2164


jorgebay:
Can we add some unit tests around this? For example:
- assert that different instances with same values produce the same hash
- assert that same values in different order produces different hash

nastra:
@jorgebay force pushed two tests. 
Ideally we should be using e.g. guava's `EqualsTester` to make sure equals/hashCode are properly implemented. The problem however is that Java's equals/hashCode contract allows hash collisions if two objects are equal but that doesn't mean that one should produce hash collisions as otherwise perf will be negatively affected when using those objects in data structures relying on hashCode

jorgebay:
Looks good, thanks!

VOTE +1

spmallette:
VOTE +1

robertdale:
`docker/build.sh -i -t -n` success
VOTE +1

dkuppitz:
VOTE +1

PR #1069, opened 2019-02-21
TINKERPOP-1992 count has negative time in profile

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1992

Not sure if this the best way to do this but basically introduced an interface called `ProfilingAware` that makes it so that `ProfileStrategy` can notify that step that "profiling" is happening so that the step has a chance to get its life in order. For this particular problem, that meant that `group()` needed to reset its reducing barrier step (if present) for its value traversal (i.e. second `by()` modulator) so that it could internally start the profile timer (which doesn't happen otherwise. 

Tested manually for both OLTP and OLAP and it all worked.

VOTE +1


dkuppitz:
A very nice and clean solution in my opinion.

VOTE +1

PR #1073, opened 2019-02-27
TINKERPOP-2166 Cache expression to obtain the method in PSerializer

jorgebay:
https://issues.apache.org/jira/browse/TINKERPOP-2166

Implement a cache for the search of the method in P deserialization.

Includes a micro benchmark:

Before:

```
Benchmark                                 Mode  Cnt      Score      Error  Units
GraphBinaryReaderWriterBenchmark.readP1  thrpt   20  74574.850 ± 2466.040  ops/s
```

With this patch:

```
Benchmark                                 Mode  Cnt        Score       Error  Units
GraphBinaryReaderWriterBenchmark.readP1  thrpt   20  1483588.719 ± 77339.678  ops/s
```

VOTE +1


nastra:
+1

spmallette:
As soon as this merges, I will backport some of this logic to `tp33` and apply it to GraphSON (TINKERPOP-2168). 

Needs CHANGELOG entry but other than that VOTE +1

PR #1076, opened 2019-03-01
TINKERPOP-2112 Fold property() so that T values can work in any order

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2112

Also optimizes a bit to capture additional possible folds to the "AddElementStep" that would have been otherwise missed before. This should enhance performance of certain mutation traversals depending on the positioning and parameterization of the `property()` steps as more folding will occur.

Note that the folding would work properly if the `T.id` was set in the first call to `property()` and that led to:

```text
gremlin> g.addV("person").
......1>   property(T.id , "id").
......2>   property(VertexProperty.Cardinality.single, "k", "v").
......3>   property(VertexProperty.Cardinality.list, "l", 1).
......4>   property("x", "y").
......5>   property(VertexProperty.Cardinality.list, "l", 2).
......6>   property("m", "m", "mm", "mm").
......7>   property("y", "z").profile()
==>Traversal Metrics
Step                                                               Count  Traversers       Time (ms)    % Dur
AddVertexStartStep({label=[person], id=[id]})                          1           1           0.080     8.48
AddPropertyStep({key=[k], value=[v]})                                  1           1           0.137    14.50
AddPropertyStep({key=[l], value=[1]})                                  1           1           0.076     8.06
AddPropertyStep({key=[x], value=[y]})                                  1           1           0.087     9.25
AddPropertyStep({key=[l], value=[2]})                                  1           1           0.081     8.65
AddPropertyStep({mm=[mm], key=[m], value=[m]})                         1           1           0.381    40.29
AddPropertyStep({key=[y], value=[z]})                                  1           1           0.102    10.78
                                            >TOTAL                     -           -           0.947        -
```

rather than an error. After this change it now works either way, meaning `T.id` can be anywhere in the mix:

```text
gremlin> g.addV("person").
......1>   property(VertexProperty.Cardinality.single, "k", "v").
......2>   property(T.id , "id").
......3>   property(VertexProperty.Cardinality.list, "l", 1).
......4>   property("x", "y").
......5>   property(VertexProperty.Cardinality.list, "l", 2).
......6>   property("m", "m", "mm", "mm").
......7>   property("y", "z").profile()
==>Traversal Metrics
Step                                                               Count  Traversers       Time (ms)    % Dur
AddVertexStartStep({id=[id], x=[y], label=[pers...                     1           1           0.195    26.49
AddPropertyStep({key=[k], value=[v]})                                  1           1           0.139    18.89
AddPropertyStep({key=[l], value=[1]})                                  1           1           0.147    19.97
AddPropertyStep({key=[l], value=[2]})                                  1           1           0.105    14.24
AddPropertyStep({mm=[mm], key=[m], value=[m]})                         1           1           0.151    20.42
                                            >TOTAL                     -           -           0.739        -
```

Of note in this second `profile()` is that the "y" property is folded all the way back to `AddVertexStep`.

VOTE +1

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
@@ -146,6 +146,7 @@ import org.apache.tinkerpop.gremlin.structure.PropertyType;
 import org.apache.tinkerpop.gremlin.structure.T;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.apache.tinkerpop.gremlin.structure.VertexProperty;
+import org.apache.tinkerpop.gremlin.structure.util.empty.EmptyGraph;
 import org.apache.tinkerpop.gremlin.util.function.ConstantSupplier;
 
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -2081,8 +2081,20 @@ else if (value instanceof Traversal)
...
+        // if empty, it will just use the regular AddPropertyStep being tacked on to the end of the traversal as usual
+        while (endStep instanceof AddPropertyStep) {
+            endStep = endStep.getPreviousStep();
+        }
+
+        // it's possible to fold the property() into the Mutating step if there are no metaproperties (i.e. keyValues)
+        // and if the cardinality is unspecified
+        if ((endStep instanceof AddVertexStep || endStep instanceof AddEdgeStep ||
+                endStep instanceof AddVertexStartStep || endStep instanceof AddEdgeStartStep) &&
                 keyValues.length == 0 && null == cardinality) {
```
dkuppitz: I think this part should also take the graph's default cardinality (`graph().features().vertex().getCardinality(key)`) into account.

spmallette: Here's what is weird when I tried this kind of thing when i was working on the PR. We basically want to fold any `single` cardinality key/value pair, so for that the logic would be:

```java
        if ((endStep instanceof AddVertexStep || endStep instanceof AddEdgeStep ||
                endStep instanceof AddVertexStartStep || endStep instanceof AddEdgeStartStep) &&
                keyValues.length == 0 &&
                (key instanceof T || (key instanceof String &&
                        ((cardinality == VertexProperty.Cardinality.single) ||
                                (null == cardinality && asAdmin().getGraph().orElse(EmptyGraph.instance()).
                                            features().vertex().getCardinality((String) key) == VertexProperty.Cardinality.single))))) {
            ((Mutating) endStep).addPropertyMutations(key, value);
        } else {
            this.asAdmin().addStep(new AddPropertyStep(this.asAdmin(), cardinality, key, value));
            ((AddPropertyStep) this.asAdmin().getEndStep()).addPropertyMutations(keyValues);
        }
```

The problem, which will occur for graphs like TinkerGraph which don't have a schema, is that:

```text
gremlin> g.addV().property(single, 'k',1).property(single,'k',2).profile()
==>Traversal Metrics
Step                                                               Count  Traversers       Time (ms)    % Dur
AddVertexStartStep({k=[1, 2]})                                         1           1           0.138   100.00
                                            >TOTAL                     -           -           0.138        -
gremlin> g.addV().property(single, 'k',1).property(single,'k',2).valueMap()
==>[k:[1,2]]
```

If the user specifies the `single` and the key/values gets folded, the specified cardinality is lost and the graph just does whatever it wants with with the arguments given to `Graph.addVertex()`. For TinkerGraph, its default cardinality is `list` so despite the user saying otherwise, it goes in as `list`. Because the old logic uses `AddPropertyStep` the cardinality is preserved as an argument and things work according to how the user specified them. 

I also tried this logic (which only takes into account the cardinality check you suggested):

```java
        if ((endStep instanceof AddVertexStep || endStep instanceof AddEdgeStep ||
                endStep instanceof AddVertexStartStep || endStep instanceof AddEdgeStartStep) &&
                keyValues.length == 0 &&
                (key instanceof T || (key instanceof String &&
                                (null == cardinality && asAdmin().getGraph().orElse(EmptyGraph.instance()).
                                            features().vertex().getCardinality((String) key) == VertexProperty.Cardinality.single)))) {
            ((Mutating) endStep).addPropertyMutations(key, value);
        } else {
            this.asAdmin().addStep(new AddPropertyStep(this.asAdmin(), cardinality, key, value));
            ((AddPropertyStep) this.asAdmin().getEndStep()).addPropertyMutations(keyValues);
        }
```

and that seemed to work a little better as `single` gets preserved properly:

```text
gremlin> g.addV().property(single, 'k',1).property(single,'k',2).profile()
==>Traversal Metrics
Step                                                               Count  Traversers       Time (ms)    % Dur
AddVertexStartStep({})                                                 1           1           0.081    31.67
AddPropertyStep({key=[k], value=[1]})                                  1           1           0.089    34.66
AddPropertyStep({key=[k], value=[2]})                                  1           1           0.086    33.67
                                            >TOTAL                     -           -           0.257        -
gremlin> 
gremlin> g.addV().property(single, 'k',1).property(single,'k',2).valueMap()
==>[k:[2]]
```

wdyt?

dkuppitz:
VOTE +1

spmallette:
this thing is still busted..............neo4j is not happy which means other stuff probably isn't happy.

spmallette:
Ok - this is finally good now. I rolled back an eager attempt at improvement around cardinality checks on the graph when folding.  That proved to be a mess because we don't do a great job enforcing behavior for that in our test suite so every graph is a little different unfortunately.  The current approach to folding seems to fix the initial problem reported with `T.id` (and some other noted issues) without breaking existing behavior.

dkuppitz:
VOTE +1

PR #1081, opened 2019-03-08
TINKERPOP-2169/2173 Responses exceeding maxContentLength cause subsequent queries to hang

divijvaidya:
https://issues.apache.org/jira/browse/TINKERPOP-2169
https://issues.apache.org/jira/browse/TINKERPOP-2173

**Problem**
1. On receiving a CorruptedFrameException, the current code does not decrement the borrowed count on the connection and marks the connection for destruction. However, the connection never gets destroyed since the it can only be destroyed when the borrowed count is 0. This leads to unnecessary connections waiting in the bin to be cleared.
2. If a connection is serving multiple requests and one of the requests gets a CorruptedFrameException, Netty closes the underlying channel and thus, other requests on the same channel receive a ChannelClosed exception. The current code marks the server host as unavailable (thus closing out other connections for the host as well) on a ChannelClosed exception whereas that is not necessarily true. Marking the host as unavailable is too aggressive in this scenario.

**Changes**
1. Connection.isDead() logic now consists of checking the underlying channel.
2. If a connection is dead, destroy the connection, irrespective of the number of borrowed items. Informing the inflight (pending) request futures about this is already done by the response handler.
3.  If a connection is returned to the pool and it is dead, replace the connection. Do not consider the host as unavailable since a single connection dead does not imply a dead host.
4.  Add a warning log whenever a host is marked as unavailable.
5.  Replace the connection on ClosedChannel exception.
6. Reset the logging level in the integration tests after each test execution.

**Testing**

- Added a test to reproduce the leak (Problem#1). The test fails before the fixes and passes after the fixes.
- `mvn verify -DskipIntegrationTests=false`  & `mvn verify install` is successful for gremlin-server & gremlin-driver

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java
@@ -18,10 +18,7 @@
  */
 package org.apache.tinkerpop.gremlin.driver;
 
-import io.netty.handler.codec.CodecException;
-import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.tinkerpop.gremlin.driver.exception.ConnectionException;
-import org.apache.tinkerpop.gremlin.driver.exception.ResponseException;
 import org.apache.tinkerpop.gremlin.driver.message.RequestMessage;
 import io.netty.bootstrap.Bootstrap;
 import io.netty.channel.Channel;
@@ -29,9 +26,6 @@ import io.netty.channel.ChannelPromise;
 import io.netty.channel.socket.nio.NioSocketChannel;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import javax.xml.ws.Response;
-import java.io.IOException;
 import java.net.URI;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
@@ -80,7 +74,6 @@ final class Connection {
     public final AtomicInteger borrowed = new AtomicInteger(0);
     private final AtomicReference<Class<Channelizer>> channelizerClass = new AtomicReference<>(null);
 
-    private volatile boolean isDead = false;
     private final int maxInProcess;
 
     private final String connectionLabel;
@@ -133,8 +126,14 @@ final class Connection {
         return Math.max(0, maxInProcess - pending.size());
     }
 
+    /**
+     * Consider a connection as dead if the underlying channel is not connected.
+     *
+     * Note: A dead connection does not necessarily imply that the server is unavailable. Additional checks
+     * should be performed to mark the server host as unavailable.
+     */
     public boolean isDead() {
-        return isDead;
+        return (channel !=null && !channel.isActive());
     }
 
     boolean isClosing() {
@@ -205,11 +204,7 @@ final class Connection {
                         if (logger.isDebugEnabled())
                             logger.debug(String.format("Write on connection %s failed", thisConnection.getConnectionInfo()), f.cause());
 
-                        // if there is a ResponseException the write failed because of something to do with the server
-                        // or client side serialization - neither of these things mean that the host is dead. the
-                        // connection should be reusable.
-                        thisConnection.isDead = ExceptionUtils.indexOfThrowable(f.cause(), ResponseException.class) == -1;
-                        thisConnection.returnToPool();
+                        handleConnectionCleanupOnError(thisConnection, f.cause());
 
                         cluster.executor().submit(() -> future.completeExceptionally(f.cause()));
                     } else {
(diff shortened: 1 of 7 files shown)


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java
```diff
@@ -133,8 +126,14 @@ public int availableInProcess() {
...
 
+    /**
+     * Consider a connection as dead if the underlying channel is not connected.
+     *
+     * Note: A dead connection does not necessarily imply that the server is unavailable. Additional checks
+     * should be performed to mark the server host as unavailable.
+     */
     public boolean isDead() {
-        return isDead;
+        return (channel !=null && !channel.isActive());
```
spmallette: i like that you got rid of the `isDead` boolean.  nice

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java
```diff
@@ -291,6 +273,23 @@ public void returnToPool() {
...
+     * In the event of an IOException (typically means that the Connection might have been closed from the server side
+     * - this is typical in situations like when a request is sent that exceeds maxContentLength and the server closes
+     * the channel on its side) or other exceptions that indicate a non-recoverable state for the Connection object
+     * (a netty CorruptedFrameException is a good example of that), the Connection cannot simply be returned to the
+     * pool as future uses will end with refusal from the server and make it appear as a dead host as the write will
+     * not succeed. Instead, the Connection needs to be replaced in these scenarios which destroys the dead channel
+     * on the client and allows a new one to be reconstructed.
+     */
+    private void handleConnectionCleanupOnError(final Connection thisConnection, final Throwable t) {
+        if (thisConnection.isDead()) {
```
spmallette: So, just to align my understanding - you don't check 

```java
if (t instanceof IOException || t instanceof CodecException) {
``` 

anymore to replace the connection and rather just check `isDead()` because `isDead()` now encompasses situations where those two exceptions would be raised?

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerIntegrateTest.java
```diff
@@ -1102,10 +1099,12 @@ public void shouldFailOnDeadHost() throws Exception {
...
             // try to re-issue a request now that the server is down
-            client.submit("1+1").all().join();
-            fail();
+            client.submit("g").all().join();
+            fail("Should throw an exception.");
         } catch (RuntimeException re) {
-            assertThat(re.getCause().getCause() instanceof ClosedChannelException, is(true));
+            // Client would have no active connections to the host, hence it would encounter a timeout
+            // trying to find an alive connection to the host.
+            assertThat(re.getCause().getCause() instanceof TimeoutException, is(true));
```
spmallette: This is a better exception than `ClosedChannelException` - cool

spmallette:
Everything about this pull request was well done - from the JIRA creation to analysis of the problem/solutions, to the actual code/tests. Thanks!

VOTE +1

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java
divijvaidya: Yes, instead of relying on specific exceptions to determine when to replace a channel vs. returning it to the pool, we explicitly check the channel itself for health status. If the channel is active, it will be returned to the pool, otherwise, it will be replaced.

Netty closes the channel on IOExceptions and certain other non-recoverable exceptions. The connection will be replaced as per the new logic which is the same behavior as earlier.

dkuppitz:
VOTE +1

divijvaidya:
1. As per developer documentation, _"Once either consensus position is reached, you are responsible for merging to the release branch and handling any merge conflicts."_
How do I merge the pull requests? Seems like I don't have write access to the repo.
2. As recommended by the developer documentation, I have created a separate pull request for the tp33 branches at https://github.com/apache/tinkerpop/pull/1082. Can you please approve that as well?
3. I don't have permissions to modify the JIRA https://issues.apache.org/jira/browse/TINKERPOP-2169. Please mark it as resolved with the correct affected versions.

jorgebay:
A committer will merge your pull requests and move the ticket to resolved (probably today).
As `tp33` is still being merged to `master` regularly, probably we will merge #1082 and then merge `tp33` into `master` (I don't think we need votes for merging #1082 that is identical to this one).

spmallette:
I've merged #1082 to `tp33` and then merged `tp33` to `master` rather than merging this PR directly to master as that produces the git history we'd like to have. I oddly had to do this:

https://github.com/apache/tinkerpop/commit/0601c5ba593474535676b77ab9785956e6dcf6da

because tests were failing. I say "oddly" because Travis was fine with what what was there before and obviously it worked in your environment. So, I'm concluding it is something environmental in some way. If there are other thoughts on the right way to handle that, please let us know.  Also note that I added some CHANGELOG entries:

https://github.com/apache/tinkerpop/commit/46e6a4da3040a006e50e821678791b42c00ac5ea

I'm going to close this PR and the related issues in JIRA. @divijvaidya thanks again for a nice PR - looking forward to further contributions from you.

spmallette:
I knew I'd made that change before somewhere for the error message and it was in my Java 11 branch:

https://github.com/apache/tinkerpop/commit/377bc2f1babe63b2c07c73a71c9c0e665121212a

So I assume, the difference has something to do with the JDK version/type being used.

PR #1083, opened 2019-03-18
TINKERPOP-2179: Have SerializationException extend IOException

newkek:
In terms of testing, wasn't really sure what more I could have put in there... Open to other suggestions.

Will open an additional separate master PR as there are things I'd like to suggest in some of the GraphBinary exceptions thrown

### DIFF
--- a/gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/SerializationExceptionTest.java
+++ b/gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/SerializationExceptionTest.java
@@ -16,23 +16,29 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.tinkerpop.gremlin.driver.ser;
+package org.apache.tinkerpop.gremlin.driver;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+import org.apache.tinkerpop.gremlin.driver.ser.SerializationException;
+import org.junit.Test;
 
 import java.io.IOException;
 
-/**
- * @author Stephen Mallette (http://stephen.genoprime.com)
- */
-public class SerializationException extends IOException {
-    public SerializationException(final String msg) {
-        super(msg);
+public class SerializationExceptionTest {
+
+    @Test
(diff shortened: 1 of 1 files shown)


spmallette:
Is there any value to adding this exception in `tp33` if we don't use it anywhere? perhaps we can just merge #1084 and not bother with this one?

newkek:
Not really, I opened a PR on tp33 because I saw you put the "affects version 3.3.5" on the ticket so I thought it meant it would need to be fixed on tp33 too, but I'm ok with just `master`

spmallette:
ah - i just picked that a bit randomly - for some reason i didn't pick up that it was only for GraphBinary. if the change is just for that format then you can just submit for `master` and close this PR and then i suppose there is a bit of commit history to clean up in #1084

newkek:
Well it does affect other things than GraphBinary, since SerializationException is thrown general in multiple places, most notably by the `MessageSerializer`s. I think it'd be good to have that change here too, but in my cases I'm only affected by the ones that are thrown when using GraphBinary.

spmallette:
ok, well, that why i put 3.3.x on the JIRA. is it too much to ask that we implement this exception consistently across IO. otherwise the PR seems kinda incomplete (and no one is going to remember to "come back later")? if you're crazy busy, just let me know and we'll figure out a way to get it done.

newkek:
> is it too much to ask that we implement this exception consistently across IO

I'd gladly do it but I'm not sure what you mean by that?

spmallette:
I'm just saying that tp33 is just getting an exception added (not even so sure you need the test as it's not really testing anything), so if there's places it should be used in relation to GraphSON/Gryo it would be nice if that PR included that. I think that. in #1084 you actually threw the exception from GraphBinary serializers - do we have places in the GraphSON/Gryo serializers where this new exception should be thrown? or should we replace some current exception thrown in those serializers with this new one? does that make sense?

spmallette:
sorry - i now get what's happened here. never mind me.

VOTE +1

newkek:
@spmallette both this PR and the `master` PR have been rebased on the latest base branches.

The master PR does not seem to have the extra commits anymore 👍

PR #1086, opened 2019-04-02
TINKERPOP-2185 bump commons-configuration to 1.10.0.redhat-1

robertdale:
https://issues.apache.org/jira/browse/TINKERPOP-2185

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -24,6 +24,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 === TinkerPop 3.3.7 (Release Date: NOT OFFICIALLY RELEASED YET)
 
 * Ensure `gremlin.sh` works when directories contain spaces
+* Bump `commons-configuration` to 1.10.0.redhat-1
 
 
 [[release-3-3-6]]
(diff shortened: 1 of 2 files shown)


spmallette:
Travis doesn't appear to be able to deal:

```text
[INFO] Building Apache TinkerPop :: Gremlin Core 3.3.7-SNAPSHOT
[WARNING] The POM for commons-configuration:commons-configuration:jar:1.10.0.redhat-1 is missing, no dependency information available
```

robertdale:
We would have to add the RH repo to the pom - https://mvnrepository.com/repos/redhat-ga
Haven't gotten around to that yet.  I'm assuming it's public at this point.  But is that something we want to do anyway?

spmallette:
not sure - this is going to gremlin-core so having a non-standard repo might mess with things like grape/plugins maybe. i wasn't 100% sure we'd use the RedHat thing...i figured it would take a bit of thought/discussion. 

of course, without something like the RedHat artifact, the version stays stuck with the security problem along 3.3.x and 3.4.x unless we choose to accept a major breaking change in either of those lines. i still don't know when we want to allow for such things (i.e. take a major breaking change to fix a security problem). i guess it's done on a case-by-case basis perhaps - so, is the seriousness of this commons-configuration issue enough to force us to bump 3.3.x and 3.4.x or can it wait for 3.5.x?

robertdale:
Closing this for lack of supporting evidence as a security fix. TINKERPOP-2185 will be fixed via major version bump.

PR #1087, opened 2019-04-02
TINKERPOP-2191 Implement EdgeLabelVerificationStrategy

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-2191

Implemented `EdgeLabelVerificationStrategy` to allow providers to encourage (or enforce) users to always specify edge labels.

VOTE +1

### DIFF
--- a/docs/src/upgrade/release-3.3.x.asciidoc
+++ b/docs/src/upgrade/release-3.3.x.asciidoc
@@ -27,6 +27,44 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 Please see the link:https://github.com/apache/tinkerpop/blob/3.3.7/CHANGELOG.asciidoc#release-3-3-7[changelog] for a complete list of all the modifications that are part of this release.
 
+=== Upgrading for Providers
+
+==== Detection of Anti-Patterns
+
+This release adds a strategy named `EdgeLabelVerificationStrategy`. The strategy will not be added by default to the traversal source, however, providers can add it explicitly to encourage (or enforce)
+users to always specify at least one edge label. `EdgeLabelVerificationStrategy` can be configured to either throw an exception if no edge label was specified, log a warning or do both.
+
+[source,text]
+----
+gremlin> // throw an exception if edge label was not specified
+gremlin> g = TinkerFactory.createModern().traversal().withStrategies(EdgeLabelVerificationStrategy.build().throwException().create())
+==>graphtraversalsource[tinkergraph[vertices:6 edges:6], standard]
+gremlin> g.V(1).out('knows')
+==>v[2]
+==>v[4]
+gremlin> g.V(1).out()
+The provided traversal contains a vertex step without any specified edge label: VertexStep(OUT,vertex)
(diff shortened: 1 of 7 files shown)


docs/src/upgrade/release-3.3.x.asciidoc
```diff
@@ -27,6 +27,44 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
...
+[source,text]
+----
+gremlin> // throw an exception if edge label was not specified
+gremlin> g = TinkerFactory.createModern().traversal().withStrategies(EdgeLabelVerificationStrategy.build().throwException().create())
+==>graphtraversalsource[tinkergraph[vertices:6 edges:6], standard]
+gremlin> g.V(1).out('knows')
+==>v[2]
+==>v[4]
+gremlin> g.V(1).out()
+The provided traversal contains a vertex step without any specified edge label: VertexStep(OUT,vertex)
```
spmallette: This is good. A couple of questions: 

1. Can the message get any more helpful somehow? Maybe it should say more about "why" this is important? Just a suggestion: "Always specify edge labels which restrict traversal paths ensuring optimal performance - the following steps need edge labels: VertexStep(OUT,vertex)".
2. I guess my second question relates to my message - It looks like the strategy throws the exception as soon as it runs into a problem. Would it be better to collect all the problems at once and report them in batch for fixing?
3. I don't suppose we have a way to convert "VertexStep(OUT,vertex)` to simple `out()` do we? Maybe `GroovyTranslator`  somehow? would be nice to actually somehow show the user exactly where they messed up in code rather than the `Step` representation, but perhaps that's not possible. I sorta think `profile()` would be more user friendly that way if it could be done. Maybe not something for this PR, but just something to think about.

docs/src/upgrade/release-3.3.x.asciidoc
```diff
@@ -27,6 +27,44 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
...
+Type ':help' or ':h' for help.
+Display stack trace? [yN]
+----
+
+[source,text]
+----
+gremlin> // log a warning if edge label was not specified (Log4j has to be configured properly)
+gremlin> g = TinkerFactory.createModern().traversal().withStrategies(EdgeLabelVerificationStrategy.build().logWarning().create())
```
spmallette: `logWarning()` is good - nice

docs/src/upgrade/release-3.3.x.asciidoc
dkuppitz: I took the first part of your suggestion, but I still let it fail/log as soon as it hits the first step with unspecified labels. Two reasons:

1. All (verification) strategies behave like that.
2. With the current `toString()` representation it could look odd if we add multiple steps (in the worst case they all look the same).

Regarding the String representation of steps: I totally agree that it should be more user-friendly, however, we should create a ticket to do that for all error messages at once.

spmallette:
VOTE +1

robertdale:
VOTE +1

PR #1089, opened 2019-04-03
TINKERPOP-2180: Make sure user.working_dir is still respected properly

nastra:


spmallette:
merged via CTR to fix integration tests for Gremlin Console

PR #1093, opened 2019-04-08
Add some additional tests for negative Text predicates

nastra:


spmallette:
Travis seems happy so that means all GLVs should be passing with the new tests:

VOTE +1

dkuppitz:
VOTE +1

robertdale:
VOTE +1

PR #1094, opened 2019-04-09
TINKERPOP-2195: Add finally block in HttpGremlinEndpointHandler to commit transaction

moayad86:
https://issues.apache.org/jira/browse/TINKERPOP-2195

An improvement is to do the serialisation first, and add a finally block to the try...catch statement, where transaction commit can take place. More information in the ticket.

### DIFF
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpGremlinEndpointHandler.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpGremlinEndpointHandler.java
@@ -245,17 +245,19 @@ public class HttpGremlinEndpointHandler extends ChannelInboundHandlerAdapter {
                                     .result(IteratorUtils.asList(o)).create();
 
                             // http server is sessionless and must handle commit on transactions. the commit occurs
-                            // before serialization to be consistent with how things work for websocket based
-                            // communication.  this means that failed serialization does not mean that you won't get
-                            // a commit to the database
-                            attemptCommit(requestArguments.getValue3(), graphManager, settings.strictTransactionManagement);
+                            // even if the serialization is failed
+                            ByteBuf buffer;
 
                             try {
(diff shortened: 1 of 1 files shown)


spmallette:
I think that your implementation should configure `ReferenceElementStrategy` (which for 3.4.0 is enabled by default):

https://github.com/apache/tinkerpop/blob/7062595b72108e9545eaceb62e45b4ea7d946ed6/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/finalization/ReferenceElementStrategy.java

and which will detach your elements at iteration time when building the `ResponseMessage`. No new transaction will be opened on serialization. If you don't like the "references" that `ReferenceElementStrategy` produces then you could write your own strategy to produce "detached" objects just as easily to same effect - just used `DetachedFactory` rather than `ReferenceFactory`:

https://github.com/apache/tinkerpop/blob/7062595b72108e9545eaceb62e45b4ea7d946ed6/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/detached/DetachedFactory.java

That should solve your problem without this change - right?

moayad86:
ReferenceElementStrategy or a strategy which uses DetachedFactory will solve the problem of detaching the objects, but we're using a custom JSON serializer to include additional information. I guess it's not going to be possible to include those information in that case.

spmallette:
I assume that the additional information you are adding in the custom serializer is something where you do more traversing on returned objects or gather data from other sources? I'm not sure a serializer is the best place for that. A serializer should only be about transforming the data to its network payload form and probably shouldn't contain other logic. 

A custom strategy on the other hand, seems like a better place to keep that sort of logic. You could "detach" and then add your additional data (nothing says you are restricted to detaching only in the strategy) Or, if that seems unsuitable for some reason then you should probably create a custom `Channelizer` and your own implementation of that REST endpoint.

moayad86:
We're using a custom Channelizer currently for that purpose, but it seems detaching the objects and including the additional data in the strategy is a reasonable solution.

spmallette:
Seems like we've resolved this by way of the workaround options i presented. Going to close this PR without merging. Thanks.

PR #1096, opened 2019-04-11
TINKERPOP-2090: Fix Sign for Checking if Connection Pool Needs to be Repopulated

bpeav:
https://issues.apache.org/jira/browse/TINKERPOP-2090

Going through the commit history, it looks like the comparison changed in this PR used to be correct but was erroneously flipped after several modifications.

_poolSize is the maximum number of connections in the pool and NrConnections is the actual number of connections currently open. The comparison would previously always be true, causing EnsurePoolIsPopulatedAsync() to always early exit and never repopulate the pool as connections in it were disposed of and removed.

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
@@ -62,7 +62,7 @@ namespace Gremlin.Net.Driver
         private async Task EnsurePoolIsPopulatedAsync()
         {
             // The pool could have been (partially) empty because of connection problems. So, we need to populate it again.
-            if (_poolSize >= NrConnections) return;
+            if (_poolSize <= NrConnections) return;
             var poolState = Interlocked.CompareExchange(ref _poolState, PoolPopulationInProgress, PoolIdle);
             if (poolState == PoolPopulationInProgress) return;
             try


spmallette:
wow - nice. VOTE +1

jseekamp:
Just curious how long it will take before this fix makes it into the gremlin.net nuget package. I'm just wondering if I would be better off forking and building myself. Right now I'm creating a new client every time I make a request to work around this issue, but that causes other connection starving issues on my servers.

FlorianHockmann:
> Just curious how long it usually takes for a fix like this to make it into the gremlin.net nuget package.

I think it makes sense in this case to publish a prerelease version of Gremlin.Net when this is merged. That way, you and other users who run into this problem can test out whether this actually fixes the problem. It would then of course be good to get feedback from you (and others) so we can either close the bug and mark it as fixed for 3.4.2 or continue investigating if the bug is still present.

@spmallette I'm not quite sure about the process of publishing prerelease / release candidate versions: Is it OK to just publish such a version for one GLV or should we always do a full release across all packages? I'm mostly wondering what will happen if we publish 3.4.2-rc1 for Gremlin.Net and then want to publish another release candidate later as that would have to use 3.4.2-rc2 then to avoid using the same version number twice.

danielcweber:
We are runnning into the same behaviour. With a prerelease on nuget, I'd happily provide feedback. I propose just naming it 3.4.2-preview1 since preview < rc in the nuget versioning scheme. alpha or beta would work too.

jseekamp:
I went ahead and created a forked version and can confirm that this pull request fixes all the issues I was seeing. I'd be happy to provide more feedback on a pre-release version as well.

spmallette:
sorry I somehow never found my way back to this PR to merge it.  all done now. 

regarding this from @FlorianHockmann 

> I'm not quite sure about the process of publishing prerelease / release candidate versions: Is it OK to just publish such a version for one GLV or should we always do a full release across all packages? I'm mostly wondering what will happen if we publish 3.4.2-rc1 for Gremlin.Net and then want to publish another release candidate later as that would have to use 3.4.2-rc2 then to avoid using the same version number twice.

I'm still hesitant around "prereleases" given discussions i've seen on various Apache lists. I think I'd like to try to avoid them for the time being, especially since we're looking to do faster releases. I'm thinking we should setup to do 3.3.7/3.4.2 in May, but need to bring that up on the dev list.

@bpeav thanks again for this PR - looks like a big fix for folks. hope to see more contributions from you in the future!

danielcweber:
Unfortunately, I can't confirm the issue to be solved. We still run into pool draining on an Azure Function. We use CosmosDB. The message is:

All 4 connections have reached their MaxInProcessPerConnection limit of 32. Consider increasing either the PoolSize or the MaxInProcessPerConnection limit.

@FlorianHockmann This has previously been linked to CosmosDB, which may or may not be the cause. Let me know how I can help, I will free up the ressources to help getting this examined.

FlorianHockmann:
> All 4 connections have reached their MaxInProcessPerConnection limit of 32. Consider increasing either the PoolSize or the MaxInProcessPerConnection limit.

This exception is currently thrown if all connections reached their limit (for which case the text makes sense) **or** if all connections in the pool are already in _Closed_  state and where therefore removed from the pool while trying to find an available connection. I just created [TINKERPOP-2215](https://issues.apache.org/jira/browse/TINKERPOP-2215) to throw a different exception in the latter case.

As a starting point to further dig into this, a PCAP file could be helpful to check what happens right before the exception gets thrown. Is it possible to create a PCAP for debugging with Azure Functions?
Our current assumption is still that the connections are closed by Azure when they are idle for a while. It would be good to know whether that's actually the case. So, is the connection idle for some time before the exception gets thrown?
In that case we could evaluate how to proceed. I can think of a few different options, but I don't know enough about the problem right now to really evaluate which approach makes most sense. I would especially like to have a confirmation that Azure closes connections even if a ping message is sent regularly which we do by default in Gremlin.Net exactly to avoid problems like this one.

danielcweber:
>Is it possible to create a PCAP for debugging with Azure Functions

Probably out of luck for PCAP and Azure Functions, but I can do it locally (there's nothing specific to Azure Functions).

>So, is the connection idle for some time before the exception gets thrown
Definitely - the exception just happens after a day or more, and there's nothing really going on at the moment on that service, so this is most probably the case.

Quick question before I dig into the sources: It sounds as if Gremlin.net does not re-open or re-create closed connections. True?

FlorianHockmann:
> Quick question before I dig into the sources: It sounds as if Gremlin.net does not re-open or re-create closed connections. True?

It does create new connections, but it throws an exception first for failed requests. The creation of new connections is actually what was fixed by this PR, so you'll have to build yourself from `master`. The method `EnsurePoolIsPopulatedAsync` is executed for every incoming request and it fills the pool with new connections if it's at its configured size.
Replacing closed connections works like this:

1. Query comes in
1. `ConnectionPool` searches for an available connection (`GetConnectionFromPool`)
1. A closed connection is encountered (in `TryGetAvailableConnection`).
1. The closed connection is removed from the pool
1. If no connection is available (e.g., because all connections were closed), a `ConnectionPoolBusyException` is thrown.
1. Another query comes in
1. `EnsurePoolIsPopulatedAsync` notices that the pool is empty and populates it again

So, the pool should recover automatically, even after throwing a `ConnectionPoolBusyException`.

spmallette:
> Our current assumption is still that the connections are closed by Azure when they are idle for a while. 

@FlorianHockmann does Gremlin.NET need a keep-alive option? I assume we don't see this problem in true Gremlin Server implementations because of the server-side keep-alive that is sending pings to the client? does Gremlin.NET even respond to those? if not, i'm not sure i understand why we don't see it in Gremlin Server too.

danielcweber:
>So, the pool should recover automatically, even after throwing a ConnectionPoolBusyException.
Ok, there's maybe a misunderstanding. I cannot confirm that the pool doesn't ever recover. It does probably. What happens is, we definitely hit 5. in your enumeration. I tested it with a network card reset, killing all connections, just like we assume CosmosDB to. The appropriate exception is then thrown.

I think Gremlin.net should not wait for the next query to recover, but try right away.

FlorianHockmann:
@spmallette 

> does Gremlin.NET need a keep-alive option?

Gremlin.Net relies on the underlying WebSocket keepalive implementation which sends by default every 30 seconds a ping. According to [this blog post](https://blogs.msdn.microsoft.com/whereismysolution/2018/03/12/does-system-net-websockets-includes-a-keepalive-mechanism-which-automatically-takes-care-of-pingpong-control-frames/), it also responds to ping messages coming from the server.
That is why I'm wondering why Azure apparently still terminates the connections. It seems to ignore those keepalive pings for its check whether a connection is still alive.
But a PCAP file that actually shows the network traffic would give us clarity that a) pings are actually sent and answered and that b) Azure is actively terminating the connection which leads to the problem.

@danielcweber 

> I tested it with a network card reset, killing all connections, just like we assume CosmosDB to. The appropriate exception is then thrown.

Yes, that's of course expected. But that doesn't tell us whether CosmosDB terminated the connections, although that's basically the only explanation I can think of. But to decide on a course of action we would also need to know why CosmosDB terminates connections. Does it really ignore keepalive pings or are they not sent / answered for some reason?

> I think Gremlin.net should not wait for the next query to recover, but try right away.

It doesn't try it right away because the assumption was that connections were closed because the server is currently not available at all. In that case, it makes more sense in my opinion to wait for the next query and try then whether the server is available again (e.g., because it was restarted).
In your case where the connections were closed because they were inactive, it could make sense to make an attempt at creating new connections right away instead of throwing an exception.
I would however prefer if we could avoid this situation from occurring in the first place instead of trying to work around it. If we can't avoid it for some reason, then we could make an attempt at creating new connections when all connections were closed before we throw an exception.

spmallette:
i see - now that you say this, i feel like you've told me this before - sorry, if that's the case. i'll try again to get Microsoft attention on this.

danielcweber:
Sorry, of course the ISP's disconnect came before CosmosDB's, so my capture is worthless. Gotta find another way of capturing the packets unless we hear from somebody from Microsoft.

jseekamp:
@danielcweber I tried a capture yesterday and could see the ping pong packets. I tested for 35 minutes and didn't see a disconnect. Are you hosting your application that uses the gremlin.net driver in Azure as well? I think it may have something to do with Azure load-balancers on the application side that causes the disconnects. 

For the time being, I added a method to purge closed connections before the call to EnsurePoolIsPopulatedAsync(). Seems to work for now. 

https://github.com/jseekamp/tinkerpop/commit/bfc7b2b6224c60e69e64dd1cebf08e49f7bd6a7a

danielcweber:
@jseekamp Awesome, thanks a lot, I'll merge it into my fork and give it a try.

>I tested for 35 minutes and didn't see a disconnect

I think we're rather talking hours than minutes, and more towards ~24h. You might be right, the disconnect may happen on the client side rather than at CosmosDb. Haven't thought of that yet.

PR #1097, opened 2019-04-12
Fix threaded transaction tests. 

ntrhieu89:
The graph thread should be refreshed to observe the result of the committed transaction (see the discussion here https://groups.google.com/forum/#!topic/gremlin-users/2JhAtXcCucE).


spmallette:
thanks - merged as CTR to tp33/master

PR #1098, opened 2019-04-16
Allow to configure processor in Gremlin-JavaScript

dwitry:
* It is currently possible to configure [opProcessor](https://tinkerpop.apache.org/docs/current/reference/#opprocessor-configurations) by creating custom `RequestMessage` in Gremlin-Java/Groovy, Gremlin.Net and Gremlin-Python
* It [was possible](https://github.com/jbmusso/gremlin-javascript/blob/master/gremlin-client/README.md#usage) to configure opProcessor in Gremlin-JavaScript 2.7.0 before API change
* For consistency, this PR adds an option to configure [opProcessor](https://tinkerpop.apache.org/docs/current/reference/#opprocessor-configurations) in new Gremlin Javascript API
* This is useful, for example, to connect to [Gremlin Server Cypher Plugin](https://github.com/opencypher/cypher-for-gremlin/tree/master/tinkerpop/cypher-gremlin-server-plugin#usage)

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -27,6 +27,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Ensure `gremlin.sh` works when directories contain spaces
 * Enabled `ctrl+c` to interrupt long running processes in Gremlin Console.
 * Implemented `EdgeLabelVerificationStrategy`
+* Gremlin-JavaScript: added parameter to configure processor in client constructor
 
 [[release-3-3-6]]
 === TinkerPop 3.3.6 (Release Date: March 18, 2019)
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/client.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/client.js
@@ -39,6 +39,7 @@ class Client {
    * @param {String} [options.traversalSource] The traversal source. Defaults to: 'g'.
    * @param {GraphSONWriter} [options.writer] The writer to use.
    * @param {Authenticator} [options.authenticator] The authentication handler to use.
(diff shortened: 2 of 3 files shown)


spmallette:
Thanks for submitting this PR. In principle, allowing the setting of the `opProcessor` is fine by me 

VOTE +1

I'd like to wait for @jorgebay or @jbmusso to provide a more full review of the change itself before we merge. I assume that this change should target the `tp33` branch and it will need a CHANGELOG entry, but other than that it looks pretty straightforward.

As a side note (not really related to the merge-ability of this PR), I think this represents more reason for bytecode to replace scripts. This sort of advanced configuration for both client and server is more than a user should have to do to get Cypher going with a TinkerPop implementation. It will be nice when we can push past scripts and into pure bytecode, but until then I guess this is what needs to be done.

jbmusso:
Looks good to me.

VOTE +1

dwitry:
@spmallette, could you please elaborate on your side note?

In this case, a client sends query string and Cypher→Gremlin translation happens on TinkerPop server side (using Cypher for Gremlin plugin). Translation cannot be done on JavaScript client side because translation works in JVM. Please clarify how do you see the use of bytecode in this scenario?

jorgebay:
Thank for your contribution, good idea to cover it with unit tests.

VOTE +1.

@dwitry Could you add the CHANGELOG entry and base it on tp33 branch?

If you don't have time to do it, I can cherry pick it and add the changelog entry when merging it.

dwitry:
I've updated the CHANGELOG entry and based PR on tp33 branch. Build failure seems to be unrelated to my change, please correct me if I'm wrong. Anything else I can do to get this merged?

spmallette:
Sorry for taking so long with this - I missed notification of the other VOTEs that came in. That travis failure is known - I've just pushed a fix for it.  This is merged now. Thanks for contributing - hope to see you around here again!

PR #1100, opened 2019-04-25
TINKERPOP-2201: Change the typing of from() and to() to accept wildcard instead of Vertex

sel-fish:
prevent query like
```
g.addV().as("first").repeat(__.addE("next").to(__.addV()).inV()).times(5).addE("next").to(__.select("first"))
```
from failing static type checking when gremlin server use security sandbox


spmallette:
Thanks for submitting this. I guess that we can accept this change since it's a change to more permissive typing and thus not a breaking change (I think I'm interpreting that right).

The travis failure is that same nonsense we've been seeing with Docker, so no problem there. I can add a changelog entry on merge.

VOTE +1

dkuppitz:
VOTE: +1

PR #1102, opened 2019-04-28
TINKERPOP-2204 Do not consume the exception while generating metadata

divijvaidya:
https://issues.apache.org/jira/browse/TINKERPOP-2204

The current implementation may lead to a client waiting indefinitely from a response from the server. This occurs in cases when the exceptions that occurred while generating status attributes or generating metadata in the TraversalOpProcessor are consumed in the catch block without sending a response back to the client.


spmallette:
Interesting. Nice find. The travis failure is known and should be resolved now if you were to rebase. 

Note that I think this problem also exists on `tp33`:

https://github.com/apache/tinkerpop/blob/tp33/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/traversal/TraversalOpProcessor.java#L544

do you agree? if so, i guess I can merge this PR as it is, then backport the change (since it's slightly different).  It would be nice if you included a CHANGELOG entry but i can handle that on merge if you'd like. 

VOTE +1

divijvaidya:
Thank you for your review. Yes, as you identified, this bug exists in TP33 as well.

Please go ahead and add the change log entry during merge for this one, I will make sure to add one in the future PR.

PR #1103, opened 2019-04-29
Create uber jar for gremlin-console

nastra:
I believe having an uber jar for the console might make things easier
for Graph providers, because one can pull in a single jar with all the
dependencies satisfied.

`mvn clean install && mvn verify -pl gremlin-console -DskipIntegrationTests=false` passes

### DIFF
--- a/gremlin-console/pom.xml
+++ b/gremlin-console/pom.xml
@@ -212,6 +212,33 @@ limitations under the License.
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


spmallette:
I inspected the uberjar that is being generated. I have some concerns about the `META-INF/` in the jar itself with respect to licensing. I'm not sure what we need to include in there to be compliant with Apache licensing, but right now there's a bit of everything. I also think LICENSE files are copying over each other with the shading plugin. I assume that as this is a binary distribution artifact that the packaging should be similar to our zip distribution and follow that pattern, but I'll research a bit.

spmallette:
Fixed up LICENSE/NOTICE issue with 6caeb93b0d8d73b6c060b997699a7c32053db8cd and merged.

PR #1105, opened 2019-05-01
TINKERPOP-2205 Change connection management to single request per channel

divijvaidya:
https://issues.apache.org/jira/browse/TINKERPOP-2205

The code in this pull request changes the server interaction mechanism of the Gremlin open source Java client. The new code addresses problems and shortcomings discussed in the linked conversation [[1]](https://lists.apache.org/thread.html/77728cb77d4eab90f15680595e653ffc6055b74db29cbd4dcd5f0339@%3Cdev.tinkerpop.apache.org%3E). More specifically, the problems addressed are as follows:
1. Difficulty in configuring the client for optimum performance.
2. Undocumented dependency of configuration parameters on each other.
3. A bad request can impact other requests on the same channel.
4. Host is marked as dead even if it is busy serving requests.
5. No way to free up server resources if the client has stopped consuming results.
6. No differentiation between retriable and non-retriable exceptions from the application code.
7. Keep alive is only sent when a query is executing, which means that a connection open for a very long time with no query being sent will be closed by the server.
8. Race condition if the server response reaches before result queue has been registered.
9. Unpredictable behaviour if the server sends an exception followed by a genuine response for the same request.
10. A concurrent hash map (tracking pending requests) is a point of contention amongst threads.
11. Concurrent requests made at the same time would all go to the same host irrespective of load balancing set up across multiple hosts. This is because the logic of detecting the least used connection & incrementing the usage count is not thread safe. Every thread checks the least used connection at the same time, gets the same connection and goes on to increment the usage of the connection.
### Changes
1. ResultSet can be closed. 
    * This allows the client to tell the server to relinquish resources associated with this request. 
2. Single request per connection. No channel multiplexing. 
    * Impact of a rogue response (such as one which causes IOException exceeding content length) does not impact the rest of the in-flight requests. 
    * Each request has its own bandwidth.
3. Removed custom keep alive logic and replaced with Netty IdleState handler. 
    * Makes the client more robust
4. Deprecated InProcess and SimultaneousUsage configuration parameters.
    * Now the customers would have to configure only a single parameter for setting concurrency of requests.
5. Throw different exceptions to the application code which makes it easy to determine what can be retried and what not. 
6. Handle errors gracefully during WebSocket handshake.
    * Makes the client robust
7. Close the websocket channel gracefully (with a close frame).
    * Server closes the channel gracefully on receiving the close frame.
8. Use EPoll instead of Nio whenever possible. 
    * Poll provides better performance on Linux platforms
9. Run chooseConnection in an async manner using executors threads. 
    * Increases thread utilization. In general a lot of effort has been made to improve thread utilization.
10. Make client resilient to multiple response from the server for the same request.
11. Client operations do not rely on the UUID of the request provided by the server.
12. Removed the complex logic for handling a request when no connections to server is available. This logic is pushed to the application now.
     * Improves transparency and provides the applications with a better control on how to handle requests beyond the client capacity.

### Backward compatibility with 3.4.x/3.3.x
**Application layer code** - This new client is fully backward compatible and requires no change in the application layer code. The only change required will be if the application layer code is relying on certain types of exceptions thrown by the client.  

**Channelizer** - Although the channelizer interface hasn’t changed, custom implementations of the channelizer will have to change their code to work with the new client.
### Limitations
1. A client generating high TPS from a single machine will have to modify the OS setting for max number of open files, since each connection corresponds to a single file in linux OS.
### Benchmarks
During benchmark testing, there was no difference in performance. I did not do an extensive benchmark with maxing out the CPU, instead, I wanted to see the throughput at default server configuration. 
The similar performance of the new code can be attributed to channels being re-used and the additional overhead is only at the bootstrap when we do more WebSocket handshakes (due to more connections) than older code.

**New code (this PR)**
```
Benchmark                      Mode  Cnt   Score   Error   Units
MyBenchmark.largeResultQuery  thrpt    5   0.652 ± 0.082  ops/ms
MyBenchmark.longQuery         thrpt    5  23.689 ± 1.239  ops/ms
MyBenchmark.shortQuery        thrpt    5  21.864 ± 4.803  ops/ms
MyBenchmark.largeResultQuery   avgt    5  49.637 ± 7.130   ms/op
MyBenchmark.longQuery          avgt    5   1.337 ± 0.079   ms/op
MyBenchmark.shortQuery         avgt    5   1.462 ± 0.343   ms/op
```
**Old code (3.4.1)**
```
Benchmark                      Mode  Cnt   Score   Error   Units
MyBenchmark.largeResultQuery  thrpt    5   0.707 ± 0.137  ops/ms
MyBenchmark.longQuery         thrpt    5  28.227 ± 1.177  ops/ms
MyBenchmark.shortQuery        thrpt    5  30.074 ± 1.438  ops/ms
MyBenchmark.largeResultQuery   avgt    5  45.536 ± 5.074   ms/op
MyBenchmark.longQuery          avgt    5   1.150 ± 0.059   ms/op
MyBenchmark.shortQuery         avgt    4   1.062 ± 0.078   ms/op
```
**Benchmark setup**
- Tested against standard Tinkerpop server v3.4.1
- Client and server running separately on EC2 m4.4xl machines
- Default settings for Tinkerpop server (no change in worker size etc.)
- Same client configuration
- Benchmark code: https://github.com/divijvaidya/driver-benchmark-tinkerpop-2205/blob/master/src/main/java/com/divijv/MyBenchmark.java

### Testing
1. Added a new test suite. 
2. All existing tests pass.
   *  gremlin-driver: mvn clean install -DskipIntegrationTests=false
    * gremlin-server: mvn clean install -DskipIntegrationTests=false
### Post merge work
1. Write a document describing how the client works.
2. Add examples of efficient usage of client.
3. Update change log.
4. Update documentation.
### Future work
1. Add a default retry strategy for timeouts while trying to obtain a connection.
2. Add a strategy to remove a fishy host from the load balancer (without impacting existing requests).


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
```diff
@@ -53,9 +57,16 @@
 
     /**
      * Initializes the {@code Channelizer}. Called just after construction.
+     * @param connection
+     *
+     * This method is deprecated since 3.4.2/3.3.7 and has been replaces by
```
spmallette: nit: could you please use this form for deprecation in the javadoc: https://github.com/apache/tinkerpop/blob/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/Graph.java#L306

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
```diff
@@ -65,34 +76,43 @@
...
 
     /**
-     * Create a message for the driver to use as a "keep-alive" for the connection. This method will only be used if
+     * Create a message for the driver to use as a "keep-alive" for the connectionPool. This method will only be used if
      * {@link #supportsKeepAlive()} is {@code true}.
+     *
+     * This method will be deprecated in future. Keep-alive is implemented using Netty ping-pong mechanism.
+     * See {@link WebSocketIdleEventHandler}
      */
+    @Deprecated
```
spmallette: I seem to remember that the point of this method was to provide a way to create keep-alive messages specific to the `Channelizer` implementation. If websockets no longer needs that because of `WebSocketIdlEventHandler` then that's fine, but I'm not sure that there will never be a need for this. On the other hand, I wonder if we'll realistically ever use this abstraction anyway - this driver pretty much just does websockets and that's it. Maybe all such abstraction could be deprecated...not sure. Thoughts?

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/ClientConnectionIntegrateTest.java
```diff
@@ -1,107 +0,0 @@
...
-        if (name.getMethodName().equals("shouldCloseConnectionDeadDueToUnRecoverableError")) {
-            final org.apache.log4j.Logger connectionLogger = org.apache.log4j.Logger.getLogger(Connection.class);
-            connectionLogger.setLevel(previousLogLevel);
-        }
-
-        rootLogger.removeAppender(recordingAppender);
-    }
-
-    /**
-     * Reproducer for TINKERPOP-2169
```
spmallette: Is TINKERPOP-2169 no longer relevant given your changes? if not, are you testing this sceneario elsewhere?

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
```diff
@@ -379,7 +380,7 @@ public void shouldEventuallySucceedOnSameServerWithScript() throws Exception {
             fail("Should not have gone through because the server is not running");
         } catch (Exception i) {
             final Throwable root = ExceptionUtils.getRootCause(i);
-            assertThat(root, instanceOf(TimeoutException.class));
+            assertThat(root, instanceOf(ConnectException.class));
```
spmallette: This is a nicer exception that the "timeout" - cool.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerSessionIntegrateTest.java
```diff
@@ -320,39 +320,47 @@ public void shouldHaveTheSessionTimeout() throws Exception {
             cluster.close();
         }
 
-        // there will be on for the timeout and a second for closing the cluster
+        this.stopServer();
```
spmallette: was there a reason you had to `stopServer()` here to get this test to pass?

spmallette:
Nice pull request. My initial review mostly focused on getting myself familiar with the changes and I only had some high-level feedback/questions to provide.  The additional integration tests you added look really good. Performance numbers look acceptable - thanks for taking that step up front - as we are basically getting similar performance with a more simplified configuration/usage model for users.

I've mentioned my concerns earlier about trying to push a change of this size/scope into the 3.4.x line.  Seeing the actual changes themselves right now didn't do much to relax those concerns even if it is all backward compatible. Note that we plan to open the 3.5.0 line of code in the next week or so (i.e. `master` will be 3.5.0) which I still feel is a better fit for a change of this scope. If we went with 3.5.0, I think you could view this as a more clean slate and just remove all the "dead" code rather than deprecating things. What do you think - can we go with 3.5.0?

I will be doing some manual testing and additional review of this PR in the coming days. Thanks again for working on this.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
divijvaidya: Changed.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
divijvaidya: I am in favor of coupling the driver's implementation with WebSocket along with providing hooks for providers to inject additional logic (Right now, they do it by overriding the `Channelizer` by a custom implementation which extends the `AbstractChannelizer`). This would simplify code and provide flexibility.

We can extract out an interface (if required) in future releases when we have a concrete implementation of an alternative protocol (HTTP/2 maybe?)

Having said that, I am completely fine with not deprecating this method and keeping it as it is, if you think that is the way to go.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/ClientConnectionIntegrateTest.java
divijvaidya: Given that now there exists a single request per connection, there is no possibility of a corrupted frame exception leading to impact on other requests using the same connection and hence, this test is not valid anymore.

For the new implementation, a similar error scenario is tested as part of `ClientSingleRequestConnectionIntegrateTest#shouldRecoverFromConnectionCloseDueToUnRecoverableError`

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
divijvaidya: Yeah, the intention is that the users (application code) should be able to distinguish between retriable and non-retriable errors. They could write retry strategy on the timeout error but the connect error might signify something which will not fix by retrying.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerSessionIntegrateTest.java
divijvaidya: Good catch. This was due to a missing piece of functionality in the new implementation.

[1] https://github.com/apache/tinkerpop/blob/master/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java#L313

Added that now and tests passes without having to stop the server to close the session.

divijvaidya:
Thank you taking time to review. I know it is quite a big change and might take some time to review.

I would be happy to jump on a call/chat to help you understand the architecture of the changes better. Let me know if that is required.

Regarding the release timeline for this change:
1. Do we have a tentative release timeline for 3.5.0 yet?
2. I will write another create slate version for 3.5 later. Couple of additional things I want to fix as part of the clean slate version is to unify the client API, better DNS resolution for potential hosts in the cluster, additional load balancing strategies, ability for providers to plug in information in the request etc.

I understand that you are not comfortable with releasing such a large change in 3.4.x, but, I am still inclined to release this as part of 3.4.x because in addition to easier configuration, this change also makes the client more robust in handling multiple error scenarios. I would like to see this in the hands of users as soon as possible, else we would have to do additional work to solve problems in existing 3.4.x client.
The only reason I can see to delay this change to 3.5 would be the backward incompatibility with custom channelizes that providers might have written extending from AbstractChannelizer. This can be addressed in the code, if required.

To address your concerns I can do the following:
- Adding additional integration/stress/smoke tests.
- Adding a config parameter to fallback to old code.
- Provide a document on working of the new code prior to merge which can be reviewed by the community for potential bugs/design deficiency.

Let me know how you want to proceed.

divijvaidya:
> I think that leads me to my problem with merging this to 3.4.x - If someone upgrades from 3.4.1 to 3.4.2 with this change, their configuration options are no longer relevant. They will just be ignored so if the user managed to get their configuration "perfect" under the old model, they now have to change that given this upgrade. Perhaps that's a simple change (maybe??), but it's a change nonetheless.

This is not true. If someone upgrades from 3.4.1 to 3.4.2 with this change, their configuration options would still remain exactly the same. For example, let's say the old configuration was as follows:
```
maxSimultaneousUsage = 32
maxInProcess = 32
maxConnectionPoolSize = 16
(we can ignore the min limits since they are an optimization supporting lazy loading rather than a requirement)
```
This configuration would mean that the user wants to execute 16x32 queries in parallel with each of the 16 connections able to serve 32 queries simultaneously.
In the new code, we decide the max number of connections in the connection pool = [maxConnectionPoolSize + Max(maxSimultaneousUsage, maxInProcess)](https://github.com/apache/tinkerpop/pull/1105/files#diff-0521cde5e923f69a46aad7852b209677R219) which, from a users perspective maintains the same behaviour as the old configuration, i.e. executing 16x32 queries in parallel.

Does this make sense? Do you agree?

> what were you thoughts on what a "switch" would look like?

I will think a bit more about how to do this elegantly but due to the nature of invasive changes, I am afraid the "switch" is going to be a boolean controlling if/else branches across the code.

spmallette:
> which, from a users perspective maintains the same behaviour as the old configuration, i.e. executing 16x32 queries in parallel.

Thanks for this clarification. I see it now - I agree.

> I am afraid the "switch" is going to be a boolean controlling if/else branches across the code.

hmm - well, unless there aren't many such if/else situations such a change might just make the whole thing messy, but i can't imagine that we'd get lucky that way. Maybe you don't need to try to do all that. How do you feel about keeping this targeted at 3.4.x but not merging for the release of 3.4.2 next week? If we did that, we would have ~45 days of testing, review, etc for 3.4.3 (release would be middle/end of july given our current two month release cycles which I believe would continue).

divijvaidya:
1. I agree with targeting this change for v3.4.3.
2. If any providers have provided a custom channelizer by extending AbstractChannelizer, they will have to modify their channelizer implementation. Is there a way we can inform providers of such implementations?
3. Is there anything else left (from an implementation perspective) that you would like me to change? I will address the documentation/examples about the working of the client in a separate PR.

spmallette:
> I agree with targeting this change for v3.4.3.

:+1: 

> If any providers have provided a custom channelizer by extending AbstractChannelizer, they will have to modify their channelizer implementation. Is there a way we can inform providers of such implementations?

You should write an entry in the upgrade documentation: 

https://github.com/apache/tinkerpop/blob/master/docs/src/upgrade/release-3.4.x.asciidoc

> Is there anything else left (from an implementation perspective) that you would like me to change? I will address the documentation/examples about the working of the client in a separate PR.

I will be doing some manual tests this week. Since we can't merge this until after we go through our release process, it should be fine to keep pushing changes to this PR (i.e. the documentation updates). I don't really like to break up doc fixes from code changes, if possible. Is there a reason you would like to see it merge ahead of those changes? since we're waiting for 3.4.3 now there is a bit of a delay to merge anyway right?

divijvaidya:
No, there is no reason to merge this before the 3.4.2 release. We can merge it after that. Meanwhile, I will continue adding changes to this branch.

divijvaidya:
1. I have added the tests as discussed in the JIRA associated with this PR.
2. There is an issue with Netty being tracked in https://github.com/netty/netty/pull/9226 which might cause some weird behaviour while closing the client. I am on it right now and will fix it separately from this PR.

divijvaidya:
This is ready for review.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
spmallette: after thinking on this for a bit, let's not deprecate this method until we specifically want to close down the entire abstraction.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
```diff
@@ -806,6 +810,7 @@ public Builder minInProcessPerConnection(final int minInProcessPerConnection) {
          * be borrowed once to have multiple requests executed against it.  This number controls the maximum
          * number of requests whereas {@link #maxInProcessPerConnection} controls the times borrowed.
          */
+        @Deprecated
```
spmallette: please add the "@deprecated As of release 3.3.4, not direct replaced" javadoc entries for all of these methods you're deprecating.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java
```diff
@@ -18,391 +18,65 @@
...
-            throw new ConnectionException(uri, "Could not open connection", ie);
-        }
-    }
-
+    @Deprecated
+    int DEFAULT_MAX_IN_PROCESS = 4;
     /**
-     * A connection can only have so many things in process happening on it at once, where "in process" refers to
-     * the maximum number of in-process requests less the number of pending responses.
+     * The parameter InProcess has been deprecated since 3.4.3. This setting
```
spmallette: please use the standard javadoc form for the deprecation message here.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
```diff
@@ -18,476 +18,81 @@
...
-
-    public boolean isClosed() {
-        return closeFuture.get() != null;
-    }
-
+public interface ConnectionPool {
+    int DEFAULT_MAX_POOL_SIZE = 8;
     /**
-     * Permanently kills the pool.
+     * Since 3.4.3 only max pool size is used and this setting is ignored.
```
spmallette: just another place where the standard javadoc deprecation message needs to be used.

spmallette:
Are there concerns with https://github.com/netty/netty/pull/9226 ? Is that a problem we introduce as a new exposed risk as a result of this PR or is it something that can happen now with how things work in `tp34`? 

I think the code changes seems fine from the best I can tell by just scanning through. I had a few nits around the "deprecated" bits of code. If you could just do a quick pass through and add the appropriate javadoc in the form we use, that would be great. For future reference, here's some additional information in our dev docs on the topic:

http://tinkerpop.apache.org/docs/current/dev/developer/#_deprecation

No need to include the `@see` link everywhere to the JIRA i don't think....perhaps a strategically placed one somewhere would suffice. 

I would really like to see the documentation changes to consider it all a "finished" PR though. Specifically I think I'd like to see:

1. User Documentation - delete the deprecated settings from the docs and modify configuration/best practices sections as needed to reflect how the driver works now.
2. User Upgrade Documentation - something to call attention to this change and explains what was done and why (much of that is here in your PR description so you could probably copy/paste a bit) as well as what users need to do on upgrade to be sure everything works well.
3. Provider Upgrade Documentation - document any things that might possibly break as a result of this change (you mentioned "providers have provided a custom channelizer" for example).

Starting the process of testing manually now. Assuming I don't run into any problems there, when you push the documentation changes, I think this PR will be basically good to go. Thanks.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
divijvaidya: ok

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
divijvaidya: Fixed. Here and other places.

divijvaidya:
> Are there concerns with netty/netty#9226 ? Is that a problem we introduce as a new exposed risk as a result of this PR or is it something that can happen now with how things work in tp34?

The impact of the problem is that the connections to the server might not close gracefully and it would be up to the server to determine that they are idle/broken and close them. I think this problem exists in the current implementation as well and hence, should not be a regression. Though, I would suggest that we should pick up this bug fix from Netty (they have a very fast release cycle) before our next release.

> I would really like to see the documentation changes to consider it all a "finished" PR though. 

I have added some documentation. Please check and let me know your thoughts about it.

spmallette:
Travis failed the Gremlin Server build - I restarted the job so let's see what happened, but the job looked like it hung with this error repeated over and over again:

```text
[ERROR] io.netty.util.concurrent.DefaultPromise - Failed to submit a listener notification task. Event loop shut down?
java.util.concurrent.RejectedExecutionException: event executor terminated
	at io.netty.util.concurrent.SingleThreadEventExecutor.reject(SingleThreadEventExecutor.java:867)
	at io.netty.util.concurrent.SingleThreadEventExecutor.offerTask(SingleThreadEventExecutor.java:328)
	at io.netty.util.concurrent.SingleThreadEventExecutor.addTask(SingleThreadEventExecutor.java:321)
	at io.netty.util.concurrent.SingleThreadEventExecutor.execute(SingleThreadEventExecutor.java:778)
	at io.netty.util.concurrent.DefaultPromise.safeExecute(DefaultPromise.java:768)
	at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:432)
	at io.netty.util.concurrent.DefaultPromise.setFailure(DefaultPromise.java:112)
	at io.netty.channel.DefaultChannelPromise.setFailure(DefaultChannelPromise.java:89)
	at io.netty.channel.AbstractChannelHandlerContext.safeExecute(AbstractChannelHandlerContext.java:1017)
	at io.netty.channel.AbstractChannelHandlerContext.close(AbstractChannelHandlerContext.java:610)
	at io.netty.channel.AbstractChannelHandlerContext.close(AbstractChannelHandlerContext.java:465)
	at io.netty.channel.DefaultChannelPipeline.close(DefaultChannelPipeline.java:1003)
	at io.netty.channel.AbstractChannel.close(AbstractChannel.java:242)
	at io.netty.channel.pool.SimpleChannelPool.close(SimpleChannelPool.java:398)
	at io.netty.channel.pool.FixedChannelPool.access$1301(FixedChannelPool.java:40)
	at io.netty.channel.pool.FixedChannelPool$6.run(FixedChannelPool.java:480)
	at io.netty.util.concurrent.GlobalEventExecutor$TaskRunner.run(GlobalEventExecutor.java:248)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.lang.Thread.run(Thread.java:748)
```

Do you know what that is?

> I have added some documentation. Please check and let me know your thoughts about it.

lgtm

spmallette:
Travis failed in what appears to be the same fashion i think - here's the full log:

https://api.travis-ci.org/v3/job/544523997/log.txt

divijvaidya:
Changes:
1. Removed excessive logging for a genuine case (when a channel serving requests gets disconnected)
2. I had to fork the FixedChannelPool from Netty (TinkerpopFixedChannelPool) until the Netty PR mentioned above is resolved and TinkerPop starts consuming that version. That fixes the issue with the event executor terminated exceptions you mentioned above. I hope that is ok!

Some of these changes are caused due to Netty version upgrade e.g. channelInactive being called on channel close (which is the correct behaviour). Hopefully, the tests should pass now. If they don't I will work on it and let you know when this PR is ready to review.

spmallette:
There are still a lot of errors in Travis:

https://api.travis-ci.org/v3/job/545093810/log.txt

I'm returning to my thinking that we should consider making this a change on the `master` branch for 3.5.0. If we did that we could:

1. skip "deprecation" take a full breaking change and just delete the dead code
1. wait for the netty release with your PR allowing both them and us to confirm stability
1. stage up for some bigger changes to better abstract aspects of the `Client` API (which i want to start on soon and will just create conflicts with this PR which will be annoying to resolve), etc. 
1. start a feature branch from `master` and merge this PR into it so that we could collaborate better and evolve this big body of change together. 

I know that you really wanted to see this on the 3.4.x line, but perhaps you might reconsider that position. Thoughts?

divijvaidya:
The latest travis build errors are due to some container issue with errors such as `incompatible event loop type: io.netty.channel.epoll.EpollEventLoop`. It was working earlier and all tests pass on my OSX laptop. This issue seems very specific to Epoll on the container. It might be a result of either Netty upgrade or some change in the container that runs the test itself and will take some time to debug!

Nevertheless, I agree with your suggestion to move this into 3.5.x. However, I am still concerned about the prevailing bugs in the 3.4.x and 3.3.x client. Perhaps we can work on small localized changes in the existing client to plug the holes while we work on the 3.5.x story? I will file some JIRAs with the tests that repro the bugs and we can tackle them together.

spmallette:
Glad you are on board with moving this larger change off to 3.5.0. i guess we need to figure out how to move this work to a feature branch. perhaps i just create a "driver-3.5.0" branch from "master" and then you target this PR at that? From there I can just merge i those changes and that branch would be the home for stabilizing the big driver change for 3.5.0?

> Perhaps we can work on small localized changes in the existing client to plug the holes while we work on the 3.5.x story? I will file some JIRAs with the tests that repro the bugs and we can tackle them together.

+1

spmallette:
Does that merge commit need to be in this branch? I think you just needed to rebase on `master` right?

divijvaidya:
@spmallette Ready to pull into a new branch!

Where shall we discuss ideas about the new client? Do we have a gitter/IRC for TinkerPop?

spmallette:
I've cherry-pick'd your commits and squashed them to the new `driver-35` branch based on `master`. we can do major refactoring there, so please target such changes at that branch.  I will periodically rebase it on master to keep it up to date. Obviously, we'll need to coordinate a bit when that happens.

> Where shall we discuss ideas about the new client?

The expectation for Apache projects is that all project communication should happen on the dev list, so feel free to send emails there to get major themes of discussion started. If we get into some minutiae or need some quick back and forth discussion we can just probably talk on IM directly and then report our discussion back to the dev list (which might just mean updating a JIRA or commenting on a PR as we have been doing). 

closing this PR now. thanks

PR #1106, opened 2019-05-01
TINKERPOP-2199 Fixed P.within()/without() behavior in gremlin-python

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2199

Made behavior for varags usage consistent with what Java does. Was never really sure what to do with this, but after reviewing the Java implementation it prefers a Collection as a single argument and basically converts the varargs to Collection, so basically just doing the same thing here.

builds with `mvn clean install`

VOTE +1


spmallette:
Had to tack in a few more commits on this after doing #1107 which addresses this issue for JS. The additional commits basically just do the same for C# that was done for python on this PR, adds some more tests in Java and fixes the io docs for GraphSON 3.0 with respect to `P.within()/without()`.

jorgebay:
Nice, VOTE +1

PR #1107, opened 2019-05-02
TINKERPOP-2206 Fixed g:List serialization

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2206

Seems like javascript arrays were being serialized as naked JSON objects rather than g:List. Java serialization seemed to forgive that for some reason, but prevented certain operations to really work right in the process like P.within() logic from TINKERPOP-2199 which was also fixed in this change.

There may be nicer ways to correct these issues from a javascript programming perspective - I just kinda hammered something in there. Happy to hear better ways to do things.

Builds with `mvn clean install`:

VOTE +1

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/io/graph-serializer.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/io/graph-serializer.js
@@ -50,27 +50,29 @@ class GraphSONWriter {
       }
       s.writer = this;
       // Insert custom serializers first
-      this._serializers.unshift(s);
+      th
+      is._serializers.unshift(s);
     });
   }
 
   adaptObject(value) {
     let s;
-    if (Array.isArray(value)) {
-      return value.map(item => this.adaptObject(item));
-    }
+    let o = value;
     for (let i = 0; i < this._serializers.length; i++) {
       const currentSerializer = this._serializers[i];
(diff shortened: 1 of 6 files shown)


dkuppitz:
VOTE +1

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/io/graph-serializer.js
```diff
@@ -50,27 +50,29 @@ class GraphSONWriter {
...
+      is._serializers.unshift(s);
     });
   }
 
   adaptObject(value) {
     let s;
+    let o = value;
     if (Array.isArray(value)) {
-      return value.map(item => this.adaptObject(item));
+      o = value.map(item => this.adaptObject(item));
```
jorgebay: I think that by serializing child items first in case of an `Array`, it causes to make two iterations of the `Array`: one here and the other one inside `ListSerializer`.

I think that, in order to maintain compat with GraphSON2 and still use `ListSerializer` for GraphSON3, we should evaluate the `if (Array.isArray(value))` condition below:

```javascript
// After the for loop
if (s) {
  s.serialize(o);
}

if (Array.isArray(value)) {
  return value.map(item => this.adaptObject(item));
}
```

spmallette:
thanks for the suggestions @jorgebay - pushed the changes in

jorgebay:
I've pushed some more minor fixes, hope you don't mind, feel free to revert.

VOTE +1

PR #1109, opened 2019-05-07
TINKERPOP-2211 Add API which allows per-request option for bytecode

divijvaidya:
https://issues.apache.org/jira/browse/TINKERPOP-2211

Currently, the client does not have an API to submit bytecode traversal with per-request overridden options. This code change brings the bytecode submission in parity with the `submit(String)` API.

**Testing**

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
@@ -151,11 +151,41 @@ public abstract class Client {
     }
 
     /**
+     * A version of {@link #submit(Bytecode)} which provides the ability to set per-request options.
+     *
+     * @param bytecode request in the form of gremlin {@link Bytecode}
+     * @param options for the request
+     *
+     * @see #submit(Bytecode)
+     */
+    public ResultSet submit(final Bytecode bytecode, final RequestOptions options) {
+        try {
+            return submitAsync(bytecode, options).get();
+        } catch (UnsupportedOperationException uoe) {
+            throw uoe;
+        } catch (Exception ex) {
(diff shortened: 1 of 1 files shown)


spmallette:
I get the idea that someone might want to submit bytecode with `RequestOptions` but I don't see how they will do that given your change. Users don't typically send bytecode through `Client` directly - they only submit it through:

```text
g = traversal().withRemote(...)
result = g.V().out().toList()
```

so...where would they set the `RequestOptions` in that case? wonder if we could `g.with()` somehow for that...maybe:

```text
g = traversal().withRemote(...)
result = g.with("request.scriptEvalTimeout", 10000).V().out().toList()
```

thoughts?

divijvaidya:
I completely agree with your suggestion to include query params using the `with()` step but let us take it as part of a separate PR.

This change is primarily for completeness sake given that there might be users who want to use the client as follows:

```
final Cluster cluster = createGremlinCluster();
final Client client = cluster.connect().alias("g");
final int queryTimeoutMillis = 200;
final RequestOptions reqOpts = RequestOptions.build().timeout(queryTimeoutMillis).create();
final ResultSet rsFuture = client.submit(EmptyGraph.instance().traversal().V().count().asAdmin().getBytecode(), reqOpts);
// Do some application layer work here while server computes the results
...
...
// Fetch the results from the server
rsFuture.all().get(60, TimeUnit.SECONDS);
```

spmallette:
I cherry-picked your the commit in this PR into the TINKERPOP-2211 branch and added the support for `with()`. It was easy enough and I even found a bug while doing it. :) 

I've opened a fresh PR at #1110 for review/VOTE and am now closing this one. Please have a look when you get a chance to make sure all is in order - thanks.

PR #1110, opened 2019-05-08
TINKERPOP-2211 Add API which allows per-request option for bytecode

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2211

Support `RequestOptions` with the Java driver using `TraversalSource.with()`. This PR replaces #1109

```text
g = traversal().withRemote(...);
g.with(ARGS_SCRIPT_EVAL_TIMEOUT, 500L).V().has(...).out().....
```

VOTE +1

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/remote/RemoteConnection.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/remote/RemoteConnection.java
@@ -41,16 +41,6 @@ public interface RemoteConnection extends AutoCloseable {
     public static final String GREMLIN_REMOTE_CONNECTION_CLASS = GREMLIN_REMOTE + "remoteConnectionClass";
 
     /**
-     * Key for configuring a per-request timeout option for a {@code RemoteConnection} .
-     */
-    public static final String PER_REQUEST_TIMEOUT = GREMLIN_REMOTE + "timeout";
-
-    /**
-     * Key for configuring a the batch size option for a {@code RemoteConnection} .
-     */
-    public static final String PER_REQUEST_BATCH_SIZE = GREMLIN_REMOTE + "batchSize";
-
-    /**
      * Submits {@link Traversal} {@link Bytecode} to a server and returns a promise of a {@link RemoteTraversal}.
      * The {@link RemoteTraversal} is an abstraction over two types of results that can be returned as part of the
      * response from the server: the results of the {@link Traversal} itself and the side-effects that it produced.
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/remote/RemoteConnection.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/remote/RemoteConnection.java
@@ -41,6 +41,16 @@ public interface RemoteConnection extends AutoCloseable {
     public static final String GREMLIN_REMOTE_CONNECTION_CLASS = GREMLIN_REMOTE + "remoteConnectionClass";
 
     /**
+     * Key for configuring a per-request timeout option for a {@code RemoteConnection} .
+     */
+    public static final String PER_REQUEST_TIMEOUT = GREMLIN_REMOTE + "timeout";
+
+    /**
+     * Key for configuring a the batch size option for a {@code RemoteConnection} .
+     */
+    public static final String PER_REQUEST_BATCH_SIZE = GREMLIN_REMOTE + "batchSize";
+
+    /**
      * Submits {@link Traversal} {@link Bytecode} to a server and returns a promise of a {@link RemoteTraversal}.
      * The {@link RemoteTraversal} is an abstraction over two types of results that can be returned as part of the
      * response from the server: the results of the {@link Traversal} itself and the side-effects that it produced.
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
@@ -151,11 +151,41 @@ public abstract class Client {
     }
 
     /**
+     * A version of {@link #submit(Bytecode)} which provides the ability to set per-request options.
+     *
+     * @param bytecode request in the form of gremlin {@link Bytecode}
+     * @param options for the request
+     *
+     * @see #submit(Bytecode)
+     */
+    public ResultSet submit(final Bytecode bytecode, final RequestOptions options) {
+        try {
+            return submitAsync(bytecode, options).get();
+        } catch (UnsupportedOperationException uoe) {
+            throw uoe;
+        } catch (Exception ex) {
+            throw new RuntimeException(ex);
+        }
+    }
+
+    /**
      * An asynchronous version of {@link #submit(Traversal)}. Results are returned as {@link Traverser} instances and
      * are therefore bulked, meaning that to properly iterate the contents of the result each {@link Traverser#bulk()}
      * must be examined to determine the number of times that object should be presented in iteration.
      */
     public CompletableFuture<ResultSet> submitAsync(final Bytecode bytecode) {
+        throw new UnsupportedOperationException("This implementation does not support Traversal submission - use a sessionless Client created with from the alias() method");
+    }
+
+    /**
+     * A version of {@link #submit(Bytecode)} which provides the ability to set per-request options.
+     *
+     * @param bytecode request in the form of gremlin {@link Bytecode}
+     * @param options for the request
+     *
+     * @see #submitAsync(Bytecode)
+     */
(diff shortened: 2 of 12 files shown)


divijvaidya:
While we are around this code base, can we also add request ID as part of the request options as well.

VOTE +1 : Code is OK, Semantics of this change is OK

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
```diff
@@ -518,9 +548,23 @@ protected void initializeImplementation() {
...
-                        .processor("traversal").addArg(Tokens.ARGS_GREMLIN, bytecode)).create());
+                // need to call buildMessage() right away to get client specific configurations, that way request specific
+                // ones can override as needed
+                final RequestMessage.Builder request = buildMessage(RequestMessage.build(Tokens.OPS_BYTECODE)
+                                                                                  .processor("traversal")
+                                                                                  .addArg(Tokens.ARGS_GREMLIN, bytecode));
+
+                // apply settings if they were made available
+                options.getBatchSize().ifPresent(batchSize -> request.add(Tokens.ARGS_BATCH_SIZE, batchSize));
+                options.getTimeout().ifPresent(timeout -> request.add(Tokens.ARGS_SCRIPT_EVAL_TIMEOUT, timeout));
```
newkek: If `ARGS_SCRIPT_EVAL_TIMEOUT` now configures bytecode queries, should it change to something without "script"? Maybe deprecate this one for now and introduce another one and eventually remove the deprecated one

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerIntegrateTest.java
```diff
@@ -411,6 +412,34 @@ public void shouldTimeOutRemoteTraversal() throws Exception {
         }
     }
 
+    @Test
+    public void shouldTimeOutRemoteTraversalWithPerRequestOption() throws Exception {
```
newkek: Would it be possible to add a unit test that verifies that when PER_REQUEST options are set within a GraphTraversal they effectively end up in the RequestMessage's RequestOptions?

Or instead just that they end up being correctly parsed in DriverRemoteConnection.submitAsync() somehow

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/remote/RemoteConnection.java
```diff
@@ -40,6 +40,16 @@
     public static final String GREMLIN_REMOTE = "gremlin.remote.";
     public static final String GREMLIN_REMOTE_CONNECTION_CLASS = GREMLIN_REMOTE + "remoteConnectionClass";
 
+    /**
+     * Key for configuring a per-request timeout option for a {@code RemoteConnection} .
+     */
+    public static final String PER_REQUEST_TIMEOUT = GREMLIN_REMOTE + "timeout";
```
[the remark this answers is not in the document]
spmallette: @newkek  you know what? i should get rid of this all together. users should just set these parameters the same way they do if they used the `RequestMessage` counterparts. in other words, there shouldn't be "counterparts" - it should just be the options the driver implementation allows. good?

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
spmallette: i've thought of that before. lots of changes but maybe that should happen........created this: https://issues.apache.org/jira/browse/TINKERPOP-2213 to be considered as part of a separate body of work.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/remote/DriverRemoteConnection.java
```diff
@@ -210,7 +214,20 @@ public static DriverRemoteConnection using(final Configuration conf) {
     @Override
     public <E> CompletableFuture<RemoteTraversal<?, E>> submitAsync(final Bytecode bytecode) throws RemoteConnectionException {
         try {
-            return client.submitAsync(bytecode).thenApply(rs -> new DriverRemoteTraversal<>(rs, client, attachElements, conf));
+            final Iterator<OptionsStrategy> itty = IteratorUtils.map(
+                    IteratorUtils.filter(bytecode.getSourceInstructions().iterator(),
```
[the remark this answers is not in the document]
spmallette: I don't know that we'd expose `RequestOptions` as the return result from a utility method because that is a class specific to the TP driver. However, I do think it would be good to extract the code that finds the `OptionsStrategy` list from `Bytecode`.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerIntegrateTest.java
spmallette: I added a unit test by extracting the `RequestOptions` to its own method on `DriverRemoteConnection`.  Method has `protected` scope which is a little weird, but it's the easiest way to test it.

spmallette:
I think I've resolved all the issues @newkek brought up and also added the request id override that @divijvaidya suggested. Please have another look - I think it's looking good.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/remote/RemoteConnection.java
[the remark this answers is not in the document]
newkek: Hm, but then unifying those means it imposes a dependency on `gremlin-driver` to be able to parse/use the options, right?

To specify the option in any traversal (not necessarily on a remote traversal connected with the `gremlin-driver`) you need a dependency to `gremlin-driver`

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Tokens.java
```diff
@@ -34,6 +34,7 @@ private Tokens() {}
     public static final String OPS_KEYS = "keys";
     public static final String OPS_CLOSE = "close";
 
+    public static final String ARGS_OVERRIDE_REQUEST_ID = "overrideRequestId";
```
divijvaidya: Users don't know that the internal implementation uses a requestId. They should just be able to set a request ID when they want instead of "overriding" it.

Could we rename this parameter from "overrideRequestId" to "requestId"?

Also for completeness, please add this param to Client. submitAsync as well.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/remote/RemoteConnection.java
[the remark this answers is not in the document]
spmallette: yeah - you would need `gremlin-driver`. it feels wrong to me to put them in `gremlin-core` because they don't really apply to `RemoteConnection` (which could be anything) - they apply to `gremlin-driver` and `gremlin-server` which is where all aspects of the protocol live. Putting them in `gremlin-core` feels like we're saying that these options are features of all `RemoteConnection`  and that's not really the intention. 

> To specify the option in any traversal (not necessarily on a remote traversal connected with the gremlin-driver) you need a dependency to gremlin-driver

I know it would be nice if there was a unified list of tokens that could be used for every driver implementation, but I dont' think that was the point of `RemoteConnection`. The idea was that each driver could supply it's own way of doing things and that feels like it extends to these sorts of configuration options (even if they are the same configuration options). I guess if you were a user and you knew that certain config options were the same then you could just treat the config as a string key and then it would work without any additional dependency.......

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Tokens.java
spmallette: agreed - done

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/remote/RemoteConnection.java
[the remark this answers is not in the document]
newkek: > I guess if you were a user and you knew that certain config options were the same then you could just treat the config as a string key and then it would work without any additional dependency.......

Yes, the point of that comment was that you'd want to use the official Tinkerpop name for the options to parse them. Falling back to using equivalent string values does not protect from if these values change in a Tinkerpop update. But I understand your other point for locating that stuff explicitly in `gremlin-driver`/`gremlin-server`

divijvaidya:
VOTE +1 : Code is OK, Semantics of this change is OK

newkek:
VOTE +1 (non-binding)

PR #1113, opened 2019-05-14
TINKERPOP-2216 add status attribute for warnings

dalaro:
JIRA link: https://issues.apache.org/jira/browse/TINKERPOP-2216

There are a lot of alternative ways to approach this.  I'm just opening this PR, with this specific approach, because it illustrates the suggested feature a bit more clearly than the prose in the JIRA issue description.

This adds a new `Tokens` string literal with the value `"warnings"`.  The console's `DriverRemoteAcceptor` checks for this status attribute every time it receives a remote resultset, printing the value above the results (to `err` when available).  If the value is a list, each element gets passed through `String#valueOf` and printed on a separate line.  Otherwise, it's just passed through `String#valueOf` and printed once.

Testing this in isolation seems a bit tricky.  I've been testing this coupled with a remote graph running in Gremlin-Server. If anyone has suggestions about reusable test harness bits underneath `gremlin-console/src/test` that I should reuse for this, I'd be interested.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Tokens.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Tokens.java
@@ -60,4 +60,19 @@ public final class Tokens {
 
     public static final String STATUS_ATTRIBUTE_EXCEPTIONS = "exceptions";
     public static final String STATUS_ATTRIBUTE_STACK_TRACE = "stackTrace";
+    /**
+     * A {@link ResultSet#statusAttributes()} key for user-facing warnings.
+     * <p>
+     * Implementations that set this key should consider using one of
+     * these two recommended value types:
+     * <ul>
+     *     <li>A {@link java.util.List} implementation containing
+     *     references for which {@link String#valueOf(Object)} produces
+     *     a meaningful return value.  For example, a list of strings.</li>
+     *     <li>Otherwise, any single non-list object for which
+     *     {@link String#valueOf(Object)} produces a meaningful return value.
+     *     For example, a string.</li>
+     * </ul>
+     */
+    public static final String STATUS_ATTRIBUTE_WARNINGS = "warnings";
 }
--- a/gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/PluggedIn.groovy
+++ b/gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/PluggedIn.groovy
@@ -100,6 +100,11 @@ class PluggedIn {
         }
 
         @Override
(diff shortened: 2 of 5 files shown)


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Tokens.java
```diff
@@ -60,4 +60,18 @@ private Tokens() {}
 
     public static final String STATUS_ATTRIBUTE_EXCEPTIONS = "exceptions";
     public static final String STATUS_ATTRIBUTE_STACK_TRACE = "stackTrace";
+    /**
+     * A {@link ResultSet#statusAttributes()} key for user-facing warnings.
+     * <p>
+     * The console checks for this key on remote results.  If the key is present,
```
spmallette: I think it would be good to generalize this javadoc a bit to not mention the "console", since any application could essentially take advantage of the warnings that are found here.

spmallette:
Testing seems hard, especially since we don't have a reliable way to trigger warnings in raw Gremlin Server - at least, I can't think of any good ones to add. Changes for this look good though I did add a comment about a javadoc improvement. I can add changelog/upgrade notes on merge. 

VOTE +1

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Tokens.java
dalaro: Sounds good, I'll try to reword it.

dalaro: I've generalized it to talk about "recommended values" rather than console behavior.  The prose describing how things are passed through `String.valueOf` is still especially convoluted.  

This is how it reads now:

```
* A `java.util.List` implementation containing references for which `String.valueOf(Object)` produces a meaningful return value. For example, a list of strings.
* Otherwise, any single non-list object for which `String.valueOf(Object)` produces a meaningful return value. For example, a string.
```

But I could change it to just this:

```
* A list of strings
* A string
```

It's not technically as accurate -- the only client consumer is the console, and it currently does pass stuff through `String.valueOf` -- but it's safe, since those two bullets describe a subset of what's actually consumable.  It's also a lot more readable.

Please let me know if you want me to take another stab at fixing that javadoc, or if you have your own edit in mind, feel free to push it on there.  I recognize it's important to get any implementations that might use this on the same page as early as possible.

dalaro:
I've done two force-pushes today, each with one change:

* Updated the javadoc comment on `STATUS_ATTRIBUTE_WARNINGS`
* Replaced a proposed star import with a regular import

dkuppitz:
VOTE +1

PR #1115, opened 2019-05-14
Alternative approach to TINKERPOP-2217. 

danielcweber:
https://issues.apache.org/jira/browse/TINKERPOP-2217

No state variable, no counter, no concurrent queue, no send-loop. Instead, the critical section is protected by a SemaphoreSlim which allows asynchronous waits. The up-queuing of requests is implicit through the semaphore. Besides that, we don't fire-and-forget the send loop anymore and more (private) methods return Task instead of void.

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
@@ -51,12 +51,11 @@ namespace Gremlin.Net.Driver
         private readonly WebSocketConnection _webSocketConnection;
         private readonly string _username;
         private readonly string _password;
-        private readonly ConcurrentQueue<RequestMessage> _writeQueue = new ConcurrentQueue<RequestMessage>();
+        private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1);
 
         private readonly ConcurrentDictionary<Guid, IResponseHandlerForSingleRequestMessage> _callbackByRequestId =
             new ConcurrentDictionary<Guid, IResponseHandlerForSingleRequestMessage>();
         private int _connectionState = 0;
-        private int _writeInProgress = 0;
         private const int Closed = 1;
 
         public Connection(Uri uri, string username, string password, GraphSONReader graphSONReader,
@@ -81,13 +80,12 @@ namespace Gremlin.Net.Driver
 
         public bool IsOpen => _webSocketConnection.IsOpen && Volatile.Read(ref _connectionState) != Closed;
 
-        public Task<ResultSet<T>> SubmitAsync<T>(RequestMessage requestMessage)
+        public async Task<ResultSet<T>> SubmitAsync<T>(RequestMessage requestMessage)
         {
             var receiver = new ResponseHandlerForSingleRequestMessage<T>(_graphSONReader);
             _callbackByRequestId.GetOrAdd(requestMessage.RequestId, receiver);
-            _writeQueue.Enqueue(requestMessage);
-            BeginSendingMessages();
-            return receiver.Result;
+            await SendMessageAsync(requestMessage).ConfigureAwait(false);
+            return await receiver.Result.ConfigureAwait(false);
         }
 
         private void BeginReceiving()
@@ -104,7 +102,7 @@ namespace Gremlin.Net.Driver
                 try
                 {
                     var received = await _webSocketConnection.ReceiveMessageAsync().ConfigureAwait(false);
-                    Parse(received);
+                    await Parse(received).ConfigureAwait(false);
                 }
                 catch (Exception e)
                 {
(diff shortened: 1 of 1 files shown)


danielcweber:
Is there any information on how to setup the integration test environment locally? Conceptually I'm convinced the approach is sound, however, something's breaking the tests.

spmallette:
it's not easy to run the integration tests outside of maven integration right now:

http://tinkerpop.apache.org/docs/current/dev/developer/#dotnet-environment

or you can run with a full docker build: `docker/build.sh -t -i`

OR you could temporarily merge in #1075 which would give you the docker hookup to start Gremlin Server with the test configuration and then you can run the .NET integration tests directly from the IDE.

danielcweber:
Mergin #1075 looks promising. I'll give it a try. Thanks!

danielcweber:
Builds properly now.

spmallette:
Note that the `master` branch is now 3.5.0, so if this needs to go to 3.4.x then it should be rebased on `tp34`

danielcweber:
Done.

spmallette:
I've lost track of what's happening with this PR - can this be closed given the fix on #1114

danielcweber:
Well I would claim that this is the more elegant solution compared to #1114.

danielcweber:
Basically, what Connection does today is manually keep track of a send-queue, starting a send-loop on the first enqueue, exiting it when it drains. We saw that the code required to make it thread-safe with lock-freedom is quite involved. The main point of that approach, however, is just to linearize sends while not blocking queued-up send requests. So what this PR does is protect the critical region by a SemaphoreSlim which allows async waits (no blocking!), and we can get rid of the explicit queue, counters etc.

jorgebay:
Thanks for your patch @danielcweber !

I'm not sure the problem you are trying to solve, is it just making the code more readable?

The `Connection` is a low level class, C# doesn't have something like Netty that simplifies sending/receiving with thread safe channels, so IMO its OK to have low level code (queues, CAS ops, ...).

Semaphores have a cost when contended, you can look at the WaitAsync() lock here: https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/shared/System/Threading/SemaphoreSlim.cs#L623

I would find very hard to +1 a patch like this without any additional tests and/or performance tests.

jorgebay:
@FlorianHockmann What do you think?

danielcweber:
The improvement in my patch is that it achieves what the original code did (queueing up items asynchronously), but by means of a higher level construct (SemaphoreSlim vs. combination of queues, CAS, etc.). I found the original code hard to reason about, as lock-free code usually is. [My original workaround](https://github.com/apache/tinkerpop/pull/1114) already fixed a bug that had slipped through unnoticed for quite some time.

Yes, SemaphoreSlim does contain a synchrounous lock, but I don't see that this is a problem per se. The driver as a whole is not designed to be lock-free in every part, which is usually just fine.

FlorianHockmann:
I agree that the implementation of `Connection` is hard to reason about and I'm definitely open to improvements on that front! But the driver is currently lock-free which is definitely a nice property to have and I would prefer if we could keep it that way. So, we need a good reason to introduce locks again in my opinion.

Apart from that, using a `SemaphoreSlim` is a bit problematic in general as it guarantees no order in which threads can enter the semaphore. In the extreme case where the user configures the pool to use just one connection with a high number of parallel requests for that connection, the situation can occur that some writes have to wait for a long time because new writes come in and are executed first.

danielcweber:
So the direction here is: No locks in any code in the driver nor in the framework classes it uses?

FlorianHockmann:
> So the direction here is: No locks in any code in the driver nor in the framework classes it uses?

No, I wouldn't go so far to ban locks completely from all code in the driver, including its dependencies. The recently introduced `CopyOnWriteCollection` for example [also uses locks](https://github.com/apache/tinkerpop/blob/c951abe4a80a3df59e46f70c8265100f00cf22a3/gremlin-dotnet/src/Gremlin.Net/Driver/CopyOnWriteCollection.cs#L41) (so the driver isn't completely lock-free, strictly speaking). I think that's ok in this case as it only needs locks to change the size of the collection, so for resizing operations of the connection pool. We want to move these operations out of the usual execution flow of requests as much as possible, ideally they are only performed by dedicated tasks executed in the background (and only in error cases).
If we have a good reason to introduce locks in the driver (and we're confident that it won't lead to deadlocks and only block threads for such a small time period that it's basically neglectable), then we can do so in my opinion.

For this specific PR, I'm not really convinced yet that the benefit of making the code easier to reason about warrants the added lock, especially because the lock needs to be acquired for every request and because this solution comes with the other disadvantage I mentioned about not having a guaranteed order with the `SemaphoreSlim`.

PR #1116, opened 2019-05-16
TINKERPOP-2219 Upgrade Netty dependency to 4.1.32

divijvaidya:
https://issues.apache.org/jira/browse/TINKERPOP-2219

**Testing**

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -371,6 +371,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Fixed behavior of `P` for `within()` and `without()` in GLVs to be consistent with Java when using varargs.
 * Cleared the input buffer after exceptions in Gremlin Console.
 * Added parameter to configure the `processor` in the gremlin-javascript `client` constructor.
+* Bumped `Netty` to 4.1.32.
 
 [[release-3-3-6]]
 === TinkerPop 3.3.6 (Release Date: March 18, 2019)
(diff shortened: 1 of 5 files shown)


spmallette:
VOTE +1

robertdale:
Need a `CHANGELOG` entry.

spmallette:
I should have wrote that I would be fine to do that on merge. just a note about this generally...we usually do a CHANGELOG entry for netty version bumps as that is a major dependency.  

Another change that needs to be there is the update to the NOTICE files for the binaries:

https://github.com/apache/tinkerpop/blob/master/gremlin-console/src/main/static/NOTICE#L51
https://github.com/apache/tinkerpop/blob/master/gremlin-server/src/main/static/NOTICE#L52

technically, there should also be a check to see that the netty NOTICE hasn't changed in any way of substance that would trigger a change to our NOTICE. @divijvaidya you can read more about LICENSE/NOTICE stuff here in case you are interested (hint: it's not interesting, but it's what we need to do to be compliant with the ASF and its license):

http://tinkerpop.apache.org/docs/current/dev/developer/#dependencies

divijvaidya:
My bad. I missed the part about dependencies in the [dev documentation](http://tinkerpop.apache.org/docs/current/dev/developer/#dependencies). I will make the changes once we resolve what version we want to upgrade to (discussion ongoing as part of the JIRA).

divijvaidya:
Updated the PR with changes to NOTICE and changelog

robertdale:
VOTE +1

dkuppitz:
VOTE +1

PR #1117, opened 2019-05-17
Prevent more queries from failing static type checking when gremlin server use security sandbox

sel-fish:
https://issues.apache.org/jira/browse/TINKERPOP-2221

**Testing:**
gremlin-core: mvn clean install -DskipIntegrationTests=false

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -32,6 +32,7 @@ This release also includes changes from <<release-3-3-7, 3.3.7>>.
 * Provided method to override the request identifier with `RequestOptions`.
 * Added option to set per-request settings on a `Traversal` submitted via `Bytecode`.
 * Fixed the Gryo registration for `OptionsStrategy` as it was not serializing state properly.
+* Changed definitions of generic in signature of more steps to pass static type checking.
 
 [[release-3-4-1]]
 === TinkerPop 3.4.1 (Release Date: March 18, 2019)
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Traversal.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Traversal.java
@@ -341,7 +341,7 @@ public interface Traversal<S, E> extends Iterator<E>, Serializable, Cloneable, A
          * @return the updated traversal
          * @throws IllegalStateException if the {@link TraversalStrategies} have already been applied
          */
-        public default <E2> Traversal.Admin<S, E2> addStep(final Step<?, E2> step) throws IllegalStateException {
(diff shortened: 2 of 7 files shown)


spmallette:
I see you're problem - for those trying to follow along it's demonstrated by:

```text
gremlin> :> g.V().has('person','name','marko').as('m').constant('x').as('l').addE(select('l')).from(select('m')).to('m')
Script8.groovy: 1: [Static type checking] - Cannot call org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal <org.apache.tinkerpop.gremlin.structure.Vertex, java.lang.String>#addE(org.apache.tinkerpop.gremlin.process.traversal.Traversal <java.lang.Object extends java.lang.Object, java.lang.String>) with arguments [org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal <A extends java.lang.Object, B extends java.lang.Object>] 
 @ line 1, column 1.
   g.V().has('person','name','marko').as('m').constant('x').as('l').addE(select('l')).from(select('m')).to('m')
   ^

1 error
Display stack trace? [yN]n
gremlin> :> g.V().has('person','name','marko').as('m').constant('x').as('l').addE((GraphTraversal<Object,String>)select('l')).from(select('m')).to('m')
==>e[3][0-x->0]
```

That cast of `GraphTraversal<Object,String>` isn't so nice. That said, now that you submit this pull request, it's made me question my approval/merging of the first one. The scope of your changes here are wide (perhaps not even complete) and that's indicative of a bigger problem I think. 

The problem that I'm starting to realize is that we're changing the `GraphTraversal` API to account for sandboxing shortcomings. That's not the right way to go imo. If we don't like that sandboxing is forcing some of these odd casts in scripts then we should improve sandboxing itself (not work around the problem by altering the `GraphTraversal` API).

So, while I appreciate this pull request I think I'm going to have to be -1 on merge. Moreover, it's my intention to revert the changes on #1100. I think we should close both `TINKERPOP-2211` and `TINKERPOP-2201` as "Wont Do" and create a new ticket that describes the problem in terms of improving the sandbox. 

I hope that makes sense. Please let me know if you have any questions. We can leave this PR open for now in case others have thoughts on the matter. Thanks.

sel-fish:
@spmallette Thanks for your response. Of course it makes sense. I myself wonder if there is a better way to deal with that. I think its better if we can directly improve the sandbox. 
As a matter of fact, when i searched solutions to fix the problem, I found a previous [commit](https://github.com/apache/tinkerpop/commit/8ff6bdf44d6a3bc922723c0646b5ad39c3c80268) which seems tend to solve the same sort of issues.

spmallette:
Thanks @sel-fish - I've reverted #1100 with 1d7ffd0e3d2b3bc12ba5c0a21a7883f987ef5106 and then created https://issues.apache.org/jira/browse/TINKERPOP-2222 - please feel free to comment further there if you like.

PR #1119, opened 2019-05-28
gt/gte/lt/lte can throw CCE if object isn't a Comparable

nastra:


nastra:
opened https://github.com/apache/tinkerpop/pull/1120 for tp33, so this can be closed

PR #1120, opened 2019-05-28
gt/gte/lt/lte can throw CCE if object isn't a Comparable

nastra:


spmallette:
I think the CCE is a bit ambiguous and for new users not terribly helpful, but I also worry that we're blindly allowing the compare to "pass" and returning no results because two objects aren't comparable. I feel like I'd like to know why I have such a problem instead. Would it be better to improve the exception somehow and give the user a better pointer to where their error was occurring?

nastra:
IMO having the CCE is the worst choice of all and it rather indicates a programming error rather than wrong input from users. 

I would suggest that we add a precondition check that makes sure both objects are either a `Number` or a `Comparable`, since that's essentially what the code expects. If not, we throw an exception telling the user that only `Number` /  `Comparable` can be used.

wdyt @spmallette / @dkuppitz ?

nastra:
Added a precondition check now that will throw an IAE when both are not null and not a Number/Comparable for gt/gte/lt/lte.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Compare.java
```diff
@@ -167,6 +181,20 @@ public Compare negate() {
...
+        return first instanceof Number && second instanceof Number;
+    }
+
+    private static boolean bothAreComparable(Object first, Object second) {
+        return first instanceof Comparable && second instanceof Comparable;
+    }
+
+    private static void checkBothAreNumberOrComparable(Object first, Object second) {
+        if (null != first && null != second && !bothAreNumber(first, second) && !bothAreComparable(first, second)) {
+            throw new IllegalArgumentException(String.format("'%s' and '%s' need to be an instance of Number or Comparable", first, second));
```
spmallette: I think that I like this approach better as the behavior is a bit more like  the old `ClassCastException`. The error message is good because it tells you what's expected and shows the actual values being compared. Perhaps it could be refined a bit further though. What do you think of this text:

```java
String.format("Cannot compare '%s' (%s) and '%s' (%s) as both need to be an instance of Number or Comparable", first, first.getClass().getSimpleName(), second, second.getClass().getSimpleName())
```

I like the idea of having the class names in there because that's what the old CCE used to have.  It's also a bit more clear that the error is a result of one or both of the values being a problem.

robertdale: You could do away with the double checking too...
```
   if (null != first && null != second) {
      if (bothAreNumber(first, second)) {
         return NumberHelper.compare((Number) first, (Number) second) < 0;
      } else if (bothAreComparable(first, second)) {
         return ((Comparable) first).compareTo(second) < 0;
      } else {
         throw new IllegalArgumentException(String.format("Cannot compare '%s' (%s) and '%s' (%s) as both need to be an instance of Number or Comparable", first, first.getClass().getSimpleName(), second, second.getClass().getSimpleName()));
      }
   } else 
      return false;
```

spmallette:
The release branches were re-opened a day or so ago so this PR can be rebased to tp33 to be on 3.3.8-SNAPSHOT.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Compare.java
nastra: addressed both of your comments and also added a check to make sure that both are actually of the same type (with tests)

nastra: @robertdale @spmallette @dkuppitz 
can you please re-review?

CHANGELOG.asciidoc
```diff
@@ -49,6 +49,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Fixed behavior of `P` for `within()` and `without()` in GLVs to be consistent with Java when using varargs.
 * Cleared the input buffer after exceptions in Gremlin Console.
 * Added parameter to configure the `processor` in the gremlin-javascript `client` constructor.
+* gt/gte/lt/lte can throw a ClassCastException if one of the object isn't a Comparable
```
spmallette: this changelog entry needs to move to 3.3.8. i can do it on merge if you don't get to it. no worries.

nastra: that would be great thanks. I haven't checked the entry anymore after rebasing and so it stayed where it was before :(

spmallette:

VOTE +1

dkuppitz:
VOTE +1

PR #1122, opened 2019-05-31
Expose Connection Status to Client

github5775:
When a consumer of the Gremlin.Net client calls to execute a Gremlin query, i.e. "SubmitAsync," and the connection is not open, there will be a costly timeout error.  I have experienced 30 to 90 second timeouts. 

By exposing connection status, the client's consumer can check the client's connection status before making this call.  My responsiveness is excellent now, and I now use the Gremlin.Net client in PROD.

Modified code to add a non-breaking HasOpenConnection bool function on the Gremlin.Net GremlinClient object.  It can be used this way:

    if(gremlinClient.HasOpenConnection)
        await gremlinClient.SubmitAsync("g.V().range(0,1).property('hello','world')");

This property could also be negated by adding this functionality inside the SubmitAsync function.  The function's code could check each time before submit, and then reconnect before submit, or just check and return null with an error if no connection.  This would remove the decision making from the developer's code, limiting the client's flexibility.

I favor the the simpler approach of exposing the pool's status (as read-only) to the client, and while this changes the interface, for the sake of brevity, that's what I implemented.  

The ideal solution would be to allow the SubmitAsync call to fail immediately if there is a connection issue, and note that in the results.  That way the consumer can decide how to react to the results info, yet there is no wait, and no change of interface.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,0 +1,1 @@
+theme: jekyll-theme-minimal
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
@@ -82,7 +82,7 @@ namespace Gremlin.Net.Driver
         /// <inheritdoc />
         public int NrRequestsInFlight => _callbackByRequestId.Count;
         /// <summary>
-        /// Provides whether the first available connection snapshot is open
+        /// Provides whether the first available connection snapshot is open 
         /// </summary>
         public bool IsOpen => _webSocketConnection.IsOpen && Volatile.Read(ref _connectionState) != Closed;
         /// <inheritdoc />
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
@@ -84,7 +84,7 @@ namespace Gremlin.Net.Driver
         /// <summary>
         /// Provides whether the first available connection snapshot is open
         /// </summary>
-        public bool IsOpen => _webSocketConnection.IsOpen && Volatile.Read(ref _connectionState) != Closed;
+        internal bool IsOpen => _webSocketConnection.IsOpen && Volatile.Read(ref _connectionState) != Closed;
         /// <inheritdoc />
         public Task<ResultSet<T>> SubmitAsync<T>(RequestMessage requestMessage)
         {
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
@@ -41,9 +41,7 @@ namespace Gremlin.Net.Driver
         void Finalize(Dictionary<string, object> statusAttributes);
         void HandleFailure(Exception objException);
     }
-    /// <summary>
-    ///  Connection Class
-    /// </summary>
+
     public class Connection : IConnection
     {
         private readonly GraphSONReader _graphSONReader;
@@ -60,9 +58,7 @@ namespace Gremlin.Net.Driver
         private int _connectionState = 0;
         private int _writeInProgress = 0;
         private const int Closed = 1;
-        /// <summary>
-        ///   Constructor
-        /// </summary>
+
         public Connection(Uri uri, string username, string password, GraphSONReader graphSONReader,
             GraphSONWriter graphSONWriter, string mimeType, Action<ClientWebSocketOptions> webSocketConfiguration)
         {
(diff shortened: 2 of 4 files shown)


github5775:
I can compile and build the code with no errors nor warning in Visual Studio, but failing here.  Please advise.  Thank you.

spmallette:
I think that Travis is just upset because you added this file:

https://github.com/apache/tinkerpop/pull/1122/files#diff-aeb42283af8ef8e9da40ededd3ae2ab2

and it has no Apache license header like:

https://github.com/apache/tinkerpop/blob/master/gremlin-server/conf/gremlin-server-classic.yaml#L1-L16

Of course, I'm not sure what that file is for - perhaps you didn't mean to add it and it can be removed? That too would fix the issue.

spmallette:
For now, I'll leave it to others to comment on the technical aspects of this pull request, but from an administrative point of view:

1. Note that this PR points at the `master` branch which is now the 3.5.x line of code - thus, this may no longer be the desired target for this body of work. Perhaps it should be rebased to `tp34`
1. The commit history is a bit muddy and includes a merge commit from `master`. Given the final disposition of the PR as it is now, the changeset is fairly small so and the commit history is mostly tweaks/cleanup/fixes/retries - it seems like it would be best to simply squash all of these commits to a single one.
1. Not sure what others think but a JIRA issue for reference might be nice given the nature of the change. At a minimum though, we should have a [CHANGELOG entry](https://github.com/apache/tinkerpop/blob/master/CHANGELOG.asciidoc) and if "responsiveness" is that much better and this is deemed a worthy improvement that we should promote on release, then it would be nice to have some upgrade docs (in which case, I think creating an issue in JIRA is probably more necessary). 

@github5775 perhaps it's smart not to act on any of these changes just yet. it's probably best to await an actual technical review of your changes before investing energy on these administrative items. thanks for contributing

github5775:
@spmallette, agree with all of your comments.  The issue is real, and I solved it for myself for now, but I would like to help solve it at the library level.  As mentioned, ideally the solution would not require any interface change...this was a short term fix.  Please let me know if I can be of any help going forward.

FlorianHockmann:
@github5775 First of all, thanks for reporting this problem and putting in the effort to improve the driver!
I would just suggest that you [create an issue](https://issues.apache.org/jira/browse/TINKERPOP/) first in the future as that allows us to discuss about the problem and different ways to handle it before you create a PR.

I'm actually not sure whether we want to expose the connection status to users. What are they supposed to do with that information?
I think it makes more sense for us to decide whether we want to fail fast in this case or attempt to re-connect to the server (and then fail if that doesn't work).

The second approach seems to make most sense as it allows us to handle the problem without the user even noticing it and if we can't reconnect, it becomes easier for users to handle that as they just have to catch an exception.
A closed connection should be an exceptional case and therefore we shouldn't require users to check before each request whether the connection is closed in my opinion.

@jorgebay What's your take on this?

jorgebay:
I agree with @FlorianHockmann, I wouldn't want users to have to check the connection status before submitting a traversal.

We can continue discussing in a ticket if necessary. In any case, thanks @github5775 for your patch and I look forward for future contributions.

github5775:
@FlorianHockmann, thank you for this excellent driver, it's what allows us to get the most out of Cosmos.  If this wasn't reality, I would have pushed off the use of a graph DB, even though it's the right fit for my use case.

Yes, conceptually, whatever I am doing externally, we can realize internally, without exposure to the consumer, other than the raising of an error.

If you could provide guidance as to which code you would like modified, I would be happy to help out.

FlorianHockmann:
I just looked into this and I'm actually not sure I really understand the problem you are trying to solve here.
You are using `Connection.IsOpen` to check if the pool has a connection that is considered _open_.
You described the problem like this:

> When a consumer of the Gremlin.Net client calls to execute a Gremlin query, i.e. "SubmitAsync," and the connection is not open, there will be a costly timeout error. I have experienced 30 to 90 second timeouts.

but I don't understand yet how this can occur and how this PR can improve the situation, considering that [the `ConnectionPool` already checks](https://github.com/apache/tinkerpop/blob/d3ad5dee67435189e7bedd1a7ee9580d4cd629f7/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs#L127) the `Connection.IsOpen` property for all connections before it uses one and then throws an exception if no connection is open.

If no connection is open, you shouldn't experience timeouts, you should be getting an exception. Is that really not the case?

github5775:
@FlorianHockmann great question.  yes, i saw this in TryGetAvailableConnection, and planned to follow this, but i had PROD issues on some other threads, so i put a band-aid, notified you via PR, and moved on. now that i see the Jira link, i will communicate there.  

without digging too deeply, i see that HasOpenConnection checks both IsOpen and _connectionState, but ConnectionPool only checks IsOpen.  this allows Submit to BeginSendingMessages. and following this down, the SendMessageAsync call may snag on a dead connection, which still returns true for IsOpen, but its state may not be correct.  now that i have some breathing room from the PROD issues, i will try to reproduce this to be able to tell you the exact line where this manifests.  then i will try to sign up for the jira group and add it as an issue.

@jorgebay looking forward to be able to contribute

FlorianHockmann:
> i see that HasOpenConnection checks both IsOpen and _connectionState, but ConnectionPool only checks IsOpen.

`HasOpenConnection` checks `Connection.IsOpen` which looks like this:

```c#
public bool IsOpen => _webSocketConnection.IsOpen && Volatile.Read(ref _connectionState) != Closed;
```

so it already uses both.

> i will try to reproduce this to be able to tell you the exact line where this manifests. then i will try to sign up for the jira group and add it as an issue.

Great! 

We don't necessarily need the exact line for a JIRA issue (although that would of course help). So, as soon as you have something that helps us to understand the issue, you can create an issue in JIRA so we can start discussing about how to resolve it.

I think we can then close the PR now?

github5775:
Yes, they both call that IsOpen, so I will look deeper.  Yes, we can close the PR.

github5775:
Just back from vacation, added [issue ](https://issues.apache.org/jira/browse/TINKERPOP-2268)in JIRA

PR #1125, opened 2019-06-06
TINKERPOP-2239 Removed previously deprecated SSL configs

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2239

These were originally deprecated way back in 3.2.10. Most folks should have moved on to the newer configurations by now.

VOTE +1


robertdale:
VOTE +1

dkuppitz:
VOTE +1

PR #1126, opened 2019-06-07
TINKERPOP-2237: Prevent exception when closing a session that doesn't exist

nastra:
https://issues.apache.org/jira/browse/TINKERPOP-2237


spmallette:
A bit surprised we didn't see a test failure somewhere as a result of this change. travis doesn't apparently run those test. Did you happen to run:

`mvn clean install -DskipTests && mvn verify -pl gremlin-server -DskipIntegrationTests=false`

does that pass? 

I think i'll add that test variation to the travis builds.

nastra:
I haven't run it because I was relying on the travis build actually. But running it now. Will comment later with the outcome

nastra:
@spmallette all tests passed

spmallette:
VOTE +1

dkuppitz:
VOTE +1

PR #1127, opened 2019-06-11
TINKERPOP-2241

sandszhouSZ:
https://issues.apache.org/jira/browse/TINKERPOP-2241

fix Client exception don't match Server exception when server throw StackOverflowError


spmallette:
Thanks for the fix. I think that this change makes sense. Could you please clean up the commit history a bit though? You have two commits on this pull request when one would be enough (just rebase/squash).  When you rebase, could you please target the `tp33` branch as I think the problem goes all the way back to the 3.3.x line of code. Once that's done I think that we can get it merged.

Also, if you want to make this pull request fully complete, please add a [CHANGELOG entry](https://github.com/apache/tinkerpop/blob/7a18d92b018050324417f56fe3481a6797539df1/CHANGELOG.asciidoc#tinkerpop-338-release-date-not-officially-released-yet) - i can handle it myself though when i merge if you like.

VOTE +1 (note to other reviewers who might come along, i will probably merge this small change via CTR)

sandszhouSZ:
thanks,i open a new pr here: https://github.com/apache/tinkerpop/pull/1131

PR #1129, opened 2019-06-11
TINKERPOP-1084 Branch option tokens should be allowed to be traversals

dkuppitz:
https://issues.apache.org/jira/browse/TINKERPOP-1084

Allow predicates and traversals to be used as options in `BranchStep`.

```
gremlin> g.V().hasLabel("person").
......1>   project("name", "age", "comments").
......2>     by("name").
......3>     by("age").
......4>     by(branch(values("age")).
......5>          option(any, bothE().count()).
......6>          option(29, constant("almost old")).
......7>          option(__.is(32), constant("looks like josh")).
......8>          option(lt(29), constant("pretty young")).
......9>          option(lt(35), constant("younger than peter")).
.....10>          option(gte(30), constant("pretty old")).fold())
==>[name:marko,age:29,comments:[3,almost old,younger than peter]]
==>[name:vadas,age:27,comments:[1,pretty young,younger than peter]]
==>[name:josh,age:32,comments:[3,looks like josh,younger than peter,pretty old]]
==>[name:peter,age:35,comments:[1,pretty old]]
```

I will squash the 2 commits before merging.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -35,6 +35,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Fixed bug in `MatchStep` where the correct was not properly determined.
 * Fixed bug where client/server exception mismatch when server throw StackOverflowError
 * Prevent exception when closing a session that doesn't exist
+* Allow predicates and traversals to be used as options in `BranchStep`.
 
 [[release-3-3-7]]
 === TinkerPop 3.3.7 (Release Date: May 28, 2019)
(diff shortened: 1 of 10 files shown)


spmallette:
Nice that we didn't have to change the API on `GraphTraversal` to support this.  A couple issues:

1. I think some upgrade docs would be nice in this case, because this is a neat feature to call attention to.
2. I think the test you added is fine, but the problem is that it doesn't validate operations for GLVs. I assume GLVs will work, but it would be nice to see a test added that would validate that.

spmallette:
VOTE +1

PR #1131, opened 2019-06-12
Fix client / server exception mismatch when server throw StackOverflo…

sandszhouSZ:
…wError


sandszhouSZ:
old pr refer:  https://github.com/apache/tinkerpop/pull/1127

PR #1133, opened 2019-06-12
Proposed typo fix and minor grammatical stuff

jon-scho:
I didn't think "heterogenous" was a word, but I looked it up, and it is!  :-)  But "heterogeneous" was intended here, right?

In the Community Contributions section, I think either of the constructions

     represent both [this] and [that]
or
     represent [this] as well as [that]

would be fine (but not "represent both [this] as well as [that]").

Are the proposed added hyphens in "Cypher to Gremlin traversal" -> "Cypher-to-Gremlin traversal" okay?


spmallette:
This was all generally fine. Not sure about the hyphen on the cypher-to-gremlin thing, so you might see that revert at some point but its merged for now. Note that the web site needs to be republished for these changes to appear. My intention is to wait for a more critical set of changes before doing that. Thanks!

PR #1138, opened 2019-06-13
typo fix to tp34

jon-scho:


spmallette:
it would be nice if you could batch up these small changes into common groups a little bit so that i can just review a bunch of modifications at once. it's a little more efficient that way. anyway, thanks for the fixes.

jon-scho:
CLASSIFICATION: UNCLASSIFIED

I was just about to ask whether that'd be helpful.  Okay, will do!

-----Original Message-----

PR #1142, opened 2019-06-14
Misc. edits, minutes 6 thru 30

jon-scho:
Misc. notes on the changes proposed here:
1.  Inserted "(JVM)" after first mention of "Java Virtual Machine" (since, later in the document, the abbreviation appears alone).
2.  Removed "etc." from end of list introduced by "e.g." (since the "etc." is implied by the use of "e.g.").
3.  Broke up the sentence that says to "recall that `id` is "reserved" for special usage in TinkerPop and is a member of the enum, `T`" into two sentences. Rationale: the reader is asked to "recall" two things (that `id` is reserved, and that it's a member of `T`), but only the first of those two things have been stated up to this point in the document (unless I've missed the other!).  :-)
4.  Slightly reworded a couple of sentences ("Think of it" -> "Think of `'id`) in a way that seems clearer to me.  (Unless I've misunderstood and messed things up there!)
5.  Inserted a '+' character at start and end of backtick-quoted passages containing a double underscore (so the double underscore won't be hidden).  (Did I do this right?)
6.  *** Removed the sentence that had said that the label for the edge was the "first parameter supplied", since it dated from a version in which the addEdge() step included a list of multiple parameters between its parentheses.  IS THIS OKAY? ***
7.  I took "your graph databases documentation" to mean "the documentation of your graph databases" (plural), so I changed "databases" to "databases'" (making it possessive); is this okay?

Places where I had questions about the wording but didn't make a change:

1.  'responsible for creating "lop"' -- well, we can see from the output of the last 2 commands entered that 'marko' created the *object* represented by v[3] and that the people who created v[3] are marko, josh, and peter ... but we have to go back to the 'Graph Traversal - Staying Simple' section to see that v[3] is 'lop' -- right?  Is the existing wording (specifically mentioning 'lop' by name) okay here?

2.  In the first sentence of the paragraph under The Final Ten Minutes -- "... we'll look at TinkerPop from a higher level and introduce different features of the stack in order to orient you with what it offers" -- , I didn't know whether the pronoun 'it' was referring to "TinkerPop" or to "the stack".  Is the wording okay as it is?

### DIFF
--- a/docs/src/tutorials/getting-started/index.asciidoc
+++ b/docs/src/tutorials/getting-started/index.asciidoc
@@ -128,7 +128,7 @@ provides him guidance on how to execute his trip around the `Graph`.
 
 There are several ways to create a `TraversalSource`. The example above uses the
 link:http://tinkerpop.apache.org/docs/x.y.z/reference/#connecting-embedded[embedded] style and is an approach
-restricted to languages using the Java Virtual Machine. Other methods are similar in form, but are not the focus of
+restricted to languages using the Java Virtual Machine (JVM). Other methods are similar in form, but are not the focus of
 this tutorial. See the Reference Documentation for more information on the different ways of
 link:http://tinkerpop.apache.org/docs/x.y.z/reference/#connecting-gremlin[connecting with Gremlin].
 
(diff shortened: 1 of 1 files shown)


spmallette:
Regarding your first question, I guess "v[3]" in the previous example isn't clear. As for your second question "TinkerPop" and "the stack" are one in the same. I've pushed some changes to help clarify.

PR #1118, opened 2019-05-22
TINKERPOP-2224 Detect and fix iterator leaks

divijvaidya:
https://issues.apache.org/jira/browse/TINKERPOP-2224

## Motivation
During the execution of the query the underlying provider layer might open some resources such as cursors or iterators which should be closed on completion of execution of the query. This is largely done by calling Traversal.close() method at the end of the traversal.
However, there are bugs in the query processor which lead to a leak of these resources. 

As an example, consider the following query:

`g.V().not(__.V().out()).fold().unfold()`

The implementation of not() step creates a child traversal which would open an iterator(aka resource) at the storage layer but due to a bug in TraversalUtil.java, it will never close this traversal (even on calling Traversal.close() on the global Traversal). This leads to open resources on the storage layer.

The purpose of this task is to add a framework for detecting an iterator leak and consequently fixing the bugs unearthed by the new framework.

## Changes
1. Add a framework to the existing test suite which would detect iterator leak during execution of every query in the test suite.
2. Fix the bugs causing iterator leak at multiple places
3. Provide an annotation to skip check for iterator leak which require fixing. To be taken up as a separate task.

## Testing
- After adding the new assertion, the existing test cases fail and after fixing the bugs, they pass.
- gremlin-core: mvn clean install -DskipIntegrationTests=fals
- gremlin-test: mvn clean install -DskipIntegrationTests=fals
- tinkergraph-gremlin: mvn clean install -DskipIntegrationTests=fals

PR #1144, opened 2019-06-18
TINKERPOP-2238 Fixed all iterator leaks

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2238

Fixed all the iterator leaks. Remove the temporary "ignore leak" annotation. 

VOTE +1


divijvaidya:
+1 - I am aware of the logic and the code change look good.

Thanks for completing this!

spmallette:
@divijvaidya i was reviewing a bit in preparation for release next week and I think i have some questions related to this change:

1. Was there a reason why `TinkerGraphIterator` didn't implement our `CloseableIterator`?
2. Going a step further, was there a reason not to somehow fold the functionality of the `TinkerGraphIterator` into `CloseableIterator` infrastructure?

It seems like, at a minimum, we should do item 1 which is pretty painless. For item 2, I guess `DefaultCloseableIterator` could take much of the functionality of the `TinkerGraphIterator` and then we'd use `CloseableIterator.asCloseable()` in these sorts of places:

https://github.com/apache/tinkerpop/pull/1118/files#diff-e146457e1c98d0e94695d31b0e5ed6f4

thoughts?

divijvaidya:
We can do 1 but I don't see a lot of benefit of doing this. Having said that, I don't have any strong opinion on this. Will change.

But for 2, I would keep prefer that `CloseableIterator` and `DefaultCloseableIterator` act as extensions to vanilla iterator operators (as their name suggests) without any domain specific business logic.

`TinkerGraphIterator` contains storage specific logic, i.e. knowledge of when to increment/decrement `StoreIteratorCounter` (different storage might release the iterator at different events) and hence, I am not in favor of adding its functionality to `DefaultCloseableIterator`.

spmallette:
>  Having said that, I don't have any strong opinion on this. Will change.

ok...i will look for a PR. i think that we should consistently try to use that `CloseableIterator` interface. it was created explicitly for this purpose (requested by OrientDB originally i believe). 

I'm going to sleep on your reasoning for 2, but what you're saying there sounds right to me.

PR #1145, opened 2019-06-18
TINKERPOP-2243 Add userAgent to RequestOptions.

BrynCooke:
https://issues.apache.org/jira/browse/TINKERPOP-2243

This allows graph providers to tailor their error messages to the client.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,7 +29,7 @@ This release also includes changes from <<release-3-3-8, 3.3.8>>.
 * Modified `NumberHelper` to better ignore `Double.NaN` in `min()` and `max()` comparisons.
 * Bump to Netty 4.1.36.
 * Bump to Groovy 2.5.7.
-* TINKERPOP-2243 Added `userAgent` to RequestOptions. Gremlin Console sends `Gremlin Console` as the `userAgent`.
+* Added `userAgent` to RequestOptions. Gremlin Console sends `Gremlin Console/<version>` as the `userAgent`.
 
 [[release-3-4-2]]
 === TinkerPop 3.4.2 (Release Date: May 28, 2019)
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,7 @@ This release also includes changes from <<release-3-3-8, 3.3.8>>.
 * Modified `NumberHelper` to better ignore `Double.NaN` in `min()` and `max()` comparisons.
 * Bump to Netty 4.1.36.
 * Bump to Groovy 2.5.7.
+* TINKERPOP-2243 Added `userAgent` to RequestOptions. Gremlin Console sends `Gremlin Console` as the `userAgent`.
 
 [[release-3-4-2]]
 === TinkerPop 3.4.2 (Release Date: May 28, 2019)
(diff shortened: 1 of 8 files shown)


CHANGELOG.asciidoc
```diff
@@ -29,6 +29,7 @@ This release also includes changes from <<release-3-3-8, 3.3.8>>.
 * Modified `NumberHelper` to better ignore `Double.NaN` in `min()` and `max()` comparisons.
 * Bump to Netty 4.1.36.
 * Bump to Groovy 2.5.7.
+* TINKERPOP-2243 Added `userAgent` to RequestOptions. Gremlin Console sends `Gremlin Console` as the `userAgent`.
```
spmallette: nit: No need to reference the JIRA here. that can be removed. we capture all the JIRAs closed on this version on release.

gremlin-console/src/main/java/org/apache/tinkerpop/gremlin/console/jsr223/DriverRemoteAcceptor.java
```diff
@@ -54,6 +54,7 @@
  */
 public class DriverRemoteAcceptor implements RemoteAcceptor {
     public static final int NO_TIMEOUT = 0;
+    public static final String USER_AGENT = "Gremlin Console";
```
spmallette: I don't want to get too fancy, but perhaps it would be better to more closely follow the User-Agent standard from [rfc2616](https://www.ietf.org/rfc/rfc2616.txt) which calls for:

```text
"User-Agent" ":" 1*( product | comment )
```

where "product" is just:

```text
product         = token ["/" product-version]
product-version = token
```

For our purposes, I think it would be nice if we did:

```text
Gremlin Console/<version>
```

You can get the version from: 

https://github.com/apache/tinkerpop/blob/2ba15943ff276167432a65ca5a6fc043c873492d/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/Gremlin.java#L38

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
```diff
@@ -1596,6 +1601,18 @@ public void shouldCloseAllClientsOnCloseOfCluster() throws Exception {
         cluster.close();
     }
 
+    @Test
+    public void shouldSendUserAgent() throws Exception {
```
divijvaidya: Can you please add a unit test for the client pattern, `g.with("userAgent", "test").V()` as well?

BrynCooke: `userAgent` is not overridable on the traversal level, it's purely a request based thing.
It's not accessible from scripts, and is only there for graph providers who have customised their request handling to pick up this extra info.

BrynCooke: Hmm, I see you mean bytecode requests. I'll update the code.

BrynCooke: Added and discovered an issue in DriverRemoteConnection.java in the process.

spmallette:
VOTE +1

dkuppitz:
VOTE +1

jorgebay:
VOTE +1

PR #1146, opened 2019-06-18
TINKERPOP-2248 Force replacement of connections on certain errors

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2248

Relying only on the `Connection.isDead()` check to determine connection replacement seemed to introduce a regression as of 2cd84ff1f3944d67d44cbba5bb032a1c57377975 The `shouldBlockRequestWhenTooBig` was failing somewhat randomly since that change. Not sure if there is a better way to make `isDead()` be the only check needed for connection replacement in this context, but this change brings the logic back that checks exception types that were forcing replacement prior to this change. 

@divijvaidya would you mind having a look at this change? perhaps you are aware of better solutions or perhaps I misinterpreted something in my analysis that you could correct?

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Improved exception and messaging for gt/gte/lt/lte when one of the object isn't a `Comparable`.
 * Added test infrastructure to check for storage iterator leak.
 * Fixed multiple iterator leaks in query processor.
+* Forced replacement of connections in Java driver for certain exception types that seem to ultimately kill the connection.
 * Changed the `reverse()` of `desc` and `asc` on `Order` to not use the deprecated `decr` and `incr`.
 * Fixed bug in `MatchStep` where the correct was not properly determined.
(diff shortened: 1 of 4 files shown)


divijvaidya:
+1 The change looks good and should fix this issue.

This is my hypothesis for what is happening:
On CorruptedFrameException, Netty sends a CloseFrame to gracefully close the WebSocket connection asynchronously and throws the exception [1]. The exception is handled by the handler and when we check for Connection.isDead(), it is actually alive at that time waiting for the response to the CloseWebSocket frame. This same connection gets allocated to the next request and at some time it receives the response to the CloseWS frame and the handler attached to the response closes the channel, in turn, closing the second request.

In the new client we plan to write for 3.5, we should handle this case in a better manner. Please note that Netty's behaviour wrt to CorruptedFrameException has changed as of 4.1.37. We need to factor that change as well in future versions [2]

[1] https://github.com/netty/netty/blob/netty-4.0.56.Final/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocket08FrameDecoder.java#L446
[2] https://github.com/netty/netty/commit/7fc718db3cd52e82901dfc3fe8b437ae1ff75d4e

PR #1042, opened 2019-01-18
TINKERPOP-2133 Implement index lookup for Text predicates

skorikov:
Use index lookup for Text predicates (startingWith,endingWith,containing) in Neo4jGraphStep to improve evaluation performance.

See: https://issues.apache.org/jira/browse/TINKERPOP-2133

PR #1149, opened 2019-06-26
Upgrade to neo4j-tinkerpop-api-impl `0.9-3.4.0`

dwitry:
* I think I've got dependency conflicts figured out. At least basic queries are executed ok and `mvn clean install -DincludeNeo4j` passes. Please clarify if I'm missing something.
* Upgrade to Neo4j  `3.4` (not `3.5`), because this is the latest available version in  `neo4j-tinkerpop-api-impl`. Migration to `3.5` will require some efforts.
* Upgrade Neo4j `3.2.3` → `3.4.11` should be straightforward. Added release notes anyway.
* Looks like #1042 could be merged after this

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,7 @@ This release also includes changes from <<release-3-4-3, 3.4.3>>.
 * Removed previously deprecated `TraversalSource.withRemote()`.
 * Removed previously deprecated SSL settings: `keyCertChainFile`, `keyFile`, `keyPassword` and `trustCertChainFile` and related infrastructure.
 * Removed previously deprecated `BulkDumperVertexProgram` and `BulkLoaderVertexProgram`.
+* Upgrade to Neo4j 3.4.11
 
 == TinkerPop 3.4.0 (Avant-Gremlin Construction #3 for Theremin and Flowers)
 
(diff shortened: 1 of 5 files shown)


spmallette:
Nice pom work and great to see upgrade docs. 

> Migration to 3.5 will require some efforts.

good to know - thanks. 

> Looks like #1042 could be merged after this

I will rebase that PR and see what happens. It still needs some work to get it to the finish line before merge but at least it might build properly as a feature branch.

Thanks for this contribution - I think neo4j users will be excited to see this.

VOTE +1

spmallette:
@dwitry i merged this to `master` a while back, but now i'm seeing some problems. when i run tests i get:

```
java.lang.RuntimeException: GraphFactory could not instantiate this Graph implementation [class org.apache.tinkerpop.gremlin.neo4j.structure.Neo4jGraph]
Caused by: java.lang.reflect.InvocationTargetException
Caused by: java.lang.NoClassDefFoundError: org/neo4j/graphdb/factory/GraphDatabaseFactory
Caused by: java.lang.ClassNotFoundException: org.neo4j.graphdb.factory.GraphDatabaseFactory
```

and it doesn't occur before this change.  Note that if you try to build this on `master` I think there's some new dependency conflicts (which is what I was trying to fix when I uncovered this) based on https://github.com/apache/tinkerpop/commit/607224a92e443f712ca86b4b38e2fe9f92e405dc but I tried changes before that one (including the very commit i merged from this PR) and I still get the problem. If this issue doesn't get resolved I will need to revert the change.

Any idea what might be wrong there?

spmallette:
wow. it took me forever to realize that i had a corrupted neo4j-kernel.jar in my .m2  directory. the jar file was present, but it was completely empty (0k in size). I have no idea what happened there. Anyway, it appears to be building now.

sorry for the false alarm @dwitry

PR #1157, opened 2019-07-08
Add ParameterizedGroovyTranslator.java / ParameterizedGroovyTranslato…

sandszhouSZ:
Add: 
  ParameterizedGroovyTranslator.java and ParameterizedGroovyTranslatorTest.java
 
Current discussion here: 
   https://issues.apache.org/jira/browse/TINKERPOP-2252?filter=-2
  Except  ParameterizedGroovyTranslatorTest i also do some Real-time interactive testing 

Thank you for your patience. If somebody have better suggestions, I will continue to optimize.

### DIFF
--- a/gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/ParameterizedGroovyTranslator.java
+++ b/gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/ParameterizedGroovyTranslator.java
@@ -1,0 +1,340 @@
+
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package org.apache.tinkerpop.gremlin.groovy.jsr223;
+
+import org.apache.commons.configuration.ConfigurationConverter;
+import org.apache.tinkerpop.gremlin.process.traversal.*;
+import org.apache.tinkerpop.gremlin.process.traversal.Translator.ScriptTranslator;
+import org.apache.tinkerpop.gremlin.process.traversal.step.TraversalOptionParent;
+import org.apache.tinkerpop.gremlin.process.traversal.strategy.TraversalStrategyProxy;
+import org.apache.tinkerpop.gremlin.process.traversal.util.ConnectiveP;
+import org.apache.tinkerpop.gremlin.process.traversal.util.OrP;
+import org.apache.tinkerpop.gremlin.structure.Edge;
+import org.apache.tinkerpop.gremlin.structure.Element;
+import org.apache.tinkerpop.gremlin.structure.Vertex;
+import org.apache.tinkerpop.gremlin.structure.VertexProperty;
+import org.apache.tinkerpop.gremlin.structure.util.StringFactory;
+import org.apache.tinkerpop.gremlin.util.function.Lambda;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * Converts bytecode to a Parameterized Groovy string of Gremlin.
+ *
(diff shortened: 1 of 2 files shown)


gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/ParameterizedGroovyTranslator.java
```diff
@@ -0,0 +1,340 @@
...
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package org.apache.tinkerpop.gremlin.groovy.jsr223;
+
+import org.apache.commons.configuration.ConfigurationConverter;
+import org.apache.tinkerpop.gremlin.process.traversal.*;
```
spmallette: Please note that we don't have a particular code style in TinkerPop but we do try to not use wildcard imports and we try to `final` variables where it makes sense to do so.

gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/ParameterizedGroovyTranslator.java
```diff
@@ -0,0 +1,340 @@
...
+
+/**
+ * Converts bytecode to a Parameterized Groovy string of Gremlin.
+ *
+ * @author Marko A. Rodriguez (http://markorodriguez.com)
+ * @author Stephen Mallette (http://stephen.genoprime.com)
+ * @author Stark Arya (sandszhou.zj@alibaba-inc.com)
+ */
+
+public class ParameterizedGroovyTranslator implements ScriptTranslator {
```
spmallette: some class level javadoc here would be good. Please provide a reference to `GroovyTranslator` and explain the difference between this one and that one.  A javadoc link from `GroovyTranslator` to this class would be nice too.

gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/ParameterizedGroovyTranslator.java
```diff
@@ -0,0 +1,340 @@
...
+        return of(traversalSource,null);
+    }
+
+    public static final ParameterizedGroovyTranslator of(final String traversalSource, final TypeTranslator typeTranslator) {
+        return new ParameterizedGroovyTranslator(traversalSource,
+                Optional.ofNullable(typeTranslator).orElseGet(DefaultParameterizedTypeTranslator::new));
+    }
+
+    //////
+    public ParameterizedGroovyResult parameterizedTranslate(final Bytecode bytecode) {
```
spmallette: I think that you need to implement this slightly differently. I don't think you want a new `parameterizedTranslate(Bytecode)` method. I think you just want:

```java
public ParameterizedGroovyResult translate(Bytecode)
```

You should just consider a new `Translator` implementation like `ScriptTranslator`:

https://github.com/apache/tinkerpop/blob/199c9106a76b09652b8f4d9cb76cbed7ac7fd418/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Translator.java#L63

that is:

```java
public interface ScriptTranslator extends Translator<String, ParameterizedGroovyResult> 
```

Personally, I think I'd prefer that `ParameterizedGroovyResult` to be named `ParameterizedScript` (and moved `gremlin-core`) so that it could be more generally used in the future if we liked (`gremlin-python` translation for example) thus:

```java
public interface ScriptTranslator extends Translator<String, ParameterizedScript> 
```

gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/ParameterizedGroovyTranslator.java
```diff
@@ -0,0 +1,340 @@
...
+
+        public Map<String,Object> getParameters() {
+            return parameters;
+        }
+    }
+
+    /**
+     * Performs standard type translation for the TinkerPop types to Parameterized Groovy.
+     */
+    public static class DefaultParameterizedTypeTranslator implements TypeTranslator {
```
spmallette: It would be nice if this could all be implemented as a simple `TypeTranslator` because then we wouldn't need this special  `Translator` implementation. Still thinking about how this might all "just work" so I could perhaps suggest more changes after these, but for now, I don't think that the constructor for `ParameterizedGroovyTranslator` should have a construction that take a simple `TypeTranslator`:

https://github.com/apache/tinkerpop/pull/1157/files#diff-1ca5e24d7542dff7f3f8fbbca51c91faR61

because for it to extract parameters it needs a special kind of `TypeTranslator` (i.e. one that captures parameters like this one). So, you might need a new `ParameterizedTypeTranslator` interface to accomplish that. I wish you didn't though. Overall this needs a bit more thought to get t right.

spmallette:
This was a pretty good start. Don't worry about the failing travis job for now - looks like Travis instability problems which we sometimes see. I offered a few immediate thoughts on your work and I think I will have more as I think it through further. Please let me know if you have questions/comments so far.

PR #1158, opened 2019-07-09
Tinkerpop 2252 Support Parameterized Groovy Script Transformation

sandszhouSZ:


sandszhouSZ:
I am very very sorry for inadvertently close the PR。
and the previous pr and suggestion here: https://github.com/apache/tinkerpop/pull/1157
issue could be found here：https://issues.apache.org/jira/browse/TINKERPOP-2252?filter=-2

some solutions here： 
1. Please note that we don't have a particular code style in TinkerPop but we do try to not use wildcard imports and we try to final variables where it makes sense to do so.
**_It is Intellij's default config and i have already fix it_**

2. some class level javadoc here would be good. Please provide a reference to GroovyTranslator and explain the difference between this one and that one. A javadoc link from GroovyTranslator to this class would be nice too.
**_I have add some javadoc before ParameterizedGroovyTranslator and explain the relation with GroovyTranslator, and i think  link from GroovyTranslator is unnecessary and it could make this relations looks  Complex_**

3. I think that you need to implement this slightly differently. I don't think you want a new parameterizedTranslate(Bytecode) method. I think you just want:
public ParameterizedGroovyResult translate(Bytecode)
**_Yes，I'm worried about changing to other files at the beginning. I have modified on your advice。_**

4. t would be nice if this could all be implemented as a simple TypeTranslator because then we wouldn't need this special Translator implementation. Still thinking about how this might all "just work" so I could perhaps suggest more changes after these, but for now, I don't think that the constructor for ParameterizedGroovyTranslator should have a construction that take a simple TypeTranslator:because for it to extract parameters it needs a special kind of TypeTranslator (i.e. one that captures parameters like this one). So, you might need a new ParameterizedTypeTranslator interface to accomplish that. I wish you didn't though. Overall this needs a bit more thought to get t right.

gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/ParameterizedGroovyTranslator.java
```diff
@@ -0,0 +1,346 @@
...
+
+    @Override
+    public String getTraversalSource() {
+        return this.traversalSource;
+    }
+
+    /**
+     * Performs standard type translation for the TinkerPop types to Parameterized Groovy.
+     */
+    public static class DefaultTypeTranslator implements TypeTranslator {
```
spmallette: It bothers me a bit that the `TypeTranslator` is being used for more than just converting types which was its original intent so that providers/users could plugin their own conversions for custom types. If we implement things the way they are now, they not only have to write two separate `TypeTranslators` (if they want to support both forms of translation) but they also have to reason about `Parameters`. I'm not sure what the solution is there, but I think that's the next thing that's concerning me about this change that I don't have a good answer for yet.

sandszhouSZ: In most cases, people just need call the following function:
        public static final ParameterizedGroovyTranslator of(final String traversalSource) 
when they need a default parameterized script translation，this cover most cases and easy enough for use.

As you have mentioned, when people want their own ParameterizedScriptTranslator.TypeTranslator Instead of DefaultTypeTranslator, they should known about Parameters，it may exceed TypeTranslator’s original intent boundary。

I think，if we  take paramterized script  as a part script and if people has some demand on paramterized script, also they need their personal paramterized script. it's necessary for them to known about Parameters and  if you worry about this  i think a better way seems like： ParameterizedScript class， we could move：
        public static final class Parameters {
to a separate file like ParameterizedBindings.java for user to directly use.

sandszhouSZ: python test fail, and i thinks it's  travis-ci problem

sandszhouSZ:
if you have new ideas, please let me known.
Thanks!
 
---Original---

spmallette:
I've had some more time to think about this. I'm having trouble envisioning the complete impact to this change so please forgive me for suggesting changes that might seem like I'm taking this design in circles.

That said, I think that everything works more nicely if the `ScriptTranslator` interface changed and simply became:

```java
public interface ScriptTranslator extends Translator<Script, String>
```

where `Script` is what you currently have named `ParameterizedScript `. Going forward, `Script` can have parameters or not so:

```java
public Optional<Map<String,Object>> getParameters()
```

Now, `ParameterizedGroovyTranslator` can go away, and we can just use `GroovyTranslator` which will return a `Script` with or without parameters based on the `TypeTranslator` given to it. I think `DefaultTypeTranslator` should simply be configurable by way of the constructor to extract parameters or not. 

I think that these changes should allow you to get rid of the `ThreadLocal` in `Parameters` hopefully (another thing that was bothering me). It allows existing extensions of `DefaultTypeTranslator` to take advantage of `Parameters` in a way that doesn't make them have to reason about that (i.e. they can just add their custom types for conversion and choose whether they opt to parameterize them or not). I think it also unifies the API more nicely for downstream usage. All of this comes at the cost of a breaking change to our APIs, but I think that's acceptable.

I'm not sure if that's all the changes that need to be done for this to be "ready", but I think I'd need to see them in place before I could say more. At least I'm seeing the bigger picture of how this fits more clearly now. Please let me know what you think.

Also, a couple of administrative notes:

1. Please prefer `git rebase` when bringing in new changes from master to your branch (i see a merge commit on this PR - please remove that when you get a chance - thanks)
2. Don't worry about travis too much - I tend to keep an eye on it and will let you know if there is a test that won't pass because of a change you made. Please consider using docker to validate your builds in addition to local maven builds as docker will run all tests across all Gremlin Language Variants - `docker/build.sh -t -i` will give you extraordinary confidence that your changes are "good".

Thanks for your patience with this pull request.

sandszhouSZ:
ScriptTranslator changes：
public interface ScriptTranslator extends Translator<String,Script> ?

Coupling ParameterizedGroovyTranslator and  GroovyTranslator  together will make the code concise exactly，but  GroovyTranslator.translate will return ParameterizedScript Instead of String ,I'm worried this incompatibility will bother users。I will be glad if you think  that's acceptable.

If you are bothered by ThreadLocal in Parameters，i  think all this could be done in Script。

spmallette:
> ,I'm worried this incompatibility will bother users。I will be glad if you think that's acceptable.

We're making the change in 3.5.0 so breaking changes are expected. It's not too large of a change and the default functionality should be the existing functionality which means that the `Script` will simply contain no parameters. So all someone would have to do to upgrade is change their code slightly to get the script string from the `Script` object....not too bad. 

> If you are bothered by ThreadLocal in Parameters，i think all this could be done in Script。

If we can do without `ThreadLocal` I think that would be really good.

> i think ParameterizedGroovyTranslatorTest should reserved to test GroovyTranslator with parameter

ok. with respect to testing, there will be other tests i will likely ask you to include, but for now that's a good start. thanks

spmallette:
Since it's been a few months, I was just wondering if you intended to come back to this PR any time soon?

sandszhouSZ:
Yes,  i will continue this PR  later  this week,  Thanks for your remind

sandszhouSZ:
Hi, spmallette. 
I merge ParameterizedScript into GroovyTranslator，use Script to save script and it's bindings.  

Parameters switch  is off as default to maintain compatibility,  but user could use self provided TypeTranslator or open parameters manually。

ParameterizedScript test in ParameterizedGroovyTranslatorTest.java, test including:
   - parameterized script checking
   - binding checking
   - eval result checking compare 
In addition to basic test based on GroovyTranslatorTest.java, I also add some test cases like:  shouldHandleArray,  shouldHandleSet, ....  to ensure that all object type in GroovyTranslator will be covered with testing。

spmallette:
Merged this with a few minor adjustments 9b84456b4cd80143a564699a277860b1618026ce 

Performance of lambda based bytecode translations is pretty impressive, though we don't really recommend using lambdas so it likely won't see much support there. Direct translation use cases (which i think was your target in submitting this PR) should benefit from this i suspect, though we typically wouldn't recommend that people take that route for sake of transactions (we've already discussed all that in JIRA - i'm just summarizing).

Thanks for sticking with this PR and getting it to the end.

PR #1160, opened 2019-07-10
TINKERPOP-2259 Default java server/driver serialization to GraphBinary

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2259

Removed Gryo configurations and updated documentation as needed. Gryo classes are all still present. Users can go back to them if they like. This change sets up TINKERPOP-2233.

VOTE +1


jorgebay:
Wow, sorry I missed this patch to review, thanks for taking care of this!

PR #1162, opened 2019-07-15
TINKERPOP-2265 Deprecate remote Traversal.sideEffects

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2265

Users should not prefer use of `cap()` to directly return side-effects as part of standard traversal iteration. Expecting removal of the deprecated bits in 3.5.0.

VOTE +1


spmallette:
@divijvaidya thanks - i shouldn't have missed all those. Should look better now.

divijvaidya:
I am familiar with the server and the java driver code changes and they look good.

VOTE +1

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/ITraversalSideEffects.cs
```diff
@@ -29,6 +29,7 @@ namespace Gremlin.Net.Process.Traversal
     /// <summary>
     ///     A <see cref="ITraversal" /> can maintain global sideEffects.
     /// </summary>
+    [Obsolete("As of release 3.3.8, not replaced, prefer use of cap()-step to retrieve side-effects as part of traversal iteration")]
```
jorgebay: Is the intention to deprecate `ITraversalSideEffects`?

spmallette: hmm - i think so. it's going to end up being a do-nothing interface ultimately. If we removed `ITraversal.SideEffects` then it wouldn't serve much purpose. We don't do any local side-effects, so without remote ones it doesn't seem useful. Unless we just keep it around for API completeness to match Java?

jorgebay: ahh, I see what you mean, makes sense to deprecate the entire interface from the GLV.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Tokens.java
```diff
@@ -30,9 +30,21 @@ private Tokens() {}
     public static final String OPS_BYTECODE = "bytecode";
     public static final String OPS_EVAL = "eval";
     public static final String OPS_INVALID = "invalid";
+    public static final String OPS_CLOSE = "close";
+
+    /**
+     * @deprecated As of release 3.3.8, not directly replaced in the protocol as side-effect retrieval after
```
FlorianHockmann: (nitpick) I think these tokens should also be deprecated the same way in Gremlin.Net's [`Token` class](https://github.com/apache/tinkerpop/blob/master/gremlin-dotnet/src/Gremlin.Net/Driver/Tokens.cs) (and I don't know about gremlin-python and gremlin-javascript).

spmallette: ok - added the obsolete annotation for .NET. javascript doesn't have side-effect functionality so nothing to deprecate there. i thought about python but it meant adding the `deprecations` library (i think) and i didnt' want to add a dependency (maybe we should...i dunno - perhaps that's a separate decision for newer versions of gremlin-python). how does everything look now?

FlorianHockmann: Looks good to me in general, but it looks like you need to add a `using` directive in `Tokens.cs` which lets the build fail right now.

PR #1163, opened 2019-07-15
TINKERPOP-1921 Added hasNext() to javascript and .NET

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1921

Python was done a long time ago, but I never returned to deal with Javascript and .NET  - so, that's done now. Gremlin Language Variants gain a bit more consistency as a result of this change. Happy to hear other takes on how to do this nicer (i don't particularly like what I did with .NET). Perhaps I'll see it differently tomorrow and make some adjustments.

VOTE +1


FlorianHockmann:
I don't have a strong opinion on this. Wrapping `IEnumerator<Traverser>` in a dedicated class that already implements `HasNext()` sounds like a good idea as it would be a better place for this logic, but we can also keep it as is in this PR if it requires too much effort to make this change.

spmallette:
What I have works which is good - it just didn't seem pretty but I'm fine to leave it like this as long as no one is feeling strongly against it.

PR #1165, opened 2019-07-23
TINKERPOP-2264 Fixed g:Date serialization for python.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2264

Got timezone of the local system out of the derser mix and standardized around what's always been expected in Java's `Date.getTime()`.

Builds with `mvn clean install :pl gremlin-python`

VOTE +1

### DIFF
--- a/gremlin-python/src/main/jython/gremlin_python/structure/io/graphsonV2d0.py
+++ b/gremlin-python/src/main/jython/gremlin_python/structure/io/graphsonV2d0.py
@@ -16,6 +16,8 @@ KIND, either express or implied.  See the License for the
 specific language governing permissions and limitations
 under the License.
 '''
+
+import calendar
 import datetime
 import json
 import time
@@ -339,20 +341,16 @@ class DateIO(_GraphSONTypeIO):
     python_type = datetime.datetime
     graphson_type = "g:Date"
     graphson_base_type = "Date"
-    epoch = datetime.datetime(1970, 1, 1)
 
     @classmethod
     def dictify(cls, obj, writer):
-        # Java timestamp expects milliseconds.
-        if six.PY3:
-            pts = (obj - cls.epoch) / datetime.timedelta(seconds=1)
-        else:
-            # Hack for legacy Python - timestamp() in Python 3.3
-            pts = (time.mktime(obj.timetuple()) + obj.microsecond / 1e6) - \
-                  (time.mktime(cls.epoch.timetuple()))
+        try:
+            timestamp_seconds = calendar.timegm(obj.utctimetuple())
+            pts = timestamp_seconds * 1e3 + getattr(obj, 'microsecond', 0) / 1e3
+        except AttributeError:
+            pts = calendar.timegm(obj.timetuple()) * 1e3
 
-        # Have to use int because of legacy Python
-        ts = int(round(pts * 1000))
+        ts = int(round(pts))
         return GraphSONUtil.typedValue(cls.graphson_base_type, ts)
 
     @classmethod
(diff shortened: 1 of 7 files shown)


gremlin-python/src/main/jython/gremlin_python/structure/io/graphsonV2d0.py
```diff
@@ -339,18 +339,17 @@ class DateIO(_GraphSONTypeIO):
...
+            pts = (obj - cls.epoch) / datetime.timedelta(seconds=1)
         else:
-            # Hack for legacy Python
-            # timestamp() in Python 3.3
-            pts = time.mktime((obj.year, obj.month, obj.day,
-			                   obj.hour, obj.minute, obj.second,
-			                   -1, -1, -1)) + obj.microsecond / 1e6
+            # Hack for legacy Python - timestamp() in Python 3.3
+            pts = (time.mktime(obj.timetuple()) + obj.microsecond / 1e6) - \
+                  (time.mktime(cls.epoch.timetuple()))
```
TheRealFalcon: If I'm understanding this right, it's taking the epoch offset and removing it from our current time.
So for example, `obj.timetuple()` will return current UTC-5 for me, and `cls.epoch.timetuple()` should return epoch-5, so we if we subtract them, we get the UTC time.

But according to the end of https://stackoverflow.com/a/5499906/37205, the epoch offset can be different for DST timezones. I don't really understand how that can be, but when I'm not in daylight time, I'm UTC-6. I think this answer is saying that if my timezone was affected, it would return epoch as epoch-6 when current time would be UTC-5.

gremlin-python/src/main/jython/gremlin_python/structure/io/graphsonV2d0.py
```diff
@@ -339,18 +339,17 @@ class DateIO(_GraphSONTypeIO):
...
     graphson_base_type = "Date"
+    epoch = datetime.datetime(1970, 1, 1)
 
     @classmethod
     def dictify(cls, obj, writer):
-        # Java timestamp expects miliseconds
+        # Java timestamp expects milliseconds.
         if six.PY3:
-            pts = obj.timestamp()
+            pts = (obj - cls.epoch) / datetime.timedelta(seconds=1)
```
aholmberg: To me this is a bit of a confusing way to get to a timestamp. I might suggest having a look at datetime handling in the Cassandra driver: https://github.com/datastax/python-driver/blob/master/cassandra/cqltypes.py#L577-L579
This technique has a few advantages:
- No special cases per python runtime
- Handles datetime inputs with timezones attached, normalizing back to UTC
- More readable (imo)

gremlin-python/src/main/jython/gremlin_python/structure/io/graphsonV2d0.py
aholmberg: I know this was not introduced in this change, but a "hack" should not be needed if an alternative conversion technique is used (see above).

gremlin-python/src/main/jython/gremlin_python/structure/io/graphsonV3d0.py
```diff
@@ -348,27 +348,26 @@ class DateIO(_GraphSONTypeIO):
...
     graphson_type = "g:Date"
     graphson_base_type = "Date"
+    epoch = datetime.datetime(1970, 1, 1)
 
     @classmethod
     def dictify(cls, obj, writer):
+        # Java timestamp expects milliseconds.
         if six.PY3:
-            pts = obj.timestamp()
+            pts = (obj - cls.epoch) / datetime.timedelta(seconds=1)
```
aholmberg: Same comments as V2d0 file.

spmallette:
thanks @aholmberg  and @TheRealFalcon - i got rid of hacks (and the "old" way in general) in exchange for `calendar` . I didn't use the full DS driver code in terms of error handling because this code should only get triggered for actual `datetime` sorts of objects. if the timestamp was a numeric i think it believe it would have ended up being handled by that serializer.

robertdale:
VOTE +1

dkuppitz:
Not an expert in Python, but the code appears to be good.

VOTE +1

PR #1168, opened 2019-07-25
TINKERPOP-2246 Use ResponseHandlerContext to ensure single final response to client

spmallette:
Merged from #1148 with deprecation included. 

@divijvaidya if this looks good to you, then i will move forward with getting it merged.


gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractEvalOpProcessor.java
```diff
@@ -191,45 +193,21 @@ protected AbstractEvalOpProcessor(final boolean manageTransactions) {
...
-        }
-    }
-
-    /**
-     * A variant of {@link #evalOpInternal(Context, Supplier, BindingSupplier)} that is suitable for use in situations
-     * when multiple threads may produce {@link ResponseStatusCode#isFinalResponse() final} response messages
-     * concurrently.
-     * @see #evalOpInternal(Context, Supplier, BindingSupplier)
-     */
-    protected void evalOpInternal(final ResponseHandlerContext rhc, final Supplier<GremlinExecutor> gremlinExecutorSupplier,
```
dimas-b: We are keeping `ResponseHandlerContext` as a deprecated class, but we are removing this `evalOpInternal` overload? If we intend to keep `ResponseHandlerContext` for backward compatibility would we also need to keep the methods that use it as parameters?

Sorry, I did not review very closely, perhaps I missed something in this regard.

spmallette: ....what's going on here?!  i fixed all that! that was the whole pont of my change. seems like my PR is messed up somehow. looking

spmallette: ok - i don't know what happened, but somehow that change got reverted somewhere along the line. thanks for noticing that @dimas-b -

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/ResponseHandlerContextTest.java
```diff
@@ -57,32 +56,32 @@
     public static Iterable<Object[]> data() {
         return Arrays.asList(new Object[][] {
                 {
-                    new BiFunction<ResponseHandlerContext, ResponseStatusCode, Void>() {
-                        @Override
-                        public Void apply(final ResponseHandlerContext context, final ResponseStatusCode code) {
-                            context.writeAndFlush(code, "testMessage");
-                            return null;
+                        new BiFunction<Context, ResponseStatusCode, Void>() {
```
divijvaidya: nit

seems like alignment is inconsistent with the next BiFunction block

divijvaidya:
Looks good. 

Vote +1: I understand the motivation of this change and am familiar with the logic of the change.

PR #1169, opened 2019-07-25
TINKERPOP-2272 Renamed steps/tokens conflicting with global function names

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2272

Used the standard underscore suffix on these problems. Didn't document a bunch purposely because the intention is to retain the old conflicting steps/tokens for the time being (both are currently present in the library so as to be backward compatible in 3.3.x and 3.4.x). In 3.5.0 all tests will be modified, documentation updated and the old conflicting names completely removed. So, for now at least the new methods are present for use and we don't have a breaking change to be concerned about.

VOTE +1


spmallette:
Note that there is a committer +1 on the JIRA issue.

robertdale:
Is there an integration test for this?

robertdale:
There were some 'remote' tests with `id`, `min`, `filter` that I changed to the underscore version and ran `mvn clean  install -DskipIntegrationTests=false` which passed.

VOTE +1

dkuppitz:
VOTE +1

spmallette:
@robertdale added a test to exercise those new underscored versions

PR #1172, opened 2019-07-30
Handle closed connections gracefully instead of bubbling up exceptions

WiredUK:
A lot of `NullReferenceException`s were being thrown and internally handled which was causing some performance issues in our codebase. This PR handles attempts to receive data when the connection has been closed.

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
@@ -103,8 +103,17 @@ namespace Gremlin.Net.Driver
             {
                 try
                 {
-                    var received = await _webSocketConnection.ReceiveMessageAsync().ConfigureAwait(false);
-                    Parse(received);
+                    var receivedMessage = await _webSocketConnection.ReceiveMessageAsync().ConfigureAwait(false);
+
+                    if (!receivedMessage.ConnectionClosed)
+                    {
+                        Parse(receivedMessage.Data);
+                    }
+                    else
+                    {
+                        await CloseAsync().ConfigureAwait(false);
+                        return;
+                    }
                 }
                 catch (Exception e)
                 {
@@ -116,6 +125,11 @@ namespace Gremlin.Net.Driver
 
         private void Parse(byte[] received)
         {
+            if (received == null || received.Length == 0)
+            {
+                return;
+            }
+
             var receivedMsg = _messageSerializer.DeserializeMessage<ResponseMessage<JToken>>(received);
 
             try
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/RecievedResult.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/RecievedResult.cs
@@ -1,4 +1,27 @@
-﻿namespace Gremlin.Net.Driver
+﻿#region License
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#endregion
+
+namespace Gremlin.Net.Driver
 {
     internal sealed class RecievedMessageResult
     {
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/RecievedResult.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/RecievedResult.cs
@@ -1,0 +1,19 @@
+﻿namespace Gremlin.Net.Driver
+{
+    internal sealed class RecievedMessageResult
+    {
+        private RecievedMessageResult(byte[] data, bool connectionClosed)
+        {
+            Data = data;
+            ConnectionClosed = connectionClosed;
+        }
+
+        public static RecievedMessageResult Success(byte[] data) => new RecievedMessageResult(data, false);
+        public static RecievedMessageResult Failed() => new RecievedMessageResult(null, true);
+
+        public byte[] Data { get; }
+        public bool ConnectionClosed { get; }
+    }
+
+
+}
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/WebSocketConnection.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/WebSocketConnection.cs
@@ -73,8 +73,13 @@ namespace Gremlin.Net.Driver
                     .ConfigureAwait(false);
         }
 
-        public async Task<byte[]> ReceiveMessageAsync()
+        public async Task<RecievedMessageResult> ReceiveMessageAsync()
         {
+            if (_client.State != WebSocketState.Open)
+            {
+                return RecievedMessageResult.Failed();
+            }
+
             using (var ms = new MemoryStream())
             {
                 WebSocketReceiveResult received;
@@ -85,7 +90,7 @@ namespace Gremlin.Net.Driver
                     ms.Write(receiveBuffer.Array, receiveBuffer.Offset, received.Count);
                 } while (!received.EndOfMessage);
 
-                return ms.ToArray();
+                return RecievedMessageResult.Success(ms.ToArray());
             }
         }
 


spmallette:
thanks for taking the time to contribute to the project. could you say some more about the performance issues - is there a noticeable improvement as a result of this change?

WiredUK:
> thanks for taking the time to contribute to the project. could you say some more about the performance issues - is there a noticeable improvement as a result of this change?

Sure Stephen, I don't have a huge amount of data, but I had a process that was taking around ~9-10s to complete and this change reduced it to ~6s. The process I was running ran approximately 12 Gremlin queries and each one threw at least 2 internal `NullReferenceException`'s, usually many more. My debug output looks something like this:

<details>
  <summary>Log output</summary>
  <p>

```
Executing Gremlin: g.V().<Redacted>
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Executing Gremlin: g.V().<Redacted>
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Executing Gremlin: g.V().<Redacted>
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Executing Gremlin: g.addV(label).<Redacted>
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Executing Gremlin: g.V().<Redacted>
Executing Gremlin: g.V().<Redacted>
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Executing Gremlin: g.V().<Redacted>
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Executing Gremlin: g.addV(label).<Redacted>
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Executing Gremlin: g.addV(label).<Redacted>
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Executing Gremlin: g.addV(label).<Redacted>
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Executing Gremlin: g.V().<Redacted>
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Executing Gremlin: g.addV(label).<Redacted>
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
Exception thrown: 'System.NullReferenceException' in Gremlin.Net.dll
```
</p>
</details>

spmallette:
Thanks for the additional information. Note that travis is failing because of Rat plugin checks - I think you just need a license header on this new file:

https://github.com/apache/tinkerpop/pull/1172/files#diff-7b8a375e9762f3f6cba2d1b3c5087d5a

WiredUK:
I've added the licence info and I think the Travis has worked now, or it has kinda worked, I'm not 100% sure!

FlorianHockmann:
Hi @WiredUK and thanks for contributing to TinkerPop!

Is the `NullReferenceException` you're getting the same as described in [TINKERPOP-2192](https://issues.apache.org/jira/browse/TINKERPOP-2192)?
If not, can you please provide a stack trace?
I would like to understand the problem better before I can really review this PR.

WiredUK:
> Is the `NullReferenceException` you're getting the same as described in [TINKERPOP-2192](https://issues.apache.org/jira/browse/TINKERPOP-2192)?

Hmm, it could be, it's definitely happening around the same part of the code. But if it is, your deduction about the cause is slightly off I think.

> If not, can you please provide a stack trace?

Sure, stack trace is:

```
at Gremlin.Net.Driver.Connection.TryParseResponseMessage(ResponseMessage`1 receivedMsg) in C:\path\Gremlin.Net\Driver\Connection.cs:line 136
at Gremlin.Net.Driver.Connection.Parse(Byte[] received) in C:\path\Gremlin.Net\Driver\Connection.cs:line 123
```

The error happens because `Parse` is receiving an empty byte array, which in turn makes `_messageSerializer.DeserializeMessage<ResponseMessage<JToken>>(received)` return `null`.

FlorianHockmann:
> But if it is, your deduction about the cause is slightly off I think.

That's definitely possible. I didn't have the line numbers for the stack trace which probably resulted in me looking in the wrong place.

> The error happens because Parse is receiving an empty byte array, which in turn makes _messageSerializer.DeserializeMessage<ResponseMessage<JToken>>(received) return null.

Isn't it then enough to just check whether we received an empty array from `_webSocketConnection.ReceiveMessageAsync()` and then skip the `Parse` method?
Then we wouldn't need an additional class just transmit information about whether we actually received anything or not because the connection was already closed.
We already [check regularly whether a connection is already closed](https://github.com/apache/tinkerpop/blob/54873cb6de65f0f610fa2a35e970b4c327a0a040/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs#L127) in the `ConnectionPool` so I would prefer if we could not add another check like that here.

WiredUK:
> Isn't it then enough to just check whether we received an empty array from `_webSocketConnection.ReceiveMessageAsync()` and then skip the `Parse` method?
> Then we wouldn't need an additional class just transmit information about whether we actually received anything or not because the connection was already closed.
> We already [check regularly whether a connection is already closed](https://github.com/apache/tinkerpop/blob/54873cb6de65f0f610fa2a35e970b4c327a0a040/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs#L127) in the `ConnectionPool` so I would prefer if we could not add another check like that here.

Maybe, but then the code gets "stuck" inside the `while(true)`, which is a tight loop, waiting for the external connection check, and only exits when an exception happens. There seems to be no way to check that all messages have been received?

FlorianHockmann:
> Maybe, but then the code gets "stuck" inside the while(true), which is a tight loop, waiting for the external connection check, and only exits when an exception happens.

But it will only call `ReceiveMessageAsync` as that shouldn't return any more if the connection is already closed, right?

WiredUK:
> > Maybe, but then the code gets "stuck" inside the while(true), which is a tight loop, waiting for the external connection check, and only exits when an exception happens.
> 
> But it will only call `ReceiveMessageAsync` as that shouldn't return any more if the connection is already closed, right?

If I just add a check around the received message being null, as per [this image](https://user-images.githubusercontent.com/7810709/62474546-06002f00-b79b-11e9-98f7-cf175dd10900.png), the errors switch to being `WebSocketException` because it continues to keep trying to receive more messages.

WiredUK:
Another option would be to return a flag from the `Parse` method which would allow breaking out of `ReceiveMessagesAsync`. If you are confident the connections ae managed elsewhere, that could work. For example https://gist.github.com/WiredUK/7b356eaa15b3c2615a3e45e28f8b950b

FlorianHockmann:
> If I just add a check around the received message being null, as per this image, the errors switch to being WebSocketException because it continues to keep trying to receive more messages.

I didn't expect that to be honest. In that case, I'd actually prefer your approach in this PR to directly close the connection if it's broken indicated by the received empty array.

> If you are confident the connections ae managed elsewhere, that could work. 

I guess it actually makes more sense to close the connection here directly if we know for sure at that point that it isn't usable any more instead of letting it sit in the `ConnectionPool` and let that detect and handle the problem.

Another point that needs to be taken care of are the requests currently awaiting a response over the connection that we close here. We can notify them similar to how the `NotifyAboutConnectionFailure` method currently does it.

@jorgebay If you have some time to spare, it would be good to have your input here, too.

jorgebay:
I agree about closing the connection eagerly and notifying the existing requests that there is a connection error (in the same way as `NotifyAboutConnectionFailure()`).

WiredUK:
The current flow is for it to throw an exception, which bubbles up and causes the connection to be closed via the `CloseConnectionBecauseOfFailureAsync` method. If that exception is no longer being thrown, we cannot call that method any more because it requires the exception object to be passed in (ultimately so it can notify)

I'm now wondering if the whole connection flow needs to be reviewed because this PR would fix the "control flow by exception" (which imo is an anti-pattern) but may have other side effects.

FlorianHockmann:
>  If that exception is no longer being thrown, we cannot call that method any more because it requires the exception object to be passed in (ultimately so it can notify)

I think we can either add a very similar method that doesn't take an exception or to just create an exception object and pass that to the method.

> I'm now wondering if the whole connection flow needs to be reviewed because this PR would fix the "control flow by exception" (which imo is an anti-pattern)

I wouldn't call it control flow by exception right now as the driver simply doesn't catch this exception as we didn't expect it. So, it's really an unexpected exception for which the error handling seems appropriate to me. This PR changes that as it checks for the error and then handles it by notifying all affected request handlers with a meaningful message instead of the `NullReferenceException` (at least per my above suggestion).

WiredUK:
The reason I said it is flow-by-exception is because of the `while(true)` loop. The only way to leave that loop is by exception. It may not have been intended to be used that way, but that is how it is currently written. 

I don't know enough about the notification process to update this PR right now, I'd need to dig into it further to make another update. Happy to take guidance though.

FlorianHockmann:
My idea was simply to do something like this if the connection was closed unexpectedly:

```c#
await CloseConnectionBecauseOfFailureAsync(
    new Exception("The connection was closed while receiving.")).ConfigureAwait(false);
```

Of course a more specific exception type than `Exception` should be used, but I haven't looked yet into which one could be a good candidate here.
This could be done in the `Connection` class. An alternative approach would simply throw a meaningful exception directly in the `WebSocketConnection` class where you currently check whether the connection was closed. That would of course again rely on exception handling to deal with this error, but I'm not even sure if that isn't the most appropriate approach here as that really is an error case in my opinion which justifies exception handling.

Both approaches terminate waiting requests and notify users about the occurred exception which is the right approach in my opinion. The fact the Cosmos DB for some reason just seems to close connections that we still consider to be active is a different problem that we need to address separately from this problem of getting a hard to understand `NullReferenceException`.

github5775:
As Florian pointed out, a provider, such as CosmosDB, may simply close a connection for lack of activity...each provider with its own specs.  Therefore, I would prefer to have state, not raise an error.  I would prefer to check state in submit before the connection error occurs.

WiredUK mentioned the Parse code checking for null data.  I also added that code to prevent error, but that may be best handled by dealing with the connection drop.  I will try to prove this by removing this if statement and seeing if my workaround prevents this as well.  Specifically, it's possible that this error only comes about when we call the database after the connection has dropped.  I'll let everyone know what happens.  https://issues.apache.org/jira/browse/TINKERPOP-2268?focusedCommentId=16903268&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16903268

plemanach:
any news on this ?

github5775:
@plemanach, have you experienced this timeout?  if so, i have a fix that has been working for a heavily used prod app with no errors for over 6 months.  you can use it while we figure out the best way to handle this.

WiredUK:
@github5775 When I opened this, it had nothing to do with a timeout, it's just the way Gremlin.NET is handling the data stream (looping rather than event-based) and program flow by exceptions. There is no timeout, certainly not for "lack of activity" as these exceptions happen immediately after opening the connection and retrieving data. My latest test throws 12 internal null reference exceptions with 3 seconds for a single Gremlin query.

github5775:
Fair enough, but I can live with that.  What I am referring to is when you make a request, that is successful, and then wait an hour, with the code still running, then when you make another request you'll hit a timeout.

What my workaround does is check the connection before submitting the query, and if the connection is not valid, it renews the connection pool, and then submits the query again.

miczyg:
> @plemanach, have you experienced this timeout? if so, i have a fix that has been working for a heavily used prod app with no errors for over 6 months. you can use it while we figure out the best way to handle this.

@github5775 we have experienced similar error with CosmosDb, can you provide the details what workaround you used for it?

github5775:
@miczyg, on my github i put a copy of gremlin.net, with my modifications.  that has been working fine for me.  i am also in the process of making our gremlin studio code available as well, but currently i just made the software available on [gremlin editor](https://gremlineditor.azurewebsites.net)

miczyg:
@github5775 as I see you are also working on web apps with CosmosDb, graph version in particular - it can be nice to share some experience on CosmosDb Gremlin connectivity and performance tips if you'd like, as MS docs are rather not comprehensive on this subject ;) You can reach me via mail: miczyg13@gmail.com

github5775:
Sure, I am hosting a prod app with cosmos graph.  Did you download the studio?  It can help you get started.

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

miczyg:
Yeah, we also have already prod app with Cosmos Graph, that is why I am now more into performance, optimization and reliability.
I checked out this studio, but unfortunately it have not started for me right away and did not have time to investigate it yet. We also have some DSL libraries that are helping us with Gremlin itself, but as I mentioned now we are looking more on non-functional topics. 
Lets move conversation to mail, not to spoil PR thread ;)

FlorianHockmann:
This PR has been open for quite some time and I would like to get back to the underlying issue @WiredUK wanted to address here.
Since the discussion thread here is quite long, I'll try to summarize the current state first so we're all on the same page:
- The problem was that the driver threw a lot of `NullReferenceException` when `ReceiveAsync` was called on the `WebSocketConnection` while it was not in the `Open` state. Otherwise, the fix to check whether the state is open before calling `ReceiveAsync` wouldn't help.
- @WiredUK mentioned that this happens in just a few seconds after starting his application. So, this is not an issue of Cosmos DB closing connections after a while (as described in [TINKERPOP-2288](https://issues.apache.org/jira/browse/TINKERPOP-2288)).
- The driver [changed in the meantime](https://github.com/apache/tinkerpop/pull/1250) to now throw a clear message when it received `null` / an empty array.

Is that summary correct or did I miss anything important or got something wrong?

While reading the comments again, I wonder whether this one actually points to a good possible solution here:

> Maybe, but then the code gets "stuck" inside the while(true), which is a tight loop, waiting for the external connection check, and only exits when an exception happens.

So, why not simply replace `while(True)` with `while(IsOpen)` in `Connection.ReceiveMessagesAsync`? If the issue here is that we try to receive messages on a closed connection, then this is the easiest fix in my opinion as we already have the information about the connection state in the `Connection` class. So we don't need to introduce a status object and return that from our `WebSocketConnection` class.

The change will actually be a bit more complicated as we need to cancel open requests which results in an exception for the user but I think that this is in general an easier solution to stop calling `ReceiveAsync` on a closed connection.

What I'm however still wondering about is why the connection is not in the `Open` state in the first place. Was it closed by the server? Or by the client? Or is the connection maybe not even open yet when we call `ReceiveAsync` the first time which would explain why @WiredUK sees the exception even within seconds of starting his application.
The last case would be very surprising to me as we're calling `ReceiveAsync` only after `await`ing for `ConnectAsync`.

github5775:
yes, how can the submit be called before the client has connected?  all of my issues were related to loss of connection at significantly subsequent time.  our web application and my gremlin studio application have never seen this case.

FlorianHockmann:
This PR has now been open for nearly two years and the last update was a year ago. So I'm closing this PR as we had no progress towards merging it and it's also not clear whether the problem this PR is trying to address is actually still a problem in the current driver. The discussion also derailed quite a bit here which makes it hard to actually concentrate on the underlying problem and a possible solution.
If the problem still persists in the current driver, then please open an issue [in JIRA](https://issues.apache.org/jira/browse/TINKERPOP/) so we can try to better understand the problem.


----

To still answer @github5775's question:

> i haven't looked at the code (is it oss?).

Of course, you can find the code of Gremlin.Net right in this repository under [gremlin-dotnet](https://github.com/apache/tinkerpop/tree/master/gremlin-dotnet)

github5775:
Makes sense @FlorianHockmann.  Thank you for your great work.

PR #1174, opened 2019-08-01
Use checkResults instead of specifying explicit order of result items

nastra:


spmallette:
Doesn't look like there is any need to assert order here, so lgtm. I will merge this as CTR in time for 3.4.3 release.

VOTE +1

PR #1175, opened 2019-08-05
Tinkerpop 2274 tp33

rpopov:
For initial review - there are still more changes to come

Resolving locale and file system - specific issues when compiling under Windows (10)

### DIFF
--- a/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/util/Log4jRecordingAppender.java
+++ b/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/util/Log4jRecordingAppender.java
@@ -18,13 +18,14 @@
  */
 package org.apache.tinkerpop.gremlin.util;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Pattern;
+
 import org.apache.log4j.AppenderSkeleton;
 import org.apache.log4j.Level;
 import org.apache.log4j.PatternLayout;
 import org.apache.log4j.spi.LoggingEvent;
-
-import java.util.ArrayList;
-import java.util.List;
 
 /**
  * Provides a way to gather logging events for purpose of testing log output.
@@ -37,7 +38,7 @@ public class Log4jRecordingAppender extends AppenderSkeleton {
 
     public Log4jRecordingAppender() {
         super();
-        setLayout(new PatternLayout("%p - %m%n"));
+        setLayout(new PatternLayout("%p - %m%n")); // note the EOLN char(s) appended
     }
 
     @Override
@@ -63,9 +64,18 @@ public class Log4jRecordingAppender extends AppenderSkeleton {
         messages.clear();
     }
 
+    /**
+     * @param regex not null
+     * @return true if there is a substring of a message matching the regular expression, where:
+     *         . matches also the EOLN char(s) defined in the layout.
+     *         $ matches the end of the string
+     */
     public boolean logContainsAny(final String regex) {
-        // chop off the line feed so that the regex doesn't have to account for that
-        return messages.stream().anyMatch(m -> m.substring(0,m.length() - 1).matches(regex));
+        Pattern pattern;
+
+        pattern = Pattern.compile( regex, Pattern.DOTALL );
+
+        return messages.stream().anyMatch(m -> pattern.matcher( m ).find());
     }
 
     public boolean logContainsAny(final String loggerName, final Level level, final String fragment) {
--- a/hadoop-gremlin/pom.xml
+++ b/hadoop-gremlin/pom.xml
@@ -237,8 +237,10 @@ limitations under the License.
                                      <message>Please install Hadoop ${hadoop.version} standalone and set HADOOP_HOME OS environment variable</message>
                                 </requireEnvironmentVariable>
                                 <requireEnvironmentVariable>
+                                     <level>WARN</level>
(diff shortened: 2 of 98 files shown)


spmallette:
Please note that travis is failing the build at this point. The error appears consistent across the different builds, so I think it's a legitimate problem with your changes and not travis instability.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/util/Log4jRecordingAppender.java
```diff
@@ -63,9 +64,18 @@ public void clear() {
...
+    /**
+     * @param regex not null
+     * @return true if there is a substring of a message matching the regular expression, where:
+     *         . matches also the EOLN char(s) defined in the layout.
+     *         $ matches the end of the string
+     */
     public boolean logContainsAny(final String regex) {
-        // chop off the line feed so that the regex doesn't have to account for that
-        return messages.stream().anyMatch(m -> m.substring(0,m.length() - 1).matches(regex));
+        Pattern pattern;
```
spmallette: minor nit - could you inline initialize `pattern` and mark it `final` - note that our typical style is to `final` all variables that don't mutate. if you could give a general review to your changes to fix those where necessary that would be great.

pom.xml
```diff
@@ -554,6 +554,11 @@ limitations under the License.
                     <artifactId>maven-clean-plugin</artifactId>
                     <version>3.1.0</version>
                 </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-enforcer-plugin</artifactId>
+                    <version>3.0.0-M2</version>
```
spmallette: any reason to go to a milestone release at 3.0.0-M2 rather than stick with the last stable version of 1.4.1? I'm not necessarily against it, but we tend to avoid non-final releases unless there is some good reason for upgrading.  Note that if we already use enforcer so you would probably want to delete the `<version>` defined here:

https://github.com/apache/tinkerpop/blob/8734301cfd692d43f911e29cd4b350beca19816f/pom.xml#L183

and allow `<pluginManagement>` to handle that.

hadoop-gremlin/pom.xml
```diff
@@ -219,6 +219,35 @@ limitations under the License.
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-failsafe-plugin</artifactId>
             </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>check-hadoop-installed</id>
```
spmallette: Could you clarify the need for enforcer for these environment variables? Was the build failing without those somehow? I don't have them set at the moment and `mvn clean install` builds fine for me.

rpopov: My idea is to check the environment the build/tests run in and warn that it is incomplete, instead of throwing cryptic exceptions. Having this stated in the POM serves also as some documentation. In this specific case I could not realize the needed build environment setup neither how the CI environment provides Hadoop libraries for the tests

pom.xml
rpopov: Yes, I will change it.

hadoop-gremlin/pom.xml
spmallette: I see where not setting `HADOOP_GREMLIN_LIBS` generates some WARN messages when you build:

```text
[WARN] org.apache.tinkerpop.gremlin.hadoop.jsr223.HadoopGremlinPlugin - Be sure to set the environmental variable: HADOOP_GREMLIN_LIBS
   [WARN] org.apache.hadoop.util.NativeCodeLoader - Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
   [WARN] org.apache.tinkerpop.gremlin.hadoop.jsr223.HadoopGremlinPlugin - Be sure to set the environmental variable: HADOOP_GREMLIN_LIBS
   [WARN] org.apache.tinkerpop.gremlin.hadoop.jsr223.HadoopGremlinPlugin - Be sure to set the environmental variable: HADOOP_GREMLIN_LIBS
   [INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.844 s - in org.apache.tinkerpop.gremlin.hadoop.jsr223.HadoopGremlinPluginTest
```

but it doesn't error, even when you run integration tests on `hadoop-gremlin`. I think those environment variables are really more for usage at runtime rather than for actual tests. You mentioned that you get "cryptic exceptions" - what are those exceptions and do they occur during build on windows?

I'd rather not force folks to setup up environment variables unless they are necessary to the build somehow. But, now that I look at it again, i assume that `<level>WARN</warn>` isn't actually enforcement (meaning the build doesn't fail if those environment variables aren't present). So perhaps this addition is ok given how you have done it even if there aren't errors specifically.

spmallette:
The tp33 branch has been re-opened for development and is set to 3.3.9-SNAPSHOT. Please feel free to rebase.

spmallette:
It looks like you got some extra commits in here somehow. You might need to clean up your git history a bit.

spmallette:
I just wanted to call attention to my previous comment from a few weeks back - unfortunately, we can't easily evaluate/review this PR unless the commit history is cleaned up a bit.

rpopov:
Replaced by https://github.com/apache/tinkerpop/pull/1188 PR

PR #1180, opened 2019-08-14
TINKERPOP-2287: Typeserializer lookup by superclass

justinchuch:
**Discussion**

If I understand the code correctly, looks like the current code could not look up the superclass serializer correctly.

For example, `InetAddress` is the superclass of `Inet4Address`, `Inet6Address`. If we remove the registry entry of `Inet4Adress` and `Inet6Address`, it could not redirect and lookup the serializer registered by the superclass (i.e. `InetAddress`) when we try to get the serializer for `Inet4Address` and `Inet6Address`.

To resolve that, I would like to propose the changes in this PR:

If the serializer of the input class type is not found, try and see if there is a serializer registered with its superclass.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ser/binary/TypeSerializerRegistry.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ser/binary/TypeSerializerRegistry.java
@@ -319,7 +319,7 @@ public class TypeSerializerRegistry {
         }
     }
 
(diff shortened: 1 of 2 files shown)


jorgebay:
I'll look at this tomorrow, thanks @justinchuch for the patch!

jorgebay:
I've moved the discussion to JIRA, see TINKERPOP-2287 .

spmallette:
Given the discussion on the JIRA issue, it seems that we won't do this one - closing.

PR #1182, opened 2019-08-26
TINKERPOP-2289 Use ip address instead of hostname for connection

thirstycrow:
https://issues.apache.org/jira/browse/TINKERPOP-2289

### DIFF
--- a/gremlin-driver/pom.xml
+++ b/gremlin-driver/pom.xml
@@ -104,6 +104,21 @@ limitations under the License.
             <artifactId>hamcrest-all</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-api-mockito</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-module-junit4</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-module-junit4-rule</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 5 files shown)


gremlin-driver/pom.xml
```diff
@@ -98,6 +98,21 @@ limitations under the License.
             <artifactId>hamcrest-all</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
```
spmallette: We use mockito for mocking and would prefer to not include additional new testing dependencies. Could you please convert your test to use that framework instead?

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/AddressingTest.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(Cluster.Builder.class)
+public class AddressingTest {
```
spmallette: Any reason not to add this test to `HostTest`? It seems like it would fit there to me. Note that the two tests there are failing as a result of your change.

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/AddressingTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test
+    public void shouldFindAllHostsWithDeterministicAddress() throws Exception {
+        String hostname = "test.tinkerpop.apache.org";
```
spmallette: Please `final` all variables (even in tests) that are immutable - it's our code style.

spmallette:
Any reason not to rebase this change to `tp33`? I imagine it would apply there as well, no?

gremlin-driver/pom.xml
thirstycrow: Powermock is required to mock the static method of `InetAddress.getAllByName`.

spmallette:
Merged - thank you.

PR #1187, opened 2019-08-29
TINKERPOP-2132 Authentication eagerly to avoid unauthorized responses

thirstycrow:
I met the problem that is already reported in TINKERPOP-2132. TINKERPOP-2205 may have addressed this problem, but it would be too long to wait for the release of tp35. So I propose a quick fix here.


thirstycrow:
@spmallette This merge request is ready for reviewing. It basically follows the same idea as your previous work in the TINKERPOP-2132 branch, which has been there for a while, but is not merged for some reason I wander. 

In this commit, I added a step that sends a validation request at the end of the constructor of `Connection`. So, we don't have to introduce a new message or change the client code to use a new channalizer. I also changed the exception handling of validation, so that `ResponseException`s do not fail the validation,  because in the case when `strictTransactionManagement` is enabled, the validation fails for sure with a `ResponseException`, but the client is still ok for correctly built messages with explicit aliases.

There are some other points I think may worth a change but not included in this commit in order to minimize the modifications.
1. The TimeoutException thrown when no host is available is a little bit misleading. Maybe a specialized NoHostAvailableException is better.
2. It's a commonly used pattern to catch checked exceptions and wrap them in RuntimeExceptions and re-throw. It might become a performance killer when exceptions happen very quickly, because of the expensive `fillInStackTrace` method invocation.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/ClientAuthenticationIntegrateTest.java
```diff
@@ -0,0 +1,82 @@
...
+    @Override
+    public Settings overrideSettings(final Settings settings) {
+        final Settings.AuthenticationSettings authSettings = new Settings.AuthenticationSettings();
+        authSettings.authenticator = Authenticator.class.getName();
+        authSettings.enableAuditLog = true;
+        return settings;
+    }
+
+    @Test
+    public void shouldBeThreadSafeWithAuthentication() throws Exception {
```
spmallette: This test passes on the tp33 branch - if it was testing the problem described in the JIRA issue I would have expected it to fail. Thoughts?

I would instead use this test:

https://issues.apache.org/jira/browse/TINKERPOP-2132?focusedCommentId=16743980&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16743980

According to my comments and to the best of my knowledge it was failing as described in the issue prior to my change in my branch. In that branch I added it to: `GremlinServerAuthIntegrateTest.java` (please just use that rather than creating a new "authentication" test class.

spmallette: Actually, I would add all the tests that I had in my branch - you can see them here: 

https://github.com/apache/tinkerpop/compare/tp33...TINKERPOP-2132

If you can get all those tests to pass then I think you've solved the problem.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
```diff
@@ -938,7 +938,8 @@ public void shouldWorkWithGraphSONExtendedV3Serialization() throws Exception {
 
     @Test
     public void shouldFailClientSideWithTooLargeAResponse() {
-        final Cluster cluster = TestClientFactory.build().maxContentLength(1).create();
+        final int maxContentLength = 128;
```
spmallette: I assume you needed to change this because the validation message is now being sent and it's greater than "1" in size?

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerAuthIntegrateTest.java
```diff
@@ -142,7 +142,7 @@ public void shouldFailAuthenticateWithPlainTextBadPassword() throws Exception {
         } catch(Exception ex) {
             final Throwable root = ExceptionUtils.getRootCause(ex);
             assertEquals(ResponseException.class, root.getClass());
-            assertEquals("Username and/or password are incorrect", root.getMessage());
+            assertEquals("Failed to authenticate", root.getMessage());
```
spmallette: i think the new error message is better, but what caused it to change?

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerAuthKrb5IntegrateTest.java
```diff
@@ -134,7 +133,7 @@ public void shouldFailWithoutClientJaasEntry() throws Exception {
             fail("This should not succeed as the client config does not contain a JaasEntry");
         } catch(Exception ex) {
             final Throwable root = ExceptionUtils.getRootCause(ex);
-            assertTrue(root instanceof ResponseException || root instanceof GSSException);
+            assertTrue(root instanceof TimeoutException);
```
spmallette: two things:

1. The reason for checking both `ResponseException` and `GSSException` is that in different environments (mac os, linux, travis, etc) we tend to see  different exceptions thrown. I've never known why that is except that we used to have a ton of unexpected test failures before we started checking both. Ever since then, Gremlin Server integration tests have been very solid. Did you do something here to ensure that we always get `TimeoutException`.
2. `TimeoutException` doesn't seem like the right exception to be thrown here (or in the test below). We have a bad client configuration - seems like any one of the other exceptions that were previously thrown were better than `TimeoutException` - thoughts?

spmallette:
> It basically follows the same idea as your previous work in the TINKERPOP-2132 branch, which has been there for a while, but is not merged for some reason I wander.

I mentioned in the JIRA ticket why I didn't merge it. It introduced a breaking change to the Gremlin Server protocol to properly fix it and I didn't want to do that for 3.3.x or 3.4.x. I think it would be best if we could avoid that if at all possible as it means previous versions of the driver won't be able to connect along those release lines.

> In this commit, I added a step that sends a validation request at the end of the constructor of Connection.

So, I suppose this approach is a possible solution. You eagerly do the authentication operation of the `Connection` by way of a validation message. I guess that the `Connection` does the websocket handshake in the constructor, so immediately sending a validation message to force authentication might not have bad consequences. It's not an elegant solution, but if this problem is really bothering people and no other non-breaking solutions can be found then I suppose we could go with it ([assuming the tests](https://github.com/apache/tinkerpop/pull/1187#discussion_r320410110) show this fix actually does solve the problem).

> not included in this commit in order to minimize the modifications.

Thank you - that is appreciated.

> The TimeoutException thrown when no host is available is a little bit misleading. Maybe a specialized NoHostAvailableException is better.

I think i made a note about that elsewhere in my review. The `TimeoutException` isn't so good. 

> It's a commonly used pattern to catch checked exceptions and wrap them in RuntimeExceptions and re-throw

Yeah - I think the exception handling could be better. The generalized `RuntimeException` isn't so nice. It would be nice to make that better if we could introduce improvements without breaking changes, but that seems hard at this point. It might need to wait until 3.5.0...not sure.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
thirstycrow: Yes, it is.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerAuthIntegrateTest.java
thirstycrow: The 'Username and/or password are incorrect' exception was thrown at validation time, and was caught and ignored. Subsequent requests will not trigger authentication again because the negotiator is already created.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerAuthKrb5IntegrateTest.java
thirstycrow: I reverted the code to check both `ResponseException` and `TimeoutException`

thirstycrow: and replaced `TimeoutException` with a dedicated `NoHostAvailableException` when no host is capable to serve the request

thirstycrow:
All the tests passed except the last one. It seems caused by some test environment problem rather than the code changes.

spmallette:
VOTE +1

On merge I re-instated `GSSException` assertions as different environments will throw that occasionally (yet another problem with exceptions in the java driver), added changelog and some upgrade docs. Thanks!

spmallette:
Gremlin Server integration tests have been a bit flaky for a number of days now. It is troubling that it is not any particular test that is a problem, but seems somewhat random and entirely bound to Travis executions (I don't seem to be able to get the failures to occur locally). Aside from this change and an additional modification i made to preserve error messages that we had before, I can't think of anything else that would be causing the problem. As such, I'm starting my investigation here, but reverting these changes in a branch to see if Travis stabilizes. If I find that the branch is stable then I imagine I will need to permanently revert the changes in the release branch until it can be determined exactly what introduced this problem. I would expect to re-open the JIRA.

I will continue updates here. I haven't pushed the branch yet, but it is called TINKERPOP-2132-revert

spmallette:
I've run and re-run Travis jobs across all release branches and they continue to fail periodically for various Gremlin Server/Driver integration tests. On the other hand, `TINKERPOP-2132-revert` branch has not failed even a single time, so I sense that something in all this change is a problem. I'm going to spend some more time testing but given that we are closing in on release, I unfortunately expect to re-open this issue and likely merge in this branch to revert the change so that we can ensure stability.

spmallette:
Unfortunately, I've officially reverted this change and merged `TINKERPOP-2132-revert` to the release branches. The issue is re-opened in JIRA.

thirstycrow:
@spmallette Can you give me a link to a failed Travis build related to this merge request? We've been using this code for production for a while. If there are hidden problems, I'd like to find it asap.

spmallette:
i'm not sure that i could at this point. (1) it was a long time ago. (2) a lot of those jobs were restarted to success or more failure and i'd be hard pressed to figure out which was which. i also found that the problem caused stability problems in DS Graph when i tried to utilize it there so Travis isn't the only environment where this problem presents itself.

PR #1188, opened 2019-09-08
[TINKERPOP-2274] Fixed the issues revealed while compiling and runnin…

rpopov:
[TINKERPOP-2274] Fixed the issues revealed while compiling and running tests under Windows

## File system differences
* Revealed that the Spark context is not closed between the tests, while the tests remove its storage files, therefore explicitluy store Spark Cintext for every test, making the tests independent and (more) corect on Spark use.
* Revealed that Spark does not close the FileInputStreams it iterates upon, so that locks hang up in the file system. Thus, not being able to fix Spark/Hadoop and specifically the use of MultiIterator, call explicitly System.gc() in order to finalize the streams remaining open after closing the context.

* In Windows the file locking is diffrent than in LINUX, so rm() tests failed. 
* See: [https://issues.apache.org/jira/browse/SPARK-12216]

## Documentaton
* Documented the convention to use only UNIX-formatted file paths as of the Storage API.
* Revealed the use and documented the Storage interface. 
    
## File names and conventions
* Delegated file parsing to File class for platform independence
* Migrated to use uniform UNIX-formatted paths, even under Windows for platform independence.
* Imposed Path / File classes use instead of manually composing file names.
* Avoided Storage file system paths constructed by concatenating as strings and separators.
* Resolved inconsistency of Storage (file system) API returning null.
* Refactored FileSystemStorage to match the use & specification. 
* Fixed some possible leaking resources. 
* Fixed inconsistencies in the use of / and home directory.
* Changed Storage.rm() to report if it succeeded to remove at least a file (in scope).
* Changed Spark tests close use the non-persistent context, this way closing it when not needed

## Changed tests
* Fixed EarlyLimitStrategyTest to allow running it
* Fixed the loger matching tests to ignore platform-specific EOLN char(s)
* Fixed org.apache.tinkerpop.gremlin.hadoop.process.computer.AbstractHadoopGraphComputerTest to delete the temorary files first before attempting to create new ones with the same name. Provided better error diagnostic messages.
* In tinkerpop-core module extracted from org.apache.tinkerpop.gremlin.TestHelper the CoreTestHelper to allow inheriting it and avoiding collision it with other TestHelper class form tinkepop-test module.
* Updated AbstractStorageCheck#AbstractStorageCheck tests to ease their understanding and help further improvement of error reporting. Fixed the tests of rm(), cp() methods.

## Fixed tests
  * org.apache.tinkerpop.gremlin.spark.SparkGremlinGryoSerializerTest
  * org.apache.tinkerpop.gremlin.spark.SparkGremlinTest
  * that used to fail with:
  * ERROR shouldSupportCopyMethods(org.apache.tinkerpop.gremlin.spark.structure.io.SparkContextStorageCheck)
  * java.lang.AssertionError
            at org.apache.tinkerpop.gremlin.spark.structure.io.SparkContextStorageCheck.shouldSupportCopyMethods(SparkContextStorageCheck.java:75)
    (NOTE: this is an example of "cryptic" error message reported on failed tests) 
  * Fixed reflection-based test to match their parameters at runtime.
    
## Dates
* Used SimpleDateFormat with explicit US locale to parse the test data. 

## POM changes
* Changed POM to dump the whole stack trace in case of test failure.
* Set POM to check if HADOOP_HOME and HADOOP_GREMLIN_LIBS env. vars are set and warn if not, in order not to fail the build with NullPointerException.
* In order to avoid MAVEN warnings, provided explicitly the most-recent version of jacoco in <pluginManagement> and removed <prerequisite>.

* Made the required HADOOP_HOME and HADOOP_GREMLIN_LIBS set up optional to allow Travis build pass, but this basically negates the use of Maven enforcer at all - in case of not installed Hadoop the tests will fail with no indication why, instead of the script to define & test its environment clearly.

## Artifacts 
* Added generation of the tests jar to publish the TestHelper class, that is used both in core and tests projects, as a preparation to avoid the collistion of TestHelper classes in both projects 
* In hadoop project added Readme.md to describe how to install hadoop, spark and OS integration.

### DIFF
--- a/gremlin-core/pom.xml
+++ b/gremlin-core/pom.xml
@@ -146,6 +146,23 @@ limitations under the License.
                     </compilerArgs>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <executions>
+                  <execution>
+                    <id>package-test-lib</id>
+                    <goals>
+                      <goal>test-jar</goal>
+                    </goals>
+                    <configuration>
+                      <includes>
+                        <include>**/*TestHelper*</include>
+                      </includes>
+                    </configuration>
+                  </execution>
+                </executions>
+            </plugin>
         </plugins>
     </build>
 </project>
--- a/gremlin-core/pom.xml
+++ b/gremlin-core/pom.xml
@@ -113,6 +113,12 @@ limitations under the License.
             <artifactId>compile-testing</artifactId>
             <version>0.10</version>
             <scope>test</scope>
+            <exclusions>
+              <exclusion>
+                <artifactId>tools</artifactId>
+                <groupId>com.sun</groupId>
+              </exclusion>
+            </exclusions>
         </dependency>
     </dependencies>
     <build>
@@ -145,23 +151,6 @@ limitations under the License.
                         <arg>-proc:none</arg>
                     </compilerArgs>
                 </configuration>
-            </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-jar-plugin</artifactId>
-                <executions>
-                  <execution>
-                    <id>package-test-lib</id>
-                    <goals>
-                      <goal>test-jar</goal>
-                    </goals>
-                    <configuration>
-                      <includes>
-                        <include>**/*TestHelper*</include>
-                      </includes>
-                    </configuration>
-                  </execution>
-                </executions>
             </plugin>
         </plugins>
     </build>
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/CoreTestHelper.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/CoreTestHelper.java
@@ -16,16 +16,12 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.tinkerpop.gremlin;
-
-import static org.junit.Assert.assertTrue;
+package org.apache.tinkerpop.gremlin.util;
 
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.Modifier;
 import java.net.URL;
 import java.util.List;
 import java.util.stream.Collectors;
@@ -41,15 +37,6 @@ public class CoreTestHelper {
     public static final String TEST_DATA_RELATIVE_DIR = "test-case-data";
 
     protected CoreTestHelper() {}
-
-    public static void assertIsUtilityClass(final Class<?> utilityClass) throws Exception {
-        final Constructor constructor = utilityClass.getDeclaredConstructor();
-
-        assertTrue(Modifier.isFinal(utilityClass.getModifiers()));
-        assertTrue(Modifier.isPrivate(constructor.getModifiers()));
-        constructor.setAccessible(true);
-        constructor.newInstance();
-    }
 
     /**
      * Creates a {@link File} reference that points to a directory relative to the supplied class in the
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/CoreTestHelper.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/CoreTestHelper.java
@@ -18,6 +18,8 @@
  */
 package org.apache.tinkerpop.gremlin.util;
 
+import org.apache.tinkerpop.gremlin.structure.io.Storage;
+
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
@@ -26,8 +28,6 @@ import java.net.URL;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-
-import org.apache.tinkerpop.gremlin.structure.io.Storage;
 
 /**
  * @author Stephen Mallette (http://stephen.genoprime.com)
@@ -113,7 +113,7 @@ public class CoreTestHelper {
         final String buildDirectory = System.getProperty("build.dir");
 
         if ( null == buildDirectory ) {
-            final String clsUri = clazz.getName().replace(".", Storage.FILE_SEPARATOR) + ".class";
+            final String clsUri = clazz.getName().replace(".", "/") + ".class";
             final URL url = clazz.getClassLoader().getResource(clsUri);
             final String clsPath = url.getPath();
             final String computePath = clsPath.substring(0, clsPath.length() - clsUri.length());
--- a/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/util/ComparatorsTest.java
+++ b/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/util/ComparatorsTest.java
@@ -18,19 +18,19 @@
  */
 package org.apache.tinkerpop.gremlin.structure.util;
 
-import org.apache.tinkerpop.gremlin.CoreTestHelper;
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import java.util.Arrays;
+import java.util.List;
+
+import org.apache.tinkerpop.gremlin.TestHelper;
 import org.apache.tinkerpop.gremlin.structure.Edge;
 import org.apache.tinkerpop.gremlin.structure.Element;
 import org.apache.tinkerpop.gremlin.structure.Property;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.junit.Test;
-
-import java.util.Arrays;
-import java.util.List;
-
-import static org.junit.Assert.assertEquals;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
 
 /**
  * @author Stephen Mallette (http://stephen.genoprime.com)
@@ -38,7 +38,7 @@ import static org.mockito.Mockito.when;
 public class ComparatorsTest {
     @Test
     public void shouldBeUtilityClass() throws Exception {
-        CoreTestHelper.assertIsUtilityClass(Comparators.class);
+        TestHelper.assertIsUtilityClass(Comparators.class);
     }
 
     @Test
--- a/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/util/ComparatorsTest.java
+++ b/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/util/ComparatorsTest.java
@@ -18,7 +18,7 @@
  */
 package org.apache.tinkerpop.gremlin.structure.util;
 
-import org.apache.tinkerpop.gremlin.TestHelper;
+import org.apache.tinkerpop.gremlin.CoreTestHelper;
 import org.apache.tinkerpop.gremlin.structure.Edge;
 import org.apache.tinkerpop.gremlin.structure.Element;
 import org.apache.tinkerpop.gremlin.structure.Property;
@@ -38,7 +38,7 @@ import static org.mockito.Mockito.when;
 public class ComparatorsTest {
     @Test
     public void shouldBeUtilityClass() throws Exception {
-        TestHelper.assertIsUtilityClass(Comparators.class);
+        CoreTestHelper.assertIsUtilityClass(Comparators.class);
     }
 
     @Test
--- a/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/util/ComparatorsTest.java
+++ b/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/util/ComparatorsTest.java
@@ -18,19 +18,19 @@
  */
 package org.apache.tinkerpop.gremlin.structure.util;
 
-import static org.junit.Assert.assertEquals;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
-
-import java.util.Arrays;
-import java.util.List;
-
 import org.apache.tinkerpop.gremlin.TestHelper;
 import org.apache.tinkerpop.gremlin.structure.Edge;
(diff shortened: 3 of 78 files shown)


okram:
Wow. Rad PR man. Love your README.

Goodness to you,
Marko.

http://rredux.com <http://rredux.com/>

gremlin-core/pom.xml
```diff
@@ -146,6 +146,23 @@ limitations under the License.
...
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <executions>
+                  <execution>
+                    <id>package-test-lib</id>
+                    <goals>
+                      <goal>test-jar</goal>
```
spmallette: we try to avoid test jars - `TestHelper`is already in `gremlin-test` - https://github.com/apache/tinkerpop/blob/06f5213a22986574751295042aaec3ce232b4cd2/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/TestHelper.java - yes, it is a small code duplication, but I'd prefer that to a special classified artifact.

spmallette:
I'm not sure if this is ready for review, but it seems like Travis isn't happy with the spark build. It seems to be timing out? That could mean that the build got slower as a result of this change or that the changes are causing the tests to simply hang.

gremlin-core/pom.xml
rpopov: Resolved as extracting from gremlin-core/src/test/.../TestHelper  classes the     - CoreTestHelper class in gremlin-core/main/src/main/java/..../util where to keep the unique code to manage directories and files, inherited by gremlin-test/...TestHelper class, which is referred everywhere (but gremlin-core). This left in gremlin-core/src/test/.../TestHelper the only truly test method which is the only duplication kept. This way no new test artifact is needed at the cost of one simple method duplicated.

rpopov:
Please review and consider the PR finalized.

pom.xml
```diff
@@ -554,6 +554,11 @@ limitations under the License.
                     <artifactId>maven-clean-plugin</artifactId>
                     <version>3.1.0</version>
                 </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-enforcer-plugin</artifactId>
```
robertdale: @rpopov enforcer plugin shows up twice here.  Should be merged to one.

gremlin-core/pom.xml
```diff
@@ -113,6 +113,12 @@ limitations under the License.
             <artifactId>compile-testing</artifactId>
             <version>0.10</version>
             <scope>test</scope>
+            <exclusions>
```
spmallette: could you please explain why this exclusion was necessary?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/CoreTestHelper.java
```diff
@@ -0,0 +1,174 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.apache.tinkerpop.gremlin.structure.io.Storage;
+
+/**
```
spmallette: I've been hesitant to move a "test" class to gremlin-core because such classes have no use outside of TinkerPop test cases. I preferred the minor code duplication that we had before. I'd say that I'm +0 for this change, however if you feel strongly that the duplication should be removed, please include some javadoc here on this class to explain that this class is really for testing purposes only and that its API may change without notice. That way if someone uses it (you'd be surprised at that classes we've made public before that people find a way to use) we absolve ourselves of some of the responsibility for maintaining the API. I can't think of any other way to restrict this class specifically to "test" except to add a new `gremlin-test-core`  module or something, but I can't think of any other code that would be helpful in such a module, so that seems like overkill.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/util/ComparatorsTest.java
```diff
@@ -18,20 +18,20 @@
  */
 package org.apache.tinkerpop.gremlin.structure.util;
 
+import static org.junit.Assert.assertEquals;
```
spmallette: There are a lot of these kinds of changes in the pull request where you just change the import order. I imagine that change was done without your knowing by your IDE. Could you please revert those sorts of changes as:

1. It's the opposite of our style - we tend to typically place static imports last and "java.*" imports after other package imports
2. It will simplify this diff for purpose of review.

Thanks.

spmallette:
Just a general question - in all the changes you made to file path calculations here, did the actual paths themselves change in any way or are all test file paths still going to the same relative location but you just make it so that all of those file paths are now accessible irrespective of operating system?

rpopov:
Yes, Stephen, you are right - all path changes are in the way to compose the path in an OS-independent way (by composing File and Path instances, not referring file separators), but the effective paths do not change.
There are other changes related to the date parsing in an uniform way, providing explicit date format.

gremlin-core/pom.xml
rpopov: com.google.testing.compile imposes a transitive dependency from JDK's tools.jar. It is not published in the Maven central. If it is not clear if tools.jar is used at all - the unit and integration tests pass with tools.jar excluded.
The import of tinkerpop-core as a Maven project in Eclipse registers tools.jar as a library, but Eclipse m2e cannot resolve it. This blocks the project's compilation in the IDE.

pom.xml
rpopov: @robertdale, the second occurrence of maven-enforcer-plugin (above) is in the <pluginManagement> section and states plugin's version to be used in every tinkerpop module. Actually, the practice to use plugins and dependencies with different ad-hoc defined versions exists in tinkerpop and I understand the *.version properties like log4j.version defined in this pom as an attempt to unify the versions. The Maven way is to define them in <dependecyManagement> and <pluginManagement> once. Thus, I would suggest a new task - unify the version management of plugins and dependencies the Maven way.

robertdale: ah, so it is. good job.

robertdale:
@rpopov I get two test errors in gremlin-core on Windows 10.  Any ideas?

```
PS C:\Users\rdale6\git\clones\tinkerpop> mvn test -rf :gremlin-core -X -DforkCount=0
Apache Maven 3.6.2 (40f52333136460af0dc0d7232c0dc0bcf0d9e117; 2019-08-27T11:06:16-04:00)
Maven home: C:\Users\rdale6\opt\apache-maven-3.6.2\bin\..
Java version: 1.8.0_221, vendor: Oracle Corporation, runtime: C:\Users\rdale6\opt\jdk1.8.0_221\jre
Default locale: en_US, platform encoding: Cp1252
OS name: "windows 10", version: "10.0", arch: "amd64", family: "windows"

...

[ERROR] Tests run: 3, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.009 s <<< FAILURE! - in org.apache.tinkerpop.gremlin.jsr223.ScriptEngineCacheTest
[ERROR] shouldGetEngineFromCache(org.apache.tinkerpop.gremlin.jsr223.ScriptEngineCacheTest)  Time elapsed: 0.002 s  <<< ERROR!
java.lang.IllegalArgumentException: There is no script engine with provided name: nashorn
        at org.apache.tinkerpop.gremlin.jsr223.ScriptEngineCache.lambda$get$0(ScriptEngineCache.java:48)
        at java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1853)
        at org.apache.tinkerpop.gremlin.jsr223.ScriptEngineCache.get(ScriptEngineCache.java:44)
        at org.apache.tinkerpop.gremlin.jsr223.ScriptEngineCacheTest.shouldGetEngineFromCache(ScriptEngineCacheTest.java:38)

...

[INFO] Running org.apache.tinkerpop.gremlin.process.traversal.dsl.GremlinDslProcessorTest
[ERROR] Tests run: 4, Failures: 4, Errors: 0, Skipped: 0, Time elapsed: 0.512 s <<< FAILURE! - in org.apache.tinkerpop.gremlin.process.traversal.dsl.GremlinDslProcessorTest
[ERROR] shouldCompileTraversalAndTraversalSourceToDefaultPackage(org.apache.tinkerpop.gremlin.process.traversal.dsl.GremlinDslProcessorTest)  Time elapsed: 0.293 s  <<< FAILURE!
java.lang.AssertionError:
Compilation produced the following errors:
/C:/Users/rdale6/git/clones/tinkerpop/gremlin-core/target/test-classes/org/apache/tinkerpop/gremlin/process/traversal/dsl/SocialPackageTraversalDsl.java:21: error: cannot find symbol
import org.apache.tinkerpop.gremlin.process.traversal.dsl.GremlinDsl;
                                                         ^
  symbol:   class GremlinDsl
  location: package org.apache.tinkerpop.gremlin.process.traversal.dsl
```

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/util/ComparatorsTest.java
rpopov: Restored the original order of imports. Please review.

rpopov:
@robertdale, I cannot reproduce the reported situation but this could be a hint:
* please check if your JRE/lib/ext of the active JAVA contains nashorn.jar - the default Nashorn implementation coming with Oracle Java 8+ (but excluded / discontinued in Java 13)
* please check if your test's classpath contains jdk.nashorn.api.scripting.NashornScriptEngine class - the default Nashorn implementation
* once the default Nashorn implementation is in jre/lib/ext, it goes to the system classloader and cannot easily be excluded from it, therefore please check in the referred above test report file in in XML and verify the values of the properties:
    * java.runtime.version
    * java.home
    * java.ext.dirs  and the contents of that extension directories for nashorn.jar

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/CoreTestHelper.java
rpopov: The problem was that there were two TestHelper classes with exactly the same qualified name with different contents that confused the IDE. In addition, the changes I introduced for unification of the test directories. Therefore I extracted the unified functionality in a new CoreTestHelper class, so that the second TestHelper could inherit from CoreHestHelper and thus reduced the duplication to a single method in a single class.

spmallette: Thanks for the explanation. I've never had the IDE problems, but as I said, i'm indifferent to the change (+0 meaning i'm not for it, but won't stand in the way of it merging as it is). I'd only ask for some javadoc to explain that this class is really for testing purposes only and that its API may change without notice. I can add it myself on merge if you're unable to get to doing that.

spmallette:
> the default Nashorn implementation coming with Oracle Java 8+ (but excluded / discontinued in Java 13)

Interesting - didn't know that. I guess we're still trying to get off Java 8 though - stuck with upstream problems with spark/hadoop ecosystem still.

pom.xml
```diff
@@ -109,9 +109,6 @@ limitations under the License.
             <url>https://github.com/twilmes</url>
         </contributor>
     </contributors>
-    <prerequisites>
-        <maven>3.3.9</maven>
```
spmallette: any reason you removed 3.3.9? 

it's interesting that we have 3.3.9 at the maven level, but enforcer only requires 3.2.5 (and our dev docs also say the minimum is 3.2.5).  I noticed that the reason we went to a 3.3.9 minimum was something related to the jacoco plugin requiring that version:

https://github.com/apache/tinkerpop/commit/5c9c8ce1e602c7766f4f7200ba5ec4e66718fbc5

I wonder if we should really be at 3.3.9 rather than 3.2.5...thoughts?

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/util/ComparatorsTest.java
spmallette: lgtm - thank you for doing that.

spmallette:
@robertdale did you ever get a chance to return to this one? I see your using Java version: 1.8.0_221 so it would be curious for nashorn to be missing. anyway, i'm running some final tests on this now around less used build options and documentation generation. if all that pans out and we sort out the "required maven version" issue I just commented on, i'm inclined to be in favor of merging this.

robertdale:
Nashorn is there. Paths, settings look good. `jjs` works.  Also, it doesn't explain the other two test failures which don't appear to have anything to do with scripting.  I don't know what's special about these three tests.

spmallette:
i noticed that you tried to do: `mvn test -rf :gremlin-core` ... any difference if you do the full business of `mvn clean install`?

also, just to be clear, i assume you only see these problems on windows? does linux/mac/whatever you normally use work ok?

robertdale:
I start out with doing the `mvn clean install` at the root.  But to get the debug and stacktrace messages, I just start from `gremlin-core`.  Exact same repository works through Ubuntu on [WSL](https://docs.microsoft.com/en-us/windows/wsl/about).  So it appears to be some windows-specific issue.  I'll try to look into it some more this weekend.

@rpopov could you provide a screenshot of the final output from maven on windows just for warm fuzzies?

spmallette:
i have a quick update in thinking on this one - currently we have a working build on linux and at least one windows environment where there is a problem.  somewhere in all the discussion on this i've pointed out that we don't really support a windows build so if the windows build is working it is a convenience mostly. Given that these changes have no real effect on Linux building and they fix a number of problems with locale and file system inconsistencies I think that if Windows is still somehow failing in certain cases, that's not a big problem - we still sorta gain for merging. 

how does that sit with you @robertdale ?

robertdale:
Yes, works for me.  The changes are sound.  Just wanted some evidence that they addressed the actual issue since I couldn't reproduce it on my side. But since then I have run into other projects that don't build so it must be something local.

VOTE +1

spmallette:
hit a problem manually testing the console and plugin installation. just had a hunch about it and bam...problem. easy fix, but now re-testing documentation generation (the other place where everything goes sideways).

PR #1189, opened 2019-09-09
TINKERPOP-2284 Added elementMap() step

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2284

Given that we only return reference objects for graph elements now, there is a lot of extra `project()`/`valueMap()` work to do for the simple and common need to return all the structure of a graph element. `elementMap()` simplifies all that to a single step without a ton of options and added features.

VOTE +1

### DIFF
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -829,6 +829,37 @@ g.V()
 
 * link:++http://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.html#drop--++[`drop()`]
 
+[[elementmap-step]]
+=== ElementMap Step
+
+The `elementMap()`-step yields a `Map` representation of the structure of an element.
+
+[gremlin-groovy,modern]
+----
+g.V().elementMap()
+g.V().elementMap('age')
+g.V().elementMap('age','blah')
+g.E().elementMap()
+----
+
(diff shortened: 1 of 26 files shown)


gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/CommonSteps.cs
```diff
@@ -244,6 +245,16 @@ private static object ToT(string enumName, string graphName)
             return T.GetByValue(enumName);
         }
 
+        private static object ToDirection(string enumName, string graphName)
+        {
+            return Direction.GetByValue(enumName);
+        }
+
+//        private static string toNormalCase(string name)
```
FlorianHockmann: (nitpick) Why this commented out code?

spmallette: it was just some left over code from trying to debug a problem - removed it and force pushed.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -829,6 +829,37 @@ g.V()
...
+[gremlin-groovy,modern]
+----
+g.V().elementMap()
+g.V().elementMap('age')
+g.V().elementMap('age','blah')
+g.E().elementMap()
+----
+
+It is important to note that the map of a vertex assumes that cardinality for each key is `single` and if it is `list`
+then only the first item encountered will be returned. As `single` is the more common cardinality for properties this
```
jorgebay: Good call!

PR #1190, opened 2019-09-11
TINKERPOP-2285 Added ResponseError in gremlin-javascript

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2285

`ResponseError` in javascript is akin to `ResponseException` in java. It includes more direct access to protocol lever server error information such as the status code and status attributes which means we don't need to pack that information into the error message itself at all. Kept that old style to the error messages just in case users were parsing those in prior versions - didn't want to break that.

Builds with `mvn clean install -pl gremlin-javascript`

VOTE +1

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/response-error.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/response-error.js
@@ -1,0 +1,47 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+'use strict';
+
+/**
+ * Represents an error obtained from the server.
+ */
+class ResponseError extends Error {
+  constructor(message, responseStatus) {
+    super(message);
+    this.name = "ResponseError";
+
+    /**
+     * Gets the server status code.
+     */
+    this.statusCode = responseStatus.code;
+
+    /**
+     * Gets the server status message.
+     */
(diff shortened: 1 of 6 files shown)


gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/response-error.js
```diff
@@ -0,0 +1,33 @@
...
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+'use strict';
+
+class ResponseError extends Error {
```
jorgebay:
```suggestion
/**
 * Represents an error obtained from the server.
 */
class ResponseError extends Error {
```

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/response-error.js
```diff
@@ -0,0 +1,33 @@
...
+    constructor(message, responseStatus) {
+        super(message);
+        this.name = "ResponseError";
+        this.statusCode = responseStatus.code;
```
jorgebay: If we add `/** description */` before a property declaration jsdoc will define the property, for example:

```javascript
  /**
   * Gets the status code.
   */
  this.statusCode = responseStatus.code;
```

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/response-error.js
```diff
@@ -0,0 +1,33 @@
...
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+'use strict';
+
+class ResponseError extends Error {
+    constructor(message, responseStatus) {
```
jorgebay: For consistency, use 2 spaces for indentation.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/response-error.js
spmallette: didn't know that - thanks

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/response-error.js
spmallette: oops - IDE needs to get fixed up

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/response-error.js
spmallette: done

spmallette:
Updated based on comments - hope I got the index.js right.  Thanks - learned a few things about js in this review.

PR #1197, opened 2019-09-22
Fix Bytecode.addSource  arguments is null bug

sandszhouSZ:
I think it is a simple bug and and could  recurrent easily,
I both fix two position and add one test case.

old jira here:
https://issues.apache.org/jira/browse/TINKERPOP-2298?jql=reporter%20in%20(currentUser())

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Bytecode.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Bytecode.java
@@ -62,13 +62,17 @@ public final class Bytecode implements Cloneable, Serializable {
      */
     public void addSource(final String sourceName, final Object... arguments) {
         if (sourceName.equals(TraversalSource.Symbols.withoutStrategies)) {
-            final Class<TraversalStrategy>[] classes = new Class[arguments.length];
-            for (int i = 0; i < arguments.length; i++) {
-                classes[i] = arguments[i] instanceof TraversalStrategyProxy ?
-                        ((TraversalStrategyProxy) arguments[i]).getStrategyClass() :
-                        (Class) arguments[i];
+            if (arguments == null)
+                this.sourceInstructions.add(new Instruction(sourceName, null));
+            else {
+                final Class<TraversalStrategy>[] classes = new Class[arguments.length];
+                for (int i = 0; i < arguments.length; i++) {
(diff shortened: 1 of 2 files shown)


spmallette:
This may be a good fix for `Bytecode` but should bytecode construction even get this far. `outE(null)` really shouldn't be a valid argument to that step. After this fix is applied, the `Bytecode` gets constructed, but what do the traversal do? does a traversal wtih `outE(null)` actually execute? does it behave as `outE()` or does it result in some other failure?

sandszhouSZ:
Good question, I think outE(null) should behave as outE()。 The reasons are:
1.  [null is a special token ](https://stackoverflow.com/questions/2707322/what-is-null-in-java)
Generally speaking, null are used as a special value to signify:
- Uninitialized state
- Termination condition
- Non-existing object
- An unknown value
I think as outE(null),  null's  meaning will be  An unknown value / Non-existing object,
as outE will be filter by arg **final String... edgeLabels** and  edgeLabels = null means
filter by unknown / Non-existing args which also means no filter at all

2. A readily sample as following,  it indicates null is just ignored directly:
gremlin> g.V(1).outE("knows", null)
==>e[7][1-knows->2]
==>e[8][1-knows->4]

3. In standard  Map structure, when the key is null  the map will return the value just if containsKey(null)  == true or return null otherwise。but as  Industrial grade product, we hardly ever met a map has null as it's key, so null here also means **Non-existing keys**

spmallette:
This issue may be touching on concepts that will be better sorted out in:

https://issues.apache.org/jira/browse/TINKERPOP-2235

Please feel free to add ideas there if you like. I'm ok with merging this as a specific bug fix to `Bytecode`. @dkuppitz any objections or other comments?

Assuming no objections from others, I think this issue goes back to 3.3.x so I would suggest you re-target your pull request to the `tp33` branch.

sandszhouSZ:
[Provide our own Null class](https://lists.apache.org/thread.html/cd6a2e8eb0c369a8814c217fe69910e0cf4613286e887e90db28e061@%3Cdev.tinkerpop.apache.org%3E) is a good choice。
I completely agree @Dmitry Novikov suggestions：
`gremlin> g.V().limit(1).project('a', 'b').by(values('name')).by(coalesce(values('notSureIfExists'), constant(Null.instance())))
==>[a:marko,b:null]
seems more elegant than just throw exception`
I look forward to this idea will come true as soon as possible

PR #1204, opened 2019-09-30
TINKERPOP-2302 add `ElementMapStep#isOnGraphComputer()` (master)

dalaro:
Adds a getter for `onGraphComputer`.  This is intended to support strategies that replace this step but want to preserve this state across replacement.

https://issues.apache.org/jira/browse/TINKERPOP-2302


spmallette:
merging via CTR

PR #1206, opened 2019-10-03
TINKERPOP-2289 Better account for localhost with driver

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2289

Pretty small add-on patch to this issue. Not sure if this is the right fix for the comments on the related JIRA - asked for feedback over there.


spmallette:
killing this. @robertdale had a nice suggestion on the JIRA for solving this problem.

PR #1208, opened 2019-10-06
Removing '_' from `from` GraphTraversal Method

ktstowell:
Removing '_' from `from` GraphTraversal Method method to match documentation (http://tinkerpop.apache.org/docs/current/reference/#addedge-step - please notice step "7" api example.) 

- Would require Major version bump due to breakages in backwards compatibility.
- Could be a patch if we left `from_` unaltered, and added `from` which proxied the same thing.

This is largely just to start a conversation so that code can be referenced in a issue.


spmallette:
Hi, as I understood it, the idea is to futureproof `from` as a reserved word in JS as described here:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import

note that we do mention the underscore for JS in `from()` documentation:

http://tinkerpop.apache.org/docs/current/reference/#from-step

ktstowell:
Awesome! Thanks for clearing that up!

PR #1209, opened 2019-10-11
TINKERPOP-2305 GraphBinary: Buffer API

jorgebay:
https://issues.apache.org/jira/browse/TINKERPOP-2305

Introduce our own `Buffer` interface as a wrapper of Netty's `ByteBuf`, allowing us to move GraphBinary serialization to `gremlin-core`.


jorgebay:
I'll add a couple of utility methods to access the underlying memory of `Buffer` for providers to use if needed.

jorgebay:
In most cases, writing to and reading from `Buffer` using built in methods will be enough. In cases where providers need to read and write directly to Netty buffers, I've exposed a couple of utility methods that allow access to the raw memory block under `Buffer` instances using Netty: `NettyBufferFactory.readRaw()` and `NettyBufferFactory.writeRaw()`.

VOTE +1

nastra:
VOTE +1

spmallette:
I think you got the fix right, but please make sure that his works:

```text
mvn clean install -pl :gremlin-io-test -Dio
```

Also, needs a few `final` declarations on variables to match general code style. Other than that, 

VOTE +1

We can merge after I re-open the release branches.

jorgebay:
Good catch about `gremlin-io-test`!

I've fixed it and declared parameters `final` in the public API.

PR #1211, opened 2019-10-30
TINKERPOP-1568 Changed Strategy Application Methodology

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-1568

Please refer to the JIRA issue for details as well as the Upgrade Docs on this PR for more information about this one. I'm generally happy with how few changes were actually involved in making this happen - if you take away my light refactoring for inclusion of the `IsRoot()` method very few files had to really change and for the most part the changes themselves were fairly straightforward. It didn't always look this way - took a fair amount of effort to get to the root of why certain changes were necessary to then refactor for greater simplicity. 

I think the only thing I wasn't really able to sort out "nicely" was `MessagePassingReductionStrategy`. It seems to need to know a lot about `profile()` for some reason and I wasn't quite able to sort out a better way to do what was necessary there. I think I'm just going to have to submit that as-is unless someone has better ideas.

VOTE +1

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Traversal.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Traversal.java
@@ -149,7 +149,7 @@ public interface Traversal<S, E> extends Iterator<E>, Serializable, Cloneable, A
     /**
      * Starts a promise to execute a function on the current {@code Traversal} that will be completed in the future.
      * Note that this method can only be used if the {@code Traversal} is constructed using
-     * {@link TraversalSource#withRemote(Configuration)}. Calling this method otherwise will yield an
+     * {@link AnonymousTraversalSource#withRemote(Configuration)}. Calling this method otherwise will yield an
      * {@code IllegalStateException}.
      */
     public default <T> CompletableFuture<T> promise(final Function<Traversal<S, E>, T> traversalFunction) {
@@ -413,7 +413,8 @@ public interface Traversal<S, E> extends Iterator<E>, Serializable, Cloneable, A
         /**
          * Apply the registered {@link TraversalStrategies} to the traversal.
          * Once the strategies are applied, the traversal is "locked" and can no longer have steps added to it.
-         * The order of operations for strategy applications should be: globally id steps, apply strategies to root traversal, then to nested traversals.
+         * The order of operations for strategy applications should be: globally id steps, apply each strategy in turn
+         * to root traversal, then recursively to nested traversals.
          *
          * @throws IllegalStateException if the {@link TraversalStrategies} have already been applied
          */
@@ -421,7 +422,8 @@ public interface Traversal<S, E> extends Iterator<E>, Serializable, Cloneable, A
 
         /**
          * Get the {@link TraverserGenerator} associated with this traversal.
-         * The traversal generator creates {@link Traverser} instances that are respective of the traversal's {@link org.apache.tinkerpop.gremlin.process.traversal.traverser.TraverserRequirement}.
+         * The traversal generator creates {@link Traverser} instances that are respective of the traversal's
+         * {@link org.apache.tinkerpop.gremlin.process.traversal.traverser.TraverserRequirement}.
          *
          * @return the generator of traversers
          */
@@ -476,20 +478,29 @@ public interface Traversal<S, E> extends Iterator<E>, Serializable, Cloneable, A
         public TraversalStrategies getStrategies();
 
         /**
-         * Set the {@link org.apache.tinkerpop.gremlin.process.traversal.step.TraversalParent} {@link Step} that is the parent of this traversal.
-         * Traversals can be nested and this is the means by which the traversal tree is connected.
+         * Set the {@link org.apache.tinkerpop.gremlin.process.traversal.step.TraversalParent} {@link Step} that is
+         * the parent of this traversal. Traversals can be nested and this is the means by which the traversal tree is
+         * connected.
          *
          * @param step the traversal holder parent step
          */
         public void setParent(final TraversalParent step);
 
         /**
-         * Get the {@link org.apache.tinkerpop.gremlin.process.traversal.step.TraversalParent} {@link Step} that is the parent of this traversal.
-         * Traversals can be nested and this is the means by which the traversal tree is walked.
+         * Get the {@link org.apache.tinkerpop.gremlin.process.traversal.step.TraversalParent} {@link Step} that is
+         * the parent of this traversal. Traversals can be nested and this is the means by which the traversal tree is
+         * walked.
          *
          * @return the traversal holder parent step
          */
         public TraversalParent getParent();
+
+        /**
+         * Determines if the traversal is at the root level.
+         */
+        public default boolean isRoot() {
+            return null == getParent() || getParent() instanceof EmptyStep;
+        }
 
         /**
          * Cloning is used to duplicate the traversal typically in OLAP environments.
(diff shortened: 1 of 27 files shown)


jorgebay:
Nice!
I'll be able to review it by early next week.

dkuppitz:
Well done, code looks so much cleaner now :+1: 

What I like most about this change is how `InlineFilterStrategy` out of sudden does what it's supposed to do.

VOTE +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
```diff
@@ -121,34 +121,54 @@ public TraverserGenerator getTraverserGenerator() {
...
+        // steps get reId'd and traverser requirements are set.
+        if (isRoot() || this.getParent() instanceof VertexProgramStep) {
+
+            // note that prior to applying strategies to children we used to set side-effects and strategies of all
+            // children to that of the parent. under this revised model of strategy application from TINKERPOP-1568
+            // it doesn't appear to be necessary to do that (at least from the perspective of the test suite). by,
+            // moving side-effect setting after actual recursive strategy application we save a loop and by
+            // consequence also fix a problem where strategies might reset something in sideeffects which seems to
+            // happen in TranslationStrategy.
+            final Iterator<TraversalStrategy<?>> strategyIterator = this.strategies.toIterator();
```
jorgebay: NIT: Any reason not to use the `for-each` syntax?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/TraversalHelper.java
```diff
@@ -465,7 +465,11 @@ public static boolean anyStepRecursively(final Predicate<Step> predicate, final
      */
     public static void applyTraversalRecursively(final Consumer<Traversal.Admin<?, ?>> consumer, final Traversal.Admin<?, ?> traversal) {
         consumer.accept(traversal);
-        for (final Step<?, ?> step : traversal.getSteps()) {
+
+        // we get accused of concurrentmodification if we try a for(Iterable)
```
jorgebay: I think it can be simplified into:

```java
traversal.getSteps().stream().filter(step -> step instanceof TraversalParent).forEach(step -> {
    for (final Traversal.Admin<?, ?> local : ((TraversalParent) step).getLocalChildren()) {
        applyTraversalRecursively(consumer, local);
    }
    for (final Traversal.Admin<?, ?> global : ((TraversalParent) step).getGlobalChildren()) {
        applyTraversalRecursively(consumer, global);
    }
});
```

I'm not sure I understand why you were getting concurrent modification exception.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalStrategies.java
```diff
@@ -69,10 +70,15 @@
...
     /**
-     * Return all the {@link TraversalStrategy} singleton instances associated with this {@link TraversalStrategies}.
+     * Return all the {@link TraversalStrategy} instances associated with this {@link TraversalStrategies}.
      */
     public List<TraversalStrategy<?>> toList();
 
+    /**
+     * Return all the {@link TraversalStrategy} instances associated with this {@link TraversalStrategies}.
+     */
+    public Iterator<TraversalStrategy<?>> toIterator();
```
jorgebay: I think this is a sign that `TraversalStrategies` should extend `Iterable<TraversalStrategy<?>>` instead.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Traversal.java
```diff
@@ -476,21 +478,30 @@ public default void reset() {
...
          *
          * @return the traversal holder parent step
          */
         public TraversalParent getParent();
 
+        /**
+         * Determines if the traversal is at the root level.
+         */
+        public default boolean isRoot() {
+            return null == getParent() || getParent() instanceof EmptyStep;
```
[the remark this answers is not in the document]
spmallette: that can be investigated: https://issues.apache.org/jira/browse/TINKERPOP-2310

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalStrategies.java
spmallette: I thought about that, but was trying not to do too much. I created an issue for further consideration: https://issues.apache.org/jira/browse/TINKERPOP-2311

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
spmallette: I guess I like iterators 😄  - will consider further adjustments with https://issues.apache.org/jira/browse/TINKERPOP-2311

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/TraversalHelper.java
spmallette: I guess you get concurrent modification because strategies may modify the list of steps while the iteration is happening...

jorgebay: ah, I see...

jorgebay:
VOTE +1

It would be good if we could tackle TINKERPOP-2311 before 3.5.0 as it's an API change.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
```diff
@@ -121,34 +121,54 @@ public TraverserGenerator getTraverserGenerator() {
...
+            while (strategyIterator.hasNext()) {
+                final TraversalStrategy<?> strategy = strategyIterator.next();
+                TraversalHelper.applyTraversalRecursively(strategy::apply, this);
```
li-boxuan, line 144: @spmallette When applying it recursively, shouldn't we also do `if (hasGraph) globalChild/localChild.setGraph(this.graph);` like before? Some [JanusGraph tests are failing](https://github.com/JanusGraph/janusgraph/pull/2619#discussion_r633085171) when upgrading to TinkerPop 3.5.0, likely due to this.

spmallette: It's been a long time since i looked at this code and as i go back to it now, i recall wanting to revisit it before release as there still is a bit of a tangle in it that i was never quite satisfied with. that revisiting didn't happen, but i'm not sure it would have raised this issue because we don't seem to have any test assertions to this effect. Other things that bothered me are slowing coming back to memory the more I stare at this. I don't imagine more breaking changes, just some additional refactoring.

While not ideal, is there a possible workaround for you, where you could alter your strategy to get the `Graph` from the parent?  Maybe just do: `Traversal.Admin<?,?> root = TraversalHelper.getRootTraversal(traversal)` for now. I've tested this approach with a contrived strategy for TinkerGraph and it works in that case.

https://issues.apache.org/jira/browse/TINKERPOP-2568

li-boxuan: Yes, thanks, that workaround works.

PR #1214, opened 2019-11-06
TINKERPOP-2235 Expand semantics of null 

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2235

This change makes `null` in traversals behave more consistently and expands its meaning to be a valid value that is not filtered away automatically by the traversal. Please see the upgrade docs for more information on the change as it provides a fair bit of detail. 

This is a breaking change heading to 3.5.0 as traversal semantics sorta shift around a bit. It does also enable the possibility of other interesting changes as described in [TINKERPOP-2312](https://issues.apache.org/jira/browse/TINKERPOP-2312). 

I've tested this manually in the console a fair bit and I'm actually surprised at how well it works without a heavy body of change (despite the size of the PR which is heavily weighted to tests and getting GLVs compliant).

@dkuppitz found some bad things during testing. pulling back my vote until they can be resolved. needs more testing....

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,6 +25,9 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 This release also includes changes from <<release-3-4-3, 3.4.3>>.
 
+* Allowed the possibility for the user of `null` in Gremlin.
+* Added a `Graph.Feature` for `supportsNullPropertyValues`.
+* Modified TinkerGraph to support `null` property values and can be configured to disable that feature.
 * Modified `null` handling in mutations to be consistent for a new `Vertex` as well as update to an existing one.
 * Upgraded to Apache Commons Configuration2.
 * Renamed `StoreStep` to `AggregateLocalStep`.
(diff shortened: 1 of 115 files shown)


spmallette:
@dkuppitz - this is good now:

```text
gremlin> g.inject(10,20,null,20,10,10).groupCount("x").dedup().as("y").project("a","b").by().by(__.select("x").select(__.select("y")))
==>[a:10,b:3]
==>[a:20,b:2]
==>[a:null,b:1]
```

not sure if there are more issues like this or if i even "fixed" it well. hopefully you can find more time to test soon.

@jorgebay i swear i spent almost as much time solving this problem as i did trying to get the c# tests to pass. i had to to this:

https://github.com/apache/tinkerpop/pull/1214/files?file-filters%5B%5D=.cs#diff-c3d769035c1f3cbd767f57d3c9e556c7R40-R41

if you know the fix to that one, please let me know, otherwise it's staying like that for the time being.

jorgebay:
> if you know the fix to that one

Gherkin support in the C# side tries to map text to method calls, with the added difficulty that generics are strict in C# (as opposed to java). IMO It's ok to leave it like this: the C# traversal is OK, is just hard for the test translator to come up with it from text.

dkuppitz:
VOTE +1

PR #1215, opened 2019-11-07
fix unit test failure in non-English environment

youdonghai:
Hello everyone,
I am a gremlin user from China.
When I tried to build tinkerpop, I encountered a problem with the unit test failure. The error log looks like this:
`java.text.ParseException: Unparseable date: "1-Jan-2018"`
This is because my operating system default language is Chinese, the January in Chinese should be "一月".
This pull request can fix this problem.


spmallette:
thanks. This issue is partially resolved for the `tp33` branch on #1188 though it has not merged yet. Let's see what happens there before merging this (as there is an additional change here that only applies to `tp34` and `master` so it still needs to be merged).

spmallette:
merge this now that #1188 is in there.  cherry-picked to `tp34` and merged to `master`

PR #1216, opened 2019-11-07
TINKERPOP-2310 and TINKERPOP-2311 Folllow up refactoring after TraversalStrategy changes

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2310
https://issues.apache.org/jira/browse/TINKERPOP-2311

A few added refactoring opportunities were noticed as a result of TINKERPOP-1568. Basically, `TraversalStrategies` now implements `Iterable` and the semantics of `Traversal.isRoot()` have been narrowed to only check for `EmptyStep` rather than also checking for `null`.

VOTE +1


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Traversal.java
```diff
@@ -480,26 +480,25 @@ public default void reset() {
         /**
          * Set the {@link org.apache.tinkerpop.gremlin.process.traversal.step.TraversalParent} {@link Step} that is
          * the parent of this traversal. Traversals can be nested and this is the means by which the traversal tree is
-         * connected.
+         * connected. If there is no parent, then it should be a {@link EmptyStep}.
```
jorgebay: :+100 for the clarification in the javadoc.

PR #1223, opened 2019-11-13
TINKERPOP-2314 Employ by(String) for Map when possible

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2314

Also improve errors around incorrect types that might be called using this modulator overload. See upgrade docs for reasoning and examples of what's happening here.

VOTE +1


dkuppitz:
Well done.

VOTE +1

PR #1222, opened 2019-11-13
TINKERPOP-1733 Support g.E().properties().hasKey('xx') & hasValue('xx')

javeme:
https://issues.apache.org/jira/browse/TINKERPOP-2313
https://issues.apache.org/jira/browse/TINKERPOP-1733

Currently, an exception will be thrown for this query: `XxProperty cannot be cast to Element`

PR #1224, opened 2019-11-26
TINKERPOP-1733 Support g.E().properties().hasKey('xx') & hasValue('xx')

javeme:
https://issues.apache.org/jira/browse/TINKERPOP-2313
https://issues.apache.org/jira/browse/TINKERPOP-1733
https://issues.apache.org/jira/browse/TINKERPOP-2318

Currently, an exception will be thrown for this query: `XxProperty cannot be cast to Element`

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,8 @@ This release also includes changes from <<release-3-3-10, 3.3.10>>.
 * Improved error messaging for `by(String)` so that it is more clear as to what the problem is
 * Bump to Netty 4.1.42
 * GraphBinary: Introduced our own `Buffer` API as a way to wrap Netty's Buffer API. Moved `GraphBinaryReader`, `GraphBinaryWriter` and `TypeSerializer<T>` to gremlin-core.
+* TINKERPOP-2318 Unify the behavior of property comparison: only compare key&value.
+* TINKERPOP-1733 Support hasKey() and hasValue() step for edge property and meta property, like `g.E().properties().hasKey('xx')`.
 
 [[release-3-4-4]]
 === TinkerPop 3.4.4 (Release Date: October 14, 2019)
--- a/docs/src/upgrade/release-3.4.x.asciidoc
+++ b/docs/src/upgrade/release-3.4.x.asciidoc
@@ -94,6 +94,176 @@ Display stack trace? [yN]n
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-2314[TINKERPOP-2314]
 
+==== hasKey() Step and hasValue() Step
+
+Previously, hasKey() Step and hasValue() Step is only supported applying to vertex property. In order to support more general scenarios, these two steps are currently allowed to be applied to edge property and meta property.
+
+Old behavior of hasKey() / hasValue() with edge property and meta property:
+
+[source,groovy]
+----
+gremlin> graph = TinkerFactory.createTheCrew()
+==>tinkergraph[vertices:6 edges:14]
+gremlin> g = graph.traversal()
+==>graphtraversalsource[tinkergraph[vertices:6 edges:14], standard]
+
+gremlin> g.E().properties().hasKey('since')
+TinkerProperty cannot be cast to Element
+
+gremlin> g.V().properties("location").properties().hasKey("startTime")
+TinkerProperty cannot be cast to Element
+
+gremlin> g.E().properties().hasValue(2010)
+TinkerProperty cannot be cast to Element
+
+gremlin> g.V().properties("location").properties().hasValue(2005)
+TinkerProperty cannot be cast to Element
+----
+
+New behavior of hasKey() with edge property:
+
+[source,groovy]
+----
+gremlin> g.E().properties().hasKey('since')
+==>p[since->2009]
+==>p[since->2010]
+==>p[since->2010]
+==>p[since->2011]
+==>p[since->2012]
+----
+
+New behavior of hasKey() with meta property:
+
+[source,groovy]
+----
+gremlin> g.V().properties().hasKey('location').properties().hasKey('startTime')
+==>p[startTime->1997]
+==>p[startTime->2001]
+==>p[startTime->2004]
+==>p[startTime->2004]
+==>p[startTime->2005]
+==>p[startTime->2005]
+==>p[startTime->1990]
+==>p[startTime->2000]
+==>p[startTime->2006]
+==>p[startTime->2007]
+==>p[startTime->2011]
+==>p[startTime->2014]
+==>p[startTime->1982]
+==>p[startTime->2009]
+----
+
+New behavior of hasValue() with edge property:
+
+[source,groovy]
+----
+gremlin> g.E().properties().hasValue(2010)
+==>p[since->2010]
+==>p[since->2010]
+----
+
+New behavior of hasValue() with meta property:
+
+[source,groovy]
+----
+gremlin> g.V().properties().hasKey('location').properties().hasValue(2005)
+==>p[endTime->2005]
+==>p[endTime->2005]
+==>p[startTime->2005]
+==>p[startTime->2005]
+----
+
+link:https://issues.apache.org/jira/browse/TINKERPOP-1733[TINKERPOP-1733]
+
+==== Properties comparison regardless of parent elements
+
+Unify the behavior of property comparison: two properties are equal only if their key&value are equal, even if their parent elements are not equal. It makes sense when comparing properties regardless of parent elements, just focus on the property itself, it would be more uniform and concise.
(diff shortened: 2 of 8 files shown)


spmallette:
initial discussion for this PR can be found on #1222 - it was rebased to `tp34`

spmallette:
thanks for moving this to `tp34`.

> I've tested the metaproperties case, and it's OK with this patch

👍 

> The Travis issue is caused by HadoopProperty dedup does not work, the root cause is Property comparison will compare the parent elements(here it's edges), but the same edges with two directions are not equal for StarOutEdge and StarInEdge. For TinkerEdge, the same edges with two directions are equal.
> Is it necessary to unify the behavior of edges comparison? like StarEdge and TinkerEdge. I would also be pleasure to submit a patch if necessary.

> The Travis issue is caused by HadoopProperty dedup does not work, the root cause is Property comparison will compare the parent elements(here it's edges), but the same edges with two directions are not equal for StarOutEdge and StarInEdge. For TinkerEdge, the same edges with two directions are equal.
> Is it necessary to unify the behavior of edges comparison? like StarEdge and TinkerEdge. I would also be pleasure to submit a patch if necessary.

sorry - i think i understand but i'm not completely sure. what's an example of the "same edges with two directions"?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/star/StarGraph.java
```diff
@@ -869,6 +870,32 @@ public Vertex inVertex() {
         }
     }
 
+    public final class StarEdgeId {
```
spmallette: This is a bit worrisome. Does this mean that if a user accesses an edge identifier in OLAP they won't get back their edge id, they will instead get a `StarEdgeId`?

javeme:
> what's an example of the "same edges with two directions"?

Such as there are two vertices: `person1` and `person2`, one edge with id "follow_1_2" linked from `person1` to `person2`, it can be expressed like this: `person1 -- follow -> person2`.
When querying edges from source vertex `person1` by `g.V(person1).outE()`, got an edge:
  `person1 -- follow -> person2`,
and querying edges from source vertex `person2` by `g.V(person2).inE()`, got an edge:
  `person2 <- follow -- person1`. 
These two edges are essentially one edge, so it makes sense their comparison should be equal, but it's not equal for StarOutEdge and StarInEdge, no such issue for TinkerEdge.

There seem to be 2 ways to fix OUT & IN StarEdge comparison:

1. Override edge id of StarEdge, like [add StarEdgeId class](https://github.com/apache/tinkerpop/pull/1224/commits/8f604df919bdb9452068eb7ea41273e7aee5b2a7). (but there are still some problems: not registered kryo)
  ```c
org.apache.tinkerpop.gremlin.driver.exception.ResponseException: Error during serialization: java.lang.IllegalArgumentException: Class is not registered: org.apache.tinkerpop.gremlin.structure.util.star.StarGraph$StarEdgeId
Note: To register this class use: kryo.register(org.apache.tinkerpop.gremlin.structure.util.star.StarGraph$StarEdgeId.class);
	at org.apache.tinkerpop.gremlin.server.GremlinResultSetIntegrateTest.shouldHandleTinkerGraphResult(GremlinResultSetIntegrateTest.java:207)
Caused by: org.apache.tinkerpop.gremlin.driver.exception.ResponseException: 
Error during serialization: java.lang.IllegalArgumentException: Class is not registered: org.apache.tinkerpop.gremlin.structure.util.star.StarGraph$StarEdgeId
Note: To register this class use: kryo.register(org.apache.tinkerpop.gremlin.structure.util.star.StarGraph$StarEdgeId.class);
```
2. Override StarEdge equals()

Which way is better, or any other ideas?

**Comment by javeme** [comment]
@spmallette 
Sorry, I made a mistake in the previous analysis, and now correct it: the  [Property comparison](https://github.com/apache/tinkerpop/blob/be4defb541e9ce99712a53200dbe7794c63f06a3/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/ElementHelper.java#L485) issue is not caused by StarEdge, but caused by DetachedProperty. when testing query like `g.V().bothE().properties().dedup()` with spark-gremlin, two identical instances of DetachedProperty with element=null(both null) are not equal, so the dedup() has no effect and there are still duplicate properties.

I have submitted a patch to fix the issue: https://github.com/apache/tinkerpop/pull/1224/commits/612f81457f15f6577daa724f605c5dcfac87f93b

**Comment by dkuppitz** [comment]
@spmallette From the old PR:

> I assume you like where this PR is going. if we make hasKey() behave uniformly for Element and Property where does that leave {{has(String)}}:
> 
> ```
> gremlin> g.V(1).properties().has('name')
> gremlin> g.V(1).properties().hasKey('name')
> ==>vp[name->marko]
> ```

What you show is the expected result, checking for a property existence is different from checking if the key of a keyed element matches a given value. Thus I don't think we can remove/deprecate `has(String)`.

```
gremlin> g.V().has("location")
==>v[1]
==>v[7]
==>v[8]
==>v[9]
gremlin> g.V().properties().hasKey("location").limit(4)
==>vp[location->san diego]
==>vp[location->santa cruz]
==>vp[location->brussels]
==>vp[location->santa fe]
gremlin> g.V().properties().hasKey("location").not(has("endTime"))
==>vp[location->santa fe]
==>vp[location->purcellville]
==>vp[location->seattle]
==>vp[location->aachen]
```

**Comment by dkuppitz** [comment]
At first glance over the PR I was worried because 2 existing tests in `ElementHelperTest` were changed, but these changes totally make sense. Actually, I have no idea why we wrote tests that verify that `null` is not equal to `null` :man_shrugging:

VOTE: +1

**Comment by javeme** [comment]
@dkuppitz Thanks for your review.

@spmallette Currently there is a problem blocking me: different results are returned with the same gremlin `g.E().properties().dedup().value()` in OLTP mode and OLAP mode：
```
OLTP mode: [0.5, 0.4, 1.0, 0.2, 0.4, 1.0]  (return TinkerProperty with element != null)
OLAP mode: [0.5, 1.0, 0.4, 0.2]            (return DetachedProperty with element = null)
```
I'm not sure what's the expected behavior for the following scenarios:
(1) the comparison of properties(of edge) with **element = null**
(2) the comparison of properties(of edge) with **element != null**:

1. Should p1_null and p2_null be equal for the following scenario? (like DetachedProperty from `g.E().properties()` with OLAP mode)
```java
p1_null = {
  key = k1
  value = v1
  element = null
}

p2_null = {
  key = k1
  value = v1
  element = null
}
```
2. Should p1 and p2 be equal for the following scenario? (like TinkerProperty from `g.E().properties()` with OLTP mode)
```java
p1 = {
  key = k1
  value = v1
  element = edge-1
}

p2 = {
  key = k1
  value = v1
  element = edge-2
}
```
I think p1 and p2 are expected to be equal even if their parent elements are not equal. When comparing properties, maybe it is more reasonable to focus on the property itself. This also makes it easier to get uniform behavior.

dkuppitz:
> I think p1 and p2 are expected to be equal even if their parent elements are not equal.

I agree, but I feel like this might cause some issues in other areas in the code...

javeme:
All checks have passed with [areEqual(final Property a, final Object b)](https://github.com/apache/tinkerpop/pull/1224/commits/6457b0fb346b7387516f94f5cca6937e924fca59#diff-302845633b76a969d0fa27dae249e400L485) update.

spmallette:
wow - this is starting to look really concise. nice to see all the tests passing with your changes. that provides some nice confidence. i have a few thoughts at this point:

* Please add some CHANGELOG entries for the changes you've put in place.
* I think this is an important enough change that it requires some upgrade documentation to help summarize for users what they can expect as a result of these changes - you would add them [here](https://github.com/apache/tinkerpop/blob/tp34/docs/src/upgrade/release-3.4.x.asciidoc).
* I assume there are no breaking changes in existing behavior here (aside from behavior that was previously undesired) - please correct me if I'm wrong.

If you're really busy, I can probably add those changes on merge, but those doc fixes would make this PR look really complete to me. Assuming those things are all settled:

VOTE +1

javeme:
@spmallette No problem, I will update CHANGELOG and documentation.

> Please add some CHANGELOG entries for the changes you've put in place.

Should we update this CHANGELOG.asciidoc in section "[[release-3-4-5]]"?

> I assume there are no breaking changes in existing behavior here

For all I know, no breaking changes. To sum up, this changes mainly include two points:

1. Unify the behavior of property comparison: two properties are equal only if their key&value are equal, even if their parent elements are not equal. Note: The "property" here refer to edge property and meta property, excluding vertex property.
2. Support hasKey() and hasValue() step for edge property and meta property.

spmallette:
> Should we update this CHANGELOG.asciidoc in section "[[release-3-4-5]]"?

yes

spmallette:
merged - thanks again @javeme - i look forward to other contributions from you in the future. well done.

javeme:
Ok, my pleasure to contribute.

PR #1226, opened 2019-12-06
Set requestId and userAgent in byte code request when they are available in requestOptions

rdtr:
It is a very small change but without this, even though we write like
```
g.with(Tokens.REQUEST_ID, "some UUID").V()...
```
the UUID won't be sent to server as `requestId` using existing client instance.

I read this thread https://github.com/apache/tinkerpop/pull/1109
and think we should do this because in `submitAsync(string, requestOptions)` we do overwrite requestId if available. Correct me if I am wrong.  I don't specifically need userAgent but just added together. 

Please let me know if I need to add a test regarding this. I couldn't find client related test files.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
@@ -565,6 +565,8 @@ public abstract class Client {
                 // apply settings if they were made available
                 options.getBatchSize().ifPresent(batchSize -> request.add(Tokens.ARGS_BATCH_SIZE, batchSize));
                 options.getTimeout().ifPresent(timeout -> request.add(Tokens.ARGS_EVAL_TIMEOUT, timeout));
+                options.getOverrideRequestId().ifPresent(request::overrideRequestId);
+                options.getUserAgent().ifPresent(userAgent -> request.add(Tokens.ARGS_USER_AGENT, userAgent));
 
                 return submitAsync(request.create());
             } catch (Exception ex) {


divijvaidya:
Thanks for making this change. And congratulations on your first change to this project 🎉

A test would be very helpful as it would catch unexpected regressions in future. You would probably want to add the test here: https://github.com/apache/tinkerpop/blob/master/gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/remote/DriverRemoteConnectionTest.java
OR
add an integration test over here: https://github.com/apache/tinkerpop/blob/master/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java

Also, can you please rebase you change against the 34 branch instead of the master? (since we would want to push this fix out with 3.4.x series)

spmallette:
Something strange has happened. This change is already on tp34. 

https://github.com/apache/tinkerpop/blob/tp34/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java#L346-L347

It's not here though on master for some reason. Where did that commit go and is this the only thing missing? rhetorical question - i'm researching still

spmallette:
oops - my mistake - looking at the wrong method signature. nothing seems missing. i'm surprised that we didn't have this there before though. i wonder how this was missed for all this time.

rdtr:
Thanks @divijvaidya , rebasing onto TP34 introduces a lot of conflicts and I feel it's easier to cherry-pick the change onto TP34 and cut a new PR. Let me do that.

@spmallette Haha, I was also surprised too when I found this today 😛

rdtr:
Created a new one: https://github.com/apache/tinkerpop/pull/1228
Let's continue the discussion there, closing.

PR #1228, opened 2019-12-06
update requestId and userAgent in RequestMessage with RequestOptions

rdtr:
a new PR for https://github.com/apache/tinkerpop/pull/1226

We miss to put requestId and userAgent in bytecode query submission while they are available in requestOptions. This commits fix the issue.

I ran the tests locally and it worked fine. Please let me know if the change any issue. Thanks.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -31,6 +31,7 @@ This release also includes changes from <<release-3-3-10, 3.3.10>>.
 * Introduced internal `Buffer` API as a way to wrap Netty's Buffer API and moved `GraphBinaryReader`, `GraphBinaryWriter` and `TypeSerializer<T>` to `gremlin-core`.
 * Unified the behavior of property comparison: only compare key&value.
 * Supported `hasKey()` and `hasValue()` step for edge property and meta property, like `g.E().properties().hasKey('xx')`.
+* Client sends overrideRequestId and userAgent to Server when they are present in requestOptions with a bytecode query request.
 
 [[release-3-4-4]]
 === TinkerPop 3.4.4 (Release Date: October 14, 2019)
--- a/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
+++ b/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
@@ -71,6 +71,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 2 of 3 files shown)


gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
```diff
@@ -64,13 +64,7 @@
...
 import java.io.File;
 import java.time.Instant;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
spmallette: thanks for the test. it looks like your IDE auto-formatted the imports. please note that we don't use wildcard imports in our code style. could you please revert that change?

spmallette:
Other than the wildcard import and a CHANGELOG entry, I think this PR looks good:

VOTE +1

divijvaidya:
VOTE +1

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
rdtr: Ah you're right, when I added `import java.util.UUID;`, IntelliJ automatically squashed the imports. Reverted it. Thanks !

rdtr:
@spmallette @divijvaidya Thanks for helping me to move this PR forward !

I restored the squashed imports and added a comment to CHANGELOG.asciidoc.
For some reason my addition to the CHANGELOG is detected as a conflict.

Could you resolve the conflict and merge this if the comment I added looks good ?

CHANGELOG.asciidoc
```diff
@@ -28,7 +28,10 @@ This release also includes changes from <<release-3-3-10, 3.3.10>>.
 * Expanded the use of `by(String)` modulator so that it can work on `Map` as well as `Element`.
 * Improved error messaging for `by(String)` so that it is more clear as to what the problem is
 * Bump to Netty 4.1.42
```
rdtr: The last line is the one I want to add in CHANGELOG, but GitHub complains that there is a conflict and I had to add these 3 lines above (GitHub still complains though). I am not clear what should be the right result here.

Please let me know if I need to modify this part.

spmallette: I don't think your PR should contain any lines beyond the last one in this diff. I would just revert your CHANGELOG modification in your branch, rebase the whole PR on tp34 and then add your one line back (and squash) so that it's clean. does that make sense? 

Worst case, if it's taking too much time and you can't figure it out, just revert your CHANGELOG modification, copy/paste your CHANGELOG entry to the comments on this PR and I'll add it manually on merge. thanks.

rdtr: @spmallette Ok, this happened because my `tp34` is behind from `upstream`. I updated `tp34` and rebased this PR onto it, now the conflict is gone. It should be good now.

rdtr:
Not this branch should be good to merge. Let me know if further tweaks are needed.

PR #1230, opened 2019-12-13
TINKERPOP-2320 allow to pass custom XmlInputFactory when instantiating GraphMLReader

rdtr:
Some provider wants to use XMLInputFactory with more secure configurations. This change makes it possible to pass XMLInputFactory when instantiating GraphMLReader.

I don't add any tests right now, I want to first confirm if this direction is OK. If yes, please suggest any tests that I need to add. Thanks !


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphml/GraphMLReader.java
```diff
@@ -343,6 +344,7 @@ public static Builder build() {
         private String vertexLabelKey = GraphMLTokens.LABEL_V;
         private boolean strict = true;
         private long batchSize = 10000;
+        private XMLInputFactory inputFactory = XMLInputFactory.newInstance();
```
rdtr: It is wasteful to always instantiate a default inputFactory considering it may not be used. Alternatively I can write in create method like:

```
        public GraphMLReader create() {
            if (this.inputFactory == null) {
                this.inputFactory = XMLInputFactory.newInstance();
            }
            return new GraphMLReader(this);
        }
```

spmallette:
I think this approach is fine. Please target the `tp33` branch so that this change goes back to 3.3.x (since this is a security issue). To make this PR complete, please consider a CHANGELOG entry, though I can add that on merge if needed. I think I'd also like to see you change the instantiation of the `XMLInputFactory` as you suggested in your comment.

Assuming all that is done:

VOTE +1

rdtr:
Sorry, rebasing directly on Github UI seems not a good solution to base onto `tp34`... I will make another PR. Thanks.

rdtr:
I just created a new PR: https://github.com/apache/tinkerpop/pull/1235
Let's continue the discussion there. Thanks!

PR #1233, opened 2019-12-16
TINKERPOP-2315 Implement clone() for all GLVs

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2315

Hard to believe we didn't have `clone()` in place across the board. There might have been some more idiomatic ways to implements this per language but I decided to stick with `clone()` as it is in Java for all languages. Just felt like it should be consistent. Happy to hear other opinions on that matter, but besides that:

VOTE +1

### DIFF
--- a/gremlin-javascript/glv/GraphTraversalSource.template
+++ b/gremlin-javascript/glv/GraphTraversalSource.template
@@ -113,6 +113,14 @@ class GraphTraversal extends Traversal {
   constructor(graph, traversalStrategies, bytecode) {
     super(graph, traversalStrategies, bytecode);
   }
+
+  /**
+   * Copy a traversal so as to reset and re-use it.
+   */
+  clone() {
+    return new GraphTraversal(this.graph, this.traversalStrategies, this.getBytecode());
+  }
+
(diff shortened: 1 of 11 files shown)


gremlin-javascript/glv/GraphTraversalSource.template
```diff
@@ -113,6 +113,14 @@ class GraphTraversal extends Traversal {
   constructor(graph, traversalStrategies, bytecode) {
     super(graph, traversalStrategies, bytecode);
   }
+
+  /**
+   * Copy a traversal so as to reset and re-use it.
+   */
+  clone() {
+    return new GraphTraversal(this.graph, this.traversalStrategies, this.getBytecode());
```
jonyfSaiflow, line 121: @spmallette 
This still references the same `graph`, `traversalStrategies`, and `bytecode` of `this`. 
This clones but will affect the cloned and original.
Not sure where to post an issue

spmallette: issues/bugs are tracked in JIRA: https://issues.apache.org/jira/projects/TINKERPOP/

PR #1235, opened 2019-12-16
TINKERPOP-2320 allow to pass custom XmlInputFactory when instantiating GraphMLReader

rdtr:
This pull request is a revised one from https://github.com/apache/tinkerpop/pull/1230.

Some provider wants to use XMLInputFactory with more secure configurations. This change makes it possible to pass XMLInputFactory when instantiating GraphMLReader.

I don't add any tests right now, I want to first confirm if this direction is OK. If yes, please suggest any tests that I need to add. Thanks !

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -27,6 +27,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Deprecated `TraversalStrategies.applyStrategies()`.
 * Deprecated Jython support in `gremlin-python`.
 * Reverted: Modified Java driver to use IP address rather than hostname to create connections.
+* Allow custom XMLInputFactory to be used with GraphMLReader.
 
 [[release-3-3-9]]
 === TinkerPop 3.3.9 (Release Date: October 14, 2019)
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphml/GraphMLReader.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphml/GraphMLReader.java
@@ -56,18 +56,19 @@ import java.util.stream.Stream;
  * @author Stephen Mallette (http://stephen.genoprime.com)
  */
 public final class GraphMLReader implements GraphReader {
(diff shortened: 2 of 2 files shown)


CHANGELOG.asciidoc
```diff
@@ -27,6 +27,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Deprecated `TraversalStrategies.applyStrategies()`.
 * Deprecated Jython support in `gremlin-python`.
 * Reverted: Modified Java driver to use IP address rather than hostname to create connections.
+* TINKERPOP-2320 Allow custom XMLInputFactory to be used with GraphMLReader.
```
spmallette: No need to add the JIRA number - those get added separately on release.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphml/GraphMLReader.java
```diff
@@ -381,7 +383,18 @@ public Builder batchSize(final long batchSize) {
...
+        /**
+         * the key to use as the inputFactory when a caller wants to pass XMLInputFactory with its own configuration.
+         */
+        public Builder xmlInputFactory(final XMLInputFactory inputFactory) {
+            this.inputFactory = inputFactory;
+            return this;
+        }
+
         public GraphMLReader create() {
+            if (this.xmlInputFactory() == null) {
```
spmallette: does this line compile? `xmlInputFactory` isn't a method is it? shouldn't you remove those parenthesis and just do: `if (this.xmlInputFactory == null) {`

CHANGELOG.asciidoc
rdtr: OK, removed it. Thanks!

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphml/GraphMLReader.java
rdtr: Sorry I was lazy... this should have been
```
if (this.inputFactory == null)
```
I updated the PR. Thanks!

spmallette:
VOTE +1

PR #1237, opened 2020-01-13
TINKERPOP-2325 - Modifications for use of indexes in FILTER

thevalzo:
I modified only the class "WhereTrafersalBuilder" and some row in SparqlToGremlinCompiler in respect to the branch master.
Please note that my optimization tecnique involves looking into all source triples for creating a more concise one. For this reason it has to be evaluated if this kind of optimization has to be done in this classes (where formally you have to translate the single triple in the single traversal step) or in the SparqlTraversalStrategy that optimize the whole traversal.

### DIFF
--- a/sparql-gremlin/src/main/java/org/apache/tinkerpop/gremlin/sparql/WhereTraversalBuilder.java
+++ b/sparql-gremlin/src/main/java/org/apache/tinkerpop/gremlin/sparql/WhereTraversalBuilder.java
@@ -47,67 +47,155 @@ class WhereTraversalBuilder {
     /**
      * Converts a general {@code Expr} to an anonymous {@link GraphTraversal}.
      */
-    static GraphTraversal<?, ?> transform(final Expr expression) {
-        if (expression instanceof E_Equals) return transform((E_Equals) expression);
-        if (expression instanceof E_NotEquals) return transform((E_NotEquals) expression);
-        if (expression instanceof E_LessThan) return transform((E_LessThan) expression);
-        if (expression instanceof E_LessThanOrEqual) return transform((E_LessThanOrEqual) expression);
-        if (expression instanceof E_GreaterThan) return transform((E_GreaterThan) expression);
-        if (expression instanceof E_GreaterThanOrEqual) return transform((E_GreaterThanOrEqual) expression);
-        if (expression instanceof E_LogicalAnd) return transform((E_LogicalAnd) expression);
-        if (expression instanceof E_LogicalOr) return transform((E_LogicalOr) expression);
-        if (expression instanceof E_Exists) return transform((E_Exists) expression);
-        if (expression instanceof E_NotExists) return transform((E_NotExists) expression);
+    public static GraphTraversal<?, ?> transform(final Expr expression, List<Triple> triples) {
+        if (expression instanceof E_Equals) return transform((E_Equals) expression, triples);
+        if (expression instanceof E_NotEquals) return transform((E_NotEquals) expression, triples);
+        if (expression instanceof E_LessThan) return transform((E_LessThan) expression, triples);
+        if (expression instanceof E_LessThanOrEqual) return transform((E_LessThanOrEqual) expression, triples);
+        if (expression instanceof E_GreaterThan) return transform((E_GreaterThan) expression, triples);
+        if (expression instanceof E_GreaterThanOrEqual) return transform((E_GreaterThanOrEqual) expression, triples);
+        if (expression instanceof E_LogicalAnd) return transform((E_LogicalAnd) expression, triples);
+        if (expression instanceof E_LogicalOr) return transform((E_LogicalOr) expression, triples);
+        if (expression instanceof E_Exists) return transform((E_Exists) expression, triples);
+        if (expression instanceof E_NotExists) return transform((E_NotExists) expression, triples);
         throw new IllegalStateException(String.format("Unhandled expression: %s", expression));
     }
 
-    private static GraphTraversal<?, ?> transform(final E_Equals expression) {
-        final Object value = expression.getArg2().getConstant().getNode().getLiteralValue();
-        return __.as(expression.getArg1().getVarName()).is(P.eq(value));
+   public static GraphTraversal<?, ?> transform(final E_Equals e, List<Triple> triples) {
+        GraphTraversal traversal = null;
+        //---MODIFIED FOR BETTER INDICES USE
+        //For the index optimization you have to find the original BGP where the Filter expression is applied, so you can create a new
+        //concise BGP easy to optimize for indexes 
+        for(final Triple triple : triples){
+            String object = triple.getObject().getName().toString();
+            String subject = triple.getSubject().getName().toString();
+            String uri = Prefixes.getURIValue(triple.getPredicate().getURI());
+            String arg1 = e.getArg1().getExprVar().getVarName();
+            
+            final Object value =  e.getArg2().getConstant().getNode().getLiteralValue();
+
+            if (object.equals(arg1)){
+                traversal = __.as(subject).has(uri, P.eq(value));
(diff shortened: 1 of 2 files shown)


spmallette:
Thank you for submitting this but note that Travis cannot build this PR:

https://travis-ci.org/apache/tinkerpop/jobs/636394809?utm_medium=notification&utm_source=github_status

Also, I'd like to see some examples of what the improvement does - specifically, could you add one or more tests that will trigger this new code and could you please paste a something here in the github comments that shows how the SPARQL compiled to Gremlin before and after this change. I think we will roll your example (and mine) into the upgrade documentation.

thevalzo:
Once I succesfully compile with mvn clean install, how can I load the sparql-gremlin package into the console? ":install org.apache.tinkerpop sparql-gremlin 3.5.0" doesn't work

spmallette:
well, 3.5.0 isn't published yet but more directly you are targetting the 3.4.x branch which means that `mvn clean install` is publishing "3.4.5-SNAPSHOT" locally:

https://github.com/apache/tinkerpop/blob/035d14b7bde1e12ce8892ddab78b12aae202c624/pom.xml#L28

you should use your `:install` command with that version.

spmallette:
@thevalzo i just thought that i'd alert you to the fact that we will be preparing for release of 3.4.5 next week. if you'd like to see this change in that version we would need to sort though this pull request by the end of this week. i'm not trying to pressure you of course...i just wanted to let you know about our release schedule in case this change was important to you and you needed it in the next possible release.

thevalzo:
Here we go, thanks @spmallette for your kind support. I haven't added more tests because the FILTER operator is already included in the test cases. In the attached images you can see how the traversal is changed, from a WhereStep clause to a HasStep, which can be optimized for the use of indexes.
original-sparql-gremlin
modified-sparql-gremlin

thevalzo:
@spmallette do you have ideas of the reason for which i'm getting a Connection reset error downloading a dependency for a plugin I didn't modify? I did not have this error on my machine.

spmallette:
that could be a travis environmental inconsistency - i restarted the job. let's see if it passes.

sparql-gremlin/src/main/java/org/apache/tinkerpop/gremlin/sparql/WhereTraversalBuilder.java
```diff
@@ -47,79 +49,248 @@
...
+        if (expression instanceof E_NotExists) return transform((E_NotExists) expression, triples);
         throw new IllegalStateException(String.format("Unhandled expression: %s", expression));
     }
 
-    private static GraphTraversal<?, ?> transform(final E_Equals expression) {
-        final Object value = expression.getArg2().getConstant().getNode().getLiteralValue();
-        return __.as(expression.getArg1().getVarName()).is(P.eq(value));
+   public static GraphTraversal<?, ?> transform(final E_Equals e, List<Triple> triples) {
+        GraphTraversal traversal = null;
+         for(final Triple triple : triples){
```
spmallette: I'm not terribly familiar with the APIs at play here so I'm not sure what to expect - if the "triples" `List` has multiple `Triple` objects in it, this function will only return a `GraphTraversal` using the last one and the others will be ignored. Is that ok?

thevalzo: In this function I iterate over "triples" to search for the variable used in FILTER condition, ( ?b in my example) so I can replace it with the label (name). Once i found it, i can discard the others triples. Maybe  I could add a breack instruction in the last IF for clarity.

spmallette: seems like you could just:

```java
            if (object.equals(arg1)){
                return  __.as(subject).has(uri, P.eq(value));
            }
```

and just return `null` at the end assuming nothing was built?

thevalzo: Yes, this seems a better solution.

spmallette:
Thank you for this pull request along with your patience/help in getting it merged. This was a very important fix. Please note that I did refactor the code a bit on merge and did write some upgrade docs as I want to make sure that this improvement gets promoted on release:

https://github.com/apache/tinkerpop/commit/7ea3dfadcc48bb2e2c66ca8285995e91cc7287a6

It would be great if you had other ideas for improving sparql-gremlin - I'd be happy to see another pull request in this area.

thevalzo:
I thank you for your precious support too, it has been a pleasure to contribute to this growing project. In the next future I don't think I will have occasion to work again on this, otherwise surely I will do again a pull request. I'll just have you know as you close the Jira ticket that the index optimization isn't complete yet, as other SPARQL operators like UNION, AND, OR remained unoptimized.

spmallette:
i'm sure there are other optimizations that remain to be added. we will simply encourage new tickets to be created for new specific fixes. thanks again

spmallette:
I'm sorry but I've had to revert this PR in preparation for release. I noticed documentation generation failures for sparql-gremlin which pointed to a problem with AND. When you had said that AND wasn't optimized I didn't take that to mean that it wasn't working at all and clearly we didn't have tests in place for that. I've added a test along with the reverting of the change:

https://github.com/apache/tinkerpop/commit/d67f7c9c757bfeeaedec4d6650640c1cc4b0aed2

I did try to fix the problem first prior to reverting, but given how close we are to release (Monday) I didn't think I was going to easily solve the problem so it was simply safer to revert. In doing that investigation though I do now better understand what these changes were doing. As I wrote in the commits on the above commit, It made me wonder if we should try to simply optimize `match()` better so that it gives providers a better chance at using an index...in that way we get improvements to `match()` for native usage and to sparql-gremlin and other language compilers. As I mentioned previously, we'll create new JIRA issues as needed for these improvements going forward.

thevalzo:
I didn't mean that AND wasn't working at all, but that it wasn't using indexes .. in fact in my tests FILTER { .. AND .. } were working. Which error it gives you?

spmallette:
i get:

```text
gremlin> g.sparql("""SELECT DISTINCT ?name
......1>          WHERE {
......2>            ?person v:label "person" .
......3>            ?person v:age ?age .
......4>            ?person e:created ?project .
......5>            ?project v:name ?name .
......6>            ?project v:lang ?lang .
......7>              FILTER (?age > 30 && ?lang = "java") }""")
The provided match pattern is unsolvable: [[MatchStartStep(person), HasStep([age.gt(30)]), MatchEndStep], [MatchStartStep(project), HasStep([lang.eq(java)]), MatchEndStep]]
Display stack trace? [yN]n
```

thevalzo:
Hi Stephen! As you said, I agree that this would be the key to perform index lookups with sparql-gremlin. Now I'm working again with Sparql-gremlin, so I would like to ask you if in this time it has been done or it is planned to do some optimization to the match() step.

spmallette:
Nothing in sparql-gremlin has changed since this issue though I did have to revert some of this change:

https://github.com/apache/tinkerpop/commit/d67f7c9c757bfeeaedec4d6650640c1cc4b0aed2#diff-73f3fdfa739b38b73256f2e3732bef23

thevalzo:
Yes, but you said previously:
"It made me wonder if we should try to simply optimize match() better so that it gives providers a better chance at using an index...in that way we get improvements to match() for native usage and to sparql-gremlin and other language compilers. As I mentioned previously, we'll create new JIRA issues as needed for these improvements going forward."

Since the index optimization of Gremlin's match() improve both native usage and sparql-gremlin, I was wondering if something has been done about it.

spmallette:
oh - i understand better now. no, nothing has been done yet to improve the `match()` algorithm or strategies related to optimizing the step itself.

PR #1239, opened 2020-01-16
TINKERPOP-2329 JavaScript GLV: upgrade ws dependency

jorgebay:
Update dependencies as long as those work with Node.js 6+ runtime.

https://issues.apache.org/jira/browse/TINKERPOP-2329

VOTE +1


jorgebay:
Also, I wanted to start a thread on the mailing list to discuss runtime version support policy (for Node.js but could apply to techs), I hope I have time in the next couple of days to do it.

spmallette:
VOTE +1

PR #1244, opened 2020-02-06
Add trustStoreType such that keystore and truststore can be different type

justinchuch:
This PR adds a little bit more flexibility to the usage of different store types between keystore and truststore. For example, `jks` in keystore and `pkcs12` in truststore.

It preserves the original behavior, such that if the user does not define the `trustStoreType`, it would use the `keyStoreType`.

----
- Ran `mvn clean install` successfully
- Ran `docker/build.sh -i -t` successfully


spmallette:
As a quick note @justinchuch given that we are trying to quickly turn around a 3.4.6 release to fix a bug in 3.4.5:

PR #1247, opened 2020-02-15
TINKERPOP-2192 Null checks added to Connection.Parse

Haapsaari-Juha:
https://issues.apache.org/jira/browse/TINKERPOP-2192

Added null check to Gremlin.Net.Driver.Connection.Parse method for receivedMsg variable and responseHandler. My first pull request to public repo, hope I did it correctly

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
@@ -124,9 +124,9 @@ namespace Gremlin.Net.Driver
             }
             catch (Exception e)
             {
-                if (_callbackByRequestId.TryRemove(receivedMsg.RequestId, out var responseHandler))
+                if (receivedMsg != null && _callbackByRequestId.TryRemove(receivedMsg.RequestId, out var responseHandler))
                 {
-                    responseHandler.HandleFailure(e);
+                    responseHandler?.HandleFailure(e);
                 }
             }
         }


spmallette:
i have no clue why docker is failing on the gremlin-python portion of the build. works fine for me locally - it's just doing `docker/build.sh` with no flags. not sure if anyone else can get it to fail locally...

spmallette:
As a quick note @Haapsaari-Juha given that we are trying to quickly turn around a 3.4.6 release to fix a bug in 3.4.5:

dzmitry-lahoda:
Change looks pretty safe. Event check for `null` in message is kind of reasonable, because  could return null https://github.com/JamesNK/Newtonsoft.Json/blob/a31156e90a14038872f54eb60ff0e9676ca4a0d8/Src/Newtonsoft.Json/JsonConvert.cs#L823 . Also at same time not sure if it hides bad JSON setting or other json related behind.  Should we fail totally in this case?

FlorianHockmann:
> Also at same time not sure if it hides bad JSON setting or other json related behind. Should we fail totally in this case?

We would still get an error in that case [from here](https://github.com/apache/tinkerpop/blob/master/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs#L136) where the `receivedMsg` is also accessed without a null check. If `receivedMsg` is `null` then `TryParseResponseMessage` will throw a `NullReferenceException` which will be caught by the try/catch in `Parse` and we notify the response handler about the exception.
I think that we should however ensure that we're not getting an exception in a `catch` clause.

dzmitry-lahoda:
>> Parse and we notify the response handler about the exception.

Will not because message is null. There there is silent error eating. 

Could we just check that deserialized object is not null and throw exception.

```
       var receivedMsg = _messageSerializer.DeserializeMessage<ResponseMessage<JToken>>(received);
            if (receivedMsg == null)
            {
                throw new Exception("receivedMsg");
            }
```

To check what happens in case of null array passed, in case of empty array passed, and do property based testing for any other array passed. So it will be more explainable failure than just eating null.

FlorianHockmann:
> Will not because message is null. There there is silent error eating.

You're right. I should have looked taken a closer look.

> Could we just check that deserialized object is not null and throw exception.

Yes, I think that might be the best option as we expect the message to never be null. So, if for some reason is actually null, then we can throw an exception which will result in a call to `CloseConnectionBecauseOfFailureAsync` which closes the connection completely and notifies all existing response handlers about the exception.

@Haapsaari-Juha: I don't know if you want to incorporate these changes into your PR as you were the first to contribute a fix for this issue or if @dzmitry-lahoda should update his PR accordingly.

> property based testing for any other array passed

Could you explain a bit further what you mean by this? Do you want to add property based testing for the internal classes of the driver?

dzmitry-lahoda:
@FlorianHockmann  I did change in my PR. Please review.

> property based testing

Given we do not have fix which only improves situation (like my PR), we would have to prove.

For this:
0. Check latest Json library and see it returns `object?` - so null is possible. DONE
1. Test with byte array being null, empty. What error we see?
2. Test byte array contains string "null", "", "[]", "{}", not null terminated string, etc.
3. Run generator of byte arrays with https://github.com/haf/expecto#property-based-tests (in C#)

Check that we never ever get null reference in this case. Or in some case we do. Than we should look into Json or into layer creating byte array. And have better semantics for error. 

But this seems no needed for now.

spmallette:
closing in favor of #1250

PR #1248, opened 2020-02-15
Fix to iOS platform

Haapsaari-Juha:
Changed Gremlin.Net.Structure.IO.GraphSON.GraphSONWriter.TryGetSerializerFor method to not use out parameter because iOS platform cannot handle these. Original TryGetSerializerFor(out IGraphSONSerializer serializer, type) causes System.NotImplementedException: byref delegate System.Linq.Expressions.Interpreter.LightLambda.CreateCustomDelegate exception on iOS platform

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/GraphSONWriter.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/GraphSONWriter.cs
@@ -114,7 +114,10 @@ namespace Gremlin.Net.Structure.IO.GraphSON
         public dynamic ToDict(dynamic objectData)
         {
             var type = objectData.GetType();
-            if (TryGetSerializerFor(out IGraphSONSerializer serializer, type))
+
+            IGraphSONSerializer serializer = TryGetSerializerFor(type);
+
+            if (serializer != null)
                 return serializer.Dictify(objectData, this);
             if (IsDictionaryType(type))
                 return DictToGraphSONDict(objectData);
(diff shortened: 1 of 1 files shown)


FlorianHockmann:
Do you have a reference for the iOS platform not being able to handle out parameters? I never heard of this (but I have also never worked on iOS so that doesn't have to say much).
It would be good to know in general whether we need to avoid certain C# features because they aren't supported on some platforms.

Haapsaari-Juha:
I'm using Xamarin Forms platform and they have limitation on iOS documented here: https://docs.microsoft.com/en-us/xamarin/ios/internals/limitations

But this code is not using any dynamic code so when digging a bit deeper in to the problem it might also be caused of this: https://github.com/dotnet/runtime/issues/31075

So issue in dotnet corefx might cause this. I don't know why CreateCustomDelegate method is called when using out IGraphSONSerializer in the code but this might be the source for the exception. But same code without using out parameter works. Could not find any reference that would say this is not allowed in iOS and I'm using out parameters with types like decimal, int, string etc. in iOS without any problems.

FlorianHockmann:
OK, thanks for the explanation. It's of course unfortunate that this doesn't work on Xamarin for some reason, but it's only a stylistic change in an internal API, so I'm ok with this.

VOTE +1

spmallette:
As a quick note @Haapsaari-Juha given that we are trying to quickly turn around a 3.4.6 release to fix a bug in 3.4.5:

spmallette:
> but it's only a stylistic change in an internal API

agree with that - @FlorianHockmann do you have time to handle merging this one?

FlorianHockmann:
> do you have time to handle merging this one?

Sure, thanks for pinging me again here. I just merged this one.

PR #1251, opened 2020-02-19
Add session mode in js driver

heljoyLiu:


heljoyLiu:
@greedy , could we set a new version node, v7.6 or other for travis-ci ? 

I found async/await failed.

image

heljoyLiu:
now, syntax is ok, but ThinkerGraph don't support transaction.

how about to skip this session test case ? I check it works in local graph which support transaction

spmallette:
We do test transactions but they only test when Neo4j is present and we only test those with Java test cases. The nature of a session and transactions within a session are already well tested that way. I think that here for a GLV you just want to ensure that the most basic aspect of a session works. The most basic tests to assert that I think would be to create a variable in one request like "x = 1" then in a second request send "x = x + 1" and assert that you get "2" as a result. you may wish to do that in a loop for multiple requests just to demonstrate the ongoing modifications to the session variable "x".

heljoyLiu:
> We do test transactions but they only test when Neo4j is present and we only test those with Java test cases. The nature of a session and transactions within a session are already well tested that way. I think that here for a GLV you just want to ensure that the most basic aspect of a session works. The most basic tests to assert that I think would be to create a variable in one request like "x = 1" then in a second request send "x = x + 1" and assert that you get "2" as a result. you may wish to do that in a loop for multiple requests just to demonstrate the ongoing modifications to the session variable "x".

en, get your point, I will change tests for session, thanks.

heljoyLiu:
Hi  @spmallette @jorgebay , I copy a sample case from java session, and it's basic aspect, thanks

spmallette:
VOTE +1

heljoyLiu:
> Looks great! thanks @heljoyLiu for including integration tests!
> 
> VOTE +1

please apply it as free, thanks

heljoyLiu:
@jorgebay hi, is here some wrong ? I will update it asap

jorgebay:
Sorry for the radio silence, I haven't got time to merge it yet... I hope I can do it soon.

spmallette:
I've merged this one and added some documentation. Thanks @heljoyLiu

heljoyLiu:
> I've merged this one and added some documentation. Thanks @heljoyLiu

Thanks @spmallette for your documentation, I will add session for other Gremlin Language Variants later as above

jorgebay:
Thanks @heljoyLiu !!! 🎉

PR #1252, opened 2020-02-24
improve some of TINKERPOP-2288

dzmitry-lahoda:
will fix some cases of `Get ConnectionPoolBusyException and then ServerUnavailableExceptions`


FlorianHockmann:
Thanks for this improvement. It also really addresses a part of TINKERPOP-2288 as it makes it easier to understand that the server is not overloaded but not reachable.

Could you change the target branch to `3.4-dev`? That way the change will land in the next 3.4.z release as well as 3.5.0.

dzmitry-lahoda:
https://github.com/apache/tinkerpop/pull/1254

PR #1257, opened 2020-03-04
dotnet: add session connection

heljoyLiu:


heljoyLiu:
Hi @spmallette  @jorgebay ,  I make this patch like gremlin-javascript, let me know if any, thanks

FlorianHockmann:
Thanks for the contribution, @heljoyLiu! The PR looks good at a first glance, but I'll need some more time to do a full review.
I'm not sure if you're aware of that but the Gremlin.NET driver has changed a lot between `3.3-dev` and `3.4-dev`, so if you want this change to go into `3.3-dev` already, then it would be good if you could create another PR that targets `3.4-dev` with this contribution. That would make it a lot easier to get your contribution into all release branches.

heljoyLiu:
> Thanks for the contribution, @heljoyLiu! The PR looks good at a first glance, but I'll need some more time to do a full review.
> I'm not sure if you're aware of that but the Gremlin.NET driver has changed a lot between `3.3-dev` and `3.4-dev`, so if you want this change to go into `3.3-dev` already, then it would be good if you could create another PR that targets `3.4-dev` with this contribution. That would make it a lot easier to get your contribution into all release branches.

ok, I will create another PR to `3.4-dev` after review, thanks @FlorianHockmann

gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
```diff
@@ -68,18 +71,54 @@ public async Task ConnectAsync()
 
         public async Task CloseAsync()
         {
+            if (!String.IsNullOrEmpty(_sessionId))
```
FlorianHockmann: (nitpick) **s**tring (it's just the usual C# code style)

gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
```diff
@@ -68,18 +71,54 @@ public async Task ConnectAsync()
...
+            }
             await _webSocketConnection.CloseAsync().ConfigureAwait(false);
         }
 
+        private async Task closeSession()
+        {
+            var msgBuilder = RequestMessage.Build(Tokens.OpsClose).Processor(Tokens.ProcessorSession);
+            msgBuilder.AddArgument(Tokens.ArgsGremlin, "session.close()").AddArgument("session", _sessionId);
+
+            var graphsonMsg = _graphSONWriter.WriteObject(msgBuilder.Create());
```
FlorianHockmann: Any reason why you didn't simply call `SendAsync` here? I think it would be good if we wouldn't have to duplicate the code to serialize and send a message.

gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
```diff
@@ -68,18 +71,54 @@ public async Task ConnectAsync()
...
+            await _webSocketConnection.SendMessageAsync(serializedMsg).ConfigureAwait(false);
+        }
+
         public bool IsOpen => _webSocketConnection.IsOpen;
 
         private async Task SendAsync(RequestMessage message)
         {
-            var graphsonMsg = _graphSONWriter.WriteObject(message);
+            var msg = message;
+            if (!String.IsNullOrEmpty(_sessionId))
```
FlorianHockmann: (nitpick) Again: **s**tring

gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
```diff
@@ -68,18 +71,54 @@ public async Task ConnectAsync()
...
+            {
+                await closeSession().ConfigureAwait(false);
+            }
             await _webSocketConnection.CloseAsync().ConfigureAwait(false);
         }
 
+        private async Task closeSession()
+        {
+            var msgBuilder = RequestMessage.Build(Tokens.OpsClose).Processor(Tokens.ProcessorSession);
+            msgBuilder.AddArgument(Tokens.ArgsGremlin, "session.close()").AddArgument("session", _sessionId);
```
FlorianHockmann: Please add a new token like `Tokens.ArgsSession` here instead of hard-coding the string here to stay consistent.

gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
FlorianHockmann: One other idea I just got, how about adding a private property like this:

```c#
private bool IsSessionEnabled => !string.IsNullOrEmpty(_sessionId);
```

then this if clause could be changed to:

```c#
if (IsSessionEnabled)
```

which is easer to understand in my opinion.
But this if course only a suggestion. Ignore it if you disagree.

gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
FlorianHockmann: Also, is the `gremlin` argument really necessary here? I don't see it being used in [the docs of session close](http://tinkerpop.apache.org/docs/3.4.6/dev/io/#_session_close_2). If it's needed, then I would suggest that we add the value also as a token maybe as something like `Tokens.DummyValue` with a string value and XML comment making it clear that the value doesn't matter.

gremlin-dotnet/src/Gremlin.Net/Driver/GremlinClient.cs
```diff
@@ -58,14 +58,15 @@ public class GremlinClient : IGremlinClient
         ///     A delegate that will be invoked with the <see cref="ClientWebSocketOptions" />
         ///     object used to configure WebSocket connections.
         /// </param>
+        /// <param name="sessionId">The session Id if Gremlin Client in session mode, defaults to null as session-less Client.</param>
         public GremlinClient(GremlinServer gremlinServer, GraphSONReader graphSONReader = null,
             GraphSONWriter graphSONWriter = null, string mimeType = null,
-            Action<ClientWebSocketOptions> webSocketConfiguration = null)
+            Action<ClientWebSocketOptions> webSocketConfiguration = null, string sessionId = null)
```
FlorianHockmann: Any reason why this isn't of type `Guid?` instead of a `string`? [It looks](http://tinkerpop.apache.org/docs/3.4.6/dev/io/#_session_eval_3) to me like this should be a `Guid`:

```js
"session" : {
  "@type" : "g:UUID",
  "@value" : "41d2e28a-20a4-4ab0-b379-d810dede3786"
}
```

heljoyLiu: I checked sessionId in document, here is `Guid`, but String request [in code](https://github.com/apache/tinkerpop/blob/master/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/session/SessionOpProcessor.java#L210), and it fails when `Guid` in args

```
[WARN] OpExecutorHandler - java.util.UUID cannot be cast to java.lang.String
java.lang.ClassCastException: java.util.UUID cannot be cast to java.lang.String
	at org.apache.tinkerpop.gremlin.server.op.session.SessionOpProcessor.getSession(SessionOpProcessor.java:203)
```
anything, we should update document or add `UUID` support in `gremlin-server`

gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
heljoyLiu: thanks, I add "session().close()" as tips for debug things,  the processor will ignore it. I will add comment for this.

gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
heljoyLiu: ok, thanks

gremlin-dotnet/src/Gremlin.Net/Driver/GremlinClient.cs
FlorianHockmann: That's strange. @spmallette, can you provide some insight on why this doesn't work with a sessionId as `UUID`/`GUID` although it's documented like that?

spmallette: The session id is a string but you often see a UUID string for convenience.

FlorianHockmann: But it's serialized as `g:UUID` in GraphSON in our docs [here](http://tinkerpop.apache.org/docs/3.4.6/dev/io/#_session_eval_3). Is that an error in the docs or am I missing something here?
But OK, that means right now for this PR that the type used is correct.

spmallette: that must be an error i the docs...unless there was a time when we accepted `String` or `UUID` types, but i don't think we did. it looks like our IO test files are bad too because those docs basically come from those files. since it's more than just a documentation change, i will go and correct all that. thanks for pointing this issue out.

gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
```diff
@@ -68,18 +76,53 @@ public async Task ConnectAsync()
...
+
+            await SendAsync(msg).ConfigureAwait(false);
+        }
+
         public bool IsOpen => _webSocketConnection.IsOpen;
 
         private async Task SendAsync(RequestMessage message)
         {
-            var graphsonMsg = _graphSONWriter.WriteObject(message);
+            var msg = message;
```
FlorianHockmann: (nitpick) This shouldn't be necessary as you can also assign `message` again below:

```c#
message = RebuildSessionMessage(message);
```

heljoyLiu: thanks, got it

heljoyLiu:
thanks, @FlorianHockmann @spmallette

heljoyLiu:
@FlorianHockmann , I will prepare another PR for `3.4-dev` , pls let me know if any wrong, thanks.

FlorianHockmann:
> I will prepare another PR for 3.4-dev , pls let me know if any wrong, thanks.

That sounds good, thanks. Gremlin.NET hasn't changed much between `3.4-dev` and `master` so we can merge that PR subsequently also into `master` and then have the change in all release branches.

heljoyLiu:
new feature PR will apply to `3.4-dev` and I create one for, close it.
thanks, @FlorianHockmann @spmallette

spmallette:
I've merged this PR even though it was closed. This change did need to go to `3.3-dev` because we need session functionality on that branch. I added a change to remove the non-standard argument from the "close" message:

https://github.com/apache/tinkerpop/commit/fb334df4845d6fef0258d12f9da08f36efae50f5

I merged `3.3-dev` to `3.4-dev` with `-s ours` to ignore these changes as they were already brought in on #1263 and #1276. I also brought in the test improvement manually to `3.4-dev` that i had added to `3.3-dev` in the above commit manually with:

https://github.com/apache/tinkerpop/commit/183b2c9e1bc4fb448b96bbe419b0bc0fca68eab7

I think we just need to merge #1275 at this point and all GLVs will be consistent on `3.3-dev` and `3.4-dev`. I expect to do a more full review of docs and testing at that point and then go on to consider what is necessary to make `master` wholly consistent. 

thanks @heljoyLiu for getting things this far.

heljoyLiu:
>  I expect to do a more full review of docs and testing at that point and then go on to consider what is necessary to make `master` wholly consistent.

I create new repo `gremlin-go sdk` [here](https://github.com/aliyun/alibabacloud-gdb-go-sdk) and it's already on some products online. 

Could you make a link on reference doc,  or do gremlin GLV have plan to support golong sdk ?

spmallette:
@heljoyLiu please post these additional questions to our dev list for further discussion where a wider range of the community is watching and can participate:

https://lists.apache.org/list.html?dev@tinkerpop.apache.org

Thanks!

PR #1259, opened 2020-03-06
TINKERPOP-2347 Remove invalid service descriptors from gremlin-shaded

olim7t:
No automated test, but I published a small project to reproduce the issue [here](https://github.com/olim7t/dummy-jpms).

It's actually only a warning when run from the command line, but I get an error in IntelliJ IDEA.


spmallette:
merge via CTR since this was a pretty simple change. minor point - i edited the CHANGELOG to a simple entry - we add in JIRAs at release time.  thanks for the contribution @olim7t

PR #1260, opened 2020-03-10
TINKERPOP-2335 Drop support for .NET Standard <2.0

FlorianHockmann:

This handles the .NET part of TINKERPOP-2335 by removing support for .NET Standard 1.3. This makes it easier for us to use newer .NET features in the future.

I also updated all .NET Core versions to the latest LTS version which is 3.1 right now. This includes the test projects, the dotnet template, the Travis file and our Docker build.
Unfortunately, I cannot try the Docker build right now as it isn't working for me on `master` at the moment. I'll add results of that once the Docker build is working again for me.

### DIFF
--- a/docs/src/upgrade/release-3.5.x.asciidoc
+++ b/docs/src/upgrade/release-3.5.x.asciidoc
@@ -211,6 +211,13 @@ All lambdas should be written using `gremlin-groovy` if they are needed.
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-2317[TINKERPOP-2317]
 
+==== .NET Standard 2.0 Only
+
+Gremlin.NET no longer targets .NET Standard 1.3, but only .NET Standard 2.0. Since .NET Core 2.0 and .NET Framework
+4.6.1 already support this .NET Standard version, most users should not be impacted by this.
+
+See: link:https://issues.apache.org/jira/browse/TINKERPOP-2335[TINKERPOP-2335]
+
 ==== Neo4j Changes
 
 There were two key changes to the neo4j-gremlin module:
--- a/gremlin-dotnet/glv/Gremlin.Net.csproj.template
+++ b/gremlin-dotnet/glv/Gremlin.Net.csproj.template
@@ -19,7 +19,7 @@ limitations under the License.
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup Label="Build">
-    <TargetFrameworks>netstandard1.3;netstandard2.0</TargetFrameworks>
+    <TargetFramework>netstandard2.0</TargetFramework>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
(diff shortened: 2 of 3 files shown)


spmallette:
Did you happen to verify if there were any issues with docfx generation as a result of this change?

FlorianHockmann:
> Did you happen to verify if there were any issues with docfx generation as a result of this change?

That shouldn't be affected from what I know. Just to be sure I also just built the docs again on this branch and it succeeded.

FlorianHockmann:
`docker/build.sh -t` now also succeeds after I rebased on `master` given [the change](https://github.com/apache/tinkerpop/commit/6735fabe94c1ac227d07a5ae91bb15c1eb202a62) from @spmallette to fix the Docker build.

So, I'm adding my VOTE +1.

spmallette:
does this mean that we have to install multiple versions of .NET to build on the different branches?

FlorianHockmann:
> does this mean that we have to install multiple versions of .NET to build on the different branches?

Good question. My understanding was always that the .NET Core SDK is completely backwards compatible which was correct, but what I didn't know is that that doesn't apply to the .NET Core runtimes which we need to execute the tests. I just tried to run the Gremlin.NET tests on `3.4-dev` with only .NET Core SDK 3.1 installed and it failed.

So, I guess that the best approach is now to update all release branches to .NET Core 3.1 which is the latest LTS version.
This has no effect on users of Gremlin.Net as that targets .NET Standard which doesn't need to be changed at all. The only change affecting users is the update in Gremlin.Net.Template, but that is a good thing in my opinion as the template is only used to create a completely new project and then using the latest LTS version sounds like the best approach to me.

Unless there are any concerns with this approach, I'm going to create a PR that targets `3.3-dev` and update to .NET Core 3.1 there. We can then merge that PR first into all release branches and I will rebase this PR afterwards. That means that all contributors (only) need .NET Core SDK 3.1 for all release branches afterwards.

spmallette:
yes - i'd like to try to keep one docker image for building, which is why i went to extremes to manually install the exact version of python we needed. if SDK 3.1 builds all the release branches then i guess we should go with that across the board. i guess the only downside is that we don't test the older branches explicitly with the version we say we support. 

i'm not sure at what point we "fork" our docker images across branches...

FlorianHockmann:
> i guess the only downside is that we don't test the older branches explicitly with the version we say we support.

What do you mean with _the version we say we support_? TinkerPop versions of past releases or .NET Core versions?
I don't really see a difference to be honest as we have only tested Gremlin.NET with a single .NET Core version in the past already. We just change that version now from 2.0 to 3.1.
And where do we say which versions we support? For TinkerPop versions, I only see the _Current Releases_ section on the [download page](http://tinkerpop.apache.org/downloads.html) and for Gremlin, we only supported .NET Standard versions explicitly and that doesn't really change.

spmallette:
Sorry, I don't mean to be confusing. Perhaps I'm just projecting what I know of python/jdk versioning on to this change and that I simply have the wrong understanding of the context for .NET. If using the new .NET SDK on the older branches doesn't increase our risk of things not working then that's fine to me. 

VOTE +1

FlorianHockmann:
No worries. I can just try to explain it a bit more detailed as .NET versioning / targeting can be confusing:

Gremlin.NET targets .NET Standard 1.3 and .NET Standard 2.0 right now. Users will include it in their applications that run on a .NET platform like .NET Core, .NET Framework, Mono or Xamarin. This PR drops support for .NET Standard 1.3 on `master` for TinkerPop 3.5.0. That has the effect that for example .NET Core 1.0 is not supported any more and users have to update to .NET Core 2.0 or higher. Since .NET Core 2.0 was released in October 2018 and .NET Core 1.x reached EOL in June 2019, that shouldn't be a big problem. The situation is similar for .NET Framework, Mono and the other platforms. .NET Standard was for example supported by .NET Framework since version 4.6.0, but .NET Standard is only supported since version 4.6.1.
This information can be found for all .NET Standard versions and platforms [in this table](https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-implementation-support).

While we only target .NET Standard with Gremlin.NET, we still need a platform to execute our tests on and to build the project. We use .NET Core for that as it's OS independent and because all .NET platforms seem to be slowly merging into .NET Core, resulting in [.NET 5](https://devblogs.microsoft.com/dotnet/introducing-net-5/) which will hopefully also make the .NET ecosystem easier to understand.
#1261 now updates our usages of .NET Core from 2.0 to 3.1 so we have the most recent SDK to build the project and to run the tests on .NET Core 3.1.
We can do that on all release branches because it really only affects our development process (apart from Gremlin.NET template).

And after #1261 is merged, this PR will really only drop support for .NET Standard 1.3.

PR #1261, opened 2020-03-11
TINKERPOP-2335 Update to .NET Core 3.1

FlorianHockmann:

This updates `3.3-dev` (and subsequently the other release branches) to .NET Core 3.1 so that we only need to have **one .NET Core SDK version installed which will be 3.1** after this PR is merged. 3.1 is the newest version right now and it is also an LTS version.

As already said on #1260, this basically only affects our build process / environment, but not users:

> This has no effect on users of Gremlin.Net as that targets .NET Standard which doesn't need to be changed at all. The only change affecting users is the update in Gremlin.Net.Template, but that is a good thing in my opinion as the template is only used to create a completely new project and then using the latest LTS version sounds like the best approach to me.

After this is merged I can rebase #1260 which will then only remove support for .NET Standard 1.3 on `master`.

`dotnet/build.sh -t` succeeds.

VOTE +1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ before_install:
   - sudo apt update
   - sudo apt install apt-transport-https
   - sudo apt update
-  - sudo apt install dotnet-sdk-2.2
+  - sudo apt install dotnet-sdk-3.1
   - sudo apt install python3.6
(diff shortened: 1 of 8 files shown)


spmallette:
VOTE +1

PR #1266, opened 2020-03-19
TINKERPOP-2351 Fixed bug in Order when enum is a key in Map

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2351

This seemed like an easy fix. Can't think of any bad side-effects. Not sure if there are other special cases we should consider, but at least this common case is solved.

```text
gremlin> g.V().valueMap(true).order(local).by(keys,desc)
==>[name:[marko],label:person,id:1,age:[29]]
==>[name:[vadas],label:person,id:2,age:[27]]
==>[name:[lop],lang:[java],label:software,id:3]
==>[name:[josh],label:person,id:4,age:[32]]
==>[name:[ripple],lang:[java],label:software,id:5]
==>[name:[peter],label:person,id:6,age:[35]]
gremlin> g.V(1).valueMap(true).order(local).by(keys,desc).next()[T.id]
==>1
gremlin> g.V(1).valueMap(true).order(local).by(keys,desc).next()[T.label]
==>person
```

When this merges forward to `3.4-dev` I will modify the test to use `elementMap()` rather than `valueMap(true)` since that method is deprecated.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -27,6 +27,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Added `trustStoreType` such that keystore and truststore can be of different types in the Java driver.
 * Added session support to gremlin-javascript.
 * Modified Gremlin Server to close the session when the channel itself is closed.
+* Fixed bug in `Order` where comparisons of `enum` types wouldn't compare with `String` values.
 * Added `maxWaitForClose` configuration option to the Java driver.
 * Deprecated `maxWaitForSessionClose` in the Java driver.
 * Remove invalid service descriptors from gremlin-shaded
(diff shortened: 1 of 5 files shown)


dkuppitz:
My first thought was that the general rule rather be: If `a` and `b` have different (non-numeric) types, use the `toString()` value for both, otherwise:

```
gremlin> g.V().map(properties().group().by(value).by(key()))
==>[29:age,marko:name]
==>[vadas:name,27:age]
==>[java:lang,lop:name]
==>[32:age,josh:name]
==>[java:lang,ripple:name]
==>[35:age,peter:name]
gremlin> g.V().map(properties().group().by(value).by(key)).order(local).by(keys, desc)
java.lang.String cannot be cast to java.lang.Integer
```

But then, taking Groovy as a reference, it probably shouldn't work this way.

```
gremlin> [id:'1','b':2,'z':3].sort()
==>b=2
==>id=1
==>z=3
gremlin> [id:'1','b':2,3:3].sort()
java.lang.String cannot be cast to java.lang.Integer
Display stack trace? [yN]
```

Thus, +1 on that change.

PR #1267, opened 2020-03-20
Add in example of using path with to or from modulators to the documentation

ctSkennerton:
When reading the reference documentation for the `path()` step I didn't realize you could use `to()` or `from()`, which are very useful features. This change adds in an example of their usage.

### DIFF
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -2189,15 +2189,15 @@ It's possible to limit the path using the <<to-step,`to()`>> or <<from-step,`fro
 
 [gremlin-groovy,modern]
 ----
-g.V().has('person','name','vardas').as('e').
+g.V().has('person','name','vadas').as('e').
       in('knows').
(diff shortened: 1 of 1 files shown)


docs/src/reference/the-traversal.asciidoc
```diff
@@ -2185,6 +2185,28 @@ g.V().out().out().path().by(
                                  __.in('created').values('name')).fold())
 ----
 
+It's possible to limit the path using the <<to-step,`to()`>> or <<from-step,`from()`>> step modulators.
+
+[gremlin-groovy,modern]
+----
+g.V().has('person','name','vardas').as('e').
```
spmallette: thanks for the pull request. i was a bit surprised to see we had no examples for that. could you please these are good examples, though, you misspelled "vadas" as "vardas" so the queries return no results. if you could please fix that, i'd be happy to get this merged.

ctSkennerton: oops, that's embarrassing. Changed.

PR #1273, opened 2020-04-01
TINKERPOP-2349 Migrate JSON serialization to System.Text.Json in Gremlin.NET

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2349

Results from a simple benchmark with BenchmarkDotNet showed that this improves the performance of the deserialization roughly by a factor of 2. The deserialization now also allocates far less memory, by a factor of ~7 which also results in less garbage collections. The performance of the serialization however stayed basically the same. It got even slightly slower by ~4% in this benchmark, but it also
allocates a bit less memory (-12%). This seems acceptable in my opinion, given that it is only a very small increase, especially compared to the differences for the deserialization and because it allocates less
memory now.

Output from BenchmarkDotNet:

Before:

|          Method |     Mean |    Error |   StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|---------------- |---------:|---------:|---------:|-------:|-------:|------:|----------:|
|   Serialization | 22.41 us | 0.616 us | 0.659 us | 1.5564 |      - |     - |   9.72 KB |
| Deserialization | 40.50 us | 1.075 us | 1.150 us | 6.3477 | 0.6104 |     - |  39.06 KB |

After:

|          Method |     Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------- |---------:|---------:|---------:|-------:|------:|------:|----------:|
|   Serialization | 23.25 us | 0.103 us | 0.091 us | 1.3733 |     - |     - |   8.58 KB |
| Deserialization | 25.89 us | 0.583 us | 0.545 us | 0.9155 |     - |     - |   5.74 KB |

Another advantage of using System.Text.Json is that it supports [`Span<T>`](https://docs.microsoft.com/en-us/archive/msdn-magazine/2018/january/csharp-all-about-span-exploring-a-new-net-mainstay) so we could make use of that in the future to reduce memory allocations in the driver even more.

VOTE +1

### DIFF
--- a/gremlin-dotnet/glv/Gremlin.Net.csproj.template
+++ b/gremlin-dotnet/glv/Gremlin.Net.csproj.template
@@ -61,15 +61,15 @@ NOTE that versions suffixed with "-rc" are considered release candidates (i.e. p
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All"/>
-    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />    
+    <PackageReference Include="System.Text.Json" Version="4.7.1" />
     <PackageReference Include="Microsoft.CSharp" Version="4.3.0" />
   </ItemGroup>
 
   <ItemGroup>
-    <None Include="../../LICENSE" Pack="true" PackagePath=""/>
-    <None Include="../../NOTICE" Pack="true" PackagePath=""/>
-    <None Include="../../../docs/static/images/gremlin-dotnet-logo_256x256.png" Pack="true" PackagePath="\"/>
+    <None Include="../../LICENSE" Pack="true" PackagePath="" />
+    <None Include="../../NOTICE" Pack="true" PackagePath="" />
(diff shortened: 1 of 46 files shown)


gremlin-dotnet/glv/Gremlin.Net.csproj.template
```diff
@@ -19,7 +19,7 @@ limitations under the License.
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup Label="Build">
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netcoreapp3.0</TargetFrameworks>
```
jorgebay, line 22: I think there's no need to add `netcoreapp3.0` as a target for the core project, given that `netcoreapp3.0` implements `netstandard2.0`.

FlorianHockmann: I wasn't sure here whether it makes sense or not. I thought that doing it like this is necessary to prevent the SDK from installing the NuGet package for `System.Text.Json` on .NET Core >= 3.0 as it's already included there. Or is that already detected automatically?

jorgebay: It will be resolved automatically for `netcoreapp3.0` users.

FlorianHockmann: Good to know. Then I'll remove this target again.

PR #1264, opened 2020-03-16
gremlin-python: add session mode client

heljoyLiu:

PR #1274, opened 2020-04-02
gremlin-python: add session mode client

heljoyLiu:


spmallette:
For ongoing reference, this PR replaces #1264.

spmallette:
Unless I'm misunderstanding something, you've removed the session "close" message on this PR. We in fact want the "close" message on `3.3-dev`. We want to consistently send the "close" message in `3.3-dev` and `3.4-dev` for all the drivers. It is only on `master` that we want to remove the sending of the message for all drivers. 

I'm sorry if all of this is confusing. I'm trying to be as specific as I can about what needs to be done and what branches need to have what changes. TinkerPop is a very mature product so all changes must be scrutinized and implemented with a plan that does its best to ensure consistency and continuity. 

For now, please don't worry about the changes to the `master` branch and removal of the "close" message. Let's get all the drivers sending the "close" message into `3.3-dev`. For that you just need to:

1.  Bring the "close" message into this PR and remove that "gremlin" argument in the "close" message
2. In #1275 just remove that "gremlin" argument in the "close" message.

After those two things are merged through all the branches, then, if you remain interested, we can worry about removal of those "close" messages in `master`. Please let me know if you have any questions.

heljoyLiu:
> For now, please don't worry about the changes to the `master` branch and removal of the "close" message. Let's get all the drivers sending the "close" message into `3.3-dev`. For that you just need to:
> 
> 1. Bring the "close" message into this PR and remove that "gremlin" argument in the "close" message
> 2. In #1275 just remove that "gremlin" argument in the "close" message.
> 
> After those two things are merged through all the branches, then, if you remain interested, we can worry about removal of those "close" messages in `master`. Please let me know if you have any questions.

I get your point now, sorry for that.

heljoyLiu:
It's ready now, sorry for late @spmallette

PR #1275, opened 2020-04-02
gremlin-javascript: remove session close method

heljoyLiu:


heljoyLiu:
new PR to remove session close. @jorgebay @spmallette thanks again

spmallette:
I think there might be some confusion. For `3.3-dev` you were just going to remove this:

```js
const args = {'gremlin': 'session.close()'};
```

Recall, that the "close" message specification does not call for the use of that "gremlin" argument. That small change was all that needed to go to `3.3-dev`.

This PR that removes the sending of the "close" message should be targeted at the `master` branch. It is on the `master` branch that the Java driver no longer supports sending it and we want to match that functionality there.

heljoyLiu:
> I think there might be some confusion. For `3.3-dev` you were just going to remove this:
> 
> ```js
> const args = {'gremlin': 'session.close()'};
> ```
> 

this is only tips or debug message, server do not care that also, I don't pay attention on "gremlin" section. 
So it means remove unnecessary section, and send request word by word. I'm aware of it now.


> This PR that removes the sending of the "close" message should be targeted at the `master` branch. It is on the `master` branch that the Java driver no longer supports sending it and we want to match that functionality there.

To the end, `3.3-dev` still remain "close" request of session in client, but `master` and `3.4-dev` target to remove this request.

sorry for this, so now I only remove the "gremlin" section in close request,  and this PR will re-target to `3.4-dev`.

spmallette:
> this is only tips or debug message, server do not care that also,

correct - the server ignores it, but we should adhere to the specification. if someone implementing a driver in Erlang looks at this code to get ideas for how it should work or perhaps someone writing a Gremlin Server implementation does the same, they will get the wrong impression (or at least get confused). 

> To the end, 3.3-dev still remain "close" request of session in client, but master and 3.4-dev target to remove this request.

sorry, not quite correct. `3.3-dev` and `3.4-dev` keep the "close" message. `master` will not.

heljoyLiu:
> correct - the server ignores it, but we should adhere to the specification.

ok ,I got it. for specification, thanks

> sorry, not quite correct. `3.3-dev` and `3.4-dev` keep the "close" message. `master` will not.

o... but PRs for .Net SDK on `3.4-dev` removed "close" request

spmallette:
> o... but PRs for .Net SDK on 3.4-dev removed "close" request

i must have missed that in review :( i thought that #1263 only resolved conflicts in `3.4-dev` and made sure everything worked given this comment from @FlorianHockmann :

https://github.com/apache/tinkerpop/pull/1257#issuecomment-595763604

@heljoyLiu i think you follow what needs to happen now. do not worry about `master`. if you could please make sure that `3.3-dev` and `3.4-dev` send the "close" message for all languages and that the "close" message does not have the "gremlin" argument. let's get that much correct before we worry about anything else. thank you.

heljoyLiu:
> @heljoyLiu i think you follow what needs to happen now. do not worry about `master`. if you could please make sure that `3.3-dev` and `3.4-dev` send the "close" message for all languages and that the "close" message does not have the "gremlin" argument. let's get that much correct before we worry about anything else. thank you.

ok, until now, follow works for me:
1. make sure that `3.3-dev` and `3.4-dev` send the "close" message for all languages
2. the "close" message does not have the "gremlin" argument

and more detail:
1. remove "gremlin" argument of javascript on `3.3-dev`
2. add the session "close" request for .Net SDK on `3.4-dev`
3. re-target session PR with session "close" request for python on `3.3.-dev`

heljoyLiu:
hi, @spmallette
 I update session close request, and remove 'gremlin' argument.

@jorgebay I move the session code to 'client', and recovery 'connection' as it was, and I think it's more simple and graceful, thanks again

gremlin-javascript/src/main/javascript/gremlin-javascript/test/helper.js
```diff
@@ -48,5 +48,5 @@ exports.getClient = function getClient(traversalSource) {
 };
 
 exports.getSessionClient = function getSessionClient(traversalSource) {
-  return new Client(serverUrl, { 'traversalSource': traversalSource, 'processor': 'session'});
-};
\ No newline at end of file
+  return new Client(serverUrl, { 'traversalSource': traversalSource, 'session': 'unique-string-id'});
```
jorgebay, line 51: NIT: For the tests, I would avoid using the same identifier for the session as the documentation. It may wrongly imply that `unique-string-id` means something.

```suggestion
  return new Client(serverUrl, { 'traversalSource': traversalSource, 'session': 'my-nodejs-session'});
```

heljoyLiu: ok, it's better to put a uuid-string, but I'm lazy here

spmallette:
I've merged this as a single squashed commit as: a9c02e4259e01847e56a24fe4a9ff93da8ab3fe5

Thanks @heljoyLiu

PR #1278, opened 2020-04-16
gremlin-python: add 'timeout' argument in client-submit

heljoyLiu:


heljoyLiu:
we could build RequestMessage with timeout args
```
    g = traversal().withGraph(Graph())
    t = g.V().has("customer_id", 851401972585122).id()

    message = RequestMessage('traversal', 'bytecode', {'gremlin': t.bytecode, 'scriptEvaluationTimeout': 50000L})
```

so close it. sorry for disturb

krlawrence:
Just a comment, I think I have a Jira open on this somewhere but could not find it in a quick search. It would be good if the Python client, and for that matter any of the GLV clients exposed easy ways to set options like the per query timeout. I get fairly regular feedback from TinkerPop users who are frustrated that the Java client has features that are enabled by special case code that the other GLVs do not yet expose. Per query timeout is probably the one that comes up the most bit others like sessions (recently added - thank you) come up as well.

heljoyLiu:
> Just a comment, I think I have a Jira open on this somewhere but could not find it in a quick search. It would be good if the Python client, and for that matter any of the GLV clients exposed easy ways to set options like the per query timeout. I get fairly regular feedback from TinkerPop users who are frustrated that the Java client has features that are enabled by special case code that the other GLVs do not yet expose. Per query timeout is probably the one that comes up the most but others like sessions (recently added - thank you) come up as well.

it's better to give users the same features all GLVs, like API(submit, submitAsync) , options, thanks.
now, I'm not sure how to set Timeout args for bytecode request, and couldn't find issue like this. 
so, I support timeout in demo graph server as bellow.
```
g.with('scriptEvaluationTimeout', 5000L).V().count()
```

krlawrence:
The problem with that in a general sense is that by the time the query is run it's too late and the Gremlin Server will not have configured the timeout in my understanding. The Java client has special code to look for that OptionsStrategy and pick the timeout out before sending to the server. I think we should do that in all the GLV clients.

sadams-rti-org:
+1
We finally gave up and just turned off the timeout because of this issue.
We use Groovy/Gremlin scripts as well as Python and Javascript GLVs

krlawrence:
I finally found the issue I remembered opening - been having odd account issues with Jira.

https://issues.apache.org/jira/browse/TINKERPOP-2296

heljoyLiu:
> I finally found the issue I remembered opening - been having odd account issues with Jira.
> 
> https://issues.apache.org/jira/browse/TINKERPOP-2296

en ,yes. I iterator OptionsStrategy of bytecode in server for the timeout args...

PR #1279, opened 2020-04-16
TINKERPOP-2288 Replace closed connections directly

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2288

There were already some comments on the earlier commit which are now unfortunately lost as I force pushed on that branch after rebasing on `3.4-dev` again.

Closed connections are now replaced automatically in the background.
If no open connection is available to answer a request, then the pool
tries it again after some time. It uses a retry policy with exponential
backoff for that, implemented with Polly.
This change also ensures that only one task performs a pool resizing
operation at a time.

These changes should ensure that:
- A connection is still returned quickly if one is available.
- Closed connections are replaced immediately, without needing to wait
    for the next incoming request.
- If the server is only unavailable temporarily (or it just closed
open connections for some reason), then the user should should not get
an exception.
He only has to wait until the connections are replaced.

TODO:
- [x] Make the retry policy configurable
- [x] Document changes


gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionPool.cs
```diff
@@ -148,26 +187,39 @@ private void ProtectIndexFromOverflowing(int currentIndex)
...
         }
 
-        private void ConsiderUnavailable()
+        private async Task ReplaceClosedConnectionsAsync()
         {
-            CloseAndRemoveAllConnectionsAsync().WaitUnwrap();
+            var poolWasPopulated = await EnsurePoolIsHealthyAsync().ConfigureAwait(false);
+            // Another connection could have been removed already, check if another population is necessary
+            if (poolWasPopulated)
+                await ReplaceClosedConnectionsAsync().ConfigureAwait(false);
```
FlorianHockmann, line 211: There was also a comment on this line [in the earlier commit](https://github.com/apache/tinkerpop/commit/5b490d310d089a82ccb8bf2e66d770299d224773#r38325980) where @spzSource asked whether this recursive call could lead to a stack overflow. I answered also inline at that commit with an explanation of why I think that a stack overflow cannot happen here.

jorgebay:
I'll try to review this week.

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1090,10 +1090,13 @@ on the `ConnectionPoolSettings` instance that can be passed to the `GremlinClien
 |Key |Description |Default
 |PoolSize |The size of the connection pool. |4
 |MaxInProcessPerConnection |The maximum number of in-flight requests that can occur on a connection. |32
+|GetOpenConnectionRetries |The number of retries to get an open connection from the pool to submit a request. |4
```
jorgebay, line 1093: NIT: Maybe we could use `GetReconnectionAttempts()` as a name?

Instead of a number, it could return an `IEnumerable<int>`, that represents milliseconds to wait between each attempt.

FlorianHockmann: > Maybe we could use GetReconnectionAttempts() as a name?

I also didn't like the name `GetOpenConnectionRetries` that much. `ReconnectionAttempts` sounds like a better name overall. I will probably omit the _Get_ prefix as it's a property and the _Get_ in `GetOpenConnectionRetries` meant the number of retries to _get_ a connection.

> Instead of a number, it could return an IEnumerable<int>, that represents milliseconds to wait between each attempt.

Do you think that users want to configure the individual wait times explicitly instead of just using the default exponential back-off? And wouldn't an `Action<int>` (or `Action<TimeSpan>`) be even more flexible and easier to configure in that case?
If we let them configure an `Action` to compute the wait times, then it should probably be an additional option though as we would still need an option to configure the number of retries in general.

jorgebay: It's true that exponential + jitter is what most people wants, my concern is whether we should allow users to configure the base delay, for example: users might want to start from 200ms, 400ms, 800ms and other people in the order of seconds.

Looking at it a little bit more, maybe there should be two settings: "number of attempts" and "base delay".

FlorianHockmann: So, we should introduce a constant factor for the exponential wait function? That would look like this:

```c#
attempt => baseDelay * Math.Pow(2, attempt)
```

and with `baseDelay = TimeSpan.FromMilliseconds(100)` they would get your wait times of 200ms, 400ms, 800ms, and so on.
With `baseDelay = TimeSpan.FromSeconds(1)` as the default we would get the current behaviour in this PR.

jorgebay: yes, that would be awesome!

FlorianHockmann: I just pushed an update that includes these changes.

jorgebay:
lgtm!

PR #1293, opened 2020-06-10
TINKERPOP-2264 update gremlin-python tests to be timezone/system-independent, related minor Docker changes

humbleengineer:
The datetime conversion tests didn't pass when running on a Mac in London, so they've been updated to use `calendar.timegm` just like the code under test was in the [original PR for TINKERPOP-2264](https://github.com/apache/tinkerpop/pull/1165).

Also, in order to run & debug the integration tests I updated gremlin-python's pom.xml to refer to a `python` executable rather than `python2`.  Since [Docker on Mac runs in its own VM](https://docs.docker.com/docker-for-mac/docker-toolbox/#the-docker-desktop-on-mac-environment) I updated the integration test `gremlin-server.sh` to explicitly publish its ports using `-p`.  Happy to revert these if they're too environment-specific.

`gremlin-python` install target builds clean using IntelliJ Maven support (I get a javadoc-related error building from command line).

### DIFF
--- a/docker/gremlin-server.sh
+++ b/docker/gremlin-server.sh
@@ -57,7 +57,7 @@ echo "Using Gremlin Server $GREMLIN_SERVER_VERSION"
 sed -e "s/GREMLIN_SERVER_VERSION\$/${GREMLIN_SERVER_VERSION}/" docker/gremlin-server/Dockerfile.template > Dockerfile
 
 docker build -t tinkerpop:${BUILD_TAG} .
-docker run ${TINKERPOP_TEST_DOCKER_OPTS} ${REMOVE_CONTAINER} -ti -v "${HOME}"/.groovy:/root/.groovy -v "${HOME}"/.m2:/root/.m2 tinkerpop:${BUILD_TAG} ${@}
+docker run ${TINKERPOP_TEST_DOCKER_OPTS} ${REMOVE_CONTAINER} -ti -p 45940:45940 -p 45941:45941 -v "${HOME}"/.groovy:/root/.groovy -v "${HOME}"/.m2:/root/.m2 tinkerpop:${BUILD_TAG} ${@}
 
 status=$?
 popd > /dev/null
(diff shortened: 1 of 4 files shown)


gremlin-python/pom.xml
```diff
@@ -311,7 +311,7 @@ limitations under the License.
                                             <fileset dir="src/main/jython"/>
                                         </copy>
                                         <!-- print versions of python and virtualenv -->
-                                        <exec dir="${project.build.directory}/python2" executable="python2"
+                                        <exec dir="${project.build.directory}/python2" executable="python"
```
spmallette, line 314: I think I'd like to keep the same assumptions we've always had about local dev environments as we're deep into 3.4.x at this point and python2 is gone in 3.5.0. Do you mind reverting this change and doing a alias or something in your environment locally to accommodate this configuration?

docker/gremlin-server.sh
```diff
@@ -57,7 +57,7 @@ echo "Using Gremlin Server $GREMLIN_SERVER_VERSION"
 sed -e "s/GREMLIN_SERVER_VERSION\$/${GREMLIN_SERVER_VERSION}/" docker/gremlin-server/Dockerfile.template > Dockerfile
 
 docker build -t tinkerpop:${BUILD_TAG} .
-docker run ${TINKERPOP_TEST_DOCKER_OPTS} ${REMOVE_CONTAINER} -ti -v "${HOME}"/.groovy:/root/.groovy -v "${HOME}"/.m2:/root/.m2 tinkerpop:${BUILD_TAG} ${@}
+docker run ${TINKERPOP_TEST_DOCKER_OPTS} ${REMOVE_CONTAINER} -ti -p 45940:45940 -p 45941:45941 -v "${HOME}"/.groovy:/root/.groovy -v "${HOME}"/.m2:/root/.m2 tinkerpop:${BUILD_TAG} ${@}
```
spmallette, line 60: I'm still not smart enough about docker to know why this worked all this time without the `-p` - @robertdale or @dkuppitz  any ideas?

PR #1294, opened 2020-06-10
TINKERPOP-2377 .NET GLV: Check for null earlier

jorgebay:
https://issues.apache.org/jira/browse/TINKERPOP-2377

I think this addresses the issue.


spmallette:
I've done about 6 runs of travis on this PR and i don't see that failure anymore. This might have fixed it, though still unclear as to why the problem was intermittent and only on travis. Anyway.....feel free to merge this one.

VOTE +1

jorgebay:
gotcha, merging this in a CTR fashion.

PR #1295, opened 2020-06-12
TINKERPOP-2383 Fixed bug in has(T,Traversal)

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2383

That has() was using PropertiesStep, but properties() doesn't process T values.

```text
gremlin> g.V().has(label, is('software'))
==>v[3]
==>v[5]
gremlin> g.V().has(id, is(6))
==>v[6]
```

VOTE +1


divijvaidya:
VOTE +1 LGTM!

PR #1298, opened 2020-06-15
[DRAFT] TINKERPOP-2352 reduce Python driver default pool sizes to 1

humbleengineer:
Reduce client pool size to 1, so that it's possible for keep-alive logic outside the client to keep the single client alive using Gremlin traversals.  Also reduce the number of worker threads to 1 by default.

This is on the assumption that most Python-based programs using Gremlin clients are essentially single-threaded, and so performance impacts should be minimal.

Draft PR while I look at how to (re-)add tests that will actually exercise multiple threads.


krlawrence:
I am not sure I agree that "most" Python Gremlin apps are single threaded. Just because Python has the Global Interpreter Lock (GIL) does not mean that the threading library is not effective in I/O bound scenarios. I regularly write Python Gremlin apps that have anywhere from 8 to 96 threads. Because requests to a Gremlin Server tend to be I/O bound the Python threading library actually works well as threads will naturally block while waiting for I/O. Rather than change these defaults I think a better solution might actually be to add as needed to the documentation giving specific examples of where different configurations make sense and include code examples of how to set those configurations up. I am aware of many good use cases for each. Here are just two cases

1. A client doing a lot of writes across multiple threads from a container/API server etc. (multiple threads definitely help here).

2. An AWS Lambda function where it makes sense to have just 1 connection as you can think of a Lambda as being a thread in that context.

humbleengineer:
Thanks for the feedback, @krlawrence.  I've reverted the code changes and added an asciidoc change instead.  I also added a new test for explictly multi-threaded use of `Client`, because I had written one to make sure this was still covered with the changed default `pool_size`.

spmallette:
Given the discussions here I cherry picked in several commits from this PR:

14986f1462c2b8d7a2d86163dd704d887ee2587d
724b3effcb04745767652609a89c9fc60561f484

Thanks!

humbleengineer:
> Given the discussions here I cherry picked in several commits from this PR:
> 
> [14986f1](https://github.com/apache/tinkerpop/commit/14986f1462c2b8d7a2d86163dd704d887ee2587d)
> [724b3ef](https://github.com/apache/tinkerpop/commit/724b3effcb04745767652609a89c9fc60561f484)
> 
> Thanks!

Thanks @spmallette!  I'd been meaning to merge in the fixes to build stability I'd seen recently and get this PR actually building, but looks like that's become moot.

PR #1303, opened 2020-07-13
Add support to provide extra arguments to the Client.submit()

Otanikotani:
The goal of this PR is to provide a capability to specify arguments on query submission via JS Client. In our use case we would like to be able to specify `scriptEvaluationTimeout` argument for certain queries (which we expect to be slow and we want to prevent accidental timeouts). 

I am by no means a JS expert, hence probably adding extraArgs as an argument to the `submit` function is not ideal, but with this change we can write something like this to increase the timeout:

```js

const traversal = gremlin.process.AnonymousTraversalSource.traversal;
const p = gremlin.process.P;
const DriverRemoteConnection = gremlin.driver.DriverRemoteConnection;

const Client = gremlin.driver.Client;

const client = new Client('<endpoint>','gmodern')
client.open();
message = `<very-long-query>`

client.submit(message, null, {"scriptEvaluationTimeout": 60000}).catch(function(error) {
  console.log("BOOOM ERROR!");
  console.log(error);
}).then(function (result) {
  console.log(result);
});

client.close();
```

Also, just to clarify - we can't use Java Client (where specifying arguments seems to be supported) as this is used in AWS JS Lambda -> AWS Neptune interaction.

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/client.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/client.js
@@ -69,17 +69,18 @@ class Client {
    * Send a request to the Gremlin Server, can send a script or bytecode steps.
    * @param {Bytecode|string} message The bytecode or script to send
    * @param {Object} [bindings] The script bindings, if any.
+   * @param {Object} [extraArgs] The extra arguments, if any.
    * @returns {Promise}
    */
-  submit(message, bindings) {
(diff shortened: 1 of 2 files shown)


spmallette:
thanks for the contribution @Otanikotani - i'm sure @jorgebay will take a look at this when he gets a moment. 

As a side note, I think this is also an issue in other GLVs as well: https://issues.apache.org/jira/browse/TINKERPOP-2296

spmallette:
Merged and added some documentation.

PR #1304, opened 2020-07-22
Support deserialization of dict as a key of another dict (python)

liranms:
Added support for dict as keys, as required by the following example statement: g.V().as_('a').out().as_('b').out().as_('c').dedup('a','b','c').groupCount().by(select('a', 'b').by(valueMap(True))).toList()

Previous implementation returned an exception: unhashable type: 'dict'

### DIFF
--- a/gremlin-python/src/main/python/gremlin_python/structure/io/graphsonV3d0.py
+++ b/gremlin-python/src/main/python/gremlin_python/structure/io/graphsonV3d0.py
@@ -42,6 +42,27 @@ _serializers = OrderedDict()
 _deserializers = {}
 
 
+class hashable_dict(dict):
+    def __hash__(self):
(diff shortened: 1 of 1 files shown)


spmallette:
huh - this change seems smart. we'd need uniform support for this in graphbinary though. also, we'd want some unit tests and potentially need to enable some of the GLV test suite since this feature is now available. Would you be able to augment this PR with at least the graphbinary support and unit tests. We can discuss the GLV tests separately once those changes are in place? Finally, It would be nice if you could create an issue in JIRA for this: https://issues.apache.org/jira/projects/TINKERPOP

spmallette:
I've pulled this change over to #1314 and made some additional changes as I commented on previously. Closing this one in favor of that.

PR #1305, opened 2020-07-23
TINKERPOP-2394 Unable to use __ class of a custom DSL when passing a script even if this class is imported

mmadoo:


spmallette:
I suppose that this change makes it the responsibility of the user to validate that they don't override any of the Gremlin DSL with their imports. I think that's ok, but this could represent a breaking change to those who might rely on the ordering of the `Customizer` implementations as they are today. As such I'm not sure if this should land in 3.4.x - perhaps it is best saved for 3.5.0 on master.

spmallette:
I was about to merge this when I wondered if the better solution was to simply rename the generated DSL class for anonymous traversals - couldn't we just rename it to something other than `__`? maybe it's simply `___` (i.e. 3 underscore instead of 2)? thoughts?

mmadoo:
I think that renaming generated DSL class to `___` may be a breaking change higher than reordering imports.
There can also be some corner case where overwrite of other classes or constants may be needed (for instance, if I have a constant `public static String id = "ID"` for a specific customer requirement, I would like to be able to use it and use `T.id` to get element ID).

spmallette:
> I think that renaming generated DSL class to ___ may be a breaking change higher than reordering imports.

sure and i was ok with that for 3.5.0. 

> There can also be some corner case where overwrite of other classes or constants may be needed (for instance, if I have a constant public static String id = "ID" for a specific customer requirement, I would like to be able to use it and use T.id to get element ID).

ok - maybe not letting Gremlin have the last word on what the imports are is ok. I will merge this as it is. thanks.

PR #1306, opened 2020-07-23
Update client.js to handle default options

gugacavalieri:
Following the docs present in [here](https://github.com/apache/tinkerpop/blob/master/gremlin-javascript/src/main/javascript/gremlin-javascript/README.md), getting started with Gremlin throws the following error:

```bash
/home/node/node_modules/gremlin/lib/driver/client.js:49
if (this._options.processor === 'session') {
  TypeError: Cannot read property 'processor' of undefined
  at new Client (/home/node/node_modules/gremlin/lib/driver/client.js:49:23)
  at new DriverRemoteConnection (/home/node/node_modules/gremlin/lib/driver/driver-remote-connection.js:53:20)
```

So, I suggest that we can pass a default empty object to the client constructor, what do you think?

Default docs reference:
```js
const g = traversal().withRemote(new DriverRemoteConnection('ws://localhost:8182/gremlin'));
```

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/client.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/client.js
@@ -44,7 +44,7 @@ class Client {
    * @param {String} [options.session] The sessionId of Client in session mode. Defaults to null means session-less Client.
    * @constructor
    */
-  constructor(url, options) {
+  constructor(url, options = {}) {
     this._options = options;
     if (this._options.processor === 'session') {
       // compatibility with old 'session' processor setting


spmallette:
wow - thanks for noticing that. that's a horrible error for someone new to get stuck with. it seems our tests only hit a variation of that `g` initialization code and not the one that takes no options, though I sense that most people would eventually make the call with an alias at a minimum - perhaps that is why no one has complained sooner. 

As ultimately options are defaulted in the manner you show, I think your change makes sense. I've cherry-picked it to 3.4-dev and merged to master (just in time for release thankfully - we begin code freeze/testing for 3.4.8 release next week) as: 7894317148b37932b9d91dc0a5e597dac47b4662

gugacavalieri:
Thanks @spmallette! Glad that we caught it just in time for the next release :smile:

PR #1307, opened 2020-07-28
[TINKERPOP-2396] add TraverserSet supplier option for AbstractStep and ExpandableStepIterator

rdtr:
This provides a way for GraphDB provider to change the logic how Traverser objects are hashed into a key in Map.

Example: I want to avoid that Element#id is used as a hash key as we have other internal representation for Element and want to use that instead.


rdtr:
OK it looks I should have paid attention to serialization logic...will check.

rdtr:
@spmallette As I replied in the JIRA ticket, on my end this change is enough to help our use case. please let me know if any further update is required.

spmallette:
I've cherry picked your change to #1320 and made some modifications. i might yet play with the method signatures a bit but the general idea is in place. also intend to do a bit more review of the change in general. hopefully this fits the need you were considering.  let's please continue discussion and collaboration on that PR rather than this one, which i will close at this point. Thanks!

rdtr:
Thanks, yes on our end we extend the default `Graph`, `Traversal` and `TraversalSource` class for our own usage so having `Traversal` sets a provider's own `TraverserSet` implementation would work for us 👍

PR #1308, opened 2020-08-06
TINKERPOP-2389 Authorization support in TinkerPop

vtslab:
[https://issues.apache.org/jira/browse/TINKERPOP-2389](https://issues.apache.org/jira/browse/TINKERPOP-2389)   
This work is by no means finished, but is a first shot along the lines of the discussion linked to in the JIRA ticket. The work includes and Authorized API, called from the AbstractChannelizer, as well as a small WhitelistAuthorizer demo implementation included in the gremlin-server tests. The current WIP PR is only intended to simplify a review of this first concept.     

As the RequestMessage object passed to the authorize() method is not easy to handle, I thought it useful to do some first preprocessing common to future Authorizer implementations:    
- split authorization requests in those for bytecode requests and those for string-based requests
- get and cast the object corresponding to Token.ARGS_GREMLIN 
This is not entirely consistent though as the other Token.ARGS_XYZ are left unpacked in the original RequestMessage.   

Any feedback is welcomed.


vtslab:
To preclude comments about completeness of the current PR, I am fully aware that the following is still missing (but not essential to designing/assessing the actual Authorizer API):
- authorize calls from the other Channelizers, as well as integration tests for these
- unit tests on the Authorizer abstract class itself
- documentation and changelog

spmallette:
@vtslab sorry i haven't had a chance to look at this yet. i'm sure i will have time next week.

spmallette:
i'm finally getting to this - sorry for the delay. the general direction here seems ok. i like that it doesn't particularly impose anything on providers who implement Gremlin Server protocol - it just introduces a companion to authentication interfaces. 

I think `Authorizer` would be better as an interface with just the:

```java
public AuthorizedUser authorize(final AuthenticatedUser user, final RequestMessage msg) throws AuthorizationException
```

and some kind of `AbstractAuthorizer` that has your "preprocessing" as a base for folks to implement from. 

I might have more to say about `AuthorizationHandler`, `AuthorizedUser` and the default `Authorization` implementation but for the most part it seems simple and straightforward enough. Now that I've looked at this in detail, I will be able to think about it more specifically in coming days. Thanks.

QwentB:
If considerfing a JWT authentication and an ABAC authorization implementation, the token will obviously contains the userId/subject but also most probably user attributes/claims that will be needed for the authorization process.
Wouldn't it make sense to use the AuthenticatedUser as the container for these attributes in a Map for example?

vtslab:
@QwentB Thanks for joining in. I am not familiar with JWT tokens, but I understand that a client receives the token after authentication. However, in the context of a secure websockets connection, once a client is authenticated there is no need to exchange tokens for authorizing traversal requests. Do I miss something? Do you only refer to gremlin-server acting as an HTTP endpoint (still present for serving languages without GLV support)? 
Do you mean AuthenticatedUser or AuthorizedUser as the container for authorization attributes?

QwentB:
@vtslab Thanks for your answer.
Yes the client receive a token after authenticating on an identity provider service, which might be part of the same infrastructure (i.e. a corporate service) or external (Google, Facebook ...). However, when the token is received by the consuming service (Janus or Gremlin here), it should be validated (signature, issuer, date time of issue, end of validity...) before being accepted as valid credentials. Considering the actual authentication scheme in Gremlin, this is a typical SASL use case and the token validation process is the "real" local authentication handled by the Authenticator.

It means that the Authenticator will have to "open" the token to extract the subject (aka the user id) but might as well extract all other claims (aka user attributes/properties) available in the token. My point is that if these attributes were to be used in the authorization process, the Authorizer must be provided with either the attributes (extracted by the Authenticator), or the token (and then reopen it and extract all attributes). IMO it make sense to deal with the token only once, for obvious performance reasons, but also for an ABAC Authorizer to be independent of the Authentication mechanism. With this in mind, the AuthenticatedUser seems to be the right container to convey these data.

The JWT authentication shall be used for both HTTP requests and WebSockets. For WebSockets, the authentication happens during the initial connection and before the request upgrade, so if the AuthenticatedUser with the attributes is available in the ChannelContext, it can used by the authorization process each time a RequestMessage is received.

vtslab:
Thanks for the explanation @QwentB, this makes sense. Apart from making unit tests, your request is really simple: allow the Authenticator implementation to pass on some user defined properties (typically a Map) inside the returned AuthenticatedUser. The current Authorizer proposal already receives the AuthenticatedUser and can read the user defined properties if it has knowledge about their meaning. TinkerPop itself has no knowledge about the user defined properties: it just allows for solutions that implement both the Authenticator and Authorizer interfaces and leverage the channelizer to store and pass on the user defined properties.

QwentB:
@vtslab functionnaly this is what I had in mind.
I'm not sure to understand your point about the WS connection closing. If the WebSocket connection is closed not further calls would be made on the Authenticator/Authorizer so their state and the presence of the properties does not really matter. 
This being said, we should make sure they will all be cleanly garbage collected and it might requires some 'dispose' action. I'm not very familiar with Netty but there should be a way to be notified when the Channel is being closed.

spmallette:
As an aside from some of the technical discussion here - thanks for participating @QwentB - I thought that there was some relevance in this thread on Gremlin Users:

https://groups.google.com/g/gremlin-users/c/Ff0sHJk7JiE/m/tsJFsNurAgAJ

A common use case that seems related to authorization is to apply some `TraversalStrategy` to a `TraversalSource` based on the user. As you can see from that thread, it's often "hard" to do that.  @vtslab any thoughts on the matter?

QwentB:
Thanks @spmallette for linking to two threads as they're obviously related.
I'm not sure if it would cover all possible use cases, but the Authorizer could return the RequestMessage instead of the AuthorizedUser. It would allow an implementations to update the RequestMessage by adding/modifying steps, add/remove Strategies or even change the "resource" that will be effectivelly bound to the request.
As pointed by @vtslab, the user's authorizations should be available in the context, at least for WebSocket requests, so they can be applied without requiring an authentication step. If the Authorizer return an updated RequestMessage, we probably need to introduce an AuthorizedUserProvider component and the channel look like something like

1. Authenticator return the AuthenticatedUser
2.

QwentB:
Thanks @spmallette for linking to two threads as they're obviously related.

I'm not sure if it would cover all possible use cases, but the Authorizer could return the RequestMessage instead of the AuthorizedUser. It would allow an implementation to do some pre-checks (script vs bytecode, VertexProgram, Lambda, use of specific strategies or steps...), but also update the RequestMessage by adding/modifying steps, add/remove Strategies or even change the "resource" that will be effectivelly bound to the request.

As pointed by @vtslab, the user's authorizations must be available in the context for WebSocket requests, so they can be applied without requiring authentication. If the Authorizer return a RequestMessage, an AuthorizedUserProvider component is probably needed to build the actual AuthorizedUser and add it to the channel context. 

The pipeline would then look like:

1. `Authenticator` returns the `AuthenticatedUser`
2. `AuthorizedUserProvider` takes the `AuthenticatedUser` as parameter and return the `AuthorizedUser` which will be added to `ChannelContext`
3. `Authorizer` takes the `AuthorizedUser` and the `RequestMessage` as parameters and return a potentialy updated `RequestMessage`

This require the AuthorizedUser to contains the authorizations for the user. To stay agnostic of any implementation, the AuthorizedUser could be specified as a parametrized class defining the authorizations' type. It might also be relevant that the AuthorizedUser contains some global metadata about the authorizations (i.e. date time of issue to force a renewal). 
In the end the AuthorizedUser could look like:
```
**
 * @param <A> the type of Authorizations
 */
public class AuthorizedUser<A> {

    private final String name;
    private final Map<String, Object> metadata;
    private final Map<String, A> authorizations;

    /**
     * Construct an AuthorizedUser instance
     *
     * @param name           the name of the AuthenticatedUser
     * @param metadata       a key/value store for information about the given authorizations
     * @param authorizations the authorizations given to the user for each resource
     */
    public AuthorizedUser(String name, Map<String, Object> metadata, Map<String, A> authorizations) {
        this.name = name;
        this.metadata = metadata;
        this.authorizations = authorizations;
    }

    public String getName() {
        return name;
    }

    public Map<String, Object> getMetadata() {
        return metadata;
    }

    public Map<String, A> getAuthorizations() {
        return authorizations;
    }
}
```

vtslab:
Thanks @QwentB for further elaborating your thoughts. There are some aspects, though, that make me hesitant to pursue them.

First, storing authorization data in the pipeline has some impact on the number of connections that can be open. That would be problematic in a scenario where many users are served who fire a query now and then. When the Authorizer caches the authorization data, however, it can simply put a limit on the size of the cache. 

Secondly, while one could modify the RequestMessage to support specific usage scenarios, it does not seem necessary. If a RequestMessage is not admissable in its present form, the Authorizer can simply reject it and add a helpful reject message. Add/remove Strategies can be done by offering the same graph via different GraphTraversalSources that differ by the Strategies applied. So, I feel that the suggestion to have the Authorizer return a possibly modified RequestMessage is out of scope for the current PR. It could be added later on, if really required.

QwentB:
Thanks for your answer @vtslab 

I agree that caching authorization have a memory impact and that it does not make sense with a "many users submitting few requests" scenario, on the other hand, if you consider the gremlin server as a storage backend accessed by a few front-end servers (so a "few users / many requests" scenario) caching the authorizations makes more sense. You're absolutely right that it should definitely have some configurable limits and eviction strategy.
All this being said, caching was just an example of the possibilities offered to potential implementations, my main point was more "separation of concerns" related as, IMO, the authorization provisioning should be separated from the authorization process itself.

Regarding the second point, I might be misunderstanding what you mean by:

> "Add/remove Strategies can be done by offering the same graph via different GraphTraversalSources that differ by the Strategies applied"

, but I guess you're referring to the statically defined traversalSources in the server's configuration file. The use case being the definition of a `graph1ReadOnly` traversal sources pointing to `graph1` but with a ReadOnlyStrategy, and then make sure `ReadOnlyUser` can only access `graph1ReadOnly`, and the same principle could be applied to a `graph1WithoutConfidentialData` traversal source with a SubgraphStrategy applied.

I agree that request message analysis coupled with such configuration provides some interesting features and will cover some use cases, but it is not applicable to an ABAC implementation where the potential combinatory of all attributes does not allow the static definition of all possible traversal sources. With the Authorizer interface returning a RequestMessage, an Authorizer could, for example, inject a SubgraphStrategy matching the exact authorizations of the user and effectively limitating the visbility.

I'm also wondering why the Authorizer is currently returning an AuthorizedUser which does not contain much and does not seems to be used anywhere afterward. If the idea is to add it to the context and use it later down the pipeline, for example, to call the "withAuthorizaton' method of an AuthorizedTraversalSource, I thing we're back to point #1 about storing data in the context.

vtslab:
The latest commits bring the PR somewhat closer to a completion. Apart from being more complete, the following points were addressed:
- the AuthorizedUser was removed from the Authorizer API. As also remarked by QwentB, it had no function.
- a mechanism was added to facilitate interaction between the Authorizer and Authentication implementations (to support one-pass interrogation of an external access control service),  by means of the AuthenticatorAuthorizer interface. Actually, I am still in doubt whether this is the best solution because it lays the burden on the TinkerPop code. Alternatives are there, e.g. by implementing the Authenticator as a singleton or even via the network. Any thoughts?

Thanks, @QwentB, for your patient explanation about ABAC. Indeed, if authorization is very granular, this cannot be supported with static GraphTraversalSources. So, now I support your suggestion to have authorize() return a RequestMessage and passing this on into the pipeline. Although this makes the pipeline less transparent, there is already a (necessary) precedent in tossing around with the RequestMessage in the SaslAuthenticationHandler. @spmallette, agree?

vtslab (commented):
Changes in test-server-start.groovy are obviously wrong (accidental from IDE) and will fail the GLV tests.

docs/src/dev/provider/index.asciidoc
```diff
@@ -1147,25 +1147,64 @@ one key value pair present (since only one `Traversal` is being submitted, there
 single alias).
 |=========================================================
 
-=== Authentication
+=== Authentication and authorization
```
spmallette, line 1150: nit: Capital "A" in "Authorization" please since it's a title. 
nit: There's a bit more formatting to do in the text like enclosing class names in backticks.

I think it would be worth adding some note here to providers to say that while Gremlin Server supports this authorization feature it is not a feature that TinkerPop requires of graph providers as part of the agreement between client and server. Graph providers may choose to implement their own methods for authorization in the manner they see fit.  I would say a similar "IMPORTANT" callout box should probably be added to the reference documentation to alert users to this notion. Finally, as you draw closer to a final body of work, this is a neat new feature that should have upgrade documentation. (and perhaps more user facing documentation?))

docs/src/reference/gremlin-applications.asciidoc
```diff
@@ -942,8 +942,10 @@ The following table describes the various YAML configuration options that Gremli
...
 |authentication.config |A `Map` of configuration settings to be passes to the `Authenticator` when it is constructed.  The settings available are dependent on the implementation. |_none_
-|authentication.enableAuditLog |The available authenticators can issue audit logging messages, binding the authenticated user to his remote socket address and binding requests with a gremlin query to the remote socket address. For privacy reasons, the default value of this setting is false. The audit logging messages are logged at the INFO level via the `audit.org.apache.tinkerpop.gremlin.server` logger, which can be configured using the log4j.properties file. |false
+|authentication.authorizer |The fully qualified classname of an `Authorizer` implementation to use.  Gremlin server does not offer `Authorizer` implementations, but link:https://tinkerpop.apache.org/docs/x.y.z/dev/provider/#_authentication_and_authorization[graph system providers can choose to do so]. |_none_
```
spmallette, line 945: > graph system providers can choose to do so

is it just up to graph system providers? couldn't users write them as well?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java
```diff
@@ -418,7 +426,10 @@ public SerializerSettings() {}
 
         /**
          * Enable audit logging of authenticated users and gremlin evaluation requests.
+         * @deprecated As of release x.y.z, replaced by {@link Settings#enableAuditLog} with slight changes in the
```
spmallette, line 429: I think you can safely put 3.5.0 as the version.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpGremlinEndpointHandler.java
```diff
@@ -189,6 +191,12 @@ public void channelRead(final ChannelHandlerContext ctx, final Object msg) {
             try {
                 logger.debug("Processing request containing script [{}] and bindings of [{}] on {}",
                         requestArguments.getValue0(), requestArguments.getValue1(), Thread.currentThread().getName());
+                if (settings.enableAuditLog) {
```
spmallette, line 194: I haven't thought this through at all but is it desirable that we would double-log here if the user has somehow used both settings (the new and deprecated one as `true`)? I see this sort of code in the `AbstractEvalOpProcessor` as well. Perhaps setting both to `true` should be considered an "illegal" configuration or the setting of the new one to `true` should override the old to `false`? Perhaps, setting the old to `true` and the new to `false` should result in a WARN to the logger to let them know they are using a deprecated settings?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/BytecodeUtil.java
```diff
@@ -18,9 +18,12 @@
  */
 package org.apache.tinkerpop.gremlin.process.traversal.util;
 
-import org.apache.tinkerpop.gremlin.process.traversal.Bytecode;
-import org.apache.tinkerpop.gremlin.process.traversal.TraversalSource;
-import org.apache.tinkerpop.gremlin.process.traversal.TraversalStrategy;
+import org.apache.tinkerpop.gremlin.jsr223.JavaTranslator;
+import org.apache.tinkerpop.gremlin.process.traversal.*;
```
spmallette, line 22: nit: looks like your IDE is making a mockery of our code style by doing a wildcard import 😄

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/BytecodeUtil.java
```diff
@@ -43,4 +47,16 @@ private BytecodeUtil() {}
                         s -> s.getOperator().equals(TraversalSource.Symbols.withStrategies) && clazz.isAssignableFrom(s.getArguments()[0].getClass())),
                 os -> (A) os.getArguments()[0]);
     }
+
+    /**
+     * Parses {@link Bytecode} to find {@link Step} objects in the step instructions.
+     * @return
+     */
+    public static Iterator<Step> findSteps(final Bytecode bytecode, final Class<?> clazz) {
```
spmallette, line 55: I sense that we will do more with `Bytecode` analysis so this class will grow in functionality. Perhaps this PR is not the time to think it all through, but I'd offer that I think that we would want to avoid doing a `Translator` operation here and try to work on pure bytecode when doing these sorts of operations. I see the main use for this method in this PR is to detect lambdas. If that is the case, then I would instead write a `hasLambda()` function here and have it simply dig through the bytecode to look for `Lambda` objects.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/AbstractChannelizer.java
```diff
@@ -200,6 +211,35 @@ private Authenticator createAuthenticator(final Settings.AuthenticationSettings
...
+            throw new IllegalStateException(String.format("Could not create/configure Authorizer %s", authorizer), ex);
+        }
+    }
+
+    private AuthenticatorAuthorizer createAuthenticatorAuthorizer(final Settings settings) {
+        try {
+            final Class<?> clazz = Class.forName(settings.authentication.authenticator);
+            final AuthenticatorAuthorizer authenticatorAuthorizer = (AuthenticatorAuthorizer) clazz.newInstance();
+            authenticatorAuthorizer.setup(settings.authentication.config);
+            authenticatorAuthorizer.configure(settings.authorization.config);
```
spmallette, line 235: `setup()` for authentication and `configure()` for authorization - could we simply unify these and go with `setup()` for both? As long as the two methods have the same name/signature then it should look like one method from an inheritance perspective, no?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/authz/AbstractAuthorizer.java
```diff
@@ -0,0 +1,113 @@
...
+            case Tokens.OPS_EVAL:
+                final String script = (String) msg.getArgs().get(Tokens.ARGS_GREMLIN);
+                authorizeString(user, script, msg);
+                break;
+            case Tokens.OPS_BYTECODE:
+                final Bytecode bytecode;
+                try {
+                    final Object bytecodeObj = msg.getArgs().get(Tokens.ARGS_GREMLIN);
+                    bytecode = bytecodeObj instanceof Bytecode ? (Bytecode) bytecodeObj :
+                            mapper.readValue(bytecodeObj.toString(), Bytecode.class);
```
spmallette, line 60: Under what circumstance does it happen that we get a `OPS_BYTECODE` but its value remains a GraphSON 2.0 string that needs to be deserialized? if so, i would think we would want to figure out how to not let that happen and keep deserialization details out of the `Authorizer`?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/StateKey.java
```diff
@@ -57,4 +58,9 @@ private StateKey() {}
      * The key for the current request.
      */
     public static final AttributeKey<RequestMessage> REQUEST_MESSAGE = AttributeKey.valueOf("request");
+
+    /**
+     * The key for the current {@link AuthenticatedUser}.
+     */
+    public static final AttributeKey<AuthenticatedUser> AUTHENTICATED_USER = AttributeKey.valueOf("authenticatedUser");
```
spmallette, line 65: nice. looks like this body of work will also close:

https://issues.apache.org/jira/browse/TINKERPOP-2418

We will need to remember to close that issue too when this merges

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerAuthIntegrateTest.java
```diff
@@ -56,7 +56,7 @@ public Settings overrideSettings(final Settings settings) {
         final Settings.AuthenticationSettings authSettings = new Settings.AuthenticationSettings();
         authSettings.authenticator = SimpleAuthenticator.class.getName();
 
-        // use a credentials graph with one user in it: stephen/password
+        // use a credentials graph with two users in it: stephen/password and marko/rainbow-dash
```
spmallette, line 59: i suppose the addition of the `marko` user was the change here:

https://github.com/apache/tinkerpop/pull/1308/files#diff-a73da218a425ffa2b99f8c287c3ff3477426bb72748d0c21894d5067485b12a8R47

just a new user...is that right? perhaps now would be a smart time to add a `README.md` to `gremlin-server/data` to remind us what the contents of those files are. i'm not even sure what `gremlin-server/data/sample.kryo` contains anymore.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/authz/Whitelist.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Optional;
+import java.util.Set;
+
+
+/**
+ * Whitelist for the WhitelistAuthorizer as configured by a YAML file.
+ *
+ * @author Marc de Lignie
+ */
+public class Whitelist {
```
spmallette, line 39: I know this is just for testing but could you please change the naming (to this class and related ones) to `AllowList`?

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/authz/WhitelistAuthorizer.java
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * Authorizes a user per request, based on a whitelist that grants access to {@link TraversalSource} instances for
+ * bytecode requests and to gremlin server's sandbox for string requests and lambdas. The {@link
+ * AuthenticationSettings}.config must have an authorizationWhitelist entry that contains the filename of the whitelist.
+ * This authorizer is for demonstration purposes only. It does not scale well in the number of users regarding
+ * memory usage and administrative burden.
+ *
+ * @author Marc de Lignie
+ */
+public class WhitelistAuthorizer extends AbstractAuthorizer {
```
spmallette, line 42: This is a fairly advanced implementation for testing. It should provide a nice example for folks to learn from. would it be good to write user docs around this example and then snippet this code explicitly in the reference documentation?

spmallette:
I gave this a review - big body of work @vtslab 🥇 

i think my review will give you a few things to chew on as this closes in on its final state. I still see a bunch of code formatting and documentation concerns I didn't comment on, but figured those could wait for your next iteration, but other than that I think I'm feeling pretty positively about the direction. 

Also, thanks to @QwentB for putting some thoughts in here. Feedback is appreciated.

docs/src/dev/provider/index.asciidoc
vtslab: Fair - I must admit that this was the easiest way to start with the documentation. Adding to the ref docs implies reorganizing the gremlin server chapter because the security section is already way too big.

docs/src/reference/gremlin-applications.asciidoc
vtslab: Indeed, if we document in the ref docs we do not need the provider term any more.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpGremlinEndpointHandler.java
vtslab: Agreed. I thought about this, but this check ends up in the gremlin server startup code, which I had not touched yet... Good to see that you are not afraid to touch your own code!

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/BytecodeUtil.java
vtslab: OK, I will do that. Note that the bytecode analysis of lambdas depends on magic strings that will come from the JavaTranslator anyway. So, this implies moving the magic strings to ByteUtil and also touching JavaTranslator (for the better!)

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/AbstractChannelizer.java
vtslab: This refers to my earlier question (that you possibly overlooked) whether we should have this AuthenticatorAuthorizer interface anyway. Only reason for its existence is for the Authorizer to have easy access to any user details not present in the AuthenticatedUser (QwentB's request). The user or provide could realize this in an alternative way by building or wrapping an Authenticator implementation into a singleton.
Agreed, that the setup/configure thing is too ugly too keep, but is also ugly to not give the AuthentionSetting and AuthorizationSetting their own config map. Not relevant if we do away with the AuthenticatorAuthorizer.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/authz/AbstractAuthorizer.java
vtslab: I will look into this, probably an oversight or some old test configs.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerAuthIntegrateTest.java
vtslab: Yes, I picked this user from the credentials-dsl example from the ref docs...
I better leave this to you, because this also relates to preparing the console and server distributions.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/authz/Whitelist.java
vtslab: Sure!

docs/src/dev/provider/index.asciidoc
spmallette: As expansive as our documentation is, we still continue to get feedback that it is confusing, incomplete, etc. Perhaps reorganization wouldn't be a really bad thing except that it means "more work". Let's see what your next iteration of this PR looks like and we can discuss further then.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/authz/WhitelistAuthorizer.java
vtslab: Fine to me: explain by example!

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/BytecodeUtil.java
spmallette: > Note that the bytecode analysis of lambdas depends on magic strings

I'm not sure what you mean by that. I would expect the `Bytecode` object to contain an actual `org.apache.tinkerpop.gremlin.util.function.Lambda` object within it. You just need to detect that, no?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/AbstractChannelizer.java
spmallette: I'm +0 with the `AuthenticatorAuthorizer` - whatever you decide on that is fine by me. I'm more concerned in just naming those methods the same thing.  I probably shouldn't have written that the methods should have the same signature I guess...just the same name. I guess we would go with `setup()` which is somewhat at odds with all our other things which use `configure()` but meh...it's fine

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/BytecodeUtil.java
vtslab: That was an excellent pointer and brought me to BytecodeHelper.getLambdaLanguage(bytecode).isPresent(). This means that I will not touch the BytecodeUtil code any more. 
You may want to merge the BytecodeUtil and  BytecodeHelper classes in a separate PR, so that you can properly deprecate one in 3.4.x and get things right for 3.5.0 (assuming there is no valid reason for these classes to both exist). In a large codebase it is easy to overlook the other if you arrive at one via your IDE goTo.

spmallette: yeah - thanks for pointing that out. i shouldnt have added `BytecodeUtil`....

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerAuthIntegrateTest.java
vtslab: Thanks, will do that.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/BytecodeUtil.java
spmallette: I've implemented TINKERPOP-2462 on 7325846aeecd8ae86f3df1565def64df1bc10b13 - going to remove `BytecodeUtil` completely

spmallette: Removed here: 6e47f532933b8a4d0ec25b8aeea8cd157584c358 - when you rebase it will be gone.

vtslab: Great! Over here, documentation is becoming completer.

vtslab:
I just discovered from the code comments that SubgraphStrategy has an unusual implementation for concatenating two withStrategies() steps: the second one simply supersedes the first one instead of &&-ing the predicates of the two SubgraphStrategies as one would expect. I can confirm this behavior in gremlin console. 
Now that we want to advocate the use of SubgraphStrategy for authorization, this behavior requires at least some lines and an example in the traversal ref docs. Personally, I would not object against modifying the semantics of SubgraphStrategy concatenation, thus introducing a breaking change. **What is your preference what to do and in what ticket/PR?** I have not checked yet whether &&-ing the predicates is easy to implement.

vtslab:
Another consideration is the authorization of OLAP requests (initially also mentioned by @FlorianHockmann). For HadoopGraph it seems sufficient to give some users access and others not. Here, one could use a new ComputerDenyStrategy.
For TinkerGraphComputer and FulgoraGraphComputer that run inside the Gremlin Server instance, OLAP also might be legitimate for some users, but here you would like to restrict the number of workers that a client can request, e.g. to 4 to give some speed up with respect to OLTP, but not drain the server. Here, one would rather have some WorkerRestrictStrategy. Of course, the two could coincide if you allow for WorkerRestrictStrategy.build.set(0).create(). Any opinions?

spmallette:
> I just discovered from the code comments that SubgraphStrategy has an unusual implementation for concatenating two withStrategies() steps: the second one simply supersedes the first one instead of &&-ing the predicates of the two SubgraphStrategies as one would expect. 

I'm not sure that we have any strategy that is designed in such a way that more than one instance can be added. Even `OptionsStrategy` has special handling such that calls to `with()` find the strategy first and modify it rather than adding multiple instances with the expectation that they merge. 

> Personally, I would not object against modifying the semantics of SubgraphStrategy concatenation, thus introducing a breaking change. What is your preference what to do and in what ticket/PR? I have not checked yet whether &&-ing the predicates is easy to implement.

Perhaps we should make the implicit rule of a single strategy instance explicit somehow. All strategies might not be mergeable as easily as say `OptionsStrategy` so perhaps we could create an interface that provides a method to merge them. If that interface isn't present on a strategy added more than once we could raise an error. I don't know if `SubgraphStrategy` will merge  predicates well...i guess it would just wrap the predicate in `and()`? That's already a really complex strategy...would hate to make it more so. 

Created https://issues.apache.org/jira/browse/TINKERPOP-2473 for further consideration/discussion.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpGremlinEndpointHandler.java
vtslab: Settings.java now logs an appropriate warning.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/StateKey.java
vtslab: There is a warning in the changelog you cannot miss before the release.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerAuthIntegrateTest.java
vtslab: Done

vtslab:
At this moment, I have no loose ends left, so I call this the end of the present iteration. I want to mention one particular choice: I stopped considering adding authorization for the WsAndHttpChannelizer for the following reasons:

- the WsAndHttpChannelizer is an almost undocumented feature
- its implementation is already kind of a hack
- authorization depends on the Authenticated user, which is provided independently on the http and ws channels and might even be different

Of course, it does not mean that it could not be added later on.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpBasicAuthorizationHandler.java
```diff
@@ -0,0 +1,116 @@
...
+    public void channelRead(final ChannelHandlerContext ctx, final Object msg) {
+        if (msg instanceof FullHttpMessage){
+            final FullHttpMessage request = (FullHttpMessage) msg;
+            try {
+                user = ctx.channel().attr(StateKey.AUTHENTICATED_USER).get();
+                if (null == user) {    // This is expected when using the AllowAllAuthenticator
+                    user = AuthenticatedUser.ANONYMOUS_USER;
+                }
+                // ToDo: move getRequestArguments to a new preceding pipeline step in the Channelizer, but @Stephen,
+                //       how about the sendAndCleanupConnection logic in HttpGremlinEndpointHandler?
```
spmallette, line 75: As they are all static methods I think you could refactor to create a small final utility class to house them - `HttpUtil` or something like that?

spmallette:
Thanks for all the changes on this. I will give it another review in greater detail after 3.4.9 is officially released.

docs/src/dev/provider/index.asciidoc
```diff
@@ -1147,25 +1147,43 @@ one key value pair present (since only one `Traversal` is being submitted, there
...
-that contains the password.  The password should be either, an encoded sequence of UTF-8 bytes, delimited by 0
-(US-ASCII NUL), where the form is : `<NUL>username<NUL>password`, or a Base64 encoded string of the former (which
-in this instance would be `AHVzZXJuYW1lAHBhc3N3b3Jk`).  Should Gremlin Server be able to authenticate with the
-provided credentials, the server will return the results of the original request as it normally does without
-authentication.  If it cannot authenticate given the challenge response from the client, it will return `UNAUTHORIZED`
-(status code `401`).
+in order to authenticate.  Gremlin Server supports the "PLAIN" SASL mechanism, which is a cleartext
+password system, for all link:https://tinkerpop.apache.org/docs/current/tutorials/gremlin-language-variants/[Gremlin Language Variants].
+Other SASL mechanisms supported for selected clients are listed in the
+link:https://tinkerpop.apache.org/docs/current/reference/#security[security section of the Gremlin Server reference documentation].
```
spmallette, line 1157: Please use the template variable for the documentation version in your link, thus: `https://tinkerpop.apache.org/docs/x.y.z/reference/#security`

docs/src/dev/provider/index.asciidoc
```diff
@@ -1147,25 +1147,43 @@ one key value pair present (since only one `Traversal` is being submitted, there
...
+In addition to authenticating users at the start of a connection, Gremlin Server allows providers to authorize users on
+a per request basis. If
+link:https://tinkerpop.apache.org/docs/x.y.z/reference/#_configuring_2[a java class is configured] that implements the
+link:https://tinkerpop.apache.org/javadocs/x.y.z/full/org/apache/tinkerpop/gremlin/server/authz/Authorizer.html[Authorizer interface],
+Gremlin Server passes each request to this `Authorizer`. The `Authorizer` can deny authorization for the request by
+throwing an exception and Gremlin Server returns `UNAUTHORIZED` (status code `401`) to the client. The `Authorizer`
+authorizes the request by returning the original request or the request with some additional constraints. Gremlin Server
+proceeds with the returned request and on its turn returns the result of the request to the client. More details on
+implementing authorization can be found in the
+link:https://tinkerpop.apache.org/docs/current/reference/#security[reference documentation for Gremlin Server security].
```
spmallette, line 1181: Please use the template variable for the documentation version in your link, thus: `https://tinkerpop.apache.org/docs/x.y.z/reference/#security`

docs/src/dev/provider/index.asciidoc
vtslab: You are right. Although a bit messy to also put in in this PR, I caught a few other instances while at the job. Just say it, if you do not want these corrections here.

vtslab:
For the HttpBasicAuthorizationHandler I chose to make it consistent with the HttpGremlinEndpointHandler because it is valid behaviour to alternate permitted and denied requests while keeping the http connection alive. Although, I added a test GremlinServerAuthzIntegrateTest#shouldKeepAuthorizingWithHttpTransport, some of the code in HttpHandlerUtil#sendError is not really tested because it only seems to affect client behaviour and gremlin server performance.

spmallette:
I'm not sure that I have any other points to make on this. I'll merge this tomorrow.  Thanks for bringing this feature all the way to where it is. Nicely done.

VOTE +1

vtslab:
VOTE +1 Found one last inconsistency in the new docs section. In the mean time a minor merge conflict has appeared that I chose not to solve any more, given your planning for tomorrow.

I assume that you realize that the adapted ref docs section on Traversal Strategies still mentions their behaviour neutrally as "inconsistent" upon multiple application (and the AllowListAuthorizer has "touchesXyz" filters to mitigate this behaviour). We created a JIRA ticket to resolve that separately and that does not preclude merging the present PR.

Thanks for all support from you and @QwentB in bringing this feature to fruition.

PR #1309, opened 2020-08-08
TINKERPOP-2369 Replace Connection on server initiated Channel close

divijvaidya:
https://issues.apache.org/jira/browse/TINKERPOP-2369

With this change, the following behaviour are modified:
1. If server closes a `Channel`, the client would detect it and replace the `Connection`. 
    Prior to this change, the client would do nothing on server initiated `Channel` close (applicable for cases when there were 
    no active requests on the channel). This would cause the load balancer to allocate requests to an already closed 
    Connection.
2. Do not throw a NPE when server sends a response for a request which is not registered with the client. Ignore it silently.
3. Ensure that replacement of a `Connection` is done only once. Prior to this change, for each failure of request using the 
    `Connection`, a new connection was being created.

**Testing**
Added tests which fail before the fixes and then succeed after the changes.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java
@@ -18,17 +18,15 @@
  */
 package org.apache.tinkerpop.gremlin.driver;
 
-import io.netty.handler.codec.CodecException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.apache.tinkerpop.gremlin.driver.exception.ConnectionException;
 import org.apache.tinkerpop.gremlin.driver.message.RequestMessage;
 import io.netty.bootstrap.Bootstrap;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.socket.nio.NioSocketChannel;
-import org.slf4j.Logger;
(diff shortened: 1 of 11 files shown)


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java
```diff
@@ -191,14 +209,6 @@ Client getClient() {
         return future;
     }
 
-    public void close() {
```
divijvaidya, line 194: This is an unused function. Removed to simplify code.

divijvaidya:
This is ready for review.

spmallette:
Thanks for this - a few minor comments/nits:

1. Could you please rename tests that start with "Test*" to our more standard "should*"
2. Have a look at where you might add more `final` declarations to match the code style.
3. I like the idea of integration tests with the `SimpleWebSocketServer` - very smart
4. I'm surprised you found as many places as you did where `Cluster.close()` wasn't called.
5. I don't see where the semantics of any of the `GremlinDriverIntegrateTest` tests changed so it seems you accomplished this fix without breaking behavioral changes in the driver...that's nice.
6. Maybe I missed it but was there a test in Gremlin Server to validate any of this change - perhaps it was already better tested by way of your `SimpleWebSocketServer` tests?
7. I assume you will polish up the commit history a bit on merge and squash things down to a few (or one) commits?

PR #1310, opened 2020-08-12
add getter method for GraphSONMessageSerializer

rdtr:
Very simple changes. I just want a way to retrieve GraphSONMapper from these Serializer.

Background:

Our implementation is heavily inspired by [HttpGremlinEndpointHandler.java](https://github.com/apache/tinkerpop/blob/6c7be04be215f7877cca12b2d681d8cd81b7fb2f/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpGremlinEndpointHandler.java) and we can retrieve MessageTextSerializer there, but this can only serialize / deserialize the whole response message.

I want to have capability to serialize / deserialize any element (e.g. some Vertex or Edge only) in the graph in HttpHandler and right now I define the same GraphSONMapper separately. But if I can just grab the GraphSONMapper I already setup for MessageTextSerializer, it would be good.

This mapper object can't be modified so it should also be safe. Please let me know if this addition is OK, if yes I will add tests + change log.

Thanks !

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ser/AbstractGraphSONMessageSerializerV1d0.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ser/AbstractGraphSONMessageSerializerV1d0.java
@@ -154,6 +154,10 @@ public abstract class AbstractGraphSONMessageSerializerV1d0 extends AbstractMess
                 .version(GraphSONVersion.V1_0);
     }
 
+    public ObjectMapper getMapper() {
+        return this.mapper;
+    }
+
     public final static class GremlinServerModule extends SimpleModule {
         public GremlinServerModule() {
             super("graphson-gremlin-server");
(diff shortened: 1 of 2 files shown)


spmallette:
I can see why this change might be necessary, though I think my preference might be to return the TinkerPop `Mapper` object instead. In that way getting a "mapper" becomes generalized to the `MessageSerializer` interface. The `MessageSerializer` would have:

```java
Mapper getMapper();
```

and then you could just construct your own from that:

```java
Mapper<ObjectMapper> mapperObjectMapper = graphsonMessagerSerializer.getMapper();
ObjectMapper objectMapper = mapperObjectMapper.createMapper()
```

I imagine there are some generics to sort out in all this and I'm not sure if it can be done without breaking changes (so perhaps only a suitable change to 3.5.0). It would also require GraphBinary to have a `Mapper` implementation. There could be other problems too. 

I just thought I'd mention this approach as it is a bit more robust than just adding getters for specific class types. Any thoughts on the matter?

rdtr:
Thank you for the suggestion @spmallette . I am happy to update my code but could you elaborate how I can add `Mapper` implementation to GraphBinary ? Can I just put placeholder methods like `GraphML` ?

GraphSon has `ObjectMapper`, Gyro has `Kyro`, I am not sure what the counter of them for GraphBinary and what to expect to implement.

spmallette:
Well, GraphBinary has `GraphBinaryReader` and `GraphBinaryWriter` so to make a `GraphBinaryMapper` implementation you'd have to create a class to hold both of those - perhaps that's just a `GraphBinaryReaderWriter` and that way you could do:

```java
public class GraphSONMapper implements Mapper<GraphBinaryReaderWriter> {

    @Override
    public GraphBinaryReaderWriter createMapper() {
       ....
    }
```

Not sure what it looks like from there, but that would at least make all the MessageSerializers consistent for retrieving these sorts of things. Does that give you some idea on how to move forward?

spmallette:
Merged and included 3ba462fca9a1fc43c559d1b328225bde649b6f96

PR #1311, opened 2020-08-15
TINKERPOP-2406 Delegate processing from event loop to worker threads

divijvaidya:
https://issues.apache.org/jira/browse/TINKERPOP-2406

**Motivation**
Netty client has two thread pools. Event loop thread pool which is responsible for handling socket events and another is a worker thread pool which is used to delegate busy work from event loop threads so that they can continue handling socket events. When programming in Netty, ideally, the event loop should never be blocked and must be kept free to respond to events.

In case of exceptional completion of readComplete which could be triggered by a server sending an error code in the response, the event loop thread is used to perform the clean up logic which could involve an expensive replaceConnection. This blocks the event loop.

**Change**
Delegate handling of erroneous response from the server (which might involve connection replacement) to the worker thread pool. After this change the event loop will have lesser pressure and event handling will become faster.


spmallette:
Could you please add a ticket in JIRA for this change? Aside from that, this makes sense - further freeing the event loop by delegating off tasks that don't need to be there is a good move. Please squash before merging. 

VOTE +1

PR #1312, opened 2020-08-20
Relax access for class members and add getters/setters

divijvaidya:
**Motivation**

Relax access for class members and add getters/setters to ease provider integration.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/GroupCountStep.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/GroupCountStep.java
@@ -41,7 +41,7 @@ import java.util.function.BinaryOperator;
 /**
  * @author Marko A. Rodriguez (http://markorodriguez.com)
  */
-public class GroupCountStep<S, E> extends ReducingBarrierStep<S, Map<E, Long>> implements TraversalParent, ByModulating {
+public final class GroupCountStep<S, E> extends ReducingBarrierStep<S, Map<E, Long>> implements TraversalParent, ByModulating {
 
     private Traversal.Admin<S, E> keyTraversal = null;
 
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/GroupCountStep.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/GroupCountStep.java
@@ -41,7 +41,7 @@ import java.util.function.BinaryOperator;
 /**
  * @author Marko A. Rodriguez (http://markorodriguez.com)
  */
-public final class GroupCountStep<S, E> extends ReducingBarrierStep<S, Map<E, Long>> implements TraversalParent, ByModulating {
+public class GroupCountStep<S, E> extends ReducingBarrierStep<S, Map<E, Long>> implements TraversalParent, ByModulating {
 
     private Traversal.Admin<S, E> keyTraversal = null;
 
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
@@ -63,6 +63,7 @@ public class DefaultTraversal<S, E> implements Traversal.Admin<S, E> {
     protected TraversalStrategies strategies;
     protected transient TraverserGenerator generator;
     protected Set<TraverserRequirement> requirements;
+
     protected boolean locked = false;
     protected Bytecode bytecode;
 
@@ -346,6 +347,10 @@ public class DefaultTraversal<S, E> implements Traversal.Admin<S, E> {
     @Override
     public void setGraph(final Graph graph) {
         this.graph = graph;
+    }
+
+    public void setLocked(boolean locked) {
+        this.locked = locked;
     }
 
     @Override
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
@@ -348,11 +348,7 @@ public class DefaultTraversal<S, E> implements Traversal.Admin<S, E> {
     public void setGraph(final Graph graph) {
         this.graph = graph;
     }
-
-    public void setLocked(boolean locked) {
-        this.locked = locked;
-    }
-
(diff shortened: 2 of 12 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/GroupCountStep.java
```diff
@@ -41,7 +41,7 @@
 /**
  * @author Marko A. Rodriguez (http://markorodriguez.com)
  */
-public final class GroupCountStep<S, E> extends ReducingBarrierStep<S, Map<E, Long>> implements TraversalParent, ByModulating {
+public class GroupCountStep<S, E> extends ReducingBarrierStep<S, Map<E, Long>> implements TraversalParent, ByModulating {
```
spmallette, line 44: We had originally made many of the steps `final` in the early days of TinkerPop 3.x to prevent extension by providers as they were changing frequently and we wanted to be able to introduce changes without worrying about breaking APIs. I'd imagine that most of the steps have largely stabilized at this point so perhaps that worry is no longer as much of a concern.  However, I still feel like these core step classes are TinkerPop's implementations and I'd not want to see our development blocked for fear of a breaking API change that affects providers (a behavioral change is different as it affects users and we would obviously be sensitive to that). 

Perhaps it would be good to start a DISCUSS thread on this on the dev list to see if there is any concerns with removing `final` for all steps with the idea that breaking API changes become a possibility for providers on patch releases if they choose to directly extend these core step classes. wdyt?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
```diff
@@ -350,6 +351,10 @@ public void setGraph(final Graph graph) {
         this.graph = graph;
     }
 
+    public void setLocked(boolean locked) {
```
spmallette, line 354: what is the intent of this change? i don't know what folks might do with that, but "unlocking" a traversal might open up bad/unexpected things for users....

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerVertex.java
```diff
@@ -40,7 +40,7 @@
 /**
  * @author Marko A. Rodriguez (http://markorodriguez.com)
  */
-public final class TinkerVertex extends TinkerElement implements Vertex {
```
spmallette, line 43: Could you please talk about the cases where folks would extend `TinkerVertex` (or `TinkerGraph` more generally)? We have one very specific one with https://github.com/ShiftLeftSecurity/tinkergraph-gremlin but i'm not sure that's relevant anymore with its successor in https://github.com/ShiftLeftSecurity/overflowdb

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/GroupCountStep.java
divijvaidya: Let me change the scope of this PR to only change visibility of certain members. We can pick up removing the final keyword after a discussion in email list.

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerVertex.java
divijvaidya: It could be extended by vendors to create a custom testing suite OR it could be used by vendors to build a TinkerGraph++ offering. IMO, we should probably relax the final requirement on classes unless it is necessary or has some specific purpose.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
divijvaidya: It's a separate conversation. I will soon raise a PR for that. I think that the way in which we apply strategies (first to all steps and then to each children one at a time), could be improved to apply all strategies at the same time to the entire traversal and sub-traversal. I will remove this change from this PR, pending further discussion on a new PR.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/GroupCountStep.java
divijvaidya: Resolving this comment since I removed the file being discussed from the scope of this PR.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
divijvaidya: Resolving this comment since I removed the file being discussed from the scope of this PR.

spmallette: note that strategy application changed/improved on `master` for 3.5.0 already:

https://tinkerpop.apache.org/docs/3.5.0-SNAPSHOT/upgrade/#_traversalstrategy_application

looking forward to hearing what you think we might change but we will need to be careful not to introduce a change that inadvertently changes traversal behavior because we change strategy application order/approach. that's why we saved the above referenced change for 3.5.0.

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerVertex.java
spmallette: I just sorta realized this issue is still hanging around here without any added reply. I did start a discussion on this on dev and didn't really get objections but I'd prefer to do this particular TinkerGraph change around a JIRA (with a bit more thought as to the repercussions. If you just back out this one change then the rest of this is good to merge from my perspective. If you agree I will create a JIRA for the TinkerGraph issue. wdyt?

divijvaidya: Sure. I will move the TinkerGraph out of the scope of this PR as well so that things become simpler.

divijvaidya: I have removed changes in TinkerGraph from this PR and this is ready for review.

spmallette:
VOTE +1

PR #1317, opened 2020-08-27
TINKERPOP-2335 Set node.js support at 10 for 3.5.0

spmallette:

Upgraded frontend plugin and centralized configuration a bit. 

VOTE +1

### DIFF
--- a/gremlin-javascript/pom.xml
+++ b/gremlin-javascript/pom.xml
@@ -29,10 +29,21 @@ limitations under the License.
         <maven.test.skip>false</maven.test.skip>
         <skipTests>${maven.test.skip}</skipTests>
         <gremlin.server.dir>${project.parent.basedir}/gremlin-server</gremlin.server.dir>
+        <npm.version>6.14.6</npm.version>
+        <node.version>v10.22.0</node.version>
     </properties>
     <build>
         <directory>${basedir}/target</directory>
         <finalName>${project.artifactId}-${project.version}</finalName>
+        <pluginManagement>
+            <plugins>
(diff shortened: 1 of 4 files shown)


gremlin-javascript/pom.xml
```diff
@@ -29,10 +29,21 @@ limitations under the License.
         <maven.test.skip>false</maven.test.skip>
         <skipTests>${maven.test.skip}</skipTests>
         <gremlin.server.dir>${project.parent.basedir}/gremlin-server</gremlin.server.dir>
+        <npm.version>6.14.6</npm.version>
+        <node.version>v10.22.0</node.version>
```
jorgebay, line 33: Good idea to extract it into a property.

gremlin-javascript/src/main/javascript/gremlin-javascript/package.json
```diff
@@ -40,4 +40,4 @@
   "engines": {
     "node": ">=6"
```
jorgebay, line 41: We should state `"node": ">= 10"` in the package.json

jorgebay (commented):
A couple of comments below.

In the meantime, I'll try it locally.

gremlin-javascript/src/main/javascript/gremlin-javascript/package.json
spmallette: oops - shouldn't have missed that - fixed now

jorgebay (approved):
I've tested it locally 👍 

Looks good, VOTE +1

PR #1318, opened 2020-08-27
TINKERPOP-2410 Release server threads waiting on connection if the connection is dead. 

divijvaidya:
https://issues.apache.org/jira/browse/TINKERPOP-2410

Consider a situation where the server has sent some results back to the client and is waiting for Netty channel to be writable again before sending rest of the results. Let's assume that the client dies (or is closed) at this instant without consuming the entire set of results. In this case, the server threads will continue to wait for the channel to become writable (which it never will since client is dead) until the timeout is hit.

With this change, if the connection becomes inactive, the loop will break and free up the threads.

## Testing
We don't have any test mechanism to check the number of busy server threads. Hence, no tests are added to repro this case. Any suggestions on how to add tests for such scenarios are welcome!

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -38,6 +38,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Avoid creating unnecessary detached objects in JVM.
 * Added support for `TraversalStrategy` usage in Javascript.
 * Added `Traversal.getTraverserSetSupplier()` to allow providers to supply their own `TraverserSet` instances.
+* Release server threads waiting on connection if the connection is dead.
 
 [[release-3-4-8]]
 === TinkerPop 3.4.8 (Release Date: August 3, 2020)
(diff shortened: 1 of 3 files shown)


spmallette:
Getting a failing test:

```
[ERROR] Failures: 
[ERROR]   GremlinServerSessionIntegrateTest.shouldBlockAdditionalRequestsDuringClose:209 
Expected: a collection containing "INFO - Rolling back open transactions on graph before killing session: shouldBlockAdditionalRequestsDuringClose\n"
     but: was "WARN - Timeout while trying to close connection on shouldBlockAdditionalRequestsDuringClose - force closing - server will close session on shutdown or expiration.\n"
[INFO] 
[ERROR] Tests run: 270, Failures: 1, Errors: 0, Skipped: 1
```

I think you just need to rebase since you fixed this on 3.4-dev right?

spmallette:
I see you brought in the fix from 3.4-dev and the test failure is cleaned up. I'm good with this change - builds with:

`mvn clean install -pl gremlin-server,gremlin-console -DskipIntegrationTests=false -DincludeNeo4j`

VOTE +1 (i assume you won't bring in that merge commit at d632f2a to the release branches and will squash the "changelog" commit)

spmallette:
Just a thought - before you got to merge this, you might want to test it on `master` to be sure all the tests are still ok.

PR #1323, opened 2020-09-03
TINKERPOP-2415 Avoid unnecessary detached objects if not required

divijvaidya:
https://issues.apache.org/jira/browse/TINKERPOP-2415

This change adds some defensive checks to avoid creating additional detached objects when not required.


spmallette:
nice one - i assume the merge commit won't go to the release branches - VOTE +1

PR #1325, opened 2020-09-07
[TINKERPOP-2416] MultiIterator should implement AutoCloseable

rdtr:
As described in the [JIRA](https://issues.apache.org/jira/browse/TINKERPOP-2416), currently `MultiIterator` does not implement `AutoCloseable`. This potentially introduces resource leak when we have multiple iterators that implement `AutoCloseable` and combine them using `IteratorUtils#concat`. The caller may expect all iterators to be closed automatically by calling `CloseableIterator#closeIterator` but right now `MultiIterator` doesn't implement AutoCloseable so even the underlying iterators are auto-closeable, they are never be closed.

This simple PR just adds AutoCloseable implementation to MultiIterator, close all inner iterators if they implement AutoCloseable.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/iterator/MultiIterator.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/iterator/MultiIterator.java
@@ -19,6 +19,7 @@
 package org.apache.tinkerpop.gremlin.util.iterator;
 
 import org.apache.tinkerpop.gremlin.process.traversal.util.FastNoSuchElementException;
+import org.apache.tinkerpop.gremlin.structure.util.CloseableIterator;
 
 import java.io.Serializable;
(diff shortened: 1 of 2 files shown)


spmallette:
There are a number of places in TinkerPop code that handle `MultiIterator` instances. I guess we don't need to explicitly `close()` those as I don't think they hold expensive resources open (that I am aware of). Any thoughts on that?

rdtr:
I agree that we don't need to explicitly call `close` in TinkerPop's code.

The change is mainly for providers. When overriding TinkerPop's API to retrieve elements from their own internal Storage layer, if the returned iterator holds any reference which must be released in the end, it is the provider's responsibility to call `close` in their own method. 

TinkerPop's base class and TinkerGraph doesn't have the resource issue so we don't need to call `close` there.

spmallette:
VOTE +1

PR #1329, opened 2020-09-17
TINKERPOP-2296 support per-request options in python

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2296

VOTE +1


rdtr:
+1, great step towards driver compatibility across languages

PR #1334, opened 2020-09-27
TINKERPOP-2427 Simplify Netty reference counting

divijvaidya:
https://issues.apache.org/jira/browse/TINKERPOP-2427

### Motivation
We have some incorrect configuration in the code such as:
1. https://github.com/apache/tinkerpop/blob/3.4-dev/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/WebSocketGremlinResponseDecoder.java#L59 (MessageToMessageDecoder automatically decreases the reference count).

1. https://github.com/apache/tinkerpop/blob/3.4-dev/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Handler.java#L270 (SimpleChannelInboundHandler automatically decreases the reference count).

which are fixed by hardcoding values such as:
* https://github.com/apache/tinkerpop/blob/3.4-dev/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/WebSocketClientHandler.java#L89

### Changes
* Do not release reference count in SimpleChannelInboundHandler and MessageToMessageDecoder since they already do that automatically.
* Retain references when we want to pass the message to the next handler.

### Testing
gremlin-driver `mvn clean install -DskipIntegrationTests=false -DincludeNeo4j`
gremlin-server `mvn clean install -DskipIntegrationTests=false -DincludeNeo4j`

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Handler.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Handler.java
@@ -121,6 +121,8 @@ final class Handler {
                 }
                 channelHandlerContext.writeAndFlush(messageBuilder.create());
             } else {
+                // SimpleChannelInboundHandler will release the frame if we don't retain it explicitely.
+                ReferenceCountUtil.retain(response);
                 channelHandlerContext.fireChannelRead(response);
             }
         }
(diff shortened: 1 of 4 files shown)


spmallette:
Glad you were able to improve this code. The `ctx.fireChannelRead(frame.retain(2));` always bugged me. I'd recommend that you merge to `master` and run tests before merging this as I think you will find some conflict there

VOTE +1

PR #1335, opened 2020-09-27
Tinkerpop-2426 Delegate WS internal handling to Netty

divijvaidya:
https://issues.apache.org/jira/browse/TINKERPOP-2426

### Motivation
Let Netty perform the underlying WebSocket related stuff such as handshake, handling ping/pong etc. instead of writing the code ourselves.

### Changes
1. Use Netty's `WebSocketClientProtocolHandler` instead of writing a custom implementation at `WebSocketClientHandler`.
1. Introduce `WsHandshakeTimeout` as a configurable parameter. Set default to 15s.
1. Rename `SimpleWebSocketServer` to more appropriately reflect its functionality in tests.

**Note**: There is no significant functionality change other than increasing handshake time from 10s to 15s. This is increased because Netty's default value for SslHandshake itself is 10s, hence WsHandshake should be larger than that. 

### Testing
gremlin-driver `mvn clean install -DskipIntegrationTests=false -DincludeNeo4j`
gremlin-server `mvn clean install -DskipIntegrationTests=false -DincludeNeo4j`

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -40,6 +40,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Added `Traversal.getTraverserSetSupplier()` to allow providers to supply their own `TraverserSet` instances.
 * Release server threads waiting on connection if the connection is dead.
 * Fixed bug where server closes HTTP connection on request error even if keep alive is set to true.
+* Java driver now delegates handling of WebSocket handshake to Netty instead of custom code. 
 
 [[release-3-4-8]]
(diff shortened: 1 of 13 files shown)


spmallette:
I didn't look at this in detail yet because gremlin-server integration tests are failing in Travis. I've already restarted them once and they failed a second time.

spmallette:
This is a good change - nice that you pushed functionality down to netty so that we don't have to test/manage it 🏆  

Please include CHANGELOG entry and the new `wsHandshakeTimeoutMillis` configuration option should be documented here:

https://tinkerpop.apache.org/docs/current/reference/#_configuration

Assuming those little changes, VOTE +1

PR #1336, opened 2020-10-01
TINKERPOP-2430 + TINKERPOP-2431 - Added recipes to documentation

bechbd:
Added recipes to the documentation for common looping patterns as well as operating on dropping elements (i.e. Get count of dropped elements)


docs/src/recipes/looping.asciidoc
```diff
@@ -0,0 +1,88 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+////
+[[looping]]
+== Looping
+One common use case when working with Gremlin is to perform complex looping statements using the `repeat()` step.  While many of the common patterns for looping within traversals are discussed in the documentation there are several more complex patterns that include additional steps which are also commonly used.  This section attempts to demonstrate how to use some of these more complex measurements.
```
spmallette, line 19: You should be free to use line breaks - asciidoc knows to keep paragraphs together. It's sometimes makes diffs easier to read with some breaks in place.

docs/src/recipes/looping.asciidoc
```diff
@@ -0,0 +1,88 @@
...
+  addE('hasParent').from('b').to('c').
+  addE('hasParent').from('d').to('c').
+  addE('hasParent').from('c').to('e').
+  addE('hasParent').from('e').to('f').
+  addE('hasParent').from('g').to('f').iterate()
+----
+
+=== Conditional Looping with Max Depth
+
+One common situation encountered when writing `repeat()` loop is the need to exit the loop when either a specific condition is met or a maximum depth is reached.  Given that graph above the following traversal demonstrates how to accomplish this type of complex exit condition.  In this example the traversal will start at vertex A and loop through all outgoing edges until it reaches vertex C or it until it has completed 3 loops.
```
spmallette, line 42: "or it until it has completed 3 loops"

docs/src/recipes/looping.asciidoc
```diff
@@ -0,0 +1,88 @@
...
+In the above case the traversal processes the `out().simplePath()` step once to move from vertex A to vertex B, a second time to move from vertex B to vertex C, and then exits the traversal because it satisfies the `hasId('C')` condition since it's reached vertex C.  If we however change that exit condition to be vertex G, we will see that this traversal now will continue to process thorough vertex X and will exit based on reaching the maximum number of loops (3).
+
+[gremlin-groovy,existing]
+----
+g.V('A').
+  repeat(out().simplePath()).
+    until(hasId('G').or().loops().is(3))
+----
+
+The key portion of the above traversal is the `until()` step.  This step is capable of accepting a `Traversal` object will keep the `repeat()` loop traversing until the evaluated output of the traversal is true, which causes the traverser to exit the `repeat()` loop.  Using the methodology allows complex logical conditions to be used as the exit criteria.
```
spmallette, line 60: accepting a `Traversal` object will keep the `repeat()`" - perhaps "object which will"?

docs/src/recipes/operating-on-dropped-elements.asciidoc
```diff
@@ -0,0 +1,27 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+////
+[[operating-on-dropped-elements]]
+== Operating on Dropped Elements
+
+One common scenario that happens when dropping elements using a traversal is the desire to perform some sort of operation on the elements being removed.  This simplest form of this scenario is returning the number of elements being dropped from the traversal, which is shown in the example below from the "modern" graph.
```
spmallette, line 20: "This simplest form of this scenario" - "The simplest"?

docs/src/recipes/operating-on-dropped-elements.asciidoc
```diff
@@ -0,0 +1,27 @@
...
+
+[gremlin-groovy,modern]
+----
+g.V().has('lang', 'java').sideEffect(drop()).count()
+----
+
+While this may seem a bit counterintuitive, what is occurring in this recipe is that the `sideEffect()` step is performing the `drop()` while each of the input traverser is still passed along to the remainder of the traversal, which in this case is a `count()` operation.  This pattern can be extended beyond a simple count to perform a variety of more complex traversals based on the dropped traversal such as removing child nodes in the case of a tree or setting properties on adjacent vertices.
```
spmallette, line 27: This is a good one - A WARNING callout might be good to point out as different graphs may have some altered behavior around the `sideEffect(drop())`. For example, while TinkerGraph is fine to return a `TinkerVertex` on which a `drop()` has been performed, other graphs might not. (at least i'm pretty sure that was the case, but I can't recall offhand which graphs do that)

spmallette:
Thanks for this documentation update - it's nice to get new recipes.

You might want to change the target branch to 3.4-dev so that this recipe gets published with the next release otherwise it will have to wait until 3.5.0 which probably won't be in the next release round. It's fine if you don't, I can bring it to 3.4-dev myself on merge.

The loop section is good - I would add a link from the `repeat()` and `loops()` steps in the `the-traversal.asciidoc` to the recipe. At the bottom of each is an "Additional References" section. Right now they all link to javadocs but I think it would help bring some notice to "Recipes" if we linked from there to relevant/related recipes. I would name the link as "Recipes - Looping" and add them after all the javadoc. 

Other than that, VOTE +1

docs/src/recipes/looping.asciidoc
bechbd: My IDE was doing this automatically so I didn't even notice.  I added them

docs/src/recipes/looping.asciidoc
bechbd: Ack

docs/src/recipes/looping.asciidoc
bechbd: Ack

docs/src/recipes/operating-on-dropped-elements.asciidoc
bechbd: Ack

docs/src/recipes/operating-on-dropped-elements.asciidoc
bechbd: Good call, I added a warning section here

bechbd:
I've updated per your comments, Stephen.  I added the links into the traversal steps but I didn't change the base branch to 3.4-dev so if you can do that on merge it would be appreciated.

spmallette:
Cherry-picked to 3.4-dev as two commits: 43cd897f83509660d4f7064dcef14dd15ef6c9ca and c3a732bf302fdf7421745fd60087fa8d8970a34b

PR #1337, opened 2020-10-05
#TINKERPOP-2432 improve toString for the Bytecode and unit test

sergsalo:
according to the documentation the Bytecode to string should produce valid Bytecode encoding. 
[
 ["V", [1]],
 ["repeat", [[
   ["out", ["knows"]]
   ["hasLabel", ["person"]]]]]
 ["times", [2]]
 ["values", ["name"]]
]

#TINKERPOP-2432

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/test/unit/bytecode-test.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/test/unit/bytecode-test.js
@@ -1,0 +1,40 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+/**
+ * @author Serhiy Salo
+ */
+'use strict';
+
+const assert = require('assert');
+const graph = require('../../lib/structure/graph');
+const graphTraversalModule = require('../../lib/process/graph-traversal');
+const __ = graphTraversalModule.statics;
+
+describe('Bytecode', () => {
+
+    describe('#toString()', () => {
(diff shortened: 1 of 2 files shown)


gremlin-javascript/src/main/javascript/gremlin-javascript/test/unit/bytecode-test.js
```diff
@@ -0,0 +1,40 @@
...
+const graph = require('../../lib/structure/graph');
+const graphTraversalModule = require('../../lib/process/graph-traversal');
+const __ = graphTraversalModule.statics;
+
+describe('Bytecode', () => {
+
+    describe('#toString()', () => {
+        it('should produce valid string representation', () => {
+            const g = new graph.Graph().traversal();
+            const bytecode = g.V().hasLabel("airport").where(__.outE("route").hasId(7753));
```
spmallette, line 35: could you please expand this test to exercise "sourceInstructions" - something like:

```js
g.with_('enabled').withStrategies(new ReadOnlyStrategy()).
  V().hasLabel("airport").where(__.outE("route").hasId(7753))
```

spmallette:
Thank you for providing this fix. 

Could you please re-target this pull request at the 3.4-dev branch (that way the fix will go to the next release at 3.4.9)? On another administrative note, we typically add a CHANGELOG entry that describes the change (which is separate from adding the JIRA, which happens on release). If you could do that, it would polish up this PR, but if not, no worries, I can add it on merge. 

@jorgebay any chance you have a few minutes to review this one?

jorgebay:
@spmallette Sure, thanks for the ping.

gremlin-javascript/src/main/javascript/gremlin-javascript/test/unit/bytecode-test.js
jorgebay: We also need tests for multiple parameters, something like `g.V().has('person','name','marko')...`

jorgebay (commented):
Thanks for your patch !

I think we can solve this by implementing a single [`toJSON`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#toJSON_behavior) on `Traversal`, instead of having a `getBytecodeAsList()` method to iterate through the whole tree.

Then `toString()` method would be something like:

```javascript
toString() {
  return JSON.stringify([
    this.sourceInstructions,
    this.stepInstructions
  ]);
```

Do you think it's worth trying the `toJSON` approach @sergsalo?

sergsalo:
I will check the toJSON and will do what you guys suggested. 

On the other note - have you guys considered of using Typescript for gremlin-javascript?

jorgebay:
> have you guys considered of using Typescript for gremlin-javascript?

I think migrating the gremlin-js codebase to TypeScript would require an effort for little gain (language preferences vary by engineer).

On the other hand, about TypeScript from the user pov, there are definitions: https://www.npmjs.com/package/@types/gremlin

and a ticket to support it within the library: https://issues.apache.org/jira/browse/TINKERPOP-2027

PR #1338, opened 2020-10-06
[TINKERPOP-2433] fix typogarphical error

bobsut:
[[TINKERPOP-2433](https://issues.apache.org/jira/browse/TINKERPOP-2433)] Fix a simple typo


spmallette:
Thanks you. For future reference there is no need to open a JIRA for small documentation fixes like this. If you plan a new section of the reference documentation, are considering extensive contextual changes to some aspect of the javadoc, etc. then a JIRA would probably be good. Otherwise, feel free to submit spelling fixes as a PR without one.

I've cherry-picked this change to 3.4-dev and merged forward to master: 60f65e659855f54d2818697c389eb125c0a547d2

PR #1339, opened 2020-10-06
TINKERPOP-2420 Per Request Options in .NET

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2420

Added per request options in .NET. Opted to not create `RequestOptions` like functionality in .NET since the way the client is exposed allows you to construct a bare `RequestMessage` with reasonable niceness - documented that approach a bit better. Also added documentation for the old GLVs that recently got this feature. 

VOTE +1

### DIFF
--- a/docs/src/reference/gremlin-variants.asciidoc
+++ b/docs/src/reference/gremlin-variants.asciidoc
@@ -216,7 +216,7 @@ List<Vertex> vertices = g.with(Tokens.ARGS_EVAL_TIMEOUT, 500L).V().out("knows").
 ----
 
 The following options are allowed on a per-request basis in this fashion: `batchSize`, `requestId`, `userAgent` and
-`evaluationTimeout` (formerly 'scriptEvaluationTimeout' which is also supported but now deprecated). Use of `Tokens`
+`evaluationTimeout` (formerly `scriptEvaluationTimeout` which is also supported but now deprecated). Use of `Tokens`
 to reference these options is preferred.
 
 anchor:java-imports[]
@@ -685,11 +685,11 @@ Some connection options can also be set on individual requests made through the 
 
 [source,python]
 ----
-vertices = g.with_('evaluationTimeout', 500L).V().out('knows').toList()
+vertices = g.with_('evaluationTimeout', 500).V().out('knows').toList()
 ----
 
 The following options are allowed on a per-request basis in this fashion: `batchSize`, `requestId`, `userAgent` and
-`evaluationTimeout` (formerly 'scriptEvaluationTimeout' which is also supported but now deprecated).
+`evaluationTimeout` (formerly `scriptEvaluationTimeout` which is also supported but now deprecated).
 
 anchor:python-imports[]
 [[gremlin-python-imports]]
@@ -966,7 +966,7 @@ result_set = client.submit('g.V().repeat(both()).times(100)', result_options={'e
 ----
 
 The following options are allowed on a per-request basis in this fashion: `batchSize`, `requestId`, `userAgent` and
-`evaluationTimeout` (formerly 'scriptEvaluationTimeout' which is also supported but now deprecated).
+`evaluationTimeout` (formerly `scriptEvaluationTimeout` which is also supported but now deprecated).
 
 [[gremlin-python-dsl]]
 === Domain Specific Languages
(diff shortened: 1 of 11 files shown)


gremlin-dotnet/src/Gremlin.Net/Driver/Tokens.cs
```diff
@@ -74,24 +79,25 @@ public class Tokens
         public static string ProcessorSession = "session";
 
         /// <summary>
-        ///     Argument name that allows to defines the number of iterations each ResponseMessage should contain - overrides the
-        ///     resultIterationBatchSize server setting.
+        ///     Argument name that allows to definition of the number of iterations each ResponseMessage should
```
FlorianHockmann, line 82: (nitpick) _allows to definition_ doesn't sound right to me (?)

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Process/Traversal/DriverRemoteConnection/GraphTraversalTests.cs
```diff
@@ -201,22 +204,21 @@ public void ShouldUseBindingsInTraversal()
...
                 {"y", true}
             };
             var g = AnonymousTraversalSource.Traversal().WithRemote(connection);
-
-            var b = new Bindings();
+            
             var countWithStrategy = g.WithStrategies(new OptionsStrategy(options)).V().Count().Next();
             Assert.Equal(6, countWithStrategy);
+            
+            Assert.Equal(ResponseStatusCode.ServerTimeout, Assert.Throws<ResponseException>(() =>
```
FlorianHockmann, line 218: (nitpick) I think it's easier to read and understand this when the two asserts are separated:

```c#
var responseException = Assert.Throws<ResponseException>(() =>
    g.With("y").With("x", "test").With(Tokens.ArgsEvalTimeout, 10).Inject(1)
        .SideEffect(Lambda.Groovy("Thread.sleep(10000)")).Iterate());
Assert.Equal(ResponseStatusCode.ServerTimeout, responseException.StatusCode);
```

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1121,6 +1121,18 @@ the remote end.
 include::../../../gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Docs/Reference/GremlinVariantsTests.cs[tags=connecting]
 ----
 
+Some connection options can also be set on individual requests made through the using `With()` step on the
```
FlorianHockmann, line 1124: _the using `With()` step_ -> _using the `With()` step_

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1121,6 +1121,18 @@ the remote end.
...
+`TraversalSource`. For instance to set request timeout to 500 milliseconds:
+
+[source,csharp]
+----
+var l = g.With(Tokens.ArgsEvalTimeout, 500).V().Out("knows").Count();
```
FlorianHockmann, line 1129: I'm not sure about this but shouldn't we use a terminal step here to actually get the count(s)? Especially new users would probably expect here to have the count stored in the variable and not a traversal object.

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1246,6 +1258,20 @@ var gremlinServer = new GremlinServer("localhost", 8182);
 var client = new GremlinClient(gremlinServer, sessionId: Guid.NewGuid().ToString()))
 ----
 
+==== Per Request Settings
+
+The `GremlinClient.submit()` functions accept an option to build a raw `RequestMessage`. A good use-case for this
```
FlorianHockmann, line 1263: (nitpick) **S**ubmit (to stay with the .NET naming conventions)

FlorianHockmann (commented):
LGTM, I only found a few nits and I also agree that `RequestOptions` shouldn't be necessary here, especially since we want to discourage people from using Gremlin scripts in the first place so that doesn't need to be so easy to use like the traversal API.

spmallette:
Addressed all your comments @FlorianHockmann - thanks!

FlorianHockmann (approved):
VOTE +1

PR #1343, opened 2020-10-08
TINKERPOP-2440 Simplify driver by delegating keepAlive logic to Netty

divijvaidya:
https://issues.apache.org/jira/browse/TINKERPOP-2440

Java driver can be simplified by delegating the keepAlive logic to Netty's IdleStateHandler instead of maintaining and writing our own custom code. 

Note that this implementation is backward compatible for 3.4.x series as the existing keepAlive logic should still continue to work for Channelizers other than `WebSocketChannelizer` (notice the if statement in the code that checks for channelizer class before scheduling keepAlive). For `WebSocketChannelizer`, the number of Ping frames sent to the server is still the same except that it is now handled by Netty.

### Testing
1. Added a new unit test
2. gremlin-server `mvn clean install -DskipIntegrationTests=false -DincludeNeo4j`
3. gremlin-driver `mvn clean install -DskipIntegrationTests=false -DincludeNeo4j`

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -43,9 +43,11 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Release server threads waiting on connection if the connection is dead.
 * Fixed bug where server closes HTTP connection on request error even if keep alive is set to true.
 * Java driver now delegates handling of WebSocket handshake to Netty instead of custom code.
+* Java driver now delegates sending keepAlive messages to Netty for `WebSocketChannelizer`.
+* Java driver now uses WebSocket compression extension ([RFC7692](https://tools.ietf.org/html/rfc7692)) for `WebSocketChannelizer`.
+* Server now supports WebSocket compression extension ([RFC7692](https://tools.ietf.org/html/rfc7692)).
 * Fixed bug with Bytecode serialization when `Bytecode.toString()` is used in Javascript.
 * Fixed "toString" for P and TextP to produce valid script representation from bytecode glv steps containing a string predicate in Javascript.
-* Add support for WebSocket compression extension.
 
 [[release-3-4-8]]
 === TinkerPop 3.4.8 (Release Date: August 3, 2020)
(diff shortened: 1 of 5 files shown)


spmallette:
Good change - Glad the `shouldKeepAliveForWebSockets()` test is back. I suppose there is nothing now testing `scheduleKeepAlive()` logic - which remains fairly complex. I haven't thought about it, but is there any way to continue to cover that logic with tests (until we, i assume, one day remove the custom keep-alive all together)?

divijvaidya:
> Good change - Glad the `shouldKeepAliveForWebSockets()` test is back. I suppose there is nothing now testing `scheduleKeepAlive()` logic - which remains fairly complex. I haven't thought about it, but is there any way to continue to cover that logic with tests (until we, i assume, one day remove the custom keep-alive all together)?

Let's keep this out of the scope of this review? Asking because we don't even have a channelizer in the code base which would use the old code path, hence, testing becomes difficult.

spmallette:
hmm - true. perhaps some aspect of keep-alive was wrong from the beginning. The `Connection` probably shouldn't have needed to know anything about keep-alive infrastructure. If you agree, I think I'd propose that we deprecate that functionality and remove it in 3.5.0 (all of this can be in a different JIRA and we can move forward with merging this as it is). Does that make sense?

Either way, VOTE +1

divijvaidya:
> hmm - true. perhaps some aspect of keep-alive was wrong from the beginning. The `Connection` probably shouldn't have needed to know anything about keep-alive infrastructure. If you agree, I think I'd propose that we deprecate that functionality and remove it in 3.5.0 (all of this can be in a different JIRA and we can move forward with merging this as it is). Does that make sense?
> 
> Either way, VOTE +1

Yes I agree that this should be deprecated in 3.5 I will remove it from master as part of the merge and do a CTR update to documentation.

spmallette:
I made some minor tweaks to deprecation messaging: 35871ebdbbfd9799f230d37013da511a9487672d and created TINKERPOP-2455 for removal of the functionality in 3.5.0

PR #1345, opened 2020-10-09
TINKERPOP-2442 Moved Translators to gremlin-core.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2442

Better testing is coming on https://issues.apache.org/jira/browse/TINKERPOP-2443 but for now I think this is about as much as we had before in a sense. More reasoning for this change can be found in the JIRA itself. 

On merge to `master` I sense we will need some additional changes as we removed Jython completely. Some documentation updates may be necessary. I will look to review those sorts of things carefully.

VOTE +1


spmallette:
The merge to `master` on this is not being friendly. We refactored the `Translator` functionality for 3.5.0......forgot about that.

PR #1348, opened 2020-10-14
TINKERPOP-2446 - Add Looping Recipe for Optional Depth

bechbd:
Add another recipe to the looping section demonstrating how to accomplish a loop with optional depth


spmallette:
looks like you based your change on your local `master` branch but then pointed the PR at `3.4-dev`.  i think you just need to rebase.

PR #1355, opened 2020-11-08
TINKERPOP-2445 Remove unnecessary initialization of serializers

divijvaidya:

## Change
Remove deprecated methods.

## Motivation
Initialization of GraphsonV1.0 is a major contributor (~28%) to initialize a client with a single connection.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -63,6 +63,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Changed all configuration keys on `AbstractWarningVerificationStrategy` implementations to `public`.
 * Deprecated `BytecodeUtil` and merged its functionality to the existing `BytecodeHelper`.
 * Added configuring implementation in HasStep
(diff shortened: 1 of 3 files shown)


spmallette:
I think that the value of the speed improvement here is worth the removal and the possible breaking change. The scope of the "break" itself seems fairly narrow and isolated - this will not affect a lot of users (if any). If I want to take a breaking change like this within patch versions I usually just email the dev list with a DISCUSS thread and explain the reasoning/scope for the breaking change (I sometimes even email the user list if i feel it is of effect to a large number of users who might want to have an objection but i don't think that's necessary for this change). I then end the dev list post with a statement of a "72 hour lazy consensus in effect" and then link the DISCUSS thread in the JIRA.

I don't think this change needs Upgrade Documentation, but a CHANGELOG entry would be good.

Note that these initializations have already been removed on `master` iirc.

Assuming all that is in place, then VOTE +1

divijvaidya:
Dev mailing list discussion for this breaking change: https://lists.apache.org/thread.html/r9bc162d489d61fcf32184cf7e45f9e06f34daa0970ac20dc09304b95%40%3Cdev.tinkerpop.apache.org%3E

PR #1356, opened 2020-11-11
TINKERPOP-2001 Add lambda support in gremlin-javascript

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2001

Pretty straightforward - follows the same pattern as python basically. No need to support native javascript lambdas at this time (anymore than there is a future need to support native python lambdas at this time) as there is no legitimate execution engine for them (though that would be neat as you could write much more fluent Gremlin in javascript that way). 

This change enabled almost 30 tests that were just being ignored and paves the way for us to write more advanced GLV tests for other purposes. 

Builds with `mvn clean install`:

VOTE +1


jorgebay:
WOOT! nice!

I'll review it shortly.

PR #1358, opened 2020-11-13
HasStep implements configuring

rdtr:
As a graph provider, we need `HasStep` followed by `with()` options. Right now `HasStep` doesn't implement `configuring` so we encounter `ClassCastException`. 

This commit simply adds a configuring implementation same as other steps like `PropertiesStep`.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -57,6 +57,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Fixed "toString" for P and TextP to produce valid script representation from bytecode glv steps containing a string predicate in Javascript.
 * Fixed a bug which could cause Java driver to hang when using `ResultSet.statusAttributes()`
 * Deprecated `BytecodeUtil` and merged its functionality to the existing `BytecodeHelper`.
+* Added configuring implementation in HasStep
 
 [[release-3-4-8]]
 === TinkerPop 3.4.8 (Release Date: August 3, 2020)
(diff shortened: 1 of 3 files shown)


spmallette:
VOTE +1

spmallette:
Could you please re-target this pull request to `3.4-dev` branch? I should have noticed that on my previous review...

Starting to think about whether `Configuring` should be a feature of all steps and what refactoring that might mean. There is a JIRA issue somewhere that talks about the use of `Parameters` on many other steps than just the `Mutating` ones. I'll go find that and review a bit. That said, I still think it will be fine if this PR merged just as it is...just going to think on it a bit longer.

divijvaidya:
> Starting to think about whether Configuring should be a feature of all steps and what refactoring that might mean

+1 

As an evolution of the language itself, I am of the opinion that we should provide the ability to add "query hints" using the `with` step to every step in the language. This provides flexibility to the providers to extend the meaning of steps in their own fashion.

spmallette:
This was the issue I was thinking of:

https://issues.apache.org/jira/browse/TINKERPOP-1329

The comments are a bit tangential to the issue so you can largely stay focused on the description. It does indeed bring up the notion of using `Parameterizing` on filter steps as the next logical place where we would implement that functionality. This is relevant because `Configuring` extends `Parameterizing` and uses the same infrastructure to support it. 

Perhaps we consider a broader implementation for 3.5.0 and stick to this more localized changed here in 3.4.x. I feel like I'd need to do some exploratory implementation to see what impacts it would have - just poking around the code isn't enlightening me just now.

PR #1360, opened 2020-11-17
TINKERPOP-2445 Parallelize connection creation

divijvaidya:

## Changes
1. Create connections in a connection pool simultaneously instead of serially.
1. Create connection pools simultaneously for multiple host endpoints instead of serially. 
1. Add a new exception `NoHostAvailableException` for cases when no host with active connections is available to send the query. This replaces TimeoutConnection which was confusing earlier.
1. Do not try to reconnect to host if the connection pool initialization to that host has failed. Attempt to reconnect requires connection but we have a failure in the constructor of connection pool itself. Log the exception and don't add host in load balancer.
1. Added powermock dependency for mocking a new constructor inside a final class.
1. Changed some String.format to concatenation as the former is adding latency during client initialization.
1. Changed `ConnectionException` to extend `RuntimeException`. This simplifies exception handling. Prior to this we had to do multiple wrap and unwraps, hence, adding unnecessary values to the stack trace.

### DIFF
--- a/gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/ser/binary/types/sample/SamplePersonSerializerTest.java
+++ b/gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/ser/binary/types/sample/SamplePersonSerializerTest.java
@@ -29,6 +29,7 @@ import org.apache.tinkerpop.gremlin.structure.io.binary.GraphBinaryIo;
 import org.apache.tinkerpop.gremlin.structure.io.binary.GraphBinaryReader;
 import org.apache.tinkerpop.gremlin.structure.io.binary.GraphBinaryWriter;
 import org.apache.tinkerpop.gremlin.structure.io.binary.TypeSerializerRegistry;
+import org.junit.Assert;
 import org.junit.Test;
 import org.mockito.internal.matchers.apachecommons.ReflectionEquals;
 
@@ -43,7 +44,6 @@ import java.util.UUID;
 
 import static org.apache.tinkerpop.gremlin.driver.ser.AbstractMessageSerializer.TOKEN_IO_REGISTRIES;
 import static org.apache.tinkerpop.gremlin.driver.ser.GraphBinaryMessageSerializerV1.TOKEN_CUSTOM;
-import static org.junit.Assert.assertThat;
 
 public class SamplePersonSerializerTest {
 
@@ -93,7 +93,7 @@ public class SamplePersonSerializerTest {
             writer.writeValue(person, buffer, nullable);
             final SamplePerson actual = reader.readValue(buffer, SamplePerson.class, nullable);
 
-            assertThat(actual, new ReflectionEquals(person));
+            Assert.assertTrue(new ReflectionEquals(person).matches(actual));
             buffer.release();
         }
     }
@@ -108,7 +108,8 @@ public class SamplePersonSerializerTest {
         final ResponseMessage deserialized = serializer.deserializeResponse(serialized);
 
         final SamplePerson actual = (SamplePerson) deserialized.getResult().getData();
-        assertThat(actual, new ReflectionEquals(person));
+
+        Assert.assertTrue(new ReflectionEquals(person).matches(actual));
     }
 
     public static class CustomIoRegistry extends AbstractIoRegistry {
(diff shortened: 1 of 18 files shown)


pom.xml
```diff
@@ -721,6 +722,18 @@ limitations under the License.
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
```
spmallette, line 725: I'm not sure what the implications are but i just checked to see if you could mock `final` classes with just mockito and came across this:

https://www.baeldung.com/mockito-final

any chance that will suffice rather than adding another testing dependency. definitely don't want to make things harder but every new bit of testing syntax seems to make it harder to get to junit5 without tons of adapters and shims (even powermock holds a `powermock-module-junit4`. Anyway, if it can't be done without the additional dependency or you see a bigger plan for testing with powermock in the future, then I suppose it's fine to keep it, otherwise my preference would be to try to do it with the test modules we already have.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
```diff
@@ -861,12 +864,28 @@ public void shouldMarkHostDeadSinceServerIsDown() throws Exception {
...
+            // the host to load balancer but it will also not remove it if it already exists there. Leave that
+            // responsibility to a client that added it. In this case, let the second client perform it's own mechanism
+            // to mark host as unavailable. The first client will discover that the host has failed either with next
+            // keepAlive message or the next request, whichever is earlier. In this case, we will simulate the second
+            // scenario by sending a new request on first client. The request would fail (since server is down) and
+            // client should mark the host unavailable.
             cluster.connect().init();
+
+            try {
+                client1.submit("1+1").all().join();
```
spmallette, line 884: If we expect an exception here (i.e. i see it's meant to be ignored) should the test `fail()` if the request succeeds?

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/ser/binary/types/sample/SamplePersonSerializerTest.java
```diff
@@ -93,7 +93,7 @@ public void readValueAndWriteValueShouldBeSymmetric() throws IOException {
             writer.writeValue(person, buffer, nullable);
             final SamplePerson actual = reader.readValue(buffer, SamplePerson.class, nullable);
 
-            assertThat(actual, new ReflectionEquals(person));
+            Assert.assertTrue(new ReflectionEquals(person).matches(actual));
```
spmallette, line 96: nit: we tend to prefer to statically import the assertion method and to avoid `assertTrue()` which provides a terrible failure message, thus we would typically prefer: `assertThat(new ReflectionEquals(person).matches(actual), is(true))`

spmallette:
I like the new behavior you have here - do you have any sense of how much initialization speed you gained in making these changes? Also, i think this is a nice improvement to write a few words about in the Upgrade Documentation - always good to call attention to performance enhancements.

pom.xml
divijvaidya: We have two requirements here. Ability to mock a final class and use spy to mock a new initialization inside method which is being tested. Using mockito to mock final classes require using an [incubating](https://github.com/mockito/mockito/wiki/What%27s-new-in-Mockito-2#unmockable) feature in mockito. I tried using that feature but it doesn't work very nicely with spy right now. Hence, powermock is the next best option.

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/ser/binary/types/sample/SamplePersonSerializerTest.java
divijvaidya: Oh yes. This is a side effect of the change when I upgraded mockito to 2.x so that we could use it to mock final classes. It's not required in current review since we are using powermock. I will revert it back.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
divijvaidya: Good point. Added.

pom.xml
spmallette: ok - thanks for the explanation. let's go with what you've done then. thanks

divijvaidya:
> I like the new behavior you have here - do you have any sense of how much initialization speed you gained in making these changes?

Once we have merge this, the removal of serializer one and another PR about shutdown I am filing soon, I will perform a benchmark and add results to upgrade documentation.

spmallette:
VOTE +1

pom.xml
spmallette: build failures on `master` are due to powermock incompatibility with java 11 and our upgrade on the mockito version. not sure how you want to best resolve those - i tried a couple quick things but couldn't get it working immediately:

https://stackoverflow.com/questions/52966897/powermock-java-11
https://stackoverflow.com/questions/62305373/how-to-use-powermock-with-mockito3-x

I sense it is something that can be resolved though, but i'll leave it to you at this point.

spmallette: Figured out the fix i think: 8ea559e13e8e877a85c02d73178a675f31ac42b0 - builds for me locally now with `mvn clean install`.  will keep an eye on travis for the full build though.

divijvaidya: Thank you for helping fix this Stephen. Appreciate it 🥇

PR #1361, opened 2020-11-19
TINKERPOP-2479 Adds config/reset options to :bytecode command

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2479
https://issues.apache.org/jira/browse/TINKERPOP-2478

Also fixes TINKERPOP-2478 where whitespace was a problem with the from/translate options. See doc changes for more information on the changes along with examples.  Note that `Date` is fine now and works by default without use of the new `config` option:

```text
gremlin> :bc from g.V().property('d', new Date())
==>{"@type":"g:Bytecode","@value":{"step":[["V"],["property","d",{"@type":"g:Date","@value":1605795962056}]]}}
```

Tested manually - we don't have a good way to test these kinds of commands.

VOTE +1


krlawrence:
I pulled the branch and built. Confirm the problem I reported is fixed.

VOTE +1

PR #1362, opened 2020-11-19
Throw error from DriverRemoteConnection on connection error

iansrobinson:
Adds a listener that throws an error from a DriverRemoteConnection if an error occurs setting up a Websocket connection. Prior to this change, Gremlin requests submitted via a failed connection would succeed, but return a null result.


spmallette:
closing in favor of #1364

PR #1363, opened 2020-11-20
TINKERPOP-2445 Remove quiet time when shutting down Java client

divijvaidya:

**Problem**: Client init and shutdown takes ~2s with 100 connections even if we aren't sending any queries at all (against a locally running server w/o ssl).

**Reason**: 95% of the time is taken by shutdown. More specifically, [this line](https://github.com/apache/tinkerpop/blob/3.4-dev/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java) where we close Netty's event loop group. The default setting for closing the event loop group contains a "quiet time" of 2s, during which it will allow more requests to be handled by the executor. Thus, it will always wait at least 2s before shutting down. This 2s is configurable and we could change it instead of using the default.

**Change**: We are changing the quiet time to 0 because the event loop group is shutdown after executor (worker threads) are shutdown in a similar manner where we don't allow any more requests.


spmallette:
i think we generally take it as implied that you are in favor of your own work by just submitting the PR but don't forget to include your own "VOTE +1" on your own PRs - just a bit of extra formality that we gathered from Apache at some point. 

VOTE +1

PR #1347, opened 2020-10-12
Tinkerpop-2437 Do not leave statusAttributes future dangling

divijvaidya:
https://issues.apache.org/jira/browse/TINKERPOP-2437

##Motivation
We do not complete the statusAttribute future if the underlying `readCompleted` is completed exceptionally.

##Testing
grenlin-server `mvn clean install -DskipIntegrationTests=false -DincludeNeo4j`

PR #1365, opened 2020-11-20
TINKERPOP-2468 Fix stochastic nature of shouldProcessSessionRequestsInOrder

divijvaidya:
https://issues.apache.org/jira/browse/TINKERPOP-2468

The test failed intermittently because it expects the results to be returned in same order in which they were sent to the server. However, sessions guarantee that the requests would be executed "in-order" on the server but they don't guarantee that the results would be returned in order as well because the results are handled by different executor threads completing different futures.

I fixed the test to not rely on order of the result but still check for order on the server.

### DIFF
--- a/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
+++ b/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
@@ -612,40 +612,42 @@ public class GremlinDriverIntegrateTest extends AbstractGremlinServerIntegration
         }
     }
 
+    /**
+     * This test validates that the session requests are processed in-order on the server. The order of results
+     * returned to the client might be different though since each result is handled by a different executor thread.
+     */
     @Test
     public void shouldProcessSessionRequestsInOrder() throws Exception {
         final Cluster cluster = TestClientFactory.open();
         try {
             final Client client = cluster.connect(name.getMethodName());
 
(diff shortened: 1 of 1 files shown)


divijvaidya:
VOTE +1

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
```diff
@@ -612,40 +612,42 @@ public void shouldProcessRequestsOutOfOrder() throws Exception {
         }
     }
 
+    /**
+     * This test validates that the session requests are processed in-order on the server. The order of results
+     * returned to the client might be different though since each result is handled by a different executor thread.
+     */
```
spmallette, line 618: just to clarify , by "different executor thread" are you referring to threads on the client being responsible for the un-ordering?

divijvaidya: Yes. Client side threads. More specifically, in [this](https://github.com/apache/tinkerpop/pull/1347/files#diff-1a4243209a3c0c168ac7e9817b3bca1a2791db382c294130ea85c6f0bcec120cR109) PR we made a change to remove the possibility of blocking the executor loop on client by the application code by passing completion of readComplete future to client's executor threads. For two different requests they might get assigned to different threads in executor pool here and return results at different time.

spmallette:
VOTE +1

PR #1366, opened 2020-11-23
Raise error from Connection on Websocket connection error.

iansrobinson:
Raise error from promise when opening a connection that fails due to an underlying Websocket connection failure.


spmallette:
I've pulled over these commits to my branch on #1364. that new test passes nicely for me. if all that looks good, then i think we are set to merge from that one. thanks!

spmallette:
Merged on #1364

PR #1369, opened 2020-12-04
TINKERPOP-2472 Decouple the driver from the IO format

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2472

This prepares Gremlin.Net for GraphBinary by decoupling the serialization format from the driver. Similar to the Java driver, it is now possible to provide the driver with a `IMessageSerializer` which it will use to serialize request messages sent to the server and response messages received from the server. Included in this PR are implementations of that new interface for GraphSON 2 and 3.
We can then add another implementation for GraphBinary in a follow up PR.

Before this PR, serialization with GraphSON / JSON was tightly integrated in the driver which made it hard to support a different IO format.

NOTE: The driver now always deserializes a result completely, e.g.,skipping serialization like it was implemented in TINKERPOP-2067 is no longer supported. However, users who need this functionality can simply implement their own `IMessageSerializer` with this functionality which should be trivial to do. This should mainly affect the [_ExRam.Gremlinq_ library](https://github.com/ExRam/ExRam.Gremlinq).
@danielcweber: Would be great if you could take a short look at these changes to see whether they produce bigger problems for Exram.Gremlinq then I anticipated.

VOTE +1


gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/GraphSONMessageSerializer.cs
```diff
@@ -0,0 +1,100 @@
...
+            return Encoding.UTF8.GetBytes(MessageWithHeader(graphSONMessage));
+        }
+
+        private string MessageWithHeader(string messageContent)
+        {
+            return $"{(char) _mimeType.Length}{_mimeType}{messageContent}";
+        }
+
+        /// <inheritdoc />
+        public virtual ResponseMessage<List<object>> DeserializeMessage(byte[] message)
```
FlorianHockmann, line 71: This is the method users need to override if they want to skip deserialization with our GraphSON reader to get the functionality from TINKERPOP-2067 back.

danielcweber:
Great abstraction and highly appreciated. From how I understand this change, and assuming this PR builds on the recent move towards System.Text.Json, this change would allow me to easily stick to Newtonsoft.Json for the foreseeable future, wouldn't it?

FlorianHockmann:
> From how I understand this change, and assuming this PR builds on the recent move towards System.Text.Json, this change would allow me to easily stick to Newtonsoft.Json for the foreseeable future, wouldn't it?

Yes, this targets `master` which is already on `System.Text.Json` which will be released with 3.5.0. And right, you can stick to `Newtonsoft.Json` with this change if you implement your own `IMessageSerializer`. You just have to implement especially the deserialization completely on your own as [the `GraphSONReader` still uses a type](https://github.com/apache/tinkerpop/blob/master/gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/GraphSONReader.cs#L103) specific to our JSON library (`JsonElement`). We have this issue for quite some time ([TINKERPOP-1696](https://issues.apache.org/jira/browse/TINKERPOP-1696)) and I haven't found a way yet to get rid of this.
But if you want to do the deserialization on your own, then this doesn't affect you.

danielcweber:
>you can stick to Newtonsoft.Json with this change if you implement your own IMessageSerializer

Very nice. 

>We have this issue for quite some time (TINKERPOP-1696) and I haven't found a way yet to get rid of this.

One possible solution I could think of is to generically parametrize these interface on the native serialization type (in this case, JsonElement), making the json-library pluggable. I'm not sure where it'll all come together though., and I'm also not sure whether this scenario is already tackled by your recent change.

spmallette:
the change you describe here makes sense to me. that abstraction seems like an important change.

@FlorianHockmann are there any documentation changes necessary for this particular change or is it largely transparent to the user when they upgrade? or perhaps the documentation really isn't necessary until we have GraphBinary working and there are multiple options for serialization?

FlorianHockmann:
> are there any documentation changes necessary for this particular change or is it largely transparent to the user when they upgrade?

It is transparent to users unless they want to use GraphSON 2 or if they for some reason want to skip deserialization completely, but apart from the ExRam.Gremlinq library I don't know why people would do that and the feature to skip deserialization was also never communicated openly.
Configuration for GraphSON 2 changed from this:

```c#
var client = new GremlinClient(new GremlinServer("localhost", 8182), new GraphSON2Reader(),
    new GraphSON2Writer(), GremlinClient.GraphSON2MimeType);
```

to this:

```c#
var client = new GremlinClient(new GremlinServer("localhost", 8182), new GraphSON2MessageSerializer());
```

which can be seen in the `GremlinVariantsTests` that are automatically included in the reference docs. I think this is also a nice little improvement for the API as users now only need to change a single parameter to change the GraphSON version instead of three.

I didn't put that into the upgrade docs as the old `GremlinClient` constructor is still there. I just deprecated it and point users to the new constructor.
Should I still put this also in the upgrade docs? I'm always unsure what kind of change is significant enough to warrant an entry there as I also don't want to clutter it with information that is irrelevant for probably >99% of our users.

Adding GraphBinary will of course make it more important for users how they can change the serialization format, but I would then just add another part to the reference docs where we already explain how to configure the client for GraphSON 2. That would of course then also deserve a section in the upgrade docs, just like [when we added it for Python](https://tinkerpop.apache.org/docs/3.4.9-SNAPSHOT/upgrade/#_python_graphbinary).

spmallette:
It sounds as though there is no real impact to users and that the real impact would be GraphBinary. I think that you can save upgrade documentation for then. 

VOTE +1

danielcweber:
@FlorianHockmann I had a look into [TINKERPOP-1696](https://issues.apache.org/jira/browse/TINKERPOP-1696) and I don't think it's an issue per se that GraphSONReader exposes the native serialization type (JsonElement), but rather that it does so from the main library. Now that you decoupled IMessageSerializer, it should be straightforward to put all things System.Text.Json into its own library (e.g. Gremlin.Net.GraphSon). Not sure how it would work otherwise since JsonElement is also over all the IGraphSonDeserializer interface. Is having a separate library a viable option? Did I miss something? I could tackle the change if needed.

FlorianHockmann:
I think @jorgebay's intention with TINKERPOP-1696 was to avoid any external types in a publicly facing API as it would make it much harder for users to update to a newer version of Gremlin.Net when we switch the JSON library if they have used these interfaces, e.g., because they have implemented their own serializers.

That problem is of course still there. But I also don't see a good solution for this. I at least couldn't get the approach to work that @jorgebay suggested in the ticket to use `dynamic`. I also wouldn't want to put the serialization logic in a separate package for this.

But overall this problem doesn't have a high priority in my opinion and it's also completely independent of this PR basically. If we want to further discuss how to tackle it, then I'd suggest that we move the discussion into the comment section of that ticket.

jorgebay (approved):
Looks great, I like the new abstractions.

VOTE +1

PR #1370, opened 2020-12-11
TINKERPOP-2490 RangeGlobalStep check high limit before retrieve next traverser

junshiguo:
https://issues.apache.org/jira/browse/TINKERPOP-2490

In FilterStep, the processNextStart() method will first retrieve next traverser and then apply filtering logic. But for RangleGlobalStep, if high limit is already hit, there will be no need to get next traverser.

e.g. g.V().limit(1). This query will touch 2 vertices although only 1 vertex will be returned.

This PR added high limit check before retrieving next traverser for filtering. Functionality is not affected, but we can expect better performance if getting next traverser is heavy.


spmallette:
Thank you for this contribution. VOTE +1

spmallette:
Unfortunately, I've had to revert this change and re-open the issue in JIRA: 46fc7f627c75e016688727551f211c432af5a0d3 

I've commented in JIRA on the reasoning for why. I sense this is not a big deal to fix, but didn't want to risk including the change given our short window for release.

PR #1371, opened 2020-12-11
TINKERPOP-2438 Introduce GremlinScriptChecker

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2438

Even though TINKERPOP-2438 is already done this is a bit of a follow on to that change. The GremlinScriptChecker performs the same exact task as the GremlinAstChecker but it does so with regex parsing which should be faster than processing a full AST. I've left the GremlinASTChecker where it is because I wonder if there isn't yet some usage for more rigorous sort of checks we might come up with in the future or perhaps have checks that only work well with AST processing. It can be removed in the future if we like.

VOTE +1

### DIFF
--- a/gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GremlinScriptChecker.java
+++ b/gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GremlinScriptChecker.java
@@ -1,0 +1,203 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.tinkerpop.gremlin.groovy.jsr223;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Processes Gremlin strings using regex so as to try to detect certain properties from the script without actual
+ * having to execute a {@code eval()} on it.
+ */
+public class GremlinScriptChecker {
+
+    public static final Result EMPTY_RESULT = new Result(0);
+    private static final List<String> tokens = Arrays.asList("evaluationTimeout", "scriptEvaluationTimeout",
+                                                             "ARGS_EVAL_TIMEOUT", "ARGS_SCRIPT_EVAL_TIMEOUT");
+
+    /**
+     * Matches single line comments, multi-line comments and space characters.
+     * <pre>
+     * 	OR: match either of the followings
+     * Sequence: match all of the followings in order
+     * / /
+     * Repeat
+     * AnyCharacterExcept\n
+     * zero or more times
+     * EndOfLine
+     * Sequence: match all of the followings in order
+     * /
+     * *
+     * Repeat
+     * CapturingGroup
+     * GroupNumber:1
+     * OR: match either of the followings
+     * AnyCharacterExcept\n
+     * AnyCharIn[ CarriageReturn NewLine]
+     * zero or more times (ungreedy)
+     * *
+     * /
+     * WhiteSpaceCharacter
+     * </pre>
+     */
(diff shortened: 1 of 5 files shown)


gremlin-groovy/src/test/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GremlinScriptCheckerTest.java
```diff
@@ -0,0 +1,300 @@
...
+    @Test
+    public void shouldIdentifyMultipleTimeouts() {
+        assertEquals(6000, GremlinScriptChecker.parse("g.with('evaluationTimeout', 1000L).with(true).V().out('knows');" +
+                "g.with('evaluationTimeout', 1000L).with(true).V().out('knows')\n" +
+                "                                                   //g.with('evaluationTimeout', 1000L).with(true).V().out('knows')\n" +
+                "                                                   /* g.with('evaluationTimeout', 1000L).with(true).V().out('knows')*/\n" +
+                "                                                   /* \n" +
+                "g.with('evaluationTimeout', 1000L).with(true).V().out('knows') \n" +
+                "*/ \n" +
+                "                                                   g.with('evaluationTimeout', 1000L).with(true).V().out('knows')\n" +
```
divijvaidya, line 89: This should probably have a `;` at the end?

divijvaidya, line 89: This should probably have a` ;` at the end?

gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GremlinScriptChecker.java
```diff
@@ -0,0 +1,81 @@
...
+ * Processes Gremlin strings using regex so as to try to detect certain properties from the script without actual
+ * having to execute a {@code eval()} on it.
+ */
+public class GremlinScriptChecker {
+
+    public static final Result EMPTY_RESULT = new Result(0);
+    private static final List<String> tokens = Arrays.asList("evaluationTimeout", "scriptEvaluationTimeout",
+                                                             "ARGS_EVAL_TIMEOUT", "ARGS_SCRIPT_EVAL_TIMEOUT");
+
+    private static final Pattern patternClean = Pattern.compile("//.*$|/\\*(.|[\\r\\n])*?\\*/|\\s", Pattern.MULTILINE);
```
divijvaidya, line 37: Documentation on this regEx would be nice. Something like:


//.*$ --> match strings starting with // followed by zero or more chars till the end of line
| --> OR
\\*(. --> match string ...
...

You can choose to copy these comments from an online regex tester. They usually print out a detailed description of the regex.

gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GremlinScriptChecker.java
```diff
@@ -0,0 +1,81 @@
...
+    private static final String timeoutTokens = "[\"']evaluationTimeout[\"']|[\"']scriptEvaluationTimeout[\"']|(?:Tokens\\.)?ARGS_EVAL_TIMEOUT|(?:Tokens\\.)?ARGS_SCRIPT_EVAL_TIMEOUT";
+    private static final Pattern patternTimeout = Pattern.compile("\\.with\\((?:" + timeoutTokens + "),(\\d*)(:?L|l)?\\)");
```
divijvaidya, line 39: Please add documentation about this regex so that reader understands what it is doing.

gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GremlinScriptChecker.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+package org.apache.tinkerpop.gremlin.groovy.jsr223;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
```
divijvaidya, line 27: Please document the behaviour for multiple queries (that we add their timeouts). This should probably go in reference docs as well.

gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GremlinScriptChecker.java
spmallette: Interesting - the tool I've always used didn't produce that nice explain stuff. I didn't realize that such a thing existed. 👍

gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GremlinScriptChecker.java
spmallette: It is documented here: https://github.com/apache/tinkerpop/pull/1371/files#diff-0c5246d79c924132b94fc6af68af66b1f81faebac2410ca9e2af663388c31bb2R73-R79

Will add something to the reference documentation.

divijvaidya:
LGTM

VOTE +1

PR #1373, opened 2020-12-18
TINKERPOP-2451 Added JavascriptTranslator for Java

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2451

Re-factored cucumber tests to match the Python pattern, which uses the Translator to generate the code that is used by the GLV to execute the cucumber tests. In this way, we test the Translator and the GLV natively.

VOTE +1


gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/translator/JavascriptTranslatorTest.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    @Test
+    public void shouldTranslateDirection() {
+        assertTranslation("Direction.BOTH", Direction.BOTH);
+    }
+
+    @Test
+    public void shouldTranslateOrder() {
+        assertTranslation("Order.desc", Order.desc);
```
jorgebay, line 99: Nice coverage!

jorgebay (approved):
Nice!
VOTE +1

PR #1376, opened 2020-12-31
TINKERPOP-2473 Prevented assignment of more than one strategy of the same type.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2473

Don't think that this is a major change in the sense that it's unlikely that anyone was relying on the old behavior. The code now just better enforces what was expected to happen.

VOTE +1


vtslab:
Nice to have this clarified. Only had time for a quick read. I believe you chose "new instance replaces old one" (from the changed code comments). This choice could be added to the changelog for clarity. The traversal ref docs have a section on Traversal Strategies that still describe the old behaviour.

spmallette:
Thanks for reminding me about the documentation. I've pushed some changes to address your points.

vtslab:
VOTE +1 Interesting to read that it was actually a bug and that replacement always was the intended behaviour.

PR #1379, opened 2021-01-12
Fix construction field assignment

mallamanis:


spmallette:
Thanks for noticing this - merged as CTR on: f8b2449fc2fb6d65a19ebc1c0af36fa9eef424c3

PR #1385, opened 2021-01-28
TINKERPOP-2512 Avoid duplicate jars in classpath

cdegroc:
https://issues.apache.org/jira/browse/TINKERPOP-2512

$CLASSPATH already contains $CP so we shouldn't include both when calling `java -cp`


spmallette:
Thank you merged this under: 63bb830d9e0271418040fafe596b3a337e60ee18

PR #1388, opened 2021-02-05
fix toString() for P class when value is array

sergsalo:
example is `P.within(["a", "b"])` predicate where ["a", "b"] is an array. In such case toString() must return `within("a","b")` and not within(a,b)


spmallette:
Thanks for your contribution. I added some tests in addition to your work:

https://github.com/apache/tinkerpop/commit/05c323af664795db024de53a0fbdebc4540fb1db

PR #1389, opened 2021-02-08
3.4 dev, handle the translation of multiple embedded traversals 

sergsalo:
handle the translation of multiple embedded traversals


sergsalo:
I need to fix couple things before this PR

PR #1392, opened 2021-02-18
add a public helper method to retrieve select key from WherePredicate…

rdtr:
Just extracted the logic to get a select key out to a public method so that we can reuse it from outside of this Step.


spmallette:
Looks like Travis is getting some build failures with this change.

spmallette:
I think that the issue is that the new method is referring the member variable of `parameters` rather than the one local to the calling method: 

https://github.com/apache/tinkerpop/pull/1392/files#diff-0e30cd3ebe2d8f6d9ac7ce6c9e91762ed7a98ff2f7682b4041faf8dbd8bb9898R68

PR #1394, opened 2021-02-20
Fixed NPE during the ResponseMessageDeserializer#createObject

egetman:
Hi guys!
It's a small change that fixes annoying NPE's on ResponseMessageDeserializer#createObject.

Background:
I use JanusGraph & Gremlin, and when some kind of runtime exceptions occurred in the server, the server responds with a null message (I assume it just gets the exception.getMessage() as-is, and it can be null). 
So the typical response at that point looks like:
```
{requestId=c7465b5e-2c12-4f66-a293-3cdf2c276e17, status={message=null, code=500, attributes={stackTrace=java.lang.ClassCastException
, exceptions=[java.lang.ClassCastException]}}, result={data=null, meta={}}}
```
and the NPE's stacktrace begins with:
```
AbstractGraphSONMessageSerializerV2d0 - Response [PooledUnsafeDirectByteBuf(ridx: 311, widx: 311, cap: 311)] could not be deserialized by org.apache.tinkerpop.gremlin.driver.ser.AbstractGraphSONMessageSerializerV2d0.
java.lang.NullPointerException: null
	at org.apache.tinkerpop.gremlin.driver.ser.AbstractGraphSONMessageSerializerV2d0$ResponseMessageDeserializer.createObject(AbstractGraphSONMessageSerializerV2d0.java:298)
```
When I was fixing that NPE, I found the duplicating code of ResponseMessage creation in:
```
AbstractGraphSONMessageSerializerV1d0
GraphSONMessageSerializerV1d0
``` 
but I didn't touch it cause I'm not sure if it accidentally looks the same or not.
If it's the same logic, maybe it's time to extract that code into some small utility component that will handle the ResponseMessage creation from the raw map?

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ser/AbstractGraphSONMessageSerializerV2d0.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ser/AbstractGraphSONMessageSerializerV2d0.java
@@ -294,7 +294,7 @@ public abstract class AbstractGraphSONMessageSerializerV2d0 extends AbstractMess
             final Map<String, Object> result = (Map<String, Object>) data.get(SerTokens.TOKEN_RESULT);
             return ResponseMessage.build(UUID.fromString(data.get(SerTokens.TOKEN_REQUEST).toString()))
                     .code(ResponseStatusCode.getFromValue((Integer) status.get(SerTokens.TOKEN_CODE)))
-                    .statusMessage(status.get(SerTokens.TOKEN_MESSAGE).toString())
+                    .statusMessage(String.valueOf(status.get(SerTokens.TOKEN_MESSAGE)))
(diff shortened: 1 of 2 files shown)


spmallette:
Thank you for offering this change. I'd not realized that the `statusMessage` might be `null`. I'd wonder if whatever `ResponseMessage` being created there is not being constructed properly, but I'm fine with your fix. If you could address a few items though, that would be great:

1. GraphSON 1 is mostly deprecated as is GraphSON 2. I'd not enter into any significant refactoring at this point around them to extract common logic. If it really bugs you I suppose you could but I'd say just fix the problem in both places (especially since it's a minor change)
2. Please add a CHANGELOG entry to say something like "Fixed `NullPointerException` in `ResponseMessage` serialization for GraphSON.
3. Please take a moment to look and see if the other serializers like Gryo and GraphBinary have this same issue. I think Gryo does but GraphBinary does not from a quick look. If there are more fixes then please adjust the CHANGELOG message as necessary.
4. Please target the `3.4-dev` branch rather than `master` for this change.

PR #1399, opened 2021-02-24
TINKERPOP-2527 Python->Groovy Translator

krlawrence:
https://issues.apache.org/jira/browse/TINKERPOP-2527

This code adds a Translator class to the Python client that will convert traversals back to Gremlin Groovy. I also included 100 or so unit tests. I have tried to cover bot step instructions and source instructions so things like strategies as well as basic Gremlin steps should get converted. This has been a common ask from Python users for a few use cases including 
1.  Writing to logs
2. Dropping out of bytecode to use sessions
3. Debugging machine generated traversals

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -58,6 +58,7 @@ This release also includes changes from <<release-3-4-3, 3.4.3>>.
 * Configured GraphSON 3.0 as the default setting for the `GraphSONMapper`.
 * Added `JavascriptTranslator` for Java.
 * Added `DotNetTranslator` for Java.
+* Added Groovy `Translator` for Python.
 * Fixed bug in `PythonTranslator` for processing `TraversalStrategy` instances in GraphBinary.
 * Fixed bug in bytecode `Bindings` where calling `of()` prior to calling a child traversal in the same parent would cause the initial binding to be lost.
 * Bumped to tornado 6.1 for gremlinpython.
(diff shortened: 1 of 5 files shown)


spmallette:
I just pushed a commit on this branch to do two basic things: 

1. Minor code formatting to PEP2 standards
2. Refactored tests to run with the standard build

Note that in doing so I noticed there was a failure on this test which made me change the assertion:

https://github.com/apache/tinkerpop/pull/1399/files#diff-f08dfe7ef61a419bb3393595484bacb700f5264213588c860625fb43e1099374R345-R346

I'm not sure exactly what happened there. Perhaps you could have a look to see if you're ok with that?

Lastly, I don't think we need to do it for this PR but I hope we can find a way to leverage our gherkin tests somehow to test translators here just as we do on the the JVM. This does look well tested however, so I'm good with how this is shaping up. We can save more advanced testing for later.

krlawrence:
Thanks for the review and the additional work. All looks great.

spmallette:
@krlawrence now that documentation is generating on `master` again I realized some sample code for Python->Groovy is needed on this PR: 

https://github.com/apache/tinkerpop/pull/1399/files#diff-aa69bbbf8224e408ea19bf3df1466e16e6309bcfb241f68b578032d9fbbbd48dR4301-R4335 

Do you have a minute to add that? If you do, be sure to pull in my changes from this branch before you add your example.

PR #1406, opened 2021-03-12
TINKERPOP-2532 Add public getter for maxBarrierSize

rngcntr:
https://issues.apache.org/jira/browse/TINKERPOP-2532

This change allows vendors to use the configured barrier sizes to enable and improve batch processing (e.g. [Janusgraph-2514](https://github.com/JanusGraph/janusgraph/issues/2514)).
I also added a test class because no class existed for `NoOpBarrierStep`. I hope my implementation of `getTraversals()` does what it is expected to do. If anyone has more experience with tests for `Step` classes and would like to issue changes, please let me know.


rngcntr:
I don't see why Travis tries to build this PR with jdk11. Can anybody help me with this?

spmallette:
it is building with jdk11 because you started with a base branch of `master` which is configured to build that way. i guess when you changed to 3.4-dev branch as a target it didn't pick that up. I restarted the build for you and that seemed to fix it. 

I will have a look at the contents of your pull requests soon enough - thanks for your contribution.

spmallette:
This was a small change and I'm a bit surprised we'd not exposed this property before so I merged it as CTR. Thank you.

PR #1411, opened 2021-04-01
add a public helper method to retrieve select key from WherePredicateStep

rdtr:
[The previous PR](https://github.com/apache/tinkerpop/pull/1392) was closed as I deleted the remote branch once and pushed again with the fix.


spmallette:
I had to tweak your change slightly - bd3df5771b27eea6982abff3bc152e95e7d71bef but this is merged now.

PR #1413, opened 2021-04-02
TINKERPOP-2457 Added max_content_length as Python driver config

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2457

I'm not sure how this will merge to `master` at the moment, or whether that merge is relevant there given the work to move from tornado to aiohttp. I sense this work will merge before that, so we'll see how it becomes relevant or not.

VOTE +1

### DIFF
--- a/gremlin-python/src/main/jython/gremlin_python/driver/tornado/transport.py
+++ b/gremlin-python/src/main/jython/gremlin_python/driver/tornado/transport.py
@@ -26,21 +26,25 @@ __author__ = 'David M. Brown (davebshow@gmail.com)'
 
 class TornadoTransport(AbstractBaseTransport):
 
+    _default_max_content_length = 10 * 1024 * 1024
+
     def __init__(self, read_timeout=None, write_timeout=None,
                  compression_options={'compression_level': 5, 'mem_level': 5},
-                 ssl_options=None):
+                 ssl_options=None, max_content_length=_default_max_content_length):
         self._loop = ioloop.IOLoop(make_current=False)
         self._ws = None
         self._read_timeout = read_timeout
         self._write_timeout = write_timeout
         self._compression_options = compression_options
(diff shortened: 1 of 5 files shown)


krlawrence:
This is a commonly asked for feature that I have had to use private workarounds for in the past. Great to see this going into the mainline code.

VOTE +1

gremlin-python/src/main/jython/gremlin_python/driver/tornado/transport.py
```diff
@@ -26,21 +26,25 @@
 
 class TornadoTransport(AbstractBaseTransport):
 
+    _default_max_content_length = 10 * 1024 * 1024
```
divijvaidya, line 29: This is probably a breaking change and should go either in 350 or in 3411 with same default as current version. In absence of explicitly setting this value, the default must be coming from the websocket framework. 

This is breaking because let's say that the current default is 20MB and a user of the client is relying on that but when we introduce a new default value of 10MB, after upgrade, their queries will suddenly start failing.

spmallette: it was set to precisely the same default as existed before. isn't that correct @krlawrence ?

krlawrence: The Tornado default is 10MB The Tornado code includes that exact calculation (10 * 1024 * 1024) in fact. This should not be a breaking change to anyone. I did a lot of testing of this before I opened the original Jira to prove that that is the current limit.

divijvaidya: Thank you for clarifying. If the default is same as before then this is definitely not breaking. That answers my question.

divijvaidya:
VOTE +1

PR #1418, opened 2021-04-20
TINKERPOP-2550 Fixed bug in java Client initialization

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2550

This change fixes a potential deadlock issue on `Client` initialization and may have some impact on https://issues.apache.org/jira/browse/TINKERPOP-2504 as well. I won't re-document the points of this change here as I went to considerable lengths to explain it in code comments. Please see the rationale for the changes there.

VOTE +1

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
@@ -18,6 +18,8 @@
  */
 package org.apache.tinkerpop.gremlin.driver;
 
+import org.apache.commons.lang.exception.ExceptionUtils;
+import org.apache.commons.lang3.concurrent.BasicThreadFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.tinkerpop.gremlin.driver.exception.ConnectionException;
@@ -41,6 +43,8 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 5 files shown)


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
```diff
@@ -511,19 +515,24 @@ protected Connection chooseConnection(final RequestMessage msg) throws TimeoutEx
...
         protected void initializeImplementation() {
+            // use a special executor here to initialize the Host instances as the worker thread pool may be
+            // insufficiently sized for this task and the parallel initialization of the ConnectionPool. if too small
+            // tasks may be schedule in such a way as to produce a deadlock: TINKERPOP-2550
+            //
+            // the cost of this single threaded executor here should be fairly small because it is only used once at
+            // initialization and shutdown. since users will typically construct a Client once for the life of their
+            // application there shouldn't be tons of thread pools being created and destroyed.
+            final BasicThreadFactory threadFactory = new BasicThreadFactory.Builder().namingPattern("gremlin-driver-init-%d").build();
+            final ExecutorService hostExecutor = Executors.newSingleThreadExecutor(threadFactory);
```
divijvaidya, line 526: starting and shutting down executor is an expensive operation which will impact the client start up/shut down time which will impact use cases when customers have short running clients such as in AWS Lambda. Is there an alternative approach that can avoid this?

spmallette: can we not take a few milliseconds to fix this in a simple way? in multiple tests of 3.4.10 vs this change, the longest time difference i've seen is like 3.5 milliseconds extra time per open/close and many times the gap is even less. too much still?

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
```diff
@@ -511,19 +515,24 @@ protected Connection chooseConnection(final RequestMessage msg) throws TimeoutEx
...
                         .join();
             } catch (CompletionException ex) {
-                Throwable cause = null;
-                Throwable result = ex;
-                if (null != (cause = ex.getCause())) {
-                    result = cause;
-                }
-
-                logger.error("", result);
+                logger.error("", ExceptionUtils.getRootCause(ex));
```
divijvaidya, line 533: ExceptionUtils.getRootCause returns null if there is no caused by in the stack trace. That is the reason, we had the prior custom written code logic.

spmallette: really?! ok - that's not what i thought happened. i will switch back.

spmallette: man...we use `getRootCause()` everywhere....................kinda want to fix that everywhere now. how many places is this a problem hiding errors?!

divijvaidya:
Vote +1

PR #1419, opened 2021-04-20
TINKERPOP-2547 Add a way to supply a callback on handshake requests with java driver

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2547

Added a `HandshakeInterceptor` interface that can be supplied to the `Cluster.Builder` which will act as a callback for the initial HTTP request that sets up the websocket through the Java driver.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -37,6 +37,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Ensured that `barrier()` additions by strategies were controlled solely by `LazyBarrierStrategy`.
 * Fixed `NullPointerException` in `ResponseMessage` deserialization for GraphSON.
 * Enabled the Gremlin.Net driver to repair its connection pool after the server was temporarily unavailable.
+* Added the ability to supply a `HandshakeInterceptor` to a `Cluster` which will provide access to the initial HTTP request that establishes the websocket.
 
 [[release-3-4-10]]
 === TinkerPop 3.4.10 (Release Date: January 18, 2021)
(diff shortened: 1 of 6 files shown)


FlorianHockmann (approved):
VOTE +1

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/RequestInterceptor.java
```diff
@@ -0,0 +1,37 @@
...
+package org.apache.tinkerpop.gremlin.driver;
+
+import org.apache.tinkerpop.gremlin.driver.handler.RequestInterceptorHandler;
+
+import java.util.function.UnaryOperator;
+
+/**
+ * This function is called when the {@link RequestInterceptorHandler} encounters an object passing through it.
+ * Implementers will need to detect the type of object to determine if they will interact with it or not. Typically,
+ * objects will be a {@code FullHttpRequest} in the case of a websocket handshake or some form of
```
divijvaidya, line 28: To begin with, shall we only restrict this to exposing the HTTP Requests Object and only at the stage of first handshake initialization with the server (that would mean that instead of being a handler this callback will be called from the place where we construct the first HTTP request)? Suggesting because invoking a function for every frame when the actual interceptor is required only for modifying the initial set of headers will have unnecessary performance penalty.

spmallette: I was trying to generalize this functionality rather than pin it particularly to `FullHttpRequest` and the initial handshake. We could say that's not useful and that only the first HTTP request matters I guess. This feature was largely your idea so I'll modify it to work the way you'd envisioned.

divijvaidya:
Thank you for picking this up. 

Vote +1 (assuming travis checks pass)

PR #1420, opened 2021-04-30
TINKERPOP-2559 Gremlin.NET: Stop sending session close message

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2559

I'm creating the PR already as it was such a trivial change but given that we're in a code freeze, this has to wait for the 3.5.0 release.

VOTE +1


spmallette:
i think you will need to re-target this to the `3.5-dev` branch. a CHANGELOG entry would be nice for 3.5.1. 

VOTE +1

FlorianHockmann:
> i think you will need to re-target this to the 3.5-dev branch. a CHANGELOG entry would be nice for 3.5.1.

Done

PR #1422, opened 2021-05-19
TINKERPOP-2569 Reconnect if java driver fails to initialize

spmallette:

Does result in a slight change in behavior as client initialization is now aware of dead hosts and will throw a `NoHostAvailableException` earlier (used to wait for a request to be submitted). This is probably better behavior. It seems unlikely that anyone would rely on this behavior for normal usage so pushing it to 3.4.x seems safe.

VOTE +1

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
@@ -111,8 +111,11 @@ final class ConnectionPool {
             logger.warn("Initialization of connections cancelled for {}", getPoolInfo(), ce);
             throw ce;
         } catch (CompletionException ce) {
-            // Some connections might have been initialized. Close the connection pool gracefully to close them.
-            this.closeAsync();
+            // since there was some error here, the host is likely unavailable at this point. if we dont mark it as
+            // such the client won't try to reconnect to the dead host. need to initialize this.open because if we
+            // don't then reconnects will fail when trying to create a new one with addConnectionIfUnderMaximum()
+            this.open = new AtomicInteger(0);
+            considerHostUnavailable();
 
             final String errMsg = "Could not initialize " + minPoolSize + " (minPoolSize) connections in pool." +
                     " Successful connections=" + this.connections.size() +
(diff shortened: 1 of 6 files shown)


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
```diff
@@ -111,8 +111,11 @@ public ConnectionPool(final Host host, final Client client, final Optional<Integ
             logger.warn("Initialization of connections cancelled for {}", getPoolInfo(), ce);
             throw ce;
         } catch (CompletionException ce) {
-            // Some connections might have been initialized. Close the connection pool gracefully to close them.
-            this.closeAsync();
+            // since there was some error here, the host is likely unavailable at this point. if we dont mark it as
+            // such the client won't try to reconnect to the dead host. need to initialize this.open because if we
+            // don't then reconnects will fail when trying to create a new one with addConnectionIfUnderMaximum()
+            this.open = new AtomicInteger(0);
+            considerHostUnavailable();
```
divijvaidya, line 118: This was intentionally removed in https://github.com/apache/tinkerpop/pull/1360 

Here's what happens if we don't remove this:

The connection has failed here. It could be due to misconfiguration such as wrong ssl credentials. `considerHostUnavailable()` will create a background task which will try to send reconnect message to the host. To send that message, it will try to borrowConnection from connection pool. But connection pool had a problem during initialization in the first place. It will try to initialise connections again. It will fail again due to ssl misconfiguration. On failure, it will call `considerHostUnavailable()` again. After some wait, borrowConnection will try to create connections again which will again fail and eventually after a timeout everything would give up and throw timeout back to application. On application side, they will observe TimedOut exception and have no clue why the connection failed because the actual reason of failure is hidden in logs from 15 min ago.

Another example of complication, on `considerHostUnavailable()` we create a new connection for reconnect message asynchronously but right after we also destroy all existing connections. In some race condition, we could end up in situation where we are adding connection for reconnect and then destroying them due to earlier logic.

IMO, to keep things simple, if a client is not able to connect to the server on initialisation, it should fail fast with an exception and let application layer handle the scenario. The application could choose to recreate the client at that stage or fail their application startup and check the logs for configuration errors.

spmallette: i'd forgotten we removed that and when i was looking at the user concern i'd remembered only how it used to work. maybe i can make some aspect of this "better" in a different fashion that doesn't bring all those problems you mentioned.

PR #1427, opened 2021-06-04
TINKERPOP-2576 Enable GitHub's Dependabot for .NET

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2576

Dependabot is now part of GitHub which comes with a new config file. This commit configures Dependabot only for Gremlin.NET so we can try it out there first.

I already enabled this change on my fork. You can see a PR created by Dependabot with this config here:
https://github.com/FlorianHockmann/tinkerpop/pull/130

We don't need the old Dependabot config file any more that was for the pre-GitHub version as soon as we enable Dependabot via GitHub by merging this PR.

VOTE +1


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "nuget"
+    directory: "/gremlin-dotnet"
+    schedule:
+      interval: "daily"
+    target-branch: "3.5-dev"
+    labels:
+      - ".NET"
```
FlorianHockmann, line 9: I think that using these two labels can make it easier to filter the PRs, e.g,, by GLV. We'll need to create this label though before merging the PR.

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "nuget"
+    directory: "/gremlin-dotnet"
+    schedule:
+      interval: "daily"
+    target-branch: "3.5-dev"
```
FlorianHockmann, line 7: I use `3.5-dev` here as the target branch instead of `3.4-dev` as `3.4-dev` still contains the source generation from templates for the GLVs which won't work with Dependabot. But I also think that it's OK to skip updates for `3.4-dev`, given that most users only use the newest version of Gremlin.NET anyway (per nuget.org download statistics), and because we haven't done much updates until now so it's not like this would make anything worse.

spmallette:
I'm going to miss my awesome rebel commit message of "Kill dependabot with deadly config!!!" at the homepage of the github repo 😈 but I'm fine with trying this. Perhaps we can enable it for more parts of TinkerPop as we go:

VOTE +1

jorgebay (approved):
VOTE +1

PR #1452, opened 2021-07-26
TINKERPOP-2522 Fix nested indentations for profiler

li-boxuan:


spmallette:
Thanks for this fix. It has been merged. Note that I added an extra annotation to your test to just make sure that case was covered: https://github.com/apache/tinkerpop/commit/fea4765ca466089703f8d0e00132a03c897aa25f

PR #1453, opened 2021-07-27
Document lessons learned releasing 3.4.12/3.5.1

FlorianHockmann:
This adds my lessons learned from the release process. I also noticed that we tag the docker images on `3.4-dev` with `latest` which we shouldn't do anymore now that we have 3.5.z releases as those should always be the latest release. So this removes the `latest` tag from the `3.4-branch` and I'll open another branch to target `3.5-dev` which keeps the `latest` tag there.

The deployment of Docker images failed for me through Maven so I had to manually posh the images. This seems to be a known issue for our Maven plugin: spotify/dockerfile-maven#344.
If I'm the only one affected by this, then we don't need to do anything from my side, but if others also run into this problem, then we should probably check whether we can replace that Maven plugin as it's also not actively maintained any more.

VOTE +1

### DIFF
--- a/docs/src/dev/developer/development-environment.asciidoc
+++ b/docs/src/dev/developer/development-environment.asciidoc
@@ -253,7 +253,6 @@ username = <username>
 password =
 
 [pypi]
-repository = https://pypi.python.org/pypi
 username = <username>
 password =
 ----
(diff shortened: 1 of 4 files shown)


spmallette:
eek - i dont like that the docker plugin is not maintained anymore. i guess that in the worst case we could go to our own hand rolled scripting approach but that's unfortunate. i didn't have problems with it for the last release so, not sure what's wrong there. let's let it sit until next release and we'll see what happens.

VOTE +1 (i think you  are free to merge this without any additional votes)

PR #1454, opened 2021-07-27
Document lessons learned releasing 3.4.12/3.5.1 [3.5-dev]

FlorianHockmann:
This contains the same commits as the other PR for 3.4-dev, just without removing the `latest` tag for the Docker images and it adds gremlint in a few more places.

VOTE +1


spmallette:
VOTE +1 (i think you are free to merge this without any additional votes)

PR #1455, opened 2021-07-29
TINKERPOP-2582 Construct traversals from gremlin-language

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2582

Produced a basic parser to use `gremlin-language` to generate a `Traversal` object. I don't think we will promote this as part of the 3.5.x line but I figured it wouldn't hurt to push it here for those who wish to experiment. 

VOTE +1


joshsh:
+1 to the concept. It's a lot of code... may I suggest a Zoom session to walk us through it?

spmallette:
i don't mind doing a zoom sesion if folks are interested. it's not terribly interesting code though. it just sorta does what it does, but there are some spots that might be worth calling attention to for future thought/work.

PR #1457, opened 2021-07-31
TINKERPOP-2591: Administrative adjustments to gremlint site

OyvindSabo:
https://issues.apache.org/jira/browse/TINKERPOP-2591

**Install gremlint from npm rather than from OyvindSabo/gremlint**
Previously the gremlint website installed the gremlint library from the old gremlint repo (OyvindSabo/gremlint). Now that it's available on npm it can be installed from there.

**Add footer displaying current gremlint version and copyright notice**
The footer is always positioned below the view content of the website, but it sticks to the bottom of the screen if there is not enough content to push it out of the screen. I avoided using any of the approaches which depend on a fixed footer height, as that could cause weird behavior for users with narrow screens where the footer text wraps.
image
image

The gremlint version is pulled from the `package-lock.json` file. Displaying the version from `package-lock.json` instead of `package.json`, for instance, enables us to display the version as `5.3.1` while still specifying it as `^5.3.1` in `package.json`. In apps created using Create React App (as the gremlint website is) only files inside the `src` folder are considered when running or building the app. To be able to access the gremlint version from `package-lock.json` I used `customize-cra` (a tool used to override and extend the default Create React App scripts) to include `package-lock.json` in the files considered to be part of the module, hence the new `config-overrides.js` file.

**Testing:**
Tested that both `npm start` and `npm run build` still works in `docs/gremlint`.
Tested that the whole TinkerPop website can still be built with `bin/generate-home.sh` and run by `npx serve target/site/home`.

VOTE +1

### DIFF
--- a/docs/gremlint/package.json
+++ b/docs/gremlint/package.json
@@ -11,8 +11,10 @@
     "@types/node": "^12.0.0",
     "@types/react": "^16.9.53",
     "@types/react-dom": "^16.9.8",
-    "gremlint": "github:OyvindSabo/gremlint#master",
+    "customize-cra": "^1.0.0",
+    "gremlint": "^3.5.1",
     "react": "^17.0.1",
+    "react-app-rewired": "^2.1.8",
     "react-dom": "^17.0.1",
     "react-scripts": "^4.0.3",
     "sharp-router": "^4.1.5",
@@ -23,8 +25,8 @@
   "scripts": {
     "predeploy": "npm run build",
     "deploy": "gh-pages -d build",
-    "start": "react-scripts start",
-    "build": "react-scripts build",
+    "start": "react-app-rewired start",
+    "build": "react-app-rewired build",
     "test": "react-scripts test",
     "eject": "react-scripts eject"
   },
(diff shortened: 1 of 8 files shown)


docs/gremlint/package.json
```diff
@@ -11,8 +11,10 @@
     "@types/node": "^12.0.0",
     "@types/react": "^16.9.53",
     "@types/react-dom": "^16.9.8",
-    "gremlint": "github:OyvindSabo/gremlint#master",
+    "customize-cra": "^1.0.0",
+    "gremlint": "^3.5.1",
```
OyvindSabo, line 15: I set the gremlint version to `^3.5.1` for now, as opposed to `^3.5.1-alpha1` or `^3.5.1-rc1` as I wasn't sure which convention we were going with and also not sure if we would actually want the website to use the alpha versions.:thinking:

spmallette: the only part in my mind where it gets tricky is with the release process, but i think you have it right. when we release 3.5.2 we dont' publish docs until after we'd have deployed that artifact to npm at which point we'd get the newer version to be live on the website at gremlint.com.

spmallette:
VOTE +1

PR #1465, opened 2021-09-01
TINKERPOP-2486 Fix load balancing across connections

simonz-bq:
https://issues.apache.org/jira/browse/TINKERPOP-2486

The method to select the least used connection is now synchronized and now includes the increment on its usage count so that it will not be fetched by multiple threads if it is not the least used connection. It now instead will decrement the usage if we deem that it will not be used. 

Also, simplified and cleaned-up some logic currently using unnecessary `while (true)` loops.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -37,7 +37,8 @@ limitations under the License.
 * Bumped to Apache Hadoop 3.3.1.
 * Bumped to Apache Spark 3.1.2.
 * Changed `NumberHelper` to properly cast to `byte` and `short` rather than default coercing to `Integer`.
-* Modified some driver defaults (maximum content length, pool size, maximum in process) to be more consistent with one another. 
+* Modified some driver defaults (maximum content length, pool size, maximum in process) to be more consistent with one another.
+* Fixed a potential connection load balancing issue due to a race condition not updating the usage count.
 
 == TinkerPop 3.5.0 (The Sleeping Gremlin: No. 18 Entr'acte Symphonique)
 
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
@@ -142,8 +142,6 @@ final class ConnectionPool {
 
         if (isClosed()) throw new ConnectionException(host.getHostUri(), host.getAddress(), "Pool is shutdown");
 
-        final Connection leastUsedConn = selectLeastUsed();
-
         if (connections.isEmpty()) {
             logger.debug("Tried to borrow connection but the pool was empty for {} - scheduling pool creation and waiting for connection", host);
             for (int i = 0; i < minPoolSize; i++) {
@@ -158,6 +156,9 @@ final class ConnectionPool {
             return waitForConnection(timeout, unit);
         }
 
+        // Get the least used valid connection
+        final Connection leastUsedConn = getLeastUsedValidConnection();
+
         if (null == leastUsedConn) {
             if (isClosed())
                 throw new ConnectionException(host.getHostUri(), host.getAddress(), "Pool is shutdown");
@@ -165,32 +166,9 @@ final class ConnectionPool {
             return waitForConnection(timeout, unit);
         }
 
-        // if the number borrowed on the least used connection exceeds the max allowed and the pool size is
-        // not at maximum then consider opening a connection
-        final int currentPoolSize = connections.size();
-        if (leastUsedConn.borrowed.get() >= maxSimultaneousUsagePerConnection && currentPoolSize < maxPoolSize) {
-            if (logger.isDebugEnabled())
-                logger.debug("Least used {} on {} exceeds maxSimultaneousUsagePerConnection but pool size {} < maxPoolSize - consider new connection",
-                        leastUsedConn.getConnectionInfo(), host, currentPoolSize);
-            considerNewConnection();
-        }
-
-        while (true) {
-            final int borrowed = leastUsedConn.borrowed.get();
-            final int availableInProcess = leastUsedConn.availableInProcess();
-
-            if (borrowed >= maxSimultaneousUsagePerConnection && leastUsedConn.availableInProcess() == 0) {
-                logger.debug("Least used connection selected from pool for {} but borrowed [{}] >= availableInProcess [{}] - wait",
-                        host, borrowed, availableInProcess);
-                return waitForConnection(timeout, unit);
-            }
-
-            if (leastUsedConn.borrowed.compareAndSet(borrowed, borrowed + 1)) {
-                if (logger.isDebugEnabled())
-                    logger.debug("Return least used {} on {}", leastUsedConn.getConnectionInfo(), host);
-                return leastUsedConn;
-            }
-        }
+        if (logger.isDebugEnabled())
+            logger.debug("Return least used {} on {}", leastUsedConn.getConnectionInfo(), host);
+        return leastUsedConn;
     }
 
     public void returnConnection(final Connection connection) throws ConnectionException {
@@ -342,7 +320,7 @@ final class ConnectionPool {
             return false;
         }
 
-        announceAvailableConnection();
+        announceAllAvailableConnection();
         return true;
     }
 
@@ -393,23 +371,12 @@ final class ConnectionPool {
             if (isClosed())
                 throw new ConnectionException(host.getHostUri(), host.getAddress(), "Pool is shutdown");
 
-            final Connection leastUsed = selectLeastUsed();
-            if (leastUsed != null) {
-                while (true) {
-                    final int inFlight = leastUsed.borrowed.get();
-                    final int availableInProcess = leastUsed.availableInProcess();
-                    if (inFlight >= availableInProcess) {
-                        logger.debug("Least used {} on {} has requests borrowed [{}] >= availableInProcess [{}] - may timeout waiting for connection",
-                                leastUsed, host, inFlight, availableInProcess);
-                        break;
-                    }
+            final Connection leastUsed = getLeastUsedValidConnection();
 
-                    if (leastUsed.borrowed.compareAndSet(inFlight, inFlight + 1)) {
-                        if (logger.isDebugEnabled())
-                            logger.debug("Return least used {} on {} after waiting", leastUsed.getConnectionInfo(), host);
-                        return leastUsed;
-                    }
-                }
+            if (leastUsed != null) {
+                if (logger.isDebugEnabled())
+                    logger.debug("Return least used {} on {} after waiting", leastUsed.getConnectionInfo(), host);
+                return leastUsed;
             }
 
             remaining = to - TimeUtil.timeSince(start, unit);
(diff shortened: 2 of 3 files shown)


divijvaidya (changes requested):

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
```diff
@@ -475,7 +474,7 @@ private void announceAvailableConnection() {
         }
     }
 
-    private Connection selectLeastUsed() {
+    private synchronized Connection selectLeastUsed() {
```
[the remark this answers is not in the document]
simonz-bq: Renamed, but this is outdated now.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
```diff
@@ -158,38 +156,40 @@ public Connection borrowConnection(final long timeout, final TimeUnit unit) thro
             return waitForConnection(timeout, unit);
         }
 
+        final Connection leastUsedConn = selectLeastUsed();
```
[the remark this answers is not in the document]
simonz-bq: I changed the way it's currently working. I'll add the top level comment in this PR but it's no longer complicated enough to warrant a specific comment aside from the docstring on the method itself.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
```diff
@@ -158,38 +156,40 @@ public Connection borrowConnection(final long timeout, final TimeUnit unit) thro
...
         }
 
+        // Currently borrowed and used connections is 1 less than borrowed which is incremented by selectLeastUsed
+        final int borrowedInUse = leastUsedConn.borrowed.get() - 1;
+
         // if the number borrowed on the least used connection exceeds the max allowed and the pool size is
         // not at maximum then consider opening a connection
         final int currentPoolSize = connections.size();
-        if (leastUsedConn.borrowed.get() >= maxSimultaneousUsagePerConnection && currentPoolSize < maxPoolSize) {
+        if (borrowedInUse >= maxSimultaneousUsagePerConnection && currentPoolSize < maxPoolSize) {
```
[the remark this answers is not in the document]
simonz-bq: I added a test case. See top level PR comment for new logic.

simonz-bq:
I updated the connection pool logic a bit to be simpler and to solve the load balancing issues due to race conditions.

In my opinion the change seems reasonable but there may be a side effect that I haven't considered or figured out.

Aside from the load balancing issues, the old and new logic work as follows for getting a connection:

1. Grab the least used connection
2. If no connection could be returned, wait for one to be available
3. If the least used connection's borrow count is higher than `maxSimultaneousUsagePerConnection`, consider making a new connection if the pool is not full. **Even if `maxSimultaneousUsagePerConnection`, this least used connection is still considered a valid connection to return**
4. If the least used connection's borrow count is higher than `maxSimultaneousUsagePerConnection` AND the least used connection has reached it's maximum processes, wait for one to be available
5. Else, return the least used connection

The old logic allowed a connection to be grabbed even if `maxSimultaneousUsagePerConnection` was exceeded, and only deemed it invalid if also the maximum processes was reached.

New:

1. Grab the least used connection that is below the `maxSimultaneousUsagePerConnection` count
2. Increment the borrow count, and if it now reaches `maxSimultaneousUsagePerConnection`, consider creation of a new connection
3. If a least used connection could not be grabbed, then wait for one to be available

spmallette:
Travis is showing two job failures - I restarted one of them as i sense it was unrelated, but the other specifically fails one of the new tests added here:

```text
[ERROR] Tests run: 3, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 68.217 s <<< FAILURE! - in org.apache.tinkerpop.gremlin.driver.ClientConnectionIntegrateTest
[ERROR] overLimitOperationsShouldDelegateToSingleNewConnection(org.apache.tinkerpop.gremlin.driver.ClientConnectionIntegrateTest)  Time elapsed: 31.629 s  <<< ERROR!
java.lang.NullPointerException
	at org.apache.tinkerpop.gremlin.driver.ClientConnectionIntegrateTest.lambda$overLimitOperationsShouldDelegateToSingleNewConnection$6(ClientConnectionIntegrateTest.java:201)
	at org.apache.tinkerpop.gremlin.driver.ClientConnectionIntegrateTest.overLimitOperationsShouldDelegateToSingleNewConnection(ClientConnectionIntegrateTest.java:199)
```

divijvaidya (commented):
1. In this change you have removed the concept of maxInProcess setting which is nice but that makes this change backward incompatible. Consider a user who has already set a value of maxInProcess to 10 and maxSimultaneous to 100. Due to the bug we weren't respecting the limit on maxSimultaneous but maxInProcess was ensuring that only 10 active requests are sent in a single connection (at `if (inFlight >= availableInProcess) {`) but now, their single connection would start having 100 requests (maxSimultaneous).

2. We might want to completely get rid of inProcess since we are not using it for making the connection selection decision, e.g. it's still used as a decision point while trying to return connection to the pool. It could be marked as deprecated in the settings as well.

Due to the above reasons, this change could be a target for 3.6.x (or perhaps 3.5.x with ample documentation) but to keep it compatible for 3.4.x, could we ensure that there is no adverse effect for users with existing configuration?

I will also think more about it and get back with some potential solutions next week.

simonz-bq:
@divijvaidya Makes sense. Let me know what the decision is so that I can change the base and perhaps mark something as deprecated.

In the meanwhile, I looked into why the build was failing due to a timeout. It turns out that a new Connection being created was not notifying all awaiters of a Connection (since they can be given to more than 1 thread) but rather only notifying one. Combined with the new logic for obtaining a Connection this should now work properly.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/ClientConnectionIntegrateTest.java
```diff
@@ -107,4 +116,102 @@ public void shouldCloseConnectionDeadDueToUnRecoverableError() throws Exception
...
+                } catch (InterruptedException | ExecutionException e) {
+                    e.printStackTrace();
+                }
+                return null;
+            }).forEach(conn -> {
+                String id = conn.getChannelId();
+                channelsSize.put(id, channelsSize.getOrDefault(id, 0) + 1);
+            });
+
+            channelsSize.entrySet().forEach(entry -> {
```
spmallette, line 155: it might be an odd chance but perhaps it would be smart to have a defensive assert prior to this line that the `channelsSize` actually has contents. If it somehow stays empty this test would pass without doing an assertions.

simonz-bq: Done

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/ClientConnectionIntegrateTest.java
```diff
@@ -107,4 +116,102 @@ public void shouldCloseConnectionDeadDueToUnRecoverableError() throws Exception
...
+            }
+
+            HashMap<String, Integer> channelsSize = new HashMap<>();
+
+            final List<Future<Connection>> executorSubmitFutures = executorServiceForTesting.invokeAll(listOfTasks);
+            executorSubmitFutures.parallelStream().map(fut -> {
+                try {
+                    return fut.get();
+                } catch (InterruptedException | ExecutionException e) {
+                    e.printStackTrace();
```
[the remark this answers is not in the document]
simonz-bq: I didnt add a log.debug, but I changed this according to something similar.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/ClientConnectionIntegrateTest.java
```diff
@@ -107,4 +115,103 @@ public void shouldCloseConnectionDeadDueToUnRecoverableError() throws Exception
...
+
+        } finally {
+            executorServiceForTesting.shutdown();
+            client.close();
+            cluster.close();
+        }
+    }
+
+    @Test
+    public void overLimitOperationsShouldDelegateToSingleNewConnection() throws InterruptedException {
```
divijvaidya, line 167: There is a lot of duplicate code between these tests that could be refactored into re-usable functions.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
```diff
@@ -342,7 +320,7 @@ private boolean addConnectionIfUnderMaximum() {
             return false;
         }
 
-        announceAvailableConnection();
```
divijvaidya, line 345: When we return a connection to the pool in `ConnectionPool.returnConnection`, we use `announceAvailableConnection` and do not announce to all waiting threads whereas here on creation of a new connection we notify all waiting threads. Semantically whether a connection is being returned to the pool or a new connection is being created are same from the perspective of a waiting thread, hence, we should not have difference between notification behaviour.

divijvaidya:
From a backward compatibility angle, I reviewed this CR for the following cases:

1. User has a value of maxInProcess > maxSimulatenousUsage

This case should not be impacted because connection would (should!) not have been borrowed and thus in process requests would never be greater than borrowed requests. 

2. User has a value of maxInProcess < maxSimulatenousUsage 

Prior to the change in this PR, at line `if (inFlight >= availableInProcess) {` requests will wait to be sent to the server in case maxSimulatenous value has been reached but now, all borrowed requests will be sent to the server,  i.e. after this PR, users with the above configuration might observe more simultaneous requests per connection than earlier. This makes this change risky for users who have adjusted the configuration after deliberate load testing on their servers.

I would **vote - 1** for this change to be released in 3.4.x and 3.5.x because of the potential of unanticipated impact of removing inProcess configuration for existing users of the client.

I would **vote +1** for this change to be released in 3.6.x / master **after the comment associated with announceAvailableConnection has been addressed**.

Separately, some additional work also needs to be taken up associated with this change (out of the scope of this PR):
1. Perf testing the impact of synchronised method introduced in this change for high throughput workload.
2. Deprecate max/minInprocess configuration

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/ClientConnectionIntegrateTest.java
simonz-bq: I think, for now anyways, the two tests are rather self contained and already complex. Keeping them as is makes them easier to understand, and I don't think there will be a reduction in complexity or even lines of code to refactor it right now.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
simonz-bq: When a new Connection is created, `inFlight` count will be at 0, thus all waiting threads to see if they can pick up this new Connection.

When a Connection is returned to the pool, `inFlight` count is reduced by 1 from its current value, which if there are any threads waiting the Connection is currently at `maxSimulatenousUsage` because otherwise there wouldn't be threads waiting for it. That is why there is only need to notify a single thread here.

simonz-bq:
The change has been rebased to target master instead of 3.4 according to @divijvaidya's request.

divijvaidya (approved):
Looks good. After fixing the change log please feel free to merge into master. We can look at merging into 3.5.x and 3.4.x as a separate PR.

CHANGELOG.asciidoc
```diff
@@ -430,6 +430,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 [[release-3-4-13]]
 === TinkerPop 3.4.13 (Release Date: NOT OFFICIALLY RELEASED YET)
 
+* Fixed a potential connection load balancing issue due to a race condition not updating the usage count.
```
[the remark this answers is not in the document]
simonz-bq: Done.

divijvaidya:
Looks good. Vote +1 for merging into master.

PR #1469, opened 2021-09-03
Added getter method for bypassTraversal in AbstractLamdaTraversal

rdtr:
Graph provider should be able to access bypassTraversal when analyzing traversal for query execution.


spmallette:
Merged - thanks.

PR #1472, opened 2021-09-10
TINKERPOP-2379 Make defaults consistent with Java

simonz-bq:
Changed the defaults where it appeared to have a corresponding setting to one in Java with a different value.

Was mentioned that the specific default changes should be mentioned in the changelog using 3.5.1 as a template in this file:  https://github.com/apache/tinkerpop/blob/3.5-dev/docs/src/upgrade/release-3.5.x.asciidoc but there didn't seem to be an apparently appropriate place to put it under 3.5.2 looking at the 3.5.1 template.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -33,7 +33,7 @@ limitations under the License.
 * Prevented metrics computation unless the traversal is in a locked state.
 * Bumped to Apache Hadoop 3.3.1.
 * Bumped to Apache Spark 3.1.2.
-* Changed some default initialization settings in Gremlin-Python and Gremlin-Dotnet to match Java.
+* Modified some driver defaults (maximum content length, pool size, maximum in process) to be more consistent with one another. 
 
 == TinkerPop 3.5.0 (The Sleeping Gremlin: No. 18 Entr'acte Symphonique)
 
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -33,6 +33,7 @@ limitations under the License.
 * Prevented metrics computation unless the traversal is in a locked state.
 * Bumped to Apache Hadoop 3.3.1.
 * Bumped to Apache Spark 3.1.2.
+* Changed some default initialization settings in Gremlin-Python and Gremlin-Dotnet to match Java.
 
 == TinkerPop 3.5.0 (The Sleeping Gremlin: No. 18 Entr'acte Symphonique)
 
--- a/docker/gremlin-server/gremlin-server-integration-krb5.yaml
+++ b/docker/gremlin-server/gremlin-server-integration-krb5.yaml
@@ -55,7 +55,7 @@ keepAliveInterval: 0
 maxInitialLineLength: 4096
 maxHeaderSize: 8192
 maxChunkSize: 8192
-maxContentLength: 1000000
+maxContentLength: 10485760
 maxAccumulationBufferComponents: 1024
 resultIterationBatchSize: 64
 writeBufferLowWaterMark: 32768
--- a/docker/gremlin-server/gremlin-server-integration-secure.yaml
+++ b/docker/gremlin-server/gremlin-server-integration-secure.yaml
@@ -55,7 +55,7 @@ keepAliveInterval: 0
 maxInitialLineLength: 4096
 maxHeaderSize: 8192
 maxChunkSize: 8192
-maxContentLength: 1000000
+maxContentLength: 10485760
 maxAccumulationBufferComponents: 1024
 resultIterationBatchSize: 64
 writeBufferLowWaterMark: 32768
--- a/docker/gremlin-server/gremlin-server-integration.yaml
+++ b/docker/gremlin-server/gremlin-server-integration.yaml
@@ -56,7 +56,7 @@ keepAliveInterval: 0
 maxInitialLineLength: 4096
 maxHeaderSize: 8192
 maxChunkSize: 8192
-maxContentLength: 1000000
+maxContentLength: 10485760
 maxAccumulationBufferComponents: 1024
 resultIterationBatchSize: 64
 writeBufferLowWaterMark: 32768
--- a/docs/src/upgrade/release-3.5.x.asciidoc
+++ b/docs/src/upgrade/release-3.5.x.asciidoc
@@ -36,9 +36,9 @@ Some default value settings were changed to be consistent across different langu
 
 * The default for `PoolSize` is now `8` instead of `4` in .NET.
 * The default for `MaxInProcessPerConnection` is now `32` instead of `16` in .NET.
+* The default for `maxContentLength` is now 10 mb in the Java builder instead of 65536.
 * The default for `pool_size` is now `8` instead of `4` in Python.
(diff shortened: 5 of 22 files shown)


FlorianHockmann:
Thanks for the contribution, but I think we need some discussion about these changes first.

The connection pool in the Java driver works differently and it also has different settings:
* `maxInProcessPerConnection` defaults to 4
* `maxSimultaneousUsagePerConnection` defaults to 16
* `maxSize` defaults to 8.

but the .NET driver only has two settings:
* `PoolSize` defaults to 4
* `MaxInProcessPerConnection` defaults to 32

and the Python driver only has the `pool_size` setting which defaults to 4.

I guess we could ignore the Java setting `maxInProcessPerConnection` here and use `maxSimultaneousUsagePerConnection` as the equivalent for the `MaxInProcessPerConnection`  in .NET (see also [TINKERPOP-2619](https://issues.apache.org/jira/browse/TINKERPOP-2619) which aims to simplify the Java connection pool settings).

In general, changing these settings would be a breaking change as users would suddenly get exceptions because they are sending more requests in parallel than allowed with reduced limits. So, this probably needs to target `master` to be released as part of 3.6.0.

Also, if we want to align the settings across drivers, then I wonder whether the Java default values are really better than the .NET or Python defaults. Ideally, we would perform some benchmarks for different scenarios and with all three drivers to find some meaningful numbers and also to verify that the same values perform similarly for the different drivers. But given that we have mostly set the defaults by guessing meaningful values until now, we don't necessarily have to do this work as part of this PR from my side.

spmallette:
Considering the administrative points first for @simonz-bq :

1. Could you please make sure PRs have the JIRA number in the title and the link in the description (the former allows for some integration to happen and the latter is just convenient)?
2. You could have just added the section "Upgrading for Users" and a sub-heading for your change. The "Upgrading for Users" is there now after a merge I just did.
3. ultimately, please squash your commits for this PR

The only other administrative point relates to where this PR should land. I had thought that it `3.5-dev` was fine but I don't think I counted on the connection pool defaults shifting in the fashion @FlorianHockmann presented. 

> then I wonder whether the Java default values are really better than the .NET or Python defaults

I think this was why I didn't expect a major breaking change @FlorianHockmann and that we could explain away default adjustments in the upgrade documentation. I was thinking mostly of  `maxContentLength` getting bigger in Java (only someone relying on a smaller number, which is unlikely, would encounter a problem) and timeouts growing a bit longer. You're right that Java may not have been the ideal model. 

@krlawrence do you have recommendations for a `maxContentLength` default? or any other defaults for that matter?

For now, let's see how the changes would actually affect folks before we decide if the change will be too radical for 3.5.x.

@simonz-bq - this issue was as much about identifying the configuration commonalities and gaps as it was about changing the defaults themselves, so I think all of this feedback is expected. From the PR, it seems that the only items in common are:

* connection pools
* max content length

is that really all there is (i'd have guessed there were various timeouts in common in some way)? and does javascript not include any commonality at all? Note that settings may not be named in the same way or may be taking defaults from the underlying websocket library being used. Moreover, if you're only looking at the "Configuration" sections of the various language variants in the Reference Documentation, this may be incomplete. Perhaps another part of this effort should be to examine/update the documentation for completeness.

simonz-bq:
I updated the PR with the following changes:

- Use maxSimultaneousUsagePerConnection as the equivalent for the MaxInProcessPerConnection in .NET
- Added JIRA number in title
- Added description in upgrading for users section
- Squashed the commits (I'm used to just doing this when merging in the PR)

I think this is a reasonable state for the PR to be in until more discussion is done.

@spmallette The timeouts matched when there was an equivalent option. In JavaScript, I did not see any current configuration for the clients in a way that was changing similar fields as the other languages. Maybe it is worth investigating settings in the underlying clients but I believe that to be outside of the scope of this ticket. Doing something like that requires an agreement that we *do* want to change the default configuration of a different library, and even beyond that it also requires a discussion of whether or not it should be exposed as configuration for the user.

krlawrence:
VOTE  -1

I'm not in favor of some of these changes. Please do not downgrade the max content sizes to match Java that is a breaking and IMO bad change. The default for example for Python has always been 10mb as that was the Tornado default and we consciously carried that over into the HTTP ASYNC/IO version of the driver.

If anything the Java values should be raised. Time and again I have to tell users how to increase the Java settings. I rarely have to tell Python users how to. I do not support this breaking change for the Python client changing max_content_length to 65536.

We should not try to blindly make all of the settings the same as the Java ones. In some cases I will assert the Java values are wrong.

spmallette:
so, allow me to summarize to get more specific here:

1. the "max content length" should be 10MB across drivers so that we don't downgrade python.
2. it sounds like we are all content with the various "pool sizes" being 8
3. it looks like we're ok with using florian's suggestion of maxSimultaneousUsagePerConnection as the equivalent for the MaxInProcessPerConnection in .NET 
4. there's no timeouts in common, so nothing to change there
5. we'll leave lower level driver defaults within specific libraries alone for now.

one thing not noted so far is that if we increase the "max content length" for the driver we need to do the same for the server, so we would need to change the various YAML files out there:

https://github.com/apache/tinkerpop/blob/master/gremlin-server/conf/gremlin-server.yaml#L48

as well as:

https://github.com/apache/tinkerpop/blob/master/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java#L146

and potential adjust failing integration tests that relied on the former 64k max. 

i suppose my final question here is whether or not this stays on 3.5-dev as florian questioned that. If you now tried to connect a 3.5.2 driver to 3.5.0 server and expected a 10mb maxContentLength to work, it wouldn't since the server would still default at 64k. Do we instead target master?

FlorianHockmann:
> i suppose my final question here is whether or not this stays on 3.5-dev as florian questioned that. If you now tried to connect a 3.5.2 driver to 3.5.0 server and expected a 10mb maxContentLength to work, it wouldn't since the server would still default at 64k. Do we instead target master?

This sounds to me like this should target master. Is there any reason why this should be released in 3.5.2?

spmallette:
i guess i'd originally pushed for 3.5.x because i wasn't completely sure what the changes would end up being and how much trouble they would cause. I'm fine with this being moved to master for the safest possible upgrade.

simonz-bq:
Rebased and changed target to `master`.

Also changed the max content length to 10mb where applicable, instead of the other way around.

spmallette:
When I wrote "so we would need to change the various YAML files out there" i did mean that there were others than the example I provided. sorry if that wasn't clear. See the other yaml in that directory of the one you changed and also look at:

https://github.com/apache/tinkerpop/blob/master/gremlin-server/src/test/resources/org/apache/tinkerpop/gremlin/server/gremlin-server-integration.yaml

which is currently set to: 1000000. Changing this may break some integration tests which I alluded to in my last comment. Further note the comment in that file at the top and also adjust the YAML that publish with docker:

https://github.com/apache/tinkerpop/tree/master/docker/gremlin-server

Perhaps the CHANGELOG needs a minor adjustment now? We didn't exactly "match Java" given the final consensus - maybe just say: "Modified driver defaults to be more consistent with one another ("pool size", "max content length", and "max in process").?

simonz-bq:
Uploaded a set of changes to the defaults. Let me know if I missed any.

spmallette:
VOTE +1

PR #1473, opened 2021-09-14
TINKERPOP-2569: Reconnect if java driver fails to initialize

xiazcy:

Made minor changes to the initialization of the Java driver Cluster and Client, such that hosts are marked as available only after successfully initializing connection pools, to fix reconnection failures when client was initialized on a dead host that restarts.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,6 +25,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 * Prevented XML External Entity (XXE) style attacks via `GraphMLReader` by disabling DTD and external entities by default.
 * Fixed a `NullPointerException` that could occur during a failed `Connection` initialization due to uninstantiated `AtomicInteger`.
+* Minor changes to the initialization of Java driver `Cluster` and `Client` such that hosts are marked as available only after successfully initializing connection pools.
 
 [[release-3-4-12]]
 === TinkerPop 3.4.12 (Release Date: July 19, 2021)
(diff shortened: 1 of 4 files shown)


xiazcy:
I'm not sure why CI wasn't automatically triggered and I don't see a way to manually trigger it, but I've ran both `GremlinDriverIntegrateTest` and `GremlinServerIntegrateTest` locally and tests that passed before making the changes also passed after the proposed changes. 

There were 3 tests inside `GremlinDriverIntegrateTest` that failed persistently for me locally, I'm not sure if I should be concerned about them? They could be due to the fact that I'm running the build on a macOS. 
- shouldFailWithScriptExecutionException
```
java.lang.AssertionError: 
Expected: a string starting with "java.lang.ArithmeticException: Division by zero\n\tat java.math.BigDecimal.divide(BigDecimal.java"
     but: was "java.lang.ArithmeticException: Division by zero
```
- shouldWorkWithGraphSONExtendedV2Serialization
```
java.lang.AssertionError: 
Expected :2021-09-15T18:07:09.294325Z
Actual   :2021-09-15T18:07:09.294Z
```
- shouldWorkWithGraphSONExtendedV3Serialization
```
java.lang.AssertionError: 
Expected :2021-09-15T18:13:39.049275Z
Actual   :2021-09-15T18:13:39.049Z
```

spmallette:
I backported several changes from 3.5-dev related to these tests which I think will allow them to pass better for you. I'm not sure that it's macos specifically, but i think possibly related to various JVM types/versions...not wholly sure. In any case, please rebase your PR on on 3.4-dev and let's see if that fixes anything (and if it actually triggers travis).  Please let me know if that change helped?

spmallette:
perhaps you were already going to fix this, but please prefer to `git rebase` rather than `git merge` so that we keep the git history nice. you should only have your single commit in this PR with just your changes (currently you can see my changes in this diff with my commit as well).

xiazcy:
Yes, sorry, just fixed it! I will keep that in mind for the future.

spmallette:
> Travis is still not triggered though.

It's probably not necessary but I'm curious if re-issuing the PR would fix it. Please go ahead and close this one and re-open and let's see what happens.

PR #1474, opened 2021-09-15
TINKERPOP-2601 Unify testing behind Gherkin

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2601

JVM based providers can now easily testing with the Gherkin feature tests which were formerly just built for testing GLVs. The benefit here is that we no longer need to do duplicate test writing (i.e. once in the old Process Test Suite in Java and then again in Gherkin). For 3.6.0 I think the plan is to maintain both suites but recommend providers implement the Gherkin version along with the transitional new "limited" process suite. Ultimately, the idea would be simple delete all tests from Java that already exist in Gherkin and to migrate all provider oriented tests to Gherkin. The latter will likely require some additional changes to both the test definitions and the various runners in each programming language. I'd think that for 3.7.0 we could begin the process of deleting the Java tests and not relying on them anymore.

Going forward, we should look to only write tests in Gherkin unless there is some semantic that we currently can't express by way of the runner, though other exceptions may exist and we don't need a hard and fast rule for that.

VOTE +1


jorgebay (approved):
This is a huge undertaking, thanks for taking care of this. It's great to mostly rely on Gherkin features for everything.

Reviewing a changeset like this is very hard, I went through parts of the java/javascript/C# and lgtm! As long as CI is happy, I think it's safe to merge.

VOTE +1

PR #1475, opened 2021-09-16
TINKERPOP-2596 Added datetime() to gremlin-language

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2596

Added GroovyTranslator support by way of a new type serializer - in this way, standard Date/Timestamp translation of old can be preserved.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -28,6 +28,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Added the `ConnectedComponent` tokens required to properly process the `with()` of the `connectedComponent()` step.
 * Fixed `DotNetTranslator` bugs where translations produced Gremlin that failed due to ambiguous step calls to `has()`.
 * Fixed bug where `RepeatUnrollStrategy`, `InlineFilterStrategy` and `MessagePassingReductionStrategy` were all being applied more than necessary.
+* Modified grammar to accept the `datetime()` function so that Gremlin scripts have a way to natively construct a `Date`.
 * Ensured `PathRetractionStrategy` is applied after `InlineFilterStrategy` which prevents an error in traverser mapping in certain conditions.
 * Deprecated `JsonBuilder` serialization for GraphSON and Gryo.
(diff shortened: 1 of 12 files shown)


spmallette:
> if I create a date using datetime("12:00"), that means "12pm UTC on January 1, 1970"?

yes. i was copying jodatime's parsing implementation here.

joshsh:
> > if I create a date using datetime("12:00"), that means "12pm UTC on January 1, 1970"?
> 
> yes. i was copying jodatime's parsing implementation here.

OK. Just seems like this will never, ever be used -- or if it used, the user may use it incorrectly, thinking the time refers to the current day of the year.

bechbd:
Yeah I agree that `datetime("12:00")` will likely never be used.  TBH I have no idea what I would even expect this to represent if I was programming it.

krlawrence:
I see two possible ways we could interpret `datetime("12:00")`.

1. 12:00 on Jan 1, 1970
2. 12:00 UTC today . This would also align well with TZ offsets such as `-5:00` and + `6:00` (this has the interesting side effect that it sort of gives us a `now()` equivalent for a current TZ)

The W3C note from 1997 defines an interesting profile along these lines:

https://www.w3.org/TR/NOTE-datetime

spmallette:
pulling Josh's note from JIRA just to make sure no one missed it:

> Option 2 kinda makes sense, if "bare" timestamps are allowed at all. Even with full date-times, there is some context dependence if the time zone is omitted – it defaults to a "local" time zone, which has to be predefined. We could assume that a reference timestamp is also predefined, and bound to a specific instant near the beginning of query evaluation. You could have even weirder things than hours and minutes without a date, e.g. seconds and milliseconds without minutes. Is it worth trying to support these oddball cases? Idk, but you could do it...

to summarize, it sounds as though josh/dave are +0 on having the bare time at all and kelvin has a decent idea with his option 2 if we did support it - i sorta like that better than the jodatime approach of basing from 1970.

spmallette:
I've opted to remove the bare time component of this PR. Should we find need for it later we can add it easily enough.

PR #1476, opened 2021-09-16
 TINKERPOP-2569: Reconnect if java driver fails to initialize

xiazcy:

Made minor changes to the initialization of the Java driver Cluster and Client, such that hosts are marked as available only after successfully initializing connection pools, to fix reconnection failures when client was initialized on a dead host that restarts.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
@@ -19,6 +19,8 @@
 package org.apache.tinkerpop.gremlin.driver;
 
 import org.apache.commons.lang3.concurrent.BasicThreadFactory;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.tinkerpop.gremlin.driver.exception.ConnectionException;
@@ -31,6 +33,7 @@ import org.apache.tinkerpop.gremlin.process.traversal.Traverser;
 import org.apache.tinkerpop.gremlin.structure.Graph;
 import org.apache.tinkerpop.gremlin.util.iterator.IteratorUtils;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
@@ -524,22 +527,25 @@ public abstract class Client {
             final BasicThreadFactory threadFactory = new BasicThreadFactory.Builder().namingPattern("gremlin-driver-initializer").build();
             final ExecutorService hostExecutor = Executors.newSingleThreadExecutor(threadFactory);
 
-            try {
-                CompletableFuture.allOf(cluster.allHosts().stream()
-                        .map(host -> CompletableFuture.runAsync(() -> initializeConnectionSetupForHost.accept(host), hostExecutor))
-                        .toArray(CompletableFuture[]::new))
-                        .join();
-            } catch (CompletionException ex) {
-                Throwable cause;
-                Throwable result = ex;
-                if (null != (cause = ex.getCause())) {
-                    result = cause;
-                }
+            // hosts are not marked as available at Cluster initialization, so we are marking hosts as available here
+            // if connection pools can be initialized successfully.
+            final List<Pair<Host, CompletableFuture<Void>>> completableFutures = new ArrayList<>();
 
-                logger.error("", result);
-            } finally {
-                hostExecutor.shutdown();
+            for (Host host : cluster.allHosts()) {
+                completableFutures.add(new ImmutablePair<>(host,
+                        CompletableFuture.runAsync(() -> initializeConnectionSetupForHost.accept(host), hostExecutor)));
             }
+
+            for (Pair<Host, CompletableFuture<Void>> completableFuture : completableFutures) {
+                try {
+                    completableFuture.getRight().join();
+                    completableFuture.getLeft().makeAvailable();
+                } catch (CompletionException ex) {
+                    logger.error("", (ex.getCause() == null) ? ex : ex.getCause());
+                }
+            }
+
+            hostExecutor.shutdown();
         }
 
         /**
@@ -567,8 +573,7 @@ public abstract class Client {
                 // added a new host to the cluster so let the load-balancer know
                 ClusteredClient.this.cluster.loadBalancingStrategy().onNew(host);
             } catch (RuntimeException ex) {
-                final String errMsg = "Could not initialize client for " + host;
-                throw new RuntimeException(errMsg, ex);
+                throw new RuntimeException(String.format("Could not initialize client for %s.", host), ex);
             }
         };
     }
@@ -742,19 +747,36 @@ public abstract class Client {
          */
         @Override
         protected void initializeImplementation() {
-            // chooses an available host at random
-            final List<Host> hosts = cluster.allHosts()
-                    .stream().filter(Host::isAvailable).collect(Collectors.toList());
-            if (hosts.isEmpty()) throw new IllegalStateException("No available host in the cluster");
-            Collections.shuffle(hosts);
-            final Host host = hosts.get(0);
+            // chooses a host at random from all hosts
+            if (cluster.allHosts().isEmpty()) {
+                throw new IllegalStateException("No available host in the cluster");
+            }
 
+            final List<Host> hosts = new ArrayList<>(cluster.allHosts());
+            Collections.shuffle(hosts);
+            // if a host has been marked as available, use it instead
+            Optional<Host> host = hosts.stream().filter(Host::isAvailable).findFirst();
+            final Host selectedHost = host.orElse(hosts.get(0));
+
+            // only mark host as available if we can initialize the connection pool successfully
+            final CompletableFuture<Void> completableFuture =
+                    CompletableFuture.runAsync(() -> initializeConnectionSetupForHost.accept(selectedHost), cluster.executor());
             try {
-                connectionPool = new ConnectionPool(host, this, Optional.of(1), Optional.of(1));
-            } catch (RuntimeException ex) {
-                logger.error("Could not initialize client for {}", host, ex);
+                completableFuture.join();
+                selectedHost.makeAvailable();
+            } catch (CompletionException ex) {
+                logger.error("", (ex.getCause() == null) ? ex : ex.getCause());
             }
         }
+
+        private Consumer<Host> initializeConnectionSetupForHost = host -> {
+            try {
+                // hosts that don't initialize connection pools will come up as a dead host
+                connectionPool = new ConnectionPool(host, this, Optional.of(1), Optional.of(1));
+            } catch (RuntimeException ex) {
+                throw new RuntimeException(String.format("Could not initialize client for %s.", host), ex);
+            }
+        };
 
         @Override
         public boolean isClosing() {
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,6 +25,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 * Prevented XML External Entity (XXE) style attacks via `GraphMLReader` by disabling DTD and external entities by default.
 * Fixed a `NullPointerException` that could occur during a failed `Connection` initialization due to uninstantiated `AtomicInteger`.
+* Minor changes to the initialization of Java driver `Cluster` and `Client` such that hosts are marked as available only after successfully initializing connection pools.
 
 [[release-3-4-12]]
 === TinkerPop 3.4.12 (Release Date: July 19, 2021)
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
@@ -1283,8 +1283,6 @@ public final class Cluster {
 
             contactPoints.forEach(address -> {
                 final Host host = add(address);
-                if (host != null)
-                    host.makeAvailable();
             });
         }
 
--- a/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
+++ b/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
@@ -1877,4 +1877,67 @@ public class GremlinDriverIntegrateTest extends AbstractGremlinServerIntegration
 
         assertEquals(100, result1.get().one().getInt());
     }
+
+    /**
+     * Client created on an initially dead host should fail initially, and recover after the dead host has restarted
+     * @param testClusterClient - boolean flag set to test clustered client if true and sessioned client if false
+     * @throws Exception
+     */
+    public void testShouldFailOnInitiallyDeadHost(final boolean testClusterClient) throws Exception {
+        logger.info("Stopping server.");
+        this.stopServer();
+
+        final Cluster cluster = TestClientFactory.build().create();
+        final Client client = testClusterClient? cluster.connect() : cluster.connect("sessionClient");
+
+        try {
+            // try to re-issue a request now that the server is down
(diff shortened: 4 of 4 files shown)


xiazcy:
I think this might be why Travis CI isn't triggering. It says "Owner apache user license exceeded." for the build error. 
image
From https://app.travis-ci.com/github/apache/tinkerpop/requests.

spmallette:
According to Apache Infra folks it seems the ASF has hit some sort of limit with Travis - [they are working on it](https://lists.apache.org/thread.html/r992d7fe9adfeef344be64a2e37a4ed1eab3955eb90b0d8ca7b447acd%40%3Cbuilds.apache.org%3E).

spmallette:
When i run the tests locally, I am getting a failure on a `mvn clean install` in `gremlin-console` (which is testing the `:remote` command using the driver):

```text
[ERROR] Tests run: 11, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.625 s <<< FAILURE! - in org.apache.tinkerpop.gremlin.console.jsr223.DriverRemoteAcceptorTest
[ERROR] shouldConnect(org.apache.tinkerpop.gremlin.console.jsr223.DriverRemoteAcceptorTest)  Time elapsed: 0.3 s  <<< ERROR!
org.apache.tinkerpop.gremlin.jsr223.console.RemoteException: Error during 'connect' - All hosts are considered unavailable due to previous exceptions. Check the error log to find the actual reason.
	at org.apache.tinkerpop.gremlin.console.jsr223.DriverRemoteAcceptor.connect(DriverRemoteAcceptor.java:116)
	at org.apache.tinkerpop.gremlin.console.jsr223.DriverRemoteAcceptorTest.shouldConnect(DriverRemoteAcceptorTest.java:133)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:365)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:273)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:238)
	at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:159)
	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:379)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:340)
	at org.apache.maven.surefire.booter.ForkedBooter.execute(ForkedBooter.java:125)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:413)
Caused by: org.apache.tinkerpop.gremlin.driver.exception.NoHostAvailableException: All hosts are considered unavailable due to previous exceptions. Check the error log to find the actual reason.
```

The failing test tries to connect the Console to a non-existent server and since it connects lazily expects it to just allow the `:remote` to be created - it seems that now it gets some form of error. Could you please have a look at the failure and explain what's failing more exactly?

Also, for this PR it might be good to try to run this command (especially since travis isn't behaving): `mvn clean install && mvn verify -pl gremlin-server,gremlin-console -DskipIntegrationTests=false` If all that passes, you should have some good confidence in your change.

xiazcy:
Hi Stephen, thank you for running the tests! I will use the command for my local tests from now on.

For the failed test, I am not sure if there is a way to fix this reconnection issue without moving away from “lazy connection”.
From my understanding of this issue, it is due to marking a host as “available” and marking client as “initialized” before confirming the host is actually available. After this point, the host are not marked as unavailable/uninitialized, and so the client thinks it’s a valid connection and does not initiate reconnection even when the host later becomes available.

To avoid this, I’ve made the change so that hosts are not marked as available upon cluster initiation, but marked as available only after successful initialization of connection pools. This would lead to `NoHostAvailableException()` and the client’s `initialized` field remaining false if one connects to a dead host at first, and upon resubmit this would lead to client calling `init()` again inside `submitAsync()`, which would lead to successful reconnection when server/host restarts.

This technically breaks the “lazy connection” idea, making it impossible to connect to a non-existent server/host, which caused that test to fail. Confirming host availability at initialization was the most intuitive change, though is “lazy connection” a concept we should hold? Thank you!

xiazcy:
I am going to try a work around that will be a simpler fix and doesn't change the original "lazy connection" structure. I'll PR on a new branch if local tests pass on it.

xiazcy:
I've added a different approach to fix this here: https://github.com/apache/tinkerpop/pull/1478. Please let me know if I should close this one. Thank you!

divijvaidya:
Thank you for making this change and trying out multiple solutions to fix the problem.

I have added a detailed note about my thoughts in #1478 and this PR is very close to what I suggested in the other PR. Please take a look at those suggestions and implement missing changes in this PR. They address lazy connection solution as well as fix the host available problem.

xiazcy:
The new changes are as follow:
1. Cluster no longer marks host as available at its `init()`, so all hosts start with the default `isAvailable` as `false`.
2. For ClusteredClients, at `initializeImplementation`, hosts are marked as available inside `initializeConnectionSetupForHost` if Connection Pool can be initialized successfully. At the end of initiating all hosts, if no hosts are available, then `NoHostAvailableException` will be thrown here. If there is at least one host made available, then any unavailable hosts will be passed into `handleUnavailableHosts` along with the executor, which will retry the connection initialization set up in the background through `host.makeUnavailable`, and the client is successfully initiated with the available host(s). 
3. For SessionedClients, hosts are randomly chosen, and if a previously marked available host is present, it will be chosen instead. Otherwise, the chosen host is tried and marked as available if Connection Pool can be initialized successfully, and if unsuccessful it will log the error and return to the Client `init()` function to throw the `NoHostAvailableException`. 

Note that for the Gremlin Console remote acceptor test, due to the above changes, `NoHostAvailableException` will be thrown when there is no live server to connect, so I have changed it to assert that remote will throw an exception instead of being successful. Please let me know if it should behave another way (e.g. catch the exception and return an error message instead, etc.)

Please also let me know if there are any questions or concerns with the changes. Thank you!

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
```diff
@@ -526,18 +527,24 @@ protected void initializeImplementation() {
...
-                    result = cause;
-                }
+                logger.error("", (ex.getCause() == null) ? ex : ex.getCause());
+            }
 
-                logger.error("", result);
-            } finally {
+            // throw an error if there is no host available after initializing connection pool.
+            if (cluster.availableHosts().isEmpty()) {
+                throw new NoHostAvailableException();
```
divijvaidya, line 539: we should shutdown the hostExecutor when this exception is thrown else it would lead to a leak

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
```diff
@@ -561,16 +568,56 @@ protected void initializeImplementation() {
...
+
+            try {
+                // hosts that don't initialize connection pools will come up as a dead host
+                hostConnectionPools.put(host, new ConnectionPool(host, ClusteredClient.this));
+
+                // added a new host to the cluster so let the load-balancer know
+                ClusteredClient.this.cluster.loadBalancingStrategy().onNew(host);
+
+                // mark this host as available
+                host.makeAvailable();
```
divijvaidya, line 613: this is already done in `reconnected()` function called in `makeUnavailable()`

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
```diff
@@ -561,16 +568,56 @@ protected void initializeImplementation() {
...
+        /**
+         * Attempt to re-initialize the {@link Host} that was previously marked as unavailable.  This method gets called
+         * as part of a schedule in {@link Host} to periodically try to re-initialize.
+         */
+        public boolean tryReInitializeHost(final Host host) {
+            logger.debug("Trying to re-initiate host connection pool on {}", host);
+
+            try {
+                // hosts that don't initialize connection pools will come up as a dead host
+                hostConnectionPools.put(host, new ConnectionPool(host, ClusteredClient.this));
```
divijvaidya, line 607: re-use `initializeConnectionSetupForHost()` ?

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
```diff
@@ -561,16 +568,56 @@ protected void initializeImplementation() {
...
+                // added a new host to the cluster so let the load-balancer know.
                 ClusteredClient.this.cluster.loadBalancingStrategy().onNew(host);
             } catch (RuntimeException ex) {
-                final String errMsg = "Could not initialize client for " + host;
-                throw new RuntimeException(errMsg, ex);
+                throw new RuntimeException(String.format("Could not initialize client for %s.", host), ex);
             }
         };
+
+        private void handleUnavailableHosts(List<Host> unavailableHosts, ExecutorService hostExecutor) {
```
divijvaidya, line 584: why do we need the hostExecutor to re-initialize the host? Asking because makeUnavailable() already uses the cluster's scheduler pool asynchronously to run the lambda. We can hand over the reconnect/re-initialize responsibility to the scheduler from here.

divijvaidya (changes requested):

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
xiazcy: I've missed this, thanks! I've put it back in the finally clause.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
xiazcy: Done.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
xiazcy: Yes, I missed that part. Changed it. Thanks!

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
xiazcy: re-using `initializeConnectionSetupForHost()` as noted below.

divijvaidya:
Looks good.

Vote +1 for merging into master, 3.5.x and 3.4.x (assuming that test run passes)

divijvaidya (approved):

spmallette:
@xiazcy i'm sorry but i think i asked you change the behavior of this test for the Gremlin Console 

```java
    @Test
    public void shouldConnect() throws Exception {
        // there is no gremlin server running for this test, but gremlin-driver lazily connects so this should
        // be ok to just validate that a connection is created
        assertThat(acceptor.connect(Arrays.asList(Storage.toPath(TestHelper.generateTempFileFromResource(this.getClass(), "remote.yaml", ".tmp")))).toString(),
                   startsWith("Configured "));
    }
```

and perhaps that was a mistake to ask you to do that. As it currently sits, we get this behavior now:

```text
gremlin> :remote connect tinkerpop.server conf/remote.yaml
ERROR org.apache.tinkerpop.gremlin.driver.Client  - Could not initialize client for Host{address=localhost/127.0.0.1:8182, hostUri=ws://localhost:8182/gremlin}
ERROR org.apache.tinkerpop.gremlin.driver.Client  - 
java.net.ConnectException: Connection refused
	at java.base/sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)
	at java.base/sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:777)
	at io.netty.channel.socket.nio.NioSocketChannel.doFinishConnect(NioSocketChannel.java:330)
	at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.finishConnect(AbstractNioChannel.java:334)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:707)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:655)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:581)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:493)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:989)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at java.base/java.lang.Thread.run(Thread.java:829)
==>Error during 'connect' - java.net.ConnectException: Connection refused
```

Not only do we get the big stacktrace, behind the scenes but the `Cluster` object is initialized and would be trying to reconnect to the server when it comes online. I think that the Gremlin Console could handle this issue a bit better by catching the `NoHostAvailableException` explicitly and reporting a more user friendly message, like "Configured <cluster> <session> but host not presently available". At that point we could change that test back to what it was and the behavior would be the same as before but with a far nicer message.

Could you please submit a PR for that to complete this issue? You could submit the work under the same JIRA ticket. 

Some additional details:

1. The exception is thrown [here](https://github.com/apache/tinkerpop/blob/3.4-dev/gremlin-console/src/main/java/org/apache/tinkerpop/gremlin/console/jsr223/DriverRemoteAcceptor.java#L111) so you would just add the catch for `NoHostAvailableException` there.
2. Rather than throw an exception in that catch you would `return` the nice error message
3. A bit of a side issue that I noticed while looking at this was the code path where an exception is propagated from that code in the link above and the `RemoteCommand` handles it [here](https://github.com/apache/tinkerpop/blob/3.4-dev/gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/commands/RemoteCommand.groovy#L55). I feel like something isn't quite right there because a `remote` will never have `close()` called on it to release the `Cluster` object. So maybe there is a leak there? Perhaps you could take a minute to analyze that and see if a little fix is necessary there?

samair:
Thanks, @spmallette, we see these problems now frequently with AWS Neptune. We have multiple pods which connect to Neptune. Sometimes one of the pods for some reason does not connect the first time and requests made from that pod fail till we manually restart. Would be very helpful if we can push this fix early.

PR #1478, opened 2021-09-23
TINKERPOP-2569: Simple Fix for Reconnect

xiazcy:

The issue is that the client's `initialized` field is set to `true` as long as hosts added, even though the hosts may be dead and the connection pools failed to initialize, and `submitAsync()` will only re-init() the client (along with connection pools) if `initialized` is set to false. So even if the server comes back up, the client will not try to reconnect. 

Original approach is to confirm host availability and throw exceptions during client `init()` when hosts are dead, so that the client's `initialization` is not set to true, however, this would fail tests that rely on the "lazy connection" concept (i.e. hosts would need to be available at the time driver connects). To avoid that, this is a simpler fix that adds a boolean flag `noLiveHostAvailable` to indicate when there is no live host at client initialization, so we can re-init() the client in the case it is `true` (in addition to checking the client `initialized` flag).

This `noLiveHostAvailable`  flag is set to `true` initially and is changed to `false` when any host initiates a connection pool successfully.

Builds successfully with `mvn clean install && mvn verify -pl gremlin-server,gremlin-console -DskipIntegrationTests=false`.


spmallette:
This flag approach seems simple enough and just costs another boolean to reason about. Let's see what @divijvaidya thinks, but it's encouraging that no tests broke and you managed to get the failing test case to pass by way of this change.

divijvaidya:
Let me try to phrase it as I understand the situation and tell me if I missed anything.

As per the current code base, the definition of client initialisation is that successful connections must have been established with at least one host in the cluster.

Currently, when the client is trying to send it's first request, `init()` will be called. `init()` tries to initializes connections across all hosts registered in the cluster and if one of the hosts is down, client will log an error. Note that if the initialization fails, the client will not throw an error or provide any response indicating that initialization has failed. This is intended behaviour because as per the definition of "successful initialization", a client is successful if at least one of the host has connected successfully. 

If no host is successful, then the client initilization **would not fail with a `NoHostAvailableException` exception**. This is because we only throw this exception if the host has been marked as "available" in the cluster. It is not necessary for a host to have active connections to be marked as "available" in the cluster. Hence, even though connection initilization to the host has failed, host is still marked as "active". This will lead to successfully setting `Client.initialized` to true which is incorrect based on the definition of client initialization. 

The code change in this PR sets a flag `noLiveHostAvailable` if at least one of the hosts fail during client initialization. 
On the next request, it then tries to init() the client again for all the hosts registered in the cluster.

Let's consider how this change will perform in different cases.

Case#1: One host registered to cluster and the host is down during client init

In such cases, init() will not throw a `NoHostAvailableException` exception. `Client.initialized` will be set to true and `Client.noLiveHostAvailable` will be false. When a second request is made to the client, it will try to init() again and if the host is alive at that time, it will succeed.

Note that the new flag is doing what the original `Client.initialized` flag should have been doing here.

Case#2: Multiple hosts registered and only one host is down during client init
In such cases, init() will not throw a `NoHostAvailableException` exception. `Client.initialized` will be set to true and `Client.noLiveHostAvailable` will be false. When a second request is made to the client, it will try to init() again against all the hosts. This will lead to successful hosts also creating a new connection pool (which is wrong and wasteful).

Due to above problems in case#2, this PR is not a suitable change.

I would propose the following alternative:

A client should:
1. Mark the host as available in the cluster only after connections to the host have been initialised.
2. During client initialization, hosts with successful connections should be marked as available and hosts with unsuccessful connection should be marked as unavailable.
3. If no hosts are successful during client init, NoHostAvailable exception should be thrown to the user and let the user handle this scenario. This is because different applications might have different behaviour. Some might want to fail their application startup while some might want to keep retrying until successful. We should provide this flexibility to the applications to decide this for themselves.
4. If some hosts are successful and some aren't, we should mark the unavailable ones as `considerHostUnavailable()`. This would keep on retrying the failed hosts in the backend while successful hosts are serving existing connections.

I would suggest you to make the changes to follow the above points and that will make the original test pass (since init() will be called on every call for case of a single host, since client.initialized will be false)

xiazcy:
Hi Divij, thank you so much for your suggestions!

Just as a response to the two cases you outlined, and to make sure we are on the same page, re-init() is actually triggered when `Client.noLiveHostAvailable` is set to `true`, and not `false`. This flag is set to `true` at beginning of `Client.initializeImplementation()`, and is set to `false` if at least one host is alive during initialization, not the other way around. This follows an "assume no live hosts unless proven otherwise" logic. 

More specifically, in the case when there are dead hosts among other live hosts, `Client.noLiveHostAvailable` is first set to `true`, and then set to `false` when any live hosts successfully initialized a connection pool inside the async call to `Client.initializeConnectionSetupForHost`. This also leaves `Client.initialized` as `true`, which shouldn’t trigger `Client.init()` at `if (!initialized || noLiveHostAvailable)` inside `Client.submitAsync()`, and if `Client.init()` is somehow called, its `if (initialized && !noLiveHostAvailable)` check would be `true` and any further initialization would be skipped. I believe this should avoid Case#2.

But in any case, this change doesn’t resolve the fact that `Client.initialized` is set as `true` when by definition it shouldn’t when no host is available, and I’m working on a way to approach this based on your suggestions.

xiazcy:
Will do, closing this.

PR #1479, opened 2021-09-23
TINKERPOP-2616: Provide better exceptions with SSL related failures

simonz-bq:
https://issues.apache.org/jira/browse/TINKERPOP-2616

NoAvailableHostException now contains the context for what caused it. A limitation to this is that it takes the last failed ConnectionPool initialization as its cause. Alternatively I considered adding a list of exceptions to NoAvailableHostException so that the user could see every failure - not sure if this is desirable though - let me know if this is something that is thought of as a good idea and I can update this PR.

Also, some of the exception handling needed some care, specifically there was a lot of `catch (Exception e)` blocks which were unnecessarily wrapping RuntimeExceptions, and any children Exceptions, in a new RuntimeException when they could be directly returned to the user.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
@@ -19,6 +19,7 @@
 package org.apache.tinkerpop.gremlin.driver;
 
 import org.apache.commons.lang3.concurrent.BasicThreadFactory;
+import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.tinkerpop.gremlin.driver.exception.ConnectionException;
@@ -66,6 +67,7 @@ public abstract class Client {
     protected final Cluster cluster;
     protected volatile boolean initialized;
     protected final Client.Settings settings;
+    protected Throwable initializationFailure = null;
 
     Client(final Cluster cluster, final Client.Settings settings) {
         this.cluster = cluster;
@@ -124,8 +126,8 @@ public abstract class Client {
     public ResultSet submit(final Traversal traversal) {
         try {
             return submitAsync(traversal).get();
-        } catch (UnsupportedOperationException uoe) {
-            throw uoe;
+        } catch (RuntimeException re) {
+            throw re;
         } catch (Exception ex) {
             throw new RuntimeException(ex);
         }
@@ -149,8 +151,8 @@ public abstract class Client {
     public ResultSet submit(final Bytecode bytecode) {
         try {
             return submitAsync(bytecode).get();
-        } catch (UnsupportedOperationException uoe) {
-            throw uoe;
+        } catch (RuntimeException re) {
+            throw re;
         } catch (Exception ex) {
             throw new RuntimeException(ex);
         }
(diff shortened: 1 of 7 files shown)


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
```diff
@@ -124,8 +126,8 @@ public Client alias(final Map<String, String> aliases) {
     public ResultSet submit(final Traversal traversal) {
         try {
             return submitAsync(traversal).get();
-        } catch (UnsupportedOperationException uoe) {
-            throw uoe;
+        } catch (RuntimeException re) {
```
divijvaidya, line 129: Thank you for these changes.

1. The top level exception thrown back to the application has changed here from `RuntimeException` to `NoHostAvailableException`. If a user has application level code that is catching the `RuntimeException` and checking the cause as `NoHostAvailableException`, that code will fail with client upgrade. Hence, this is a backward incompatible change since it modifies the exception contract.
For 3.4.x, could we please keep the contract as same (i.e. by wrapping NoHostAvailableException in RuntimeException)?

2. All other changes in nested exceptions look good (we can still have `NoHostAvailableException` with a nice cause).

3. I would vote that we can take this break in exception contract for 3.5.x series since that is still going through some stabilization with respect to other behaviours as well such as null semantics. @spmallette agree?

Overall, I think we do the following:
1. Merge this to master.
2. Create a different version with NoHostAvailableException as a 1st level cause to merge into 3.4 branch.

divijvaidya:
Vote +1 for merge into master and 3.5 branch.

It needs modification to merge into 3.4 branch though which could be done in a separate PR.

spmallette:
> I would vote that we can take this break in exception contract for 3.5.x series since that is still going through some stabilization with respect to other behaviours as well such as null semantics. 

For this particular modification, I'd be ok with taking this minor breaking change in 3.5-dev as long as there is an upgrade doc entry to support it. 

Backporting the change to 3.4-dev without the breaking change also makes sense to me.  Assuming all that is in place:

VOTE +1

spmallette:
@simonz-bq i've merged this - here's some additional points:

1. `master` failed on this test `shouldEnableSslAndFailIfProtocolsDontMatch` - something is different there where we don't get an `SSLException` there but a `ClosedChannelException`. I'm not sure why but i changed the behavior temporarily ( e7c889967a5c40de6e4d5086e8e7d547ba6f406b ) Could you please investigate that and see if we can get `SSLException` there so that behavior can remain consistent? 
2. now that this is merged, could you backport the change in a non-breaking way as @divijvaidya suggested? i suppose that means, retaining the old exception contract but pushing `SSLException` into the "cause". if you think there is something particularly hard in doing that please let us know. we're in the "very minor enhancements and bug fixes only" phase of 3.4-dev so if it's anything terribly complicated perhaps we can just count it out of scope.

You can just submit whatever fix PRs are necessary on this JIRA issue. Taken together, that should close this one out.

spmallette:
I've figured this out now. The JDK version did have something to do with it, but it wasn't a JDK8 v JDK11 issue. The failure on JDK11 could be recreated on JDK8 if you used the newest version of JDK8 (at 292). This change that disabled TLS1.1 by default was causing the server to close the channel early producing a different sort of exception to the client (i.e. ChannelClosedException):

https://access.redhat.com/documentation/en-us/openjdk/8/html/release_notes_for_openjdk_8.0.292/openjdk8-windows-features-and-enhancements#disabled_tls_1_0_and_1_1_versions

To fix, I modified the test to use TLS1.2 on the server and TLS1.3 on the client and that allowed the server to configure SSL properly and allowed the test semantics to assert an SSLException. 

@simonz-bq please backport the changes of this commit (dc051d8283ef4278940813591b4bd3dbdbe7974a ) to #1484 - as I look now I was a bit hasty in that commit as I was just eager to see it working and pushed. perhaps you could also:

1. add a comments to that test so that in some distant future if this test starts failing we'll recall that "Server TLS version must be enabled by default in the jdk for this test to work properly"
2. fix the `fail()` message in that test to mention the right versions of TLS being used.

PR #1481, opened 2021-09-24
TINKERPOP-2621 Allowed grammar to support empty arguments to within/without()

spmallette:
This functionality was allowed by Java/Groovy so it seems to follow that we have the same here:

```text
gremlin> P.within()
==>within([])
```

Builds with `mvn clean install` which means that the variants were content with this syntax as well. 

VOTE +1


joshsh:
VOTE +1

divijvaidya:
Could we also please create a JIRA to write an optimizer to handle empty within()/without() steps from the query? Some examples of how the optimizers will behave are as follows (I can add more details in the JIRA later):
1. `has(key, within(''))` is always false in which case we can replace `within('')` with `false`. 
2. `has(hey, without(''))` is always true in which case this step/`HasContainer` can be removed from the traversal as a filter.

spmallette:
> Could we also please create a JIRA to write an optimizer to handle empty within()/without() steps from the query? 

Sure, i was more focused on just getting this to work for this PR but the thought crossed my mind that we could probably treat these cases a bit better.  Here's the JIRA: https://issues.apache.org/jira/browse/TINKERPOP-2625

PR #1291, opened 2020-06-02
TINKERPOP-2143 JavaScript GLV: Support browsers

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2143

The original PR for this is at #1070. The changes were simply brought here for ongoing rebasing until someone picks up the work and completes it. As it stands there are failing tests on a standard build.

PR #1482, opened 2021-09-30
change js remote to use isomorphic ws so can be used in browser also

gerardtoconnor:
current js implementation can only be used on node/server, and main reason is due to dependency on ws and its advanced options. if we use `isomorphic-ws` the client can be used on browser also, while still resolving to `ws`, with advanced options if deployed on node/server

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
@@ -23,7 +23,7 @@
 'use strict';
 
 const EventEmitter = require('events');
-const WebSocket = require('ws');
+const WebSocket = require('isomorphic-ws');
 const util = require('util');
 const utils = require('../utils');
 const serializer = require('../structure/io/graph-serializer');


spmallette:
thanks for this @gerardtoconnor but aren't there other things that need to be change for gremlin-javascript to work in a browser? specifically i'm thinking of the open PR #1291 and the work that remained there. cc/ @jorgebay

jorgebay:
Thanks for looking into this @gerardtoconnor !

Makes sense to start switching to something that supports both node.js and the browser.

I think it requires deeper changes:
- Support Buffer in browser: e.g https://github.com/feross/buffer
- Support for custom headers or sane defaults for setting the mime type.
- Some kind of testing

@gerardtoconnor what are your thoughts on that?  :)

gerardtoconnor:
Thanks @spmallette & @jorgebay I had missed those deeper change requirements so thank you for the insight. The existing PR does not look too far off, although needs tests etc, do we feel that we can work on that then or has there been any major stumbling  points? @spmallette are others able to contribute to your fork or should I re-fork and open a new PR (I'd rather not and keep it under that PR).

spmallette:
you can definitely submit a PR against the TINKERPOP-2143 branch. I've just rebased it for you on master:

https://github.com/apache/tinkerpop/tree/TINKERPOP-2143

i would think that you could close this PR at this point and just submit a fresh one from that branch when you are ready. It would be excellent if you could finish this work off for us. thank you!

PR #1483, opened 2021-10-01
TINKERPOP-2620 Fixed NPE related to null arguments to certain steps

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2620

Resolved some troubles related to elementMap(), valueMap(), values() and properties(). See the JIRA for examples.


krlawrence:
Thanks for cleaning these up. Always nice to get rid of bad user facing exceptions. VOTE +1

PR #1502, opened 2021-11-19
TINKERPOP-2626 Prevent premature close of traversal

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2626

This is not a complete fix for this issue but does get rid of the premature close. Opted to try to close the Traversal as part of hasNext() as in a sense it is a way to complete iteration of the traversal. Without that change certain tests begin to fail so some additional change besides removing the premature close had to be added.

VOTE +1

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
@@ -193,7 +193,9 @@ public class DefaultTraversal<S, E> implements Traversal.Admin<S, E> {
     @Override
     public boolean hasNext() {
         if (!this.locked) this.applyStrategies();
-        return this.lastTraverser.bulk() > 0L || this.finalEndStep.hasNext();
+        final boolean more = this.lastTraverser.bulk() > 0L || this.finalEndStep.hasNext();
+        if (!more) CloseableIterator.closeIterator(this);
+        return more;
     }
 
     @Override
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -23,6 +23,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 [[release-3-4-13]]
 === TinkerPop 3.4.13 (Release Date: NOT OFFICIALLY RELEASED YET)
 
+* Fixed `RangeGlobalStep` which was prematurely closing the iterator.
 * Prevented XML External Entity (XXE) style attacks via `GraphMLReader` by disabling DTD and external entities by default.
 * Improved error message for failed serialization for HTTP-based requests.
 * Fixed a `NullPointerException` that could occur during a failed `Connection` initialization due to uninstantiated `AtomicInteger`.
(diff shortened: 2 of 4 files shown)


saikiranboga (approved):

divijvaidya:
LGTM. Vote +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
```diff
@@ -193,7 +193,9 @@ public void applyStrategies() throws IllegalStateException {
     @Override
     public boolean hasNext() {
         if (!this.locked) this.applyStrategies();
-        return this.lastTraverser.bulk() > 0L || this.finalEndStep.hasNext();
+        final boolean more = this.lastTraverser.bulk() > 0L || this.finalEndStep.hasNext();
+        if (!more) CloseableIterator.closeIterator(this);
```
FlorianHockmann, line 197: Just a comprehension question: Why does it make sense to close the iterator for `hasNext()`, but not do the same for the other terminal steps?

FlorianHockmann (approved):
I added one question where I'm probably just lacking the context as I haven't worked much on this core part of TinkerPop, but the change LGTM in general.

VOTE +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
spmallette: the other terminal steps tend to call `fill()` which performs a close properly.

divijvaidya:
I thought of a bug in this implementation. 

When we call `hasNext()` first time and there are no more results, we will close the traversal. `hasNext()` is supposed to be an idempotent operation. If we call `hasNext()` again on this closed traversal, we may encounter unexpected behaviour because we are again trying to find a result on a traversal that has already been closed (and the underlying storage implementation might have released resources).

The fix would be to add a lifecycle state in Traversal.java. Set the lifecycle state to closed when close() is called. Once it is closed, any call to hasNext should return false and any call to next() should return NoSuchElement() exception.

spmallette:
after looking at this a bit, i'm not sure i see how we can get this change in without some form of breaking change to the `Traversal` API. @divijvaidya maybe i'm not envisioning the change as you were seeing it happen...thoughts?

divijvaidya:
> after looking at this a bit, i'm not sure i see how we can get this change in without some form of breaking change to the `Traversal` API. @divijvaidya maybe i'm not envisioning the change as you were seeing it happen...thoughts?

I am proposing the following changes, let me know if they sound good to you.
1. Add an abstract method to `Traversal.java` called `notifyClosed()`
2. That method should be called inside `close()` method of `Traversal.java`
3. `DefaultTraversal.java` should have a member called `isClosed` which would be set to `true` inside it's implementation of `notifyClosed()`. `DefaultTraversal` would also have a method called `isClosed()` which will return the value of `isClosed` member.
4. In the `hasNext()` method of `DefaultTraversal`, it will check if `isClosed()` is true and if yes, then it will return false. Note that this is not breaking in the sense that today the behaviour of iterator hasNext() after being closed in undefined. We are correcting the semantics here such that you cannot access the iterator when it is closed already. Same change for `next()`, it should return NoSuchElement exception when called on closed iterator.

spmallette:
I had tried a similar approach - I stopped when i noticed around item 3 that `isClosed()` could only work nicely on an implementation like `DefaultTraversal`. As there are other `Traversal` implementations to consider putting it where it ideally goes as a member of `Traversal` won't come without a breaking change which is why i asked for your thoughts.

I can implement this in 3.5.2 as you describe it and then take a breaking change in 3.6.0 after this merges to `master` to promote `isClosed()` to `Traversal`.

PR #1503, opened 2021-11-19
TINKERPOP-2650 Removed deprecated gremlin-python step names

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2650

Gremlin step names that conflicted with python keywords have been removed. Users will need to prefer the proper underscore suffixed step names.

VOTE +1


krlawrence:
My build system is currently having issues building `glv-python` but I read over the changes and they look good. 

VOTE +1

PR #1506, opened 2021-11-24
TINKERPOP-2656 update python translator to provide an option for non synantic sugar translations

tkolanko:
https://issues.apache.org/jira/browse/TINKERPOP-2656

This PR implements two changes to the python translator

1. The `SymbolHelper` map has been updated to include the missing
reserved words/global functions differences for built in Python keywords
https://tinkerpop.apache.org/docs/3.5.1/reference/#gremlin-python-differences
2. A new translator type, `NoSugarTranslator`, has been added which
translates the Gremlin query without using any of the syntax sugar.

Given the following query:
```
g.V().range(0, 10).has("person", "name", "marko").limit(2).values("name")
```

Default translator:
```
g.V()[0:10].has('person','name','marko')[0:2].name
```

No sugar translator:
```
g.V().range_(0,10).has('person','name','marko').limit(2).values('name')
```

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -23,6 +23,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 [[release-3-5-2]]
 === TinkerPop 3.5.2 (Release Date: NOT OFFICIALLY RELEASED YET)
 
+* Added a `NoSugarTranslator` translator to `PythonTranslator` which translates Gremlin queries to Python without syntactic sugar (ex `g.V().limit(1)` instead of `g.V()[0:1]`)
 * Added support for `g.Tx()` in .NET.
 * Added support for `with()` constant options to `io()`.
 * Changed `GroovyTranslator` to generate code more compatible to Java with `Date` and `Timestamp`.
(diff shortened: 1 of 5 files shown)


spmallette:
> The SymbolHelper map has been updated to include the missing
reserved words/global functions differences for built in Python keywords

I made a similar change on #1530 headed to `master`. I was going to backport it but this works too. 

as this is your second PR, i'd point out that we like a CHANGELOG entry included with changes - a bullet or two that describes the nature of your changes (this is separate from including JIRA information which you also see in the CHANGELOG which is added when we do a release). If you'd like to update this PR with such an entry you would add it for 3.5.2 here:

https://github.com/apache/tinkerpop/blob/master/CHANGELOG.asciidoc#tinkerpop-352-release-date-not-officially-released-yet

Otherwise, no worries this time - I can add it when I merge.  From an administrative perspective, It would be nice if you  could rebase/squash this PR to a single commit so the git history stays nice/compact.

Thanks for the contribution! Hope to see you again 🙂 

VOTE +1

tkolanko:
Ok thanks - I'll add the changelog update and also update the docs to mention that there is an option to disable the syntax sugar translations to python (and squash everything down to one commit)

PR #1507, opened 2021-11-24
Fix a rare function call causing a crash

Shushishtok:
Relevant Discord discussion:
https://discord.com/channels/838910279550238720/844295727487647744/913113345341657098
https://discord.com/channels/838910279550238720/894915500709117972/913119924103282719

Following a rare crash that I've encountered while working with Gremlin, a small fix that might cover a specific case that can throw an error.

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/process/translator.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/process/translator.js
@@ -80,7 +80,7 @@ class Translator {
             } else if (Array.isArray(params[k])) {
               const parts = [];
               for (const param of params[k]) {
-                parts.push(translate(param.getBytecode(), true));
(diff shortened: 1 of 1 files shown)


spmallette:
This test triggers the problem:

```js
    it('should produce valid script representation from a traversal object', function () {
      const g = new graph.Graph().traversal();
      const script = new Translator('g').translate(g.V([1, 2, 3]));
      assert.ok(script);
      assert.strictEqual(script, 'g.V([1, 2, 3])');
    });
```

and should be incorporated. the added issue is that the fix in this PR is only partial because the value in the array isn't a `Traversal` and thus a recursive call to `translate()` just fails. 

Created an issue for tracking as this needs a bit more thought:

https://issues.apache.org/jira/browse/TINKERPOP-2658

spmallette:
closing this one in favor of #1509

PR #1510, opened 2021-12-07
TINKERPOP-2660 Added back close message to drivers

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2660

Adding them back may just be temporary, so the message remains deprecated on the server side. Session semantics needs better definition going forward at which point the final disposition for this message can be established.

VOTE +1


jorgebay (approved):
lgtm, VOTE +1

FlorianHockmann (approved):
VOTE +1

PR #1514, opened 2021-12-08
TINKERPOP-2635 Better by() consistency

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2635
https://issues.apache.org/jira/browse/TINKERPOP-2636
https://issues.apache.org/jira/browse/TINKERPOP-2608

Covers a number of wide changes around `by()` modulators which makes it work consistently for all its cases. See the upgrade doc changes for more information.

VOTE +1


anassisi2:
Hi Stephen,
I'm not sure where to post the issue below, it seems you are working on gremlin-javascript.
I get a "ReferenceError: outE is not defined" when I use it after local or repeat like:

g.V().hasLabel('assembly', 'part').
  local(outE("version").order().by('ver', desc).limit(1)).inV().
  path().
  by(__.elementMap()).
  by(__.valueMap()).
  by(__.valueMap()).
  toList();        

I'm using ver 3.5.1 installed via Npm.

Thank you,
Andrea

spmallette:
you may ask question in Discord, stackoverflow, etc. Quickly though, you probably need to just import `outE()`:

https://tinkerpop.apache.org/docs/current/reference/#gremlin-javascript-imports

```js
const __ = gremlin.process.statics;
```

and then you should be able to reference it as `__.outE()` or import it as `outE()` explicitly.

anassisi2:
Thank you Stephen!

PR #1516, opened 2021-12-14
TINKERPOP-2663 Allowed for Vertex/ReferenceVertex in grammar

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2663

This is helpful for `addE()` with `from()` and `to()` to be consistent with `Vertex` object signatures.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -34,6 +34,7 @@ limitations under the License.
 * Improved behavior of `V()` and `E()` when `null` is an argument producing a filtering behavior rather than an exception.
 * Prevented metrics computation unless the traversal is in a locked state.
 * Added syntax to Gremlin grammar to explicitly define `byte`, `short` and `BigInteger`.
+* Added syntax to Gremlin grammar to allow construction of a reference `Vertex`.
 * Created a way to produce a corpus of Gremlin traversals via `FeatureReader` and `DocumentationReader` in `gremlin-language`.
 * Exposed Gherkin tests as part of the provider test suite.
 * Packaged Gherkin tests and data as standalone package as a convenience distribution.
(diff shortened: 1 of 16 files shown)


gremlin-language/src/main/antlr4/Gremlin.g4
```diff
@@ -740,24 +742,28 @@ traversalMethod_write
     ARGUMENT AND TERMINAL RULES
 **********************************************/
 
+structureVertex
+    : NEW ('Vertex'|'ReferenceVertex') LPAREN genericLiteral COMMA stringLiteral RPAREN
```
divijvaidya, line 746: ReferenceVertex can be constructed without a label too in which case the label is automatically assigned to "vertex". See https://github.com/apache/tinkerpop/blob/7f7d3a485c7f100f98047b71672a0c2c9ab855b4/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/reference/ReferenceElement.java#L61 

I think what you are doing is correct, i.e. to have users specify vertex label explicitly but for completeness sake, please add a comment here on why we don't allow a reference vertex with just an ID.

spmallette: added a comment to the grammar file.

CHANGELOG.asciidoc
```diff
@@ -34,6 +34,7 @@ limitations under the License.
 * Improved behavior of `V()` and `E()` when `null` is an argument producing a filtering behavior rather than an exception.
 * Prevented metrics computation unless the traversal is in a locked state.
 * Added syntax to Gremlin grammar to explicitly define `byte`, `short` and `BigInteger`.
+* Added syntax to Gremlin grammar to allow construction of a reference `Vertex`.
```
divijvaidya, line 37: Shall we add an example in the changelog here to let the users know on what is possible in the syntax now?

divijvaidya:
Vote +1

Minor comment to add an example in the changelog so that users can see what new syntax is available. Other than that no concerns.

CHANGELOG.asciidoc
spmallette: I think we will handle all the docs around gremlin-language with this: https://issues.apache.org/jira/browse/TINKERPOP-2644

PR #1517, opened 2021-12-14
TINKERPOP-2667 Allowed fold()/addAll to merge Map objects

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2667

The following works now:

```text
gremlin> g.inject([a: 1],[b:2]).fold([], addAll)
==>[[a:1],[b:2]]
gremlin> g.inject([a: 1],[b:2]).fold([:], addAll)
==>[a:1,b:2]
```

Interestingly, this functionality used to result in an exception so it really isn't a change in behavior which means it could slip in on 3.5-dev.

VOTE +1


krlawrence:
This is a lot simpler than the way we currently have people do it with `unfold/group/fold`

VOTE +1

PR #1523, opened 2021-12-16
Added an `AnonymizingTypeTranslator` for use with `GroovyTranslator` …

mikepersonick:
…which strips PII (anonymizes any String, Numeric, Date, Timestamp, or UUID data).

  https://issues.apache.org/jira/browse/TINKERPOP-2666


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/translator/AnonymizingTypeTranslator.java
```diff
@@ -0,0 +1,170 @@
...
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package org.apache.tinkerpop.gremlin.process.traversal.translator;
+
+import org.apache.tinkerpop.gremlin.process.traversal.*;
```
divijvaidya, line 22: Please expand all imports to match the convention of rest of the code base.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/translator/AnonymizingTypeTranslatorTest.java
```diff
@@ -0,0 +1,88 @@
...
+import org.junit.Test;
+
+import java.util.Arrays;
+
+public class AnonymizingTypeTranslatorTest {
+
+    private void testAnonymize(final String query, final String expected) {
+        try {
+            final Bytecode bc = (Bytecode) GremlinQueryParser.parse(query, new NoOpTerminalVisitor());
+            final Translator.ScriptTranslator translator = GroovyTranslator.of("g", new AnonymizingTypeTranslator());
```
divijvaidya, line 37: would this work if the `GraphTraversalSourceAlias` is not "g" (which is not necessary in Gremlin language)?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/translator/AnonymizingTypeTranslator.java
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * This default implementation keeps a map from type (Java Class) to another map from instances to anonymized
+     * token.
+     */
+    public static class DefaultAnonymizer implements Anonymizer {
+        /*
+         * Map<ClassName, Map<Object, AnonymizedValue>>
+         */
+        private final Map<String, Map<Object, String>> simpleNameToObjectCache = new HashMap<>();
```
divijvaidya, line 62: Correct me if I am wrong here but this map will not be garbage collected until JVM is destroyed (since it is allocated on static initialization of this class). With large number of queries, this will eat up heap memory quickly. 

Is it possible to use weak reference here or clear the map or move this class outside (non-inner)?

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/translator/AnonymizingTypeTranslatorTest.java
mikepersonick: The test wouldn't work, but the test uses "g". Should be fine. For production code you would set up the GroovyTranslator differently.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/translator/AnonymizingTypeTranslator.java
mikepersonick: The map is not static, just the class. So the cache will be garbage collected when the Anonymizer goes out of scope (right after translation).

divijvaidya: Yes I understand that. I was wondering if there is a possibility where the class does not go out of scope through the lifecycle of JVM but seems like we create a new instance of anonymizer every time and it goes out of scope quite quickly, so my point is not valid.

divijvaidya (approved):

divijvaidya:
Vote +1

PR #1533, opened 2021-12-21
Tinkerpop 2665 - Add ability for property() to take a map

bechbd:
Added the ability for property() to take a map as an argument per https://issues.apache.org/jira/browse/TINKERPOP-2665


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/__.java
```diff
@@ -980,7 +981,18 @@ protected __() {
...
+        start().property(cardinality, key, value, keyValues);
+    }
+
+    /**
+     * @see GraphTraversal#property(Object)
+     */
+    public static <A> GraphTraversal<A, A> property(final Object value) {
+        return __.<A>
+
+                start().property(value);
```
spmallette, line 995: is this extra spacing here intended? could you please fold both this one and the previous back to one line?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -2364,14 +2365,57 @@ else if (value instanceof Traversal)
      * @since 3.0.0-incubating
      */
     public default GraphTraversal<S, E> property(final Object key, final Object value, final Object... keyValues) {
```
spmallette, line 2367: could you please have a look at the javadoc for both of these `property()` methods - they likely need updating given all this change, right?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -2364,14 +2365,57 @@ else if (value instanceof Traversal)
...
+                return this.property((VertexProperty.Cardinality) key, value, null == keyValues ? null : keyValues[0],
                         keyValues != null && keyValues.length > 1 ?
                                 Arrays.copyOfRange(keyValues, 1, keyValues.length) :
-                                new Object[]{}) :
-                this.property(null, key, value, keyValues);
+                                new Object[]{});
+            }
+        } else  { //handles if cardinality is not the first parameter
+            if (value instanceof Map) { //Handle the property(Cardinality, Map) signature
+                Map<Object, Object> map = (Map)value;
```
spmallette, line 2383: please give a look to your local variable declarations in this PR - we dont have many style rules but we like to prefer the use of the `final` keyword where it can be used.

gremlin-language/src/main/antlr4/Gremlin.g4
```diff
@@ -592,12 +592,16 @@ traversalMethod_properties
 traversalMethod_property
 	: 'property' LPAREN traversalCardinality COMMA genericLiteral COMMA genericLiteral (COMMA genericLiteralList)? RPAREN #traversalMethod_property_Cardinality_Object_Object_Object
 	| 'property' LPAREN genericLiteral COMMA genericLiteral (COMMA genericLiteralList)? RPAREN #traversalMethod_property_Object_Object_Object
+	| 'property' LPAREN genericLiteralMap? RPAREN # traversalMethod_property_Object
```
spmallette, line 595: should there be a question mark here? doesn't that make the argument optional so that it will also match `property()`? similar sort of question in the definition below....

i forget how antrl matches but i think it's in the order of definition. perhaps it would then be best to put these more specific sorts of definitions above the other two. it doesn't look like the conflict but in my mind we should go more specific to more general (genericLiteralMap -> genericLiteral).

gremlin-language/src/main/antlr4/Gremlin.g4
```diff
@@ -592,12 +592,16 @@ traversalMethod_properties
...
+	| 'property' LPAREN traversalCardinality COMMA genericLiteralMap? RPAREN # traversalMethod_property_Cardinality_Object
 	;
 
 traversalMethod_propertyMap
 	: 'propertyMap' LPAREN stringLiteralList RPAREN
 	;
 
+
```
spmallette, line 603: nit: some extra lines to remove

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/language/grammar/TraversalMethodVisitorTest.java
```diff
@@ -696,6 +700,30 @@ public void testTraversalMethod_property_Object_Object_Object() throws Exception
         compare(g.V().property(1,2,"key", 4), eval("g.V().property(1,2,'key',4)"));
     }
 
+    @Test
+    public void testTraversalMethod_property_Object() throws Exception {
+        LinkedHashMap<Object, Object> map = new LinkedHashMap<>();
```
spmallette, line 705: can this work without a `LinkedHashMap` here? i think it would be preferable to test with a regular old `HashMap`.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -598,15 +598,22 @@ previous step to enable vertex and edge creation with all its properties in one
...
+g.addV().property(T.label,'person').valueMap().with(WithOptions.tokens) <6>
+----
+
+<1> Properties can also take a Map as an argument.
+<2> For vertices, a cardinality can be provided for <<vertex-properties,vertex properties>>.
+<3> If a cardinality is specified for a Map then that cardinality will be used for all properties in the map.  If you need different cardinalities per property then you should individually add the property values.
+<4> It is possible to select the property value (as well as key) via a traversal.
+<5> For vertices, the `property()`-step can add meta-properties.
+<5> The label value can be specified as a property only at the time a vertex is added and if one is not specified in the addV()
+<5> If passing in a map the token T.label must be surrounded by parenthesis when using Groovy
```
spmallette, line 616: good addition to the docs, though is your numbering right there (i.e. three "5" in a row)? i think that this change should also include some simple Upgrade Documentation.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/__.java
bechbd: No that was an auto-formatting change.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/language/grammar/TraversalMethodVisitorTest.java
bechbd: If you test with a regular HashMap you have no guarantee that the items will appear in the Bytecode in the same order as the eval() statement.  It will compile and run but randomly fail do to the non-deterministic aspects.

spmallette: ok, so the `LinkedHashMap` is for testing purposes

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/__.java
```diff
@@ -37,6 +37,7 @@
 
 import java.util.Collection;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
```
spmallette, line 40: nit: can this import be removed?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/GenericLiteralVisitor.java
```diff
@@ -30,10 +30,7 @@
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
```
spmallette, line 33: We dont have many code style requirements but one is to not wildcard import. you're IDE probably was the culprit there.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/TraversalMethodVisitor.java
```diff
@@ -1145,6 +1148,23 @@ public GraphTraversal visitTraversalMethod_property_Object_Object_Object(final G
                 GenericLiteralVisitor.getGenericLiteralList(ctx.genericLiteralList()));
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Traversal visitTraversalMethod_property_Cardinality_Object(
```
spmallette, line 1155: the other code style item we have is to mark variables `final` when possible.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -2272,6 +2273,10 @@ else if (value instanceof Traversal)
      * and if the {@link Element} is a {@link VertexProperty}.  This method is the long-hand version of
      * {@link #property(Object, Object, Object...)} with the difference that the {@link VertexProperty.Cardinality}
      * can be supplied.
+     * <p/>* 
+     * If a {@link Map} is supplied and if supported by the {@link Graph} then each of the key/value pairs in the map will
+     * be added as property.  This method is the long-hand version of looping through the 
+     * {@link #property(Cardinality, Object, Object, Object...)} method for each key/value pair supplied.
```
spmallette, line 2279: should this javadoc have been changed? this particular method doesn't do any looping through the `Map` does it?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -2364,14 +2373,58 @@ else if (value instanceof Traversal)
      * @since 3.0.0-incubating
      */
     public default GraphTraversal<S, E> property(final Object key, final Object value, final Object... keyValues) {
```
spmallette, line 2375: haven't these changes now created a situation where you can't do: `property('data', [k:100])` with the expectation that you would store a `Map` at "data"? I don't think we want that to happen. right?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -2364,14 +2373,58 @@ else if (value instanceof Traversal)
...
+                }
+                return this;
+            } else {
+                return this.property(null, key, value, keyValues);
+            }
+        }
+    }
+    
+    /**
+     * When a {@link Map} is supplied and if supported by the {@link Graph} then each of the key/value pairs in the map will
```
spmallette, line 2403: Not sure about the language used in each of these javadoc entries - "if supported by the Graph" isn't really a condition, is it? All graphs will generally support this given the way in which you implemented it, right?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -2364,14 +2373,58 @@ else if (value instanceof Traversal)
...
+            final Map<Object, Object> map = (Map)value;
+            for (Map.Entry<Object, Object> entry : map.entrySet()) {
+                property(null, entry.getKey(), entry.getValue());
+            }
+            return this;
+        } else {
+            throw new IllegalArgumentException("property(object) must have a Map argument");
+        }
     }
-
```
spmallette, line 2374: nit:  looks like some unintended changes here/below here in `GraphTraversal.java`

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -2364,14 +2373,58 @@ else if (value instanceof Traversal)
...
+     * <p /> 
+     * This method is effectively calls {@link #property(VertexProperty.Cardinality, Object, Object, Object...)}
+     * as {@code property(null, key, value, keyValues}.
+     *
+     * @param value     the value for the property
+     * @return the traversal with the last step modified to add a property
+     * @see <a href="http://tinkerpop.apache.org/docs/${project.version}/reference/#addproperty-step" target="_blank">AddProperty Step</a>
+     * @since 3.0.0-incubating
+     */
+    public default GraphTraversal<S, E> property(final Object value) {
```
spmallette, line 2417: I can't remember the reason for using `Object` here for the "value" argument. If this step only accepts `Map` then is there a reason why we don't just define it as `final Map<String,Object> map`? That would get rid of the `instanceof` check and related exception.

spmallette:
As a final note on this round of review, please add some Upgrade Documentation to help draw attention to this change.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/GenericLiteralVisitor.java
bechbd: Yeah, this was definitely an IDE thing

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
bechbd: Yeah, that is a good point, I've removed it.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
bechbd: I think this went to my earlier offline comment on why not provide more specific data types.  Based on that discussion I've updated this to be `Map<Object, Object> ` as we cannot use a string for the key and use the `T` enum.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
bechbd: Got a bit overzealous here, it's been reverted,

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
bechbd: Good catch, yes that is an error.  Fixed.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/TraversalMethodVisitor.java
bechbd: Fixed

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
bechbd: Yeah not sure how that happened, reverted

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/__.java
bechbd: Removed

spmallette:
Merged on #1546 1546

PR #1538, opened 2022-01-05
TINKERPOP-2651 Update to .NET 6

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2651

Note that this only affects our build process and the Gremlin.Net.Template. Users of Gremlin.Net will not be affected by this change as that still targets .NET Standard 2.0 (and additionally .NET Standard 1.3 in `3.4-dev`).

This does not necessarily have to be merged before code freeze for 3.4.13/3.5.2/3.6.0. I'm just opening the PR now as I have the contribution ready, but I can also simply retarget this on `3.5-dev` if we don't get it merged before code freeze.

I verified that the Docker build also still works with `docker/build.sh -t`.

VOTE +1

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -176,10 +176,10 @@ jobs:
         with:
           java-version: '11'
           distribution: 'temurin'
-      - name: Set up .NET 3.1.x
+      - name: Set up .NET 6.0.x
         uses: actions/setup-dotnet@v1
         with:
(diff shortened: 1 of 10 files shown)


spmallette:
`3.5-dev` is good. VOTE +1

FlorianHockmann:
Retargeted on `3.5-dev`. I will merge this after code freeze is lifted if no other reviews come in until then.

PR #1539, opened 2022-01-05
TINKERPOP-2679 update javascript driver to support stream processing

tkolanko:
This PR implements a change to javascript driver for allowing an optional callback which will be run with the result set of each chunk of data returned from the gremlin server rather than waiting for the entire query to finish.

### DIFF
--- a/docs/src/reference/gremlin-variants.asciidoc
+++ b/docs/src/reference/gremlin-variants.asciidoc
@@ -1721,6 +1721,32 @@ IMPORTANT: The preferred method for setting a per-request timeout for scripts is
 with bytecode may try `g.with(EVALUATION_TIMEOUT, 500)` within a script. Scripts with multiple traversals and multiple
 timeouts will be interpreted as a sum of all timeouts identified in the script for that request.
 
+
+==== Processing results as they are returned from the Gremlin server
+
+
+The Gremlin JavaScript driver maintains a WebSocket connection to the Gremlin server and receives messages according to the `batchSize` parameter on the per request settings or the `resultIterationBatchSize` value configured for the Gremlin server. When submitting scripts the default behavior is to wait for the entire result set to be returned from a query before allowing any processing on the result set. 
+
+The following examples assume that you have 100 vertices in your graph.
+
+[source,javascript]
+----
+const result = await client.submit("g.V()");
+console.log(result.toArray()); // 100 - all the vertices in your graph
+----
+
+When working with larger result sets it may be beneficial for memory management to process each chunk of data as it is returned from the gremlin server. The Gremlin JavaScript driver can accept an optional callback to run on each chunk of data returned.
+
+[source,javascript]
+----
+
+await client.submit("g.V()", {}, { batchSize: 25 }, (data) => {
+  console.log(data.toArray().length); // 25 - this callback will be called 4 times (100 / 25 = 4)
+})
+----
+
+If the callback is provided, each chunk of data will be processed as it comes in, rather than waiting for the entire result set to be returned.
+
 [[gremlin-javascript-dsl]]
 === Domain Specific Languages
 
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/client.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/client.js
@@ -82,16 +82,20 @@ class Client {
    * @param {Number} [requestOptions.batchSize] The size in which the result of a request is to be "batched" back to the client
    * @param {String} [requestOptions.userAgent] A custom string that specifies to the server where the request came from.
    * @param {Number} [requestOptions.evaluationTimeout] The timeout for the evaluation of the request.
+   * @param {callback} [onDataMessageHandler] A callback to handle data as it comes in from the gremlin server when submitting string scripts.
    * @returns {Promise}
    */
-  submit(message, bindings, requestOptions) {
-    const requestIdOverride = requestOptions && requestOptions.requestId
+  submit(message, bindings, requestOptions, onDataMessageHandler) {
+    const requestIdOverride = requestOptions && requestOptions.requestId;
     if (requestIdOverride) delete requestOptions['requestId'];
 
-    const args = Object.assign({
-      gremlin: message,
-      aliases: { 'g': this._options.traversalSource || 'g' }
-    }, requestOptions)
+    const args = Object.assign(
+      {
+        gremlin: message,
+        aliases: { 'g': this._options.traversalSource || 'g' },
+      },
+      requestOptions
+    );
 
     if (this._options.session && this._options.processor === 'session') {
       args['session'] = this._options.session;
@@ -99,15 +103,33 @@ class Client {
 
     if (message instanceof Bytecode) {
       if (this._options.session && this._options.processor === 'session') {
-        return this._connection.submit('session', 'bytecode', args, requestIdOverride);
+        return this._connection.submit(
+          'session',
+          'bytecode',
+          args,
+          requestIdOverride,
+          onDataMessageHandler
+        );
       } else {
-        return this._connection.submit('traversal', 'bytecode', args, requestIdOverride);
+        return this._connection.submit(
+          'traversal',
+          'bytecode',
+          args,
+          requestIdOverride,
+          onDataMessageHandler
+        );
       }
     } else if (typeof message === 'string') {
       args['bindings'] = bindings;
       args['language'] = 'gremlin-groovy';
       args['accept'] = this._connection.mimeType;
-      return this._connection.submit(this._options.processor || '','eval', args, requestIdOverride);
+      return this._connection.submit(
+        this._options.processor || '',
+        'eval',
+        args,
+        requestIdOverride,
+        onDataMessageHandler
+      );
     } else {
       throw new TypeError("message must be of type Bytecode or string");
     }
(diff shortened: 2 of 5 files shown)


spmallette:
Thanks for submitting this - it may have to wait for 3.5.3 though as we're getting ready for release. Is that ok for you?

tkolanko:
@spmallette that works just fine - thanks!

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1721,6 +1721,32 @@ IMPORTANT: The preferred method for setting a per-request timeout for scripts is
...
+const result = await client.submit("g.V()");
+console.log(result.toArray()); // 100 - all the vertices in your graph
+----
+
+When working with larger result sets it may be beneficial for memory management to process each chunk of data as it is returned from the gremlin server. The Gremlin JavaScript driver can accept an optional callback to run on each chunk of data returned.
+
+[source,javascript]
+----
+
+await client.submit("g.V()", {}, { batchSize: 25 }, (data) => {
```
jorgebay, line 1743: I think mixing promises and callbacks in the same API can be confusing and prone to misuse.

I think we should expose a different method for "streaming" or grouping into smaller sets of results, for example with async iterables:

```javascript
const stream = client.stream(traversal);
for await (const item of stream) {
  statement
}
```

or regular callbacks
```javascript
client.forEach(traversal, item => {
 // called for each item
}, err => {
  // called at the end or when there's an error
});
```

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
```diff
@@ -290,13 +296,31 @@ class Connection extends EventEmitter {
...
+          );
+        }
         this._clearHandler(response.requestId);
         return handler.callback(null, new ResultSet(utils.emptyArray, response.status.attributes));
       case responseStatusCode.partialContent:
-        handler.result = handler.result || [];
-        handler.result.push.apply(handler.result, response.result.data);
+        if (this._onDataMessageHandlers[response.requestId]) {
+          this._onDataMessageHandlers[response.requestId](
+            new ResultSet(response.result.data, response.status.attributes)
```
jorgebay, line 309: Maybe instead of having 4 resultsets, the user wants to access each individual item.

jorgebay (commented):
Good idea, I think the js api needs a way to access the results as they come.

I've commented mostly on the API below.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
tkolanko: That would mean different parsing would have to take place between partial content and final parsing/if the total results were less than the batch size.

docs/src/reference/gremlin-variants.asciidoc
jorgebay: > we don't operate on traversals

I used the variable name `traversal` for the string script containing the traversal, I edited the code example for clarity.

tkolanko: ah ok, let me take a look at how complex it would be to add an async iterator to websocket messages. Through a quick google search there are [some](https://github.com/alanshaw/it-ws) [libraries](https://github.com/alanshaw/it-ws) but they don't seem to be maintained

jorgebay: I think from the user POV, having a thing (`ResultSet`, `Stream`, ...) that represents the result of the execution is more familiar than having 1 per "batch". The batch is a low level thing.

There are several db client libraries that use this approach (blocking iterators in other languages, async iterators in node.js, streams or callbacks). I [worked on the paging api for Cassandra](https://github.com/datastax/nodejs-driver/tree/master/doc/features/paging) that uses several of these approaches (see [here for example](https://github.com/datastax/nodejs-driver#row-streaming-and-pipes)), but these are popular patterns across node.js db client libraries (see [mysql](https://github.com/mysqljs/mysql#streaming-query-rows))

I think that we should return an object that fires events or a stream of result items, we should clearly signal when all the data has been received or when there was an error. If we use callbacks, I think we shouldn't mix it with promises to get the error.

jorgebay: I think an async iterator is harder to implement in this case given that the tinkerpop protocol doesn't allow to fetch the next page (it will continue to serve the next "partial content").

I think an event emitter (emitter of result items) or a stream fit better in this case.

tkolanko: @jorgebay I implemented a new `client.stream` which returns a readable stream. Can you take a look and let me know what you think.

jorgebay: Nice!
I'll try to review it this week

spmallette: we're getting a bit closer to 3.5.3 code freeze i think. i was wondering if there was anything left to do on this PR to have it ready for merge. cc/ @jorgebay

tkolanko: Some of the existing tests don't like the changes I made to reduce duplicate code the original submit and the new stream method. I'll back out of those and leave the duplicated code for now; it's not that much. If we want to have this fixed up for 3.5.3 I can take a crack at it this weekend.

spmallette:
@tkolanko i just thought i'd follow up on this PR which seems to pass CI builds now. is this done from your perspective?

tkolanko:
> @tkolanko i just thought i'd follow up on this PR which seems to pass CI builds now. is this done from your perspective?

Yeah, good on my side if @jorgebay is good with it

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/client.js
```diff
@@ -73,41 +72,118 @@ class Client {
...
+    } else if (typeof message === "string") {
+      args["bindings"] = bindings;
+      args["language"] = "gremlin-groovy";
+      args["accept"] = this._connection.mimeType;
+      return this._connection.stream(
+        this._options.processor || "",
+        "eval",
+        args,
+        requestIdOverride
+      );
```
jorgebay, line 186: this code block is identical to the previous one, except from the underlying method that it's called.

Would it be possible to extract to something like `executeOnConnection()` or something like that, where the connection method is a parameter as well?

For example, it will have:
```javascript
return this._connection.[methodName](/* ... */)
```

jorgebay (commented):
Looks great, I've added some minor final suggestions below.

The formatting changes made reviewing a little harder but the logic is 👌

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/client.js
tkolanko: That's what I originally did in https://github.com/apache/tinkerpop/pull/1539/commits/fe1e468945ea3a36c53c2b452627bfffd20e0239 but it ended up breaking all the client tests that overwrote the internal connection object and I couldn't figure out why. I'll take another look to see if I can get the tests to pass with that

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/client.js
```diff
@@ -46,13 +45,13 @@ class Client {
    */
   constructor(url, options = {}) {
     this._options = options;
-    if (this._options.processor === 'session') {
+    if (this._options.processor === "session") {
```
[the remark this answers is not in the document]
tkolanko: Yeah, it's most likely my editor. I'll get these fixed up

Would you be willing to accept a seperate pr that enabled eslint with airbnb config + prettier and just accept it's formatting?

tkolanko:
@spmallette I think this will need to wait until after 3.5.3 so I can make the changes @jorgebay requested

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/client.js
jorgebay: It's just for maintainability but if you see that's too much work, we can move forward without extracting a common method.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/client.js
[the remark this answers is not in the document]
jorgebay: Sure, we feel free to create a JIRA ticket.
My main concern is flattening the history too much, so we can start with eslint w/ few presets, it's fine by me.

tkolanko:
@spmallette @jorgebay Here's where we are with this PR:

1. I've fixed up the formatting issues and will create a new JIRA ticket to implement a standard for linting the JS files
2. The tests are passing, however, there is a duplicated block of code that should be extracted out. The current client tests are overriding an internal object and they did not play nice with having that block of code extracted out. I need more time to figure out why that is. We can either hold off on this PR and fixed that up or we can merge this PR in and I can create a JIRA ticket to figure out a better way to remove the duplicated block of code

jorgebay (approved):
This is a huge step forward in terms of user facing API. Thanks for looking into the proposed changes.

VOTE +1

jorgebay:
I'll leave it open for a few more hours, to give the opportunity for other committers to chime in.

If there aren't any remarks by then, I'll merge it.

PR #1546, opened 2022-01-12
Added ability to specify a Map for property() step

bechbd:
This pull request extends the `property()` step to take a `Map` argument by adding two new signatures:

```
property(Map)
property(Cardinality, Map)
```
When called, each individual key/value pair in the map is saved as a property to the element.
When the cardinality is specified, that cardinality will be applied to all elements in the map as they are saved to the element.

```
gremlin> g.addV('person').property(['first_name': 'Dave', 'last_name': 'B']).valueMap()
==>[last_name:[B],first_name:[Dave]]
gremlin> g.addV('person').property(single, ['first_name': 'Dave', 'last_name': 'B']).valueMap()
==>[last_name:[B],first_name:[Dave]]
```

Known limitations
If users need different cardinalities per property, then please use the existing pattern of stringing multiple property() calls together.


spmallette:
I think this looks good now - code review was on #1533 - merging.

PR #1547, opened 2022-01-12
Add GraphBinary serialization support to gremlin-javascript

ihoro:
https://issues.apache.org/jira/browse/TINKERPOP-2229

This is the first tiny step draft, there is a lot of chore work ahead. And this is the reason I would like to request comments in order to make sure I’m going the right direction and to apply comments/suggestions at the early stages.

Shall I continue with such structure?

PR #1549, opened 2022-01-15
gremlin-javascript: Tune Writer/Reader abstraction border

ihoro:
The idea is to simply extract GraphSON related logic from connection class back to respective serializer implementation. As a result we get a generic “door” for different serializer implementations, e.g. GraphBinary.

As [suggested](https://github.com/apache/tinkerpop/pull/1547#discussion_r784821882) by @divijvaidya.


jorgebay (approved):
Extracted abstractions look great, VOTE +1

spmallette:
VOTE +1

PR #1553, opened 2022-01-21
Gremlin Value Expressions 2.0 / Ternary Boolean Logics

mikepersonick:
https://issues.apache.org/jira/browse/TINKERPOP-2687

Ready for final review and merge.

TODO (later, under a separate ticket):
- Gherkin and Parser support for NaN/Inf
- Gherkin tests for Comparability


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/OrderabilityComparator.java
```diff
@@ -167,6 +174,20 @@ private static Object transform(Object o) {
         return o;
     }
 
+    /**
+     * Return true if the two objects are of the same comparison type, even if they are not the same Class.
+     */
+    public static boolean comparable(final Object f, final Object s) {
+        if (f == null || s == null)
+            return f== s; // both in the null space
```
mikepersonick, line 182: fix

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/FilterStep.java
```diff
@@ -34,9 +36,22 @@ public FilterStep(final Traversal.Admin traversal) {
...
         while (true) {
-            final Traverser.Admin<S> traverser = this.starts.next();
-            if (this.filter(traverser))
-                return traverser;
+            try {
+                final Traverser.Admin<S> traverser = this.starts.next();
+                if (this.filter(traverser))
+                    return traverser;
+            } catch (GremlinTypeErrorException ex) {
+                if (this instanceof BinaryReductionStep || getTraversal().getParent() == EmptyStep.instance()) {
```
mikepersonick, line 44: @spmallette Is this the right way to test for top-level traversal?

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/TernaryBooleanLogicsTest.java
```diff
@@ -0,0 +1,317 @@
...
+        // P.neq
+        checkHasNext(true, g.inject(Double.NaN).is(P.neq(Double.NaN)));
+        // P.lt
+        checkHasNext(false, g.inject(Double.NaN).is(P.lt(Double.NaN)));
+        // P.lte
+        checkHasNext(false, g.inject(Double.NaN).is(P.lte(Double.NaN)));
+        // P.gt
+        checkHasNext(false, g.inject(Double.NaN).is(P.gt(Double.NaN)));
+        // P.gte
+        checkHasNext(false, g.inject(Double.NaN).is(P.gte(Double.NaN)));
```
mschmidt00, line 62: Does it make sense to also add some tests for nested predicates? Something like "1<2 == 2<3", with more interesting values such as NULL's an NaN's?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Contains.java
```diff
@@ -52,9 +52,19 @@
     within {
         @Override
         public boolean test(final Object first, final Collection second) {
-            return first instanceof Number
-                    ? IteratorUtils.anyMatch(second.iterator(), o -> Compare.eq.test(first, o))
-                    : second.contains(first);
+            GremlinTypeErrorException typeError = null;
```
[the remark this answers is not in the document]
mikepersonick: Without just calls !within

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/FilterStep.java
mikepersonick: The reference implementation doesn't really work like that. There is nothing to forward it to. This FilterStep (if it's at the top level of the query) has the solution and makes the call whether to filter it or not. It is not embedded in a target, the solution has already left the target. If an exception is thrown here it goes straight to the client, there is no where higher to catch it inside of the query execution layer. For Graph providers that push this down obviously that changes.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/CompareTest.java
```diff
@@ -95,7 +87,96 @@
...
+                {Compare.gte, new Object(), new Object(), GremlinTypeErrorException.class},
+                {Compare.gte, new Object(), new Object(), GremlinTypeErrorException.class},
+
+                /*
+                 * NaN has pretty much the same comparability behavior against any argument (including itself):
+                 * P.eq(NaN, any) = FALSE
+                 * P.neq(NaN, any) = TRUE
+                 * P.lt/lte/gt/gte(NaN, any) = ERROR -> FALSE
+                 */
+                {Compare.eq,  NaN, NaN, false},
```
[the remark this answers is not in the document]
mikepersonick: Not TBD, NaN is done for compare

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/CompareTest.java
```diff
@@ -95,7 +87,96 @@
...
+                {Compare.lte, null, 1, GremlinTypeErrorException.class},
+                {Compare.lte, 1, null, GremlinTypeErrorException.class},
+
+                {Compare.eq,  NaN, null, false},
+                {Compare.neq, NaN, null, true},
+                {Compare.gt,  NaN, null, GremlinTypeErrorException.class},
+                {Compare.gte, NaN, null, GremlinTypeErrorException.class},
+                {Compare.lt,  NaN, null, GremlinTypeErrorException.class},
+                {Compare.lte, NaN, null, GremlinTypeErrorException.class},
         }));
```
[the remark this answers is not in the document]
mikepersonick: p.not is tested in the Xor connective test below, but I will be adding significant additional test coverage so I will add some more not testing.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Compare.java
```diff
@@ -78,25 +77,19 @@ public Compare negate() {
...
-                if (bothAreComparableAndOfSameType(first, second)) {
-                    return ((Comparable) first).compareTo(second) > 0;
-                }
-                throwException(first, second);
-            }
-
-            return false;
+            // NaN takes precedence
+            if (eitherAreNaN(first, second))
+                throwTypeError();
```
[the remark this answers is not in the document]
mschmidt00: ack

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/CompareTest.java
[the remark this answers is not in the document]
mschmidt00: ack, having a look now

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Text.java
```diff
@@ -37,6 +37,8 @@
     startingWith {
         @Override
         public boolean test(final String value, final String prefix) {
```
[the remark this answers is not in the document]
mikepersonick: Based on what I'm seeing that must be a parse error. @spmallette?

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/ContainsTest.java
```diff
@@ -47,7 +52,12 @@
                 {Contains.within, 100, Arrays.asList(1, 2, 3, 4, 10), false},
                 {Contains.without, 10L, Arrays.asList(1, 2, 3, 4, 10), false},
                 {Contains.within, "test", Arrays.asList(1, 2, 3, "test", 10), true},
-                {Contains.without, "testing", Arrays.asList(1, 2, 3, "test", 10), true}
+                {Contains.without, "testing", Arrays.asList(1, 2, 3, "test", 10), true},
+
+                {Contains.within, Double.NaN, Arrays.asList(Double.NaN), false},
+                {Contains.within, Double.NaN, Arrays.asList(0, Double.NaN), false},
+                {Contains.without, Double.NaN, Arrays.asList(Double.NaN), true},
+                {Contains.without, Double.NaN, Arrays.asList(0, Double.NaN), true},
```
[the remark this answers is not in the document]
mikepersonick: I'm not sure we can - Contains uses P.eq/neq equality semantics. No ERRORs

divijvaidya (commented):
~~You will probably need changes in `IsStep` as well to catch the exception and change to `false` e.g. 
3.5.x behaviour~~

Never mind. Found the changes made to `FilterStep` which should cover that.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/AndStep.java
```diff
@@ -33,10 +34,18 @@ public AndStep(final Traversal.Admin traversal, final Traversal<S, ?>... travers
...
     protected boolean filter(final Traverser.Admin<S> traverser) {
+        GremlinTypeErrorException typeError = null;
         for (final Traversal.Admin<S, ?> traversal : this.traversals) {
-            if (!TraversalUtil.test(traverser, traversal))
-                return false;
+            try {
+                if (!TraversalUtil.test(traverser, traversal))
+                    return false;
+            } catch (GremlinTypeErrorException ex) {
+                // hold onto it until the end in case any other arguments evaluate to FALSE
```
divijvaidya, line 43: I did not understand this. why not short circuit here? (given that the exception is converted to false anyways and the expression will always be false irrespective of the result of other traversals.)

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/FilterStep.java
divijvaidya: I agree with what Mike is saying. When a traversal is executed, exceptions are directly propagated outside to the server and from server to the client. One option for "guaranteed-top-level" would be catching this at server and converting to a result but I would not be in favour of that approach since it leaks the query execution semantics to the server and introduce strong coupling.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/BinaryReductionStep.java
```diff
@@ -0,0 +1,29 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.tinkerpop.gremlin.process.traversal.step.filter;
+
+import org.apache.tinkerpop.gremlin.process.traversal.GremlinTypeErrorException;
+
+/**
+ * Any {@link FilterStep} marked as a BinaryReductionStep will catch {@link GremlinTypeErrorException}s produced by
```
divijvaidya, line 24: I am a bit apprehensive about maintainability of BinaryReductionStep. If a new step is added in future, it would be upto the implementors good intentions to add this interface. Instead, can we make it more explicit? 

Some possible ideas:
1. Do we need this interface? Can we say that all FilterSteps convert `GremlinTypeErrorException` into a boolean?What are the cases where they don't?
2. Why do we wait to convert this exception into false at the top level? Why not convert it in the lowest granularity step itself?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/AndStep.java
mikepersonick: Check out the table in the semantics document (in this PR). There is a description of AND/OR/XOR semantics:

`ERROR && FALSE -> FALSE`

`AND` might be wrapped by an operator that handles ERROR differently from FALSE (e.g. `NOT`), so we need to wait on propagating the ERROR until we've seen all the arguments.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/BinaryReductionStep.java
mikepersonick: 1. Child traversals need an opportunity to reduce from ternary to binary. We can't say all FilterSteps convert, because of (2) propagation.
2. We need to propagate `ERROR` up to the last minute before reducing to `FALSE`, since it can be combined in interesting ways via `AND/OR/NOT`.

It seems extremely unlikely to me there would be another step added that would need to worry about this. Any normal filter step would not. It's a quasi-top level filter like .where and .filter that might need to worry about binary reduction in the context of a child traversal.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/AndStep.java
divijvaidya: Understood. Your comment made me think about this change in the semantics itself.

My understanding is that we are changing the semantics to say "illegal comparisons are treated as false i.e. not comparable" unlike before where we used to throw an exception on illegal comparisons. I also understand that the new change in preferred since it removes "null" as a result for a comparison evaluation and beings it closer to equality/equivalence semantics.

But why do we need ternary logic? Why can't Illegal comparison simply be false (instead of error)? (please feel free to point me to a document or discussion thread if this has already been discussed before).

I am advocating for illegal comparison to be treated as `false` at predicate evaluation itself to keep things simpler for customers. It's easy to understand "illegal comparisons are treated as false" but it is difficult to navigate the granular semantics defined in the table in this PR (the one involving XOR).

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/BinaryReductionStep.java
divijvaidya: Understood. Resolving this thread.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Text.java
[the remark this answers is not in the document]
spmallette: yes - i would expect a parser error at most here.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/FilterStep.java
spmallette: checking for a top-level traversal is generally done with `Traversal.isRoot()`. other than that i think you have this right.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/AndStep.java
mikepersonick: We are trying to bring things more in line with SPARQL, which has well defined logics for error expressions. There are certain value expressions that simply cannot be proved as true or false. There are some nice properties we get by using `ERROR` instead of `FALSE` for these. For example any numeric comparison to NaN is an `ERROR` - P.lt/lte/gt/gte - all `ERROR`, and all reduce to `FALSE` if evaluated on their own. When enclosed by a `NOT`, if we simply immediately used `FALSE` instead of `ERROR`, we would get things like `NOT(P.lt(NaN)) = TRUE`, which is logically inconsistent with `P.gte(NaN) = FALSE`.

It may just be that we need to beef up the documentation.

divijvaidya: aahh... the `NOT(P.lt(NaN)) = TRUE` is a really good example to explain this. Thank you Mike.

spmallette:
VOTE +1

divijvaidya (approved):
Looks good. Vote +1 assuming the integration test pass.

PR #1555, opened 2022-01-31
TINKERPOP-2681 mergeE() and mergeV()

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2681
https://issues.apache.org/jira/browse/TINKERPOP-2661

Please see the upgrade docs for more information on this change.  It should be noted that there isn't any reference documentation here and perhaps the recipes need some review. @krlawrence is going to have at look at those sorts of changes separately from this body of work. 

Since this change is on the heavy side I thought I'd put it out here for review despite there being some tweaks that I may notice as I review the diff.  Generally speaking though, 

VOTE +1


krlawrence:
I have built this branch and run several tests, all of which worked. I think this is a huge win for Gremlin users. The upsert pattern - whether using fold/coalese or injected maps is something most users struggle with and query engines find hard to optimize well.

VOTE +1

bechbd:
I built this branch and tried it out, and it is looking great.  So much easier to grok than the current `coalesce()` pattern

VOTE +1

divijvaidya:
The other +1 seem to be from testing the usage perspective. I am currently in midst of reviewing this PR from code perspective. Please wait for my feedback before merging.

docs/src/upgrade/release-3.6.x.asciidoc
```diff
@@ -29,6 +29,71 @@ Please see the link:https://github.com/apache/tinkerpop/blob/3.6.0/CHANGELOG.asc
...
+......2>   elementMap()
+==>[id:2,label:person,name:vadas,age:30]
+----
+
+The `fold().coalesce(unfold(), ...)` pattern was even more complicated for upserting edges, but the following example
+demonstrates how much easier `mergeE()` is to follow:
+
+[source,text]
+----
+// prior to 3.6.0, use fold().coalesce(unfold(), ...)
```
FlorianHockmann, line 66: (nitpick) This comment says that `fold().coalesce(unfold(), ...)` is used but the Gremlin traversal doesn't use `fold()` and `unfold()`. Maybe we should really change this traversal to also use `fold()` and `unfold()`?

CHANGELOG.asciidoc
```diff
@@ -27,6 +27,8 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Changed TinkerGraph to allow identifiers to be heterogeneous when filtering.
 * Prevented values of `T` to `property()` from being `null`.
 * Added `fail()` step.
+* Added `mergeV()` and `mergeE()` steps.
+* Moved `TraversalOptionParent.Pick` to it's own class as `Pick`.
```
FlorianHockmann, line 31:
```suggestion
* Moved `TraversalOptionParent.Pick` to its own class as `Pick`.
```

FlorianHockmann (commented):
I only performed a very high-level review, mostly just of the upgrade docs, because I wanted to get a gist of how the new steps work.

docs/src/upgrade/release-3.6.x.asciidoc
```diff
@@ -29,6 +29,71 @@ Please see the link:https://github.com/apache/tinkerpop/blob/3.6.0/CHANGELOG.asc
...
+step it should be straightforward for providers to optimize as part of their implementations. The following example
+demonstrates just how much easier implementing a basic upsert of a vertex has gotten:
+
+[source,text]
+----
+// prior to 3.6.0, use fold().coalesce(unfold(), ...)
+gremlin> g.V().has('person','name','vadas').has('age', 27).
+......1>   fold().
+......2>   coalesce(unfold().property('age',30),
+......3>            addV('person').property('name','vadas').property('age',27)).
```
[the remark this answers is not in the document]
spmallette: > I find the example here already quite advanced and not obvious to understand.

well, i think that's somewhat indicative of why these steps needed to be added. i tried to pick an easy one already....you should see a "hard" one. 😄  that said, i've added a more simple example along the lines you described.

docs/src/upgrade/release-3.6.x.asciidoc
spmallette: i fixed that up.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/reference/ReferenceVertex.java
```diff
@@ -38,6 +38,10 @@ private ReferenceVertex() {
 
     }
 
+    public ReferenceVertex(final Object id) {
```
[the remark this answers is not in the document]
spmallette: good point. a "reference" really just needs to have enough information to identify it so that it is attachable back to the graph. the label has never been required to do that and ends up being a bit superfluous for the case where the user wants to construct a reference independently of the graph as in the case:

```java
g.V(1).addE('knows').to(new ReferenceVertex(2))
```

which I believe is allowed in different language variants currently.  Historically speaking, `ReferenceVertex` grew out of a need in OLAP (where the label has more relevance) and then later became the currency of Gremlin Server network serialization. Not sure if this constructor is dangerous for its convenience. I'd be inclined to leave it

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/WsGremlinBinaryRequestDecoder.java
```diff
@@ -76,6 +76,7 @@ protected void decode(final ChannelHandlerContext channelHandlerContext, final B
             try {
                 objects.add(serializer.deserializeRequest(messageBytes.discardReadBytes()));
             } catch (SerializationException se) {
+                logger.warn(se.getMessage());
```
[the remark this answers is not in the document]
spmallette: i went in a two day hole of debugging because I couldn't figure out where that a serialization problem from here was causing trouble. just wanted some additional information in the logs here to help for that sort of problem in the future. i will change the log messages to a general message and stacktrace.

gremlin-test/features/sideEffect/Group.feature
```diff
@@ -264,17 +264,17 @@ Feature: Step - group()
...
-    And using the parameter xx1 defined as "m[{\"marko\":[\"666\"], \"noone\":[\"blah\"]}]"
-    And the traversal of
-      """
-      g.withSideEffect("a", xx1).V().group("a").by("name").by(__.outE().label().fold()).cap("a").unfold().group().by(keys).by(select(values).order(Scope.local).by(Order.asc))
-      """
-    When iterated to list
-    Then the result should be unordered
-      | result |
-      | m[{"ripple":[], "peter":["created"], "noone":["blah"], "vadas":[], "josh":["created", "created"], "lop":[], "marko":["666", "created", "knows", "knows"]}] |
+#  Scenario: g_withSideEffectXa__marko_666_noone_blahX_V_groupXaX_byXnameX_byXoutE_label_foldX_capXaX
```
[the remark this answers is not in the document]
spmallette: added further context in comments and created: https://issues.apache.org/jira/browse/TINKERPOP-2699

functionally the Gremlin works fine...the test frameworks just have a problem asserting the behavior.

docs/src/upgrade/release-3.6.x.asciidoc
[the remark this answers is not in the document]
FlorianHockmann: With not obvious to understand I actually also meant the example below with the new `mergeV()` step because of the `onMatch` option.
But the example you added above is definitely easier to understand in my opinion and first reading that example together with the description for this example should also make that example easy enough to understand.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MergeVertexStep.java
```diff
@@ -0,0 +1,293 @@
...
+        return super.processNextStart();
+    }
+
+    /**
+     * Use the {@code Map} of search criteria to most efficiently return a {@code Stream<Vertex>} of matching elements.
+     * Providers might override this method when extending this step to provide their own optimized mechanisms for
+     * matching the list of vertices. This implementation is only optimized for the {@link T#id} so any other usage
+     * will simply be in-memory filtering which could be slow.
+     */
+    protected Stream<Vertex> createSearchStream(final Map<Object,Object> search) {
```
mikepersonick, line 157: I would consider moving this out of here and maybe make it part of the Graph API - give providers a chance to optimize it. Doesn't need to be this PR though.

gremlin-dotnet/build/generate.groovy
```diff
@@ -31,6 +31,16 @@ import java.nio.file.Paths
...
 
+// getting an exception like:
+// > InvocationTargetException: javax.script.ScriptException: groovy.lang.MissingMethodException: No signature of
+// > method: org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.DefaultGraphTraversal.mergeE() is applicable for
+// > argument types: (String) values: [4ffdea36-4a0e-4681-acba-e76875d1b25b]
+// usually means bindings are not being extracted properly by VarAsBindingASTTransformation which typically happens
+// when a step is taking an argument that cannot properly resolve to the type required by the step itself. there are
+// special cases in that VarAsBindingASTTransformation class which might need to be adjusted. Editing the
+// GremlinGroovyScriptEngineTest#shouldProduceBindingsForVars() with the failing step and argument can typically make
+// this issue relatively easy to debug and enforce.
```
mikepersonick, line 42: Why not write a test for this in [GremlinGroovyScriptEngineTest](https://github.com/mikepersonick/tinkerpop/blob/TINKERPOP-2695/gremlin-groovy/src/test/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GremlinGroovyScriptEngineTest.java#L516)

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/process/traversal/step/map/TinkerMergeEdgeStep.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public TinkerMergeEdgeStep(final MergeEdgeStep step) {
+        super(step.getTraversal(), step.isStart(), step.getSearchCreateTraversal());
+        if (step.getOnMatchTraversal() != null) this.addChildOption(Merge.onMatch, step.getOnMatchTraversal());
+        if (step.getOnCreateTraversal() != null) this.addChildOption(Merge.onCreate, step.getOnCreateTraversal());
+        if (step.getCallbackRegistry() != null) this.callbackRegistry = step.getCallbackRegistry();
+    }
+
+    @Override
+    protected Stream<Edge> createSearchStream(final Map<Object, Object> search) {
```
mikepersonick, line 51: Same as above, perhaps this search stream business should be part of the Graph API with a default impl.

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/process/traversal/step/map/TinkerMergeVertexStep.java
```diff
@@ -0,0 +1,88 @@
...
+public class TinkerMergeVertexStep<S> extends MergeVertexStep<S> {
+    public TinkerMergeVertexStep(final MergeVertexStep step) {
+        super(step.getTraversal(), step.isStart(), step.getSearchCreateTraversal());
+        if (step.getOnMatchTraversal() != null) this.addChildOption(Merge.onMatch, step.getOnMatchTraversal());
+        if (step.getOnCreateTraversal() != null) this.addChildOption(Merge.onCreate, step.getOnCreateTraversal());
+        if (step.getCallbackRegistry() != null) this.callbackRegistry = step.getCallbackRegistry();
+    }
+
+    @Override
+    protected Stream<Vertex> createSearchStream(final Map<Object, Object> search) {
```
mikepersonick, line 49: -> Graph API

gremlin-dotnet/build/generate.groovy
mikepersonick: Looks like there is something there. Is it this?

mikepersonick (commented):
Overall looks quite nice. I vote +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MergeEdgeStep.java
```diff
@@ -0,0 +1,340 @@
...
+        return super.processNextStart();
+    }
+
+    /**
+     * Use the {@code Map} of search criteria to most efficiently return a {@code Stream<Edge>} of matching elements.
+     * Providers might override this method when extending this step to provide their own optimized mechanisms for
+     * matching the list of edges. This implementation is only optimized for the {@link T#id} so any other usage
+     * will simply be in-memory filtering which could be slow.
+     */
+    protected Stream<Edge> createSearchStream(final Map<Object,Object> search) {
```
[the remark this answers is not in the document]
spmallette: Might not be a bad idea, though a lot of providers are just extending steps or developing their own implementations wholly. My impression is that the Graph API has become a bit less used in favor of customizations by way of strategies as I did for TinkerGraph. I'd be ok leaving that as an open issue for a future change.

gremlin-dotnet/build/generate.groovy
spmallette: yeah there is a test for it there...i just wanted a comment/reminder here in these scripts because i sometimes run afoul of this problem and then forget the reason. others would probably be completely lost without this pointer.

gremlin-test/features/map/MergeEdge.feature
```diff
@@ -0,0 +1,201 @@
...
+      g.addV("person").property("name", "marko").property("age", 29)
+      """
+    And using the parameter xx1 defined as "m[{\"t[label]\": \"self\", \"weight\":\"d[0.5].d\"}]"
+    And the traversal of
+      """
+      g.V().mergeE(xx1)
+      """
+    When iterated to list
+    Then the result should have a count of 1
+    And the graph should return 1 for count of "g.E().hasLabel(\"self\").has(\"weight\",0.5)"
```
divijvaidya, line 34: please also assert that it is connected to the correct vertex here (marko).

gremlin-test/features/map/MergeEdge.feature
```diff
@@ -0,0 +1,201 @@
...
+      """
+    And using the parameter xx1 defined as "m[{\"t[label]\": \"knows\", \"D[OUT]\":\"v[100]\", \"D[IN]\":\"v[101]\", \"weight\":\"d[0.5].d\"}]"
+    And the traversal of
+      """
+      g.mergeE(xx1)
+      """
+    When iterated to list
+    Then the result should have a count of 1
+    And the graph should return 1 for count of "g.V().has(\"person\",\"name\",\"marko\").outE(\"knows\").has(\"weight\",0.5).inV().has(\"person\",\"name\",\"vadas\")"
+    And the graph should return 2 for count of "g.V().has(\"person\",\"name\",\"marko\").outE(\"knows\").inV().has(\"person\",\"name\",\"vadas\")"
```
divijvaidya, line 70: nit

outE().inV() could be replaced with out()

gremlin-test/features/map/MergeEdge.feature
```diff
@@ -0,0 +1,201 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+@StepClassMap @StepMergeE
+Feature: Step - mergeE()
```
divijvaidya, line 19: Please add the following test scenarios (similar cases for mergeV as well):
1. Uses g.mergeE() but does not specify OUT in the map
2. Uses g.mergeE() but does not specify IN in the map 
3. Uses g.mergeE() but does not specify either OUT or IN in the map 
4. Uses g.mergeE() but does not specify LABEL in the map 
5. Uses g.V(1).mergeE() but specifies OUT in the map which is different than 1
6. Wrong syntax for the map (this case could go in grammar test)
7. lower case OUT/IN in the map (this case could go in grammar test) 
8. test for mergeE(traversal) --> (I am quite curious to see the use case for this scenario)
9. mergeE() at the beginning of a child traversal
10. mergeE() in the middle a child traversal 
11. mergeV() with insertion of property having different cardinality (list/set/single)
12. mergeV() when existing vertex has single cardinality but new vertex wants to add same property with set cardinality

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MergeVertexStep.java
```diff
@@ -0,0 +1,293 @@
...
+        stream = stream.map(v -> {
+            // if no onMatch is defined then there is no update - return the vertex unchanged
+            if (null == onMatchTraversal) return v;
+
+            // assume good input from GraphTraversal - folks might drop in a T here even though it is immutable
+            final Map<String, Object> onMatchMap = TraversalUtil.apply(traverser, onMatchTraversal);
+            onMatchMap.forEach((key, value) -> {
+                // trigger callbacks for eventing - in this case, it's a VertexPropertyChangedEvent. if there's no
+                // registry/callbacks then just set the property
+                if (this.callbackRegistry != null && !callbackRegistry.getCallbacks().isEmpty()) {
```
divijvaidya, line 198: I missed any tests in this PR which test the event stream for these cases

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MergeVertexStep.java
```diff
@@ -0,0 +1,293 @@
...
+                    final EventStrategy eventStrategy = getTraversal().getStrategies().getStrategy(EventStrategy.class).get();
+                    final Property<?> p = v.property(key);
+                    final Property<Object> oldValue = p.isPresent() ? eventStrategy.detach(v.property(key)) : null;
+                    final Event.VertexPropertyChangedEvent vpce = new Event.VertexPropertyChangedEvent(eventStrategy.detach(v), oldValue, value);
+                    this.callbackRegistry.getCallbacks().forEach(c -> c.accept(vpce));
+                }
+                v.property(key, value);
```
divijvaidya, line 205: This line assumes that the cardinality will be the default one which is incorrect.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MergeEdgeStep.java
```diff
@@ -0,0 +1,339 @@
...
+        } else {
+            stream = IteratorUtils.stream(graph.edges());
+            directionUsedInLookup = Optional.empty();
+        }
+
+        // in-memory filter is not going to be nice here. it will be up to graphs to optimize this step as they do
+        // for other Mutation steps
+        stream = stream.filter(e -> {
+            // try to match on all search criteria skipping T.id as it was handled above
+            return search.entrySet().stream().filter(
```
divijvaidya, line 185: nit

This lambda could be extracted into a `Predicate`

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MergeVertexStep.java
spmallette: tests here: https://github.com/apache/tinkerpop/pull/1555/files#diff-7f146792ffd876947ef179432d27ec9a862e43d81728898b632c363e3ab3069c

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MergeVertexStep.java
spmallette: ok - we can try to detect it with `getCardinality()` off of graph features.

gremlin-test/features/map/MergeEdge.feature
spmallette: thanks for all the test ideas. this was an area i needed to return to. adding a test index as comments to each test file to make it easier to understand what is being tested as well

spmallette:
not sure who else is looking at this but you may want to pump the brakes a bit. i found some holes in `mergeE()` while writing more tests and kelvin, as he was writing documentation, suggested a small refinement to me privately to allow the `Direction` options to take an `T.id` in addition to a `Vertex` which should streamline the syntax a bit. he also suggested a slightly larger adjustment that would allow for `from` and `to` usage rather than `OUT` and `IN` which is sort of appealing as it matches the Gremlin edge creation syntax. still thinking about the impact of that one, but I like how it sounds. I will post back when i have some more code pushed for all this as well as the existing feedback already on this PR. Thanks for all the attention this has been given.

docs/src/dev/io/graphson.asciidoc
```diff
@@ -2819,6 +2779,16 @@ The following `Bytecode` example represents the traversal of `g.V().hasLabel('pe
 }
 ----
 
+==== Merge
```
mikepersonick, line 2782: Why no onCreate too?

spmallette: we only show one example for most of these. it's (hopefully) understood that any enum value of `Merge` would follow this pattern.

tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/TinkerGraphWorld.java
```diff
@@ -106,6 +106,7 @@ public GraphTraversalSource getGraphTraversalSource(final LoadGraphWith.GraphDat
 
         private static final List<String> TAGS_TO_IGNORE = Arrays.asList(
                 "@StepDrop",
+                "@StepInject",
```
mikepersonick, line 109: Why? Won't this cause a lot of tests (i.e. Comparability/Orderability) to get ignored?

spmallette: Those are for OLAP tests and OLAP doesn't support `inject()` (among other things)

spmallette:
I think I've gotten this solidified now. Hopefully there is no more significant change. Thanks for the early reviews. @krlawrence is still planning to follow up with reference documentation and will arrive on a later PR in time for release. Note the addition of the new Gremlin step semantic documentation - hopefully it establishes a nice pattern for future expansion to cover the other steps.

PR #1557, opened 2022-02-01
Inf/NaN Support for Parser/Gherkin and Comparability Feature Tests

mikepersonick:
https://issues.apache.org/jira/browse/TINKERPOP-2695


mikepersonick:
@spmallette If you wouldn't mind reviewing this one - after looking at mergeV/mergeE it seems possible that I missed some related changes necessary for the parser changes I made.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/translator/GroovyTranslator.java
```diff
@@ -181,24 +182,30 @@ protected String getSyntax(final TraversalOptionParent.Pick o) {
 
         @Override
         protected String getSyntax(final Number o) {
```
spmallette, line 184: if you go further down in this file you will find `LanguageTypeTranslator`  which is a `TypeTranslator` for `gremlin-language`. It's an extension of this one. I would think you need some changes there too so that we can generate inf/nan that matches the grammar.

spmallette:
> after looking at mergeV/mergeE it seems possible that I missed some related changes necessary for the parser changes I made.

it's getting harder to remember all the places where changes matter. nothing is standing out to me as glaringly missing. i would like if we didn't have to use a binding for "Infinity" in the scriptengine but we don't really have anything else like this and it seems like the easiest way to do it.

spmallette:
VOTE +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/translator/GroovyTranslator.java
```diff
@@ -419,24 +426,30 @@ protected Script produceScript(final Vertex o) {
...
             if (o instanceof Long)
                 return o + "L";
-            else if (o instanceof Double)
-                return o + "D";
-            else if (o instanceof Float)
-                return o + "F";
-            else if (o instanceof Integer)
+            if (o instanceof Integer)
                 return o + "I";
-            else if (o instanceof Byte)
```
krlawrence, line 430: Just curious why the else's are removed - can "o" be more than one thing?

mikepersonick: I was seeing a mix and match of cases with and without else so I just got rid of all of them. In this case it's fine since every case is a return statement. This would be a lot cleaner in a more modern version of Java where we could use a switch.

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/P.cs
```diff
@@ -85,79 +85,79 @@ public P Or(P otherPredicate)
...
         {
-            var value = args.Length == 1 ? args[0] : args;
+            var value = args != null && args.Length == 1 ? args[0] : args;
             return new P("eq", value);
         }
 
         public static P Gt(params object[] args)
         {
-            var value = args.Length == 1 ? args[0] : args;
+            var value = args != null && args.Length == 1 ? args[0] : args;
```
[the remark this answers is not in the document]
FlorianHockmann: Hey Mike, sorry if this came off the wrong way! I didn't have much time on Wednesday to review this and therefore really only checked the .NET changes. So I didn't see all the work you put into comparability semantics, including handling of `null`.

I just read through (most) of the added docs where you describe the improved semantics and checked the added feature tests. Now I understand why it makes sense to allow `null` here and I definitely welcome these changes! It's good to have a clear definition of how we handle infinity and `null`.
This will also help with TINKERPOP-2348 where we want to clearly communicate to users of Gremlin.NET where `null` is allowed as an argument or a possible return value and where not. When I initially worked on that, I ran into the problem that I wasn't sure in a lot of places whether we actually want to accept `null` there or not. So, any clarifications we make on how we handle `null` in various places are definitely welcome 😄

FlorianHockmann (approved):
VOTE +1

PR #1558, opened 2022-02-01
Updated Roadmap for 3.6 and 3.7

bechbd:
Updated roadmap to move string manipulation functions and case-insensitive search to 3.7


divijvaidya:
Seems like we are nearing the release for 3.6.x and these features not close to implementation so I agree with this change. 

Will wait for Stephen or Kelvin to +1 this though.

spmallette:
@bechbd this is fine, but could you please clean up the git history? there is a merge commit in here now. please prefer `git rebase` over `git merge` when updating your branches to keep the history straight.

spmallette:
cherrypicked as d1a14720bcbd951b58feacfc497e0a3c59daaea7 to keep the git history clean

PR #1559, opened 2022-02-02
TINKERPOP-2682 Enable WS compression in .NET

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2695

WebSocket support was only added in .NET 6 which made it necessary to explicitly add .NET 6 as an additional target framework. Compression will now be enabled by default on .NET 6 as it's only available there. This made it necessary to add conditional compilation based on the target framework.

### Package validation

I also enabled [package validation](https://docs.microsoft.com/en-us/dotnet/fundamentals/package-validation/overview) so we a) notice if we introduce breaking changes by accident and b) ensure that our code compiled against .NET Standard 2.0 can also be run against .NET 6. This already helped me to keep the breaking change here to a minimum. It's just an added optional parameter here to be able to disable compression. We could refactor the `GremlinClient` constructors in the future in a way to avoid this problem of having to add optional parameters for new config options which is always a breaking change. The validation is executed on `dotnet pack` which gets already executed via `mvn verify` and is therefore also part of our GH actions.

Compatibility errors that we know about and that we accept should be added to the `CompatibilitySuppressions.xml` file which  can also be generated via

```
dotnet pack Gremlin.Net.csproj /p:GenerateCompatibilitySuppressionFile=true
```

### Possibly vulnerable to attacks (CRIME / BREACH)

WebSocket compression can be problematic due to attacks like CRIME/BREACH as noted in TINKERPOP-2682. So, I added an option to easily disable it for applications that might be vulnerable and added docs to make users aware of this.

I guess we should discuss how we want to deal with this in general as it's not specific to .NET. So, should we simply add similar notes to the Java/Python docs?

### Benchmarks

Lastly, some quick benchmarks I ran on my machine:

**GraphBinary:**

Without compression:
Packets transmitted: 3594
Bytes transmitted: 7MB
Runtime: 1min 1.5sec

With compression:
Packets transmitted: 62 (1.7%)
Bytes transmitted: 34KB (0.5%)
Runtime: 1.3sec

**GraphSON 3:**

Without compression:
Packets transmitted: 12029
Bytes transmitted: 25MB
Runtime: 1min 40sec

With compression:
Packets transmitted: 116 (1.0%)
Bytes transmitted: 103KB (0.4%)
Runtime: 1min 39sec

So, both for GraphSON as well as for GraphBinary the network traffic shrinks by a factor of ~100 while the total runtime is not affected. Note that this was just a very simple benchmark (I used the same traversal as @spmallette used in the PRs for Python and Java, just with a smaller number of iterations to keep the runtime to a minimum). The results will of course vary for different traversals / graphs.

VOTE +1

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Gremlin.Net.csproj
+++ b/gremlin-dotnet/src/Gremlin.Net/Gremlin.Net.csproj
@@ -18,7 +18,7 @@ limitations under the License.
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup Label="Build">
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <LangVersion>8</LangVersion>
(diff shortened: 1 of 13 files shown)


gremlin-dotnet/src/Gremlin.Net/Gremlin.Net.csproj
```diff
@@ -67,14 +67,19 @@ NOTE that versions suffixed with "-rc" are considered release candidates (i.e. p
     <RepositoryUrl>https://github.com/apache/tinkerpop</RepositoryUrl>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>    
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+    <EnablePackageValidation>true</EnablePackageValidation>
+    <PackageValidationBaselineVersion>3.5.2</PackageValidationBaselineVersion>
```
FlorianHockmann, line 71: When merging this to `master`, we'll have to see whether we already have additional breaking changes there. Then we can decide whether we want to change the baseline version there to `3.6.0` (as soon as that's released) to only check for breaking changes within the 3.6 line or keep it as is to check for breaking changes also between 3.6 and 3.5.

spmallette:
> WebSocket compression can be problematic due to attacks like CRIME/BREACH....I guess we should discuss how we want to deal with this in general as it's not specific to .NET. So, should we simply add similar notes to the Java/Python docs?

I wasn't aware of this issue. Maybe you should create an issue for this in JIRA so that it's not forgotten? I don't think compression is configurable in java at this point. cc/ @divijvaidya  

Thanks for doing some benchmarks. GraphSON network traffic size differences always look so crazy to me whenever we flip compression on. 😄 

VOTE +1

jorgebay (approved):
lgtm, VOTE +1

FlorianHockmann:
> I wasn't aware of this issue. Maybe you should create an issue for this in JIRA so that it's not forgotten?

Done: [TINKERPOP-2700](https://issues.apache.org/jira/browse/TINKERPOP-2700)

PR #1560, opened 2022-02-03
docs: Fix GraphBinary enum like types specification

ihoro:
Enum like types (Barrier, Cardinality, Column, Direction, Operator, Order, Pick, Pop, Scope, T) should provide enum element name as a fully-qualified string, i.e. as `{type_code}{type_info}{value_flag}{value}`.


jorgebay:
VOTE +1

spmallette:
thanks for finding these little issues in our docs @ihoro VOTE +1

PR #1564, opened 2022-02-11
TINKERPOP-2518 Enable more Gherkin scenarios

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2518

There were three categories of scenarios ignored that are now enabled:
* Some scenarios passed without any change. I don't know why they were ignored in the first place. Maybe we've fixed something in the meantime but forgot to enable them again.
* Some scenarios expected a decimal type as the result but the server sent a double. I simply changed the expected type there to reflect what we get back from the server.
* One scenario required a more complex assertion as it expects an array of dictionaries which means that we have to perform a deep equality check.

The only kind of scenario(s) (we only have one of them on `3.5-dev`) that is still ignores uses `null` as a key in a map which is simply not supported in .NET. So, that's a feature that just cannot support in .NET.

I'll create a different PR for `master` as we have additional scenarios there that are not on `3.5-dev`.

VOTE +1


spmallette:
excellent - thanks! VOTE +1

PR #1565, opened 2022-02-11
TINKERPOP-2518 Enable more Gherkin scenarios (master)

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2518

This includes the changes from #1564 which targets `3.5-dev` and adds a commit to fix scenarios that are only on `master`.

This uncovered a bug in the `HasId step that threw a `NullReferenceException` if the second argument was `null`. We probably have the same problem with other steps that are just not tested yet for `null`. I however haven't touched them here as I'm not sure yet where we actually want to allow `null`.

Ignoring the tag `AllowNullPropertyValues` automatically in .NET simply didn't work as we checked for the tag on the feature, but it's on the scenario.

VOTE +1


gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs
```diff
@@ -772,8 +772,16 @@ private GraphTraversal(ICollection<ITraversalStrategy> traversalStrategies, Byte
         /// </summary>
         public GraphTraversal<S, E> HasId (object id, params object[] otherIds)
         {
-            var args = new List<object>(1 + otherIds.Length) {id};
```
FlorianHockmann, line 775: @mikepersonick This change is basically the same you did in #1557 for the `P` predicates. We have other steps here that also take a `params` array argument and then access its `.Length` without checking whether it's `null` first (e.g., the `HasKey` step below).
As I mentioned in the PR description, I'm not sure whether we want to allow `null` everywhere or maybe throw an `ArgumentNullException` in some places where we're sure that `null` isn't a valid argument.
Any thoughts on that?

Otherwise, I'd suggest that we create a ticket for this issue and then address it in a follow-up PR where we can concentrate on just that topic.

FlorianHockmann:
Looks like [the scenarios where I changed the expected result type from `decimal` to `double`](https://github.com/apache/tinkerpop/pull/1564/commits/599d2eeb7bb1c8d88cb4561e494e9dab75feb8da#diff-5722803f3bf83fc6999422a5d9a46d3c40799f86b1566216b30a24fb9c3a3ce4) are now failing here for TinkerGraph, although they passed for `3.5-dev`. I'm not sure yet why they passed for `3.5-dev`, but not for `master` here. Will have to look into that.

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs
mikepersonick: Florian, thanks the review and merge on the other PR. I’m out today so I’ll take a closer look on Monday, but I guess my general feeling is that whatever the semantics are on `null` (or anything else), it’s probably best to handle them in the query execution layer rather than the language bindings.

FlorianHockmann:
> Looks like [the scenarios where I changed the expected result type from decimal to double](https://github.com/apache/tinkerpop/pull/1564/commits/599d2eeb7bb1c8d88cb4561e494e9dab75feb8da#diff-5722803f3bf83fc6999422a5d9a46d3c40799f86b1566216b30a24fb9c3a3ce4) are now failing here for TinkerGraph, although they passed for 3.5-dev. I'm not sure yet why they passed for 3.5-dev, but not for master here. Will have to look into that.

I guess I shouldn't have changed these scenarios at all as we're using `BigDecimal` per default there in Java and not `double` so `decimal`/`BigDecimal` should really be the expected type. So, I've updated #1564 to instead allow `double` values in the Gherkin .NET tests when `decimal` is expectedand also updated this branch afterwards.
Hope this fixes the problem also for `master` here.

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs
FlorianHockmann: Thanks for your input. I've created an issue for that: https://issues.apache.org/jira/browse/TINKERPOP-2705
because it warrants its own issue & PR in my opinion so we can concentrate only on that topic.

FlorianHockmann:
@mikepersonick Do you still like to review this one? Otherwise, I'll merge this and #1564.

PR #1570, opened 2022-02-17
Fixed issue with property(Map) being passed a null

bechbd:
Fixed issue where property(Map) was throwing an NPE when passed a null.  This now behaves as a no-op and returns the input values


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -2473,6 +2473,8 @@ else if (value instanceof Traversal)
                     property(key, entry.getKey(), entry.getValue());
                 }
                 return this;
+            } else if (value == null) { // Just return the input if you pass a null
+                return this;
```
spmallette, line 2477: just a side effect to consider here...by returning `this` we don't modify the traversal/bytecode which means that if you were to do a `profile()`, `toString()`, use a `Translator`, etc. you wouldn't get and exact representation of the Gremlin that you executed. i've generally tried to avoid this problem in other cases with `null` arguments, but have questioned how important it is at times. thoughts?

mikepersonick: What's the alternative?

spmallette: I guess you would avoid folding to an existing `AddPropertyStep` and include a "no-op" form for a new instance of it. So `AddPropertyStep` would need some changes to do this.

bechbd: Since it performs a no-op here, I am not sure, it really matters too much that it doesn't get translated.  However, I think the bigger question is if we expect the Translators to provide an exact translation of the query or if it can translate to an "optimized" version of the query.  

I suspect that even without the `null` case, a general Map would result in multiple `property()` steps being chained together instead of a single step.

Thoughts on the need to provide a verbatim translation or not?

spmallette: I think it's ok for now to overlook this concern, but in the long run it should probably be addressed somehow. Created: https://issues.apache.org/jira/browse/TINKERPOP-2714 to track.

spmallette:
@bechbd looks like you need to rebase this - but other than that VOTE +1

PR #1571, opened 2022-02-18
TINKERPOP-2707 Close of parent connection also calls close on spawned child sessions.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2707

Not closing these sessions could end in a leak if a user opts to get a little loose with the expected API calls for properly handling sessions. This change should prevent accidents.

VOTE +1

### DIFF
--- a/gremlin-python/src/main/python/gremlin_python/driver/driver_remote_connection.py
+++ b/gremlin-python/src/main/python/gremlin_python/driver/driver_remote_connection.py
@@ -54,6 +54,11 @@ class DriverRemoteConnection(RemoteConnection):
         self.__headers = headers
         self.__session = session
         self.__transport_kwargs = transport_kwargs
+
+        # keeps a list of sessions that have been spawned from this DriverRemoteConnection
+        # so that they can be closed if this parent session is closed.
+        self.__spawned_sessions = []
+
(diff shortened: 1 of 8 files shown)


gremlin-python/src/main/python/gremlin_python/driver/driver_remote_connection.py
```diff
@@ -74,7 +79,19 @@ def __init__(self, url, traversal_source="g", protocol_factory=None,
...
+        # close this client and any DriverRemoteConnection instances spawned from this one
+        # for a session
+        if len(self.__spawned_sessions) > 0:
+            logging.info("closing spawned sessions from DriverRemoteConnection with url '%s'", str(self._url))
+            for spawned_session in self.__spawned_sessions:
+                spawned_session.close()
+
+        if self.__session:
+            logging.info("closing DriverRemoteConnection with url '%s' with session '%s'",
+                         str(self._url), self.__session)
```
lyndonb-bq, line 91: Does this print correctly? I thought the session object in the DriverRemoteConnection class was technical a uuid object, I am wondering if self.__session should be wrapped in an `str` like so:
```suggestion
                         str(self._url), str(self.__session))
```

PR #1572, opened 2022-02-19
TINKERPOP-2467 Follow python naming conventions for Gremlin syntax

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2467

Retained the old camel case and just deprecated it. Added snake case along side it. PythonTranslator still generates the old style for now which is helpful because it continues to test the deprecation working all the way through to the new functions.

Don't think there is a huge rush to drop the deprecated steps. We can move the translator for the next major release and that will set us up to remove the deprecated steps at our leisure in the major following that (or later if desired). The main point of this PR is to at least introduce the consistently named steps so that they can begin to see usage. 

Builds with `mvn clean install -pl gremlin-python -DincludeeNeo4j`


mikepersonick:
lgtm +1

divijvaidya:
Every time we change syntax significantly, we add additional overhead for users to consume the new version of TinkerPop. What is the motivation of merging this to master (tracking 3.6.x)? I would probably see this as a candidate for 4.0.

Also, please document this in the developers guide "how to add a new step" section so that we can maintain consistency for steps in future.

spmallette:
I don't see additional overhead to users if the old naming is still present. The deprecation can stay indefinitely if we desire.

divijvaidya:
> I don't see additional overhead to users if the old naming is still present. The deprecation can stay indefinitely if we desire.

My bad. I read this PR as changing the syntax and not just deprecation.

PR #1573, opened 2022-02-23
TINKERPOP-2712: Ensure new properties are added before triggering the associated event.

rngcntr:
Also fix minor JavaDoc copy/paste error.

For further description, see [TINKERPOP-2712](https://issues.apache.org/jira/browse/TINKERPOP-2712)


rngcntr:
It appears the change is a little bit trickier than I initially expected. Due to the earlier modification of the graph, the events are now initialized with wrong parameters. I will have another look at it and fix the events if possible.

PR #1574, opened 2022-02-24
Service call() and element()

mikepersonick:
https://issues.apache.org/jira/browse/TINKERPOP-2680
  https://issues.apache.org/jira/browse/TINKERPOP-2713

Only thing missing is reference documentation update.


mikepersonick:
Converted back to draft pending some updates to the Service Metadata API.

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -611,3 +611,84 @@ resolve to a `Map`.
 
 See: link:https://github.com/apache/tinkerpop/tree/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MergeVertexStep.java[source],
 link:https://tinkerpop.apache.org/docs/current/reference/#mergev-step[reference]
+
+=== call()
```
spmallette, line 615: nice - thanks for adding to this section of docs. could you please move it above `mergeV/E()` to hold alphabetical order? also, i'm not sure I recall what i was thinking when i developed the style for this documentation, but i pointed the docs at `master` and `current` in the links - i would think those should be switched to the documentation system's dynamic form of x.y.z. For example:

https://tinkerpop.apache.org/docs/x.y.z/reference/#call-step

https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/service/Service.java

The documentation system will replace that "x.y.z" variable with the current version on generation. If you don't mind, please make the changes to `mergeV/E()` as well.

docs/src/upgrade/release-3.6.x.asciidoc
```diff
@@ -776,3 +776,10 @@ semantics.
 
 See: link:https://tinkerpop.apache.org/docs/3.6.0/dev/provider/#_gremlin_semantics[Gremlin Semantics]
 See: link:https://tinkerpop.apache.org/docs/3.6.0/dev/developer/#_for_committers[Developer Documentation]
+
+===== Service Call API
```
spmallette, line 780: could you please add the link to the associated JIRA? Also could you please include a section for `element()`?

docs/src/dev/provider/gremlin-semantics.asciidoc
spmallette, line 615: i know `element()` was sort of a secondary thing here, but for completeness sake, could you add a section for that here as well?

spmallette:
> Only thing missing is reference documentation update.

do you intend to add the reference documentation for `call()` and `element()` on this PR or are you trying to merge this first?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/DefaultGremlinBaseVisitor.java
```diff
@@ -115,6 +115,31 @@ protected void notImplemented(final ParseTree ctx) {
 	 * {@inheritDoc}
 	 */
 	@Override public T visitTraversalSourceSpawnMethod_io(final GremlinParser.TraversalSourceSpawnMethod_ioContext ctx) { notImplemented(ctx); return null; }
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public T visitTraversalSourceSpawnMethod_call_empty(GremlinParser.TraversalSourceSpawnMethod_call_emptyContext ctx) { notImplemented(ctx); return null; }
```
spmallette, line 122: nit: `final`

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/TraversalMethodVisitor.java
```diff
@@ -1653,6 +1653,35 @@ public Traversal visitTraversalMethod_to_Vertex(final GremlinParser.TraversalMet
         return graphTraversal.to(StructureElementVisitor.instance().visitStructureVertex(ctx.structureVertex()));
     }
 
+    @Override
+    public Traversal visitTraversalMethod_element(GremlinParser.TraversalMethod_elementContext ctx) {
```
spmallette, line 1657: nit: `final`

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/CallStep.java
```diff
@@ -0,0 +1,338 @@
...
+        this(traversal, isStart, service, params, null);
+    }
+
+    public CallStep(final Traversal.Admin traversal, final boolean isStart, final String service, final Map params,
+                    final Traversal.Admin<S, Map> mapTraversal) {
+        super(traversal);
+
+        this.isStart = isStart;
+        this.serviceName = service;
+        this.params = params == null ? new LinkedHashMap() : params;
```
spmallette, line 89: minor thing i suppose, but while digging through this class, i sorta kept confusing `params` with our typical usage of `parameters` - maybe we could rename `params` to something else that's more descriptive in relation to `call()`?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/service/Service.java
```diff
@@ -0,0 +1,287 @@
...
+     * Return the max barrier size. Default is all upstream solutions.
+     */
+    default int getMaxBarrierSize() {
+        return Integer.MAX_VALUE;
+    }
+
+    /**
+     * Execute a Start service call.
+     */
+    default CloseableIterator<R> execute(ServiceCallContext ctx, Map params) {
```
spmallette, line 164: nit: `final`

gremlin-javascript/build/generate.groovy
```diff
@@ -54,15 +54,10 @@ radishGremlinFile = new File("${projectBaseDir}/gremlin-javascript/src/main/java
...
-gremlinGroovyScriptEngine = new GremlinGroovyScriptEngine(new GroovyCustomizer() {
-    public CompilationCustomizer create() {
-        return new RepeatASTTransformationCustomizer(new AmbiguousMethodASTTransformation())
-    }
-}, new GroovyCustomizer() {
-    public CompilationCustomizer create() {
-        return new RepeatASTTransformationCustomizer(new VarAsBindingASTTransformation())
-    }
-})
+gremlinGroovyScriptEngine = new GremlinGroovyScriptEngine(
```
spmallette, line 57: just to keep these `generate.groovy` files looking as similar as possible, could you also implement this change in for .net and python please?

hadoop-gremlin/src/test/java/org/apache/tinkerpop/gremlin/hadoop/HadoopGraphFeatureIntegrateTest.java
```diff
@@ -75,6 +77,9 @@
         add(Pair.with("g_V_matchXa_followedBy_count_isXgtX10XX_b__a_0followedBy_count_isXgtX10XX_bX_count", skipReasonLength));
     }};
 
+    private static final List<String> TAGS_TO_IGNORE = Arrays.asList(
```
spmallette, line 80: do we need to define it this way? any reason to not just add this tag to the `CucumberOptions` annotation above?

spmallette:
Finished with a first pass of this - one final thing i noticed...i didn't see gherkin tests for the new `element()` step.

mikepersonick:
Rebased, comments addressed (element() gherkin tests added).

I need to talk to Kelvin about the reference doc. Best to just merge this if it's ready.

divijvaidya (commented):
Following things are missing from this review. Are you planning to pick them up as follow-up tasks?

1. Add queries in process suite integration tests for these steps.
2. Add the new types such a ServiceRegistry to the serializers. A general question I have is around how the serializers (particularily graphbinary) will recognize the new tokens added in this reviews.

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -488,6 +488,107 @@ fully demonstrative of Gremlin step semantics. It is also hard to simply read th
...
+
+* `String` - The name of the service call.
+* `Map` - A collection of static parameters relevant to the particular service call.
+* `Traversal` - A traversal used to dynamically build at query time a collection of parameters relevant to the service
+call.
+
+*Modulation:*
+
+* `with(key, value)` - Sets an additional static parameter relevant to the service call.
+* `with(key, Traversal)` - Sets an additional dynamic parameter relevant to the service call.
```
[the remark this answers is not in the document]
spmallette: If it is a child it must be anonymous or there would be an exception.

spmallette:
> Add queries in process suite integration tests for these steps.

I'd agree that `element()` should probably have at least some process suite tests. doing that for `call()` seems somewhat questionable since they are a bit TinkerGraph specific, right (I almost wonder if they need to be in the Gherkin suite at all and not just live in `tinkergraph-gremlin`? 

by 3.7.0 hopefully we can just be on gherkin and not have to duplicate tests anymore.

> Add the new types such a ServiceRegistry to the serializers. A general question I have is around how the serializers (particularily graphbinary) will recognize the new tokens added in this reviews.

I don't think we need serializers added, Maybe I'm not thinking of something but users are not sending `SeviceRegistry` instances back and forth during remote Gremlin calls. It's like JanusGraph users not working with remote instances of the management API. 

There might be some need to add these classes for Gryo to suit OLAP sorts of cases. I assume `call()` will work fine in OLAP...we should probably test to be sure and if it doesn't document it. A test would quickly demonstrate if serialization uses cases aren't happy.

docs/src/dev/provider/gremlin-semantics.asciidoc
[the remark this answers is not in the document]
mikepersonick: Shown in the table above

mikepersonick:
Just to echo what Stephen said:

> Following things are missing from this review. Are you planning to pick them up as follow-up tasks?
> 
> 1. Add queries in process suite integration tests for these steps.
> 2. Add the new types such a ServiceRegistry to the serializers. A general question I have is around how the serializers (particularily graphbinary) will recognize the new tokens added in this reviews.

1. We decided to skip process suite tests in favor of just feature tests. Especially since these tests are only testable via Tinkergraph. That's why I put them directly in the Tinkergraph package.
2. All the ServiceRegistry stuff is part of the Graph API, not part of Gremlin or the remoting API.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/service/Service.java
```diff
@@ -0,0 +1,287 @@
...
+
+import static org.apache.tinkerpop.gremlin.util.tools.CollectionFactory.asMap;
+
+/**
+ * Service call with I input type and R return type. Services can return {@link Traverser}s or raw values (which will be
+ * converted into Traversers by {@link CallStep}.
+ *
+ * @author Mike Personick (http://github.com/mikepersonick)
+ */
+public interface Service<I, R> {
```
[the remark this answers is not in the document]
mikepersonick: Yes, good idea.

PR #1577, opened 2022-03-01
Fix reference docs for aggregate without parameters

danielcweber:
...because, as it states a few lines below:

Returns:
the traversal with an appended AggregateGlobalStep


spmallette:
merged - thanks for noticing and fixing that

PR #1580, opened 2022-03-04
TINKERPOP-2705 Improve handling of null arguments in Gremlin.NET

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2705

Changes made:
* `GraphTraversal`/`GraphTraversalSource`: Ensure that no `NullReferenceException` gets thrown. Either allow `null` as an argument or throw an `ArgumentNullException` if it doesn't seem sensible and if Java also doesn't allow `null`.
* `__`: Simply forward all `null` arguments to the steps in `GraphTraversal` so they can be handled there and we don't need to duplicate the logic whether we want to allow `null` or throw.

I had to adapt the `DotNetTranslator` a bit as the logic to avoid problems where multiple overloads would be possible was to eager. This resulted in a `params object[]` argument that should be set to `null` for a scenario, being actually tested with an `object[]` which had `null` as an element. This let the scenarios pass although users would get a `NullReferenceException` since they probably won't add the unnecessary `object` cast.
Example:
`HasKey('name', null)` -> threw a `NullReferenceException`
`HasKey('name', (object) null)` -> passed

VOTE +1


mikepersonick:
+1

I am very much in favor of this as a general rule: "Simply forward all null arguments to the steps in GraphTraversal so they can be handled there and we don't need to duplicate the logic whether we want to allow null or throw." The language bindings should be "dumb" as far as semantics are concerned, null semantics included.

PR #1582, opened 2022-03-08
TINKERPOP-2716 implement eslint and prettier for code formatting

tkolanko:
@jorgebay @krlawrence 

Here is the PR which implements eslint+prettier. Note that only the prettier rules are enabled to enforce code stanards for things like single quotes vs double quotes, spaces vs tabs etc etc

I tried turning on the airbnb config but this projects current code base is highly divergent from the airbnb style guide:

```
 ✨  Summary of failing ESLint rules                          
errors  14 warnings   0 rule: yoda
errors  51 warnings   0 rule: vars-on-top
errors  39 warnings   0 rule: strict
errors   8 warnings   0 rule: spaced-comment
errors  33 warnings   0 rule: prettier/prettier
errors  35 warnings   0 rule: prefer-template
errors   5 warnings   0 rule: prefer-spread
errors   4 warnings   0 rule: prefer-object-spread
errors  23 warnings   0 rule: prefer-destructuring
errors   8 warnings   0 rule: prefer-const
errors 193 warnings   0 rule: prefer-arrow-callback
errors  22 warnings   0 rule: one-var
errors  59 warnings   0 rule: object-shorthand
errors   4 warnings   0 rule: no-void
errors  66 warnings   0 rule: no-var
errors   4 warnings   0 rule: no-useless-escape
errors   3 warnings   0 rule: no-useless-constructor
errors  89 warnings   0 rule: no-use-before-define
errors  16 warnings   0 rule: no-unused-vars
errors  48 warnings   0 rule: no-unused-expressions
errors   1 warnings   0 rule: no-unneeded-ternary
errors 253 warnings   0 rule: no-underscore-dangle
errors   1 warnings   0 rule: no-undef-init
errors 198 warnings   0 rule: no-undef
errors  23 warnings   0 rule: no-shadow
errors  13 warnings   0 rule: no-sequences
errors   3 warnings   0 rule: no-return-assign
errors   6 warnings   0 rule: no-restricted-syntax
errors   2 warnings   0 rule: no-restricted-globals
errors  17 warnings   0 rule: no-redeclare
errors   5 warnings   0 rule: no-prototype-builtins
errors  45 warnings   0 rule: no-plusplus
errors  31 warnings   0 rule: no-param-reassign
errors   7 warnings   0 rule: no-nested-ternary
errors  10 warnings   0 rule: no-multi-assign
errors   1 warnings   0 rule: no-empty
errors  11 warnings   0 rule: no-else-return
errors   0 warnings   2 rule: no-console
errors  11 warnings   0 rule: no-cond-assign
errors   1 warnings   0 rule: no-buffer-constructor
errors  18 warnings   0 rule: no-bitwise
errors  19 warnings   0 rule: new-cap
errors   8 warnings   0 rule: max-classes-per-file
errors   1 warnings   0 rule: lines-between-class-members
errors  28 warnings   0 rule: lines-around-directive
errors   1 warnings   0 rule: import/no-useless-path-segments
errors  10 warnings   0 rule: import/newline-after-import
errors   2 warnings   0 rule: import/extensions
errors   1 warnings   0 rule: getter-return
errors   0 warnings 211 rule: func-names
errors   2 warnings   0 rule: eqeqeq
errors  49 warnings   0 rule: dot-notation
errors   2 warnings   0 rule: default-case
errors   7 warnings   0 rule: consistent-return
errors  50 warnings   0 rule: class-methods-use-this
errors   6 warnings   0 rule: camelcase
errors 210 warnings   0 rule: block-scoped-var
errors  10 warnings   0 rule: arrow-body-style
🔥  2000 problems in total (1787 errors, 213 warnings)
```

I'm not sure the best way to deal with the errors:
1. We could enable airbnb then fix all errors
2. We could enable airbnb but turn off all failing rules, then over time enable a rule and fix all errors related to the rule
3. We could enable airbnb, turn off all failing rules and leave things as is

Some of the fixes will involve quite a bit, for example `no-underscore-dangle` will also require changes to the tests since some tests override internal objects prefixed with a _.

Let me know if you want to continue forward with just code stlying as a PR, it may be better to do everything in one shot but that is going to take a large effort


jorgebay:
Thanks for taking care of this! 

I like prettier but given that is a codebase that's been around for sometime, is hard to implement a strict set of rules.

I would remove prettier rules and include [eslint:recommended](https://eslint.org/docs/rules/), with quotes: single and a few more that you feel that would help to unify the codebase. fwiw I've used these ones in the past on an old codebase: https://github.com/datastax/nodejs-driver/blob/master/.eslintrc.js

Another thing to have in mind would be to only include rules which errors can automatically be fixed (no manual changes).

gremlin-javascript/src/main/javascript/gremlin-javascript/.eslintrc.js
```diff
@@ -0,0 +1,151 @@
...
+  },
+  parserOptions: {
+    ecmaVersion: 2017,
+  },
+  ignorePatterns: ["test/**/*.js", "doc/**/*.js"],
+  extends: "eslint:recommended",
+  rules: {
+    indent: ["error", 2, { SwitchCase: 1 }],
+    "linebreak-style": ["error", "unix"],
+    quotes: ["off", "single"],
```
jorgebay, line 35: I think this should be `["error", "single"]

jorgebay:
I'll merge it tomorrow, if there isn't further feedback.

tkolanko:
@jorgebay Should we add `npm run lint` to the CI checks as well?

jorgebay:
Ideally yes, but it's not a blocker.

gremlin-javascript/pom.xml
```diff
@@ -236,6 +236,16 @@ file.write(file.getText("UTF-8").replaceFirst(/"version": "(.*)",/, "\"version\"
                             <failOnError>true</failOnError>
                         </configuration>
                     </execution>
+                    <execution>
+                      <id>npm lint</id>
+                      <phase>lint</phase>
```
jorgebay, line 241: I don't remember much about maven but I think this needs to be `<phase>integration-test</phase>` to be able to run.

[I don't see it on the build](https://github.com/apache/tinkerpop/runs/5646493848?check_suite_focus=true).

@spmallette can you help us with maven foo?

tkolanko: I missed the arguments section when I copied the block over, I pushed up a new commit and I think I got it

tkolanko: that worked but it turns out eslint dropped support for v10 node after version 7. I downgraded locally and everything seems fine, a new check should be running soon

FlorianHockmann: > you would want to probably tie this to the "verify" phase which would happen after integration tests

Wouldn't it make more sense to let a linter run before executing costly integration tests? If my code isn't formatted correctly, I wouldn't want to wait after all integration tests were executed to get that information, only so I can change the formatting and then having to execute all integration tests again to see whether the linter is now happy.
How about tying this to the `validate` phase instead which happens first and [the docs](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#a-build-lifecycle-is-made-up-of-phases) describe it as:

> validate the project is correct and all necessary information is available

jorgebay: yeah, I ideally an earlier phase would be great.

I don't have a strong opinion tho, I don't want to drag this PR further and I would like to merge @tkolanko in the next few days.

spmallette: the trouble i had with `validate` was that it was too early if we wanted lint to consider generated sources. if we don't care about that then it's fine to change it. doesn't really bother me either way.

jorgebay:
Looks awesome! thanks @tkolanko !

PR #1585, opened 2022-03-10
Initial upload of docs for mergeE and mergeV

krlawrence:
Here is an initial PR containing the suggested documentation for `mergeV` and `mergeE`. Given the power, and likely rapid uptake, of these new steps, I have tried to do a fairly detailed coverage. To make the examples a little more fresh, I chose to show adding dogs to the graph and diverge from just showing yet another way to create the "Toy graph".  If people decide that they don't like this approach we can fall back to adding the toy graph.  I just get a lot of comments from users asking why all the examples in our docs use the same graph and felt that a slightly different data set made sense to demonstrate the power of these new steps. 

Not looking for a VOTE right away. Let's discuss and make any changes first.

### DIFF
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -2383,6 +2383,281 @@ link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gre
 link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.html#mean-org.apache.tinkerpop.gremlin.process.traversal.Scope-++[`mean(Scope)`],
 link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gremlin/process/traversal/Scope.html++[`Scope`]
 
+
+[[mergeedge-step]]
+=== MergeEdge Step
+
+The `mergeE()` -step is used to add edges and their properties to a graph in a "create
+if not exist" fashion. The `mergeE` step can also be used to find edges matching a given
+pattern. This step was added to Gremlin as part of the TinkerPop 3.6 release. Prior to
+that, it was necessary to write more complex queries using steps such as `coalesce`,
+`fold` and `unfold`, or injecting maps into queries and writing additional steps to
+manipulate them. The `mergeE` step makes these types of operation much simpler for the
+query writer. The input passed to `mergeE` can be either a map, or a child traversal that
+produces a map.
+
+NOTE: There is a corresponding `mergeV` step that can be used when creating vertices.
+
+Additionally, `optional` steps may be combined with `mergeE`, to take action depending on
+whether a vertex was created, or already existed. There are various ways that `mergeE` can
+be used. The simplest being to provide a single map of keys and values, along with the
+source and target vertex IDs, as a parameter.  An ID and a label may also be provided but
+this is optional. If the `T.id` or `T.label` are provided, they need to be enclosed in
+parenthesis. The `mergeE` step can be used directly from the GraphTraversalSource - `g`,
+or in the middle of a traversal. For a match with an existing vertex to occur, all values
+in the map must exist on a vertex; otherwise, a new vertex will be created. The examples
+that follow show how `mergeE` can be used to add relationships between dogs in the graph.       
+
+[gremlin-groovy]
+----
+g.mergeV([(T.id):1,(T.label):'Dog','name':'Toby'])
+g.mergeV([(T.id):2,(T.label):'Dog','name':'Brandy']) <1>
+g.mergeE([(T.label):'Sibling','created':'2022-02-07',(Direction.from):1,(Direction.to):2]) <2>
+g.E().elementMap()
+----
+
+<1> Create two vertices with ID values of 1 and 2.
+<2> Create a "Sibling" relationship between the vertices.
+
+For a `mergeE` step to succeed, both the `from` and `to` vertices must already exist. It
+is not possible to create new vertices directly using `mergeE`, but `mergeV` and `mergeE`
+steps can be combined, in a single query, to achieve that goal.
+
+NOTE: The `mergeE` step will not create vertices that do not exist. In those cases an
+error will be returned. 
+
+If the `Direction` enum has been statically included, its explicit use can be omitted from
+the query.
+
+[gremlin-groovy]
+----
+g.mergeV([(T.id):1,(T.label):'Dog','name':'Toby'])
+g.mergeV([(T.id):2,(T.label):'Dog','name':'Brandy'])
+g.mergeE([(T.label):'Sibling','created':'2022-02-07',(from):1,(to):2])
+g.E().elementMap()
+----
+
+One or more `option` steps can be used to control the behavior when an edge is created of
+updated.
+
+[gremlin-groovy]
+----
+g.mergeV([(T.id):1,(T.label):'Dog','name':'Toby'])
+g.mergeV([(T.id):2,(T.label):'Dog','name':'Brandy'])
+
+g.withSideEffect('map',[(T.label):'Sibling',(from):1,(to):2]).
+  mergeE(select('map')).
+    option(Merge.onCreate,['created':'2022-02-07']).  <1>
+    option(Merge.onMatch,['updated':'2022-02-07'])    <2>
+g.E().elementMap()
+----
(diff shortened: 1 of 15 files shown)


spmallette:
This change also includes some adjustments to the grammar to allow for `Map` definition to be a bit more Groovy-like. In particular, it allows for this type of parsing:

```groovy
[a: 100, "b", 200, id: 300, (T.label): 'person')]
```

It is of note, that unlike Groovy, an enum value like `id` will take precedence as the enum rather than the `String` value. If you want the string value to override the enum it should be placed in quotes.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -2302,6 +2302,281 @@ link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gre
 link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.html#mean-org.apache.tinkerpop.gremlin.process.traversal.Scope-++[`mean(Scope)`],
 link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gremlin/process/traversal/Scope.html++[`Scope`]
 
+
+[[mergeedge-step]]
+=== MergeEdge Step
+
+The `mergeE()` -step is used to add edges and their properties to a graph in a "create
+if not exist" fashion. The `mergeE` step can also be used to find edges matching a given
+pattern. This step was added to Gremlin as part of the TinkerPop 3.6 release. Prior to
```
spmallette, line 2311: > This step was added to Gremlin as part of the TinkerPop 3.6 release. Prior to....

This feels more like "Upgrade Documentation" to me rather than matching the style of our reference guide.  I've actually written something similar there. Not sure if you want to move this content there to replace mine, as you add a bit more detail, but I don't think it should stay here.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -2302,6 +2302,281 @@ link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gre
...
+that, it was necessary to write more complex queries using steps such as `coalesce`,
+`fold` and `unfold`, or injecting maps into queries and writing additional steps to
+manipulate them. The `mergeE` step makes these types of operation much simpler for the
+query writer. The input passed to `mergeE` can be either a map, or a child traversal that
+produces a map.
+
+NOTE: There is a corresponding `mergeV` step that can be used when creating vertices.
```
spmallette, line 2318: nit: hyperlink - `<<mergevertex-step, mergeV()-step>>`

docs/src/reference/the-traversal.asciidoc
```diff
@@ -2302,6 +2302,281 @@ link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gre
...
+
+Additionally, `optional` steps may be combined with `mergeE`, to take action depending on
```
spmallette, line 2320: not "optional", but "`option()` step modulators".

docs/src/reference/the-traversal.asciidoc
```diff
@@ -2302,6 +2302,281 @@ link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gre
...
 link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gremlin/process/traversal/Scope.html++[`Scope`]
 
+
+[[mergeedge-step]]
+=== MergeEdge Step
+
+The `mergeE()` -step is used to add edges and their properties to a graph in a "create
+if not exist" fashion. The `mergeE` step can also be used to find edges matching a given
+pattern. This step was added to Gremlin as part of the TinkerPop 3.6 release. Prior to
+that, it was necessary to write more complex queries using steps such as `coalesce`,
```
spmallette, line 2312: from a formatting perspective reference documentation refers to step names with double parens typically: so "`coalesce`" would be "`coalesce()`" or "`coalesce()`-step". Could you please review your content to fix that formatting a bit to keep things looking consistent?

docs/src/reference/the-traversal.asciidoc
```diff
@@ -2302,6 +2302,281 @@ link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gre
...
+produces a map.
+
+NOTE: There is a corresponding `mergeV` step that can be used when creating vertices.
+
+Additionally, `optional` steps may be combined with `mergeE`, to take action depending on
+whether a vertex was created, or already existed. There are various ways that `mergeE` can
+be used. The simplest being to provide a single map of keys and values, along with the
+source and target vertex IDs, as a parameter.  An ID and a label may also be provided but
+this is optional. If the `T.id` or `T.label` are provided, they need to be enclosed in
+parenthesis. The `mergeE` step can be used directly from the GraphTraversalSource - `g`,
```
spmallette, line 2325: > If the `T.id` or `T.label` are provided, they need to be enclosed in parenthesis.

This isn't true anymore after my follow-on commit to this one. I couldn't figure out a place to properly discuss `Map` initialization in the documentation. Perhaps it should just be included here and in mergev section as a `NOTE:`?

spmallette: hmm - i just realized our examples are being processed by Groovy and not the `gremlin-language`, so what you wrote is technically correct in that context. not sure how we want to rectify that now, but it needs to be clarified a bit.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -2302,6 +2302,281 @@ link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gre
...
+
+[gremlin-groovy]
+----
+g.mergeV([(T.id):1,(T.label):'Dog','name':'Toby'])
+g.mergeV([(T.id):2,(T.label):'Dog','name':'Brandy'])
+g.mergeE([(T.label):'Sibling','created':'2022-02-07',(from):1,(to):2])
+g.E().elementMap()
+----
+
+One or more `option` steps can be used to control the behavior when an edge is created of
```
spmallette, line 2359: nit: "of" -> "or"

docs/src/reference/the-traversal.asciidoc
```diff
@@ -2302,6 +2302,281 @@ link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gre
...
+----
+
+<1> Create three dogs.
+<2> Stream the edge maps into `mergeE` steps.
+
+
+[[mergevertex-step]]
+=== MergeVertex Step
+
+The `mergeV()` -step is used to add vertices and their properties to a graph in a "create
```
spmallette, line 2402: similar feedback here to what I already wrote in `mergeE()` regarding "upgrade docs" and the link to `mergeE()` in the NOTE.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -2302,6 +2302,281 @@ link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gre
...
+
+NOTE: If `mergeV` is given an empty map; such as `mergeV([:])`, it will match, and
+return, every vertex in the graph. This is the same behavior seen with `V([])`.
+
+[gremlin-groovy]
+----
+g.mergeV([(T.id):300]).
+  option(Merge.onCreate,[(T.id):300, (T.label):'Dog', 'name':'Toby', 'age':10]). <1>
+  option(Merge.onMatch,['age':11]) <2>
+
```
spmallette, line 2489: you might want to remove these linefeeds - i'm pretty sure that they will render as empty `gremlin>` lines when the docs get generated.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -2302,6 +2302,281 @@ link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gre
...
+g.V(400).valueMap().with(WithOptions.tokens)  <5>
+----
+
+<1> Use the first map to search for a vertex with an ID of 400.
+<2> If the vertex was not found, use the second map to create it.
+<3> If the vertex was found, add an `updated` property.
+<4> Pixel exists now, so we will take this option.
+<5> The `updated` property has now been added.
+
```
spmallette, line 2579: as previously noted for `mergeE()` please add "Additional References" here.

PR #1586, opened 2022-03-14
TINKERPOP-2719 Avoid unintentionally opening another transaction

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2719

`AbstractOpProcessor` already has a mechanism to avoid calling `hasNext()` after the iteration is already complete. We simply copied this mechanism to `SessionOpProcessor` and `TraversalOpProcessor`.

This should avoid problems where a transaction is used again after it was already closed.

Co-authored-by: Florian Grieskamp <florian.grieskamp@gdata.de>
Co-authored-by: kdrblkbs <kdrblkbs@gmail.com>

VOTE +1

### DIFF
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/traversal/TraversalOpProcessor.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/traversal/TraversalOpProcessor.java
@@ -383,9 +383,14 @@ public class TraversalOpProcessor extends AbstractOpProcessor {
                         break;
                     }
 
+                    // track whether there is anything left in the iterator because it needs to be accessed after
+                    // the transaction could be closed - in that case a call to hasNext() could open a new transaction
+                    // unintentionally
+                    final boolean moreInIterator = itty.hasNext();
+
                     try {
                         // only need to reset the aggregation list if there's more stuff to write
-                        if (itty.hasNext())
+                        if (moreInIterator)
                             aggregate = new ArrayList<>(resultIterationBatchSize);
                         else {
                             // iteration and serialization are both complete which means this finished successfully. note that
@@ -407,7 +412,7 @@ public class TraversalOpProcessor extends AbstractOpProcessor {
                         throw ex;
                     }
 
-                    if (!itty.hasNext()) iterateComplete(nettyContext, msg, itty);
+                    if (!moreInIterator) iterateComplete(nettyContext, msg, itty);
 
                     // the flush is called after the commit has potentially occurred.  in this way, if a commit was
                     // required then it will be 100% complete before the client receives it. the "frame" at this point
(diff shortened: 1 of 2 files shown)


spmallette:
Please include a CHANGELOG but other than that: VOTE +1

divijvaidya:
VOTE +1

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/traversal/TraversalOpProcessor.java
```diff
@@ -383,9 +383,14 @@ protected void handleIterator(final Context context, final Iterator itty, final
                         break;
                     }
 
+                    // track whether there is anything left in the iterator because it needs to be accessed after
+                    // the transaction could be closed - in that case a call to hasNext() could open a new transaction
+                    // unintentionally
+                    final boolean moreInIterator = itty.hasNext();
```
divijvaidya, line 389: Instead of creating a new variable, can we reset the value of existing variable `hasMore` here?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/traversal/TraversalOpProcessor.java
```diff
@@ -407,7 +412,7 @@ protected void handleIterator(final Context context, final Iterator itty, final
                         throw ex;
                     }
 
-                    if (!itty.hasNext()) iterateComplete(nettyContext, msg, itty);
+                    if (!moreInIterator) iterateComplete(nettyContext, msg, itty);
```
divijvaidya, line 415: can use hasMore here...

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/traversal/TraversalOpProcessor.java
FlorianHockmann: `hasMore` is only set to `false` in line 401 and that is in the `else` block of the if condition: `if (moreInIterator)`. So if we also use `hasMore` here instead, then `hasMore` cannot become `false` any more.
This additional `itty.hasNext()` is needed here to check whether more elements are available after the last `next()`. This is also explained in the comment above starting at line 337:

```
            // there is a need to check hasNext() on the iterator because if the channel is not writeable the
            // previous pass through the while loop will have next()'d the iterator and if it is "done" then a
            // NoSuchElementException will raise its head. also need a check to ensure that this iteration doesn't
            // require a forced flush which can be forced by sub-classes.
```

This is also in line with the implementation in `AbstractOpProcessor` where it's implemented exactly like this. We really just copied the logic from there.

FlorianHockmann:
> Please include a CHANGELOG

done

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/traversal/TraversalOpProcessor.java
FlorianHockmann: Thanks for the detailed explanation! Our understanding of your earlier suggestion was wrong based on that. Our assumption was that you wanted to get rid of the additional call `itty.hasNext()` completely.

But then it definitely sounds good to reuse `hasMore` hereinstead of adding `moreInIterator`. Pushed an updated commit that incorporates this change also for the `AbstractOpProcessor` which already used `moreInIterator` before.

divijvaidya:
Thank for making the change. Looks good to me.

Vote +1.

PR #1587, opened 2022-03-16
return future when Client.submitAsync is called

sputnik13:
driver.Client.submitAsync is not returning the result of self.submit_async, so the future is not being passed forward to the caller.

While it's understood this interface is deprecated, it should still behave as it previously did in 3.4 until it is removed.


lyndonbauto (approved):

lyndonbauto:
Thanks for this @sputnik13 - I am going to propagate this to master then merge this PR.

PR #1592, opened 2022-03-23
TINKERPOP-2726 Fix boolean translation in Python

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2726

This ensures that the GroovyTranslator in Python creates valid boolean values `true` /  `false` instead of `True` / `False`.

VOTE +1


spmallette:
a CHANGELOG entry would be nice - other than that VOTE +1

spmallette:
i think manual changelog entries are the better accounting of modifications because they often have better wording than the JIRAs and might also be more detailed in terms of changes. the generated JIRA list is just for reference imo

FlorianHockmann:
Added the CHANGELOG entry.

mikepersonick:
+1 but again, it worries me that we hadn't already caught this. See my comment on the GroovyTransalator PR. For each of these translators, we should run the whole Gherkin test suite through them and then compare the bytecode for the original and the translated. Or something like that.

PR #1593, opened 2022-03-23
TINKERPOP-2367 Add a GroovyTranslator to .NET

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2367

I still need to generate the docs to inspect the changes, but I think this is ready for review already.


FlorianHockmann:
The generated docs also look good to me, so: VOTE +1

spmallette:
nice addition -  would be good to add some Upgrade Documentation to call attention to this feature, but other than that VOTE +1

FlorianHockmann:
> would be good to add some Upgrade Documentation to call attention to this feature

Oh right, I only looked at the PR that added the translator to Python and didn't see any upgrade docs there, but the upgrade docs for 3.5.0 definitely had them:
https://tinkerpop.apache.org/docs/current/upgrade/#_translators

I'll add something to the docs.

mikepersonick:
I vote +1 but I am a little concerned about the test coverage. What would be great is if we could run the whole Gherkin test suite through this translator and then back through the DotNetTranslator and then compare the bytecode of the original and round trip version.

PR #1594, opened 2022-03-24
Gremlin-Go GA - Preview Ready

lyndonbauto:
# What This Pull Request Contains
This has the changes for Milestone 1 through 4 of the gremlin-go driver. The expectation is that this driver, though not fully featured, will work for almost all use cases.

Users should be able to use the driver like any other GLV, though there are some advanced features missing (and it has not been optimized with connection pooling yet).

# Cucumber Test Suite Considerations
We have all tests passing in the Cucumber test suite, have integrated the driver into the TinkerPop build infrastructure and have the driver ready for an actual release.

There has been some minor changes since the Cucumber test suite was last run, but I will re-run it and will update the PR with the result as we expect 100% pass rate there.

# Shoutout to Persons Involved
I would like to thank @xiazcy, @simonz-bq, @L0Lmaker, @vkagamlyk, @divijvaidya, @krlawrence and @spmallette for continued work they have all been putting into this either through implementing features or through code reviews.

# Outstanding Items from Prior Review
I wanted to get a PR up against 3.5-dev sooner than later with the impeding code freeze, so with that in mind I left a couple comments open from the past PR, which I will be addressing shortly. The comments are from @divijvaidya, and are follows:

> 1. For Client.Close() ===> Please document whether this API is idempotent or not. Ideally, a close API call should be idempotent. You can achieve this by adding a lifecycle state variable with client which goes from INITIALIZED -> CLOSING -> CLOSED. If close() API is called when the client is not in INITIALIZED state, then it should be a no-op.
> 2. Please log errors at error level logging before passing the error to the user application at all public interfaces such as NewClient or client.submit().
> 3. We are missing integration tests against the server (unless I missed something), which can validate the client behaviour in scenario such as:
>    1. server initiated abrupt WS connection disconnect
>    2. websocket handshake times out with the server
>    3. server disconnects a connection but client sends a query on that connection afterwards etc.
> 4. Please add a test which validates that all go routines are destroyed on client close


spmallette:
> We are missing integration tests against the server (unless I missed something), which can validate the client behaviour in scenario such as:

I'm not sure that we can do this right now. The only driver that has any tests like this is Java and that is because the test setup is such that we can easily start/stop/reconfigure Gremlin Server. There is no such mechanism to do so for non-JVM language variants. I do think this sort of testing is important but it will imply a whole range of test framework development to allow this to happen and I'm not sure what that would look like as it would require per-test mechanisms for that start/stop/reconfigure functionality. 

Unless there is an easy way to do this that I'm not aware of I'd say that such testing would have to be part of a different/larger body of work which builds a Gherkin-like integration test framework for driver behaviors that would cover connection management, sessions, etc.

divijvaidya:
> > We are missing integration tests against the server (unless I missed something), which can validate the client behaviour in scenario such as:
> 
> I'm not sure that we can do this right now. The only driver that has any tests like this is Java and that is because the test setup is such that we can easily start/stop/reconfigure Gremlin Server. There is no such mechanism to do so for non-JVM language variants. I do think this sort of testing is important but it will imply a whole range of test framework development to allow this to happen and I'm not sure what that would look like as it would require per-test mechanisms for that start/stop/reconfigure functionality.
> 
> Unless there is an easy way to do this that I'm not aware of I'd say that such testing would have to be part of a different/larger body of work which builds a Gherkin-like integration test framework for driver behaviors that would cover connection management, sessions, etc.

My main concern is that GA status for a component here in this project would mean that users can start using this in production scenarios. However, we can't confidently say that the client is production ready unless we have tested it on various scenarios involving client-server interaction.

> The only driver that has any tests like this is Java and that is because the test setup is such that we can easily start/stop/reconfigure Gremlin Server. There is no such mechanism to do so for non-JVM language variants. 

I understand what we did not have integration tests for other non-java drivers in the past. But we don't have to keep following the same practices from the past if they were incorrect. Given the scope of usage TinkerPop clients has across the industry, I would argue that we need to release well tested and stable software if we are making something GA.

We can still merge this PR into mainline without integration tests and call it Beta (not recommended for production usage) until we have sufficient testing in place. But I don't agree with calling this ready for production use (GA / preview-GA).

Now, as far as how to do the technical implementation goes, we can either do it against the TinkerPop server or a mock WebSocket server. The ultimate aim is to assert the client-server contract, e.g. does the client respect a close WebSocket frame from the server correctly.

**Option A: With live server**

1. Write an API in the server available in debug mode which will trigger the specific test behaviour we want, e.g. close the WS connection with a WS frame.

2. Use a shell script to start the server and then run a go file containing the integration test. The integration test assumes that a server is running on port 8182

3. Invoke the "test" API from point 1 to make the server do the desired behaviour. 

**Option B: With mock WS server** (All the scenarios that I wrote in the PR comment can be handled by a mock WS server)

Gorilla WebSocket library (the same one that client uses in this PR) already has example tests on how to mock a server. See: https://github.com/gorilla/websocket/blob/master/client_server_test.go#L62 

we can mock a server with WebSocket connection similarly and make it send a close websocket to see how the client reacts and assert the reaction. Similarly for other test scenarios I mentioned, we can mock the connection "disconnect" and can mock the "WS handshake timeout" using the mocks. We don't really need a full TinkerPop server to do this.

divijvaidya:
(Please correct me if I am wrong, there's a lot of files so I might have missed this)
Another missing component from this PR is the handling of Ping WebSocketFrame from the server.

Note that Gremlin server could have [WebSocket Idle monitor enabled](https://github.com/apache/tinkerpop/blob/master/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/AbstractChannelizer.java#L170). In such cases, server would detect an idle websocket connection and [send a ping frame](https://github.com/apache/tinkerpop/blob/master/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/UnifiedHandler.java#L328) to the client. But this implementation of client does not have a PingFrame handler from the server (and sending a pong websocket frame back).

spmallette:
> My main concern is that GA status for a component here in this project would mean that users can start using this in production scenarios. 

i feel like we have confusion and i'm not sure where it is coming from. i dont think anyone in the community is in favor of calling this "ready for production". that's why it's an "early release" an "alpha" an "preview" a "milestone" a "release candidate" a "beta"....not sure what you want to call it, but it is not going out with an "official" version stamp with "3.5.3" - it will be "v3.5.3-rc1" (or the like). 

we've released in this fashion before and it is designed to allow an easy way for users to consume some new feature before it is fully ready. typically these early releases are not widely promoted. For instance, we don't include the change in the Upgrade Documentation, Reference Documentation, etc.  However we'll likely make mention of it in a mailing list post and or directly contact those whom we know would be interested in giving it an early try knowing it may not be fully baked.

does that clarify the plan better?

divijvaidya:
> does that clarify the plan better?

Yep. Absolutely. That does make me comfortable with that fact that we done't have integration tests here. Thanks for being patient. Appreciate it! :)

spmallette:
There are a number of little odds/ends that I'd like to push on this, but I think we should focus in on merging it and I can do those changes after the fact. 

VOTE +1

krlawrence:
VOTE +1 to get this body of work merged and then doing any additional iteration so that we have a preview version of the client merged it. We have a lot of users looking for Go clients and getting their feedback ASAP will be very beneficial.

PR #1596, opened 2022-03-28
TINKERPOP-2727 Allow for a null value to has()

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2727

Seems to apply when the null is on the first arg of a two arg overload - it tries to call on has(T,...) which ends in IllegalArgumentException, which never quite surfaces to the JavaTranslator properly. Another example of why bytecode is lacking and why gremlin-language is likely going to be its successor.

VOTE +1


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/JavaTranslator.java
```diff
@@ -266,6 +267,23 @@ private Object invokeMethod(final Object delegate, final Class<?> returnType, fi
...
+                        // form of GraphTraversal can't be produced because of validations for has(T, ...) but in
+                        // other language it might be allowed which means that has((T) null, ...) from something like
+                        // python will end up as has((String) null, ...) which filters rather than generates an
+                        // exception. calling the T version even in a non-JVM language seems odd and more likely the
+                        // caller was shooting for the String one, but ugh who knows. this issue showcases again how
+                        // badly bytecode should either change to use gremlin-language and go away or bytecode should
+                        // get a safer way to be translated to a traversal with more explicit calls that don't rely
+                        // on reflection.
+                        if (methodName.equals(GraphTraversal.Symbols.has) && newArguments.length > 0 && null == newArguments[0] &&
+                            method.getParameterTypes()[0].isAssignableFrom(org.apache.tinkerpop.gremlin.structure.T.class)) {
```
divijvaidya, line 283: Please correct me if I am wrong here but the tests in this PR do not test for this condition that is mentioned in this comment `has((T) null, ...)` . Could we please add a test for it?

divijvaidya (commented):
Interesting use case. I had always assumed that key in has() is strictly a String and didn't realize it could be `null` as well.

Can we add a line at https://tinkerpop.apache.org/docs/current/reference/#has-step to specify the behaviour of `has()` step when key is `null` (i.e. filtering)? It is currently non obvious for a user of the language and is missing from JavaDocs as well.

FlorianHockmann (commented):
Thanks for fixing this so fast!

> I had always assumed that key in has() is strictly a String and didn't realize it could be null as well.

> It is currently non obvious for a user of the language and is missing from JavaDocs as well.

Agreed. I'm also not really sure where `null` is allowed and how the steps behave when they get `null` as an argument. Adding this to the ref / Java docs would make this more obvious.

Apart from that, this PR looks good to me.

spmallette:
i think i've addressed the issues here. for documentation of `has()` i added some details to the reference docs but ultimately i think we need to grow the[ Gremlin Semantics section](https://github.com/apache/tinkerpop/blob/master/docs/src/dev/provider/gremlin-semantics.asciidoc) to cover these sorts of details.

gremlin-test/features/filter/Has.feature
```diff
@@ -804,6 +813,48 @@ Feature: Step - has()
...
+    """
+    When iterated to list
+    Then the result should be unordered
+      | result |
+      | v[marko] |
+      | v[vadas] |
+      | v[josh] |
+      | v[peter] |
+
+  Scenario: g_V_hasXlabel_constantXpersonXX
```
FlorianHockmann, line 844: Shouldn't this be named `g_V_hasXlabel_isXpersonXX`?

docs/src/reference/the-traversal.asciidoc
```diff
@@ -1434,6 +1436,15 @@ the key,value pairs for those vertices.
 <6> Find all age-properties and emit their value.
 <7> Find all vertices that do not have an age-property and emit their name.
 <8> Find all "person" vertices that have a name property that starts with the letter "m".
+<9> Property key values cannot be `null` and therefore produces no matches.
```
FlorianHockmann, line 1439: Not that important, but _produce_ without an _s_, right?

docs/src/reference/the-traversal.asciidoc
```diff
@@ -1434,6 +1436,15 @@ the key,value pairs for those vertices.
...
+<10> An example of `has()` where the argument is a `Traversal` and does not quite behave the way most expect.
+
+Item 10 in the above set of examples bears some discussion. The behavior is not such that the result of the `Traversal`
+is used as the comparing value for `has()`, but the current `Traverser`, which in this case is the vertex `label`, is
+given to the `Traversal` to behave as a filter itself. In other words, if the `Traversal` (i.e. `is('person')`) returns
+a value then the `has()` is effectively `true`. A common mistake is to try to use `select()` in this context where one
+would do `has('name', select('n'))` to try to inject the value of "n" into the step to get `has('name', <value-of-n>)`,
```
FlorianHockmann, line 1446: Oh wow, good explanation. I expected `has()` to behave this way with the `is()` step, but I would also have expected it to work with `select()` like you described here. Good to know!

FlorianHockmann (approved):
I added two more nits, but this PR is also good to go without addressing them.

VOTE +1

docs/src/reference/the-traversal.asciidoc
divijvaidya, line 1439: nit (feel free to ignore)

perhaps rephrase to
> Property Key is always stored as `String` and therefore an equality check with `null` will produce no result.

divijvaidya (approved):
LGTM. Vote +1

docs/src/reference/the-traversal.asciidoc
spmallette: uhh - depending on how you read the subject of the sentence it could go either way. i intended "produces" so i'll reword.

PR #1605, opened 2022-03-29
TINKERPOP-2525 Extended gherkin tests to cover SubgraphStrategy

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2525

@FlorianHockmann please take a look at the .NET changes. I wasn't familiar with the CompatibilitySuppressions.xml file and not sure I liked how i dealt with the `checkAdjacentVertices` option that was missing from `SubgraphStrategy`. Perhaps you could suggest a better way than how I did it?

VOTE +1


gremlin-dotnet/src/Gremlin.Net/Process/Traversal/Strategy/Decoration/SubgraphStrategy.cs
```diff
@@ -53,6 +54,10 @@ public SubgraphStrategy() : base(JavaFqcn)
                 Configuration["edges"] = edges;
             if (vertexProperties != null)
                 Configuration["vertexProperties"] = vertexProperties;
+
+            // hate to write this by default so only going to write it if it's false which is the non-default value
+            if (!checkAdjacentVertices)
+                Configuration["checkAdjacentVertices"] = checkAdjacentVertices;
```
FlorianHockmann, line 60: Really just a nit, but my IDE (Rider) is complaining that `checkAdjacentVertices` will always be `false` here and suggests to change this to `Configuration["checkAdjacentVertices"] = false;` to make this more obvious.

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/Strategy/Decoration/SubgraphStrategy.cs
```diff
@@ -53,6 +54,10 @@ public SubgraphStrategy() : base(JavaFqcn)
                 Configuration["edges"] = edges;
             if (vertexProperties != null)
                 Configuration["vertexProperties"] = vertexProperties;
+
+            // hate to write this by default so only going to write it if it's false which is the non-default value
```
FlorianHockmann, line 58: Are we 100% sure that `true` will stay the default? Otherwise, we should maybe really just always add this so we don't at some point change the default in Java to `false` and then forget to update this here.

FlorianHockmann (commented):
> I wasn't familiar with the CompatibilitySuppressions.xml file

This compatibility check should ensure that we don't add breaking changes by accident. So, we have to add an entry to this file if we're adding a breaking change that we are OK with.
Here, I would ideally like the check to ignore this because you're only adding it on `master` which will be released in 3.6.0 and we're OK with breaking changes there in general.
So, we could use 3.6.0 as the version to compare against (`PackageValidationBaselineVersion` in the csproj), but since we haven't released 3.6.0 yet, we don't have a version to compare against.

Therefore, I'd keep this as you did it in this PR and then we can change the version to compare against on `master` to 3.6.0 after the release and then remove this entry again, or we could also
keep comparing against 3.5.2 and then also keep the entry as it doesn't really hurt in my opinion.

TLDR: I would have handled this exactly like you did here.

The scenarios seem to pass for Python, but isn't this option also missing [here](https://github.com/apache/tinkerpop/blob/d1dfe8e4ef5d61c44f375b74e3199e99394984fe/gremlin-python/src/main/python/gremlin_python/process/strategies.py#L81)?

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/Strategy/Decoration/SubgraphStrategy.cs
spmallette: decided to change it to always write it.

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/GherkinTestRunner.cs
```diff
@@ -182,16 +182,12 @@ private void OutputResults(List<ResultFeature> results)
                 foreach (var resultScenario in resultFeature.Scenarios)
                 {
                     totalScenarios++;
-                    WriteOutput($"  Scenario: {resultScenario.Key.Name}");
```
FlorianHockmann, line 185: This shortens the output to make it easier to see which scenarios are actually failing, right?

FlorianHockmann (approved):
LGTM, VOTE +1

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/Strategy/Decoration/SubgraphStrategy.cs
```diff
@@ -43,8 +43,9 @@ public SubgraphStrategy() : base(JavaFqcn)
         /// <param name="vertices">Constrains vertices for the <see cref="ITraversal" />.</param>
         /// <param name="edges">Constrains edges for the <see cref="ITraversal" />.</param>
         /// <param name="vertexProperties">Constrains vertex properties for the <see cref="ITraversal" />.</param>
+        /// <param name="checkAdjacentVertices">Determines if filters are applied to the adjacent vertices of an edge.</param>
         public SubgraphStrategy(ITraversal vertices = null, ITraversal edges = null,
-            ITraversal vertexProperties = null)
+            ITraversal vertexProperties = null, bool checkAdjacentVertices = true)
```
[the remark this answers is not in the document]
spmallette: this was what i wanted. i tried it but it didn't work for some reason yesterday. today, it's fine. 🤷

gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/GherkinTestRunner.cs
spmallette: yeah - added a note about that to the commit comment. it was getting way too hard to find the errors.

FlorianHockmann: Yeah makes sense, I've also ran into that problem in the past.

PR #1639, opened 2022-05-02
Gremlin Go Milestone 6

lyndonbauto:
This has the changes for Milestone 5 of the gremlin-go driver. I would like to thank @xiazcy, @simonz-bq, @L0Lmaker, @vkagamlyk, @divijvaidya, @spmallette, and @krlawrence for continued help they have all been putting into this.

We have also added enhanced documentation and performance testing.

We are planning to host a bug bash on May 4 [in discord](https://discord.gg/mQzHUyNP?event=966289876133761044), everyone, with or without prior Golang experience, is welcome to join.

Pull request comments and all bugs introduced via the discord event will be addressed in this pull request unless otherwise stated.

This pull request will remain open for at least two weeks, so the community has a chance to look at it and critique it.


.github/workflows/build-test.yml
```diff
@@ -227,6 +227,11 @@ jobs:
           docker-compose up --exit-code-from gremlin-go-integration-tests
           docker-compose down
 
+      - name: Upload to Codecov
```
spmallette, line 230: was there ever any investigation into whether ASF infra involvement was required to make this work?

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1897,3 +1897,300 @@ pip install gremlinpython
 pip install aiohttp
 python example.py
 ----
+
+[[gremlin-go]]
+== Gremlin-Go
+
+image:gremlin-go.png[width=130,float=right] Apache TinkerPop's Gremlin-Go implements Gremlin within the link:https://go.dev/[Go] language and can therefore be used on different operating systems. Go's syntax has the similar constructs as Java including
+"dot notation" for function chaining (`a.b.c`) and round bracket function arguments (`a(b,c)`). Something unlike Java is that Gremlin-Go requires a 
+`gremlingo` prefix when using the namespace (`a(b())` vs `gremlingo.a(gremlingo.__.b())`). Anyone familiar with Gremlin-Java will be able to work
```
spmallette, line 1906: `gremlingo.a(gremlingo.__.b())` is a bit misleading since the double-underscore isn't supported except by way of `var` initialization and then you wouldn't use the `gremlingo` prefix, right?

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1897,3 +1897,300 @@ pip install gremlinpython
...
+----
+results, err := g.With("evaluationTimeout", 500).V().Out("knows").ToList()
+----
+
+The following options are allowed on a per-request basis in this fashion: `batchSize`, `requestId`, `userAgent` and
+`evaluationTimeout` (formerly `scriptEvaluationTimeout` which is also supported but now deprecated).
+
+anchor:go-imports[]
+[[gremlin-go-imports]]
+=== Common Imports
```
spmallette, line 1967: this section should discuss the pattern of aliasing (i.e. assigning to `var`/`const`) to get a more Gremlin like feel:

```go
const Local = gremlingo.Local
var Neq = gremlingo.P.Neq
var __ = gremlingo.T__
var In = __.In
```

On the subject of naming, I think `T__` has been controversial for a while now but no one has come up with anything groundbreakingly better. The lesser of two evils is to just go with renaming `T__` to the full naming of `AnonymousTraversal` and then using the aliasing above to get it to be `__` like regular Gremlin. Furthermore, I think that enums like `Local` probably need their associated prefixes as in `Scope.Local`.

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1897,3 +1897,300 @@ pip install gremlinpython
...
+client, err := gremlingo.NewClient("ws://localhost:8182/gremlin") <2>
+----
+
+<1> Import the Gremlin-Go module.
+<2> Opens a reference to `localhost` - note that there are various configuration options that can be passed
+to the `Client` object upon instantiation as keyword arguments.
+
+Once a `Client` instance is ready, it is possible to issue some Gremlin:
+
+[source,go] //todo: clarify replacement one with One
```
spmallette, line 2116: todo ??

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1897,3 +1897,300 @@ pip install gremlinpython
...
+=== Limitations
+
+* There is no default `set` type in Go. Any set type code from server will be deserialized into slices with the list 
+type implementation. To input a set into Gremlin-Go, a custom struct which implements the `gremlingo.Set` interface 
+will be serialized as a set. `gremlingo.NewSimpleSet` is a basic implementation of a set that is provided by Gremlin-Go 
+that can be used to fulfill the `gremlingo.Set` interface if desired.
+* The `path` data type serialization is currently incomplete as labels are represented as list of lists instead of list 
+of sets. Fully functional Path serialization will be implemented when set is implemented in the next milestone. Path 
+can be successfully deserialized.
+* Traversal step functions currently take `string` arguments with double quotes only. Operations using Gremlin keywords,
```
spmallette, line 2182: is this still true? how are the gherkin tests passing - we surely have a `by(T.label)` somewhere in those tests?

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1897,3 +1897,300 @@ pip install gremlinpython
 pip install aiohttp
 python example.py
 ----
+
+[[gremlin-go]]
```
spmallette, line 1901: looks like this was added to the end of the document. the languages are in alphabetical order now.

gremlin-go/Dockerfile
```diff
@@ -15,23 +15,34 @@
...
+RUN apk update && apk add dos2unix
+RUN dos2unix /opt/docker-entrypoint.sh && apk del dos2unix
+
+ARG NEO4J_VERSION
 # Installs Neo4j libraries to this image so that we can test variants with transactions,
 # but only only port 45940 is configured with the neo4j graph as the neo4j-empty.properties
 # is statically pointing at a temp directory and that space can only be accessed by one
 # graph at a time.
-RUN /opt/gremlin-server/bin/gremlin-server.sh install org.apache.tinkerpop neo4j-gremlin $NEO4J_VERSION
+RUN /opt/gremlin-server/bin/gremlin-server.sh install org.apache.tinkerpop neo4j-gremlin ${NEO4J_VERSION:-3.5.4}
```
spmallette, line 39: how does this work? 3.5.4 doesn't exist yet. it should be 3.5.4-SNAPSHOT (i.e. this should match whatever gremlin server is which should match whatever version gremlin-go is which should match whatever the root pom.xml is.)

gremlin-go/README.md
```diff
@@ -22,58 +22,24 @@ for early testing purposes only.
 
 -->
 
-# Getting Started
```
spmallette, line 25: If we follow the pattern of the other README, we find they are fairly minimal and end user focused. Now that reference docs have been added for gremlin-go, I think it would be best to minimize the README contents to use end-user stuff, expand the reference/developer docs where necessary and then link to them from here.

gremlin-go/docker-compose.yml
```diff
@@ -24,6 +24,8 @@ services:
     build:
       context: ../
       dockerfile: gremlin-go/Dockerfile
+      args:
+        NEO4J_VERSION: 3.5.3
```
spmallette, line 28: hardcoding to a previous version isn't going to work. tests should run against the most recent build (i.e. whatever is in the current top-level pom.xml).

gremlin-go/docker/generate-all.groovy
```diff
@@ -0,0 +1,79 @@
+/*
```
spmallette, line 1: we've agreed that we're trying to establish a testing pattern here that we can extend to all GLVs, so i'll continue to ignore this duplication. just commenting here to continue to keep it forefront that we're working towards that goal and that test configuration copies is not going to be the approach taken.

gremlin-go/docker/gremlin-server-integration-secure.yaml
```diff
@@ -20,6 +20,7 @@ port: 45941
 evaluationTimeout: 30000
 graphs: {
   graph: conf/tinkergraph-empty.properties,
+  test: conf/tinkergraph-empty.properties,
```
spmallette, line 23: what is this additional "test" configuration for?

gremlin-go/driver/performance-tests/java-test/src/main/java/PerformanceTest.java
```diff
@@ -0,0 +1,165 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// DO NOT USE - THIS WILL BE DELETED
```
spmallette, line 20: We have a little `util` namespace in `gremlin-driver` where you could tuck this:

https://github.com/apache/tinkerpop/tree/master/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/util

then you wouldn't need this random pom.xml.  i'd also add your comments here to the class as javadoc rather than just floating.

gremlin-go/docker/gremlin-server-integration-secure.yaml
lyndonbauto: Cucumber and integration tests run in parallel, Cucumber was modifying the graph so we added an empty test graph to also use for testing in integration so that there wasn't race conditions causing intermittent failures with the parallel tests.

docs/src/reference/gremlin-variants.asciidoc
lyndonbauto: Both of these things are going to be worked on separately from this ticket, thanks for raising this!

gremlin-go/README.md
vkagamlyk: README is also used as package documentation https://pkg.go.dev/github.com/lyndonbauto/tinkerpop/gremlin-go/v3@v3.5.3-rc15

docs/src/reference/gremlin-variants.asciidoc
lyndonbauto: There are outdated, will remove

gremlin-go/Dockerfile
xiazcy: The -3.5.4 makes it default to 3.5.4, but can be changed via the "NEO4J_VERSION" build argument in docker-compose. I don't think there is a way to have the compose file dynamically parse the pom file, but I will make a run script to obtain the Gremlin server/Neo4j version and execute the docker-compose for now. Ultimately this will be resolved with the docker design works.

lyndonbauto:
All comments have been either addressed or have been acknowledged with plan to address at later date. 

Future pull requests are going to be incremental against 3.5-dev instead of these bulk milestone based pull requests.

PR #1659, opened 2022-05-16
TINKERPOP-2745: Add getter method for properties

li-boxuan:
This commit adds VertexProgram::getVertexPropertyKeys method
which returns a list of (property key, class type) pairs. The
pairs contain vertex property keys that are read or written in
the execution of the vertex program. Graph providers could
leverage this information to create schema ahead.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/VertexProgram.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/VertexProgram.java
@@ -20,6 +20,7 @@
 package org.apache.tinkerpop.gremlin.process.computer;
 
 import org.apache.commons.configuration2.Configuration;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.tinkerpop.gremlin.process.traversal.traverser.TraverserRequirement;
 import org.apache.tinkerpop.gremlin.structure.Element;
 import org.apache.tinkerpop.gremlin.structure.Graph;
@@ -27,6 +28,7 @@ import org.apache.tinkerpop.gremlin.structure.Vertex;
 
 import java.lang.reflect.Constructor;
 import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 
@@ -47,6 +49,10 @@ import java.util.Set;
 public interface VertexProgram<M> extends Cloneable {
 
     public static final String VERTEX_PROGRAM = "gremlin.vertexProgram";
+
+    default List<Pair<String, Class>> getVertexPropertyKeys() {
+        return Collections.emptyList();
+    }
 
     /**
      * When it is necessary to store the state of the VertexProgram, this method is called.
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/clustering/connected/ConnectedComponentVertexProgram.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/clustering/connected/ConnectedComponentVertexProgram.java
@@ -21,6 +21,7 @@ package org.apache.tinkerpop.gremlin.process.computer.clustering.connected;
 import org.apache.commons.configuration2.BaseConfiguration;
 import org.apache.commons.configuration2.Configuration;
 import org.apache.commons.configuration2.ConfigurationUtils;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.tinkerpop.gremlin.process.computer.GraphComputer;
 import org.apache.tinkerpop.gremlin.process.computer.Memory;
 import org.apache.tinkerpop.gremlin.process.computer.MemoryComputeKey;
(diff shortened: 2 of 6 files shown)


li-boxuan:
I am not sure how to test them in the TinkerPop codebase, but I have verified them using JanusGraph.

mikepersonick:
Would you mind taking a look at these and see if there is a way to add tests somehow?

https://github.com/apache/tinkerpop/blob/master/gremlin-test/features/map/PageRank.feature
https://github.com/apache/tinkerpop/blob/master/gremlin-test/features/map/PeerPressure.feature
https://github.com/apache/tinkerpop/blob/master/gremlin-test/features/map/ShortestPath.feature

And in the associated Java tests here:

https://github.com/apache/tinkerpop/tree/master/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/clustering/connected/ConnectedComponentVertexProgram.java
```diff
@@ -43,12 +44,10 @@
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.apache.tinkerpop.gremlin.structure.VertexProperty;
 
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.Set;
+import java.util.*;
```
spmallette, line 47: nit: please check your IDE settings - looks like it added a wildcard import. the tinkerpop style is to be explicit with import

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/clustering/connected/ConnectedComponentVertexProgram.java
```diff
@@ -76,6 +75,14 @@ public class ConnectedComponentVertexProgram implements VertexProgram<String> {
 
     private ConnectedComponentVertexProgram() {}
 
+    @Override
+    public List<Pair<String, Class>> getVertexPropertyKeys() {
+        return Arrays.asList(
+            org.apache.commons.lang3.tuple.Pair.of(HALTED_TRAVERSERS, Object.class),
```
spmallette, line 81: Please prefer `org.javatuples.Pair` as that is used throughout the project.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/VertexProgram.java
```diff
@@ -48,6 +50,10 @@
 
     public static final String VERTEX_PROGRAM = "gremlin.vertexProgram";
 
+    default List<Pair<String, Class>> getVertexPropertyKeys() {
```
spmallette, line 53: Is there a reason you can't get this information from `getVertexComputeKeys()`? Seems like it would be better if `VertexComputeKey` was modified to include the data type which is what you seem to want to add.  It feels like, with this change you'd be making the `VertexProgram` developer define the keys twice which leaves open the possibility for divergence.

Here's an example from this PR in `ConnectedComponentVertexProgram`:

```java
@Override
    public List<Pair<String, Class>> getVertexPropertyKeys() {
        return Arrays.asList(
            org.apache.commons.lang3.tuple.Pair.of(HALTED_TRAVERSERS, Object.class),
            org.apache.commons.lang3.tuple.Pair.of(ACTIVE_TRAVERSERS, Object.class),
            org.apache.commons.lang3.tuple.Pair.of(COMPONENT, Object.class));
    }
```

It defines `COMPONENT` when i think it should be `property`.

I think using the existing `getVertexComputeKeys()` approach also would solve the testing issue that @mikepersonick brought up because that method is implicitly tested already - you wouldn't really need to add anything new. Perhaps you would add some basic unit test to validate the new behavior in the `VertexComputeKey`?

You targeted `3.5-dev` so you need to do this in a fashion that is non-breaking. I sense that is possible if you add a new `VertexComputeKey.of(String, boolean, Class<?>)` and have the old `of(String,boolean)` call the new method with `Object.class` for the last argument. Then you could go through all of the existing `VertexProgram` instances and set the types explicitly. I  don't think that will break anything. 

Assuming this all makes sense and I'm not missing anything, please make sure to include a CHANGELOG entry when you're done.

li-boxuan: > It defines COMPONENT when i think it should be property

You are right, I was wrong. I used the default property name in the test so didn't catch this bug.

> Is there a reason you can't get this information from getVertexComputeKeys()

Thanks, I totally forgot about it 😅 . That should mostly suffice the use case I described, except that I am not 100% sure why other vertex programs don't contain `HALTED_TRAVERSERS` and `ACTIVE_TRAVERSERS` in `vertexComputeKeys`. Anyways, this PR is not very useful now so I am closing it. Thank you @mikepersonick and @spmallette for your reviews!

PR #1665, opened 2022-05-19
Gremlin-Go: Updated Docker Testing Set Up

xiazcy:
The current gremlin server version is now being dynamically parsed by an added run.sh script, which passes the version as environment variable and executes the docker compose for Gremlin Go testing. 

Installation of Neo4j is now moved from build stage to run stage of the server image, and a wait-for-server.sh script is added to docker compose that checks the server status before starting the integration tests. 

Docker testing section under README has been updated accordingly.


gremlin-go/wait-for-server.sh
```diff
@@ -0,0 +1,32 @@
...
+# under the License.
+#
+
+set -e
+
+host="$1"; shift
+port="$1"; shift
+
+until curl "$host:$port?gremlin=100-1"; do
+  >&2 echo "Gremlin server is unavailable - sleeping"
```
spmallette, line 27: not sure if this is an issue, but should there be a limit to how many times it tries? could this wait infinitely and hang if there is a problem?

gremlin-go/docker/gremlin-server-integration.yaml
```diff
@@ -18,15 +18,16 @@
...
 graphs: {
-  graph: conf/tinkergraph-empty.properties,
-  immutable: conf/tinkergraph-empty.properties,
-  classic: conf/tinkergraph-empty.properties,
-  modern: conf/tinkergraph-empty.properties,
-  crew: conf/tinkergraph-empty.properties,
-  grateful: conf/tinkergraph-empty.properties,
-  sink: conf/tinkergraph-empty.properties,
-  tx: conf/neo4j-empty.properties
+  graph: scripts/tinkergraph-empty.properties,
```
spmallette, line 23: was this intentional? i thought `tinkergraph-empty.properties` was in in `conf/`: https://github.com/apache/tinkerpop/tree/master/gremlin-server/conf

gremlin-go/docker-compose.yml
```diff
@@ -37,6 +39,7 @@ services:
       - .:/go_app
       - ../gremlin-test/features:/gremlin-test
     environment:
+      - HOST_SERVER=gremlin-test-server
       - CUCUMBER_FEATURE_FOLDER=/gremlin-test
       - GREMLIN_SERVER_URL=ws://gremlin-test-server:45940/gremlin
       - GREMLIN_SERVER_BASIC_AUTH_URL=wss://gremlin-test-server:45941/gremlin
```
spmallette, line 45: out of curiosity do you envision any issues with getting the kerberos version of the server running under this model?

gremlin-go/docker/gremlin-server-integration.yaml
xiazcy: Yes, this was intentional. The `tinkergraph-empty.properties` in the `conf` folder is for regular server set up, the `tinkergraph-empty.propertie` for testing was in the `test/scripts` folder. 
It is more of an oversight before as the difference between them didn't have an impact in 3.5-dev, but for 3.6-dev, integration tests needed the `gremlin.tinkergraph.service` set up for server, so I switched to the `test/scripts` folder here.

gremlin-go/wait-for-server.sh
xiazcy: Yes, it would hang is there was a problem with the server, I will add a limit/guard for that.

gremlin-go/docker/gremlin-server-integration.yaml
spmallette: hmm - what is the difference? the files aren't identical?

xiazcy: Right, the files are not identical. 
Example from 3.6, `.properties` in the `conf` only contains a simple empty graph https://github.com/apache/tinkerpop/blob/3.6-dev/gremlin-server/conf/tinkergraph-empty.properties, where as in the the `scripts` the Text Search and Degree Centrality services are enabled (which are needed for the cucumber `g.call()` tests) https://github.com/apache/tinkerpop/blob/3.6-dev/gremlin-server/src/test/scripts/tinkergraph-empty.properties.

spmallette: hmm - too many config/properties files. please make a note that unification would be ideal and that as you look to replicate this implementation pattern to the other language variants we should have that goal in mind. i actually think we might have missed something by not changing `gremlin-server/conf/tinkergraph-empty.properties` as that is the file that ships with the packaged Gremlin Server. would be nice if `call()` was ready to work there. none of this is a blocker for this PR.

spmallette:
VOTE +1, pending resolution of comments.

gremlin-go/wait-for-server.sh
xiazcy: Added a max limit field with usage in the script.

gremlin-go/docker/gremlin-server-integration.yaml
xiazcy: Sounds good, I will keep that in mind going forward. Thanks Stephen!

gremlin-go/docker-compose.yml
xiazcy: I cannot say for certain as I have not tried setting up a server with kerberos nor am I familiar enough with kerberos, but from what I understand, we can install kerberos to docker and containerize the KDC, server and client, exposing ports should allow communications between them in docker compose (I imagine issues, if any, would arise here). Do you have any particular concerns we should keep in mind?

PR #1667, opened 2022-05-24
Added getters to high and low properties in RangeLocalStep

steigma:


lyndonbauto:
VOTE+1 Thanks @steigma. This has passed the 1 week lazy consensus so I will deal with merging this.

PR #1669, opened 2022-05-26
Fix deadlock on non response-specific read errors

simonz-bq:
- Fixed a potential deadlock that would occur whenever reading from the server fails in a fashion that is not specific to a specific request, e.g. the connection is dead
- Fixed a potential edge case where a ResultSet would not return an error (it would only log it) alongside an incomplete set of results

### DIFF
--- a/gremlin-go/driver/resultSet.go
+++ b/gremlin-go/driver/resultSet.go
@@ -35,6 +35,7 @@ type ResultSet interface {
 	GetRequestID() string
 	IsEmpty() bool
 	Close()
+	unlockedClose()
 	Channel() chan *Result
 	addResult(result *Result)
 	One() (*Result, bool, error)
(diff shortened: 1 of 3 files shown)


gremlin-go/driver/resultSet.go
```diff
@@ -120,6 +121,19 @@ func (channelResultSet *channelResultSet) Close() {
 	}
 }
 
+// Close and remove from the channelResultSet from the container without locking container. Meant for use when calling
```
lyndonbauto, line 124: Is `Close` still used? Overall this LGTM and I think solves this well - just wondering if we can remove the other Close() now.

lyndonbauto: If it is used, perhaps we should have
`LockedClose()`
and
`unlockedClose()`

and both should call 
```
func (channelResultSet *channelResultSet) close(locked bool) {
	if !channelResultSet.closed {
		channelResultSet.channelMutex.Lock()
		channelResultSet.closed = true
		if lock {
     			channelResultSet.container.delete(channelResultSet.requestID)
		} else {
			delete(channelResultSet.container.internalMap, channelResultSet.requestID)
		}
		close(channelResultSet.channel)
		channelResultSet.channelMutex.Unlock()
		channelResultSet.sendSignal()
	}
}```

Comment (simonz-bq) [reply gremlin-go/driver/resultSet.go:124]: Close is still used in the happy path where a ResultSet is fully read, as well as being an exposed point for the user to mark it as completed. This makes it hard for us to modify the API around that.

File: gremlin-go/driver/resultSet.go [hunk 2]
```diff
@@ -120,6 +121,19 @@ func (channelResultSet *channelResultSet) Close() {
...
+// function already locks the container.
+func (channelResultSet *channelResultSet) unlockedClose() {
```
vkagamlyk, line 126: what happens if 2 go routines this method simultaneously?

lyndonbauto: I think the idea is that this should never be called outside of closeAll and closeAll uses a mutex to ensure there is no concurrent usage.

vkagamlyk: it makes sense

lyndonbauto:
VOTE+1

Because this is a minor bug fix and I have full context on this issue I am going to my vote alone.

PR #1671, opened 2022-05-27
TINKERPOP-2749: Gremlin Javascript, Gremlint Windows build error fixes

L0Lmaker:
Currently, Tinkerpop does not build on Windows. Running `mvn clean install` causes errors in `gremlin-javascript` and other modules.

This PR fixes the errors caused by `gremlin-javascript`. It also adds a Windows build check to the GitHub Actions workflows.

Issues addressed:
1. Fixed an issue where groovy script defined in `gremlin-javascript/pom.xml` receives an incorrectly formatted path for Windows. Single `\` characters are invalid in groovy and using `${project.basedir}` returns a string with non-escaped `\` characters.
2. Fixed an issue where lint/line-break checks would fail with message: `Expected linebreaks to be 'LF' but found 'CRLF'`
3. Fixed an issue where relative addressing is used, which was causing `'.' is not recognized as an internal or external command, operable program or batch file.` errors.
4. Fixed an issue where incorrectly formatted paths taken from `pom.xml` properties would lead to `test-server-start.groovy` initializing integration test server incorrectly on Windows.

Link to Issue: [TINKERPOP-2749](https://issues.apache.org/jira/browse/TINKERPOP-2749)

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/.eslintrc.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/.eslintrc.js
@@ -31,9 +31,9 @@ module.exports = {
   extends: ['eslint:recommended', 'prettier'],
   plugins: ['prettier'],
   rules: {
-    'prettier/prettier': ['error'],
+    'prettier/prettier': ['error', { endOfLine: 'auto' }],
     indent: ['error', 2, { SwitchCase: 1 }],
-    'linebreak-style': ['error', 'unix'],
+    'linebreak-style': 0,
     quotes: ['error', 'single'],
     semi: ['error', 'always'],
     'no-constant-condition': ['error', { checkLoops: false }],
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -126,7 +126,10 @@ jobs:
     name: javascript
     timeout-minutes: 15
     needs: smoke
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
     steps:
       - uses: actions/checkout@v3
       - name: Set up JDK 11
--- a/gremlin-javascript/pom.xml
+++ b/gremlin-javascript/pom.xml
@@ -116,10 +116,11 @@ limitations under the License.
                         <configuration>
                             <scripts>
                                 <script>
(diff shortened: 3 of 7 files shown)


lyndonbauto:
@L0Lmaker Looks like we got a conflict now, could you rebase?

gremlin-javascript/src/main/javascript/gremlin-javascript/.eslintrc.js
```diff
@@ -31,9 +31,9 @@ module.exports = {
   extends: ['eslint:recommended', 'prettier'],
   plugins: ['prettier'],
   rules: {
-    'prettier/prettier': ['error'],
```
FlorianHockmann, line 34: The changes in this file allow line endings to be `CRLF`, right? Is that really necessary? I guess we'd prefer if only Unix style line endings were committed, but maybe I'm missing something here?

FlorianHockmann (commented):
Great work, I just executed the tests locally on my Windows machine and they passed! It's really nice to get better support for developing on Windows :tada:
Also good to have this tested in CI so we can be sure that we don't accidentally break Windows support in the future.

I've only added one question, but these changes look already good to me (a rebase is of course still necessary).

VOTE +1

gremlin-javascript/src/main/javascript/gremlin-javascript/.eslintrc.js
L0Lmaker: There are two ways we could resolve this. Either allow for `CRLF` formatted files or convert the existing javascript files that use the `CRLF` format to `LF`.  The former felt like a simpler fix which is why I disabled the check for line separators.

If we want to keep it only Unix style, we could go ahead with converting all the JS files that are affected.

lyndonbauto:
I think we are okay to move forward with this once the actions pass. I will handle merging this through.

gremlin-javascript/src/main/javascript/gremlin-javascript/.eslintrc.js
lyndonbauto: I think we can leave this one just to move this along. If someone wants to go through and convert all the line endings, that can be done separately.

FlorianHockmann: Oh, I wasn't aware that there are already files with `CRLF`. Then I understand the change. Thanks for the explanation!

PR #1674, opened 2022-06-01
Bump logback version to 1.2.9 or greater

simonz-bq:
Fixes https://issues.apache.org/jira/browse/TINKERPOP-2746

### DIFF
--- a/gremlin-console/pom.xml
+++ b/gremlin-console/pom.xml
@@ -503,7 +503,7 @@ limitations under the License.
                             <repository>tinkerpop/gremlin-console</repository>
                         </configuration>
                     </plugin>
-                </plugins>                
+                </plugins>
             </build>
         </profile>
     </profiles>
--- a/pom.xml
+++ b/pom.xml
@@ -171,7 +171,7 @@ limitations under the License.
         <jcabi.version>1.1</jcabi.version>
         <junit.version>4.13.1</junit.version>
         <kerby.version>2.0.1</kerby.version>
-        <logback.version>1.2.3</logback.version>
+        <logback.version>1.2.9</logback.version>
         <metrics.version>3.0.2</metrics.version>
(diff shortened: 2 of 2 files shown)


spmallette:
logback version is defined in the root pom.xml in `<dependencyManagement>`:

https://github.com/apache/tinkerpop/blob/master/pom.xml#L822-L826

as a result, i don't think you need to add the `<version>` to all the pom.xml individually.

simonz-bq:
> logback version is defined in the root pom.xml in `<dependencyManagement>`:
> 
> https://github.com/apache/tinkerpop/blob/master/pom.xml#L822-L826
> 
> as a result, i don't think you need to add the `<version>` to all the pom.xml individually.

The Tinkerpop pom.xml is complex so I can't say with certainty if it's necessary.

My thought process is that if you build using the main pom.xml, then the dependency will be grabbed based on the version specified, and therefore satisfies the dependency requirement of the individual pom.xml when those dependencies are resolved.

However, if someone were to modularly use the individual maven projects as a standalone, I had a concern that then the dependency would not enforce a non-vulnerable versioning, hence why I added it.

If someone can confirm that this is not an issue, then I think removing the added `<version`> to the individual pom is good.

spmallette:
you can remove it. the point of `<dependencyManagement>` is to centralize dependency definitions so you don't need to repeat yourself throughout the sub-modules. you should be able to prove it to yourself by doing a `mvn dependency:tree -pl <sub-module>` and check the logback version you find there.

>  if someone were to modularly use the individual maven projects as a standalone

just fyi, they would have to make a lot of changes to the a sub-project pom to get it to work in this fashion, so i would hope they would just skip trying to do that.

lyndonbauto:
VOTE+1 I will merge this through sometime today.

PR #1676, opened 2022-06-02
TINKERPOP-2693 Finish graphbinary support in gremlin-python

vkagamlyk:
Finish graphbinary support in gremlin-python

https://issues.apache.org/jira/browse/TINKERPOP-2693

### DIFF
--- a/gremlin-python/src/main/python/gremlin_python/driver/driver_remote_connection.py
+++ b/gremlin-python/src/main/python/gremlin_python/driver/driver_remote_connection.py
@@ -59,7 +59,7 @@ class DriverRemoteConnection(RemoteConnection):
         # so that they can be closed if this parent session is closed.
         self.__spawned_sessions = []
 
-        if message_serializer is None:
+        if message_serializer is None and graphson_reader is not None and graphson_writer is not None:
             message_serializer = serializer.GraphSONMessageSerializer(
                 reader=graphson_reader,
                 writer=graphson_writer)
--- a/gremlin-python/src/main/python/gremlin_python/driver/driver_remote_connection.py
+++ b/gremlin-python/src/main/python/gremlin_python/driver/driver_remote_connection.py
@@ -59,7 +59,7 @@ class DriverRemoteConnection(RemoteConnection):
         # so that they can be closed if this parent session is closed.
         self.__spawned_sessions = []
 
-        if message_serializer is None and graphson_reader is not None and graphson_writer is not None:
+        if message_serializer is None:
             message_serializer = serializer.GraphSONMessageSerializer(
                 reader=graphson_reader,
                 writer=graphson_writer)
--- a/gremlin-python/src/main/python/gremlin_python/structure/io/graphbinaryV1.py
+++ b/gremlin-python/src/main/python/gremlin_python/structure/io/graphbinaryV1.py
@@ -31,10 +31,11 @@ from struct import pack, unpack
 from aenum import Enum
 from datetime import timedelta
 from gremlin_python import statics
-from gremlin_python.statics import FloatType, FunctionType, IntType, LongType, TypeType, DictType, ListType, SetType, \
-                                   SingleByte, ByteBufferType, GremlinType, SingleChar
-from gremlin_python.process.traversal import Barrier, Binding, Bytecode, Cardinality, Column, Direction, Operator, \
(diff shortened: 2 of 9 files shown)


gremlin-python/src/main/python/gremlin_python/driver/driver_remote_connection.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, url, traversal_source="g", protocol_factory=None,
         # so that they can be closed if this parent session is closed.
         self.__spawned_sessions = []
 
-        if message_serializer is None:
+        if message_serializer is None and graphson_reader is not None and graphson_writer is not None:
```
lyndonbauto, line 62: Did you encounter a case where graphson_read/graphson_writer are None?

Just wondering why this is required here but not in client.py

gremlin-python/src/main/python/gremlin_python/structure/io/graphbinaryV1.py
```diff
@@ -255,15 +257,15 @@ class LongIO(_GraphBinaryTypeIO):
     @classmethod
     def dictify(cls, obj, writer, to_extend, as_value=False, nullable=True):
         if obj < -9223372036854775808 or obj > 9223372036854775807:
-            raise Exception("TODO: don't forget bigint")
+            raise Exception("Value too big, please use bigint Gremlin type")
```
lyndonbauto, line 260: Should we just promote the type to bigint here? That way we handle this implictly and explicitly. Meaning you could do something like:

`g.V().....property("bigint_explicit", bigint(1)).property("bigint_implicit", 9223372036854775807 + 1)`.

gremlin-python/src/main/python/tests/driver/test_client.py
```diff
@@ -27,10 +27,13 @@
 from gremlin_python.process.strategies import OptionsStrategy
 from gremlin_python.structure.graph import Graph
 from gremlin_python.driver.aiohttp.transport import AiohttpTransport
+from gremlin_python.statics import *
```
lyndonbauto, line 30: best practice is to import specific items
```suggestion
from gremlin_python.statics import bigint, long
```

lyndonbauto: I see the other test just imports * already. Perhaps we can ignore this comment since the other tests don't consider that.

lyndonbauto (changes requested):
Couple minor comments, let me know when they are addressed and I will re-review.

Thanks for taking on this work @vkagamlyk

gremlin-python/src/main/python/gremlin_python/structure/io/graphbinaryV1.py
vkagamlyk: implicit conversion can introduce new issues, for example 
`g.inject(1).repeat(__.addV('person').property('name', __.loops())).times(9223372036854775807 + 1).count()`

leads to server exception
`[ERROR] TraversalOpProcessor - Could not deserialize the Traversal instance
java.lang.IllegalStateException: null: .times(BigInteger)`

if you think that this is not an issue, then I will add implicit conversion

gremlin-python/src/main/python/gremlin_python/driver/driver_remote_connection.py
vkagamlyk: this code is related to setting default serializer

gremlin-python/src/main/python/gremlin_python/structure/io/graphbinaryV1.py
lyndonbauto: Fair point. I am not sure myself I will tag some people who might know - @krlawrence @spmallette any thoughts?

gremlin-python/src/main/python/gremlin_python/driver/driver_remote_connection.py
vkagamlyk: reverted in https://github.com/apache/tinkerpop/pull/1676/commits/ff4e16c18236af61491e680cbf810664aba38e56

lyndonbauto:
LGTM VOTE+1

lyndonbauto (approved):

lyndonbauto:
Going to merge on lazy consensus.

PR #1680, opened 2022-06-03
TINKERPOP-2708: unhandledRejection upon connection failure

xiazcy:
As suggested in https://issues.apache.org/jira/browse/TINKERPOP-2708, deprecated and removed the functionality of the connection option `connectOnStartup`. Since it is generally not good to return promises in constructors, removing the call to `open()` in the `Connection` constructor seemed to be the way to resolve unhandledRejection and any potential race conditions.

There is a slight API behaviour change as a result. Setting `connectOnStartup` true now will log a deprecated warning to the user. The `Connection` constructor will no longer initiate connection in the background by default, so users will need to call `open()` explicitly after creating a connection object. However, open() is called anyways when connection submits traversals, removing this shouldn't affect any normal traversal functionality. 

As this may be considered a breaking change, I added a change-log entry, are there additional docs I should edit? 

As a side note, this is related to https://issues.apache.org/jira/browse/TINKERPOP-2381. I believe removing connectOnStartup functionality resolves the issues listed. I'm not sure if there is a need to refactor how Client invokes `connection.open()`, since inside `connection.open()`, it does check if it has been called previously.


spmallette:
> As this may be considered a breaking change, I added a change-log entry, are there additional docs I should edit?

breaking changes are typically documented in the Upgrade Documentation so that we can call attention to important changes more easily with greater detail than the CHANGELOG - it is probably worth adding an item to say that this parameter no longer does anything and that you would want to call `open()` explicitly to "connect on startup". also, if it is a breaking change, there should be a "breaking" label on the JIRA. finally, i would assume the deprecation to be removed on merge to master for 3.7.0, so perhaps you should consider a follow-on PR to get rid of the warning message and parameter all together on that branch.

separately, there is a merge commit in the history. could you please remove that? other than that VOTE +1

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
```diff
@@ -103,13 +103,15 @@ class Connection extends EventEmitter {
     this._pingIntervalDelay = this.options.pingInterval || pingIntervalDelay;
     this._pongTimeoutDelay = this.options.pongTimeout || pongTimeoutDelay;
 
-    if (this.options.connectOnStartup !== false) {
-      this.open();
+    if (this.options.connectOnStartup) {
```
lyndonbauto, line 106: The above comment that was updated used to say the default was true, but that is not entirely correct is it? Technically it was just `!== false`?

Just want to ensure this console warning will not be printing out when a user does not set this issue.

lyndonbauto:
VOTE+1

PR #1682, opened 2022-06-03
Specify ujson version of 2.0.0 or greater

simonz-bq:
Fixes https://issues.apache.org/jira/browse/TINKERPOP-2631

Since ujson is an optional dependency, hard enforcing a version may cause problems for users.

Instead, a version >= 2.0.0 is specified to be automatically resolved by the Python package manager if someone were to use install gremlin-python with ujson enabled. A warning is also given when using a version below 2.0.0 is detected.

### DIFF
--- a/gremlin-python/src/main/python/setup.py
+++ b/gremlin-python/src/main/python/setup.py
@@ -79,8 +79,9 @@ setup(
         'PyHamcrest>=1.9.0,<2.0.0'
     ],
     install_requires=install_requires,
-    extra_require={
-        'kerberos': 'kerberos>=1.3.0,<2.0.0'    # Does not install in Microsoft Windows
+    extras_require={
+        'kerberos': 'kerberos>=1.3.0,<2.0.0',    # Does not install in Microsoft Windows
+        'ujson': 'ujson>=2.0.0'
     },
     classifiers=[
         "Intended Audience :: Developers",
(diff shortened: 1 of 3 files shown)


gremlin-python/src/main/python/setup.py
```diff
@@ -79,8 +79,9 @@
         'PyHamcrest>=1.9.0,<2.0.0'
     ],
     install_requires=install_requires,
-    extra_require={
-        'kerberos': 'kerberos>=1.3.0,<2.0.0'    # Does not install in Microsoft Windows
+    extras_require={
+        'kerberos': 'kerberos>=1.3.0,<2.0.0',    # Does not install in Microsoft Windows
+        'ujson': 'ujson>=2.0.0'
```
lyndonbauto, line 84: Possibly a stupid question, but why is this in extra_require instead of [install_require](https://github.com/apache/tinkerpop/pull/1682/files#diff-f7172d5670efe2a25eb77a7385821c9e7bfa40ec98b215b8fee4ffae7e0105faR47)

simonz-bq: It's a non-built-in version of json for Python, and was considered optional if the user wanted to use it in their system or they already had the dependency from another part on their system. Putting it in install_require will download the ujson dependency and use it always.

This can't prevent someone from using ujson, but it will show them a warning if they're under version 2. Specifying it here also gives Python developers a place to look, and allows people to use it if they want with `pip install gremlin-python[ujson]`

lyndonbauto:
Thanks @simonz-bq, will merge now with lazy consensus.

PR #1688, opened 2022-06-08
adding multi source djikstra algorithm

nikita15p:
Signed-off-by: nikita15p <nikita15p@gmail.com>

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/search/path/MultiSourceDjistraVertexProgram
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/search/path/MultiSourceDjistraVertexProgram
@@ -1,0 +1,624 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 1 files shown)


spmallette:
thanks for offering this. here's some general comments:

1. This looks mostly like a copy/paste of the existing `ShortestPathVertexProgram` with some minor change to `collectShortestPaths()`. is that the only change? If it is, is then I think you'd want to rethink how this is implemented so that you could offer this new functionality and make better re-use of existing code.
2. For an "algorithm" I wouldn't typically be in favor of accepting a PR with just the `VertexProgram` implementation. We would want it accessible from Gremlin as we do with other algorithms which could mean a Gremlin language change (maybe?) which would in turn require some discussion. That's just my opinion though...perhaps others would have different feeling. 
3. Tests?
4. There's more administrative things to bring up (like documentation), if this were to be considered for merge, but that can all be saved for a later when the bigger items already listed are more settled.

nikita15p:
@spmallette  thanks for comments. So this functionality offers multi source djikstra algorithm and it leverages shortestPath code...So shall I be imporing ShortestPathVertexProgram methods then?

spmallette:
I'm not sure what you mean by:

> So shall I be imporing ShortestPathVertexProgram methods then?

spmallette:
closing this PR as there hasn't been any activity on it. if you're still interested in it, please feel free to open a new one with the feedback provided addressed.

PR #1690, opened 2022-06-08
TINKERPOP-2749: Gremlin Dotnet Windows Build fixes

L0Lmaker:
Link to Issue: [TINKERPOP-2749](https://issues.apache.org/jira/browse/TINKERPOP-2749)

This PR addresses Errors that cause the Windows build of Gremlin Dotnet to fail.

Issues Addressed:
1. Fixed an issue where groovy script defined in `gremlin-dotnet/src/pom.xml` receives an incorrectly formatted path for Windows. Single `\` characters are invalid in groovy and using `${project.basedir}` returns a string with non-escaped `\` characters in Windows.

_Note: The .NET windows build workflow has not yet been activated as there were inconsistent failures popping up on GitHub actions. Around 1-2 out of 10 workflow runs would fail due to inconsistent test failures. This issue needs to be more thoroughly investigated. Locally, however, the build has no issues on Windows with these changes._

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/gremlin-dotnet/src/pom.xml
+++ b/gremlin-dotnet/src/pom.xml
@@ -55,15 +55,16 @@ limitations under the License.
                         <configuration>
                             <scripts>
                                 <script>
-def mavenVersion = "${project.version}"
-def file = new File("${project.basedir}/Gremlin.Net/Gremlin.Net.csproj")
-file.write(file.getText("UTF-8").replaceFirst(/&lt;Version&gt;(.*)&lt;\/Version&gt;/, "&lt;Version&gt;" + mavenVersion + "&lt;/Version&gt;"))
+                                    def mavenVersion = "${project.version}"
+                                    def platformAgnosticBaseDirPath = new File(".this-definitely-does-not-exist").absolutePath.replace("\\.this-definitely-does-not-exist", "").replace("/.this-definitely-does-not-exist", "").replace("\\", "/")
+                                    def file = new File(platformAgnosticBaseDirPath + "/gremlin-dotnet/src/Gremlin.Net/Gremlin.Net.csproj")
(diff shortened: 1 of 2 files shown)


FlorianHockmann:
This is only about getting the Maven build to run without building the .NET code, i.e., without adding `-P gremlin-dotnet` which requires the .NET SDK to be installed, right? Just to be clear: It's ok in my opinion to first concentrate only on the maven build. I just wanted to ask so we're sure what's included here and what is not. It would be great in that case if you could create a follow-up issue in Jira where you note what's still TODO after TINKERPOP-2749 is done.

L0Lmaker:
The .NET code builds successfully as well, but there is an issue where intermediate GitHub actions runs fail for the windows build (~1-2 out of 10). This issue needs to be more thoroughly investigated so for now I haven't added the Windows build in the GHA jobs. My apologies, I should've explained why I didn't add it in. I have updated the PR description to include this detail.

Ofcourse! I will create a follow-up issue detailing all the rest of the work for this issue in JIRA.

FlorianHockmann:
Thanks for the explanation!

I also just tried it on my machine again and got it working, including the tests. The build somehow only converts some special characters sometimes, but I'm not sure what causes this and it might be specific to my machine. So we can ignore this for now from my side.

VOTE +1

lyndonbauto:
VOTE+1

spmallette:
Cherry-picked as 4675011d960c60b499b1c6300591fdd2968ac8c6 to get rid of the merge commit. I had a conflict merging to 3.6-dev and wasn't sure how to proceed so I kept what was in 3.6-dev unchanged.  

@L0Lmaker could you please have a look at how to port this change to 3.6-dev (if necessary) and issue another PR to that branch?

L0Lmaker:
Created a follow-up issue on JIRA.
Link to issue: [TINKERPOP-2760](https://issues.apache.org/jira/browse/TINKERPOP-2760)

PR #1694, opened 2022-06-10
Gremlin-Go: Updated cucumber/gremlin.go to match translator updates - 3.5-dev only

xiazcy:
It looks like the regenerated `gremlin.go` file was not uploaded to 3.5-dev after the strategies update in the go translator, which will result in file changes after building locally. Only affects 3.5-dev, both 3.6-dev and master are up-to-date.


lyndonbauto:
VOTE+1. 

Going to merge this through on my vote only and skip lazy consensus because this is just an autogenerated file that was supposed to be checked in before and was missed. I will investigate if the 3.6-dev build is correct and fix it if not.

PR #1696, opened 2022-06-10
Gremlin-Go: Removed unneeded helper function from go cucumber suite

xiazcy:
The helper function `strategyFactory` was no longer needed with the updated translator and `gremlin.go` file, so can be removed from the cucumber suite in go.


lyndonbauto:
Build failure here is due to a floating gremlin-python test bug which I have now fixed so not a concern.

VOTE+1.

PR #1697, opened 2022-06-10
TINKERPOP-2734 Fix potential SessionedClient NPE

simonz-bq:
Potentially fixes https://issues.apache.org/jira/browse/TINKERPOP-2734

Could not reproduce the issue without given reproduction steps, and also seems that it would require specific server circumstance.

However, investigation of the log traces the reporter included in the ticket gives clues of where to look, and a potential but unlikely NPE was logically deduced. Unsure if this will fix this specific user's problem but worth fixing regardless.

As a side effect, also reduces an unneeded coupling due to inheritance on the Client type.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -35,6 +35,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Fixed bug in `PartitionStrategy` where the use of `AbstractLambdaTraversal` caused an unexpected exception.
 * Fixed bug where close requests for sessions were improperly validating the request in the `UnifiedChannelizer`.
 * Deprecated and removed functionality of the `connectOnStartup` option in `gremlin-javascript` to resolve potential `unhandledRejection` and race conditions.
+* Fixed potential `NullPointerException` in `gremlin-driver` where initialization of a `ConnectionPool` would fail but not throw an exception due to centralized error check being satisfied by a different process.
 
 [[release-3-5-3]]
 === TinkerPop 3.5.3 (Release Date: April 4, 2022)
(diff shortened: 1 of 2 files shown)


alexey-temnikov (commented):
LGTM as a possible fix, as now an exception should be thrown if the connection pool is not initialized.

spmallette:
generally speaking, i don't really see anything wrong with this change. i dont' think it introduces anything bad and the decoupling of logic is probably a good idea. could we question the theory of this change a bit though? i suppose the assumption here that `connectionPool` is somehow `null` when it gets to `SessionedClient.chooseConnection()` and that is the NPE. if so, what circumstance allows that to happen and how does this change either potentially fix or better expose that problem?

alexey-temnikov:
@spmallette, in the reported error logs the cause exception stack trace got swallowed - `Could not initialize client for Optional[..]` which may indicate the connection pool hasn't been initialized in `SessionedClient.initializeImplementation()`. By rethrowing the error and decoupling the code instead of NPE we should be able to see a proper underlying stack trace which could lead to the root cause.

spmallette:
> which may indicate the connection pool hasn't been initialized in SessionedClient.initializeImplementation()

I understand what it is doing, but I'm more asking, is there a theory as to why it is not getting initialized?

simonz-bq:
> if so, what circumstance allows that to happen and how does this change either potentially fix or better expose that problem?

In the case that initializing the connection pool had a failure *but* somehow the cluster has available hosts `cluster.availableHosts().isEmpty()` in the list (not necssarily alive), there wouldn't be a connection pool initialized but an error would not be throw because of this check being possibly false.

spmallette:
sounds good @simonz-bq - that's what i saw too. i could envision where you had a multiple hosts and the session happened to pick an unavailable one at random and then separately a different process could find an available one so that by the time the first session reach the centralized error check it would find an available host.  Could you please amend the commit message to include the description of this potentiality? Also, please add a CHANGELOG entry. I'd say this is good to merge following those two things.

PR #1698, opened 2022-06-13
Complete ResponseStatusCodeEnum in .NET (3.6-dev)

danielcweber:
Add missing response status codes to the enum in .NET. Currently, it is not possible to use the new fail()-operator since it'll 
return a new status code (595) which makes the `IndicatesError` method throw.

@FlorianHockmann Maybe this justifies an out-of-band release of a patched Gremlin.NET ?

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/Messages/ResponseStatusCode.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/Messages/ResponseStatusCode.cs
@@ -137,9 +137,14 @@ namespace Gremlin.Net.Driver.Messages
                 case ResponseStatusCode.Authenticate:
                     return false;
                 case ResponseStatusCode.Unauthorized:
+                case ResponseStatusCode.Forbidden:
+                case ResponseStatusCode.TooManyRequests:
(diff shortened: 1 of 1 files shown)


FlorianHockmann (commented):
Good find and thanks for fixing this!

Unfortunately, most of these status code were already present before 3.6.0. So, we ideally need a separate PR for `3.5-dev` to add those codes already there. That should be all except for `SERVER_ERROR_FAIL_STEP(595)` if I'm not mistaken. Could you maybe create a separate PR for `3.5-dev`?

Regarding an out-of-band release only for Gremlin.NET: I think we could only do something like a release candidate for the next version, so `3.5.4-rc1` / `3.6.1-rc1` as the versioning of Gremlin.NET is aligned with that of TinkerPop in general.
If the `fail()` step is important for you, then we can probably make that happen. Publishing such release candidates doesn't require that much effort.

danielcweber:
PR for the 3.5-dev-branch to be found in #1710.

danielcweber:
@FlorianHockmann

>If the fail() step is important for you...

It'll become important to us once AWS Neptune moves to 3.6. Thanks for considering an rc-release!

PR #1699, opened 2022-06-13
fix: Traversal.graph is empty in StepStrategy.apply() with `count().is(0)`

javeme:
This PR fixed:
`traversal.getGraph()` is empty in `StepStrategy.apply()` method `VertexStepStrategy.apply(Traversal.Admin traversal)` for the following Gremlin:
```java
g.V(3).repeat(inE('child').outV().simplePath())
 .until(or(inE().count().is(0),loops().is(eq(2))))
 .path()
```

#### Initial data:
```
g.addV('node').property(id, 1).as('1')
 .addV('node').property(id, 2).as('2')
 .addV('node').property(id, 3).as('3')
 .addV('node').property(id, 4).as('4')
 .addE('child').from('1').to('2')
 .addE('child').from('2').to('3')
 .addE('child').from('4').to('3')
```

#### Cause analysis:
In the case of a non `count().is(0)` statement, the graph context is passed here:
https://github.com/apache/tinkerpop/blob/f8af06198ee5d112599fb938f97ff9707a83c81c/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java#L138

So we can get the graph context when Strategy.apply() is called:
https://github.com/apache/tinkerpop/blob/f8af06198ee5d112599fb938f97ff9707a83c81c/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java#L152

But when using the `count().is(0)` statement, the graph context is missing because the original traversal is replaced with a new traversal by `__.start()`:
https://github.com/apache/tinkerpop/blob/7f7d3a485c7f100f98047b71672a0c2c9ab855b4/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/CountStrategy.java#L154

The `__.start()` method implementation:
https://github.com/apache/tinkerpop/blob/2c24493f7643139cd8197ed27a03180a924a5704/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/__.java#L59

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -41,6 +41,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Deprecated and removed functionality of the `connectOnStartup` option in `gremlin-javascript` to resolve potential `unhandledRejection` and race conditions.
 * Fixed potential `NullPointerException` in `gremlin-driver` where initialization of a `ConnectionPool` would fail but not throw an exception due to centralized error check being satisfied by a different process.
 * Fixed a bug where the JavaScript client would hang indefinitely on traversals if the connection to the server was terminated.
+* Fixed Traversal.graph is empty in StepStrategy.apply() with `count().is(0)`.
 
 [[release-3-5-3]]
 === TinkerPop 3.5.3 (Release Date: April 4, 2022)
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/CountStrategy.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/CountStrategy.java
@@ -164,6 +164,7 @@ public final class CountStrategy extends AbstractTraversalStrategy<TraversalStra
                                 } else {
                                     inner = __.identity().asAdmin();
                                 }
+                                traversal.getGraph().ifPresent(graph -> inner.setGraph(graph));
                                 if (prev != null)
                                     TraversalHelper.replaceStep(prev, new NotStep<>(traversal, inner), traversal);
                                 else


javeme:
The following is the stack trace when `traversal.getGraph().get()` returned null:

```java
2022-06-13 23:18:00 [gremlin-server-exec-2] [WARN] o.a.t.g.s.h.HttpHandlerUtil - Invalid request - responding with 500 Internal Server Error and Error encountered evaluating script: g.V(3).repeat(inE('child').outV().simplePath()).until(or(inE().count().is(0),loops().is(eq(2)))).path()
java.lang.NullPointerException: null
	at com.baidu.hugegraph.traversal.optimize.HugeVertexStepStrategy.apply(HugeVertexStepStrategy.java:73) ~[classes/:?]
	at com.baidu.hugegraph.auth.HugeGraphAuthProxy$TraversalStrategyProxy.apply(HugeGraphAuthProxy.java:1722) ~[classes/:?]
	at org.apache.tinkerpop.gremlin.process.traversal.util.TraversalHelper.applyTraversalRecursively(TraversalHelper.java:467) ~[gremlin-core-3.5.1.jar:3.5.1]
	at org.apache.tinkerpop.gremlin.process.traversal.util.TraversalHelper.applyTraversalRecursively(TraversalHelper.java:475) ~[gremlin-core-3.5.1.jar:3.5.1]
	at org.apache.tinkerpop.gremlin.process.traversal.util.TraversalHelper.applyTraversalRecursively(TraversalHelper.java:475) ~[gremlin-core-3.5.1.jar:3.5.1]
	at org.apache.tinkerpop.gremlin.process.traversal.util.TraversalHelper.applyTraversalRecursively(TraversalHelper.java:475) ~[gremlin-core-3.5.1.jar:3.5.1]
	at org.apache.tinkerpop.gremlin.process.traversal.util.DefaultTraversal.applyStrategies(DefaultTraversal.java:151) ~[gremlin-core-3.5.1.jar:3.5.1]
	at org.apache.tinkerpop.gremlin.process.traversal.util.DefaultTraversal.hasNext(DefaultTraversal.java:221) ~[gremlin-core-3.5.1.jar:3.5.1]
	at org.apache.tinkerpop.gremlin.util.iterator.IteratorUtils.fill(IteratorUtils.java:62) ~[gremlin-core-3.5.1.jar:3.5.1]
	at org.apache.tinkerpop.gremlin.util.iterator.IteratorUtils.list(IteratorUtils.java:85) ~[gremlin-core-3.5.1.jar:3.5.1]
	at org.apache.tinkerpop.gremlin.util.iterator.IteratorUtils.asList(IteratorUtils.java:382) ~[gremlin-core-3.5.1.jar:3.5.1]
	at org.apache.tinkerpop.gremlin.server.handler.HttpGremlinEndpointHandler.lambda$channelRead$1(HttpGremlinEndpointHandler.java:221) ~[gremlin-server-3.5.1.jar:3.5.1]
	at org.apache.tinkerpop.gremlin.util.function.FunctionUtils.lambda$wrapFunction$0(FunctionUtils.java:36) ~[gremlin-core-3.5.1.jar:3.5.1]
	at org.apache.tinkerpop.gremlin.groovy.engine.GremlinExecutor.lambda$eval$0(GremlinExecutor.java:278) ~[gremlin-groovy-3.5.1.jar:3.5.1]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) [?:1.8.0_111]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_111]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) [?:1.8.0_111]
	at com.baidu.hugegraph.auth.HugeGraphAuthProxy$ContextTask.run(HugeGraphAuthProxy.java:1838) [classes/:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [?:1.8.0_111]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) [?:1.8.0_111]
	at java.lang.Thread.run(Thread.java:745) [?:1.8.0_111]
```

spmallette:
sorry no one has had a chance to look at this until now. i think you've identified the problem well. A few things:

1. you supplied an example in the PR description where this was failing, but could you turn that into an actual Gremlin test? 
1. did you mean to target `master`? i assume this problem goes all the way back to 3.5.3, right? if so could you please retarget your PR for `3.5-dev` branch? 
1. can you please add a CHANGELOG entry?
1. i'm concerned this is a wider issue within strategies. i just recently fixed a similar issue in `PartitionStrategy`. i can't help wondering if there is a bigger problem here that needs to be addressed. i dont think strategy authors should have to worry about this detail.  i suppose that needs to be looked at separately from this PR though.

javeme:
Thanks for your review.

1. I will try to add a gremlin test later.
2. Done.
3. Done.
4. Yes, we may need to find a general way to avoid this case, checking one by one is hard work and often miss something.

spmallette:
i have this more generalized fix i'm considering: https://github.com/apache/tinkerpop/commit/8a7049f5685c57bf6b523f41b8b3025523e89b52

i don't like that it adds yet another round of recursion. i'm questioning if strategy application should even be responsible for this, but its always been this way. i don't think this adds any significant performance penalty as strategy application is already pretty fast. in any event, if you have a way to test this fix on your end, that would be helpful but i'm leaning toward this approach for at least the branches we can't take breaks in.

javeme:
I think it's ok to assign graph instance after each strategy is applied, it ensures that the `apply()` of subsequent strategies can get the graph instance.

So do we need to merge this PR as a bug record, or abandon it?

javeme:
I have compiled and tested [this fix](https://github.com/apache/tinkerpop/commit/8a7049f5685c57bf6b523f41b8b3025523e89b52) on hugegraph backend, it works as expected.

spmallette:
if that fix works then i think we would just close this PR and assume the problem satisfied by the change you tested. thanks for bringing this up. i'll close this after the other fix gets merged.

spmallette:
merged the fix noted above as 1a548e808922a0eae23b586fe4d6567238989299

spmallette:
@javeme this is a bit old now, but it seems we must revisit it. the fix mentioned here is being rolled back for 3.5.6/3.6.3. that extra recursion is proving too costly. please see the latest work on this here: https://github.com/apache/tinkerpop/pull/2026 if you find the `Graph` still isn't there for some reason, you can pull it with `Traversal.getRoot(traversal).getGraph()` or in the worst case, we might want to look into figuring out where else we might get the `Graph` assigned as part of traversal construction, but so far, five graph implementations seem ok with how this change works.

javeme:
Thank you for your notice.

According to https://github.com/apache/tinkerpop/pull/2026 it seems that the `Graph` is set in the lock() method, right?
https://github.com/apache/tinkerpop/pull/2026/files#diff-33e6e540ffbb42c4d9ac4a78e83fbb9c28168d0731340b6d1b2e2a680324f081R320

PR #1700, opened 2022-06-13
gremlin-go: support per-request arguments in bytecode

jroimartin:
The issue that triggered this change is that per-request options were not working with gremlin-go. For instance setting a per-request timeout this way:

```go
	result, err := g.With("evaluationTimeout", 5000).V().Count().Next()
	if err != nil {
		// Handle error.
	}
```

A very similar problem was described some time ago for the Python GLV in the JIRA issue [TINKERPOP-2296](https://issues.apache.org/jira/browse/TINKERPOP-2296) and solved in the pull-request [apache/tinkerpop#1329](https://github.com/apache/tinkerpop/pull/1329).

Stephen Mallette explains the root cause of the issue in a comment in JIRA:

> There is no real strategy application in GLVs - the strategies are simply shipped to the server as bytecode configurations. By that point, it's already bypassed the point where the server will care about the timeout (or other request setting) overrides.

The Python GLV fixes this issue by:

1. Converting the "with" options into an OptionsStrategy.
2. Extracting a whitelisted set of options from the OptionsStrategy.
3. Sending the extracted options as part of the bytecode request.

However, this means that the generated bytecode will not always match the code written by the users. E.g.

```
	with(k, v) -> withStrategies(OptionsStrategy({k: v}))
```

So, this PR takes a slightly different approach. It extracts the whitelisted set of per-request options in makeBytecodeRequests() and adds them to the bytecode request without modifying the generated bytecode.

### DIFF
--- a/gremlin-go/driver/connection_test.go
+++ b/gremlin-go/driver/connection_test.go
@@ -1128,19 +1128,63 @@ func TestConnection(t *testing.T) {
 		g := getTestGraph(t, testNoAuthUrl, testNoAuthAuthInfo, testNoAuthTlsConfig)
 		defer g.remoteConnection.Close()
 
-		_, err := g.With("evaluationTimeout", 10).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Next()
-		assert.NotNil(t, err)
+		reqArgsTests := []struct {
+			msg       string
+			traversal *GraphTraversal
+			nilErr    bool
+		}{
+			{
+				"Traversal must time out (With)",
+				g.
+					With("evaluationTimeout", 10).
+					Inject(1).
+					SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}),
+				false,
+			},
+			{
+				"Traversal must finish (With)",
+				g.
+					With("evaluationTimeout", 10000).
+					Inject(1).
+					SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}),
+				true,
+			},
+			{
+				"evaluationTimeout is overridden and traversal must time out (With)",
+				g.
+					With("evaluationTimeout", 10000).With("evaluationTimeout", 10).
+					Inject(1).
+					SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}),
+				false,
+			},
+			{
+				"Traversal must time out (OptionsStrategy)",
+				g.
+					WithStrategies(OptionsStrategy(map[string]interface{}{"evaluationTimeout": 10})).
+					Inject(1).
+					SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}),
+				false,
+			},
+			{
+				"Traversal must finish (OptionsStrategy)",
+				g.
+					WithStrategies(OptionsStrategy(map[string]interface{}{"evaluationTimeout": 10000})).
+					Inject(1).
+					SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}),
+				true,
+			},
+		}
 
-		_, err = g.With("evaluationTimeout", 10000).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Next()
-		assert.Nil(t, err)
+		gotErrs := make([]<-chan error, len(reqArgsTests))
 
-		_, err = g.With("evaluationTimeout", 10000).With("evaluationTimeout", 10).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Next()
-		assert.NotNil(t, err)
+		// Run tests in parallel.
+		for i, tt := range reqArgsTests {
+			gotErrs[i] = tt.traversal.Iterate()
+		}
 
-		_, err = g.WithStrategies(OptionsStrategy(map[string]interface{}{"evaluationTimeout": 10})).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Next()
-		assert.NotNil(t, err)
-
-		_, err = g.WithStrategies(OptionsStrategy(map[string]interface{}{"evaluationTimeout": 10000})).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Next()
-		assert.Nil(t, err)
+		// Check error promises.
+		for i, tt := range reqArgsTests {
+			assert.Equal(t, <-gotErrs[i] == nil, tt.nilErr, tt.msg)
+		}
 	})
 }
--- a/gremlin-go/driver/connection_test.go
+++ b/gremlin-go/driver/connection_test.go
@@ -1121,4 +1121,26 @@ func TestConnection(t *testing.T) {
 		// This routine.
 		assert.Equal(t, startCount, runtime.NumGoroutine())
 	})
+
+	t.Run("Test per-request arguments", func(t *testing.T) {
+		skipTestsIfNotEnabled(t, integrationTestSuiteName, testNoAuthEnable)
+
+		g := getTestGraph(t, testNoAuthUrl, testNoAuthAuthInfo, testNoAuthTlsConfig)
+		defer g.remoteConnection.Close()
+
+		_, err := g.With("evaluationTimeout", 10).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Next()
+		assert.NotNil(t, err)
+
+		_, err = g.With("evaluationTimeout", 10000).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Next()
+		assert.Nil(t, err)
+
+		_, err = g.With("evaluationTimeout", 10000).With("evaluationTimeout", 10).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Next()
+		assert.NotNil(t, err)
+
+		_, err = g.WithStrategies(OptionsStrategy(map[string]interface{}{"evaluationTimeout": 10})).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Next()
+		assert.NotNil(t, err)
+
+		_, err = g.WithStrategies(OptionsStrategy(map[string]interface{}{"evaluationTimeout": 10000})).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Next()
+		assert.Nil(t, err)
+	})
 }
--- a/gremlin-go/driver/request.go
+++ b/gremlin-go/driver/request.go
@@ -91,11 +91,10 @@ func makeBytecodeRequest(bytecodeGremlin *Bytecode, traversalSource string, sess
 // allowedReqArgs contains the arguments that will be extracted from the
 // bytecode and sent with the request.
 var allowedReqArgs = map[string]bool{
-	"evaluationTimeout":       true,
-	"scriptEvaluationTimeout": true,
-	"batchSize":               true,
-	"requestId":               true,
-	"userAgent":               true,
+	"evaluationTimeout": true,
+	"batchSize":         true,
+	"requestId":         true,
+	"userAgent":         true,
 }
 
 // extractReqArgs extracts request arguments from the provided bytecode.
--- a/gremlin-go/driver/request.go
+++ b/gremlin-go/driver/request.go
@@ -75,12 +75,96 @@ func makeBytecodeRequest(bytecodeGremlin *Bytecode, traversalSource string, sess
 		newProcessor = sessionProcessor
 		newArgs["session"] = sessionId
 	}
+
+	for k, v := range extractReqArgs(bytecodeGremlin) {
+		newArgs[k] = v
+	}
+
 	return request{
 		requestID: uuid.New(),
 		op:        bytecodeOp,
 		processor: newProcessor,
 		args:      newArgs,
 	}
+}
+
+// allowedReqArgs contains the arguments that will be extracted from the
+// bytecode and sent with the request.
+var allowedReqArgs = map[string]bool{
+	"evaluationTimeout":       true,
+	"scriptEvaluationTimeout": true,
+	"batchSize":               true,
+	"requestId":               true,
+	"userAgent":               true,
+}
+
+// extractReqArgs extracts request arguments from the provided bytecode.
+func extractReqArgs(bytecode *Bytecode) map[string]interface{} {
+	args := make(map[string]interface{})
+
+	for _, insn := range bytecode.sourceInstructions {
+		switch insn.operator {
+		case "withStrategies":
+			for k, v := range extractWithStrategiesReqArgs(insn) {
+				args[k] = v
+			}
+		case "with":
+			if k, v := extractWithReqArgs(insn); k != "" {
+				args[k] = v
+			}
+		}
+	}
+
+	return args
+}
+
+// extractWithStrategiesReqArgs extracts request arguments from the passed
+// "withStrategies" source instruction. Only OptionsStrategy is considered.
+func extractWithStrategiesReqArgs(insn instruction) map[string]interface{} {
+	args := make(map[string]interface{})
+
+	for _, strategyInterface := range insn.arguments {
+		strategy, ok := strategyInterface.(*traversalStrategy)
+		if !ok {
+			// (*GraphTraversalSource).WithStrategies accepts
+			// TraversalStrategy parameters only. Thus, this
+			// should be unreachable.
+			continue
+		}
+
+		if strategy.name != decorationNamespace+"OptionsStrategy" {
+			continue
+		}
+
+		for k, v := range strategy.configuration {
+			if allowedReqArgs[k] {
+				args[k] = v
+			}
+		}
+	}
+
+	return args
+}
+
+// extractWithReqArgs extracts a request argument from the passed "with" source
+// instruction.
+func extractWithReqArgs(insn instruction) (key string, value interface{}) {
+	if len(insn.arguments) != 2 {
+		// (*GraphTraversalSource).With accepts two parameters. Thus,
+		// this should be unreachable.
+		return "", nil
+	}
+
+	key, ok := insn.arguments[0].(string)
+	if !ok {
+		return "", nil
+	}
+
+	if !allowedReqArgs[key] {
+		return "", nil
+	}
+
+	return key, insn.arguments[1]
 }
 
 func makeBasicAuthRequest(auth string) (req request) {
--- a/gremlin-go/driver/request.go
+++ b/gremlin-go/driver/request.go
@@ -108,7 +108,7 @@ func extractReqArgs(bytecode *Bytecode) map[string]interface{} {
 				args[k] = v
 			}
 		case "with":
-			if k, v := extractWithReqArgs(insn); k != "" {
+			if k, v := extractWithReqArg(insn); k != "" {
 				args[k] = v
 			}
 		}
@@ -145,9 +145,9 @@ func extractWithStrategiesReqArgs(insn instruction) map[string]interface{} {
 	return args
 }
 
-// extractWithReqArgs extracts a request argument from the passed "with" source
+// extractWithReqArg extracts a request argument from the passed "with" source
 // instruction.
-func extractWithReqArgs(insn instruction) (key string, value interface{}) {
+func extractWithReqArg(insn instruction) (key string, value interface{}) {
 	if len(insn.arguments) != 2 {
 		// (*GraphTraversalSource).With accepts two parameters. Thus,
 		// this should be unreachable.
--- a/docs/src/reference/gremlin-variants.asciidoc
+++ b/docs/src/reference/gremlin-variants.asciidoc
@@ -1451,7 +1451,7 @@ results, err := g.With("evaluationTimeout", 500).V().Out("knows").ToList()
 ----
 
 The following options are allowed on a per-request basis in this fashion: `batchSize`, `requestId`, `userAgent` and
-`evaluationTimeout` (formerly `scriptEvaluationTimeout` which is also supported but now deprecated).
+`evaluationTimeout`.
 
 anchor:go-imports[]
 [[gremlin-go-imports]]
@@ -1627,7 +1627,7 @@ resultSet, err := client.Submit("g.V().repeat(both()).times(100)", map[string]in
 ----
 
 The following options are allowed on a per-request basis in this fashion: `batchSize`, `requestId`, `userAgent` and
-`evaluationTimeout` (formerly `scriptEvaluationTimeout` which is also supported but now deprecated).
+`evaluationTimeout`.
 
 IMPORTANT: The preferred method for setting a per-request timeout for scripts is demonstrated above, but those familiar
 with bytecode may try `g.with("evaluationTimeout", 500)` within a script. Scripts with multiple traversals and multiple
--- a/gremlin-go/docker/gremlin-server-integration-secure.yaml
+++ b/gremlin-go/docker/gremlin-server-integration-secure.yaml
@@ -18,7 +18,7 @@
 host: 0.0.0.0
 port: 45941
 evaluationTimeout: 30000
-channelizer: org.apache.tinkerpop.gremlin.server.channel.UnifiedChannelizer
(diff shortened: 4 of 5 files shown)


xiazcy:
Thank you so much for submitting the PR @jroimartin, we have enabled the GitHub workflows and will be taking a closer look into this.

gremlin-go/driver/request.go
```diff
@@ -83,6 +88,85 @@ func makeBytecodeRequest(bytecodeGremlin *Bytecode, traversalSource string, sess
...
+	"batchSize":               true,
+	"requestId":               true,
+	"userAgent":               true,
+}
+
+// extractReqArgs extracts request arguments from the provided bytecode.
+func extractReqArgs(bytecode *Bytecode) map[string]interface{} {
+	args := make(map[string]interface{})
+
+	for _, insn := range bytecode.sourceInstructions {
```
simonz-bq, line 105: Nit: It's good practice to avoid abbreviations.

simonz-bq:
Overall, from a code perspective LGTM. Will let someone more familiar with the feature itself review as well.

gremlin-go/driver/request.go
jroimartin: Thanks for the review!

Do you mean using `insn` instead of `instruction` or `sourceInstruction`?

I usually try to stick to the following criteria described by Russ Cox:

> Descriptive names are fine for things that merit description at every mention.
>
> Receiver names typically don’t, nor do names scoped to a small amount of code (like a small loop body or even a small function).

(Reference: https://twitter.com/_rsc/status/1507319374612414472)

Also, I didn't want to shadow the `instruction` type.

That being said, I'm more than happy to change it if that fits better with the coding style of the project. What do you suggest?

gremlin-go/driver/request.go
```diff
@@ -83,6 +88,85 @@ func makeBytecodeRequest(bytecodeGremlin *Bytecode, traversalSource string, sess
 	}
 }
 
+// allowedReqArgs contains the arguments that will be extracted from the
+// bytecode and sent with the request.
+var allowedReqArgs = map[string]bool{
+	"evaluationTimeout":       true,
+	"scriptEvaluationTimeout": true,
```
spmallette, line 95: i wouldn't bother supporting `scriptEvaluationTimeout` - it's a deprecated argument where `evaluationTimeout` is preferred.

spmallette:
> It extracts the whitelisted set of per-request options in makeBytecodeRequests() and adds them to the bytecode request without modifying the generated bytecode.

I don't think that's a bad approach and I can't think of any problems with this offhand. Perhaps python (other GLVs) should be done that way too. Generally speaking, it's nicer when the bytecode matches what was originally constructed. 

Normally I'd suggest a CHANGELOG entry or even upgrade docs, but since the Go GLV isn't officially released yet, we can forego that as you've basically just added a feature that should have been present anyway from the start. 😄 

VOTE +1

gremlin-go/driver/request.go
jroimartin: Thanks for the review!

Would it be a problem to keep it for now? If I'm not wrong, this option is still supported by other GLVs and mentioned in the Tinkerpop reference. By supporting it we anticipate to users that try to use the old name and don't understand why the timeout is being ignored or why it does not works with all GLVs. I feel it is better to be consistent among GLVs and, once the setting is fully deprecated, removing the support from all the languages at once. What do you think?

Some instances,

**gremlin-python**

https://github.com/apache/tinkerpop/blob/7ad487b75992a4c46871ae9aab2d5451c133eb8b/gremlin-python/src/main/python/gremlin_python/driver/driver_remote_connection.py#L171

**Tinkerpop reference**

In the different GLV sections:

> The following options are allowed on a per-request basis in this fashion: batchSize, requestId, userAgent and evaluationTimeout (formerly scriptEvaluationTimeout which is also supported but now deprecated).

Which suggests that `scriptEvaluationTimeout` should still work although is not recommended to use it.

jroimartin:
> Normally I'd suggest a CHANGELOG entry or even upgrade docs, but since the Go GLV isn't officially released yet, we can forego that as you've basically just added a feature that should have been present anyway from the start.

I've just checked the docs and it is already there! :smile: 

https://github.com/apache/tinkerpop/blob/24384fb9932e028c753ae73c0fa7817665e80713/docs/src/reference/gremlin-variants.asciidoc?plain=1#L1457-L1466

gremlin-go/driver/request.go
spmallette: `scriptEvaluationTimeout` was technically removed in 3.5.0: https://issues.apache.org/jira/browse/TINKERPOP-2295 if there are remnants of it still around, that is probably a mistake. there are some graphs that quietly still support it in 3.5.x, but i don't think TinkerPop needs to continue to quietly do the same. and gremlin-go is new so why bother to allow folks to write code that uses deprecated/removed things. Just seems like we'd be adding things to forget to remove later 🙂

jroimartin: That makes sense. I removed `scriptEvaluationTimeout` both from the gremlin-go docs and code. I didn't touch any other GLV because it feels like it should go in a different PR.

eb9b358 should fix it. PTAL :slightly_smiling_face: 

Thanks!

jroimartin:
In the meantime I also renamed `extractWithReqArgs` to `extractWithReqArg` given that it returns one single request argument.

gremlin-go/driver/connection_test.go
```diff
@@ -1121,4 +1121,26 @@ func TestConnection(t *testing.T) {
...
+
+		_, err = g.With("evaluationTimeout", 10000).With("evaluationTimeout", 10).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Next()
+		assert.NotNil(t, err)
+
+		_, err = g.WithStrategies(OptionsStrategy(map[string]interface{}{"evaluationTimeout": 10})).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Next()
+		assert.NotNil(t, err)
+
+		_, err = g.WithStrategies(OptionsStrategy(map[string]interface{}{"evaluationTimeout": 10000})).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Next()
+		assert.Nil(t, err)
+	})
```
lyndonbauto, line 1145: Does this test take 5000 ms * 5 seconds to execute? If so can we possible use `Iterate()` and then check the error promise to run the requests async?

jroimartin: I pushed 866364c. Is that what you mean?

However, now the test is flaky due to what I think is an unrelated issue. The gremlin-server fails sometimes with:

```
gremlin-test-server             | [ERROR] GremlinGroovyScriptEngine$GroovyCacheLoader - Script compilation FAILED gremlinscriptengine__ggremlinscriptengine__g.with.with("evaluationTimeout","evaluationTimeout",10000L10L))..injectinject((1L1L)).sideEffect(.sideEffect({Thread.sleep(5000)}).none(){Thread.sleep(5000)}).none() took 3ms {}
gremlin-test-server             | org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
gremlin-test-server             | Script4.groovy: 1: expecting ')', found '10L' @ line 1, column 102.
gremlin-test-server             |    ut","evaluationTimeout",10000L10L))..inj
gremlin-test-server             |                                  ^
gremlin-test-server             |
gremlin-test-server             | 1 error
gremlin-test-server             |
gremlin-test-server             |       at org.codehaus.groovy.control.ErrorCollector.failIfErrors(ErrorCollector.java:311)
gremlin-test-server             |       at org.codehaus.groovy.control.ErrorCollector.addFatalError(ErrorCollector.java:151)
gremlin-test-server             |       at org.codehaus.groovy.control.ErrorCollector.addError(ErrorCollector.java:121)
gremlin-test-server             |       at org.codehaus.groovy.control.ErrorCollector.addError(ErrorCollector.java:133)
...
```

It seems to be trying to compile the following traversal: `gremlinscriptengine__ggremlinscriptengine__g.with.with("evaluationTimeout","evaluationTimeout",10000L10L))..injectinject((1L1L)).sideEffect(.sideEffect({Thread.sleep(5000)}).none(){Thread.sleep(5000)}).none()`, which is obviously wrong. I don't know enough about the code base, but it looks like a data race while building the query string.

I ran the tests against gremlin-server 3.5.3.

lyndonbauto: @jroimartin no that wasn't what I meant, I should have been more specific. 

Take a look at the [Iterate() function](https://pkg.go.dev/github.com/apache/tinkerpop/gremlin-go/v3/driver#Traversal.Iterate). This is typically used when someone is inserting data, it ignores the return value and runs in the background, providing a channel with an error instead of blocking and returning an error.

It is use **IN PLACE** of `Next()`

How this should look in code (untested, so might need minor edit to compile):
```
, err := g.With("evaluationTimeout", 10).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Next()
		assert.NotNil(t, err)

		_, err = g.With("evaluationTimeout", 10000).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Next()
		assert.Nil(t, err)
```

could instead be something like:
```
err1 := g.With("evaluationTimeout", 10).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Iterate()

err2 := g.With("evaluationTimeout", 10000).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Iterate()
....

assert.Nil(t, <-err1)
assert.Nil(t, <-err2)
....

```

jroimartin: @lyndonbauto sorry, but isn't this essentially the same that I did in 866364ce328c7a91bac62366815224f30c0f9601?

I did it table-driven because this way it is easier to add new test cases (you just have to add a new entry to the table) and everything related to a test case is grouped together (each table entry contains the description of the test case, the traversal and the expected error).

```go
reqArgsTests := []struct {
	msg       string
	traversal *GraphTraversal
	nilErr    bool
}{
	{
		"Traversal must time out (With)",
		g.With("evaluationTimeout", 10).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}),
		false,
	},
	{
		"Traversal must finish (With)",
		g.With("evaluationTimeout", 10000).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}),
		true,
	},
	{
		"evaluationTimeout is overridden and traversal must time out (With)",
		g.With("evaluationTimeout", 10000).With("evaluationTimeout", 10).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}),
		false,
	},
	{
		"Traversal must time out (OptionsStrategy)",
		g.WithStrategies(OptionsStrategy(map[string]interface{}{"evaluationTimeout": 10})).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}),
		false,
	},
	{
		"Traversal must finish (OptionsStrategy)",
		g.WithStrategies(OptionsStrategy(map[string]interface{}{"evaluationTimeout": 10000})).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}),
		true,
	},
}
```

Of course, if you prefer, I can change it to what you proposed:

```go
err1 := g.With("evaluationTimeout", 10).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Iterate()
err2 := g.With("evaluationTimeout", 10000).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Iterate()
err3 := g.With("evaluationTimeout", 10000).With("evaluationTimeout", 10).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Iterate()
err4 := g.WithStrategies(OptionsStrategy(map[string]interface{}{"evaluationTimeout": 10})).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Itearte()
err5 := g.WithStrategies(OptionsStrategy(map[string]interface{}{"evaluationTimeout": 10000})).Inject(1).SideEffect(&Lambda{"Thread.sleep(5000)", "gremlin-groovy"}).Iterate()

assert.NotNil(t, <-err1)
assert.Nil(t, <-err2)
assert.NotNil(t, <-err3)
assert.NotNil(t, <-err4)
assert.Nil(t, <-err5)
```

Which is shorter, but then the asserts are not next to the test cases, which IMHO make it harder to read and maintain. For instance when you want to add a new test case in the middle of the existing ones.

What do you think? Of course, I'm more than happy to change it if that helps to follow the coding guidelines of the project of if I missed something.

Regarding the gremlin-server error. Am I doing anything wrong? Should I send a bug report?

lyndonbauto: Oh sorry  I missed the iterate calls, my bad! I am not sure about the gremlin-server error tbh. @spmallette have you seen anything like this before? odd that it is inconsistent. I wonder if we can reproduce it with another glv.

jroimartin: @lyndonbauto I've been able to reproduce it using gremlin-python with the following set-up (it's a quick hack):

**repro.py**

```python
#!/usr/bin/env python3

from gremlin_python.driver.driver_remote_connection import DriverRemoteConnection
from gremlin_python.process.anonymous_traversal import traversal

g = traversal().with_remote(DriverRemoteConnection('ws://localhost:45940/gremlin','g'))

try:
    g.inject(1).sideEffect(lambda: "Thread.sleep(1000)").iterate()
except Exception as e:
    print(e)
```

**stress.bash**

```bash
#!/bin/bash

for i in $(seq 0 250); do
	python3 repro.py &
	pids[${i}]=$!
done

wait ${pids[*]}
```

I'm running gremlin-server this way (from the gremlin-go directory):

```
$ GREMLIN_SERVER=3.5.4-SNAPSHOT docker-compose run --rm --service-ports gremlin-test-server
```

The reproducer:

```
$ ./stress.bash

Received error message '{'requestId': '0a358421-df3c-4965-8535-bf66337289c8', 'status': {'message': 'startup failed:\nScript4.groovy: 1: unexpected token: 1 @ line 1, column 44.\n1 error\n', 'code': 597, 'attributes': {'stackTrace': 'org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:\nScript4.groovy: 1: unexpected token: 1 @ line 1, column 44.\n   engine__g.inject((int) 1(int) 1).sideEff\n                                 ^\n\n1 error\n\n\tat org.codehaus.groovy.control.ErrorCollector.failIfErrors(ErrorCollector.java:311)\n\tat org.codehaus.groovy.control.ErrorCollector.addFatalError(ErrorCollector.java:151)\n\tat org.codehaus.groovy.control.ErrorCollector.addError(ErrorCollector.java:121)\n\tat org.codehaus.groovy.control.ErrorCollector.addError(ErrorCollector.java:133)\n\tat org.codehaus.groovy.control.So
...
```

In the gremlin-server logs:

```
[ERROR] ? - Script compilation FAILED gremlinscriptengine__g.inject((int) 1(int) 1).sideEffect().sideEffect({Thread.sleep(1000)}{Thread.sleep(1000)}))..nonenone()() took 11ms {}
org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
Script4.groovy: 1: unexpected token: 1 @ line 1, column 44.
   engine__g.inject((int) 1(int) 1).sideEff
                                 ^

1 error

        at org.codehaus.groovy.control.ErrorCollector.failIfErrors(ErrorCollector.java:311)
        at org.codehaus.groovy.control.ErrorCollector.addFatalError(ErrorCollector.java:151)
        at org.codehaus.groovy.control.ErrorCollector.addError(ErrorCollector.java:121)
        at org.codehaus.groovy.control.ErrorCollector.addError(ErrorCollector.java:133)
        at org.codehaus.groovy.control.SourceUnit.addError(SourceUnit.java:325)
...
```

xiazcy: Hi @jroimartin, just to make sure it's not due to the gremlin-go test server set up. Have you tried running the python script against the built gremlin server it self, instead of through docker compose? This would be running the file `gremlin-server/target/apache-tinkerpop-gremlin-server-3.5.4-SNAPSHOT-standalone/bin/gremlin-server.sh`, just starting the default server should be enough for the inject. Note the port will be different, `8182` instead of `45940`.

jroimartin: @xiazcy very good point. Actually, I have **not** been able to reproduce it with a bare gremlin-server. Specifically, I tried:

```
$ docker run --rm -ti -p 8182:8182 tinkerpop/gremlin-server:3.5.4-SNAPSHOT
```

and

```
$ docker run --rm -ti -p 8182:8182 tinkerpop/gremlin-server:3.5.3
```

jroimartin: I've been comparing the configuration used by the gremlin-go test server and a regular gremlin-server and I think I managed to reproduce the issue with the regular gremlin-server.

In the generated docker container (`tinkerpop/gremlin-server:3.5.4-SNAPSHOT`) I modified `/opt/gremlin-server/conf/gremlin-server.yaml` in the following way:

```diff
--- gremlin-server.yaml 2022-06-22 11:30:53.452370098 +0200
+++ gremlin-server-repro.yaml   2022-06-22 11:27:50.000000000 +0200
@@ -18,7 +18,7 @@
 host: 172.17.0.2
 port: 8182
 evaluationTimeout: 30000
-channelizer: org.apache.tinkerpop.gremlin.server.channel.WebSocketChannelizer
+channelizer: org.apache.tinkerpop.gremlin.server.channel.UnifiedChannelizer
 graphs: {
   graph: conf/tinkergraph-empty.properties}
 scriptEngines: {
```

Again I'm not familiar with the code base of the project, but I hope this can serve as a hint for you.

BTW given that this looks like a race condition and it is not always easy to reproduce them in a reliable way, take this information with a grain of salt. I don't want to waste people time with inaccurate information :slightly_smiling_face:

jroimartin: Full backtrace of one of the errors:

```
org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
Script4.groovy: 1: unexpected token: . @ line 1, column 53.
   .inject((int) 1).sideE.inject(.inject((i
                                 ^

1 error

        at org.codehaus.groovy.control.ErrorCollector.failIfErrors(ErrorCollector.java:311)
        at org.codehaus.groovy.control.ErrorCollector.addFatalError(ErrorCollector.java:151)
        at org.codehaus.groovy.control.ErrorCollector.addError(ErrorCollector.java:121)
        at org.codehaus.groovy.control.ErrorCollector.addError(ErrorCollector.java:133)
        at org.codehaus.groovy.control.SourceUnit.addError(SourceUnit.java:325)
        at org.codehaus.groovy.antlr.AntlrParserPlugin.transformCSTIntoAST(AntlrParserPlugin.java:224)
        at org.codehaus.groovy.antlr.AntlrParserPlugin.parseCST(AntlrParserPlugin.java:192)
        at org.codehaus.groovy.control.SourceUnit.parse(SourceUnit.java:226)
        at org.codehaus.groovy.control.CompilationUnit$1.call(CompilationUnit.java:201)
        at org.codehaus.groovy.control.CompilationUnit.applyToSourceUnits(CompilationUnit.java:965)
        at org.codehaus.groovy.control.CompilationUnit.doPhaseOperation(CompilationUnit.java:642)
        at org.codehaus.groovy.control.CompilationUnit.processPhaseOperations(CompilationUnit.java:618)
        at org.codehaus.groovy.control.CompilationUnit.compile(CompilationUnit.java:595)
        at groovy.lang.GroovyClassLoader.doParseClass(GroovyClassLoader.java:401)
        at groovy.lang.GroovyClassLoader.access$300(GroovyClassLoader.java:89)
        at groovy.lang.GroovyClassLoader$5.provide(GroovyClassLoader.java:341)
        at groovy.lang.GroovyClassLoader$5.provide(GroovyClassLoader.java:338)
        at org.codehaus.groovy.runtime.memoize.ConcurrentCommonCache.getAndPut(ConcurrentCommonCache.java:147)
        at groovy.lang.GroovyClassLoader.parseClass(GroovyClassLoader.java:336)
        at groovy.lang.GroovyClassLoader.parseClass(GroovyClassLoader.java:320)
        at groovy.lang.GroovyClassLoader.parseClass(GroovyClassLoader.java:262)
        at org.apache.tinkerpop.gremlin.groovy.jsr223.GremlinGroovyScriptEngine$GroovyCacheLoader.lambda$load$0(GremlinGroovyScriptEngine.java:821)
        at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1700)
        at java.base/java.util.concurrent.CompletableFuture.asyncSupplyStage(CompletableFuture.java:1714)
        at java.base/java.util.concurrent.CompletableFuture.supplyAsync(CompletableFuture.java:1931)
        at org.apache.tinkerpop.gremlin.groovy.jsr223.GremlinGroovyScriptEngine$GroovyCacheLoader.load(GremlinGroovyScriptEngine.java:819)
        at org.apache.tinkerpop.gremlin.groovy.jsr223.GremlinGroovyScriptEngine$GroovyCacheLoader.load(GremlinGroovyScriptEngine.java:814)
        at com.github.benmanes.caffeine.cache.BoundedLocalCache$BoundedLocalLoadingCache.lambda$new$0(BoundedLocalCache.java:3117)
        at com.github.benmanes.caffeine.cache.LocalCache.lambda$statsAware$0(LocalCache.java:144)
        at com.github.benmanes.caffeine.cache.BoundedLocalCache.lambda$doComputeIfAbsent$16(BoundedLocalCache.java:1968)
        at java.base/java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1908)
        at com.github.benmanes.caffeine.cache.BoundedLocalCache.doComputeIfAbsent(BoundedLocalCache.java:1966)
        at com.github.benmanes.caffeine.cache.BoundedLocalCache.computeIfAbsent(BoundedLocalCache.java:1949)
        at com.github.benmanes.caffeine.cache.LocalCache.computeIfAbsent(LocalCache.java:113)
        at com.github.benmanes.caffeine.cache.LocalLoadingCache.get(LocalLoadingCache.java:67)
        at org.apache.tinkerpop.gremlin.groovy.jsr223.GremlinGroovyScriptEngine.getScriptClass(GremlinGroovyScriptEngine.java:569)
        at org.apache.tinkerpop.gremlin.groovy.jsr223.GremlinGroovyScriptEngine.eval(GremlinGroovyScriptEngine.java:376)
        at java.scripting/javax.script.AbstractScriptEngine.eval(AbstractScriptEngine.java:233)
        at org.apache.tinkerpop.gremlin.groovy.jsr223.GremlinGroovyScriptEngine.eval(GremlinGroovyScriptEngine.java:298)
        at org.apache.tinkerpop.gremlin.server.handler.AbstractSession.fromBytecode(AbstractSession.java:443)
        at org.apache.tinkerpop.gremlin.server.handler.AbstractSession.process(AbstractSession.java:264)
        at org.apache.tinkerpop.gremlin.server.handler.SingleTaskSession.run(SingleTaskSession.java:70)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.lang.Thread.run(Thread.java:829)
```

xiazcy: @jroimartin Actually yes, `WsAndHttpChannelizer` would work, and since ultimately we just need HTTP requests enabled, we may not even need to switch back to `UnifiedChannelizer`. Could you make that change as well with your PR? Thank you!

xiazcy:
LGTM.
@lyndonbauto would you have any additional comments? Thanks!
Also I've opened a ticket to track the UnifiedChannelizer issue: https://issues.apache.org/jira/browse/TINKERPOP-2765. It will be looked at separately since moving forward it shouldn't affect gremlin-go with our changed to `WsAndHttpChannelizer`.

gremlin-go/driver/connection_test.go
xiazcy: Thank you!

lyndonbauto:
> LGTM. @lyndonbauto would you have any additional comments? Thanks! Also I've opened a ticket to track the UnifiedChannelizer issue: https://issues.apache.org/jira/browse/TINKERPOP-2765. It will be looked at separately since moving forward it shouldn't affect gremlin-go with our changed to `WsAndHttpChannelizer`.

I am good.
VOTE+1

PR #1709, opened 2022-06-16
TINKERPOP-2617 Clean up Java connection pool logic

simonz-bq:
Removed duplicated code for fetching a connection form the ConnectionPool.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
@@ -142,8 +142,7 @@ final class ConnectionPool {
 
         if (isClosed()) throw new ConnectionException(host.getHostUri(), host.getAddress(), "Pool is shutdown");
 
-        final Connection leastUsedConn = selectLeastUsed();
-
+        // Initialize connections if the connection pool is empty
         if (connections.isEmpty()) {
             logger.debug("Tried to borrow connection but the pool was empty for {} - scheduling pool creation and waiting for connection", host);
             for (int i = 0; i < minPoolSize; i++) {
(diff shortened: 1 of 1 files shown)


lyndonbauto:
@divijvaidya do you have time to look at this? I think you have the most context on Java connection pooling.

divijvaidya:
Let me get to a first review by end of this week @simonz-bq . Meanwhile, it would be great if we could get some testing around the edge cases that you mentioned which aren't correctly handled in the current code.

simonz-bq:
> Let me get to a first review by end of this week @simonz-bq . Meanwhile, it would be great if we could get some testing around the edge cases that you mentioned which aren't correctly handled in the current code.

@divijvaidya the additional edge cases logic was removed, but I forgot to update the initial commit comment. It was causing existing integration tests to fail and hang and investigation into how to fixing them proved to be difficult and complex for what was being considered as a easy win.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
```diff
@@ -382,35 +344,49 @@ private Connection waitForConnection(final long timeout, final TimeUnit unit) th
...
 
-            final Connection leastUsed = selectLeastUsed();
             if (leastUsed != null) {
                 while (true) {
                     final int inFlight = leastUsed.borrowed.get();
                     final int availableInProcess = leastUsed.availableInProcess();
-                    if (inFlight >= availableInProcess) {
-                        logger.debug("Least used {} on {} has requests borrowed [{}] >= availableInProcess [{}] - may timeout waiting for connection",
-                                leastUsed, host, inFlight, availableInProcess);
+                    if (inFlight >= maxSimultaneousUsagePerConnection && availableInProcess == 0) {
```
divijvaidya, line 367: We currently have a conditional check for whether the selected connection can accept this new request at two places at line 351 and line 367. But the current code has different conditions for checking the same logic condition. We should unify that.

This conditional should be abstracted into a function or a functional interface, let's say, `canAcceptNewRequestsOnConnection()` and re-use the same conditional check on line 351 
`if (!canAcceptNewRequestsOnConnection() && currentPoolSize < maxPoolSize)`.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
```diff
@@ -382,35 +344,49 @@ private Connection waitForConnection(final long timeout, final TimeUnit unit) th
...
         long remaining = timeout;
         long to = timeout;
-        do {
-            try {
-                awaitAvailableConnection(remaining, unit);
-            } catch (InterruptedException e) {
-                Thread.currentThread().interrupt();
-                to = 0;
+
+        Connection leastUsed = selectLeastUsed();
```
divijvaidya, line 348: Perhaps refactor as following for returning the happy case early. It helps in readability since the reader now understands that the code after this line is handling logic for case when no connection is immediately available.

You can do something like:
```
// scenario 1: when connection pool is empty
if (connections.empty()) {
    for (size 1..minPoolSize)
        createNewConnection()
    return waitForConnection()
}

// scenario 2: happy case when a connection is available
Connection leastUsed = selectLeastUsed();
if (leastUsed != null && leastUsed.canAcceptNewRequests)
    return leastUsed

// scenario 3: wait for a connection to become available
maybeCreateNewConnection() // checks if it is possible to create a new connection, if yes, creates one.
return waitForConnection() // doesn't need to consider the older leastUsed, it will wait first and then ask another.

fun waitForConnection {
  do {
      awaitForAvailableConnection
      find least used
      if (leastUsed.canAcceptNewRequests)
            return leastUsed
  } while (timeout pending)
}
```

divijvaidya (changes requested):
Thank you for making this change. It was a long pending change which would solve problems with client connectivity.

Since we are changing some inherent logic on how a connection is selected (by modifying the conditions for when a connection is eligible for taking request), it's important to back this change with tests. I would suggest the testing the following scenarios:

Scenario 1:
Connection pool is empty
Wait till connect pool size reaches minConnPoolSize
Assert that no new connections are scheduled for creation
Request should get one of the new connections created

Scenario 1.1:
Connection pool is empty
Do not wait till connect pool size reaches minConnPoolSize
Assert that 1 new connection is scheduled for creation
Request should get one of the new connections created

Scenario 2:
ConnectionPool is at max size
All connections have reached max borrowed count
Assert that no new connection is scheduled for creation
One connection decreases borrowed count and in process count
Assert that this one connection gets to accept this request.

Scenario 2.1:
ConnectionPool is at not at max size
All connections have reached max borrowed count
Assert that new connection is scheduled for creation
No existing connection reduces borrowed count
New connection finished creation
Assert that this new connection gets to accept this request.

Scenario 2.2:
ConnectionPool is at not at max size
All connections have reached max borrowed count
Assert that new connection is scheduled for creation
One connection decreases borrowed count and in process count
New connection finished creation
Assert that this new connection gets to accept this request (because it's least used)

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
simonz-bq: I am confused as to what is being proposed here.

This is more or less the way the existing logic is written. The ticket was created with the acceptance criteria to avoid the duplication of the checking in scenario 2 and 3, where now it is combined into a single point of checking in `waitForConnection`. 

Whilst I agree that it is now less readable, unfortunately this is the result. 

I will add some comments to make it more clear what is happening.

simonz-bq:
Regarding the ask for more tests, the logic did not change. This is a refactor and the functionality behaves similar to before and is ultimately just for the sake of reducing code duplication.

Originally I had made some minor logic adjustments that were supposed to improve the logic, but it lead to nebulous failures with existing tests, and how to fix them or write new ones is very esoteric. Investigating how to fix them exceeded a time box that we'd expected for what was supposed to be a simple code refactor.

More work can definitely be done regarding the connection pooling, but I think doing so here would exceed the scope of the ticket and what we have here is sufficient for the ask at hand.

xiazcy:
Been open for a while with not much activity, we are closing this PR.

lyndonbauto:
@DKZed Where are you at with these changes?

PR #1714, opened 2022-06-16
TINKERPOP-2751: Throw exception if transaction attempted on non transaction supported Graph

L0Lmaker:
Link to issue: [TINKERPOP 2751](https://issues.apache.org/jira/browse/TINKERPOP-2751)

If a transaction is attempted on a graph that does not support transactions, the client hangs as it waits for a response from the server. The change in this PR handles that case and responds to the client with a `transactionNotSupported()` exception.

Ideally, we would want the client to fail at `tx.begin()` but this would involve a much bigger change on the server and across the GLVs. For now, users should be notified that they need to use a transaction-enabled graph when attempting to perform a graph operation (rollback or commit)

### DIFF
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/session/SessionOpProcessor.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/session/SessionOpProcessor.java
@@ -526,6 +526,11 @@ public class SessionOpProcessor extends AbstractEvalOpProcessor {
                 throw new IllegalStateException(String.format(
                         "Bytecode in request is not a recognized graph operation: %s", bytecode.toString()));
             }
+        } else {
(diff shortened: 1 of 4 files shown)


gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/session/SessionOpProcessor.java
```diff
@@ -526,6 +526,11 @@ protected void handleGraphOperation(final Bytecode bytecode, final Graph graph,
                 throw new IllegalStateException(String.format(
                         "Bytecode in request is not a recognized graph operation: %s", bytecode.toString()));
             }
+        } else {
+            UnsupportedOperationException uoe = Graph.Exceptions.transactionsNotSupported();
```
spmallette, line 530: nit: final

lyndonbauto:
Nice - I remember briefly looking for the culprit on this issue. VOTE+1

PR #1716, opened 2022-06-17
TINKERPOP-2761: fix version key race in Manifest

lionelfleury:
- use a more specific key to avoid race on version
- see: https://github.com/JanusGraph/janusgraph/discussions/3004

Fixes https://issues.apache.org/jira/browse/TINKERPOP-2761


spmallette:
thanks for contributing, but i'm not sure i understand the discussion that was referenced and thus not sure i understand why the manifest version needs a special `tinkerpopVersion`. could you please explain further what you mean by a "race on version"?

spmallette:
> In the mentioned thread, it seems to read the version key from hbase jar

sorry - i read that thread but github didn't display the discussion in whole and i missed some additional comments

as i recall there isn't any standard to manifest files keys. rather than camel case of `tinkerpopVersion` could you please change it to `tinkerpop-version` ?

PR #1724, opened 2022-06-21
gremlin-go: improve docker-compose compatibility, fix run.sh command line parsing

jroimartin:
The current `docker-compose.yml` file is not compatible with the docker-compose version packaged by some widely used Linux distributions. For instance, Debian bullseye (current stable) [packages docker-compose 1.25.0](https://packages.debian.org/bullseye/docker-compose).

docker-compose 1.25.0 [supports up to compose file format 3.7](https://github.com/docker/compose/releases/tag/1.25.0) and requires the version to be explicitly set in the `docker-compose.yml` file.

By setting the version of the `docker-compose.yml` file to 3.7 we support docker-compose 1.22.0+ and the only required change is adding the "version" field. Also, [version 3 has not been deprecated](https://docs.docker.com/compose/compose-file/compose-versioning/).

Finally, `run.sh` accepts a gremlin-server version as argument, however it is not used when running tests. This PR also fixes this.

I have tested this PR both with docker-compose 1.25.0 and 2.6.0 (latest docker-ce version).

### DIFF
--- a/gremlin-go/docker-compose.yml
+++ b/gremlin-go/docker-compose.yml
@@ -15,6 +15,8 @@
 #    specific language governing permissions and limitations
 #    under the License.
 
+version: "3.7"
+
 services:
 
   gremlin-test-server:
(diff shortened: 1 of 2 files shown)


xiazcy (commented):
LGTM. Agree for a version with better compatibility, and the export on the run script is much cleaner. Thanks!

PR #1725, opened 2022-06-21
Fix a javadoc comment in Cluster.Builder regarding maxInProcessPerCon…

rdtr:
Just fixed a comment. No unit test should be required.


mikepersonick:
shipit +1

PR #1727, opened 2022-06-21
Release msg in finally block on HttpGremlinEndpointHandler

rdtr:
Just to make sure `msg` is released at the end instead of adding release logic everywhere.


kenhuuu:
Hi @rdtr , were you planning on addressing the review comment from above about adding a unit test? Some of the changes you are requesting will be updated in master with the move to HTTP-only, but those changes won't exist in 3.6.x/3.7.x. Is this change something you would like to see go into 3.6.x/3.7.x still?

PR #1732, opened 2022-06-22
fixed null exception

nikita15p:
Fixed "Caused by: java.lang.NullPonterException"
Signed-off-by: nikita15p <nikita15p@gmail.com>

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/search/path/ShortestPathVertexProgram.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/search/path/ShortestPathVertexProgram.java
@@ -452,11 +452,14 @@ public class ShortestPathVertexProgram implements VertexProgram<Triplet<Path, Ed
                 otherV = edge.outVertex();
 
             // only send message if the adjacent vertex is not yet part of the current path
-            if (!currentPath.objects().contains(otherV)) {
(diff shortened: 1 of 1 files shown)


divijvaidya:
Thank you for the code change @nikita15p. Could we add a unit test here which fails prior to this change and passes after? It would help us catch any regression in future.

nikita15p:
@divijvaidya
Same condition `currentPath != null` is already available at line 408, since we are accessing objects() method.                                  
Is there any test case for that? Basically this is the obvious fix which I am recommending since I came across exception.

spmallette:
i think the check at 408 is part of the design for `extendPath()` where `null` is expected. i think @divijvaidya is questioning the circumstances around the `currentPath` being `null` in `processEdges()` where it doesn't seem clear how `currentPath` would end up that way. understanding "how" would better explain what is happening there and if simply ignoring the subsequent code path is the right fix.

lyndonbauto:
@nikita15p bump - are you able to make a unit test for this situation? Would be good to know how this is happening.

nikita15p:
I faced this problem in my code and have fixed it locally. Will write a UT
and update the PR. Couldn't work on this till now.

Thanks and Regards,
Nikita

xiazcy:
Thank you for raising the PR. We've decided to close this one as there has not been any activity in the past 3 months.

Please feel free to open another PR to fix this if you are still interested, providing an explanation on how the current path leads to the exception and adding a unit test.

PR #1733, opened 2022-06-22
gremlin-python: enable opt-out for some pkg reqs

skieffer:
For `gremlin-python` users who plan to use a custom transport, there may be no need to install the `aiohttp` or `nest_asyncio` packages. These packages bring quite a few recursive requirements with them as well.

These packages should _not_ be moved to `extras_require` in `setup.py`, since this would break existing deployments.

Instead, there is a pattern for opting out of requirements at installation time, using environment variables. It is discussed for example [here](https://github.com/aio-libs/aioredis-py/issues/663#issuecomment-555555554).

EDIT: Associated JIRA ticket: <https://issues.apache.org/jira/browse/TINKERPOP-2752>

### DIFF
--- a/gremlin-python/src/main/python/setup.py
+++ b/gremlin-python/src/main/python/setup.py
@@ -45,11 +45,21 @@ from gremlin_python import __version__
 version = __version__.version
 
 install_requires = [
-    'nest_asyncio',
-    'aiohttp>=3.8.0,<=3.8.1',
     'aenum>=1.4.5,<4.0.0',
     'isodate>=0.6.0,<1.0.0'
 ]
+
+# Users who plan to use a custom transport (instead of AiohttpTransport) can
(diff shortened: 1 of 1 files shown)


skieffer:
Hi @krlawrence, thanks for taking a look at this.

There are no dependency conflicts. For me the concern is partly about wasted MBs in a docker image (only about 12MB, but it counts), and partly about wanting to keep misleading clutter out of automatically generated licensing info. I don't want to say, "This software uses packages X, Y, and Z" if it doesn't actually use them.

In any case, I realize the problem is really not with `gremlinpython` per se, but is fundamentally about needing better control over pip and dependency resolution. Therefore I'm trying to tackle the more general problem over at `pip-tools`: https://github.com/jazzband/pip-tools/pull/1645

I think the latter would actually be my preferred solution, so, in the meantime, I guess I won't push too hard for this PR.

PR #1736, opened 2022-06-24
TINKERPOP-2754 Fix indefinite hanging on JS client on server restart

simonz-bq:
The JS implementation of the websocket resolves traversal promises by waiting for either "error" (reject) or "message" (resolve or reject depending on the contained data). However, upon the connection terminating (server restart), the "close" event is fired instead, so therefore we must let all waiting response handlers on the websocket know to reject the waiting promises.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -35,6 +35,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Fixed bug in `PartitionStrategy` where the use of `AbstractLambdaTraversal` caused an unexpected exception.
 * Fixed bug where close requests for sessions were improperly validating the request in the `UnifiedChannelizer`.
 * Deprecated and removed functionality of the `connectOnStartup` option in `gremlin-javascript` to resolve potential `unhandledRejection` and race conditions.
+* Fixed a bug where the JavaScript client would hang indefinitely on traversals if the connection to the server was terminated.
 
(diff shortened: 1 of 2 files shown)


divijvaidya:
Thank you for the PR. Could we add a unit test which tests the driver behaviour on abrupt close of server and validates that the traversals have been correct closed? It would help us catch regressions related to close behaviour in the future.

simonz-bq:
> Thank you for the PR. Could we add a unit test which tests the driver behaviour on abrupt close of server and validates that the traversals have been correct closed? It would help us catch regressions related to close behaviour in the future.

Tests have been added.

gremlin-javascript/src/main/javascript/gremlin-javascript/test/integration/client-tests.js
```diff
@@ -107,5 +107,48 @@ describe('Client', function () {
...
+
+      await closingClient.close();
+      for await (const result of readable) {
+        // Consume the stream
+      }
+
+      assert.strictEqual(isRejected, true);
+    });
   });
 });
```
lyndonbauto, line 154:
```suggestion
});

```

lyndonbauto:
Nice, thanks for adding this! 

VOTE+1

divijvaidya (approved):
LGTM! Vote +1

Since we don't have 3 votes on this PR yet, we are going to wait for a 7 day window before merging this change.

PR #1738, opened 2022-06-24
Timeout updates for Gremlin-JS for GitHub Actions workflow failures

L0Lmaker:
Gremlin JS Windows GitHub Actions workflow would often fail due to timeout issues. This PR increases the timeout for Integration and Cucumber tests and also the Actions Workflow to enable the process to have enough time to run.


divijvaidya:
Do we know which tests are flaky wrt the time it takes to execute them? Increasing the timeout might be a superficial fix when the root cause are problem such as https://github.com/apache/tinkerpop/pull/1736

lyndonbauto:
I second Divij's comment.

If we have tests that are expected to take 5 mins and we are timing out in 5 mins (for example) then I am okay with increasing timeout. But if we are increasing timeout because we have inconsistent behaviour we should root cause that.

lyndonbauto:
Code changes LGTM otherwise.

PR #1739, opened 2022-06-27
Improved pathing changes made for Windows Build Support

L0Lmaker:
Pathing changes introduced to support the windows build of `gremlin-javascript`, `gremlin-dotnet`, and `gremlint` would make it so that building Tinkerpop or its modules from anywhere other than the Tinkerpop root would not work. 

The changes in this PR would make it so that maven commands such as `install`, `verify`, `compile` and so on would work as expected from outside the root.

### DIFF
--- a/gremlin-dotnet/src/pom.xml
+++ b/gremlin-dotnet/src/pom.xml
@@ -56,14 +56,14 @@ limitations under the License.
                             <scripts>
                                 <script>
                                     def mavenVersion = "${project.version}"
-                                    def platformAgnosticBaseDirPath = new File(".this-definitely-does-not-exist").absolutePath.replace("\\.this-definitely-does-not-exist", "").replace("/.this-definitely-does-not-exist", "").replace("\\", "/")
-                                    def file = new File(platformAgnosticBaseDirPath + "/gremlin-dotnet/src/Gremlin.Net/Gremlin.Net.csproj")
+                                    def platformAgnosticBaseDirPath = project.basedir.getAbsolutePath().replace("\\", "/")
+                                    def file = new File(platformAgnosticBaseDirPath + "/Gremlin.Net/Gremlin.Net.csproj")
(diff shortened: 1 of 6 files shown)


lyndonbauto:
LGTM
VOTE+1

L0Lmaker:
> I think this is definitely already a good improvement and most of the tests seem to pass this way, but I still run into problems with the Gherkin tests in `gremlin-dotnet` when executed from that directory:
> 
> ```shell
> mvn verify -pl :gremlin-dotnet,:gremlin-dotnet-tests -P gremlin-dotnet -DincludeNeo4j
> ```
> 
> leads to errors such as:
> 
> ```
> java.lang.IllegalStateException: data/tinkerpop-modern.xml does not exist
> ```
> 
> so it looks like more paths need to be changed to get full support for building and testing from other paths than the project's root.

I am able to reproduce the same error. I checked with an older version of the code without the windows build support changes in commit 38a3bd7 on `3.5` and the same error exists there as well. 

Should I document this in [TINKERPOP-2760](https://issues.apache.org/jira/browse/TINKERPOP-2760), which I made for pending windows tasks or create a new ticket for it?

L0Lmaker:
> Could you however please squash your commits? Would be better to have only one commit here instead of six.

Squashed in commit 6aa0d21a3b9b9ed33825d838e38e182ce15d15f8

FlorianHockmann:
> I am able to reproduce the same error. I checked with an older version of the code without the windows build support changes in commit https://github.com/apache/tinkerpop/commit/38a3bd76e35a83fbb56e157f4874633920f436b5 on 3.5 and the same error exists there as well.
>
> Shall I document this in [TINKERPOP-2760](https://issues.apache.org/jira/browse/TINKERPOP-2760), which I made for pending windows tasks or create a new ticket for it?

If it's not specific to Windows, then I think that a separate ticket makes more sense.


VOTE +1

L0Lmaker:
> If it's not specific to Windows, then I think that a separate ticket makes more sense.

No, it is not specific to Windows, I will create a new ticket for it.

PR #1743, opened 2022-06-30
Cast certain traversal step inputs as int32 for serialization in Gremlin-Go

simonz-bq:
Certain traversal steps in Gremlin Server only take in `int` as arguments, and by default numbers passed through in Go, and therefore by extension Gremlin-Go, are passed in as `int`, which in Go are a larger data type and is different from `int32`, which is what matches Java's `int`. The default serialization for such numbers ends up being a `long`, which causes the server to return a failure due to a type mismatch. 

This change makes it so that when a user uses Gremlin-Go, they no longer have to enter numbers in certain steps as `Times(int32(10))` and instead can now just simply do `Times(10)`. It will now cast the input as `int32` which will then be serialized to a Java `int`, if the number itself is valid to fit in an `int32`. Otherwise, it will be rejected by the server, as expected.


gremlin-go/driver/graphTraversal.go
```diff
@@ -669,6 +670,30 @@ func (g *GraphTraversal) Write(args ...interface{}) *GraphTraversal {
...
+		case uint64:
+			args[i] = int32(val)
+		case int8:
+			args[i] = int32(val)
+		case int16:
+			args[i] = int32(val)
+		case int64:
+			args[i] = int32(val)
+		case int:
+			args[i] = int32(val)
```
jroimartin, line 689: `uint16` and `uint` are missing. Is it intended?

Other than that, I remember running into this problem when I started using gremlin-go and neither the solution nor the error were obvious. I ended up greping the repo to find a solution. So, in my opinion, this is a good change for people new to gremlin-go. However, I would update the `Times` documentation to mention that the passed value will be casted to `int32`.

gremlin-go/driver/graphTraversal.go
```diff
@@ -669,6 +670,30 @@ func (g *GraphTraversal) Write(args ...interface{}) *GraphTraversal {
...
 }
 
+func int32Args(args ...interface{}) []interface{} {
+	for i, arg := range args {
+		switch val := arg.(type) {
+		case uint8:
+			args[i] = int32(val)
+		case uint32:
+			args[i] = int32(val)
+		case uint64:
```
xiazcy, line 680: Should we include `uint64` and `int64`? Just in case someone enters a number larger than int32, casting would not give the correct result. I wonder if we should just propagates those types through, and let server error?

gremlin-go/driver/graphTraversal.go
```diff
@@ -587,7 +587,8 @@ func (g *GraphTraversal) TimeLimit(args ...interface{}) *GraphTraversal {
 
 // Times adds the times step to the GraphTraversal.
 func (g *GraphTraversal) Times(args ...interface{}) *GraphTraversal {
-	g.Bytecode.AddStep("times", args...)
+	// Gremlin server only accepts times step with integer argument values
+	g.Bytecode.AddStep("times", int32Args(args)...)
```
xiazcy, line 591: Current cucumber test suite still uses int32() cast for the Times arguments (example [here](https://github.com/lyndonbauto/tinkerpop/blob/7ada477e466479d5378d72a476df6dd2b8a78941/gremlin-go/driver/cucumber/gremlin.go#L63)), the `GolangTranslator.java` will need to be updated at this [line](https://github.com/lyndonbauto/tinkerpop/blob/7ada477e466479d5378d72a476df6dd2b8a78941/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/translator/GolangTranslator.java#L276) and `gremlin.go` file regenerated to reflect this change.

gremlin-go/driver/graphTraversal.go
```diff
@@ -669,6 +670,30 @@ func (g *GraphTraversal) Write(args ...interface{}) *GraphTraversal {
 	return g
 }
 
+func int32Args(args ...interface{}) []interface{} {
+	for i, arg := range args {
+		switch val := arg.(type) {
+		case uint8:
```
xiazcy, line 676: `uint8`, `int8`, `int16` are not needed. Times() accept numbers shorter than integers.

gremlin-go/driver/graphTraversal.go
xiazcy: `uint16` is serialized as an `int32` so that was intentional. Not sure about `uint`.

xiazcy (changes requested):
Now, there are other steps which also need int32 casting if the user enters `int` (examples can be found in `gremlin.go` by searching `int32(`), given we are updating this one, should we look into the others and address the same way?

gremlin-go/driver/graphTraversal.go
jroimartin: Then we might need to exclude also `int` and `uint`, right? I feel that would make this patch not very useful. If I'm not wrong, the common use case `.Times(123)` would not be covered anymore.

From the Go's spec:

```
There is also a set of predeclared integer types with implementation-specific sizes:

uint     either 32 or 64 bits
int      same size as uint
```

jroimartin:
@simonz-bq @xiazcy would it be an option to deal with this kind of casts inside the function [makeBytecodeRequest](https://github.com/apache/tinkerpop/blob/40508a427756d088d38b2236837c09db5dd1f2cd/gremlin-go/driver/request.go#L66)?

It could be modified to return `(request, error)` and, then `(*Client).submitBytecode` could [return with error if it fails](https://github.com/apache/tinkerpop/blob/40508a427756d088d38b2236837c09db5dd1f2cd/gremlin-go/driver/client.go#L161). That would allow to check that the cast is valid before sending it to the server.

I'm asking because it seems that the root cause of the problem is dealing with this inside the step itself which cannot return an error.

gremlin-go/driver/graphTraversal.go
xiazcy: Yeah, main reason we are running into the issue is we needed to serialize go int into java long since it could be 64 bits. I'm more in favour of leaving letting the error be returned by the server instead of casting, and perhaps adding documentations so users know what to do in those case.

xiazcy:
> `makeBytecodeRequest` could be modified to return `(request, error)` and `(*Client).submitBytecode` could [return with error if it fails](https://github.com/apache/tinkerpop/blob/40508a427756d088d38b2236837c09db5dd1f2cd/gremlin-go/driver/client.go#L161). That would allow to check that the conversion is valid before sending it to the server.
> 
@jroimartin That may work, and I suppose it would involve checking within a range of bytecode steps that needs casting and perform the cast? Although in that case would we want to add this extra check for every bytecode request we make?

jroimartin:
> I suppose it would involve checking within a range of bytecode steps that needs casting and perform the cast?

That's what I had in mind.

>  Although in that case would we want to add this extra check for every bytecode request we make?

I guess it's a trade-off:

- Updating the GLV docs only: users might be confused at first, but it would be documented and it does not add tricky nuances to the API or extra over-head.
- Implementing it inside the `Times` step: it would cover some of the most common use cases, but bugs due to truncated integers can be very tricky and even more confusing.
- Implementing it inside `makeBytecodeRequest`: adds an overhead to every request. Although bytecode is already traversed to extract per-request arguments, so this could be maybe "merged" somehow to traverse the bytecode only once.

xiazcy:
> * Implementing it inside the `Times` step: it would cover some of the most common use cases, but bugs due to truncated integers can be very tricky and even more confusing.

Yeah, while it is unlikely for users to enter very large numbers, we probably should handle the case, and I don't see how we can do that here when we can't return errors in a traversal step.

> * Implementing it inside `makeBytecodeRequest`: adds an overhead to every request. Although bytecode is already traversed to extract per-request arguments, so this could be maybe "merged" somehow to traverse the bytecode only once.

I would agree this may be the most ideal if we can minimize the overhead, to make it more natural for users without the need to cast.

gremlin-go/driver/graphTraversal.go
xiazcy: Also, the reason to suggest excluding `uint64` and `int64` is because I think we should enforce the fact that these steps only takes 32-bit integers or less, so we shouldn't allow users to consciously input types that is larger. 

The ambiguity comes with `int`, as that is the default integer type for go if one doesn't specify the type, so that would be the main case we need to handle. 

I'm not entirely sure how `uint` would be handled, since a 32-bit `uint` would technically fit to a long in java. Perhaps if we can figured out a good way to cast, this can be handled a similar way.

gremlin-go/driver/graphTraversal.go
simonz-bq: I don't think this matters too much. Passing in a int32 cast is still a valid input, and the cucumber tests are meant to test input-output, not the semantics of how it's entered into gremlin-go. There are other examples of inputs not being exactly as you'd expect a real user to pass things in.

gremlin-go/driver/graphTraversal.go
```diff
@@ -669,6 +670,38 @@ func (g *GraphTraversal) Write(args ...interface{}) *GraphTraversal {
 	return g
 }
 
+const maxInt32 = 2147483647
```
jroimartin, line 673: What about using `math.MaxInt32` instead?

gremlin-go/driver/graphTraversal.go
```diff
@@ -669,6 +670,38 @@ func (g *GraphTraversal) Write(args ...interface{}) *GraphTraversal {
...
+				args[i] = int32(val)
+			}
+		case int:
+			if val <= maxInt32 {
+				args[i] = int32(val)
+			}
+		case int64:
+			if val <= maxInt32 {
+				args[i] = int32(val)
+			}
```
jroimartin, line 697: Does it make sense to also check for `val >= math.MinInt32` in the case of `int` and `int64`? I guess the server would fail anyway, but being strict we could be truncating big negative numbers.

gremlin-go/driver/graphTraversal.go
```diff
@@ -669,6 +670,38 @@ func (g *GraphTraversal) Write(args ...interface{}) *GraphTraversal {
...
+		default:
+			break
```
jroimartin, line 699: I think the default case can be removed given that it is a no-op.

jroimartin (commented):
@simonz-bq so the idea with a0a6db8e5aa3dd68f2f61d8377fab7e81288c700 is to convert to `int32` only if possible and allow the server to fail otherwise, right?

I think it is a good approach. It fulfills all the mentioned requirements: it covers the most common use cases, it does not truncate the user data and it does not add unnecessary overhead on every request.

The only thing I see is that we could fail faster because at the time of the conversion check we already know that the server will fail, but I guess it is not a big deal?

xiazcy (approved):
LGTM. Thanks Simon!

jroimartin:
LGTM. Although, given my little knowledge of the project, mine are mostly suggestions and a way of learning more about the internals :slightly_smiling_face:

PR #1747, opened 2022-07-04
Add documentation for gremlin-python event loop issue workarounds.

kenhuuu:
Add documentation that explains how to workaround gremlin-python's need to use asyncio event loops for asynchronous applications that already have an event loop running. This issue has been reported multiple times (most recently in https://issues.apache.org/jira/browse/TINKERPOP-2755).


krlawrence:
Note that the Gremlin Python client already has this built in. You just need to provide a parameter when you create the transport.
```
connection = DriverRemoteConnection(endpoint,'g',
                 transport_factory=lambda:AiohttpTransport(call_from_event_loop=True))
```

PR #1748, opened 2022-07-05
Set an exception thrown instead of the root cause during the Client initialization

rdtr:
When we print an exception by logger, it is more helpful if we have full stack trace instead of just printing the exact root cause.


spmallette:
merged as 10f35d945c8c0f482b93f75e5dc39e2c4d4a9a78 starting at 3.5-dev

PR #1749, opened 2022-07-06
Gremlin-Go: Add int8, uint, and uint64 serialization

xiazcy:
Patching the go GraphBinary serializer with these 3 types. 

Serializing `int8` as Short, since there is no signed 8-bit integer type in GraphBinary I/O ([Byte](https://tinkerpop.apache.org/docs/3.6.0/dev/io/#_byte_3) is an unsigned 8-bit integer). Serializing both `uint` and `uint64` as BigIntegers, as they don't fit into Long.


vkagamlyk (approved):
LGTM

kenhuuu (commented):
LGTM

PR #1750, opened 2022-07-06
Gremlin-Python: added return in SingleByte and SingleChar & fixed if check bug

xiazcy:
Added missing return statements in SingleByte and SingleChar classes (e.g. `SingleByte(2)` will return `None` without the `return`), and fixed bug in SingleChar `if` check.


vkagamlyk (approved):
LGTM

kenhuuu (commented):
LGTM

PR #1753, opened 2022-07-12
Gremlin-Go: Fixed bug for GHA, increased server wait retries & fixed traversal int casting

xiazcy:
GHA was not capturing the exit code from `docker-compose up` because `docker-compose down` was executed separately. Now exiting upon the compose up exit code at the end of `run.sh`, and GHA should correctly capture failed golang tests.

Fixed the failing golang test related to recent [PR](https://github.com/apache/tinkerpop/pull/1743) that was missed due to the above mistake. 

Doubled the wait-for-server retry count as sometimes neo4j takes longer to get installed in GHA, to prevent premature exit of the tests.

### DIFF
--- a/gremlin-go/run.sh
+++ b/gremlin-go/run.sh
@@ -44,7 +44,8 @@ GREMLIN_SERVER_VERSION=$(grep tinkerpop -A2 pom.xml | grep -Po '(?<=<version>)([
 export GREMLIN_SERVER="${1:-$GREMLIN_SERVER_VERSION}"
 echo "$GREMLIN_SERVER"
 
-# Passes current gremlin server version into docker compose as environment variable
+# Passes current gremlin server version into docker compose as environment variable & removes all service containers
 docker-compose up --build --exit-code-from gremlin-go-integration-tests
-# Removes all service containers
+EXIT_CODE=$?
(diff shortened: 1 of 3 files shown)


gremlin-go/run.sh
```diff
@@ -44,7 +44,5 @@ GREMLIN_SERVER_VERSION=$(grep tinkerpop -A2 pom.xml | grep -Po '(?<=<version>)([
 export GREMLIN_SERVER="${1:-$GREMLIN_SERVER_VERSION}"
 echo "$GREMLIN_SERVER"
 
-# Passes current gremlin server version into docker compose as environment variable
-docker-compose up --build --exit-code-from gremlin-go-integration-tests
-# Removes all service containers
-docker-compose down
+# Passes current gremlin server version into docker compose as environment variable & removes all service containers
+docker-compose up --build --abort-on-container-exit --exit-code-from gremlin-go-integration-tests && docker-compose down
```
jroimartin, line 48: Do we really need `docker-compose down` if we abort on container exit?

jroimartin:
LGTM. Good catch with https://github.com/apache/tinkerpop/pull/1743!

PR #1754, opened 2022-07-12
gremlin-go: fix deadlock on (*DriverRemoteConnection).Close

jroimartin:
Further analysis shows that apache/tinkerpop#1734 is not the right fix. A deadlock can still happen in `(*gremlinServerWSProtocol).readLoop` if `errorCallback` is called before the connection protocol is closed and, because of this, `(*gremlinServerWSProtocol).close` calls `(*gremlinServerWSProtocol).wg.Wait`.

This commit adds the parameter "wait" to `protocol.close`. So, the call to `(*gremlinServerWSProtocol).wg.Wait` can be skipped when called from within an error callback. Also, the test that was introduced has been updated.

### DIFF
--- a/gremlin-go/driver/connection.go
+++ b/gremlin-go/driver/connection.go
@@ -55,8 +55,10 @@ type connectionSettings struct {
 func (connection *connection) errorCallback() {
 	connection.logHandler.log(Error, errorCallback)
 	connection.state = closedDueToError
-	err := connection.protocol.close()
-	if err != nil {
+
+	// This callback is called from within protocol.readLoop. Therefore,
+	// it cannot wait for it to finish to avoid a deadlock.
+	if err := connection.protocol.close(false); err != nil {
 		connection.logHandler.logf(Error, failedToCloseInErrorCallback, err.Error())
 	}
 }
@@ -68,7 +70,7 @@ func (connection *connection) close() error {
 	connection.logHandler.log(Info, closeConnection)
 	var err error
 	if connection.protocol != nil {
-		err = connection.protocol.close()
+		err = connection.protocol.close(true)
 	}
 	connection.state = closed
 	return err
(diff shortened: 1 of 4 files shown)


jroimartin:
@simonz-bq @lyndonbauto you might want to give this one a look because you already have the context of #1734. Thanks!

simonz-bq:
@jroimartin, correct me if I'm wrong, but another deadlock can occur here. Let's call them user for invoked `close` and readLoop for error-caused `close`.

1. `readLoop` errors and calls the `errorCallback`, which in turn calls `close`
2. `readLoop` holds the `close` lock
3. `readLoop` finishes closing, releases the `close` lock, goes back to the loop, and invokes `Done()` before returning
5. user holds the `close` lock
6. user waits for the waitgroup

There is a race condition here where we assume step 6 happens before step 3 finishes.

I *think* this can be solved by saving the `closed` state before to a variable, and then checking that && wait to solve this.

gremlin-go/driver/connection.go
```diff
@@ -55,8 +55,10 @@ type connectionSettings struct {
 func (connection *connection) errorCallback() {
 	connection.logHandler.log(Error, errorCallback)
 	connection.state = closedDueToError
-	err := connection.protocol.close()
-	if err != nil {
+
+	// This callback is called from within protocol.readLoop. Therefore,
```
simonz-bq, line 59: Can we modify protocol.readLoop?

Simply I would just like all `protocol.wg.Done()` to be removed, and instead added before the `for` as `defer(protocol.wg.Done()`.

simonz-bq (approved):

gremlin-go/driver/connection.go
jroimartin: Sure thing. Done in c93661f. I did the same in `transporter.writeLoop`.

simonz-bq (approved):
LGTM, thanks!

PR #1757, opened 2022-07-16
Gremlin-Go: Update CI to run on Windows

xiazcy:
With these changes, users should be able to run `mvn clean install` in Windows with 3.5-dev successfully. 

Main changes are:
1. The removal of the plugin `dockerfile-maven-plugin`, as it is fails on Windows without clear resolution, we are using raw docker build execution in server for image building.
2. The use of `docker-compose` directly instead of the `run.sh` script, as bash script will not work in Windows. The `run.sh` script will remain as an optional tool in gremlin-go for uses to run the tests. 

A flag `-DskipImageBuild` is introduced to allow skipping the image building on incompatible platforms. Currently GHA Windows Servers fails when building linux based images (alpine jdk), and so the server image building have been disabled in Spark windows GHA while we investigate alternative resolutions. 

For 3.6-dev, running `mvn clean install -pl :gremlin-go` will work in Windows (with a `.glv` file) with this change, however, there are errors in other modules which were not encountered in 3.5-dev. These will be investigated and addressed separately.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -92,7 +92,7 @@ jobs:
       - name: Build with Maven Windows
         if: runner.os == 'Windows'
         run: |
-          mvn clean install -pl -:gremlin-javascript,-:gremlin-python,-gremlin-dotnet,-:gremlin-dotnet-source,-:gremlin-dotnet-tests,-:gremlint -q -DskipTests -Dci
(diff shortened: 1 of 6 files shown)


kenhuuu:
LGTM. This seems helpful for Windows users that want an easy way to build Docker images for the server through Maven.

PR #1758, opened 2022-07-18
TINKERPOP-2772 Add Spark utility to load vertices as RDD

li-boxuan:


spmallette:
thanks @li-boxuan - Just to be clear - the general point of this change is to expose this helper method that i suppose is designed to be called directly by users:

```
public static JavaPairRDD<Object, VertexWritable> loadVertices(
        final org.apache.commons.configuration2.Configuration sparkGraphConfiguration,
        final JavaSparkContext sparkContext)
```

is that right?

li-boxuan:
@spmallette Your understanding is correct. I didn’t add doc yet in case the community does not like this idea.

spmallette:
Seems like a helpful function. No breaking change. Please add javadoc, upgrade docs and CHANGELOG entry. not sure if user documentation is necessary here, but if you think it would be helpful, feel free to add that. i think it would suffice to just have example usage in the javadoc though. Not sure if you want to leave this targeted as master branch, but if you want it in a current release line, i don't see any reason you couldn't rebase on 3.5-dev.

PR #1761, opened 2022-07-18
TINKERPOP-2769: Add server response for tasks that never run because of timeout.

kenhuuu:
Add handling for situations where a task that is supposed to respond to the client is cancelled before it runs because it times out.
Fixes https://issues.apache.org/jira/browse/TINKERPOP-2769


jroimartin:
@kenhuuu I tested the PR and I can confirm that both the gremlin-go and the gremlin-python reproducers are not able to trigger the issue anymore. Thanks!

spmallette:
Could you please add a CHANGELOG entry and do a quick review for `final` declarations where needed? nice looking tests, so other than that, this one looks ok i think. i may want to give it a second look though before merging as this area of code with timeouts is always tricky.

PR #1765, opened 2022-07-19
Allow connections from IPv6 for `test-server-start.groovy` server

L0Lmaker:
Attempting to reach localhost on Windows with a client can be resolved with IPv6 and IPv4 addresses. The WebSocket client tries to establish a connection to the IPv6 address with a timeout of 1 sec and attempts to do this 2 times. Gremlin Server does not accept connection for IPv6, so it refuses it. Then the client connects to IPv4 address.

With 0.0.0.0 server will accept all connections, including IPv6.


kenhuuu:
LGTM

PR #1766, opened 2022-07-22
Fix dotnet tests for Windows GHA

vkagamlyk:
Enabled running dotnet tests on Windows on GitHub Actions.

Fixed dotnet tests for Windows GHA:
- test ShouldUseSpecifiedEvaluationTimeout. The test checks that EvalTimeout parameter works. Usual test request waiting time on server before execution is ~10-25 seconds, so I increased timeout to have time to spare.
- test ShouldHandleCloseMessageForInFlightRequestsAsync. In dotnet driver, sending requests and reading responses is carried out in different threads without synchronization. If an error occurs, it is processed by the reading thread (Connection.NotifyAboutConnectionFailure method), so there is a race condition between the two threads to read/update the connection.NrRequestsInFlight property (based on _callbackByRequestId field).

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -187,7 +187,10 @@ jobs:
     name: .NET
     timeout-minutes: 20
     needs: smoke
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
     steps:
       - uses: actions/checkout@v3
       - name: Set up JDK11
--- a/gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Driver/MessagesTests.cs
+++ b/gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Driver/MessagesTests.cs
@@ -102,7 +102,7 @@ namespace Gremlin.Net.IntegrationTest.Driver
             using (var gremlinClient = new GremlinClient(gremlinServer))
             {
                 const long timeOutInMs = 1L;
-                const int scriptSleepTimeInMs = 30000;
+                const int scriptSleepTimeInMs = 60000;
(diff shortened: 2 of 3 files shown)


.github/workflows/build-test.yml
```diff
@@ -187,7 +187,10 @@ jobs:
     name: .NET
     timeout-minutes: 20
     needs: smoke
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
```
L0Lmaker, line 193: Windows build will fail at touch command, suggestion for `dotnet` workflow:
```
dotnet:
    name: .NET
    timeout-minutes: 20
    needs: smoke
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ windows-latest ]
    steps:
      - uses: actions/checkout@v3
      - name: Set up JDK11
        uses: actions/setup-java@v3
        with:
          java-version: '11'
          distribution: 'temurin'
      - name: Set up .NET 6.0.x
        uses: actions/setup-dotnet@v2
        with:
          dotnet-version: '6.0.x'
      - name: Create .glv files - Ubuntu
        if: contains(matrix.os, 'ubuntu')
        run: |
          touch gremlin-dotnet/src/.glv
          touch gremlin-dotnet/test/.glv
      - name: Create .glv files - Windows
        if: contains(matrix.os, 'windows')
        run: |
          echo $null >> gremlin-dotnet/src/.glv
          echo $null >> gremlin-dotnet/test/.glv
      - name: Build with Maven
        run: |
          mvn clean install -pl -:gremlin-javascript,-:gremlin-python,-:gremlint -q -DskipTests -Dci
          mvn verify -pl :gremlin-dotnet,:gremlin-dotnet-tests -P gremlin-dotnet -DincludeNeo4j
```

L0Lmaker: It doesn't seem like it failed so it seems this might not be an issue after all. Hiding comment as it's resolved.

PR #1768, opened 2022-07-27
TINKERPOP-2765 Fix concurrency issue during script translation in the GremlinGroovyScriptEngine.

kenhuuu:
An issue occurs when the GremlinGroovyScriptEngine is called from multiple threads to evaluate bytecode that contains lambdas. The default translator is not thread-safe which results in queries being interlaced with one another. This is fixed by creating a new instance of the translator per translation so that the translation process isn't interfered with. This primarily affects the UnifiedChannelizer because it currently uses the same instance of the translator from multiple threads.

Fixes https://issues.apache.org/jira/browse/TINKERPOP-2765

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,6 +25,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 * Changed label generation in `PathProcessorStrategy` to be more deterministic.
 * Bumped to Apache `commons-configuration` 2.8.0 to fix security vulnerability.
+* Fixed issue where the `GremlinGroovyScriptEngine` reused the same translator concurrently which lead to incorrect translations.
 
 [[release-3-5-4]]
 === TinkerPop 3.5.4 (Release Date: July 18, 2022)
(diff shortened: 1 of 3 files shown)


spmallette:
Do you feel like this is just a patch for the stated problem in the JIRA? Would it be better if the translator were just made to be thread-safe:

https://issues.apache.org/jira/browse/TINKERPOP-2675

I'm not against merging this as it is to solve the problem, but i'm curious about the bigger issue of general thread-safety and whether or not it would make sense to refactor as TINKERPOP-2675 suggests or if your opinion would be to leave that all as it is.

kenhuuu:
> Do you feel like this is just a patch for the stated problem in the JIRA? Would it be better if the translator were just made to be thread-safe:
> 
> https://issues.apache.org/jira/browse/TINKERPOP-2675
> 
> I'm not against merging this as it is to solve the problem, but i'm curious about the bigger issue of general thread-safety and whether or not it would make sense to refactor as TINKERPOP-2675 suggests or if your opinion would be to leave that all as it is.

This is something that I thought about when investigating the fix, but I decided on making this simple patch so that it could be merged into 3.5-dev. I think it does make more sense to refactor the translators but that would likely require a breaking change so I thought it wasn't something we would want to do for 3.5-dev.

spmallette:
Can you please add a CHANGELOG entry and submit a second PR for 3.6-dev (the merge will likely conflict given the heavy refactoring that took place between the two branches)? I suggest you do the CHANGELOG for this PR first, then I can merge this one through the branches. I'll ignore conflicts for 3.6-dev/master and then you can issue the second PR to 3.6-dev.

spmallette:
>  I'll ignore conflicts for 3.6-dev/master and then you can issue the second PR to 3.6-dev.

@kenhuuu could you please submit the PR against 3.6-dev for these changes as the merge is now complete:

kenhuuu:
Sure, I made a PR for 3.6-dev at https://github.com/apache/tinkerpop/pull/1778. Thanks.

PR #1785, opened 2022-08-17
Export Result interface

good41898:
The `Result` struct does not allow for any assigning of data, which makes it impossible to mock out for testing. While the `ResultSet` interface and its methods can be mocked out, they must match the original function signature, and return the `Result` type. Since you cannot assign to the unexported `Result` type, it is not possible to mock out returning data from the various methods of `ResultSet`.

### DIFF
--- a/gremlin-go/driver/connection_test.go
+++ b/gremlin-go/driver/connection_test.go
@@ -21,6 +21,9 @@ package gremlingo
 
 import (
 	"crypto/tls"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/text/language"
 	"math/big"
 	"os"
 	"reflect"
@@ -30,10 +33,6 @@ import (
 	"sync"
 	"testing"
 	"time"
-
-	"github.com/google/uuid"
-	"github.com/stretchr/testify/assert"
-	"golang.org/x/text/language"
 )
 
 const personLabel = "Person"
--- a/gremlin-go/driver/connection_test.go
+++ b/gremlin-go/driver/connection_test.go
@@ -21,9 +21,6 @@ package gremlingo
 
 import (
 	"crypto/tls"
-	"github.com/google/uuid"
-	"github.com/stretchr/testify/assert"
-	"golang.org/x/text/language"
 	"math/big"
 	"os"
 	"reflect"
@@ -33,6 +30,10 @@ import (
 	"sync"
 	"testing"
 	"time"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/text/language"
 )
 
 const personLabel = "Person"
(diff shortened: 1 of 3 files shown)


good41898:
https://issues.apache.org/jira/browse/TINKERPOP-2785

xiazcy:
Hi, @good41898, thank you for raising the issue. We will take a look. 
Also a quick note before diving into the issue, please make sure all tests pass locally with your changes. As it currently stands, just exporting the field in the `Result` struct without changing any reference to it will generate errors. Thanks!

good41898:
Thanks @xiazcy ! I was able to successfully run unit tests locally for Result and ResultSet packages, but am running into connection issues with the local gremlin server while running `./run.sh 3.6.1` even though I have confirmed both the gremlin server container and integration test container are up and running, and on the same docker network. Thanks

lyndonbauto:
@good41898 Hi there, thanks for using the Golang driver, cool to see people are putting it into things now.

First off, did you need some help with running the tests? If you need more assistance there feel free to jump into the TinkerPop Discord and send us some more details, you can @ myself or yang to get our attention quicker.

On the note of this code. I am unsure why this is required. The typical technique from my understanding of Golang (which I will admit is minimal) is to mock by making your own interface, for example: https://github.com/apache/tinkerpop/blob/master/gremlin-go/driver/gorillaTransporter_test.go#L35-L66

Please correct my if I am wrong in my understanding. 

Anyway I will approve your workflow to run since @xiazcy has suggested it might create some errors.

lyndonbauto:
Looks like tests are passing. I am OK with moving this through if there is an actual need, just need you to get back to me on this part

> On the note of this code. I am unsure why this is required. The typical technique from my understanding of Golang (which I will admit is minimal) is to mock by making your own interface, for example: https://github.com/apache/tinkerpop/blob/master/gremlin-go/driver/gorillaTransporter_test.go#L35-L66

before we proceed @good41898

good41898:
You are correct that the typical technique for mocking is by making your own interface! 

To give an example of the issue, I have abstracted the gremlin-go driver into an interface like this:

`GremlinGoDriver interface {
   Submit(traversalString string) (gremlingo.ResultSet, error)
   Close()
}`

Which contains the two methods of the driver that I use. While unit testing, I am able to mock out any methods (like the Submit() and Close() method) as long as the function signatures for the mocked methods match the signatures of the actual methods. Where I am running into issues is when trying to mock returned data from a `ResultSet` object. I am able to create a new struct containing the result set so I can mock those methods (like the ResultSet.All() method) as well, like this:

```
type MockResultSet struct {
  ResultSet
 }
 
 func (m MockResultSet) All() ([]*Result, error) {
    var res []*Result
    return  res
 }
```
  
However, since the function signatures must match, this mocked All() method must return a Result object, which I cannot assign data to as the result field is non-exported.

Let me know if this is helpful/ makes sense. Thanks.

good41898:
closing pull request, will open a new one based off of `3.5-dev`

PR #1786, opened 2022-08-18
Export Data field

good41898:
The `Result` struct does not allow for any assigning of data, which makes it impossible to mock out for testing. While the `ResultSet` interface and its methods can be mocked out, they must match the original function signature, and return the `Result` type. Since you cannot assign to the unexported `Result` type, it is not possible to mock out returning data from the various methods of `ResultSet`.

### DIFF
--- a/gremlin-go/driver/result.go
+++ b/gremlin-go/driver/result.go
@@ -27,17 +27,17 @@ import (
 
 // Result Struct to abstract the Result and provide functions to use it.
 type Result struct {
-	result interface{}
+	Result interface{}
 }
 
 // String returns the string representation of the Result struct in Go-syntax format.
 func (r *Result) String() string {
-	return fmt.Sprintf("result{object=%v class=%T}", r.result, r.result)
+	return fmt.Sprintf("result{object=%v class=%T}", r.Result, r.Result)
 }
 
 // GetString gets the string representation of the result.
 func (r *Result) GetString() string {
-	return fmt.Sprintf("%v", r.result)
+	return fmt.Sprintf("%v", r.Result)
 }
 
 // GetInt gets the result by coercing it into an int, else returns an error if not parsable.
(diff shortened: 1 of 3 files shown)


good41898:
To give an example of the issue, I have abstracted the gremlin-go driver into an interface like this:
```
GremlinGoDriver interface {
   Submit(traversalString string) (gremlingo.ResultSet, error)
   Close()
}
```

Which contains the two methods of the driver that I use. While unit testing, I am able to mock out any methods (like the Submit() and Close() method) as long as the function signatures for the mocked methods match the signatures of the actual methods. Where I am running into issues is when trying to mock returned data from a `ResultSet` object. I am able to create a new struct containing the result set so I can mock those methods (like the ResultSet.All() method) as well, like this:

```
type MockResultSet struct {
  ResultSet
 }
 
 func (m MockResultSet) All() ([]*Result, error) {
    var res []*Result
    return  res
 }
```
  
However, since the function signatures must match, this mocked All() method must return a Result object, which I cannot assign data to as the result field is non-exported.

Let me know if this is helpful/ makes sense. Thanks.

good41898:
https://issues.apache.org/jira/browse/TINKERPOP-2785

gremlin-go/driver/result.go
```diff
@@ -27,17 +27,17 @@ import (
 
 // Result Struct to abstract the Result and provide functions to use it.
 type Result struct {
-	result interface{}
+	Result interface{}
```
xiazcy, line 30: One minor item, to avoid confusion with the struct and field names being the same, I'd suggest to change the field name to `Data` instead. 
```suggestion
	Data interface{}
```
So we'll have `r.Result` changed to `r.Data`. 
The rest LGTM. Thanks!

good41898: done! Thanks 👍

xiazcy (approved):
LGTM, thanks! The failing action seem to be in python, so we should be good with go.

PR #1797, opened 2022-08-29
Dev Environments: Dockerize Testing for Python and Gremlin Console (Python-based tests only)

xiazcy:
Dockerized `gremlin-python` build and testing. Building, testing, and packaging of the python GLV are now executed inside of Docker, and python installation is no longer a system requirement. Outputs and distribution packages and are placed in the `target/` folder as before. Deployment of python still requires local Python installations. 

Dockerized `gremlin-console` python-based integration tests. Python 3.8 is used for the image because the version of pytest used isn't compatible with later versions of Python. Building our own Docker image also allows for greater control over the version of Python and Java used.

The location of the python test output was changed in order to fit better with the bind mounted volume for Docker.

Minor documentation updates included. We will be looking at the Develop Docs more holistically for more appropriate updates in a separate PR.

 Conflicts will be left alone for now, as we plan to rebased after the previous [PR](https://github.com/apache/tinkerpop/pull/1788) is merged.

### DIFF
--- a/gremlin-python/pom.xml
+++ b/gremlin-python/pom.xml
@@ -158,6 +158,11 @@ limitations under the License.
                                             <env key="BUILD_DIR" value="${project.build.directory}/python3"/>
                                             <arg line="down"/>
                                         </exec>
+                                        <exec executable="docker" failonerror="true">
+                                            <env key="PYTHONPATH" value=""/>
+                                            <env key="BUILD_DIR" value="${project.build.directory}/python3"/>
+                                            <arg line="image prune --filter label=maintainer=dev@tinkerpop.apache.org -f"/>
+                                        </exec>
                                     </target>
                                 </configuration>
                             </execution>
--- a/gremlin-python/pom.xml
+++ b/gremlin-python/pom.xml
@@ -83,102 +83,31 @@ limitations under the License.
             </activation>
             <build>
                 <plugins>
-                    <!-- need to create python-reports directory at this point or else pytest can't write the report to it -->
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-antrun-plugin</artifactId>
                         <executions>
+                            <!-- copy source files in python directory to target/, to keep testing separate from packaging.
+                            We use target/python3 for testing and target/python-packaged for distribution tasks.-->
                             <execution>
-                                <id>create-python-reports-directory</id>
-                                <phase>initialize</phase>
-                                <goals>
-                                    <goal>run</goal>
-                                </goals>
-                                <configuration>
-                                    <target>
-                                        <mkdir dir="${project.build.directory}/python-reports"/>
-                                        <mkdir dir="${project.build.directory}/python-packaged/env"/>
-                                    </target>
-                                </configuration>
-                            </execution>
-                            <!-- copy files in python directory to target/py and run virtual env to sandbox python.
-                                 there is no need to "activate" the virtualenv because all calls to python occur
-                                 directly from bin/ -->
-                            <execution>
-                                <id>setup-py-env</id>
+                                <id>setup-env</id>
                                 <phase>process-resources</phase>
                                 <goals>
                                     <goal>run</goal>
                                 </goals>
                                 <configuration>
                                     <tasks>
-                                        <!-- seems like we need a few different copies of the same source. all the
-                                             different python stuff doesn't seem to want to share. we use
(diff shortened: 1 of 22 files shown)


xiazcy:
Also, please note that the files changed will overlap with the previous [PR](https://github.com/apache/tinkerpop/pull/1788) as they share the same set up. 

The PR-specific changes are in the `gremlin-python` and `gremlin-console` folders.

vkagamlyk:
Build/test time for GLV with and without Docker

Testing environment:

m4.xlarge AWS EC2 instance (4 vCPU, 16GiB RAM)

OS: Ubuntu 22.04.1 LTS

Java: openjdk version "11.0.16" 2022-07-19
OpenJDK Runtime Environment (build 11.0.16+8-post-Ubuntu-0ubuntu122.04)
OpenJDK 64-Bit Server VM (build 11.0.16+8-post-Ubuntu-0ubuntu122.04, mixed mode, sharing)

Docker: version 20.10.17, build 100c701

.glv files add with `touch {gremlin-python,gremlin-javascript,gremlin-go}/.glv`
Tests performed with command `sudo time mvn clean install`
First test run skipped.

| branch |.glv added | Average time |
| --- | --- | --- |
| 3.5 | no | 20:47.52 |
| 3.5 | yes | 24:56.99 |
| dockerize-others | yes | 28:55.19 |

Build time increase is ~15%

gremlin-python/pom.xml
```diff
@@ -83,102 +83,31 @@ limitations under the License.
...
                                         </copy>
-                                        <exec dir="${project.build.directory}/python3" executable="python3"
-                                              failonerror="true">
-                                            <arg line="--version"/>
-                                        </exec>
-                                        <exec dir="${project.build.directory}/python3" executable="virtualenv"
-                                              failonerror="true">
-                                            <arg line="--version"/>
-                                        </exec>
-                                        <exec dir="${project.build.directory}/python3" executable="virtualenv"
```
spmallette, line 134: as i sit here admiring all this pom.xml code getting simplified behind docker, i wonder if these changes will allow releases to actually work. for example, here we get rid of the virtualenv setup of python3, wheel, setuptools, etc. which would all be needed on release day (i think) when we deploy to pypi with `-Dpypi`: https://github.com/apache/tinkerpop/pull/1797/files#diff-924caaf2e60b71e4f805fe7d35f9c8895cb8d23ae258b12b7497b21d91ab639fR249-R269

thoughts?

xiazcy: That is one thing I wasn't sure of, how much the impact would be to release, not only with the set up removed, but also with the packaging taken inside of docker. I don't have any experiences with `pypi` releases though, I don't imagine there is any `mock release` to try? In any case, we could move the necessary python environment set up into the release profile to make sure it still works.

Some other options could be to move the packaging stage for python back into maven, so we keep the environment set up, to be sure it works the same way as before for package and release. Then down the road we could consider putting both package and release into Docker.

spmallette: > In any case, we could move the necessary python environment set up into the release profile to make sure it still works.

i was thinking you should just do this for now. in the slightly longer term (i.e. next release manager tries this approach out and documents it), maybe we just go to native release instructions. 

> Then down the road we could consider putting both package and release into Docker.

yes - that's a good long term goal

gremlin-python/src/main/python/setup.py
```diff
@@ -75,7 +75,7 @@
     tests_require=[
         'pytest>=4.6.4,<7.2.0',
         'mock>=3.0.5,<4.0.0',
-        'radish-bdd==0.8.6',
+        'radish-bdd==0.13.4',
```
xiazcy, line 78: Just to note that I've bumped the radish test suite to the latest version. 

While it no longer matters inside of Docker (tests pass for either version), it resolves the error `Setup script exited with error in radish-parse_type setup command: use_2to3 is invalid.` for the previous pom set up.

gremlin-python/pom.xml
xiazcy: The environment set up is added to the deploy profile in commit [ca0cdf0](https://github.com/apache/tinkerpop/pull/1797/commits/ca0cdf047710895f38a143fd0b9eada3604620f6). There could be issues we won't realized until the actual release process though (hopefully not).

xiazcy:
VOTE +1. Given the two weeks review period, and that there are no additional comments, I will merge this through today.

vkagamlyk (commented):
VOTE +1

PR #1808, opened 2022-09-13
TINKERPOP-2798 Mid-traversal E()

vkagamlyk:
- Added support for mid-traversal E()
- Added unit tests
- Added Cucumber tests
- Updated doc

Jira ticket https://issues.apache.org/jira/browse/TINKERPOP-2798


docs/src/dev/future/index.asciidoc
```diff
@@ -49,7 +49,6 @@ line of 22H2.
 
 * Add support for traversals as parameters for `V()`, `is()`, and `has()` (includes `Traversal` arguments to `P`)
 * Geospatial support for TinkerPop (link:++https://lists.apache.org/list?dev@tinkerpop.apache.org:2021-7:DISCUSS%20geo-spatial++[DISCUSS Thread])
-* Add mid-traversal `E()` support
```
spmallette, line 52: could you please create a JIRA for this and reference it here rather than deleting the line? if folks want to know the status they could just look in on the ticket. i believe that's how it was being done prior to the 3.6.x release. it would be good to get the JIRA on this PR too so that it all links together nicely.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -1236,8 +1236,8 @@ g.V()
 [[e-step]]
 === E Step
 
-The `E()`-step is meant to read edges from the graph and is usually used to start a `GraphTraversal`. Unlike the related
-<<v-step,V()>> it cannot be used mid-traversal.
+The `E()`-step is meant to read edges from the graph and is usually used to start a `GraphTraversal`, but can also
+be used mid-traversal.
```
spmallette, line 1240: please add an example of mid-traversal `E()` here

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -331,6 +331,21 @@ public default GraphTraversal<S, Vertex> V(final Object... vertexIdsOrElements)
         return this.asAdmin().addStep(new GraphStep<>(this.asAdmin(), Vertex.class, false, ids));
     }
 
+    /**
+     * A {@code E} step is usually used to start a traversal but it may also be used mid-traversal.
+     *
+     * @param edgeIdsOrElements vertices to inject into the traversal
```
spmallette, line 337: "vertices" should be "edges" in the javadoc there.

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs
```diff
@@ -141,73 +159,73 @@ private GraphTraversal(ICollection<ITraversalStrategy> traversalStrategies, Byte
         /// <summary>
         ///     Adds the aggregate step to this <see cref="GraphTraversal{SType, EType}" />.
         /// </summary>
-        public GraphTraversal<S, E> Aggregate (Scope scope, string sideEffectKey)
+        public GraphTraversal<S, E_> Aggregate (Scope scope, string sideEffectKey)
         {
             Bytecode.AddStep("aggregate", scope, sideEffectKey);
-            return Wrap<S, E>(this);
+            return Wrap<S, E_>(this);
```
spmallette, line 165: ugh - that's too bad....

spmallette:
needs a CHANGELOG entry and Upgrade Documentation, but looking pretty good. thanks!

docs/src/dev/future/index.asciidoc
vkagamlyk: Added ticket https://issues.apache.org/jira/browse/TINKERPOP-2798

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs
vkagamlyk: Class was defined as `public class GraphTraversal<S, E> : DefaultTraversal<S, E>`
[C# don't allow to have method with same name as type](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0102)

Possible solutions:
1. rename `E` in class definition (internal effect only, I used it)
2. rename `E` function

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
vkagamlyk:
```suggestion
     * @param edgeIdsOrElements edges to inject into the traversal
```

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs
spmallette: yeah, i was just making an observation. not much we can do there. personally, i don't mind the underscore but maybe @FlorianHockmann or others could offer a better suggestion?

docs/src/dev/future/index.asciidoc
vkagamlyk: added in https://github.com/apache/tinkerpop/pull/1808/commits/13501c50233d1692d2cd3dd8185e61e9050f8fe3

docs/src/reference/the-traversal.asciidoc
vkagamlyk: added in https://github.com/apache/tinkerpop/pull/1808/commits/13501c50233d1692d2cd3dd8185e61e9050f8fe3

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
vkagamlyk: fixed in https://github.com/apache/tinkerpop/pull/1808/commits/13501c50233d1692d2cd3dd8185e61e9050f8fe3

vkagamlyk:
> needs a CHANGELOG entry and Upgrade Documentation, but looking pretty good. thanks!

CHANGELOG updated in https://github.com/apache/tinkerpop/pull/1808/commits/13501c50233d1692d2cd3dd8185e61e9050f8fe3

gremlin-dotnet/src/Gremlin.Net/Gremlin.Net.csproj
```diff
@@ -67,7 +67,7 @@ NOTE that versions suffixed with "-rc" are considered release candidates (i.e. p
     <RepositoryUrl>https://github.com/apache/tinkerpop</RepositoryUrl>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>    
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
-    <EnablePackageValidation>true</EnablePackageValidation>
+    <EnablePackageValidation>false</EnablePackageValidation>
```
FlorianHockmann, line 70: Please revert this change. Package validation ensures that we don't accidentally add breaking changes without being aware of it.
In this case, we accept the breaking change of renaming the generic type parameter from `E` to `E_` of `GraphTraversal`. So, we can silence the warning for this change. You can do this by adding the following entry to the file `CompatibilitySuppressions.xml` in `gremlin-dotnet/src/Gremlin.Net`:

```xml
<!-- Renamed generic type parameter name in GraphTraversal -->
<Suppression>
  <DiagnosticId>CP0001</DiagnosticId>
  <Target>T:Gremlin.Net.Process.Traversal.GraphTraversal`2</Target>
  <Left>lib/netstandard2.0/Gremlin.Net.dll</Left>
  <Right>lib/netstandard2.0/Gremlin.Net.dll</Right>
  <IsBaselineSuppression>true</IsBaselineSuppression>
</Suppression>
```

behind the other two suppression entries there.

I realize though that this compatibility checking and how to deal with its warnings isn't easy to understand, especially for non-.NET developers who just want to modify Gremlin and therefore also have to make changes in .NET. So, I'm wondering whether we should keep the feature at all. Nevertheless, that shouldn't be part of this PR, so please add the suppression for now and reenable the validation here.

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs
FlorianHockmann: The C# [naming convention suggests](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/generics/generic-type-parameters#type-parameter-naming-guidelines) that names of generic type parameters like these should start with a _T_. We didn't follow that and instead just reused the names from Java _S_ and _E_, but I'm wondering if we should simply change these into _TS_ and _TE_ or even _TStart_ and _TEnd_ to be more descriptive and comply with the naming convention that most C# devs are used to.
This would then also solve this problem here.

If you don't have any objections, then I'll create a separate issue in Jira to change these names in a different PR. But this shouldn't block this PR here. So, I suggest that we keep it for now as you've did in this PR with `E_` and then we can later change that to `TE` / `TEnd` or whatever.

That's definitely better than changing the name of the step `E()` itself as that would lead to a lot more confusion in my opinion.

spmallette: @vkagamlyk would you mind taking the extra step to fix it on this PR? or an immediate follow-up PR? i'd hate for us to forget this change when we have a chance to instituted it on 3.7.x and let this release with the underscore.

CHANGELOG.asciidoc
```diff
@@ -30,6 +30,10 @@ This release also includes changes from <<release-3-6-XXX, 3.6.XXX>>.
 * Added warning on vertex property cardinality mismatch when reading GraphML.
 * Bumped to `ws` 8.x for `gremlin-javascript`.
 
+==== Improvements
+
+* TINKERPOP-2798 Add support for mid-traversal E()
```
spmallette, line 35: The JIRA entries to CHANGELOG are added on release. You should add an entry above. That entry is typically meant to be more descriptive than the JIRA title or even more specific as you might do multiple entries per JIRA item.

gremlin-test/features/map/Edge.feature
```diff
@@ -0,0 +1,51 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+@StepClassMap @StepE
+Feature: Step - E()
```
spmallette, line 19: I assume it should work without any trouble but I think it's worth adding some tests for where your new mid-traversal `E()` is taking some arguments (including `null`).

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs
vkagamlyk: @FlorianHockmann is it better to rename TStart/TEnd in all affected classes or only fix `GraphTraversal<TStart, TEnd >` in this PR?

gremlin-test/features/map/Edge.feature
vkagamlyk: added in https://github.com/apache/tinkerpop/pull/1808/commits/edd2160056fcd2816e565ee7da757479dc139502

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs
FlorianHockmann: I would do it in all affected classes to stay consistent.

vkagamlyk: renamed in https://github.com/apache/tinkerpop/pull/1808/commits/fc3480cc5a217a215e48ae13396082effd59085a

gremlin-dotnet/src/Gremlin.Net/Gremlin.Net.csproj
vkagamlyk: @FlorianHockmann  Should I add all renamed classes and methods to `CompatibilitySuppressions.xml`?

How to make exclusion for method with generic param like `Gremlin.Net.Process.Traversal.GraphTraversalSource.Inject<S>(params S[])`?

CHANGELOG.asciidoc
vkagamlyk: updated in https://github.com/apache/tinkerpop/pull/1808/commits/41240b356c3cde71f673952d27ef061983402c9d

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs
vkagamlyk: Should I propagate these changes to `3.5-dev` and `3.6-dev`?

gremlin-dotnet/src/Gremlin.Net/Gremlin.Net.csproj
FlorianHockmann: You can generate the file automatically:

 ```bash
dotnet pack gremlin-dotnet/src/Gremlin.Net /p:GenerateCompatibilitySuppressionFile=true
```

or apparently also if you temporarily add this to the Gremlin.Net.csproj:

```xml
<GenerateCompatibilitySuppressionFile>true</GenerateCompatibilitySuppressionFile>
```

as described [here](https://docs.microsoft.com/en-us/dotnet/fundamentals/package-validation/overview#suppress-compatibility-errors).

This will remove the comments that we currently have in the file, but I guess those aren't that helpful any way so it shouldn't be a problem.

vkagamlyk: updated in https://github.com/apache/tinkerpop/pull/1808/commits/5028d7a6078f828a5b8d69426464e67bb5853f22

spmallette:
please rebase/squash these commits before merge but other than that - VOTE +1

FlorianHockmann (approved):
Everything looks good, but for some reason there is no update to the `Gremlin.cs` class which includes the .NET traversals for all scenarios, just like `gremlin.go` for Go and `gremlin.js` for JS.

You probably just forgot to `git add` it as it needed to be present to run the Gherkin tests for the added scenarios. I also just tried it locally with your branch and I got an updated version with these traversals.

Feel free to merge this directly after adding these missing changes.

VOTE +1

vkagamlyk:
VOTE +1

PR #1817, opened 2022-09-29
Disable compatibility check against base version

FlorianHockmann:
I added these checks in https://github.com/apache/tinkerpop/commit/963e625219e9c5e6a144486f72d3dc0220ebe8df because I've also added condition compilation there based on the used .NET version. `SendMessageUncompressedAsync` is only included for .NET 6 and higher.
I think it still makes sense to have a basic package validation to check for compatibility between different .NET versions which is enabled by `EnablePackageValidation` in the csproj.

However, I've also added a version compatibility check there which should ensure that we don't accidentally add breaking changes. It turned out in the meantime that this check mostly adds a burden on .NET development which is especially problematic if someone wants to modify Gremlin in general which means that all GLVs have to be modified and then runs into these compatibility warnings.

So, I don't think anymore that these checks against a baseline version (which was 3.5.2) add enough value to accept this complexity in .NET development.

VOTE +1


vkagamlyk:
VOTE +1

PR #1818, opened 2022-09-29
TINKERPOP-2794 Add support for cancellation 

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2794

This enables users to cancel the evaluation of a single traversal / the execution of a submitted script.
Cancellation on the server is however out of scope for this contribution as we first need TINKERPOP-2210 for that. So this frees up resources on the client and requests that were not successfully submitted to the server yet can also be cancelled completely.

The first commit comes from #1817 which I used here already as this change adds a lot of changes that are technically breaking changes and would therefore require a lot of suppressions for compatibility warnings.

VOTE +1


vkagamlyk (commented):
LGTM
VOTE +1

FlorianHockmann:
I rebased on `3.5-dev` and now a test failed, but I don't it's really caused by anything from the rebase. It's probably a timing issue. I'll look into that.

FlorianHockmann:
> I rebased on 3.5-dev and now a test failed, but I don't it's really caused by anything from the rebase. It's probably a timing issue. I'll look into that.

I think I've found and solved the problem with the latest update here. The test probably just checked already whether the message was sent before that actually happened as sending messages happens asynchronously in the background. Adding a simple delay should solve this.
I ran the GitHub Action for the .NET build 5 times in a row and it passed every time.

PR #1819, opened 2022-09-29
Gremlin Arrow Flight Proposal

vkagamlyk:
Added proposal document for Arrow Flight usage

### DIFF
--- a/docs/src/dev/future/index.asciidoc
+++ b/docs/src/dev/future/index.asciidoc
@@ -91,6 +91,7 @@ story.
 |=========================================================
 |Proposal |Description |Targets |Resolved
 |link:https://github.com/apache/tinkerpop/blob/master/docs/src/dev/future/proposal-equality-1.asciidoc[Proposal 1] |Equality, Equivalence, Comparability and Orderability Semantics - Documents existing Gremlin semantics along with clarifications for ambiguous behaviors and recommendations for consistency. |3.6.0 |N
+|link:https://github.com/apache/tinkerpop/blob/master/docs/src/dev/future/proposal-arrow-flight-2[Proposal 2] |Gremlin Arrow Flight. |4.0.0 |N
 |=========================================================
 
 = Appendix
--- a/docs/src/dev/future/proposal-arrow-flight-2.asciidoc
+++ b/docs/src/dev/future/proposal-arrow-flight-2.asciidoc
@@ -1,0 +1,77 @@
+////
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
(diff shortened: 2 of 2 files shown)


joshsh:
"Gremlin Server and drivers should replace the network layer with Arrow Flight" --> What factors make it infeasible to support WebSocket and Arrow Flight in parallel?

vkagamlyk:
> "Gremlin Server and drivers should replace the network layer with Arrow Flight" --> What factors make it infeasible to support WebSocket and Arrow Flight in parallel?

My goal is to simplify development and support. WebSocket and Arrow Flight in parallel is a complication

joshsh:
> 

You don't think it would be safer to leave the WebSocket solution intact until the Arrow Flight solution has proven itself end-to-end? Stage 1, in this proposal, already involves replacing the transport layer.

vkagamlyk:
> > 
> 
> You don't think it would be safer to leave the WebSocket solution intact until the Arrow Flight solution has proven itself end-to-end? Stage 1, in this proposal, already involves replacing the transport layer.

I think it's not difficult to use another endpoint for Arrow Flight and keep existing WebSocket solution for a transitional period

joshsh:
> I think it's not difficult to use another endpoint for Arrow Flight and keep existing WebSocket solution for a transitional period

I suggest updating the proposal to reflect that.
Stage 0: clone the end-to-end WebSocket solution and begin migrating it to Arrow Flight, keeping the original solution intact
Stage n: assess the WebSocket solution and retire it if no longer needed

joshsh:
Even at Stage n, caution will be necessary. I believe there are plenty of "dark" Gremlin clients which aren't part of the official code base or known Gremlin variants, and talk directly to Gremlin Server using WebSocket.

spmallette:
Thanks for submitting this proposal for discussion. I see this targets version 4.x. Maybe it's ok to take the big break there from websockets. If this proposal were targeting 3.x, we'd have to keep websocket communication for sure. 

If the proposal sticks to targeting 4.x then it seems like one of the big questions affecting the staging of an implementation would be a schema proposal for 4.x. I assume having that proposal would help you understand how to use the standard flight serialization and thus skip past using the current serializers as part of this implementation.

joshsh:
Ah, I didn't realize this was targeted at 4.x only. Disregard my "hot swap" concerns, then. For schema, I would like to take a common specification of a TinkerPop model like [this](https://github.com/CategoricalData/hydra/blob/main/hydra-haskell/src/main/haskell/Hydra/Impl/Haskell/Sources/Ext/Tinkerpop/V3.hs) or [this](https://github.com/CategoricalData/hydra/blob/main/hydra-haskell/src/main/haskell/Hydra/Impl/Haskell/Sources/Ext/Tinkerpop/Typed.hs) and transform it into concrete APIs like [this](https://categoricaldata.github.io/hydra/hydra-java/javadoc/hydra/ext/tinkerpop/v3/package-summary.html) in a number of languages (Java, Scala, Haskell right now; working on Python support, and there will be others). We should take a fresh look at requirements for the schema language.

PR #1821, opened 2022-10-05
TINKERPOP-2808 Improving ARM Compatibility

cole-bq:
https://issues.apache.org/jira/browse/TINKERPOP-2808

Compatibility fixes for ARM based Macs

Updated base images for gremlin-server and gremlin-console docker images to support arm64:

The previously used adopt openjdk11 alpine linux image does not support arm64. In fact none of the major openjdk distributions maintain a jdk11 alpine image for arm64. Changed base image to alpine and installed oracle openjdk 11. Oracle openjdk 11 was chosen as it is already maintained in the APK community repository.

Added jansi as an explicit dependency of gremlin-console. Previously it was a dependency of jline which itself is a dependency of groovy. The console is able to run just fine on x64 architectures but it cannot run on arm JDK's without importing Jansi directly. This change resolves [TINKERPOP-2584](https://issues.apache.org/jira/browse/TINKERPOP-2584) at least for ARM Macs.

Updated dev docs with instructions to configure docker desktop for better performance on apple silicon.

### DIFF
--- a/docs/src/dev/developer/development-environment.asciidoc
+++ b/docs/src/dev/developer/development-environment.asciidoc
@@ -569,6 +569,10 @@ You should see exit code 0 upon successful completion of the test suites. Run `d
 service containers (not needed if you executed Maven commands or `run.sh`), or `docker-compose down --rmi all` to 
 remove the service containers while deleting all used images.
 
+Note for running docker with MacOS on ARM processors: Docker's performance is extremely poor on ARM Mac's in its
+default configuration. It is recommended to enable both the "New Virtualization Framework" and "VirtioFS" under
+Docker Desktop Settings -> Experimental Features.
+
 [[intellij]]
 == Intellij Usage
 
--- a/gremlin-console/Dockerfile
+++ b/gremlin-console/Dockerfile
@@ -15,13 +15,15 @@
 # specific language governing permissions and limitations
 # under the License.
 
-FROM adoptopenjdk/openjdk11:alpine-slim
+FROM alpine
 
 LABEL maintainer="dev@tinkerpop.apache.org"
 
 ARG GREMLIN_CONSOLE_DIR
 
-RUN apk add --no-cache --update bash
+RUN apk add --no-cache --update \
+    bash \
+    openjdk11-jdk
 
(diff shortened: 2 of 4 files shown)


vkagamlyk:
Please provide size of Gremlin Server docker image.

everything else is fine, VOTE +1

cole-bq:
> Please provide size of Gremlin Server docker image.
> 
> everything else is fine, VOTE +1

New image size is 372 MB, essentially the same size as before.

spmallette:
nice - people have definitely been asking for this.

> Oracle openjdk 11 was chosen as it is already maintained in the APK community repository

it sounded like there wasn't based on what you wrote in the description of the PR, but  i'll ask again...was there any other option besides Oracle openjdk? 

I once tried to change our docker image to use what i use on my system: https://sdkman.io/ which would then let us easy install whatever jdk, maven, etc. that we wanted. love that thing. i can't remember what problems I had but it never worked right. that might be a nice little thing to do if it's easy. not saying it necessarily needs to be done on this PR, but I guess I'd thought most devs moved on from Oracle JDK at this point and i thought temurin/corretto were what folks gravitated to.

cole-bq:
> nice - people have definitely been asking for this.
> 
> > Oracle openjdk 11 was chosen as it is already maintained in the APK community repository
> 
> it sounded like there wasn't based on what you wrote in the description of the PR, but i'll ask again...was there any other option besides Oracle openjdk?
> 
> I once tried to change our docker image to use what i use on my system: https://sdkman.io/ which would then let us easy install whatever jdk, maven, etc. that we wanted. love that thing. i can't remember what problems I had but it never worked right. that might be a nice little thing to do if it's easy. not saying it necessarily needs to be done on this PR, but I guess I'd thought most devs moved on from Oracle JDK at this point and i thought temurin/corretto were what folks gravitated to.

Yes Oracle openjdk is the only option in the apk repository. I made a trivial attempt at using sdkman for the image but I ran into problems with it. If there is enough interest in using temurin or corretto I could look into it further but for now Oracle is the easiest and simplest option.

spmallette:
VOTE +1

xiazcy:
VOTE +1.
Thanks Cole!

PR #1822, opened 2022-10-06
Fix permission issues with Docker generated files

xiazcy:
Docker containers run and create files with root permission by default. Adding `chown` to change permission of Docker created files to current user, so `sudo` isn't needed for maven operations.


xiazcy:
Invoked CTR to merge this, as this is a minor change to dev testing environment only.

PR #1824, opened 2022-10-07
JIRA TINKERPOP-2809 bump jackson-databind to 2.14.0-rc1

acoady:


acoady:
This is closed in favour of deferring until 2.14 is released and not using a rc version

PR #1833, opened 2022-10-17
TINKERPOP-2814 Add configuration for SSL handshake timeout.

kenhuuu:
Resolves https://issues.apache.org/jira/browse/TINKERPOP-2814

Note: The default timeouts were increased because they may be too low and can contribute to connection exceptions. This increase can be removed if it is deemed that the defaults shouldn't be changed in 3.5-dev.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
@@ -35,6 +35,7 @@ import io.netty.handler.codec.http.websocketx.CloseWebSocketFrame;
 import io.netty.handler.codec.http.websocketx.WebSocketVersion;
 import io.netty.handler.codec.http.websocketx.extensions.compression.WebSocketClientCompressionHandler;
 import io.netty.handler.ssl.SslContext;
+import io.netty.handler.ssl.SslHandler;
 import io.netty.handler.timeout.IdleStateHandler;
 
 import org.slf4j.Logger;
@@ -86,6 +87,7 @@ public interface Channelizer extends ChannelHandler {
 
         protected static final String PIPELINE_GREMLIN_SASL_HANDLER = "gremlin-sasl-handler";
         protected static final String PIPELINE_GREMLIN_HANDLER = "gremlin-handler";
+        public static final String PIPELINE_SSL_HANDLER = "gremlin-ssl-handler";
 
         public boolean supportsSsl() {
             return cluster.connectionPoolSettings().enableSsl;
@@ -124,7 +126,12 @@ public interface Channelizer extends ChannelHandler {
             }
 
             if (sslCtx.isPresent()) {
-                pipeline.addLast(sslCtx.get().newHandler(socketChannel.alloc(), connection.getUri().getHost(), connection.getUri().getPort()));
+                SslHandler sslHandler = sslCtx.get().newHandler(socketChannel.alloc(), connection.getUri().getHost(), connection.getUri().getPort());
+                // TINKERPOP-2814. Remove the SSL handshake timeout so that handshakes that take longer than 10000ms
+                // (Netty default) but less than connectionSetupTimeoutMillis can succeed. This means the SSL handshake
+                // will instead be capped by connectionSetupTimeoutMillis.
+                sslHandler.setHandshakeTimeoutMillis(0);
+                pipeline.addLast(PIPELINE_SSL_HANDLER, sslHandler);
             }
 
             configure(pipeline);
@@ -187,7 +194,7 @@ public interface Channelizer extends ChannelHandler {
                     new WebSocketClientHandler.InterceptedWebSocketClientHandshaker13(
                             connection.getUri(), WebSocketVersion.V13, null, true,
                             httpHeaders, maxContentLength, true, false, -1,
-                            cluster.getHandshakeInterceptor()), cluster.getConnectionSetupTimeout());
+                            cluster.getHandshakeInterceptor()), cluster.getConnectionSetupTimeout(), supportsSsl());
 
             final int keepAliveInterval = toIntExact(TimeUnit.SECONDS.convert(
                     cluster.connectionPoolSettings().keepAliveInterval, TimeUnit.MILLISECONDS));
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
@@ -993,9 +993,6 @@ public final class Cluster {
         /**
          * Sets the duration of time in milliseconds provided for connection setup to complete which includes WebSocket
          * handshake and SSL handshake. Beyond this duration an exception would be thrown.
-         *
-         * Note that this value should be greater that SSL handshake timeout defined in
-         * {@link io.netty.handler.ssl.SslHandler} since WebSocket handshake include SSL handshake.
          */
         public Builder connectionSetupTimeoutMillis(final long connectionSetupTimeoutMillis) {
             this.connectionSetupTimeoutMillis = connectionSetupTimeoutMillis;
--- a/gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/WebSocketClientBehaviorIntegrateTest.java
+++ b/gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/WebSocketClientBehaviorIntegrateTest.java
@@ -26,6 +26,7 @@ import org.junit.rules.TestName;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.log4j.Level;
+import org.apache.tinkerpop.gremlin.driver.exception.NoHostAvailableException;
 import org.apache.tinkerpop.gremlin.driver.ser.Serializers;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.apache.tinkerpop.gremlin.util.Log4jRecordingAppender;
@@ -41,6 +42,7 @@ import static org.hamcrest.core.Is.is;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 
 public class WebSocketClientBehaviorIntegrateTest {
     @Rule
@@ -67,7 +69,12 @@ public class WebSocketClientBehaviorIntegrateTest {
         rootLogger.addAppender(recordingAppender);
 
         server = new SimpleSocketServer();
-        server.start(new TestWSGremlinInitializer());
+        if (name.getMethodName().equals("shouldAttemptHandshakeForLongerThanDefaultNettySslHandshakeTimeout") ||
+                name.getMethodName().equals("shouldPrintCorrectErrorForRegularWebSocketHandshakeTimeout")) {
+            server.start(new TestWSNoOpInitializer());
+        } else {
+            server.start(new TestWSGremlinInitializer());
+        }
     }
 
     @After
@@ -308,5 +315,60 @@ public class WebSocketClientBehaviorIntegrateTest {
                 recordingAppender.getMessages().stream()
                         .filter(str -> str.contains("Considering new connection on"))
                         .count());
+    }
+
+    /**
+     * (TINKERPOP-2814) Tests to make sure that the SSL handshake is now capped by connectionSetupTimeoutMillis and not
+     * the default Netty SSL handshake timeout of 10,000ms.
+     */
+    @Test
+    public void shouldAttemptHandshakeForLongerThanDefaultNettySslHandshakeTimeout() {
+        final Cluster cluster = Cluster.build("localhost").port(SimpleSocketServer.PORT)
+                .minConnectionPoolSize(1)
+                .maxConnectionPoolSize(1)
+                .connectionSetupTimeoutMillis(20000) // needs to be larger than 10000ms.
+                .enableSsl(true)
+                .create();
+
+        final Client.ClusteredClient client = cluster.connect();
+        final long start = System.currentTimeMillis();
+
+        Exception caught = null;
+        try {
+            client.submit("1");
+        } catch (Exception e) {
+            caught = e;
+        } finally {
+            // Test against 15000ms which should give a big enough buffer to avoid timing issues.
+            assertTrue(System.currentTimeMillis() - start > 15000);
+            assertTrue(caught != null);
+            assertTrue(caught instanceof NoHostAvailableException);
+            assertTrue(recordingAppender.getMessages().stream().anyMatch(str -> str.contains("SSL handshake not completed")));
+        }
+    }
+
+    /**
+     * Tests to make sure that the correct error message is logged when a non-SSL connection attempt times out.
+     */
+    @Test
+    public void shouldPrintCorrectErrorForRegularWebSocketHandshakeTimeout() {
+        final Cluster cluster = Cluster.build("localhost").port(SimpleSocketServer.PORT)
+                .minConnectionPoolSize(1)
+                .maxConnectionPoolSize(1)
+                .connectionSetupTimeoutMillis(100)
+                .create();
+
+        final Client.ClusteredClient client = cluster.connect();
+
+        Exception caught = null;
+        try {
+            client.submit("1");
+        } catch (Exception e) {
+            caught = e;
+        } finally {
+            assertTrue(caught != null);
+            assertTrue(caught instanceof NoHostAvailableException);
+            assertTrue(recordingAppender.getMessages().stream().anyMatch(str -> str.contains("WebSocket handshake not completed")));
+        }
     }
 }
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -47,6 +47,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Bumped to Groovy 2.5.19.
 * Bumped to Netty 4.1.85.
 * Bumped `ivy` to 2.5.1 to fix security vulnerability
+* Removed default SSL handshake timeout. The SSL handshake timeout will instead be capped by setting `connectionSetupTimeoutMillis`.
 
 ==== Bugs
 
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Settings.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Settings.java
@@ -407,9 +407,6 @@ final class Settings {
          * Duration of time in milliseconds provided for connection setup to complete which includes WebSocket
          * handshake and SSL handshake. Beyond this duration an exception would be thrown if the handshake is not
          * complete by then.
-         *
-         * Note that this value should be greater that SSL handshake timeout defined in
-         * {@link io.netty.handler.ssl.SslHandler} since WebSocket handshake include SSL handshake.
          */
         public long connectionSetupTimeoutMillis = Connection.CONNECTION_SETUP_TIMEOUT_MILLIS;
     }
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/WebSocketClientHandler.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/WebSocketClientHandler.java
@@ -18,6 +18,7 @@
  */
 package org.apache.tinkerpop.gremlin.driver.handler;
 
+import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPromise;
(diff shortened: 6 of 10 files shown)


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
```diff
@@ -205,6 +205,7 @@ private static Builder getBuilderFromSettings(final Settings settings) {
                 .maxConnectionPoolSize(settings.connectionPool.maxSize)
                 .minConnectionPoolSize(settings.connectionPool.minSize)
                 .connectionSetupTimeoutMillis(settings.connectionPool.connectionSetupTimeoutMillis)
+                .sslHandshakeTimeoutMillis(settings.connectionPool.sslHandshakeTimeoutMillis)
```
divijvaidya, line 208: Instead of adding yet another configuration for the users to tweak on the client, is it possible to define this as a percentage of `connectionSetupTimeoutMillis`?

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
```diff
@@ -986,13 +996,26 @@ public Builder port(final int port) {
          * handshake and SSL handshake. Beyond this duration an exception would be thrown.
          *
          * Note that this value should be greater that SSL handshake timeout defined in
-         * {@link io.netty.handler.ssl.SslHandler} since WebSocket handshake include SSL handshake.
+         * {@link io.netty.handler.ssl.SslHandler} since WebSocket handshake include SSL handshake. This value should be
+         * less than {@link Builder#maxWaitForConnection}.
```
divijvaidya, line 1000: please update comment to include `connectionSetupTimeoutMillis`

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java
```diff
@@ -59,14 +59,15 @@ final class Connection {
 
     public static final int MAX_IN_PROCESS = 4;
     public static final int MIN_IN_PROCESS = 1;
-    public static final int MAX_WAIT_FOR_CONNECTION = 16000;
+    public static final int MAX_WAIT_FOR_CONNECTION = 25000;
```
divijvaidya, line 62: Changing default could be a breaking change for users. As an example, a user may be relying on connection setup to be at 15s based on which they may have configured throttling rules in their application. 

I would suggest to target the change in default values for v3.7.x

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
kenhuuu: I understand the concern of there being too many configuration options for the user already, but I'm not sure if there is a reasonable default for this as a percentage of connectionSetupTimeoutMillis. Would it make sense then to just set this value to 0 (no timeout). This would allow the user to focus on just setting a proper value for connectionSetupTimeoutMillis which will cap the SSL handshake anyway.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java
kenhuuu: This seems like more of a behavioral change than a complete breaking change. However, if you think this is still too big of a behavioral change then I will remove it and we can look to change it in 3.7 as you suggested.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
spmallette: i'm pretty sure that we dont have any other timeouts interrelated to `connectionSetupTimeoutMillis` (but please double check my recollection) so I think I'm ok with using a percentage here. if we do it as a percentage, then i'd be ok with 0 as the default. that seems to make sense.

divijvaidya: 0 makes sense to me as well.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java
divijvaidya: You are right, it's not an interface / contract change but if we release this as a patch change, a customer may have to change their application code to consume the patch version. 

As an example, let's say that the customer has set exponential backoff/retry for connection setup based on the current default values but now we are changing the default values and their tuning for retries would break. That is the reason, I would like to add this to 3.7.x.

@spmallette thoughts?

spmallette: I'm fine with it waiting for 3.7.x.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
kenhuuu: Hi Divij, could you clarify some recent comments? Did you mean to say that we shouldn't add another configuration for the SSL timeout and instead we should just internally set this to a percentage of connectionSetupTimeoutMillis? Or, did you mean that we should add a configuration but it should be percentage based and the default could be 0.

kenhuuu: It was also brought to my attention that a potential issue with setting the default to 0 is that it might mask the underlying cause of the timeout issue. The user would see that there was a connection setup timeout but they wouldn't know that it was caused by a timeout in the SSL handshake. Netty won't throw a SSL handshake related exception if the value is set to 0.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
kenhuuu: Could you clarify this? Did you mean adding a param line? If so, I didn't do this as it seemed a bit redundant and it isn't done for any of the other params in this class.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
kenhuuu: @divijvaidya do you have any more thoughts?

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
divijvaidya: I am sorry sorry, copy/paste attempt resulted in wrong comment. 

What I wanted to say was to update the java doc comment since "Note that this value should be greater that SSL handshake timeout defined in SslHandler" is not true anymore. We should replace this with information with the new timeout we are introducing `sslHandshakeTimeoutMillis`

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
kenhuuu: Hi @divijvaidya could you please follow up on these questions? Thanks.

divijvaidya: Hey @kenhuuu 

Thank you for your patience so far. I am trying to explore the possibility of having only one timeout that the customer has to manage i.e. the connection timeout. As you rightly pointed out, we need to ensure that our implementation will not lose details of where the timeout occurred.

The `SslHandler` class in Netty has public methods which allow to check whether the ssl handshake has been completed. For example, one can use `SslHandler#handshakeFuture.isComplete()` to check whether the handshake has been completed or not. Alternatively one can listen to `SslHandshakeCompletionEvent` as mentioned in https://netty.io/4.1/api/io/netty/handler/ssl/SslHandler.html

> Beside using the handshake [ChannelFuture](https://netty.io/4.1/api/io/netty/channel/ChannelFuture.html) to get notified about the completion of the handshake it's also possible to detect it by implement the [ChannelInboundHandler.userEventTriggered(ChannelHandlerContext, Object)](https://netty.io/4.1/api/io/netty/channel/ChannelInboundHandler.html#userEventTriggered-io.netty.channel.ChannelHandlerContext-java.lang.Object-) method and check for a [SslHandshakeCompletionEvent](https://netty.io/4.1/api/io/netty/handler/ssl/SslHandshakeCompletionEvent.html). 

Hence, we could potentially do an implementation such as the following in our client:
1. Set SslHandler#setHandshakeTimeout to 0 [which means that handshake is not bound by the timeout](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SslHandler.java#L2105). 
2. Set a timeout on overall connection setup (which we already do).
3. When timeout hits for overall connection setup, check what stage is the connection setup at? Is the ssl handshake complete (can be checked using the future as as mentioned above)? If not, then cancel it. Is the websocket handshake complete? Is the SASL handshake complete?
4. Depending on the stage where the overall connection is at, we can generate an appropriate error message to the application.

What do you think about this approach?

kenhuuu: Yes, I think this is a good approach. Thank you for the suggestion.

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/WebSocketClientBehaviorIntegrateTest.java
```diff
@@ -271,4 +276,28 @@ public void shouldNotCreateReplacementConnectionWhenClientClosesConnection() thr
...
+                .create();
+
+        final Client.ClusteredClient client = cluster.connect();
+        final long start = System.currentTimeMillis();
+
+        try {
+            client.submit("1");
+        } finally {
+            // Test against 15,000ms which should give a big enough buffer to avoid timing issues.
+            assertTrue(System.currentTimeMillis() - start > 15000);
```
divijvaidya, line 300: If I understand this correctly, we expect the connection setup to fail and we verify that the failure is after the threshold of ssl handshake timeout, hence proving that the timeout is bound by the overall connection setup limit and not by ssl. Is that right?

Could we also add an assertion that we expect the connection setup to fail? I know that it is not what this test is checking but it will help guard the purpose of this test is future cases. Also let's validate that the correct error message is being printed and we haven't regressed in the aspect of providing the detailed/correct error log.

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/WebSocketClientBehaviorIntegrateTest.java
```diff
@@ -271,4 +276,28 @@ public void shouldNotCreateReplacementConnectionWhenClientClosesConnection() thr
...
+                .enableSsl(true)
+                .create();
+
+        final Client.ClusteredClient client = cluster.connect();
+        final long start = System.currentTimeMillis();
+
+        try {
+            client.submit("1");
+        } finally {
+            // Test against 15,000ms which should give a big enough buffer to avoid timing issues.
```
divijvaidya, line 299: nit
not important but FYI, in some geographies (such as Germany) a comma in numbers is read as decimal and folks in Germany will probably read this as 15ms! Not using comma at all keeps things independent of such geographic irregularities.

divijvaidya (commented):
This looks much better now. Thank you!

I added some minor comments, once we are through them (and test suite passes), we should be ready to merge,

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/WebSocketClientBehaviorIntegrateTest.java
kenhuuu: Yes, your understanding of the test is correct. Sure, I'll add those assertions.

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/WebSocketClientBehaviorIntegrateTest.java
kenhuuu: Thanks for the reminder. I'll remove those commas.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
```diff
@@ -123,7 +124,9 @@ protected void initChannel(final SocketChannel socketChannel) throws Exception {
             }
 
             if (sslCtx.isPresent()) {
-                pipeline.addLast(sslCtx.get().newHandler(socketChannel.alloc(), connection.getUri().getHost(), connection.getUri().getPort()));
+                SslHandler sslHandler = sslCtx.get().newHandler(socketChannel.alloc(), connection.getUri().getHost(), connection.getUri().getPort());
+                sslHandler.setHandshakeTimeoutMillis(0);
```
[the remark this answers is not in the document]
kenhuuu: I recently noticed an exception that is being thrown from the SslHandler when setting this value to 0.

`[ERROR] org.apache.tinkerpop.gremlin.driver.Handler$GremlinResponseHandler - Could not process the response
io.netty.handler.ssl.SslClosedEngineException: SSLEngine closed already
	at io.netty.handler.ssl.SslHandler.wrap(SslHandler.java:859)
	at io.netty.handler.ssl.SslHandler.wrapAndFlush(SslHandler.java:798)
	at io.netty.handler.ssl.SslHandler.flush(SslHandler.java:779)
	at io.netty.handler.ssl.SslHandler.flush(SslHandler.java:1971)
	at io.netty.handler.ssl.SslHandler.closeOutboundAndChannel(SslHandler.java:1940)
	at io.netty.handler.ssl.SslHandler.close(SslHandler.java:730)
	at io.netty.channel.AbstractChannelHandlerContext.invokeClose(AbstractChannelHandlerContext.java:622)
	at io.netty.channel.AbstractChannelHandlerContext.close(AbstractChannelHandlerContext.java:606)
	at io.netty.channel.CombinedChannelDuplexHandler$DelegatingChannelHandlerContext.close(CombinedChannelDuplexHandler.java:505)
	at io.netty.channel.ChannelOutboundHandlerAdapter.close(ChannelOutboundHandlerAdapter.java:77)
	at io.netty.channel.CombinedChannelDuplexHandler.close(CombinedChannelDuplexHandler.java:316)
	at io.netty.channel.AbstractChannelHandlerContext.invokeClose(AbstractChannelHandlerContext.java:622)
	at io.netty.channel.AbstractChannelHandlerContext.close(AbstractChannelHandlerContext.java:606)
	at io.netty.channel.AbstractChannelHandlerContext.close(AbstractChannelHandlerContext.java:472)
	at io.netty.handler.codec.http.websocketx.WebSocketClientProtocolHandshakeHandler$2.run(WebSocketClientProtocolHandshakeHandler.java:122)
	at io.netty.util.concurrent.PromiseTask.runTask(PromiseTask.java:98)
	at io.netty.util.concurrent.ScheduledFutureTask.run(ScheduledFutureTask.java:170)
	at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:174)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:167)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:470)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:503)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:995)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at java.base/java.lang.Thread.run(Thread.java:829)`

I've tried several workarounds to get rid of this issue with closing the SslHandler. These workarounds include subclassing SslHandler to modify how close works if the handshake times out, removing the SslHandler from the pipeline and swallowing the exception if the handshake failed. However, all of these workarounds have drawbacks so instead I propose that we go back to your initial recommendation which is to make the SSL handshake timeout a percentage of `connectionSetupTimeoutMillis`. I feel like 0.9*connectionSetupTimeoutMillis would still enable configurability for this value while preventing any timing issues with the WebSocket handshake timeout.

What are you thoughts on this @divijvaidya ? Thanks.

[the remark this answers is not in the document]
divijvaidya: Please tell me more. I don't think I understood the cause of the exception above.

From the above exception stack trace, it looks like when we try to close SslHandler (during termination of a connection), it is already closed and hence, throws an exception. Why is it already closed? What does setting the timeout to 0 vs. say 100 has to do with it?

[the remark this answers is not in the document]
kenhuuu: That "SSLEngine closed already" exception is just a general error message because the handshake future wasn't given a specific cause. When updating the handshake future with an SSLHandshakeException, the stack trace looks like

`[ERROR] org.apache.tinkerpop.gremlin.driver.Handler$GremlinResponseHandler - Could not process the response
javax.net.ssl.SSLHandshakeException: SSL handshake timed out.
	at org.apache.tinkerpop.gremlin.driver.handler.WebSocketClientHandler.userEventTriggered(WebSocketClientHandler.java:127)
	at io.netty.channel.AbstractChannelHandlerContext.invokeUserEventTriggered(AbstractChannelHandlerContext.java:346)
	at io.netty.channel.AbstractChannelHandlerContext.invokeUserEventTriggered(AbstractChannelHandlerContext.java:332)
	at io.netty.channel.AbstractChannelHandlerContext.fireUserEventTriggered(AbstractChannelHandlerContext.java:324)
	at io.netty.handler.codec.http.websocketx.WebSocketClientProtocolHandshakeHandler$2.run(WebSocketClientProtocolHandshakeHandler.java:121)
	at io.netty.util.concurrent.PromiseTask.runTask(PromiseTask.java:98)
	at io.netty.util.concurrent.ScheduledFutureTask.run(ScheduledFutureTask.java:170)
	at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:174)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:167)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:470)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:503)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:995)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at java.base/java.lang.Thread.run(Thread.java:829)`

For more information, this is what the 3.5.4 driver logs show:
`[ERROR] org.apache.tinkerpop.gremlin.driver.Handler$GremlinResponseHandler - Could not process the response
io.netty.handler.ssl.SslHandshakeTimeoutException: handshake timed out after 10000ms
	at io.netty.handler.ssl.SslHandler$7.run(SslHandler.java:2113)
	at io.netty.util.concurrent.PromiseTask.runTask(PromiseTask.java:98)
	at io.netty.util.concurrent.ScheduledFutureTask.run(ScheduledFutureTask.java:170)
	at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:174)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:167)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:470)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:503)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:995)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at java.base/java.lang.Thread.run(Thread.java:829)`

If you think this new stack trace is fine then we can keep the default SSL handshake timeout as 0. Otherwise, it may be cleaner to just use a percentage of connectionSetupTimeoutMillis. And just to be clear, this is just an exception that is logged in logs. The user won't see this exception as it is part of the close/cleanup process happening in the background.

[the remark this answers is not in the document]
divijvaidya: I think that the new stack trace i.e. `[ERROR] org.apache.tinkerpop.gremlin.driver.Handler$GremlinResponseHandler - Could not process the response javax.net.ssl.SSLHandshakeException: SSL handshake timed out` is fine as it provides the reason for connection failure i.e. ssl timeout.

I am saying this because the alternative to go for a %age based timeout adds another knob that the users have to tune and they don't have good heuristics to tune it since the %age is a function of network weather at that moment. That is the reason I have an inclination to avoid adding a new configuration or adding a %age based solution.

kenhuuu:
I think I've addressed the remaining issues. Are there any other comments or issues?

xiazcy:
VOTE +1. Given the amount of comments and resolved issues, we are merging this today.

cole-bq:
LGTM

PR #1837, opened 2022-10-20
TINKERPOP-2806 Provide method for provider plugins to get notified on script/query processing

tkolanko:
This PR adds 3 new default methods to the GraphManager interface allowing custom GraphManagers to opt into receiving notifications based on the lifecycle of scripts/traversals.

AbstractEvalOpProcessor and TraversalOpProcessor get the GraphManager from context and call the lifecycle methods as needed.

I created a proposal thread on https://lists.apache.org/thread/b3k8b4j29gswyk39pl4cfsvd5ofozh4l but there wasn't much interaction, I'm hoping we can get some more discussion from a PR

### DIFF
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractEvalOpProcessor.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractEvalOpProcessor.java
@@ -236,11 +236,7 @@ public abstract class AbstractEvalOpProcessor extends AbstractOpProcessor {
                     graphManager.onQueryError(msg, t);
                     if (managedTransactionsForRequest) attemptRollback(msg, ctx.getGraphManager(), settings.strictTransactionManagement);
                 })
-                .afterTimeout(b -> {
-                  final String errorMessage = String.format(
-                  "A timeout occurred within the script during evaluation of [%s] - consider increasing the limit given to TimedInterruptCustomizerProvider",
-                  msg);
-                  TimeoutException t = new TimeoutException(errorMessage);
+                .afterTimeout((b, t) -> {
                   graphManager.onQueryError(msg, t);
                 })
                 .beforeEval(b -> {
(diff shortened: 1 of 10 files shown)


gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractEvalOpProcessor.java
```diff
@@ -231,9 +233,18 @@ protected void evalOpInternal(final Context ctx, final Supplier<GremlinExecutor>
         final GremlinExecutor.LifeCycle lifeCycle = GremlinExecutor.LifeCycle.build()
                 .evaluationTimeoutOverride(seto)
                 .afterFailure((b,t) -> {
+                    graphManager.onQueryError(msg, t);
                     if (managedTransactionsForRequest) attemptRollback(msg, ctx.getGraphManager(), settings.strictTransactionManagement);
                 })
+                .afterTimeout(b -> {
```
spmallette, line 239: this is unfortunate. i assume there is no way to get the original time-out exception thrown? do we really have to construct a fresh exception here just for purpose of getting one to `onQueryError()`?

spmallette: sorry for the delay in reviewing your latest changes. unfortunately, i think the breaking change forces this into 3.7.x rather than 3.6.x. there are several providers who rely on this API. Maybe you could turn it back into a non-breaking change though? Since it is just a callback, you could add the new signature for `afterTimeout()` as an overload. You could then have `GremlinExecutor` call both. Deprecate the `Consumer` version and ensure the javadoc makes clear that users should only utilize one of those two methods as a single timeout triggers them both. Would that work?

tkolanko: @spmallette In https://github.com/apache/tinkerpop/pull/1837/commits/718c0830d0725655b8c5999832e12f917991a9fc I deprecated the use of `afterTimeout` as a consumer and I had the deprecated method create a new `BiConsumer` which calls the updated `afterTimeout` method. I also reverted the `GremlinExecutor` tests to use the deprecated method to make sure they are still working, I can either move them to the new method or leave things as they are

spmallette:
Cherry-picked as:

24e74da0acd07f0dae7087acdccaece9a4d54533
ca9106926a7f4dbb6720da48df2509fe71bb1251
f13ce6e1ed583fd798978b836665cdb07c737f14

and then added an extra commit with some little fixes/docs: cf211fd59c3352ee7a2f1a38d9992df695939b07

sorry this took a while to get merged. thanks for your contribution.

PR #1838, opened 2022-10-20
TINKERPOP-2480 Add User-Agent request header to Java Driver and Server

cole-bq:

Adds a user agent to gremlin-driver which is sent during the web socket handshake.

This behaviour is enabled by default but can be disabled by setting the enableWsHandshakeUserAgent configuration to false.

### DIFF
--- a/docs/src/dev/provider/index.asciidoc
+++ b/docs/src/dev/provider/index.asciidoc
@@ -848,6 +848,14 @@ Server returns for a single request.  Again, this description of Gremlin Server'
 out-of-the-box configuration.  It is quite possible to construct other flows, that might be more amenable to a
 particular language or style of processing.
 
+It is recommended but not required that a driver include a `User-Agent` header as part of any web socket
+handshake request to Gremlin Server. Gremlin Server uses the user agent in building usage metrics
+as well as debugging. The standard format for connection user agents is:
+
+`"[Application Name] [GLV Name].[Version] [Language Runtime Version] [OS].[Version] [CPU Architecture]"`
+For example:
+`"MyTestApplication Gremlin-Java.3.5.4 11.0.16.1 Mac_OS_X.12.6.1 aarch64"`
+
 To formulate a request to Gremlin Server, a `RequestMessage` needs to be constructed.  The `RequestMessage` is a
 generalized representation of a request that carries a set of "standard" values in addition to optional ones that are
 dependent on the operation being performed.  A `RequestMessage` has these fields:
--- a/docs/src/reference/gremlin-variants.asciidoc
+++ b/docs/src/reference/gremlin-variants.asciidoc
@@ -329,6 +329,7 @@ The following table describes the various configuration options for the Gremlin 
 |serializer.config |A `Map` of configuration settings for the serializer. |_none_
 |username |The username to submit on requests that require authentication. |_none_
 |workerPoolSize |Size of the pool for handling background work. |available processors * 2
+|enableUserAgentOnConnect |Enables sending a user agent to the server during connection requests. More details can be found in provider docs link:https://tinkerpop.apache.org/docs/x.y.z/dev/provider/#_graph_driver_provider_requirements[here].|true
 |=========================================================
 
 Please see the link:https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gremlin/driver/Cluster.Builder.html[Cluster.Builder javadoc] to get more information on these settings.
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
@@ -27,10 +27,11 @@ import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelPipeline;
 import io.netty.channel.socket.SocketChannel;
-import io.netty.handler.codec.http.EmptyHttpHeaders;
-import io.netty.handler.codec.http.websocketx.CloseWebSocketFrame;
+import io.netty.handler.codec.http.HttpHeaders;
 import io.netty.handler.codec.http.HttpClientCodec;
 import io.netty.handler.codec.http.HttpObjectAggregator;
+import io.netty.handler.codec.http.DefaultHttpHeaders;
+import io.netty.handler.codec.http.websocketx.CloseWebSocketFrame;
 import io.netty.handler.codec.http.websocketx.WebSocketVersion;
 import io.netty.handler.codec.http.websocketx.extensions.compression.WebSocketClientCompressionHandler;
 import io.netty.handler.ssl.SslContext;
@@ -178,10 +179,14 @@ public interface Channelizer extends ChannelHandler {
                 throw new IllegalStateException("To use wss scheme ensure that enableSsl is set to true in configuration");
 
             final int maxContentLength = cluster.connectionPoolSettings().maxContentLength;
+            final HttpHeaders httpHeaders = new DefaultHttpHeaders();
+            if(connection.getCluster().isUserAgentOnConnectEnabled()) {
+                httpHeaders.set(UserAgent.USER_AGENT_HEADER_NAME, UserAgent.USER_AGENT);
+            }
             handler = new WebSocketClientHandler(
                     new WebSocketClientHandler.InterceptedWebSocketClientHandshaker13(
-                            connection.getUri(), WebSocketVersion.V13, null,true,
-                            EmptyHttpHeaders.INSTANCE, maxContentLength, true, false, -1,
+                            connection.getUri(), WebSocketVersion.V13, null, true,
+                            httpHeaders, maxContentLength, true, false, -1,
                             cluster.getHandshakeInterceptor()), cluster.getConnectionSetupTimeout());
 
             final int keepAliveInterval = toIntExact(TimeUnit.SECONDS.convert(
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
@@ -34,7 +34,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.net.ssl.KeyManagerFactory;
-import javax.net.ssl.TrustManager;
 import javax.net.ssl.TrustManagerFactory;
 
 import java.io.File;
@@ -205,6 +204,7 @@ public final class Cluster {
                 .maxConnectionPoolSize(settings.connectionPool.maxSize)
                 .minConnectionPoolSize(settings.connectionPool.minSize)
                 .connectionSetupTimeoutMillis(settings.connectionPool.connectionSetupTimeoutMillis)
+                .enableUserAgentOnConnect(settings.enableUserAgentOnConnect)
                 .validationRequest(settings.connectionPool.validationRequest);
 
         if (settings.username != null && settings.password != null)
@@ -561,6 +561,14 @@ public final class Cluster {
         return builder.build();
     }
 
+    /**
+     * Checks if cluster is configured to send a User Agent header
+     * in the web socket handshake
+     */
+    public boolean isUserAgentOnConnectEnabled() {
+        return manager.isUserAgentOnConnectEnabled();
+    }
+
     public final static class Builder {
         private List<InetAddress> addresses = new ArrayList<>();
         private int port = 8182;
@@ -597,6 +605,7 @@ public final class Cluster {
         private HandshakeInterceptor interceptor = HandshakeInterceptor.NO_OP;
         private AuthProperties authProps = new AuthProperties();
         private long connectionSetupTimeoutMillis = Connection.CONNECTION_SETUP_TIMEOUT_MILLIS;
+        private boolean enableUserAgentOnConnect = true;
 
         private Builder() {
             // empty to prevent direct instantiation
@@ -993,6 +1002,16 @@ public final class Cluster {
             return this;
         }
 
+        /**
+         * Configures whether cluster will send a user agent during
+         * web socket handshakes
+         * @param enableUserAgentOnConnect true enables the useragent. false disables the useragent.
+         */
+        public Builder enableUserAgentOnConnect(final boolean enableUserAgentOnConnect) {
+            this.enableUserAgentOnConnect = enableUserAgentOnConnect;
+            return this;
+        }
+
         List<InetSocketAddress> getContactPoints() {
             return addresses.stream().map(addy -> new InetSocketAddress(addy, port)).collect(Collectors.toList());
         }
@@ -1045,6 +1064,7 @@ public final class Cluster {
         private final int workerPoolSize;
         private final int port;
         private final String path;
+        private final boolean enableUserAgentOnConnect;
 
         private final AtomicReference<CompletableFuture<Void>> closeFuture = new AtomicReference<>();
 
@@ -1057,6 +1077,7 @@ public final class Cluster {
             this.authProps = builder.authProps;
             this.contactPoints = builder.getContactPoints();
             this.interceptor = builder.interceptor;
+            this.enableUserAgentOnConnect = builder.enableUserAgentOnConnect;
 
             connectionPoolSettings = new Settings.ConnectionPoolSettings();
             connectionPoolSettings.maxInProcessPerConnection = builder.maxInProcessPerConnection;
@@ -1228,6 +1249,14 @@ public final class Cluster {
         @Override
         public String toString() {
             return String.join(", ", contactPoints.stream().map(InetSocketAddress::toString).collect(Collectors.<String>toList()));
+        }
+
+        /**
+         * Checks if cluster is configured to send a User Agent header
+         * in the web socket handshake
+         */
+        public boolean isUserAgentOnConnectEnabled() {
+            return enableUserAgentOnConnect;
         }
     }
 }
--- a/gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/WebSocketClientBehaviorIntegrateTest.java
+++ b/gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/WebSocketClientBehaviorIntegrateTest.java
@@ -89,6 +89,44 @@ public class WebSocketClientBehaviorIntegrateTest {
     }
 
     /**
+     * Tests that client is correctly sending user agent during web socket handshake by having the server return
+     * the captured user agent.
+     */
+    @Test
+    public void shouldIncludeUserAgentInHandshakeRequest() {
+        final Cluster cluster = Cluster.build("localhost").port(SimpleSocketServer.PORT)
+                .minConnectionPoolSize(1)
+                .maxConnectionPoolSize(1)
+                .serializer(Serializers.GRAPHSON_V2D0)
+                .create();
+        final Client.ClusteredClient client = cluster.connect();
+
+        // trigger the testing server to return captured user agent
+        String returnedUserAgent = client.submit("1", RequestOptions.build()
+                        .overrideRequestId(TestWSGremlinInitializer.USER_AGENT_REQUEST_ID).create()).one().getString();
+        assertEquals(UserAgent.USER_AGENT, returnedUserAgent);
+    }
+
+    /**
+     * Tests that no user agent is sent to server when that behaviour is disabled.
+     */
+    @Test
+    public void shouldNotIncludeUserAgentInHandshakeRequestIfDisabled() {
+        final Cluster cluster = Cluster.build("localhost").port(SimpleSocketServer.PORT)
+                .minConnectionPoolSize(1)
+                .maxConnectionPoolSize(1)
+                .serializer(Serializers.GRAPHSON_V2D0)
+                .enableUserAgentOnConnect(false)
+                .create();
+        final Client.ClusteredClient client = cluster.connect();
+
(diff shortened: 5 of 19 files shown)


spmallette:
i realize this feature is focused on the Java driver, but perhaps we could make the server do something with the agent in the PR? could we at least have a debug level log message output a line with the requestid/agent value? or perhaps we dont need a discrete message? maybe just update the server logging that currently exists for writing a `RequestMessage` to include some of that information where it might be helpful in debugging?

alexey-temnikov:
@spmallette, absolutely, this is intent. We also wanted to explore options if we can add client version information to the exception that occurred in the server to help with troubleshooting.

vkagamlyk:
VOTE +1

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
```diff
@@ -178,10 +180,14 @@ public void configure(final ChannelPipeline pipeline) {
                 throw new IllegalStateException("To use wss scheme ensure that enableSsl is set to true in configuration");
 
             final int maxContentLength = cluster.connectionPoolSettings().maxContentLength;
+            HttpHeaders httpHeaders = new DefaultHttpHeaders();
```
spmallette, line 183: nit: please review changes to make sure variables that can be `final` are marked as such.

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -329,6 +329,7 @@ The following table describes the various configuration options for the Gremlin
 |serializer.config |A `Map` of configuration settings for the serializer. |_none_
 |username |The username to submit on requests that require authentication. |_none_
 |workerPoolSize |Size of the pool for handling background work. |available processors * 2
+|enableWsHandshakeUserAgent |Configures the channelizer to send a user agent during web socket handshakes. |true
```
spmallette, line 332: 1. It doesn't necessarily have to be here in this table, but i think the user will need more details as to what effect this setting has. 
2. `enableWsHandshakeUserAgent` says what it is but isn't really clear at first read? maybe `enableUserAgentOnConnect` would be more descriptive?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/WsUserAgentHandler.java
```diff
@@ -0,0 +1,54 @@
...
+import io.netty.util.AttributeKey;
+import org.apache.tinkerpop.gremlin.driver.util.UserAgent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Channel handler which extracts a user agent header from a web socket handshake if present
+ * then logs the user agent and stores it as a channel attribute for future reference.
+ */
+public class WsUserAgentHandler extends ChannelInboundHandlerAdapter {
```
spmallette, line 34: is `info` the right log level here? seems we would trigger this for each connection made by a driver. maybe that would flood the logs with too much stuff? maybe we `debug` logging is better and some sort of aggregation of agent types reported as metrics (which by default only periodically writes to the logs)? thoughts?

cole-bq: I agree, best to keep logs clearer and aggregated user agents will be more useful too.

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -682,6 +683,14 @@ This command will generate a new Maven project in a directory called "app" with
 `com.my`. Please see the `README.asciidoc` in the root of each generated project for information on how to build and
 execute it.
 
+[[gremlin-java-user-agent]]
+=== User Agent
```
spmallette, line 687: i wouldn't make this a separate top-level section as GLVs have a set form. i was going to suggest you could make it a fourth level section under `gremlin-java-configuration` but i feel like this the "user agent" is just going to then have to be replicated for each GLV. I'm starting to think about this differently now. This new use of user agent is technically a standard that drivers should adhere to. I think you should document it as such in the [Provider Documentation](https://github.com/apache/tinkerpop/blob/master/docs/src/dev/provider/index.asciidoc#graph-driver-provider-requirements) where you should include the expected format, examples, etc. Then you can just add a link to that from your little configuration description if folks want to know what the "use agent" is. Finally in the Upgrade Documentation you should make mention of this new feature in the Provider section to call attention to it. how does that sound?

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
```diff
@@ -993,6 +1002,16 @@ public Builder connectionSetupTimeoutMillis(final long connectionSetupTimeoutMil
             return this;
         }
 
+        /**
+         * Configures whether cluster will send a user agent during
+         * web socket handshakes
+         * @param enableUserAgentOnConnect true enables the useragent. false disables the useragent.
+         */
+        public Builder enableUserAgentOnConnect(boolean enableUserAgentOnConnect) {
+            this.wsHandshakeUserAgentEnabled = enableUserAgentOnConnect;
```
spmallette, line 1011: it seems like you went with the rename i suggested - please rename the underlying variables as well to keep things consistent.

nit: `final`

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/WsUserAgentHandler.java
spmallette: you went with `UserAgent` for the metric name, but please double check to see if that is consistent with patterns already established for naming metrics in Gremlin Server. i think naming prefixes are all lower case with a hyphen between words if there are more than one.  also, since you added a metric, please document it [here](https://tinkerpop.apache.org/docs/current/reference/#metrics).

cole-bq: I've updated the name to follow the existing pattern. Unfortunately because the user agent itself is being used as the final part of the metric name, that portion of it will not match the existing format.

spmallette: > that portion of it will not match the existing format.

I believe that is fine as I think other metrics work that way as well. as long as the static prefix part of the metric is named according to standard the dynamic right-hand-side can do as it wishes.

spmallette:
VOTE +1

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/WebSocketClientBehaviorIntegrateTest.java
```diff
@@ -88,6 +89,44 @@ public void shutdown() {
         rootLogger.removeAppender(recordingAppender);
     }
 
+    /**
+     * Tests that client is correctly sending user agent during web socket handshake by having the server return
+     * the captured user agent.
+     */
+    @Test
+    public void shouldIncludeUserAgentInHandshakeRequest() {
```
divijvaidya, line 97: Please add the following test scenarios:

1. Validate that the user agent is sent only once per WebSocket connection.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/WsUserAgentHandler.java
```diff
@@ -0,0 +1,60 @@
...
+    @Override
+    public void userEventTriggered(ChannelHandlerContext ctx, java.lang.Object evt){
+        if(evt instanceof WebSocketServerProtocolHandler.HandshakeComplete){
+            final HttpHeaders requestHeaders = ((WebSocketServerProtocolHandler.HandshakeComplete) evt).requestHeaders();
+
+            if(requestHeaders.contains(UserAgent.USER_AGENT_HEADER_NAME)){
+                final String userAgent = requestHeaders.get(UserAgent.USER_AGENT_HEADER_NAME);
+
+                ctx.channel().attr(USER_AGENT_ATTR_KEY).set(userAgent);
+                logger.debug(String.format("New Connection on channel [%s] with user agent [%s]", ctx.channel().id().asShortText(), userAgent));
```
divijvaidya, line 51: I would recommend to use [slf4j's parameterized format for logging with](https://www.slf4j.org/faq.html#logging_performance) `{}`. That form ensures that the arguments in the log line are not evaluated is the log level is not satisfied. In the correct scenario, String.format() will always be calculated irrespective of the logging level set for the server.

Same comment for other log lines.

docs/src/dev/provider/index.asciidoc
```diff
@@ -848,6 +848,14 @@ Server returns for a single request.  Again, this description of Gremlin Server'
 out-of-the-box configuration.  It is quite possible to construct other flows, that might be more amenable to a
 particular language or style of processing.
 
+It is recommended but not required that a driver include a `User-Agent` header as part of any web socket
```
divijvaidya, line 851: What are the side effects of enabling it? We should help the users make this decision. Please add a jmh benchmark for handshake w/ and w/o the user agent. User agent will add additional latency to the handshake and hence, connection setup will be slower but by how much?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/channel/WebSocketChannelizer.java
```diff
@@ -108,7 +109,7 @@ public void configure(final ChannelPipeline pipeline) {
                 closeOnProtocolViolation(false).allowExtensions(true).maxFramePayloadLength(settings.maxContentLength).build();
         pipeline.addLast(PIPELINE_REQUEST_HANDLER, new WebSocketServerProtocolHandler(GREMLIN_ENDPOINT,
                 null, false, false, 10000L, wsDecoderConfig));
-
+        pipeline.addLast("ws-user-agent-handler", new WsUserAgentHandler());
```
divijvaidya, line 112: Please create a JIRA to implement this for HTTP calls as well.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/util/UserAgent.java
```diff
@@ -0,0 +1,55 @@
...
+    static {
+        String applicationName = "";
+        try {
+            applicationName = ((String)(new javax.naming.InitialContext().lookup("java:app/AppName"))).replace(' ', '_');
+        } catch (NamingException e) {
+            applicationName = "NotAvailable";
+        };
+
+        final String glvVersion = Gremlin.version().replace(' ', '_');
+        final String javaVersion = System.getProperty("java.version", "NotAvailable").replace(' ', '_');
```
divijvaidya, line 46: The information on how to configure the user agent is missing in documentation. We should add the information that the user agent is constructed from System Properties and could be overridden/set by using `-D`.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/channel/WebSocketChannelizer.java
divijvaidya: This handler should go "after" authN has taken place. Else an attacker may perform a DoS attack by sending unauthenticated large string in user agent.

divijvaidya (changes requested):
Thank you for the contribution! This is going to be very useful to the providers to understand the popular drivers with their users.

I have some concerns about possibility of DoS attack and the performance impact of this feature. Left a bunch of comments about it.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/util/UserAgent.java
```diff
@@ -0,0 +1,55 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.tinkerpop.gremlin.driver.util;
+
+import org.apache.tinkerpop.gremlin.util.Gremlin;
+import javax.naming.NamingException;
+
+public class UserAgent {
```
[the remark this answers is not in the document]
cole-bq: Thanks for all of your feedback Divij, could I ask for some more elaboration of your expectations regarding decoupling the driver and server code for the user agent? Keeping things decoupled was something I kept in mind while implementing my changes and I'm not sure what further changes you are looking for in that regard. As it stands all the code which generates and sends the user agent exists in gremlin-driver (which will be re-implemented in each GLV in a future PR). All the code which extracts the user agent, logs it for debugging, and adds it to the server metrics is wrapped into `WsUserAgentHandler` in gremlin-server. The only use of driver code from the server currently (excluding tests) is that the server uses the driver's constant for the "User-Agent" header name. This constant could easily be duplicated in the server to avoid referencing driver code from the server.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/WsUserAgentHandler.java
cole-bq: Thanks I will incorporate this change

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/WebSocketClientBehaviorIntegrateTest.java
cole-bq: That sounds like a good test to have, I will add it.

docs/src/dev/provider/index.asciidoc
cole-bq: How much value do you see in having such a benchmark? My understanding is that the jmh benchmarks aren't currently setup for measuring anything like the driver handshake latency. I also have concerns about how usefulness of any result as the latency will be highly dependent on network conditions. A typical user agent in the format we are using is under 100 bytes and it is only sent once per connection in the handshake. Personally I don't foresee there being any noticeable side effects other than the obvious added logging and metrics.

Some sort of benchmark can certainty be added but I don't believe it will be trivial to add and I worry it won't offer much value for the investment put into creating it.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/channel/WebSocketChannelizer.java
cole-bq: Unfortunately I'm not sure this reordering is possible without substantial change. The user agent handler triggers on a HandShake.Complete event right as the handshake is finishing. The authentication and authorization handlers take effect after the handshake is finished. Since the user agent is sent as an http header though, it is limited by MaxHeaderSize which defaults to 8192. This should be sufficient protection against maliciously large headers being sent from unauthenticated users. I will add a test which confirms this behavior.

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/WebSocketClientBehaviorIntegrateTest.java
cole-bq: On further inspection I'm less sure that any test of this nature would be useful. Since the user agent is sent as an http header during the handshake, it is by definition only sent once per connection. Once the handshake is done and the connection established, there isn't really anywhere to check for a user agent being sent again.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/util/UserAgent.java
cole-bq: While implementing this change I didn't have any intentions of making the user agent customizable. It is true that the user agent being constructed from system properties means that they can be overridden in maven with `-D` but this override only applies to that single execution. I'm not sure of any specific use cases for this behavior and such I'm not sure where it should be documented. I was wondering if you have anywhere in mind where it would be convenient, I don't want to dump extra information in places where it is not useful. Thanks.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/channel/WebSocketChannelizer.java
cole-bq: I have made this JIRA for the HTTP implementation. https://issues.apache.org/jira/browse/TINKERPOP-2830

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/util/UserAgent.java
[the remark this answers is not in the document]
divijvaidya: I was thinking in the following direction:

1. UserAgent.java is a model (POJO) that describes the UserAgent. It contains fields such as JavaVersion etc. along with the getter & setters. It contains methods "serializeToString" and "deserializeFromString". This model resides in a place which is common to all drivers, communication protocols and server. I would suggest placing it in https://github.com/apache/tinkerpop/tree/master/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver at same level as Host.java. This POJO/modelling for UserAgent would help us from code extensibility perspective in future if we want to add enums or constraints checks on values of user agent.

2. Both HTTP and WebSocket will create this POJO using a static method in the above file, perhaps, UserAgent#Generate() 


The goal of this structure is to ensure that code is extensible to HTTP in future and that individual methods could be unit tested easily. I don't have a strong opinion on this except for things such as WS_HANDSHAKE_USER_AGENT which should be made agnostic of WebSocket (the same could be used for HTTP).

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/util/UserAgent.java
divijvaidya: > I'm not sure of any specific use cases for this behavior
Let's remove it in that case. We can always add it later if someone needs it.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/channel/WebSocketChannelizer.java
divijvaidya: > The authentication and authorization handlers take effect after the handshake is finished.

AFAIK, AuthN is done as part of the handshake itself. But irrespective of that, I think our new handler only executes when a handshake is complete. In which case our new handler's code will not execute for non authenticated requests, hence, there is no risk (contrary to what I mentioned earlier). Is my understanding right here?

docs/src/dev/provider/index.asciidoc
divijvaidya: There are users of TinkerPop which open one connection per request and hence, they are sensitive to handshake latency. Theoretically the latency incurred will be due to increased size of header (hence longer header parsing time from bytebuffer to Java object + longer time to send larger object over network). 

If it is not trivial to calculate this, then we can ignore this since I believe there are larger bottlenecks in handshake than header parsing right now.

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/WebSocketClientBehaviorIntegrateTest.java
divijvaidya: sure, feel free to ignore the suggestion of this new test.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/util/UserAgent.java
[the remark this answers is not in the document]
cole-bq: Thanks for responding so soon. I agree that all of these changes would provide future extensibility benefits and I especially agree with the naming change to be web socket agnostic. I have pushed a small change where I have moved the UserAgent class to your suggested location and renamed WS_HANDSHAKE_USER_AGENT to just USER_AGENT. I have opted to keep USER_AGENT as a public statically initialized string constant as it is the most lightweight method of accessing the user agent and it meets the current needs. I certainly would not oppose modifying this further in the manner you've suggested sometime in the future if additional extensibility is required.

docs/src/dev/provider/index.asciidoc
cole-bq: I would like to ignore this for now. It seems that designing a good benchmark would take a reasonable amount of time and it is not clear to me how I would normalize results to account for various network conditions.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/channel/WebSocketChannelizer.java
cole-bq: Unfortunately in my tests, the user agent handler always triggers before the authentication handler regardless of the order the handlers are added to the pipeline. This is the order of events as I understand it:

The handshake is completely controlled by Netty's `WebSocketServerProtocolHandler`. When the handshake is completed, that handler fires off a `WebSocketServerProtocolHandler.HandshakeComplete` event which contains the request headers with the user agent. This event is the only reasonable place I am aware of that the user agent can be extracted. After this event has fired, the handshake is complete and the connection has been successfully upgraded to a web socket connection. The authentication and authorization handlers then do their work by exchanging messages with the client through the open web socket connection. This cannot be done until after the web socket has been established.

I don't see any good way to delay the user agent handler from triggering until after authentication. We will need to rely on the maxHeaderSize config limiting the size of any incoming header. I am also working on a change which will impose a hard limit on the number of unique user agents being added to the metrics to protect against a malicious user trying to fill memory with millions of garbage user agents.

cole-bq:
@divijvaidya Do you have any further feedback for this PR? Are there any other changes you would like to see included? Thanks.

divijvaidya (approved):
Thank you for making the changes. This change looks good to me.

Vote +1

PR #1843, opened 2022-10-24
TINKERPOP-2824 Properties on response elements

vkagamlyk:
Current status: 
- added property deserialization to all GLV's
- added `materializeProperties` request option to server, Go, .net, Python and Javascript clients
- added integration tests for Go, .net, Python and Javascript

Next:
- community review

Сan be discussed:
- format of properties in Javascript, Python and Go clients.

Jira ticket: https://issues.apache.org/jira/browse/TINKERPOP-2824


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/detached/DetachedVertex.java
```diff
@@ -74,6 +75,38 @@ protected DetachedVertex(final Vertex vertex, final boolean withProperties) {
...
+                || detachOptions.getDetachMode() == DetachStrategy.DetachMode.CUSTOM && detachOptions.getProperties() == null)
+            return;
+
+        final Iterator<VertexProperty<Object>> propertyIterator = vertex.properties();
+        if (!propertyIterator.hasNext())
+            return;
+
+        this.properties = new HashMap<>();
+
+        if (detachOptions.getDetachMode() == DetachStrategy.DetachMode.ALL) {
```
cole-bq, line 92: There is some code duplication between this case and DetachedVertex(vertex, true). Could potentially have the above constructor (DetachedVertex(final Vertex vertex, final boolean withProperties)) make calls to this constructor to avoid duplication.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/finalization/DetachStrategy.java
```diff
@@ -0,0 +1,156 @@
...
+                    .orElseGet(() -> traversal.getSteps().size());
+            traversal.addStep(index, new DetachElementStep<>(traversal, detachOptions));
+        }
+    }
+
+    public static DetachStrategy create(final Configuration configuration) {
+        return new DetachStrategy(DetachMode.valueOf(configuration.getString(ID_MODE)), configuration.getStringArray(ID_PROPERTIES));
+    }
+
+    public static final String ID_MODE = "mode";
```
kenhuuu, line 65: Static variables generally go near the top of the file.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/finalization/DetachStrategy.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        public String[] getProperties() {
+            return this.properties;
+        }
+
+        public static final class Builder {
+
+            private DetachMode detachMode = DetachMode.NONE;
+            private String[] properties;
```
kenhuuu, line 108: No default value may lead to a potential null pointer exception.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/detached/PropertyUtil.java
```diff
@@ -0,0 +1,35 @@
...
+import org.apache.tinkerpop.gremlin.structure.Property;
+
+/**
+ * Helper class that is used to calculate whether a property matches a template.
+ * Can be expanded to support wildcards or Regexp.
+ *
+ * @author Valentyn Kahamlyk
+ */
+public class PropertyUtil {
+    public static boolean Match(final DetachStrategy.DetachOptions detachOptions, final Property property) {
```
kenhuuu, line 32: This method should be named "match".

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/finalization/DetachStrategy.java
vkagamlyk: yes, but https://github.com/apache/tinkerpop/blob/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/computer/traversal/strategy/decoration/VertexProgramStrategy.java#L189

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/detached/DetachedVertex.java
vkagamlyk: fixed in https://github.com/apache/tinkerpop/pull/1843/commits/8290b2049fb4876fd7ddeafd18a2345222f85597

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/finalization/DetachStrategy.java
vkagamlyk: null is expected default. Is it better to explicitly assign null?

docs/src/reference/the-traversal.asciidoc
```diff
@@ -4916,6 +4916,28 @@ g.V().hasLabel('person'). <1>
...
 
+=== DetachStrategy
+
+`DetachStrategy` manages the fields that will be included in the result. The strategy adds a step to the end of the traversal 
+and does not affect the previous steps.
+There are 3 options for detachMode: `ALL` for all properties, `NONE` for no properties, `CUSTOM` for a specific list of properties.
+`id` and `label` always included in response.
+Same for Vertexes and Edges.
+
+WARNING: by default Gremlin Server configured to use `ReferenceElementStrategy` which removes all properties.
```
spmallette, line 4927: 1. What happens if `ReferenceElementStrategy` and `DetachStrategy` are both used at the same time?
2. I wonder what should happen to `ReferenceElementStrategy` in 3.7.x...is it even relevant anymore as there is `DetachStrategy` with mode `NONE` which is the same thing basically? I guess one returns a `ReferenceElement` and the other a `DetachedElement` with no properties so there is that difference. thoughts?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/finalization/DetachStrategy.java
```diff
@@ -0,0 +1,186 @@
...
+ */
+public final class DetachStrategy extends AbstractTraversalStrategy<TraversalStrategy.FinalizationStrategy> implements TraversalStrategy.FinalizationStrategy {
+
+    private static final DetachStrategy INSTANCE = new DetachStrategy(DetachMode.ALL, null);
+    private DetachMode detachMode = DetachMode.ALL;
+    private List<String> properties;
+
+    private DetachStrategy() {}
+
+    private DetachStrategy(final DetachMode detachMode, final List<String> properties) {
```
spmallette, line 54: would a `Set` be better than a `List` for the properties? i think it might be better to refer to these things as "keys" as "property" tends to denote a `Property` object typically in TinkerPop parlance.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/finalization/DetachStrategy.java
```diff
@@ -0,0 +1,186 @@
...
+    @Override
+    public Configuration getConfiguration() {
+        final Map<String, Object> map = new HashMap<>();
+        map.put(STRATEGY, DetachStrategy.class.getCanonicalName());
+        map.put(ID_MODE, this.detachMode.toString());
+        map.put(ID_PROPERTIES, this.properties);
+        return new MapConfiguration(map);
+    }
+
+    public enum DetachMode {
```
spmallette, line 98: it's too bad this enum addition forces all those serializer changes. is there other precedent for this in other strategies? in other words, do any other strategies introduce their own enums? maybe we can avoid this with some shift in the way we do the syntax?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/detached/DetachedFactory.java
```diff
@@ -115,4 +116,74 @@ public static <D> D detach(final Object object, final boolean withProperties) {
             return (D) object;
         }
     }
+
+    // detached with options
+    public static DetachedVertex detach(final Vertex vertex, final DetachStrategy.DetachOptions detachOptions) {
```
spmallette, line 121: i think we'd want to deprecate the `withProperties` overloads for these methods and prefer this new one. thoughts?

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/finalization/DetachStrategyProcessTest.java
```diff
@@ -0,0 +1,65 @@
...
+public class DetachStrategyProcessTest extends AbstractGremlinProcessTest {
+
+    @Test
+    @LoadGraphWith(MODERN)
+    @IgnoreEngine(TraversalEngine.Type.COMPUTER)
+    public void shouldReturnOnlyRequestedFields() {
+
+        final Vertex v = g.V(1).next();
+        final Vertex vWithAllProps = g
+                .withStrategies(DetachStrategy.build().detachMode(DetachStrategy.DetachMode.ALL).properties("name").create())
```
spmallette, line 45: did you think about not allowing invalid configurations in the builder?

docs/src/reference/the-traversal.asciidoc
vkagamlyk: 1. The result will be Vertex without properties. I don't think it makes sense to use them at the same time.
2. I think `ReferenceElementStrategy` may be deprecated in 3.7.
There are some small differences between ReferenceElement and DetachedElement that need to be aligned before. For example https://github.com/apache/tinkerpop/blob/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/reference/ReferenceElement.java#L50 and https://github.com/apache/tinkerpop/blob/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/detached/DetachedElement.java#L49
3. I consider default may be `DetachStrategy` configured to skip properties. I think users who want to get properties can remove the strategy from server startup or use it from the client when needed. Most current users don't expect to get properties and it doesn't make sense to load the server and network.
Another option is to remove `ReferenceElementStrategy` and allow users to choose the settings themselves, with the recommendation to skip properties by default.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/finalization/DetachStrategy.java
vkagamlyk: There are no other strategies that would add enums. May I just use a string value?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/detached/DetachedFactory.java
vkagamlyk: This overloads used in some other cases like Gryo Serializers, `DedupGlobalStep`, `EventStrategy`, etc.
So it seems to me that it is more convenient to keep `withProperties` as shortcut

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/finalization/DetachStrategyProcessTest.java
vkagamlyk: throw exception when non-empty `properties` and detachment mode is not CUSTOM?

docs/src/reference/the-traversal.asciidoc
spmallette: > I think ReferenceElementStrategy may be deprecated in 3.7.
There are some small differences between ReferenceElement and DetachedElement that need to be aligned before.

I'm not sure that `ReferenceElement` can be replaced by `DetachedElement` for all use cases. The original point of `ReferenceElement` is to avoid the the memory costs of even having an empty `Map<String, List<Property>> properties` declaration at all for OLAP use cases. `ReferenceElement` is meant to be as light as possible. If it could be proven that `DetachedElement` could have that same memory footprint I suppose the entire `ReferenceElement` infrastructure could be deprecated. 

As far as `ReferenceElementStrategy` goes.......I think that can be deprecated as part of this because it is not use in OLAP iirc. It is only used for the case `DetachStrategy` is and is behaviorally identical for practical purposes.

> I consider default may be DetachStrategy configured to skip properties. I think users who want to get properties can remove the strategy from server startup or use it from the client when needed. Most current users don't expect to get properties and it doesn't make sense to load the server and network.

i'd agree. so the only real change users will get is that they will now get a `DetachedElement` rather than a `ReferenceElement` and hopefully they haven't been coding to the implementation and will not notice the difference as they would be working with the interface of `Element`.  The upgrade documentation will need to be sure to include this notice.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/finalization/DetachStrategy.java
spmallette: we do have a precedent for using integer constants for `with()`-step: https://github.com/apache/tinkerpop/blob/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/WithOptions.java 

maybe we should follow that pattern here and save the serialization expansion. maybe the rule is to save enum for core language elements, like `Scope`, `Direction`, etc.

docs/src/reference/the-traversal.asciidoc
vkagamlyk: I can change detach mode `NONE` or add detach mode `REFERENCE` to build `ReferenceElement`.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/finalization/DetachStrategy.java
vkagamlyk: done in https://github.com/apache/tinkerpop/pull/1843/commits/7d82c1b134de7b32e0f4f7f37262fbbcf1124654

docs/src/reference/the-traversal.asciidoc
spmallette: that's interesting. We often refer to getting a `ReferenceElement` as "reference detachment" so that might be a neat way to centralize all detachment functions behind a single strategy. seems like a good idea imo

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
```diff
@@ -34,15 +34,16 @@ public class GremlinScriptChecker {
...
      * An empty result whose properties return as empty.
      */
-    public static final Result EMPTY_RESULT = new Result(null, null);
+    public static final Result EMPTY_RESULT = new Result(null, null, null);
 
     /**
      * At least one of these tokens should be present somewhere in the Gremlin string for {@link #parse(String)} to
      * take any action at all.
      */
+    // todo: why not used `Tokens`?
```
spmallette, line 43: not sure what this todo was for, but you should include the `Tokens` version here too in the caps form.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
```diff
@@ -94,6 +95,11 @@ public class GremlinScriptChecker {
      */
     private static final String requestIdTokens = "[\"']requestId[\"']|(?:Tokens\\.)?REQUEST_ID";
 
+    /**
+     * Regex fragment for the materializeProperties to look for.
+     */
+    private static final String materializePropertiesTokens = "[\"']materializeProperties[\"']";
```
spmallette, line 101: and here include the `Tokens.` form

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
```diff
@@ -162,7 +168,10 @@ public class GremlinScriptChecker {
      * </pre>
      */
     private static final Pattern patternWithOptions =
-            Pattern.compile("\\.with\\(((?:" + timeoutTokens + "),(?<to>\\d*)(:?L|l)?)|((?:" + requestIdTokens + "),[\"'](?<rid>.*?))[\"']\\)");
+            Pattern.compile("\\.with\\((((?:"
```
spmallette, line 171: every time i touch this, i've double checked to make sure there is no significant performance penalty. this checker is executed a lot so we'd want to know if a slow down was happening here. could you just validate that there is no significant change with a quick benchmark? i don't expect anything given what i recall from the last update here, but i think it's worth continuing to validate assumptions on this one. maybe even make this a permanent benchmark in gremlin-benchmarks so that we dont have to start from scratch each time?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
```diff
@@ -228,11 +244,20 @@ public Optional<String> getRequestId() {
             return null == requestId ? Optional.empty() : Optional.of(requestId);
         }
 
+        /**
+         * Gets the value of the request identifier supplied using the {@code with()} source step. If there are
```
spmallette, line 248: nit: prefer `link` for javadoc when referring elsewhere in the codebase that we can hyperlink to and `code` when referring to the item being documented, external libs or jdk classes especially on public methods like this one.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/types/EdgeSerializer.java
```diff
@@ -71,8 +73,12 @@ protected void writeValue(final Edge value, final Buffer buffer, final GraphBina
...
         // here as an option for the future as we've waffled this soooooooooo many times now
         context.write(null, buffer);
-        // we don't serialize properties for graph vertices/edges. they are "references", but we leave a place holder
-        // here as an option for the future as we've waffled this soooooooooo many times now
-        context.write(null, buffer);
+        if (value.properties() == null) {
+            context.write(null, buffer);
+        }
+        else {
+            List<?> asList = IteratorUtils.toList(value.properties());
```
spmallette, line 80: nit: please do a check for variables that can be marked `final`

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptCheckerTest.java
```diff
@@ -192,12 +194,41 @@ public void shouldIdentifyMultipleRequestIds() {
                 getRequestId().get());
     }
 
+    @Test
```
spmallette, line 197: please include test variations for `Tokens` styled naming when that's added given the previous comments.

gremlin-javascript/src/main/javascript/gremlin-javascript/test/cucumber/feature-steps.js
```diff
@@ -131,17 +131,30 @@ Given(/^using the parameter (.+) defined as "(.+)"$/, function (paramName, strin
...
+  } else if (p instanceof graphModule.Path) {
+    removeProperties(p.objects)
+  }
+
+  return p
+}
+
 When('iterated to list', function () {
-  return this.traversal.toList().then(list => this.result = list).catch(err => this.result = err);
+  return this.traversal.toList().then(list => this.result = removeProperties(list)).catch(err => this.result = err);
```
spmallette, line 151: could you please explain what's happening here with `removeProperties()` and related changes elsewhere referring to "properties excluded from verification"? i'm not sure i follow

vkagamlyk: Current state:
1. different GLV's use different ways to compare elements, for example Java and .NET consider Elements equal when ID's equal. But in Javascript used memberwise comparison, so Elements equal only when all properties equal.
2. Javascript client already contains property deserialization, and format is different for GraphSON and GraphBinary.
For GraphSON properties is object, for GraphBinary used plain array.

As a result Vertex loaded with GraphSON is not equal to same Vertex got with GraphBinary due to different properties format, so I excluded properties from elements for Javascript Gherkin tests to pass existing tests.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
vkagamlyk: There's a bit of a mess with names for this constants. 
For example we have constant for [evaluationTimeout](https://github.com/apache/tinkerpop/blob/master/gremlin-util/src/main/java/org/apache/tinkerpop/gremlin/util/Tokens.java#L78) and can use `Token.ARGS_EVAL_TIMEOUT`, not plain string. Same for `requestId`, but not for `scriptEvaluationTimeout`. Also `Tokens.ARGS_SCRIPT_EVAL_TIMEOUT` is not defined and I don't understand why.

I think better to use constants when possible and this like can looks like
```
private static final Set<String> tokens = new HashSet<>(Arrays.asList(Tokens.ARGS_EVAL_TIMEOUT, Tokens.ARGS_SCRIPT_EVAL_TIMEOUT,
                                                 Tokens.REQUEST_ID, Tokens.ARGS_MATERIALIZE_PROPERTIES));
```
but made in the same style as before.

FlorianHockmann (commented):
Great change! I just added a few comments but basically all of them are only nitpicks.

Apart from that, I don't see any documentation for this new functionality. I think this needs to be documented in a lot of places (upgrade docs, changelog, reference docs, I/O docs).
We should also definitely document that this changes the default behavior which will now include properties for all returned vertices and edges.

Do you want to do that in a follow-up PR?

gremlin-dotnet/src/Gremlin.Net/Structure/Edge.cs
```diff
@@ -52,6 +55,15 @@ public Edge(object? id, Vertex outV, string label, Vertex inV)
         /// </summary>
         public Vertex OutV { get; }
 
+        /// <summary>
+        /// Get property by key
+        /// </summary>
+        /// <returns>property or null when not found</returns>
+        public Property? Property(string key)
```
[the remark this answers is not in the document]
vkagamlyk: I considered 3 options:
1. minimal change, just extend `Vertex` and `Edge` classes with new typed `Property` method like `public Property? Property(string key)`
2. non-breaking change, add dynamic `Property` method to the base `Element` class like `public dynamic? Property(string key)`, it's less convenient for users
3. breaking change, make `Element' type generic `Element<PropertyType>` and method like `public T? Property(string key)`

first one is safest and at the same time user-friendly, so I chose it

gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphBinary/Types/EdgeSerializer.cs
```diff
@@ -45,15 +46,17 @@ public EdgeSerializer() : base(DataType.Edge)
         {
             await writer.WriteAsync(value.Id, stream, cancellationToken).ConfigureAwait(false);
             await writer.WriteNonNullableValueAsync(value.Label, stream, cancellationToken).ConfigureAwait(false);
-            
+
+            // !!! await writer.WriteValueAsync(value.Label, stream, false, cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
vkagamlyk: I forgot to remove this like

gremlin-server/src/test/scripts/generate-all.groovy
```diff
@@ -61,7 +61,7 @@ globals << [hook : [
 
 // add default TraversalSource instances for each graph instance
 globals << [gclassic : traversal().withEmbedded(classic).withStrategies(ReferenceElementStrategy)]
-globals << [gmodern : traversal().withEmbedded(modern).withStrategies(ReferenceElementStrategy)]
+globals << [gmodern : traversal().withEmbedded(modern)]
```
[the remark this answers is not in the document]
vkagamlyk: Yes, I used `gmodern` for tests

vkagamlyk:
> Great change! I just added a few comments but basically all of them are only nitpicks.
> 
> Apart from that, I don't see any documentation for this new functionality. I think this needs to be documented in a lot of places (upgrade docs, changelog, reference docs, I/O docs). We should also definitely document that this changes the default behavior which will now include properties for all returned vertices and edges.
> 
> Do you want to do that in a follow-up PR?

I plan to add documentation update to this PR, I first wanted the PR to be more or less formed

gremlin-dotnet/src/Gremlin.Net/Structure/Edge.cs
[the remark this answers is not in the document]
FlorianHockmann: Sorry, I didn't look closely enough and completely forgot that a `Vertex` has a `VertexProperty` which isn't the same as the `Property` of an `Edge`. Then I agree with your approach here.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
spmallette: We have to check all the forms. if you just did:

```
private static final Set<String> tokens = new HashSet<>(Arrays.asList(Tokens.ARGS_EVAL_TIMEOUT, Tokens.ARGS_SCRIPT_EVAL_TIMEOUT,
                                                 Tokens.REQUEST_ID, Tokens.ARGS_MATERIALIZE_PROPERTIES));
```

then if the user sent a string query of: `g.with(Tokens.ARGS_EVAL_TIMEOUT, 100)` the checker wouldn't find it. `Tokens.ARGS_EVAL_TIMEOUT` is basically a valid part of the Gremlin string someone might send.

spmallette: I suppose you could do this:

```java
private static final Set<String> tokens = new HashSet<>(Arrays.asList(Tokens.ARGS_EVAL_TIMEOUT, Tokens.ARGS_SCRIPT_EVAL_TIMEOUT,
"ARGS_EVAL_TIMEOUT", "ARGS_SCRIPT_EVAL_TIMEOUT",
                                                 Tokens.REQUEST_ID, Tokens.ARGS_MATERIALIZE_PROPERTIES
                                                 "REQUEST_ID", "ARGS_MATERIALIZE_PROPERTIES"));
```

But then the other problem is that the script checker is in `gremlin-core` and `Tokens` is in `gremlin-driver`. It's a little messed up. I suppose the right way to resolve it would be to officially make these settings part of `gremlin-language` but that seems out of scope for your work. You could create something in JIRA for that if you want to come back to it.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
spmallette: please update the "explain" in the javadoc: https://regex101.com/

gremlin-javascript/src/main/javascript/gremlin-javascript/test/cucumber/feature-steps.js
spmallette: ugh - that's really bad. i agree with florian that needs to be fixed. if no other GLVs need this special handling then i'd say create a JIRA to track explicitly since this is a behavioral change and just include the fix here in this PR so we don't need this `removeProperties()` workaround.

gremlin-go/driver/graphBinary.go
```diff
@@ -1052,8 +1052,8 @@ func vertexReader(data *[]byte, i *int) (interface{}, error) {
 	return vertexReaderNullByte(data, i, true)
 }
 
-// {fully qualified id}{unqualified label}{[unused null byte]}
-func vertexReaderNullByte(data *[]byte, i *int, unusedByte bool) (interface{}, error) {
+// {fully qualified id}{unqualified label}{fully qualified properties}
+func vertexReaderNullByte(data *[]byte, i *int, readProperties bool) (interface{}, error) {
```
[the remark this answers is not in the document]
vkagamlyk: sounds good, renamed in https://github.com/apache/tinkerpop/pull/1843/commits/31abd69ac156047a4de748011ea768d8bf9cdae9

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
vkagamlyk: "explain" updated in https://github.com/apache/tinkerpop/pull/1843/commits/b8582c750db1982a33ba41f4a9fda976ecc31eaa

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
vkagamlyk: and 2 years ago you preferred to use `code` :) and I did the same

fixed in https://github.com/apache/tinkerpop/pull/1843/commits/143a2bf5b39a26412e501bc3a3f6f564d32d9afd

spmallette: ha - i code reviewed myself i guess

gremlin-javascript/src/main/javascript/gremlin-javascript/test/cucumber/feature-steps.js
vkagamlyk: I created ticket: https://issues.apache.org/jira/browse/TINKERPOP-2872

Perhaps it is also worth discussing to get the same properties format for GraphSON and GraphBinary.

spmallette: as long as javascript matches the defined format for GraphSON i don't think we need to change anything: https://tinkerpop.apache.org/docs/current/dev/io/#_vertex_3

vkagamlyk: This issue is related to the internal representation of the data in JS client.

When data comes in GraphSON, when properties is stored as Object where key is property name and value are array of properties, like described in GraphSON specification. To get property used code `age = vertex.properties.age[0]`.

For GraphBinary used simple array `property[]`. To get same property 'age = vertex.properties[1]'

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
vkagamlyk: Benchmark                                                      Mode  Cnt       Score       Error  Units
GremlinScriptCheckerBenchmark.testParseAll                    thrpt   20  203909.105 ±  8915.473  ops/s
GremlinScriptCheckerBenchmark.testParseMaterializeProperties  thrpt   20  484647.497 ± 15267.578  ops/s
GremlinScriptCheckerBenchmark.testParseRequestId              thrpt   20  382702.687 ±  6140.784  ops/s

vkagamlyk: benchmark added in https://github.com/apache/tinkerpop/pull/1843/commits/8c36b01b673982875d9d9e04d91f3ddce73860f5

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
vkagamlyk: added in https://github.com/apache/tinkerpop/pull/1843/commits/8c36b01b673982875d9d9e04d91f3ddce73860f5

gremlin-python/src/main/python/tests/structure/io/test_functionalityio.py
```diff
@@ -25,6 +25,34 @@
...
+
+def test_vertex_without_properties(remote_connection):
+    g = Graph().traversal().withRemote(remote_connection)
+    vertex = g.with_('materializeProperties', 'tokens').V(1).next()
+    assert vertex.id == 1
+    # empty array for GraphBinary and missing field for GraphSON
+    assert vertex.properties is None or len(vertex.properties) == 0
+
+
+def test_edge(remote_connection):
```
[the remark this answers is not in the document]
vkagamlyk: added in https://github.com/apache/tinkerpop/pull/1843/commits/9e9520b1c4aefa027445a351af19900cdd0bdeb9

docs/src/reference/gremlin-applications.asciidoc
```diff
@@ -2250,50 +2250,36 @@ to as "detached elements" and cases where properties are not included are "refer
...
 [source,groovy]
 ----
-globals << [g : traversal().withEmbedded(graph).withStrategies(ReferenceElementStrategy)]
+g.with('materializeProperties', 'tokens').V()
 ----
 
-This configuration is global to Gremlin Server and therefore all methods of connection will always return elements
-without properties. If this strategy is not included, then there are other considerations to take into account such as
-the connection type (i.e. script or bytecode) and the serializer.
+The `tokens` value for the `materializeProperties` means that need to return only `id` and `label`. Another option
```
FlorianHockmann, line 2263:
```suggestion
The `tokens` value for the `materializeProperties` means that only `id` and `label` should be returned. Another option
```

docs/src/reference/gremlin-applications.asciidoc
```diff
@@ -2250,50 +2250,36 @@ to as "detached elements" and cases where properties are not included are "refer
...
+`all` used to indicate that all properties should be returned and is the default value.
```
FlorianHockmann, line 2264: I think this needs to be:

> Another option, `all`, _can be_ used [...]

FlorianHockmann (commented):
Just reviewed the added docs. Another place of the docs that needs to be updated are the IO docs, e.g., [here](https://tinkerpop.apache.org/docs/current/dev/io/#_vertex_4).

docs/src/upgrade/release-3.7.x.asciidoc
```diff
@@ -29,6 +29,28 @@ Please see the link:https://github.com/apache/tinkerpop/blob/3.7.0/CHANGELOG.asc
 
 === Upgrading for Users
 
+==== Properties on Elements
```
spmallette, line 32: This is a major piece of functionality. I think the upgrade documentation needs more content to really impress upon folks just what impact this change has for them. some background on the previous behavior would help, a link to the "history" of why it was that way would be good, some example code, etc. 

I can help with polishing up wording, but please add a bit more here for us to review.

spmallette:
i think you need multiple CHANGELOG entries on this one. you added one feature but a lot of things changed like, major changes to different serializers,  changes to where `HaltedTraverserStrategy` was used, `GremlinScriptChecker` changes, etc.

docs/src/upgrade/release-3.7.x.asciidoc
spmallette, line 32: i can't recall, but are there any compatibility issues to call attention to here? like can a 3.6.x driver work nicely with 3.7.x or the reverse? maybe there is something to call attention to here with that.

vkagamlyk: different behavior for different GLV's
Python, Go: deserialization error for GraphBinary, skip properties for GraphSON
.NET:  skip properties
Javascript: should work ok

also serialization can't be managed with `materializeProperties` option

docs/src/upgrade/release-3.7.x.asciidoc
```diff
@@ -29,9 +29,73 @@ Please see the link:https://github.com/apache/tinkerpop/blob/3.7.0/CHANGELOG.asc
...
+g.With(Tokens.ArgMaterializeProperties, "tokens").V(1).Next()
+----
+
+See: link:https://issues.apache.org/jira/browse/TINKERPOP-2824[TINKERPOP-2824]
+
+
 ==== Gremlin.NET: Nullable Annotations
 
-Gremlin.NET now uses link:https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references#nullable-variable-annotations[nullable annotations]
+Gremlin.NET now uses link:https://learn.microsoft.com/en-us/dotnet/sharp/nullable-references#nullable-variable-annotations[nullable annotations]
```
Cole-Greer, line 98: I believe this edit was added by mistake. New link results in a 404

CHANGELOG.asciidoc
```diff
@@ -38,6 +38,9 @@ This release also includes changes from <<release-3-6-XXX, 3.6.XXX>>.
 * Reduces dependency from `gremlin-server` onto `gremlin-driver` to a test scope only.
 * Added `RequestOptions` and `RequestOptionsBuilder` types to Go GLV to encapsulate per-request settings and bindings.
 * Added `SubmitWithOptions()` methods to `Client` and `DriverRemoteConnection` in Go GLV to pass `RequestOptions` to the server.
+* Changed default behavior for returning properties on Elements for OLTP queries. Properties now returned.
+* Detachment is no longer performed in `TraverserIterator`.
+* Added `materializeProperties` request option to control properties serialization.
```
spmallette, line 43: I'd include: 

* Modified serializers in <languages> to handle serialization and deserialization of properties.
* Added functional properties to the graph structure components for <languages or just "non-JVM languages" if you did them all but Java>.
* Modified the `GremlinScriptChecker` to extract the `materializeProperties` request option.

spmallette:
it seems like existing tests are running and passing, but none of the existing tests validate properties coming back. so, what tests ensure that `materializeProperties` is behaving properly?

gremlin-server/src/test/scripts/generate-all.groovy
```diff
@@ -61,7 +61,7 @@ globals << [hook : [
 
 // add default TraversalSource instances for each graph instance
 globals << [gclassic : traversal().withEmbedded(classic).withStrategies(ReferenceElementStrategy)]
```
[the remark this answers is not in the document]
vkagamlyk: to reduce the number of problems with tests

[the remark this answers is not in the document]
spmallette: sorry - what am i forgetting? what tests rely on `ReferenceElementStrategy` to be present? 

ideally, `ReferenceElementStrategy` should be a thing of the past, no?

docs/src/upgrade/release-3.7.x.asciidoc
spmallette: your recent work shows `ReferenceElement`-type objects are no longer returned - you get a `DetachedElement` from remote requests. Folks may not have been coding to the `Element` interfaces as they likely should have been and therefore could been working directly with "reference" implementations. I think upgrade docs should point out the change and the appropriate way to avoid problems in the future - i.e. use the interfaces like `Vertex` and `Edge`.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/types/VertexPropertySerializer.java
```diff
@@ -66,7 +66,10 @@ protected void writeValue(final VertexProperty value, final Buffer buffer, final
         // we don't serialize the parent vertex, let's hold a place for it
         context.write(null, buffer);
 
-        final List<?> asList = IteratorUtils.toList(value.properties());
-        context.write(asList, buffer);
+        // Neo4jVertexProperty throw UnsupportedOperationException for properties,
+        // so for now VertexProperty will be serialized without inner properties
+        // final List<?> asList = IteratorUtils.toList(value.properties());
+        // context.write(asList, buffer);
+        context.write(null, buffer);
```
spmallette, line 73: I assume that this is temporary? not all graphs support meta/multiproperties so this is a likely to be a general problem. if you had the `Graph` or `g` you could test `features()` to see what the graph supported and then either write the metaproperties or not, but i don't think you have access to that here. 

Any idea why this works for GraphSON without omitting the metaproperties?

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerSerializationIntegrateTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Test
+    public void shouldDeserializeVertexPropertiesForGremlin() {
+        final Vertex vertex = client.submit("gmodern.V(1)").one().getVertex();
+
+        assertVertexWithProperties(vertex);
+    }
+
+    @Test
+    public void shouldSkipVertexPropertiesForGremlin() {
```
spmallette, line 87: nit: "ForGremlin" - prefer "ForScripts"

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerSerializationIntegrateTest.java
```diff
@@ -0,0 +1,149 @@
...
+import org.junit.runners.Parameterized;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(Parameterized.class)
+public class GremlinServerSerializationIntegrateTest extends AbstractGremlinServerIntegrationTest {
```
spmallette, line 46: given the issues with metaproperties/neo4j that i noted elsewhere, I think this test needs to include assertions for metaproperties and multiproperties to ensure all of that is working.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerSerializationIntegrateTest.java
```diff
@@ -0,0 +1,149 @@
...
+        final Vertex vertex = client.submit("gmodern.V(1)").one().getVertex();
+
+        assertVertexWithProperties(vertex);
+    }
+
+    @Test
+    public void shouldSkipVertexPropertiesForGremlin() {
+        final Vertex vertex = client.submit("gmodern.with('materializeProperties', 'tokens').V(1)").one().getVertex();
+
+        assertEquals(Collections.emptyIterator(), vertex.properties());
```
spmallette, line 90: at risk of redundancy, but for sanity, please assert that the `T` values are present in addition to asserting the properties are not.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/types/VertexPropertySerializer.java
vkagamlyk: GraphSON uses workaround by getting properties starting from `keys`. It only works when `Neo4jVertexProperty.vertexPropertyNode` is null or empty.

spmallette: what is the plan to re-enable this?

vkagamlyk: I can do it like in GraphSON, but then there will be a risk of getting a problem if some provider wants to use `vertexPropertyNode`.
Do you have other suggestions?

spmallette: I'm not sure i understand the nature of the risk you're describing, but the `keys()` implementation for `Neo4jGraph` is strange anyway. It returns an empty `Set`  while `properties()` throws an exception...that's pretty inconsistent.  I"m not even sure `vertexPropertyNode` is relevant - i have a feeling it is dead code and `keys()` was missed in some refactoring at some point.

Neo4j really isn't the issue though. The real issue is how it works for any `Graph` implementation that doesn't support metaproperties. As the current approach has worked for years in GraphSON/Gryo I feel like we can fairly assume it should be safe for GraphBinary. 

That said, perhaps we should do better than assume. You could encode the idea that `properties()` / `keys()` should not throw exceptions but instead return empty iterables.  You'd do that by modifying test semantics for `FeatureSupportTest` and probably `VertexPropertyTest` (there might be others that fail as a result of this change). This change would require Upgrade Docs for providers to take note of. I think you'd want to wholly remove `VertexProperty.Exception.metaPropertiesNotSupported` basically and see what breaks.

vkagamlyk: `VertexProperty.Exception.metaPropertiesNotSupported` used when `VertexProperty` doesn't support getting and setting meta-properties. I changed getter behavior to return empty Iterable, but kept Exception for setter.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerSerializationIntegrateTest.java
vkagamlyk: added in https://github.com/apache/tinkerpop/pull/1843/commits/6884272f19b8ed97a9de3f53033e651b3fa437bf

spmallette:
could you please rebase to get rid of the conflicts and squash to clean up the commit history (you don't need to do this as one commit i guess, but 80+ tiny commits is a bit much 🙂)? other than that, I think this is generally ready to merge. let's see how it looks after rebase for a final review.

vkagamlyk:
Replaced with squashed https://github.com/apache/tinkerpop/pull/1992

PR #1845, opened 2022-10-25
Better error for addE() where the FROM or TO traverser gives no results

metawilm-aws:


metawilm-aws:
Sorry, something went wrong with this PR. Please see https://github.com/apache/tinkerpop/pull/1860 instead

PR #1846, opened 2022-10-25
Better error for group().by() if the by() traverser gives no results

metawilm-aws:


spmallette:
Given relatively recent changes in `by()` behavior over 3.5.x and 3.6.x, this error situation isn't likely to happen - https://tinkerpop.apache.org/docs/current/upgrade/#_consistent_by_behavior unless there is a specific test case that demonstrates how to arrive in this state, it's perhaps best to leave this as it is for now.

PR #1848, opened 2022-10-26
EXAMPLE: Deserialization of integer based datatypes lose the type definition

epankala:


epankala:
This PR as is is incomplete fix and just an example on how it works for me and how I suspect issue could be resolved.

This PR also does not handle other 'int' based types only long and bigint. 

It is possible that there are also other than 'int' based unhandled cases in the deserialization.

Also at least float seemed to be transmitted as 0x07 (Double at https://github.com/apache/tinkerpop/blob/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/DataType.java) back from the server so additionally maybe something is off with python serialization of 'float' type.

xiazcy:
Hi @epankala, thank you for raising this issue. We have created a JIRA ticket to track the process: https://issues.apache.org/jira/browse/TINKERPOP-2848. 

Would you still be interested in investigating this further?

xiazcy:
Closing due to inactivity. Please feel free to follow-up in the JIRA ticket.

PR #1862, opened 2022-11-09
TINKERPOP-2826 bump ivy to 2.5.1

acoady:


acoady:
We need an update to spark-core which also uses IVY 2.5.0 which is not available yet

PR #1867, opened 2022-11-16
TINKERPOP-2348 Add nullable annotations

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2348

This only adds nullable annotations to the projects Gremlin.Net and the Gremlin.Net.Template, but not to the test projects. I intend to add it there in a follow-up PR to at least somehow limit the scope of this commit. This takes already care of everything user-facing.

This unfortunately includes a few breaking changes (not necessarily complete):
- Message types:
  * Some set property accessors removed
- Structure types:
  * Some set property accessors removed
- `DefaultTraversal.Bytecode` now abstract
- `GraphBinaryReader`/`Writer`
  * `ReadValueAsync` / `WriteValueAsync` were split into two methods, for nullable and non-nullable values
- Changed exceptions: Instead of `NullReferenceExceptions`, `IOExceptions` should be thrown during serialization

Most of these breaking changes are also mentioned in the upgrade docs. I haven't added all of them there as I found it to be unlikely that anyone is really affected by them.

VOTE +1


gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs
```diff
@@ -96,15 +99,15 @@ private GraphTraversal(ICollection<ITraversalStrategy> traversalStrategies, Byte
...
         {
             if (edgeIdsOrElements == null)
             {
-                Bytecode.AddStep("E", new object[] { null });
+                Bytecode.AddStep("E", new object?[] { null });
             }
             else
             {
-                var args = new List<object>(edgeIdsOrElements.Length);
+                var args = new List<object?>(edgeIdsOrElements.Length);
```
vkagamlyk, line 110: why is the `edgeIdsOrElements` array copied through the list?

FlorianHockmann: This file was originally generated automatically by some Groovy script that created basically one C# implementation of every Java implementation of a Gremlin step. It simply created a new list for all arguments so it could add all arguments to that list and then pass the list to `Bytecode.AddStep()`.
Now that we no longer generate this code automatically, we could of course improve / optimize it by hand and remove these unnecessary creation of lists for example. However I think that should be done in a separate PR.

vkagamlyk:
VOTE +1

jorgebay (approved):
Looks good to me.

VOTE +1

PR #1873, opened 2022-11-25
[TINKERPOP-2831] Use FastNoSuchElementException to improve performance

ministat:
Janusgraph's JanusGraphVertexDeserializer uses the following code to create a TinkerVerex, which leads to a lot of NuSuchElementException and hurts the performance.

FastNoSuchElementException can fix the performance issue.

    public TinkerVertex getOrCreateVertex(final long vertexId, final String label, final TinkerGraph tg) {
        TinkerVertex v;

        try {
            v = (TinkerVertex)tg.vertices(vertexId).next();
        } catch (NoSuchElementException e) {
            if (null != label) {
                v = (TinkerVertex) tg.addVertex(T.label, label, T.id, vertexId);
            } else {
                v = (TinkerVertex) tg.addVertex(T.id, vertexId);
            }
        }

        return v;
    }

### DIFF
--- a/tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerGraphIterator.java
+++ b/tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerGraphIterator.java
@@ -18,6 +18,7 @@
  */
 package org.apache.tinkerpop.gremlin.tinkergraph.structure;
 
+import org.apache.tinkerpop.gremlin.process.traversal.util.FastNoSuchElementException;
 import org.apache.tinkerpop.gremlin.structure.util.CloseableIterator;
 import org.apache.tinkerpop.gremlin.util.iterator.StoreIteratorCounter;
 
@@ -60,7 +61,7 @@ public class TinkerGraphIterator<E> implements CloseableIterator<E> {
     @Override
     public E next() {
         if (!hasNext()) {
-            throw new NoSuchElementException();
+            throw FastNoSuchElementException.instance();
         }
 
         final E ret = next;
--- a/tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerGraphIteratorTest.java
+++ b/tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerGraphIteratorTest.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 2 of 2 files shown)


ministat:
@cole-bq I'm interested with your exploration. Using Java's Exception to stop the iteration is not recommended since it has to fill stack trace which hurts the performance. The workaround now is to use Exception without stack trace.

I checked all places which catches the NoSuchElementException. Currently, there are some other places which catch the NoSuchElementException to stop the traversal, and none of them care about the stacktrace, just wants to break the loop or traversal, like this patch wants to address.

Now, there are some other graph DB which already refers to Tinkerpop and depends on the exception to do something in their logic, like Janusgraph mentioned in this issue. So, if you prefer to some other solution, please consider the backward compatibility.  

```
src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/TraversalUtil.java:45:        } catch (final NoSuchElementException e) {
src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/TraversalUtil.java:113:        } catch (final NoSuchElementException e) {
src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/AbstractStep.java:156:            } catch (final NoSuchElementException e) {
src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Traversal.java:187:        } catch (final NoSuchElementException ignored) {
src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Traversal.java:212:        } catch (final NoSuchElementException ignored) {
src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Traversal.java:267:        } catch (final NoSuchElementException ignore) {
src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Traversal.java:280:        } catch (final NoSuchElementException ignore) {
```

cole-bq:
@ministat I've done a deeper exploration into this issue and I don't believe that any modification to tinkerpop is necessary. Based on my initial glance at the code I was under the impression that `TinkergraphIterator.hasNext()` would be a slow operation due to its use of exceptions to determine if there were any elements remaining. In my testing, the TinkergraphIterator returned by `tg.vertices(vertexId)` always uses FastNoSuchElementException internally during calls to `hasNext()`.

Based on my findings I would suggest modifying your example to:
```
public TinkerVertex getOrCreateVertex(final long vertexId, final String label, final TinkerGraph tg) {
    TinkerVertex v;
    Iterator<Vertex> iter = tg.vertices(vertexId);
    if(iter.hasNext()) {
        try {
            v = (TinkerVertex) iter.next();
        } catch (NoSuchElementException e) {}
    } else {
        if (null != label) {
                v = (TinkerVertex) tg.addVertex(T.label, label, T.id, vertexId);
            } else {
                v = (TinkerVertex) tg.addVertex(T.id, vertexId);
            }
    }
    return v;
}
```
Based on my testing this will give the same or slightly better performance than switching `next()` to use `FastNoSuchElementException` and not guarding calls to next() with hasNext().

I would recommend that we continue using NoSuchElementException as is. Even if the stack trace for it isn't currently being used, it seems to me that it may be useful at some point in the future. If `hasNext()` can be used to avoid any performance issues I don't see a benefit in removing the stack trace.

ministat:
@cole-bq No problem. I can change on Janusgraph side to avoid the performance downgrade.

xiazcy:
Closing PR as there is alternative solution.

PR #1875, opened 2022-11-28
TINKERPOP-2348 Add nullable annotations also to tests

FlorianHockmann:
This is a follow-up to #1867 to now also enable nullable annotations to tests.

https://issues.apache.org/jira/browse/TINKERPOP-2348

Only the generated Gherkin test traversals are missing as we would have to let the Java `DotNetTranslator` produce C# code with nullable annotations. I think we should do that at some point in the future as nullable annotations are only enabled since .NET 6 by default and only for new projects. So, many users are probably still not using them.

Adding the annotations also in tests revealed two (well actually three) more places where `null` should be allowed in Gremlin.Net itself: In `Traversal.Equals()` and in the `From()`/`To()` steps if the argument is a `Vertex`. While `null` is still not really allowed there, Gremlin throws a more meaningful exception message if it gets `null` and we might change that in the future to somehow allow `null` maybe.

VOTE +1


gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/CommonSteps.cs
```diff
@@ -127,20 +127,20 @@ public void TranslateTraversal(string traversalText)
                 throw new InvalidOperationException("g should be a traversal source");
             }
 
-            if (ScenarioData.CurrentFeature.Tags.Select(t => t.Name).ToList().Contains("@GraphComputerOnly"))
+            if (ScenarioData.CurrentFeature!.Tags.Select(t => t.Name).ToList().Contains("@GraphComputerOnly"))
```
vkagamlyk, line 130: why is there `.ToList()` needed? 
```suggestion
            if (ScenarioData.CurrentFeature!.Tags.Any(t => t.Name == "@GraphComputerOnly"))
```

FlorianHockmann: It shouldn't be needed. I updated the commit to include your suggestion.

vkagamlyk:
VOTE +1

PR #1876, opened 2022-11-28
TINKERPOP-2480 Add User-Agent to Dotnet Driver

cole-bq:

Adds a user agent to gremlin-dotnet which is sent as a request header during the web socket handshake.

This behavior is enabled by default but can be disabled by setting the ConnectionPoolSettings.EnableUserAgentOnConnect configuration to false.

Note: There are no tests included as part of this PR. My intention is to add integration tests to ensure that the user agent is being sent correctly once [TINKERPOP-2819](https://issues.apache.org/jira/browse/TINKERPOP-2819) has been merged. The existing dotnet glv tests are sufficient to show that this change does not introduce any faults into the driver. I have conducted manual tests to ensure that the changes are currently functioning correctly. My thoughts are that with a potential release approaching, it is worthwhile to have the user agent introduced now, with full validation coming soon.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -36,6 +36,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Async operations in .NET can now be cancelled. This however does not cancel work that is already happening on the server.
 * Bumped to `snakeyaml` 1.32 to fix security vulnerability.
 * Added user agent to web socket handshake in java driver. Can be controlled by a new enableUserAgentOnConnect configuration. It is enabled by default.
+* Added user agent to web socket handshake in Gremlin.Net driver. Can be controlled by `EnableUserAgentOnConnect` in `ConnectionPoolSettings`. It is enabled by default.
 * Added logging in .NET.
 * Added `addDefaultXModule` to `GraphSONMapper` as a shortcut for including a version matched GraphSON extension module.
 * Modified `GraphSONRecordReader` and `GraphSONRecordWriter` to include the GraphSON extension module by default.
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/WebSocketSettings.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/WebSocketSettings.cs
@@ -41,6 +41,8 @@ namespace Gremlin.Net.Driver
         /// </summary>
         public Action<ClientWebSocketOptions> WebSocketConfigurationCallback { get; set; }
 
+        public bool EnableUserAgentOnConnect { get; set; }
+
 #if NET6_0_OR_GREATER
         /// <summary>
         ///     Gets or sets whether compressions will be used. The default is true. (Only available since .NET 6.)
--- a/gremlin-dotnet/src/Gremlin.Net/Process/Utils.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Process/Utils.cs
@@ -22,8 +22,10 @@
 #endregion
 
 using System;
+using System.Reflection;
 using System.Runtime.ExceptionServices;
 using System.Threading.Tasks;
+using static System.Runtime.InteropServices.RuntimeInformation;
 
 namespace Gremlin.Net.Process
 {
@@ -32,6 +34,9 @@ namespace Gremlin.Net.Process
     /// </summary>
     internal static class Utils
     {
+        public static string UserAgent => _userAgent ??= GenerateUserAgent();
+        private static string _userAgent;
+        
         /// <summary>
         /// Waits the completion of the provided Task.
         /// When an AggregateException is thrown, the inner exception is thrown.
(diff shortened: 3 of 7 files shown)


gremlin-dotnet/src/Gremlin.Net/Process/Utils.cs
```diff
@@ -63,5 +66,23 @@ public static void Forget(this Task task)
...
+
+        /// <summary>
+        ///  Returns a user agent for connection request headers.
+        ///
+        /// Format:
+        /// "[Application Name] [GLV Name].[Version] [Language Runtime Version] [OS].[Version] [CPU Architecture]"
+        /// </summary>
+        private static string GenerateUserAgent()
+        {
+            var applicationName = Assembly.GetExecutingAssembly().GetName().Name.Replace(' ', '_');
```
FlorianHockmann, line 78: `Name` could be `null` which would throw a `NullReferenceException`. Please use conditional access:
```suggestion
            var applicationName = Assembly.GetExecutingAssembly().GetName().Name?.Replace(' ', '_');
```
That way, `applicationName` will be `null` which isn't that nice, but at least better than getting an exception in my opinion.

gremlin-dotnet/src/Gremlin.Net/Process/Utils.cs
```diff
@@ -63,5 +66,23 @@ public static void Forget(this Task task)
...
+            var driverVersion = AssemblyName.GetAssemblyName("Gremlin.Net.dll")?.Version.ToString().Replace(' ', '_');
```
FlorianHockmann, line 79: `GetAssemblyName()` can't return `null` if I'm not mistaken. It will instead throw if it cannot find the assembly (which should never happen), but `Version` can be `null`. So please change this to:
```suggestion
            var driverVersion = AssemblyName.GetAssemblyName("Gremlin.Net.dll").Version?.ToString().Replace(' ', '_');
```

gremlin-dotnet/src/Gremlin.Net/Process/Utils.cs
```diff
@@ -63,5 +66,23 @@ public static void Forget(this Task task)
...
+            var languageVersion = Environment.Version.ToString().Replace(' ', '_');
+            var osName = Environment.OSVersion.Platform.ToString().Replace(' ', '_');
+            var osVersion = Environment.OSVersion.Version.ToString().Replace(' ', '_');
+            var cpuArchitecture = System.Runtime.InteropServices.RuntimeInformation.OSArchitecture.ToString().Replace(' ', '_');
```
FlorianHockmann, line 83: (nitpick) Please wrap this long line:
```suggestion
            var cpuArchitecture = System.Runtime.InteropServices.RuntimeInformation.OSArchitecture.ToString()
                .Replace(' ', '_');
```

or import this static classes so they don't need to be written out here with there namespaces.

gremlin-dotnet/src/Gremlin.Net/Process/Utils.cs
```diff
@@ -32,6 +33,8 @@ namespace Gremlin.Net.Process
     /// </summary>
     internal static class Utils
     {
+        public static readonly string UserAgent = GenerateUserAgent();
```
FlorianHockmann, line 36: A public field is usually discouraged in C#. It isn't a big problem here as this class is `internal`, but I would still suggest that only make properties and methods `public`. My suggestion is to turn this into a simple property with a backing field:
```suggestion
        public static string UserAgent => _userAgent ??= GenerateUserAgent();
        private static string _userAgent;
```

gremlin-dotnet/src/Gremlin.Net/Driver/WebSocketSettings.cs
```diff
@@ -22,6 +22,7 @@
 #endregion
 
 using System;
+using System.ComponentModel;
```
FlorianHockmann, line 25: Remove unused import

CHANGELOG.asciidoc
```diff
@@ -35,6 +35,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Dockerized all test environment for .NET, JavaScript, Python, Go, and Python-based tests for Console, and added Docker as a build requirement.
 * Async operations in .NET can now be cancelled. This however does not cancel work that is already happening on the server.
 * Bumped to `snakeyaml` 1.32 to fix security vulnerability.
+* Added user agent to web socket handshake in dotnet driver. Can be controlled by EnableUserAgentOnConnect in `ConnectionPoolSettings`. It is enabled by default.
```
FlorianHockmann, line 38: Please either change this from _dotnet driver_ to _the Gremlin.Net driver_ or _the .NET driver_.
And _EnableUserAgentOnConnect_ should also be marked as code, so: \`EnableUserAgentOnConnect\`

FlorianHockmann (commented):
The changes look good to me basically. I added some comments, but they are basically only about stylistic issues.
I've also debugged this locally on my machine and the user-agent it produced looked good to me.

gremlin-dotnet/src/Gremlin.Net/Process/Utils.cs
cole-bq: Thanks I have made this change and also provided a default value of "NotAvailable" in case the result is null.

cole-bq:
Thanks for all the comments @FlorianHockmann. I have pushed an updated commit with all of your suggestions implemented.

FlorianHockmann (approved):
VOTE +1

PR #1878, opened 2022-11-29
TINKERPOP-2804 Moved gherkin files to gremlin-test resources

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2804

Taking this approach makes it easier for providers to reference the files from the classpath, otherwise they likely need to have copies of them on the local file system which isn't so good. This is a breaking change but it only affects providers. They will have a minor adjustment to make to their test infrastructure on upgrade.

VOTE +1


vkagamlyk:
.net local test run without docker: `System.IO.DirectoryNotFoundException : Could not find a part of the path 'C:\_Projects\tinkerpop-apache\gremlin-test\features'.`
because https://github.com/apache/tinkerpop/blob/3.6-dev/gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/GherkinTestRunner.cs#L385

vkagamlyk:
VOTE +1

PR #1882, opened 2022-12-02
TINKERPOP-2813 Make gremlin-driver more optimistic about host availability

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2813

Summarized the major points of the change in the Upgrade Docs. Might be best to start there if reviewing this. Made some other adjustments to log better data for debugging purposes with the `ConnectionPool`.  Ended up making a number of adjustments to the thread pools to which various jobs were being submitted. Untangling that led to the addition of two pools which helped bring back parallel host initialization and got rid of random usage of the fork/join pool. I sense that something could be done to eliminate some of these pools but I believe it will come at the cost of a much larger refactoring effort to untangle how some of these jobs get scheduled up. I dunno.....not super pleased with it but gotta go pencils down at some point.

there've been many days of continuous testing on these changes at this point from various folks and all results have been favorable. calling out @kenhuuu specifically, he did some basic performance testing and profiling and didn't find any problems to note:

```text
Test Case:
    2450 short queries (return in milliseconds)
    1280 mix of short and medium queries

Setup:
    m5.2xlarge
    3.5.4 Gremlin Server with gremlin-server.yaml
    Defaults for connection pooling and in flight per connection settings for Java driver

Results for Three Run Average.
    3.5.4:
        Short: 29912ms
        Mixed: 65689ms

    TINKERPOP-2813 branch:
        Short: 30161ms
        Mixed: 65422ms
```

### DIFF
--- a/docs/src/reference/gremlin-variants.asciidoc
+++ b/docs/src/reference/gremlin-variants.asciidoc
@@ -658,6 +658,31 @@ social.persons("marko").knows("josh");
 NOTE: Using Maven, as shown in the `gremlin-archetype-dsl` module, makes developing DSLs with the annotation processor
 straightforward in that it sets up appropriate paths to the generated code automatically.
 
+[[gremlin-java-troubleshooting]]
+=== Troubleshooting
+
+*Max frame length of 65536 has been exceeded*
+
+This error occurs when the driver attempts to process a request/response that exceeds the configured maximum size.
+The most direct way to fix this problem is to increase the `maxContentLength` setting in the driver. Ideally, the
+`maxContentLength` set for the driver should match the setting defined on the server.
+
+*TimeoutException*
+
+A `TimeoutException` is thrown by the driver when the time limit assigned by the `maxWaitForConnection` is exceeded
+when trying to borrow a connection from the connection pool for a particular host. There are generally two scenarios
+where this occurs:
+
+1. The server has actually reached its maximum capacity or the driver has just learned that the server is unreachable.
+2. The client is throttling requests.
+
+The latter of the two can be addressed from the driver side in the following ways:
+
+* Increase the `maxWaitForConnection` allowing the client to wait a bit longer for a connection to become available.
+* Increase the number of requests allowed per connection by increasing the `maxSimultaneousUsagePerConnection` and
+`maxInProcessPerConnection` settings.
+* Increase the number of connections available in the connection pool by increasing the `maxConnectionPoolSize`.
+
 anchor:java-application-examples[]
 anchor:gremlin-archetypes[]
 [[gremlin-java-examples]]
--- a/docs/src/reference/gremlin-variants.asciidoc
+++ b/docs/src/reference/gremlin-variants.asciidoc
@@ -674,7 +674,7 @@ when trying to borrow a connection from the connection pool for a particular hos
 where this occurs:
 
 1. The server has actually reached its maximum capacity or the driver has just learned that the server is unreachable.
-2. The client is throttling requests.
+2. The client is throttling requests when the pool is exhausted.
 
 The latter of the two can be addressed from the driver side in the following ways:
 
@@ -682,6 +682,9 @@ The latter of the two can be addressed from the driver side in the following way
 * Increase the number of requests allowed per connection by increasing the `maxSimultaneousUsagePerConnection` and
 `maxInProcessPerConnection` settings.
 * Increase the number of connections available in the connection pool by increasing the `maxConnectionPoolSize`.
+
+The exception and logs (assuming they are enabled) should contain information about the state of the connection pool
+along with its connections which can help shed more light on which of these scenarios caused the problem.
 
 anchor:java-application-examples[]
 anchor:gremlin-archetypes[]
--- a/docs/src/upgrade/release-3.5.x.asciidoc
+++ b/docs/src/upgrade/release-3.5.x.asciidoc
@@ -50,6 +50,43 @@ var client = new GremlinClient(new GremlinServer("localhost", 8182), loggerFacto
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-2471[TINKERPOP-2471]
 
+==== gremlin-driver Host Availability
+
+The Java drivers approach to determine the health of the host to which it is connected is important to how it behaves.
+The driver has generally taken a pessimistic approach to make this determination, where a failure to borrow a
+connection from the pool would constitute enough evidence to mark the host as dead. Unfortunately a failure to borrow
+from the pool is not the best indicator for a dead host. Intermittent network failure, an overly busy client, or other
+temporary issues could have been at play while the server was perfectly available.
+
+The consequences for marking a host dead are fairly severe as the connection pool is shutdown. While it is shutdown
+in an orderly fashion so as to allow for existing requests to complete if at all possible, it immediately blocks new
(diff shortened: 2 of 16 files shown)


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
```diff
@@ -495,49 +496,45 @@ public Client alias(final Map<String, String> aliases) {
...
 
-            // you can get no possible hosts in more than a few situations. perhaps the servers are just all down.
-            // or perhaps the client is not configured properly (disables ssl when ssl is enabled on the server).
-            if (!possibleHosts.hasNext())
-                throwNoHostAvailableException();
-
-            final Host bestHost = possibleHosts.next();
+            // try a random host if none are marked available. maybe it will reconnect in the meantime. better than
+            // going straight to a fast NoHostAvailableException as was the case in versions 3.5.4 and earlier
+            final Host bestHost = possibleHosts.hasNext() ? possibleHosts.next() : chooseRandomHost();
```
vkagamlyk, line 513: For example we have a cluster of 3 instances.
For some reason (reconfiguration, scaling, network fluke) all instances restarted.
After some time can we connect to one random host and use it, but all the others will remain disconnected.
Maybe it makes sense to check all unavailable hosts when the user sends a request, but not more than once every 10-60 seconds?

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -658,6 +658,31 @@ social.persons("marko").knows("josh");
...
+*TimeoutException*
+
+A `TimeoutException` is thrown by the driver when the time limit assigned by the `maxWaitForConnection` is exceeded
+when trying to borrow a connection from the connection pool for a particular host. There are generally two scenarios
+where this occurs:
+
+1. The server has actually reached its maximum capacity or the driver has just learned that the server is unreachable.
+2. The client is throttling requests.
+
+The latter of the two can be addressed from the driver side in the following ways:
```
FlorianHockmann, line 679: Not sure if that's necessary at all, but how about mentioning that the exception message / the error log message should contain information about the state of the `ConnectionPool` and its `Connections` to understand which of the two cases caused the problem?

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -658,6 +658,31 @@ social.persons("marko").knows("josh");
...
+`maxContentLength` set for the driver should match the setting defined on the server.
+
+*TimeoutException*
+
+A `TimeoutException` is thrown by the driver when the time limit assigned by the `maxWaitForConnection` is exceeded
+when trying to borrow a connection from the connection pool for a particular host. There are generally two scenarios
+where this occurs:
+
+1. The server has actually reached its maximum capacity or the driver has just learned that the server is unreachable.
+2. The client is throttling requests.
```
FlorianHockmann, line 677: To be honest, when I just read this once it was not directly clear to me what that means. Why is it throttling requests?
Simply adding "[...] because the pool is exhausted" would already help.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/ClientConnectionIntegrateTest.java
```diff
@@ -110,4 +124,100 @@ public void shouldCloseConnectionDeadDueToUnRecoverableError() throws Exception
...
+                    // this condition inside this thread or the test locks up
+                    hadFailOtherThanTimeout.compareAndSet(false, !(ex.getCause() instanceof TimeoutException));
+                } finally {
+                    latch.countDown();
+                }
+            });
+        }, "worker-shouldSucceedWithJitteryConnection").start();
+
+        // wait long enough for the jitters to kick in at least a little
+        while (latch.getCount() > 500) {
```
FlorianHockmann, line 174: (nitpick) Is this `500` because then half the requests have been executed? In that case, I'd change this to `requests / 2` to make this clear.

docs/src/upgrade/release-3.5.x.asciidoc
```diff
@@ -50,6 +50,48 @@ var client = new GremlinClient(new GremlinServer("localhost", 8182), loggerFacto
...
+The driver has generally taken a pessimistic approach to make this determination, where a failure to borrow a
+connection from the pool would constitute enough evidence to mark the host as dead. Unfortunately a failure to borrow
+from the pool is not the best indicator for a dead host. Intermittent network failure, an overly busy client, or other
+temporary issues could have been at play while the server was perfectly available.
+
+The consequences for marking a host dead are fairly severe as the connection pool is shutdown. While it is shutdown
+in an orderly fashion so as to allow for existing requests to complete if at all possible, it immediately blocks new
+requests producing an immediate `NoHostAvailableException` (assuming there are actually no other hosts to route
+requests to). There is then some delay to reconnect to the server and re-establish each of the connections in the pool,
+which might take some time especially if there was a large pool. If an application were sending hundreds of request
```
FlorianHockmann, line 65:
```suggestion
which might take some time especially if there was a large pool. If an application were sending hundreds of requests
```

FlorianHockmann (commented):
Just reviewed this, mostly to better understand the improvement you're making to the driver here myself.

Added some inline comments, but they are just nits or questions I had.

kenhuuu:
LGTM. I've tested this and it greatly improves the handling of temporary errors.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/ClientConnectionIntegrateTest.java
```diff
@@ -21,20 +21,34 @@
...
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.number.OrderingComparison.greaterThan;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 public class ClientConnectionIntegrateTest extends AbstractGremlinServerIntegrationTest {
+    private static final org.slf4j.Logger logger = LoggerFactory.getLogger(ClientConnectionIntegrateTest.class);
```
vkagamlyk, line 51: looks redundant, tests use log4j

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/ClientConnectionIntegrateTest.java
```diff
@@ -110,4 +124,100 @@ public void shouldCloseConnectionDeadDueToUnRecoverableError() throws Exception
...
+        // wait for requests to complete
+        assertTrue(latch.await(30000, TimeUnit.MILLISECONDS));
+
+        // make sure we had some failures for sure coming out the factory
+        assertThat(connectionFactory.getNumberOfFailures(), is(greaterThan(0L)));
+
+        // if there was a exception in the worker thread, then it had better be a TimeoutException
+        assertThat(hadFailOtherThanTimeout.get(), is(false));
+
+        connectionFactory.jittery = false;
```
vkagamlyk, line 187: This `connectionFactory` is not used more in this scope
```suggestion
```

spmallette:
@FlorianHockmann Thanks for the review. I addressed your points. More refinements to the documentation are coming I think, as @kenhuuu will be adding a follow-on PR with more tests. I'll follow up that with a few more documentation adjustments.

PR #1883, opened 2022-12-02
 TINKERPOP-2480 Add User-Agent to Javascript Driver

cole-bq:

Adds a user agent to gremlin-javascript which is sent as a request header during the web socket handshake.

This behavior is enabled by default but can be disabled by setting the enableUserAgentOnConnect option to false.

Note: There are no tests included as part of this PR. My intention is to add integration tests to ensure that the user agent is being sent correctly once [TINKERPOP-2819](https://issues.apache.org/jira/browse/TINKERPOP-2819) has been merged. The existing javascript glv tests are sufficient to show that this change does not introduce any faults into the driver. I have conducted manual tests to ensure that the changes are currently functioning correctly. My thoughts are that with a potential release approaching, it is worthwhile to have the user agent introduced now, with full validation coming soon.

This PR also adds a configuration section to the JS reference documentation in order to be consistent with the other GLV's. The configuration table is populated based on the JSDoc comments from the Connection, Client, and DriverRemoteConnection constructors.


gremlin-javascript/src/main/javascript/gremlin-javascript/.eslintrc.js
```diff
@@ -26,7 +26,7 @@ module.exports = {
     node: true,
   },
   parserOptions: {
-    ecmaVersion: 2017,
+    ecmaVersion: 2020,
```
vkagamlyk, line 29: is this a breaking change?

cole-bq: I bumped the version on eslint as it was blocking the use of the ?? operator which has been available since node 14. It was my understanding that tinkerpop required node >= 16. Upon further investigation, package.json node only needs to be >=10 for tinkerpop. I will revert this change and find a replacement for ?? wherever I've used it.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
```diff
@@ -98,6 +99,10 @@ class Connection extends EventEmitter {
     this.isOpen = false;
     this.traversalSource = options.traversalSource || 'g';
     this._authenticator = options.authenticator;
+    if (options.enableUserAgentOnConnect === undefined) {
+      options.enableUserAgentOnConnect = true;
+    }
+    this._enableUserAgentOnConnect = options.enableUserAgentOnConnect;
```
vkagamlyk, line 105:
```suggestion
    this._enableUserAgentOnConnect = options.enableUserAgentOnConnect !== false;
```

cole-bq: Good suggestion, I like that better.

jorgebay:
Thanks @cole-bq, looks great!

xiazcy:
Thanks Cole! VOTE +1

vkagamlyk:
VOTE +1

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/utils.js
```diff
@@ -24,6 +24,8 @@
 'use strict';
 
 const crypto = require('crypto');
+const os = require('os');
+const gremlinVersion = require(__dirname + '/../package.json').version;
```
alexey-temnikov, line 28: Let's confirm this still works even when the target client application is compiled in minimized js package (so reference to JSON file is not lost).

cole-bq: I tried using webpack to package and minimize a test client app and the driver version was still being correctly loaded into the user agent without issue.

PR #1914, opened 2022-12-19
Fix RequestId Override

Cole-Greer:
I believe this fix is small enough for review to be done in CTR fashion. If any commiter agrees with this I would like to merge this in without a lengthy initial review.

Fixes a bug where python and go glv's were not using the correct request_id when overriden. This is detailed in TINKERPOP-2841 and TINKERPOP-2844. This commit is not sufficient to close those tickets as further investigation is needed to ensure the other per-request settings work as intended, this commit provides an immediate fix to the broken request_id functionality.

Also fixes a typo in the python reference docs.


vkagamlyk:
Thanks Cole!
VOTE +1

xiazcy:
VOTE +1. Merging this small bug fix as a CTR.

PR #1922, opened 2022-12-29
Add LICENSE and NOTICE files for gremlin-go

Cole-Greer:
Just a quick update to license files. Adding LICENSE and NOTICE to gremlin-go as per recent discussion in the dev list. Also a minor fix to filepaths referenced in the root LICENSE file.


xiazcy:
Since we've discussed on dev list, I'm merging this via CTR

PR #1937, opened 2023-01-11
Improved user agent handling of cases where gremlin version is not available

Cole-Greer:
It was observed in the java driver that the UserAgent class could fail to load if the manifest file was excluded from the jar. This is because the user agent relies on the manifest to extract the current version of gremlin. This PR adds a bit of extra error handling and will insert a value of "NotAvailable" in case of any failures. A similar issue was also present in JS which is also fixed here.

This PR also updates Gremlin.version() to return "VersionNotFound" if the manifest read fails as this is preferable to the current behavior of throwing a ExceptionInInitializerError.

Thanks @kenhuuu for discovering this issue.


vkagamlyk:
Thank you Cole!
VOTE +1

xiazcy:
VOTE +1

lyndonbauto:
VOTE+1.

PR #1938, opened 2023-01-11
[TINKERPOP-2844] Adds tests for per-request-settings in GLV's

Cole-Greer:
https://issues.apache.org/jira/browse/TINKERPOP-2844

Creates a test where a driver sends a request to gremlin-socket-server with all per-request settings set. gremlin-socket-server extracts these arguments and returns them in a string to the driver.

I expanded the scope of the ticket a bit and added this test to Java, .net, and JS in addition to Python. Go is excluded from this commit as the test is failing and needs additional work. The go per request settings issue is being tracked by [TINKERPOP-2841](https://issues.apache.org/jira/browse/TINKERPOP-2841)

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/test/integration/client-behavior-tests.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/test/integration/client-behavior-tests.js
@@ -47,7 +47,6 @@ describe('Client', function () {
             assert.equal(connectionClosed, true);
 
             let result = await client.submit('1', null, {requestId: settings.SINGLE_VERTEX_REQUEST_ID})
-            console.log("result received: "+JSON.stringify(result));
             assert.ok(result);
         });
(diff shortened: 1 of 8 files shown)


gremlin-javascript/src/main/javascript/gremlin-javascript/test/integration/client-behavior-tests.js
```diff
@@ -64,5 +63,15 @@ describe('Client', function () {
 
             await noUserAgentClient.close();
         });
+        it('should send per request settings to server', async function () {
+            let resultSet = await client.submit('1', null, {
```
vkagamlyk, line 67: better to use `const` to declare immutable variable

vkagamlyk:
VOTE +1

xiazcy:
Thanks! Merging now given lazy consensus.

PR #1939, opened 2023-01-11
[TINKERPOP-2841] Test and Fix Per-Request Settings in Go GLV

Cole-Greer:
https://issues.apache.org/jira/browse/TINKERPOP-2841

In my testing, the go glv was not sending per-request options correctly to the server. Go was passing request options inside the gremlin script bindings map instead of within the args portion of the request message.

This PR re-implements the per-request settings by adding new Client.SubmitWithOptions() and DriverRemoteConnection.SubmitWithOptions() methods. Both original Submit() methods are unchanged in their behavior so this will not break any users.

This PR introduces a new `RequestOptions` struct and builder which is used to configure per-request settings and passed into the new SubmitWithOptions() methods.


xiazcy:
LGTM, thanks Cole! VOTE +1

vkagamlyk:
VOTE +1

PR #1940, opened 2023-01-13
TINKERPOP-2847 Compare and store request UUIDs as lower case strings in gremlin-javascript.

kenhuuu:
Fixes https://issues.apache.org/jira/browse/TINKERPOP-2847.

Storing and comparing the request IDs in lower case prevents issues that arise when the supplied ID and returned ID are using different cases to represent the hex values of the UUID.


vkagamlyk:
VOTE +1

xiazcy:
VOTE +1 pending change log entry.

lyndonbauto:
VOTE+1.

PR #1946, opened 2023-01-16
[TINKERPOP-2767] Fix server not handling StackOverflowError in bytecode traversals

Cole-Greer:
https://issues.apache.org/jira/browse/TINKERPOP-2767

The existing error handling in TraversalOpProcessor and SessionOpProcessor during bytecode iteration was catching all exceptions sending an appropriate error response to the client. The OpProcessor however was not catching a StackOverflowError which could be induced by running a query which contains a large repeat step. This Error was being caught by FutureTask.run() but GremlinServer never wait's on this task or checks the results which caused this Error to be lost. A similar issue was found to exist in AbstractSession.process().

This PR adjusts the existing error handling code to catch any Throwable during bytecode iteration so clients will receive error messages and codes for server errors as well as exceptions. Any errors are re-thrown such that the evalFuture FutureTask will continue to have an exception set correctly (although GremlinServer currently does not use this for anything).

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -28,6 +28,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Changed `PartitionStrategy` to force its filters to the end of the chain for `Vertex` and `Edge` read steps, thus preserving order of the `has()`.
 * Added `RequestOptions` and `RequestOptionsBuilder` types to Go GLV to encapsulate per-request settings and bindings.
 * Added `SubmitWithOptions()` methods to `Client` and `DriverRemoteConnection` in Go GLV to pass `RequestOptions` to the server.
+* Fixed bug in which `gremlin-server` would not respond to clients if an `Error` was thrown during bytecode traversals.
 
 [[release-3-5-5]]
 === TinkerPop 3.5.5 (Release Date: January 16, 2023)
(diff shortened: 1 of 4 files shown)


xiazcy:
I think there is the question of do we want to send all of the error messages to the client? I feel like we should handle the hanging in the driver, but it may not be necessary to dump all stack trace?

Cole-Greer:
> 

I structured this PR in this manner to mirror the behaviour of gremlin script executions in addition to this being the solution with the smallest changes. I think it's fair to have a discussion if this is actually the ideal behaviour. Perhaps this is worthy of a discussion in the dev list? I believe if we decide a different response behaviour is preferable for bytecode traversals, that we should apply that same change to script-based traversals.

Cole-Greer:
Based on discussion in the dev list, I created [TINKERPOP-2866](https://issues.apache.org/jira/browse/TINKERPOP-2866) which I would consider separate from the issue this PR is resolving. I would ask that we proceed with merging this PR if there are no objections.

xiazcy:
I don't have a problem with this change. I'd add a CHANGELOG entry since this is alters expected error output on driver. Otherwise VOTE +1.

vkagamlyk:
VOTE +1

xiazcy:
Just to note that we'll need a PR to 3.6-dev separately for these changes.

PR #1947, opened 2023-01-17
Adds CodeQL vulnerability scanning to Github Actions

Cole-Greer:
Adds new workflow which runs CodeQL vulnerability scanning on any pushes or pull requests to master, 3.5-dev, or 3.6-dev.

Also includes a small fix in go to prevent a possible lossy type conversion.

CodeQL will result in a failing check if a push or PR introduces a new vulnerability. It will not fail the checks for vulnerabilities which were pre-existing in the branch. Existing vulnerabilities are tracked and listed in Github under Security -> Vulnerability and Alerts -> Code scanning (Only visible to committers)

There are 3 vulnerabilities which have been found in the initial scan. I would ask that a committer dismiss the go warning as a "false positive". The warning is due to a potentially unsafe type conversion from a uint64 to a uint. During it's initialization, the size of the uint64 is being bounded to the size of uint making a lossy conversion impossible.

The remaining 2 vulnerabilities are due to inefficient regex in prism.js which is is included in our sources as part of the website. This seems like a low priority to me and I would suggest we either dismiss these warnings as "won't do" or leave them open for added visibility with no immediate plan for a patch.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,99 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master", "3.5-dev", "3.6-dev" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "master", "3.5-dev", "3.6-dev" ]
+  schedule:
+    - cron: '22 8 * * 1'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'go', 'java', 'javascript', 'python' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
+        # Use only 'java' to analyze code written in Java, Kotlin or both
+        # Use only 'javascript' to analyze code written in JavaScript, TypeScript or both
+        # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      # Initializes the CodeQL tools for scanning.
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v2
+        with:
+          languages: ${{ matrix.language }}
+          # If you wish to specify custom queries, you can do so here or in a config file.
+          # By default, queries listed here will override any specified in a config file.
+          # Prefix the list here with "+" to use these queries and those in the config file.
+
+          # Details on CodeQL's query packs refer to : https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs
+          # queries: security-extended,security-and-quality
+
+
+      # Autobuild attempts to build any compiled languages  (C/C++, C#, Go, or Java).
+      # If this step fails, then you should remove it and run the build manually (see below)
+      - name: Autobuild
+        uses: github/codeql-action/autobuild@v2
+
+      # ℹ️ Command-line programs to run using the OS shell.
+      # 📚 See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun
+
+      #   If the Autobuild fails above, remove it and uncomment the following three lines.
+      #   modify them (or add more) to build your code if your project, please refer to the EXAMPLE below for guidance.
+
+      # - run: |
+      #   echo "Run, Build Application using script"
+      #   ./location_of_script_within_repo/buildscript.sh
+
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v2
+        with:
+          category: "/language:${{matrix.language}}"
+
+  # .Net is run as a separate job as the integration tests take far too long to build
+  analyze_dotnet:
+    name: Analyze .Net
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,104 @@
...
+      # - run: |
+      #   echo "Run, Build Application using script"
+      #   ./location_of_script_within_repo/buildscript.sh
+
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v2
+        with:
+          category: "/language:${{matrix.language}}"
+
+  # .Net is run as a separate job as the integration tests take far too long to build
```
FlorianHockmann, line 67: Hmm, that's unfortunate and I'm wondering why this is the case. The normal build without executing any tests usually shouldn't take long.
But sure, manually building .NET should also not be a problem.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,104 @@
...
+    name: Analyze .Net
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
```
FlorianHockmann, line 78: Not that important, but does it really make sense to use `matrix` here if the `run` below is completely specific to .NET? I think it would be simpler without the `matrix`.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,104 @@
...
+        with:
+          languages: ${{ matrix.language }}
+
+      - run: |
+          echo "Build .NET"
+          dotnet build gremlin-dotnet/src/Gremlin.Net;
+          dotnet build gremlin-dotnet/src/Gremlin.Net.Template
+          dotnet build gremlin-dotnet/test/Gremlin.Net.UnitTest
+          dotnet build gremlin-dotnet/test/Gremlin.Net.Benchmarks
+#          Excluded due to extremely long (~45m) build time in github actions
```
FlorianHockmann, line 97: Wow, that's really extreme. Would be interesting to know of course what's taking so long here, but it's also not that important, especially for this PR. I also don't think that we really need to perform the scanning on test projects as we don't ship them to users.

FlorianHockmann (commented):
Thanks a lot for tackling this! The changes also look good to me basically. The .NET scan however includes this in its logs:

```
Analysis produced the following diagnostic data:

|          Diagnostic          |      Summary       |
+------------------------------+--------------------+
| Extraction errors            | 1 result (1 error) |
| Compilation error            | 8 results          |
| Compilation message          | 16 results         |
| Successfully extracted files | 272 results        |
```

Doesn't this mean that it found something? And shouldn't that be reported?

.github/workflows/codeql.yml
Cole-Greer: The `matrix` here is unnecessary. My reasoning for including it was in case eventually another language also needs to be excluded from autobuild. I don't have a strong preference either way. I will remove the `matrix` to simplify the current version of it.

.github/workflows/codeql.yml
Cole-Greer: Yeah I was pretty surprised by that result but in my tests it was very consistent and reproducible. I'm sure a build system could be found which doesn't run into whatever bottleneck this is hitting. It doesn't seem like a much of a priority at this time though.

Cole-Greer:
> Thanks a lot for tackling this! The changes also look good to me basically. The .NET scan however includes this in its logs:
> 
> ```
> Analysis produced the following diagnostic data:
> 
> |          Diagnostic          |      Summary       |
> +------------------------------+--------------------+
> | Extraction errors            | 1 result (1 error) |
> | Compilation error            | 8 results          |
> | Compilation message          | 16 results         |
> | Successfully extracted files | 272 results        |
> ```
> 
> Doesn't this mean that it found something? And shouldn't that be reported?

The diagnostics here is referring to what code it was able to analyze, not what the results of that analysis was. The extraction error there means that there was one file which codeQL failed to extract for analysis. We could potentially learn more from running codeQL in debug mode. According to the codeQL [docs](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/troubleshooting-the-codeql-workflow#extraction-errors-in-the-database): "A small number of extractor errors is healthy and typically indicates a good state of analysis."

The results of a single job can be viewed from the "code scanning results" section of the checks tab ([here for this PR](https://github.com/apache/tinkerpop/pull/1947/checks?check_run_id=10709124951)).

All of the analysis results for the repo are also collected in the security tab [here](https://github.com/apache/tinkerpop/security/code-scanning). (Hopefully that link works, it is only accessible to committers).

FlorianHockmann:
> The diagnostics here is referring to what code it was able to analyze, not what the results of that analysis was. The extraction error there means that there was one file which codeQL failed to extract for analysis. We could potentially learn more from running codeQL in debug mode. According to the codeQL [docs](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/troubleshooting-the-codeql-workflow#extraction-errors-in-the-database): "A small number of extractor errors is healthy and typically indicates a good state of analysis."

Thanks for the explanation, that's good to know.

> All of the analysis results for the repo are also collected in the security tab [here](https://github.com/apache/tinkerpop/security/code-scanning). (Hopefully that link works, it is only accessible to committers).

Yep, that's working. Looks good to me.

I also agree with your assessment of the 3 warnings it found. I'll just leave the Go one open for someone with more Go experience to confirm & mark as a FP.

Overall, this is good to go from my side. VOTE +1
But this probably has to wait a bit as we're currently in code freeze.

xiazcy:
LGTM, merging this in now.

PR #1950, opened 2023-01-24
TINKERPOP-2647 Deactivate active plugins during Uninstallation

AmatyaAvadhanula:
Fixes https://issues.apache.org/jira/browse/TINKERPOP-2647

When a plugin is used and uninstalled without being deactivated first, the console would fail to start up. This PR aims to deactivate such plugins when uninstall is called to prevent errors.

Ran `mvn clean install` to success and performed manual testing in the Gremlin Console

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -36,6 +36,7 @@ This release also includes changes from <<release-3-6-XXX, 3.6.XXX>>.
 * Moves `SimpleSocketServer` and its initializers to a new `gremlin-tools/gremlin-socket-server` module.
 * Configures `gremlin-socket-server` to build a docker image which can be used for testing GLV's. (Can be skipped with -DskipImageBuild)
 * Reduces dependency from `gremlin-server` onto `gremlin-driver` to a test scope only.
+* TINKERPOP-2647 Disable active plugins on uninstall to prevent errors during startup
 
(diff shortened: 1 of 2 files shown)


Cole-Greer:
Thanks for your contributions here. Just so you are aware, we are currently in the middle of a code freeze for the upcoming 3.5.5 and 3.6.2 releases, as such there likely won't be much activity on PR's for the time being and it won't be merged until after the releases are done.

I'm not too familiar with the console codebase so I cannot comment with too much authority here but the changes look good to me. I would prefer if the solution did not require building and iterating through so many hash sets. I would hope there is a simple way to check if the specific module being uninstalled also needs to be deactivated without looking at all the plugins. Without looking into it further though I'm not sure if this is indeed possible without further changes. There should never be that many plugins installed at any time so I don't think it's a big issue as it is currently implemented.

One ask I have is that you base these changes off of the 3.5-dev branch instead of master. Our branch structure is that we target non breaking changes towards the earliest active development branch (currently 3.5-dev) and when a PR is merged, the changes will be pulled forward into the newer development branches as well (3.6-dev, master).

Please let me know if you have any questions about this or if I can help in any way.

AmatyaAvadhanula:
Thank you @Cole-Greer. I'll look further to see if the approach can be simplified and base the changes off of 3.5-dev.

PR #1952, opened 2023-01-30
patch: Adding for null checks in generateUserAgent() to avoid undefined runtime errors

iAmmarTahir:
While using Gremlin-javascript in production, I faced an undefined error in `lib/utils.js`. Further investigation revealed that the fields were not correctly checked for nulls. This PR aims to add proper null checks to avoid undefined errors on runtime.

The error screenshot:
Screenshot 2023-01-30 at 4 24 22 PM

Gremlin version: v3.6.2
Node version: v14.17.4
NPM version: 7.19.1

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/utils.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/utils.js
@@ -90,10 +90,23 @@ function generateUserAgent() {
   } catch (e) {
     driverVersion = 'NotAvailable';
   }
-  const runtimeVersion = process?.version?.replace(' ', '_');
-  const osName = os?.platform().replace(' ', '_');
-  const osVersion = os?.release().replace(' ', '_');
-  const cpuArch = process?.arch?.replace(' ', '_');
+  let runtimeVersion = osName = osVersion = cpuArch = 'NotAvailable';
+  
+  // process != null is equal to process != null && process != undefined
+  if(process != null) {
+    if(process.version != null) {
+      runtimeVersion = process?.version?.replace(' ', '_');
+    }
+    if(process.arch != null) {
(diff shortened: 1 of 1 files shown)


Cole-Greer:
@iAmmarTahir Thanks for catching this error and submitting this fix. I would like to add that this would be further improved if a default value of 'NotAvailable' was provided as in ``` const cpuArch = process?.arch?.replace(' ', '_') || 'NotAvailable';```

Also I would like to ask if you could target this PR to 3.5-dev or 3.6-dev. Our master branch contains the code which will eventually make up the next major release (3.7.0). To have your changes brought into upcoming minor releases (3.5.6, 3.6.3) the changes will need to be in the corresponding dev branches. You should always target the earliest branch you wish to support as changes will be merged forward (eg. if your PR is merged into 3.5-dev, we will also merge it forward into 3.6-dev and master).

One wrinkle is that the 3.5.x branch is using node v10 which does not support the `?.` operator. Therefore some less elegant null checks will be needed for that branch. If you are willing to submit a node v10 compatible fix to 3.5-dev that would be fantastic. Otherwise I would ask that you add the default value of 'NotAvailable' and target this PR to 3.6-dev and I can look into a fix for the 3.5.x branch.

Cole-Greer:
I also want to note that this PR failed the JS CI tests as the linter is currently configured for ecma 2017 which is roughly equivalent to node v10. I have submitted a [PR](https://github.com/apache/tinkerpop/pull/1959) to bump this to version 2021 which is fully supported by node v16 and will prevent the linter from rejecting the use of `?.`

iAmmarTahir:
@Cole-Greer I have addressed the issues that you mentioned and removed optional chaining to support compatibility for NodeJS v10. I will target this branch towards `3.5-dev`

PR #1960, opened 2023-02-03
patch: Fixing generateUserAgent() to handle null and undefined properly

iAmmarTahir:
This PR aims to handle null and undefined values properly in `generateUserAgent()`

Screenshot 2023-01-30 at 4 24 07 PM

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/utils.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/utils.js
@@ -90,10 +90,24 @@ function generateUserAgent() {
   } catch (e) {
     driverVersion = 'NotAvailable';
   }
-  const runtimeVersion = process.version.replace(' ', '_');
-  const osName = os.platform().replace(' ', '_');
-  const osVersion = os.release().replace(' ', '_');
-  const cpuArch = process.arch.replace(' ', '_');
+  let runtimeVersion;
+  let osName;
+  let osVersion;
+  let cpuArch;
+  runtimeVersion = osName = osVersion = cpuArch = 'NotAvailable';
+  if (process != null) {
+    if (process.version != null) {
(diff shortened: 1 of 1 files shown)


iAmmarTahir:
@Cole-Greer I have created a new PR with your suggested changes and removed the optional chaining `.?` to support older NodeJS versions. Let me know if there any changes to make.

Cole-Greer:
@iAmmarTahir Thanks that looks great! To get this merged now it will need 3 VOTE +1's from tinkerpop committers. Alternatively if it gets a single VOTE +1 and a week passes, it can be merged with "lazy consensus". You can typically expect some committers to review a PR within a couple of days.

I'm not a tinkerpop committer so I cannot give an official vote here but I will give a VOTE +1 (non-binding) as I think this PR should be accepted.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/utils.js
```diff
@@ -90,10 +90,24 @@ function generateUserAgent() {
...
-  const osName = os.platform().replace(' ', '_');
-  const osVersion = os.release().replace(' ', '_');
-  const cpuArch = process.arch.replace(' ', '_');
+  let runtimeVersion;
+  let osName;
+  let osVersion;
+  let cpuArch;
+  runtimeVersion = osName = osVersion = cpuArch = 'NotAvailable';
+  if (process != null) {
+    if (process.version != null) {
```
vkagamlyk, line 99: Can `version` be undefined?
```suggestion
    if (process.version) {
```
same question for `arch`

iAmmarTahir: Yes, it makes sense. If the process is not null or undefined, then it must have a version and arch

vkagamlyk:
@iAmmarTahir Thank you for fixing this issue!
VOTE +1

PR #1961, opened 2023-02-03
TINKERPOP-2864 GraphBinaryMessageSerializer supports DefaultComputerResult

ministat:
Fix the issue of "Serializer for type org.apache.tinkerpop.gremlin.process.computer.util.DefaultComputerResult not found" when serializing the result of "graph.compute(SparkGraphComputer).program(CloneVertexProgram.build().create()).submit().get()"

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/GraphBinaryWriter.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/GraphBinaryWriter.java
@@ -87,29 +87,37 @@ public class GraphBinaryWriter {
         }
 
         final Class<?> objectClass = value.getClass();
-        final TypeSerializer<T> serializer = (TypeSerializer<T>) registry.getSerializer(objectClass);
 
-        if (serializer instanceof CustomTypeSerializer) {
-            // It's a custom type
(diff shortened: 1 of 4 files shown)


spmallette:
GraphBinary is meant to return types that can be returned by the Gremlin language, not any random script. Random scripts should convert their results to one of the primitive types GraphBinary exposes. So your `toString()` of `ComputerResult` is at least one approach you could take there. It's tempting but I don't think we should get in the business of doing a quiet `toString()` by default. I feel like the user should know if a serializer isn't present directly by exception, then they alter their request accordingly.

ministat:
@spmallette All right. That is fine if GraphBinary design is to handle the return by Gremlin language instead of random script. I'll use toString() in my script to make it work.

ministat:
@spmallette The UT in TypeSerializerFailureTests expects to see an exception when serializing DefaultRemoteTraverser which is similar as DefaultComputerGraph. That is why my PR failed on that case. So, the design of GraphBinary does not target to serialize all types.

ministat:
Close this PR since it is not the design of GraphBinaryMessageSerialier.

spmallette:
let's see if anyone else feels differently about this before you close it.

spmallette:
i think folks are inclined to leave this as it is. closing.

PR #1963, opened 2023-02-06
Added Go GLV verification to validate-distribution.sh

Cole-Greer:
It came to my attention that the Go GLV was missing from validate-distribution.sh. Pushing a quick update here to fix that.

I've also added a bit of configuration to make the mac user experience a little smoother. If this is not desirable I will remove it and only add the Go fix.


vkagamlyk:
Thank you, @Cole-Greer 
VOTE +1

xiazcy:
VOTE +1

PR #1968, opened 2023-02-09
[TINKERPOP-2841] Fix Per-Request Settings in Go GLV 3.5

Cole-Greer:
https://issues.apache.org/jira/browse/TINKERPOP-2841

This PR is cherry picking the changes introduced in https://github.com/apache/tinkerpop/pull/1939 back to 3.5.x with the socket server based tests removed for compatibility. There are no changes to the existing external interface (other than adding a new SubmitWithOptions() method). This is not a breaking change for any users.


vkagamlyk:
VOTE +1

PR #1986, opened 2023-03-10
TINKERPOP-2373 Bump to Groovy 4.0.9/GMavenPlus 2.1.0

kenhuuu:
Fixes https://issues.apache.org/jira/browse/TINKERPOP-2373

Groovy 4 packages the indy version by default so it isn't necessary to use the classifier anymore. asm-util is a new dependency for groovysh but clashes with another asm dependency so we exclude it from groovysh since it isn't required when running (Groovy will use the version of asm from jarjarasm). Removed usages of "def class" since this no longer seems to be supported by Groovy. Updated the parameter size for shouldProvideBetterExceptionForMethodCodeTooLarge test as it no longer triggers the exception with the previous, smaller size.

### DIFF
--- a/gremlin-console/src/main/static/NOTICE
+++ b/gremlin-console/src/main/static/NOTICE
@@ -18,15 +18,10 @@ Original source copyright:
 Copyright (c) 2008 Alexander Beider & Stephen P. Morse.
 
 ------------------------------------------------------------------------
-Apache Groovy 2.5.15 (AL ASF)
+Apache Groovy 4.0.9 (AL ASF)
 ------------------------------------------------------------------------
-This product includes/uses ANTLR (http://www.antlr2.org/)
-developed by Terence Parr 1989-2006
-
-This product bundles icons from the famfamfam.com silk icons set
-http://www.famfamfam.com/lab/icons/silk/
-Licensed under the Creative Commons Attribution Licence v2.5
-http://creativecommons.org/licenses/by/2.5/
+This product includes/uses ANTLR4 (https://github.com/antlr/antlr4)
+Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
 
 ------------------------------------------------------------------------
 Apache Ivy 2.5.1 (AL ASF)
--- a/gremlin-console/src/test/java/org/apache/tinkerpop/gremlin/console/jsr223/GephiRemoteAcceptorIntegrateTest.java
+++ b/gremlin-console/src/test/java/org/apache/tinkerpop/gremlin/console/jsr223/GephiRemoteAcceptorIntegrateTest.java
@@ -20,12 +20,12 @@ package org.apache.tinkerpop.gremlin.console.jsr223;
 
 import com.github.tomakehurst.wiremock.junit.WireMockRule;
 import org.apache.commons.io.input.NullInputStream;
+import org.apache.groovy.groovysh.Groovysh;
 import org.apache.tinkerpop.gremlin.console.GremlinGroovysh;
 import org.apache.tinkerpop.gremlin.console.Mediator;
 import org.apache.tinkerpop.gremlin.jsr223.console.RemoteException;
 import org.apache.tinkerpop.gremlin.structure.Graph;
 import org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerFactory;
-import org.codehaus.groovy.tools.shell.Groovysh;
 import org.codehaus.groovy.tools.shell.IO;
 import org.junit.Before;
 import org.junit.Rule;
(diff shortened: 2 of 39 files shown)


kenhuuu:
VOTE +1

gremlin-console/src/main/static/NOTICE
```diff
@@ -18,15 +18,10 @@ Original source copyright:
...
 
 ------------------------------------------------------------------------
-Apache Groovy 2.5.15 (AL ASF)
+Apache Groovy 4.0.9 (AL ASF)
 ------------------------------------------------------------------------
-This product includes/uses ANTLR (http://www.antlr2.org/)
-developed by Terence Parr 1989-2006
-
-This product bundles icons from the famfamfam.com silk icons set
-http://www.famfamfam.com/lab/icons/silk/
```
Cole-Greer, line 27: Was this piece intended to be removed? Are the icons no longer bundled?

gremlin-console/src/test/java/org/apache/tinkerpop/gremlin/console/jsr223/GephiRemoteAcceptorIntegrateTest.java
```diff
@@ -20,12 +20,12 @@
...
 import com.github.tomakehurst.wiremock.junit.WireMockRule;
 import org.apache.commons.io.input.NullInputStream;
+import org.apache.groovy.groovysh.Groovysh;
 import org.apache.tinkerpop.gremlin.console.GremlinGroovysh;
 import org.apache.tinkerpop.gremlin.console.Mediator;
 import org.apache.tinkerpop.gremlin.jsr223.console.RemoteException;
 import org.apache.tinkerpop.gremlin.structure.Graph;
 import org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerFactory;
-import org.codehaus.groovy.tools.shell.Groovysh;
 import org.codehaus.groovy.tools.shell.IO;
```
vkagamlyk, line 29: is this import required?

gremlin-console/src/main/static/NOTICE
spmallette: The NOTICE file for groovy seems to have even more contents now: https://github.com/apache/groovy/blob/GROOVY_4_0_9/NOTICE

not sure what's actually packaged in the groovy jars though.

spmallette:
VOTE +1 - pending console/server LICENSE/NOTICE review. also need to make sure there aren't any new binaries in those zip distributions that came as a result of this upgrade as that might affect LICENSE/NOTICE as well.

gremlin-console/src/main/static/NOTICE
kenhuuu: > Was this piece intended to be removed? Are the icons no longer bundled?

The NOTICE file doesn't exist anymore in the groovy-console 4.0.9 jar, however, it does exist in https://github.com/apache/groovy/blob/GROOVY_4_0_9/subprojects/groovy-console/NOTICE so maybe we should add it back? In any case, groovy-console is only included with gremlin-server and not gremlin-console.

kenhuuu: > The NOTICE file for groovy seems to have even more contents now: https://github.com/apache/groovy/blob/GROOVY_4_0_9/NOTICE
> 
> not sure what's actually packaged in the groovy jars though.

I went through every groovy 4.0.9 jar included in our distribution and uber jars and they don't contain any more information than what I included in this PR. But, as I mentioned above, it is possible that they aren't packaging the NOTICE files in META-INF anymore so it might be better to just add everything from the NOTICE file you linked?

gremlin-console/src/test/java/org/apache/tinkerpop/gremlin/console/jsr223/GephiRemoteAcceptorIntegrateTest.java
kenhuuu: No, this particular class has been renamed. Although, not all of imports that we use have been.

vkagamlyk:
VOTE +1

kenhuuu:
> VOTE +1 - pending console/server LICENSE/NOTICE review. also need to make sure there aren't any new binaries in those zip distributions that came as a result of this upgrade as that might affect LICENSE/NOTICE as well.

Yes, there were two jar added due to the upgrade. I'll need to modify the LICENSE for one of them.

gremlin-console/src/main/static/NOTICE
spmallette: my understanding is that for binary distributions, if the packaged jars do not have the NOTICE included then we do not need it in our NOTICE. if you've gone through all the groovy jars in both server/console and have only included what's there then i think we're good. 

the groovy NOTICE files we're looking at in code may be all source NOTICE-related which has a different set of rules to adhere to. we dont package groovy source so i think we're ok there.

PR #1989, opened 2023-03-13
TINKERPOP-2747 Separate AuthInfo struct vs AuthInfoProvider interface for dynamic credentials

disq:
This creates an `AuthInfoProvider` interface with the methods necessary to read credential/header information, changes the `AuthInfo` signature slightly to satisfy that interface, and updates call sites so that each call site calls the necessary interface method to re-read the credentials.

The interface looks like:
```go
type AuthInfoProvider interface {
	GetHeader() http.Header
	GetBasicAuth() (ok bool, username, password string)
}
```

It also adds a new `NewDynamicAuth(f func() AuthInfoProvider)` helper function that creates a generator based `AuthInfoProvider`, so that each time `GetHeader()` or `GetBasicAuth()` methods are called the given function is called to get a brand new `AuthInfoProvider` (which itself can be `AuthInfo` or something else) and the results are relayed to the caller.

With this, `AuthInfo`-related constructors (`BasicAuthInfo()` and `HeaderAuthInfo()`) are unchanged, but there's now the ability to have any object provide the values (implementing `AuthInfoProvider`) and they are called on-demand for each connection attempt, as opposed to having them as static values.

Note: This is my first contribution, and I couldn't get the local gremlin-server to play nicely to get the tests to run, but I've tried it elsewhere with AWS Neptune auth, like so:

```go
// ... code to set up aws-sdk-v2 credential chain ...
		signer := v4.NewSigner()

		gen := func() gremlingo.AuthInfoProvider {
			if err := signer.SignHTTP(ctx, cr, req, emptyStringSHA256, "neptune-db", "us-east-1", time.Now()); err != nil {
				panic(err) // not ideal, but it's always nil
			}
			return gremlingo.HeaderAuthInfo(req.Header)
		}
		return gremlingo.NewDynamicAuth(gen), nil
```

### DIFF
--- a/gremlin-go/driver/authInfo.go
+++ b/gremlin-go/driver/authInfo.go
@@ -69,13 +69,13 @@ func HeaderAuthInfo(header http.Header) *AuthInfo {
 
 // DynamicAuth is an AuthInfoProvider that allows dynamic credential generation.
 type DynamicAuth struct {
-	fn func() *AuthInfo
+	fn func() AuthInfoProvider
 }
 
 var _ AuthInfoProvider = (*DynamicAuth)(nil)
 
 // NewDynamicAuth provides a way to generate dynamic credentials with the specified generator function.
-func NewDynamicAuth(f func() *AuthInfo) *DynamicAuth {
+func NewDynamicAuth(f func() AuthInfoProvider) *DynamicAuth {
 	return &DynamicAuth{fn: f}
 }
 
--- a/gremlin-go/driver/authInfo.go
+++ b/gremlin-go/driver/authInfo.go
@@ -66,3 +66,25 @@ func BasicAuthInfo(username string, password string) *AuthInfo {
 func HeaderAuthInfo(header http.Header) *AuthInfo {
 	return &AuthInfo{Header: header}
 }
+
+// DynamicAuth is an AuthInfoProvider that allows dynamic credential generation.
+type DynamicAuth struct {
+	fn func() *AuthInfo
+}
+
+var _ AuthInfoProvider = (*DynamicAuth)(nil)
+
+// NewDynamicAuth provides a way to generate dynamic credentials with the specified generator function.
+func NewDynamicAuth(f func() *AuthInfo) *DynamicAuth {
+	return &DynamicAuth{fn: f}
+}
+
+// GetHeader calls the stored function to get the header dynamically.
+func (d *DynamicAuth) GetHeader() http.Header {
+	return d.fn().GetHeader()
+}
+
+// GetHeader calls the stored function to get basic authentication dynamically.
+func (d *DynamicAuth) GetBasicAuth() (bool, string, string) {
+	return d.fn().GetBasicAuth()
+}
--- a/gremlin-go/driver/authInfo.go
+++ b/gremlin-go/driver/authInfo.go
@@ -72,7 +72,12 @@ type DynamicAuth struct {
 	fn func() AuthInfoProvider
 }
 
-var _ AuthInfoProvider = (*DynamicAuth)(nil)
+var (
+	_ AuthInfoProvider = (*DynamicAuth)(nil)
+
+	// NoopAuthInfo is a no-op AuthInfoProvider that can be used to disable authentication.
+	NoopAuthInfo = NewDynamicAuth(func() AuthInfoProvider { return &AuthInfo{} })
+)
 
 // NewDynamicAuth provides a way to generate dynamic credentials with the specified generator function.
 func NewDynamicAuth(f func() AuthInfoProvider) *DynamicAuth {
--- a/gremlin-go/driver/authInfo.go
+++ b/gremlin-go/driver/authInfo.go
@@ -21,8 +21,14 @@ package gremlingo
 
 import "net/http"
 
-// AuthInfo is an option struct that allows authentication information to be specified.
-// Authentication can be provided via http.Header Header directly.
+// AuthInfoProvider is an interface that allows authentication information to be specified.
+type AuthInfoProvider interface {
+	GetHeader() http.Header
+	GetBasicAuth() (ok bool, username, password string)
+}
+
+// AuthInfo is an option struct that allows authentication information to be specified statically.
+// Authentication can be provided via http.Header directly.
 // Basic authentication can also be used via the BasicAuthInfo function.
 type AuthInfo struct {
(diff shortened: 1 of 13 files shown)


gremlin-go/driver/authInfo.go
```diff
@@ -21,29 +21,40 @@ package gremlingo
...
+// Authentication can be provided via http.Header directly.
 // Basic authentication can also be used via the BasicAuthInfo function.
 type AuthInfo struct {
 	Header   http.Header
 	Username string
 	Password string
 }
 
-// getHeader provides a safe way to get a header from the AuthInfo even if it is nil.
+var _ AuthInfoProvider = (*AuthInfo)(nil)
```
disq, line 39: This is so that we make sure `*AuthInfo` satisfies the `AuthInfoProvider` interface (compile-time check). More below with `*DynamicAuth`.

gremlin-go/driver/client.go
```diff
@@ -21,9 +21,10 @@ package gremlingo
 
 import (
 	"crypto/tls"
-	"golang.org/x/text/language"
 	"runtime"
 	"time"
+
+	"golang.org/x/text/language"
```
disq, line 27: The imports were out of order so they were sorted by `goimports` automatically.

gremlin-go/driver/client_test.go
```diff
@@ -40,13 +41,13 @@ func TestClient(t *testing.T) {
 				settings.AuthInfo = testNoAuthAuthInfo
 			})
 		defer client.Close()
-		assert.Nil(t, err)
+		assert.NoError(t, err)
```
disq, line 44: Errcheck assertion is traditionally done with `assert.NoError`. There are couple more files with this slight inconvenience which I chose not to touch for now.

Cole-Greer:
Thanks @disq for submitting this PR. I saw you were having some trouble running the tests locally. The basic requirements for running the tests are JDK 8 or 11 (17 is currently unsupported), and docker needs to be running. You also will need to create an empty file with the path `tinkerpop/gremlin-go/.glv` to activate the maven profile for the go GLV. Then you should be able to run all of the tests by running `mvn clean install`. The relevant docs for this are [here](https://tinkerpop.apache.org/docs/current/dev/developer/#development-environment) if you haven't found them yet. These tests also run in our github actions on your PR if you don't want to setup the local environment just yet.

There appears to be one test failing in the checks right now. The maven logs can be a bit much to search through, I like to search for `❌` in the logs to find the failing test quickly. It appears that `gremlin-go/driver/gorillaTransporter.go:68` can sometimes result in a nil pointer dereference.

I'll spend some time tomorrow taking a closer look and seeing if I can help out at all with the tests. If you do try to get the tests running locally, feel free to reach out here or on discord for more specific help with any issues.

gremlin-go/driver/connection.go
```diff
@@ -42,15 +42,15 @@ type connection struct {
...
-	authInfo         			*AuthInfo
-	tlsConfig         			*tls.Config
-	keepAliveInterval 			time.Duration
-	writeDeadline     			time.Duration
-	connectionTimeout 			time.Duration
-	enableCompression 			bool
-	readBufferSize				int
-	writeBufferSize				int
-	enableUserAgentOnConnect		bool
+	authInfo                 AuthInfoProvider
```
Cole-Greer, line 45: I do have some concerns with potentially breaking things for some users by changing this type from `*AuthInfo` to `AuthInfoProvider`. As the fields in AuthInfo were all public, there are likely users who are modifying these fields directly instead of using the interface.

It might not be a big deal as the current expected use case is users construct the AuthInfo struct, pass it into a new client, then never touch it again. I don't believe the AuthInfo is publicly accessible anywhere after it has been passed into the client.

I want to think about this a little bit more but I believe it is ok and won't impact any users.

gremlin-go/driver/authInfo.go
```diff
@@ -55,3 +66,30 @@ func BasicAuthInfo(username string, password string) *AuthInfo {
...
+// DynamicAuth is an AuthInfoProvider that allows dynamic credential generation.
+type DynamicAuth struct {
+	fn func() AuthInfoProvider
+}
+
+var (
+	_ AuthInfoProvider = (*DynamicAuth)(nil)
+
+	// NoopAuthInfo is a no-op AuthInfoProvider that can be used to disable authentication.
+	NoopAuthInfo = NewDynamicAuth(func() AuthInfoProvider { return &AuthInfo{} })
```
disq, line 79: This is implemented as dynamic auth so that the returning value can't be changed, e.g. by setting the public fields of the AuthInfo struct after use.

gremlin-go/driver/connection.go
disq: > There appears to be one test failing in the checks right now. The maven logs can be a bit much to search through, I like to search for `❌` in the logs to find the failing test quickly. It appears that `gremlin-go/driver/gorillaTransporter.go:68` can sometimes result in a nil pointer dereference.

Thank you for the clarification on how to run the tests locally.

The issue seems to be one of the tests setting `authInfo` to `nil`, which is unexpected (and isn't done anywhere else in the codebase as far as I can tell) so now `gorillaTransporter.getAuthInfo()` has a nil-check and returns `NoopAuthInfo` if `nil`. `NoopAuthInfo` is an idempotent way of specifying a zero-value for the `AuthInfoProvider` and can be used publicly as well.

disq: > I don't believe the AuthInfo is publicly accessible anywhere after it has been passed into the client.

Since `DriverRemoteConnection.settings` itself is unexported (as per Go export rules, it would have to be named `Settings`) I don't see a way of how one would go about changing any setting after initialization.

vkagamlyk:
LGTM
VOTE +1

Cole-Greer:
VOTE +1 (Non-binding). Thanks for fixing those issues, looks good to me. One small ask is if you could add a short description to the 3.5.6 section of the [CHANGELOG](https://github.com/apache/tinkerpop/blob/3.5-dev/CHANGELOG.asciidoc). Just a short bullet point on it should do. Once the changelog is updated and the additional votes are in we should be good to merge this.

xiazcy:
LGTM, VOTE +1.

kenhuuu:
VOTE +1. Thanks for contributing this.

PR #2000, opened 2023-03-22
Updates Java UserAgent Error Handling

Cole-Greer:
A concern was raised over the possibility of an Error being thrown during the initialization of the UserAgent class resulting in the class to fail to load and giving users an obscure NoClassDefFoundError. I'm upgrading the existing catch(Exception)'s to catch(Throwable)'s to catch these potential Error's and ensure that the user agent class will always load correctly. Any missing data due to caught errors will be replaced with a default value of "NotAvailable"

Missing data in the user agent on rare occasions is a small concern and preferable to the driver crashing altogether.


vkagamlyk:
VOTE +1

kenhuuu:
VOTE +1

xiazcy:
VOTE +1

PR #2004, opened 2023-03-28
Adds Missing .glv file warnings to maven

Cole-Greer:
Creates new profiles in the glv poms which will simply echo out a warning message if a profile is skipped due to a missing .glv file.

### DIFF
--- a/gremlin-dotnet/src/pom.xml
+++ b/gremlin-dotnet/src/pom.xml
@@ -186,6 +186,43 @@ limitations under the License.
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>gremlin-dotnet-src-skipped-warning</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+                <file>
+                    <missing>.glv</missing>
+                </file>
+            </activation>
+            <properties>
+                <packaging.type>pom</packaging.type>
+            </properties>
+            <build>
+                <plugins>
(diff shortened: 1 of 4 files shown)


vkagamlyk:
VOTE+1

xiazcy:
I'm slightly on the fence about this one. It'll be great to have warnings which tells you that you aren't building the GLVs, but I'm not sure if it's necessarily worth adding an entire profile for each of the GLVs for this, given that there is documentation about them, and also we are considering removing the need to use `.glv` in 3.7.0 (as they were initially added due to GLV tests required environment set-up, but we are now using Docker which should be able to remove that need).
Would there be an easier way to add the warning in an existing profile?

Cole-Greer:
> I'm slightly on the fence about this one. It'll be great to have warnings which tells you that you aren't building the GLVs, but I'm not sure if it's necessarily worth adding an entire profile for each of the GLVs for this, given that there is documentation about them, and also we are considering removing the need to use `.glv` in 3.7.0 (as they were initially added due to GLV tests required environment set-up, but we are now using Docker which should be able to remove that need). Would there be an easier way to add the warning in an existing profile?

This was the best method for adding some sort of warning that I could find. It is possible there are better solutions but I don't think it's worth investing much time in. I think this warning would be helpful for current developers, we can remove these profiles once the `.glv` system is torn out.

PR #2007, opened 2023-03-29
TINKERPOP-2852 Deploying multi-arch docker images

Cole-Greer:
Replaces spotify plugin with io.fabric8 docker-maven-plugin. The new plugin has the benefit of supporting multi-arch images through buildx for deployments and it is an actively maintained plugin as opposed to the spotify plugin.

The new plugin continues to fetch auth credentials from the same location so no changes to the release environment will be necessary.

I have tested the full set of .Net gherkin and integration tests against the new server image with no issues. I have also used [container-diff](https://github.com/GoogleContainerTools/container-diff) to compare the new image to the one that we currently build using the mvn exec plugin and docker compose. These are the results:

```
-----File-----

These entries have been added to colegreer/tinkerpop:server-3.5.6-SNAPSHOT: None

These entries have been deleted from colegreer/tinkerpop:server-3.5.6-SNAPSHOT: None

These entries have been changed between colegreer/tinkerpop:server-3.5.6-SNAPSHOT and tinkerpop/gremlin-server:3.5.6-SNAPSHOT:
FILE                                                             SIZE1         SIZE2
/opt/gremlin-server/lib/gremlin-server-3.5.6-SNAPSHOT.jar        243K          242.8K
/etc/ssl/certs/java/cacerts                                      155.3K        155.3K


-----Apt-----

Packages found only in colegreer/tinkerpop:server-3.5.6-SNAPSHOT: None

Packages found only in tinkerpop/gremlin-server:3.5.6-SNAPSHOT: None

Version differences: None
```
I haven't yet identified the minute difference in the jar file between the two images. I can investigate this further if anyone views it as a potential concern.

Unfortunately the ARM64 console image has been unstable in my tests and I've opted to exclude it for now. The configuration for multi-arch console images is still included in this PR so an ARM console image can be added to the deployment in the future with a one liner update.

Additional architectures such as `ppc64le` could be added to the deployment simply by listing them in the `docker.platforms` property in the console and server poms. I would expect all architectures supported by the [alpine](https://hub.docker.com/_/alpine) base image should work although I lack the resources to test this.

### DIFF
--- a/docs/src/upgrade/release-3.5.x.asciidoc
+++ b/docs/src/upgrade/release-3.5.x.asciidoc
@@ -30,24 +30,10 @@ complete list of all the modifications that are part of this release.
 
 === Upgrading for Users
 
-=== Upgrading for Providers
+==== ARM64 Support for Gremlin Server Docker Image
 
-==== Local steps should handle non-Iterables
-
-`index` steps and local steps `count`, `dedup`, `max`, `mean`, `min`, `order`, `range`, `sample`, `sum`, `tail` should
-work correctly with not only `Iterable` input, but also with arrays and single values.
-
-Examples of queries that should be supported:
-
-[source,groovy]
-----
-g.inject(1).max(local)
-----
-
-[source,java]
-----
-g.inject(new Integer[] {1,2},3).max(Scope.local).toList()
-----
+Gremlin server docker image now supports both AMD64 and ARM64. Multi-architecture image can now be found at
+https://hub.docker.com/r/tinkerpop/gremlin-server
 
 === Upgrading for Providers
 
@@ -66,6 +52,23 @@ expensive step to propagate that data throughout the traversal in between each s
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-2855[TINKERPOP-2855],
 link:https://issues.apache.org/jira/browse/TINKERPOP-2888[TINKERPOP-2888]
+
+===== Local steps should handle non-Iterables
+
+`index` steps and local steps `count`, `dedup`, `max`, `mean`, `min`, `order`, `range`, `sample`, `sum`, `tail` should
+work correctly with not only `Iterable` input, but also with arrays and single values.
+
+Examples of queries that should be supported:
+
+[source,groovy]
+----
+g.inject(1).max(local)
+----
+
+[source,java]
+----
+g.inject(new Integer[] {1,2},3).max(Scope.local).toList()
+----
 
 == TinkerPop 3.5.5
 
--- a/gremlin-server/pom.xml
+++ b/gremlin-server/pom.xml
@@ -32,6 +32,7 @@ limitations under the License.
         -->
         <maven.exec.skip>false</maven.exec.skip> <!-- default -->
         <skipImageBuild>${maven.exec.skip}</skipImageBuild>
+        <docker.platforms>linux/amd64,linux/arm64/v8</docker.platforms>
     </properties>
     <dependencies>
         <dependency>
@@ -207,6 +208,38 @@ limitations under the License.
                 </executions>
             </plugin>
         </plugins>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>io.fabric8</groupId>
+                    <artifactId>docker-maven-plugin</artifactId>
+                    <configuration>
+                        <images>
+                            <image>
+                                <name>tinkerpop/gremlin-server:${project.version}</name>
+                                <build>
+                                    <dockerFile>${project.build.directory}/../Dockerfile</dockerFile>
+                                    <args>
+                                        <GREMLIN_SERVER_DIR>target/apache-tinkerpop-${project.artifactId}-${project.version}-standalone</GREMLIN_SERVER_DIR>
+                                    </args>
+                                    <tags>
+                                        <!-- docker.tags.* properties are empty if project.version is a pre-release
+                                            resulting in no additional tags -->
+                                        <tag>${docker.tags.majorVersion.minorVersion}</tag>
+                                        <tag>${docker.tags.latest}</tag>
+                                    </tags>
+                                    <buildx>
+                                        <platforms>
+                                            <platform>${docker.platforms}</platform>
+                                        </platforms>
+                                    </buildx>
+                                </build>
+                            </image>
+                        </images>
+                    </configuration>
+                </plugin>
+            </plugins>
+        </pluginManagement>
     </build>
 
     <profiles>
@@ -233,7 +266,6 @@ limitations under the License.
                 </plugins>
             </build>
         </profile>
-
 
         <profile>
             <!--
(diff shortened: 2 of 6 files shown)


pom.xml
```diff
@@ -345,6 +345,55 @@ limitations under the License.
...
+                    <execution>
+                        <!-- sets the docker.tags.latest property to "latest" if a prerelease version was used,
+                            empty string otherwise -->
+                        <id>build-helper-regex-set-docker-latest-tag</id>
+                        <phase>validate</phase>
+                        <goals>
+                            <goal>regex-property</goal>
+                        </goals>
+                        <configuration>
+                            <name>docker.tags.latest</name>
```
FlorianHockmann, line 390: I guess we have to remove this part for `3.5-dev` and only add it for `3.6-dev` as `latest` should always point to the latest `3.6` image, but not a `3.5` one.

pom.xml
```diff
@@ -345,6 +345,55 @@ limitations under the License.
...
+                        <configuration>
+                            <name>true.when.is.prerelease.version</name>
+                            <value>${only.when.is.prerelease.version}</value>
+                            <regex>${project.version}</regex>
+                            <replacement></replacement>
+                            <failIfNoMatch>false</failIfNoMatch>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <!-- sets the docker.tags.majorVersion.minorVersion property to
```
FlorianHockmann, line 365: Isn't this exactly the other way around? The property should only bet set to `major.minor` if it is **not** a prerelease version. (same below for `latest`)

FlorianHockmann (commented):
Great contribution in general! However, I'm not sure if the tagging is working correctly. I just tried it locally by setting the version to `3.5.7` and then generated the images via:

```sh
mvn versions:set -DnewVersion=3.5.7 -DgenerateBackupPoms=false
mvn clean install -DskipTests
mvn clean install -pl :gremlin-server,:gremlin-console -DdockerImages -DskipTests
```

This is resulted in the following images:

```sh
$ docker image ls
REPOSITORY                  TAG               IMAGE ID       CREATED          SIZE
tinkerpop/gremlin-console   3.5               81ff2adc9827   20 seconds ago   344MB
tinkerpop/gremlin-console   3.5.7             81ff2adc9827   20 seconds ago   344MB
tinkerpop/gremlin-console   latest            81ff2adc9827   20 seconds ago   344MB
tinkerpop/gremlin-server    3.5.7             535bcd720bc6   54 seconds ago   381MB
```

Why is there a difference in the tags between console and server? The tags for the console image look correct to me if `latest` is removed, but the server image misses the tag: `3.5`.

pom.xml
Cole-Greer: This is true but to the best of my knowledge not a new issue. I haven't been able to run the old plugin in my environment so I can't verify this with 100% certainty but `latest` is the default tag for the old plugin and my understanding is that a new `latest` tag gets pushed with every deploy. I believe that currently we are reliant on the fact that typically 3.5.x gets deployed first, and is immediately followed by 3.6.x. I'm not aware of any way to resolve this automatically as there is no awareness within maven if the current deployment is for the "old branch" (3.5) and should not have a latest tag, or if it's for the "new branch" (3.6) and needs a latest tag.

I implemented it this way in order to best replicate the old behaviour. Perhaps it would be better to always have the latest tag disabled unless the release manager sets something like `-Ddocker.latest` during the deployment.

pom.xml
Cole-Greer: Thanks for catching that. I have fixed the comments.

pom.xml
FlorianHockmann: Yes, there is no way to handle this automatically. That is why we only have the logic to tag `latest` present on the `3.6-dev` branch and on `master` right now, but not on the `3.5-dev` branch. That way, we can't accidentally tag a Docker image built on `3.5-dev` as `latest`.

Cole-Greer: Ahh I wasn't paying close attention to the differences between branches and missed that. I will comment out the `latest` tag piece for `3.5-dev`, and it can be un-commented when it gets merged up.

Alternatively I can remove it entirely here and submit an additional PR for 3.6-dev.

FlorianHockmann: I think commenting out makes sense as it clearly documents this on all branches :-)

FlorianHockmann (commented):
I just tried again to build the docker images with these changes after changing the versions manually to `3.5.7`, but this resulted in an error for the Console image:

```
[ERROR] Failed to execute goal io.fabric8:docker-maven-plugin:0.42.0:build (docker:build) on project gremlin-console: Execution docker:build of goal io.fabric8:docker-maven-plugin:0.42.0:build failed: Given Docker name 'tinkerpop/gremlin-console:${parsedVersion.majorVersion}.${parsedVersion.minorVersion}' is invalid:
[ERROR]    * tag part '${parsedVersion.majorVersion}.${parsedVersion.minorVersion}' doesn't match allowed pattern '^[\w][\w.-]{0,127}$'
[ERROR] See http://bit.ly/docker_image_fmt for more details
```

It worked however for the server image for some reason:

```
$ docker image ls
REPOSITORY                                                       TAG               IMAGE ID       CREATED          SIZE
tinkerpop/gremlin-server                                         3.5.7             c1fd1b9fba2d   3 minutes ago    381MB
```

I already tried to find a difference in the config for the server and the console image, but couldn't find a possible cause for this.
Is it working for you when you change the version to a non-prerelease one?

Cole-Greer:
> Is it working for you when you change the version to a non-prerelease one?

@FlorianHockmann That is strange. The tags were working as expected in my testing before submitting the PR. More investigation and testing is needed, I have converted this back to a draft PR for the meantime.

Cole-Greer:
@FlorianHockmann Looks like I accidentally removed too much when I attempted to remove the `latest` tag from 3.5 which was the cause of that error you observed. I have fixed that and run several tests deploying images with different versions to a personal dockerhub repo. You can see all of the tags here (https://hub.docker.com/repository/docker/colegreer/gremlin-console/tags?page=1&ordering=last_updated, https://hub.docker.com/repository/docker/colegreer/gremlin-server/tags?page=1&ordering=last_updated).

I tested deployments of `3.5.6-SNAPSHOT`, `3.5.7`, `3.6.3-SNAPSHOT`, and `3.6.4` to ensure we are getting the correct tags for all of them.

When running these tests I temporarily moved the `docker:push` execution from the `deploy` phase to the `install` phase as I wanted to test docker deployment without triggering any of the other deploy phase goals. It's also noteworthy that the new plugin does not attempt to build the full multi-arch image during the build goal. The reasoning for this being that multi-arch images are much larger in size and slow to build. The full multi-arch image is not built until the `push` goal is executed. This isn't ideal for us but there doesn't appear to be a good workaround with this plugin. Ultimately I don't think it is a big issue though, as we only need the multi-arch images during deployment.

pom.xml
```diff
@@ -345,11 +345,62 @@ limitations under the License.
...
+                            <replacement></replacement>
+                            <failIfNoMatch>false</failIfNoMatch>
+                        </configuration>
+                    </execution> -->
                     <execution>
                         <id>parse-version</id>
                         <goals>
                             <goal>parse-version</goal>
-                        </goals>
+                            </goals>
```
FlorianHockmann, line 403: (nitpick) Wrong indentation

FlorianHockmann (approved):
OK, that was really a small error then. The tags definitely look good now!

I also agree with your assessment regarding when the multi-arch image is built. I don't think that we should try to change this now and would just start with how the plugin behaves out of the box. If we later find a good reason why we would need the multi-arch image already before deployment, then we can still look into how to get that done.

VOTE +1

edit: Could you please add an entry to the CHANGELOG and maybe also a small note to the upgrade docs to make users aware of this improvement?

gremlin-server/pom.xml
```diff
@@ -32,6 +32,8 @@ limitations under the License.
         -->
         <maven.exec.skip>false</maven.exec.skip> <!-- default -->
         <skipImageBuild>${maven.exec.skip}</skipImageBuild>
+        <docker.platforms>linux/amd64,linux/arm64/v8</docker.platforms>
+
```
lyndonbauto, line 36: nit:
```suggestion
        <docker.platforms>linux/amd64,linux/arm64/v8</docker.platforms>
```

gremlin-server/pom.xml
```diff
@@ -207,6 +209,39 @@ limitations under the License.
                 </executions>
             </plugin>
         </plugins>
+
+        <pluginManagement>
```
lyndonbauto, line 213:
```suggestion
        <pluginManagement>
```

gremlin-server/pom.xml
```diff
@@ -378,60 +412,28 @@ limitations under the License.
...
                             </execution>
                         </executions>
-                        <configuration>
-                            <repository>tinkerpop/gremlin-server</repository>
-                        </configuration>
                     </plugin>
                 </plugins>
             </build>
         </profile>
+
```
lyndonbauto, line 437:
```suggestion
```

lyndonbauto:
Made a couple nits, feel free to commit or ignore them.

So does this now support building on x64 and aarch64? 

Also does it support cross platform docker image building or just building for the given machines architecture?

Cole-Greer:
@lyndonbauto 

> So does this now support building on x64 and aarch64?

As of TinkerPop 3.5.5/3.6.3 both x64 and aarch64 were able to build and run TinkerPop out of the box.

> Also does it support cross platform docker image building or just building for the given machines architecture?

The new docker plugin introduced here uses docker buildx under the hood. In theory a user can run this from any platform with docker installed and build an image for any other platform they want. My expectation is that we could build images for all architectures supported by the [alpine base image](https://hub.docker.com/_/alpine/tags) simply by adding them to the `<docker.platforms>` list. The only reason the list is currently limited to x86 and arm is that is all I am able to test. Also in my testing the ARM console image was unstable on my ARM Mac so I opted not to include it in the deployment here. I'm currently unsure what's causing this instability as the console runs fine on ARM without docker.

lyndonbauto:
@Cole-Greer Sweet, that's really cool mean.

VOTE+1

docs/src/upgrade/release-3.5.x.asciidoc
```diff
@@ -30,24 +30,8 @@ complete list of all the modifications that are part of this release.
...
-[source,groovy]
-----
-g.inject(1).max(local)
-----
-
-[source,java]
-----
-g.inject(new Integer[] {1,2},3).max(Scope.local).toList()
-----
+Gremlin server docker image now supports both AMD64 and ARM64. Multi-architecture image can now be found at
```
FlorianHockmann, line 33: Could you please add a heading for this?

xiazcy:
VOTE +1

PR #2009, opened 2023-03-31
TINKERPOP-2919 Improve performance of FilterRankingStrategy

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2919

Shortcut some of the recursion in the strategy with a cache that basically captures `TraversalParent` information formerly calculated and recalculated. Reduced execution time of the sample traversal I had from about 1 minute 48 seconds to just a few milliseconds.

VOTE +1


vkagamlyk:
As far as I understand everything works as before, but without multiple recalculations.
The only improvement I see is caching of `stepRanking`. It used in `apply`, but probably `getStepRank` and `getMaxStepRank` also can benefit from caching. Another possible approach to add 'getRank()' to the base Step interface and remove related code from `FilterRankingStrategy`.

LGTM, VOTE +1

PR #2010, opened 2023-04-01
fixed broken link in the-traversal.asciidoc

Derposoft:
When I clicked the old broken link, it just reloaded the document which confused me. Removing the hyphen here should point it to the right place.


spmallette:
thanks for the fix - cherry-picked to 3.5-dev on 6ebb499ad26c24d0d270d3881ca595f54b13e6b1

PR #2012, opened 2023-04-05
[TINKERPOP-2890] Avoid exceptions on local scope based steps where possible

vkagamlyk:
Several local steps extended to work with input other than `Iterable`.
`g.inject(1).dedup(local)` is now valid query.

Affected steps: count, dedup, max, mean, min, order, range, sample, sum, tail.


spmallette:
Could you please add some feature tests for each of the steps that were changed so that this functionality is validated?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/TailLocalStep.java
```diff
@@ -54,7 +54,7 @@ protected S map(final Traverser.Admin<S> traverser) {
                         start instanceof Collection ? ((Collection) start).size() :
                                 start instanceof Path ? ((Path) start).size() :
                                         start instanceof Iterable ? IteratorUtils.count((Iterable) start) :
-                                                this.limit;
+                                                IteratorUtils.count(IteratorUtils.asIterator(start));
```
spmallette, line 57: is this meant as some sort of safety? `RangeLocalStep.applyRange()` seems to satisfy the issue, no? no exception thrown here:

```
gremlin> g.inject([1,2,3],[1],10).tail(local)
==>3
==>1
==>10
```

vkagamlyk: also to handle arrays as input
```
gremlin> g.inject([1,2,3] as int[], [1]).tail(local,1)
==>[1, 2, 3]
==>1
```

gremlin-test/features/filter/Dedup.feature
```diff
@@ -306,4 +306,15 @@ Feature: Step - dedup()
       | ripple |
       | peter |
       | marko |
-      | josh  |
\ No newline at end of file
+      | josh  |
+
+    Scenario: g_injectX1X_dedupXlocalX_unfold
```
spmallette, line 311: this was the contrived example from the JIRA, but i think it comes with the downside that you had to add OptOut for them in computer tests. could you come up with tests that don't require that OptOut please?

vkagamlyk:
VOTE+1

kenhuuu:
VOTE +1

xiazcy:
VOTE +1

PR #2014, opened 2023-04-07
TINKERPOP-2922 Fixed bugs in translator and grammar around Map

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2922

The grammar wasn't parsing groovy syntax for Map keys that hand parens that weren't enum related. Also didn't handle complex objects as keys. The translator was modified to produce slightly more idiomatic Map keys for string values.

VOTE +1


vkagamlyk:
VOTE+1

PR #2017, opened 2023-04-11
[TINKERPOP-2901] Incorrect result caused by has(key, predicate)

Cole-Greer:
https://issues.apache.org/jira/browse/TINKERPOP-2901

One of the optimizations done by TinkerGraphStepStrategy is push down `has()` steps into HasContainer's within TinkerGraphStep. When this occurs, the normal binary reduction of ternary boolean states gets skipped. This PR addresses this by replicating the binary reduction logic within TinkerGraphStep when the filtering is performed.

This PR addresses cases where `GremlinTypeErrorException` is able to leak to the user.

Fixes 2 cases:
1: If has step gets merged into a TinkerGraphStep by TinkerGraphStepStrategy, the evaluation
of the has containers will still be guarded by the same ternary boolean binary reduction logic
as in FilterStep class.

2: If the parent step is not a FilterStep and thus is not equipped to handle a GremlinTypeErrorException,
the error will be reduced to false instead of propagating.

### DIFF
--- a/tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/process/traversal/step/sideEffect/TinkerGraphStep.java
+++ b/tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/process/traversal/step/sideEffect/TinkerGraphStep.java
@@ -19,8 +19,11 @@
 package org.apache.tinkerpop.gremlin.tinkergraph.process.traversal.step.sideEffect;
 
 import org.apache.tinkerpop.gremlin.process.traversal.Compare;
+import org.apache.tinkerpop.gremlin.process.traversal.GremlinTypeErrorException;
 import org.apache.tinkerpop.gremlin.process.traversal.P;
 import org.apache.tinkerpop.gremlin.process.traversal.step.HasContainerHolder;
+import org.apache.tinkerpop.gremlin.process.traversal.step.filter.BinaryReductionStep;
+import org.apache.tinkerpop.gremlin.process.traversal.step.filter.FilterStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.map.GraphStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.util.HasContainer;
 import org.apache.tinkerpop.gremlin.process.traversal.util.AndP;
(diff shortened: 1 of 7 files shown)


mikepersonick:
Looks good.

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/process/traversal/step/sideEffect/TinkerGraphStep.java
```diff
@@ -132,8 +133,20 @@ private <E extends Element> Iterator<E> iteratorList(final Iterator<E> iterator)
...
+                        list.add(e);
+                } catch (GremlinTypeErrorException ex) {
+                    if (getTraversal().isRoot()) {
+                        /*
+                         * If we are at the top level of the query, we do a binary reduction from
+                         * ERROR -> FALSE and filter the solution quietly.
+                         */
+                    } else {
+                        // not a ternary -> binary reducer, pass the ERROR on
+                        throw ex;
```
mikepersonick, line 147: What happens if the GremlinTypeErrorException happens in a non-root traversal like inside a Repeat? There I think we should reduce. It's only when the parent traversal is a value expression (like Not) that we should not reduce.

Cole-Greer: I agree with that. I will put together an update which also checks the parent step of non-root traversals and triggers a reduction if the parent is not a value expression.

Cole-Greer: I've pushed some changes which modify the error propagation rules such that the error will propagate if and only if the parent step is a FilterStep. I haven't completely convinced myself that this is the right rule yet. I believe more tests are still needed.

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/process/traversal/step/sideEffect/TinkerGraphStep.java
```diff
@@ -132,8 +135,20 @@ private <E extends Element> Iterator<E> iteratorList(final Iterator<E> iterator)
         try {
             while (iterator.hasNext()) {
                 final E e = iterator.next();
-                if (HasContainer.testAll(e, this.hasContainers))
-                    list.add(e);
+                try {
+                    if (HasContainer.testAll(e, this.hasContainers))
+                        list.add(e);
+                } catch (GremlinTypeErrorException ex) {
+                    if (getTraversal().isRoot() || !(getTraversal().getParent().asStep() instanceof FilterStep)) {
```
vkagamlyk, line 142: probably better to do not catch/throw exception for non-root and FilterStep.
code might look like
```
if (getTraversal().isRoot() || !(getTraversal().getParent().asStep() instanceof FilterStep)) {
  try { doWork(); } catch (GremlinTypeErrorException ex) {}
} else { 
  doWork();
}
```

Cole-Greer: Thanks for the suggestion, I have refactored the code accordingly.

vkagamlyk:
VOTE+1

xiazcy:
I think where to reduce the error may need a deeper thought/discussion. I'm good with this change only fixing the error leak and explain how current reduction works in documentation.
VOTE +1 pending doc updates.

mikepersonick:
+1

PR #2018, opened 2023-04-12
TINKERPOP-2899 Changed hashCode implementation for LP/P traversers.

kenhuuu:
Fixes https://issues.apache.org/jira/browse/TINKERPOP-2899

In some instances, a graph may contain elements whose Ids are sequentially increasing integers. Because the hashCode for Path implementations are calculated based on these same Ids, you run into a situation where both parts of the hashCode calculation are related by some constant. H(A)-H(B)=C where C is a constant. In this case, XOR is a bad candidate for creating the final hashCode because XOR will reduce to some small pool of potential output. This occurs because the most significant bits will be the same for H(A) and H(B) which will cause them to become 0 when XOR'd. This commit changes the algorithm to be a more standard algorithm for this case which is m * H(A) + H(B).

Some quick performance numbers using Kelvin's air-routes data loaded into TinkerGraph via io(graphml()).readGraph()

| Query | Current Implementation time in ms | Proposed Implementation time in ms |
| --- | --- | --- |
| g.E().hasLabel('route').sample(1000).inV().path() | 55403 | 1243 |
| g.E().as('a').hasLabel('route').sample(1000).as('b').dedup('b').inV() | 53653 | 1134 |
| g.withBulk(false).E().as('a').hasLabel('route').sample(1000).as('b').dedup('b').inV() | 53924 | 1357 |
| g.withBulk(false).E().hasLabel('route').sample(1000).inV().path() | 51906 | 1041 |

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -40,6 +40,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Bumped to `snakeyaml` 2.0 to fix security vulnerability.
 * Bumped to Apache `commons-configuration` 2.9.0 to fix security vulnerability.
 * Improved `count` step optimization for negative values in input for 'eq' comparison.
+* Fixed performance issue when using SampleGlobalStep with a traverser that has either a LABELED_PATH or PATH requirement.
 
 [[release-3-5-5]]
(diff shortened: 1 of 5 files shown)


vkagamlyk:
VOTE+

spmallette:
VOTE +1

xiazcy:
VOTE +1

PR #2019, opened 2023-04-12
TINKERPOP-2863 Update hasId step mid-traversal processing

xiazcy:
https://issues.apache.org/jira/browse/TINKERPOP-2863

Aligned mid-traversal hasId() processing of lists with TinkerGraphStep filtering, to use P.within to check within lists of ids instead of using P.eq to compare entire list. 

This preserves the effect where that one cannot query list typed ids (because we look within the list) that has been present historically with vertex/edge id filtering in g.V()/g.E(). 

It also fixes the behavior where arrays in the varargs are unrolled but not lists.


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -1608,35 +1608,31 @@ public default GraphTraversal<S, E> hasId(final Object id, final Object... other
...
+                // when hasId() is pushed down to g.V() step, it unwraps any lists in ids into their individual elements,
+                // here we will do that with lists as well to remain consistent (TINKERPOP-2863)
+                ids.addAll((Collection<?>) id);
+            } else
+                ids.add(id);
+
+            // flattening ids from other lists in varargs works cleaner with lists, as otherwise any list will need to
+            // be turned into array first
             for (final Object i : otherIds) {
                 if (i.getClass().isArray()) {
```
vkagamlyk, line 1627:
```suggestion
                if (i instanceof Object[]) {
```

vkagamlyk:
LGTM, VOTE+1

kenhuuu:
VOTE +1

spmallette:
This needs feature tests to enforce the behavior.

xiazcy:
> This needs feature tests to enforce the behavior.

Added in https://github.com/apache/tinkerpop/pull/2019/commits/be12e6cfbd6d16501fb0d680786261a7caae35d9

PR #2020, opened 2023-04-12
TINKERPOP-2863 Update hasId step mid-traversal processing - 3.6-dev

xiazcy:
Same changes as #2019, but for 3.6-dev instead, as there was an added null check in hasId() for 3.6, which led to changes to the Gherkin test suites. 

It appears that the previous implementation of hasId() omitted adding `null` to the Bytecode step if `otherIds` were `null`, which unintentionally removed tailing nulls from traversals and in the test files.

I'll squash all commits into one when this is ready to merge.


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -2446,7 +2442,7 @@ public default GraphTraversal<S, TraversalMetrics> profile() {
      * and if the {@link Element} is a {@link VertexProperty}.  This method is the long-hand version of
      * {@link #property(Object, Object, Object...)} with the difference that the {@link VertexProperty.Cardinality}
      * can be supplied.
-     * <p/>* 
+     * <p/>*
```
xiazcy, line 2445:
```suggestion
     * <p/>* 
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -2528,7 +2524,7 @@ public default GraphTraversal<S, E> property(final VertexProperty.Cardinality ca
      * the {@link Graph} will be used.
      * <p/>
      * If a {@link Map} is supplied then each of the key/value pairs in the map will
-     * be added as property.  This method is the long-hand version of looping through the 
+     * be added as property.  This method is the long-hand version of looping through the
```
xiazcy, line 2527:
```suggestion
     * be added as property.  This method is the long-hand version of looping through the 
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -2561,14 +2557,14 @@ public default GraphTraversal<S, E> property(final Object key, final Object valu
             return this.property(null, key, value, keyValues);
         }
     }
-    
+
     /**
      * When a {@link Map} is supplied then each of the key/value pairs in the map will
-     * be added as property.  This method is the long-hand version of looping through the 
+     * be added as property.  This method is the long-hand version of looping through the
```
xiazcy, line 2563:
```suggestion
     * be added as property.  This method is the long-hand version of looping through the 
```

PR #2022, opened 2023-04-14
[TINKERPOP-2924] Refactor PropertyMapStep for overwrite functionality possibilities

porunov:
Make all private fields and methods as protected to be able to overwrite some of it's functionality. This allows to overwrite some functionality for optimization purposes.

https://issues.apache.org/jira/browse/TINKERPOP-2924

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/PropertyMapStep.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/PropertyMapStep.java
@@ -61,8 +61,8 @@ public class PropertyMapStep<K,E> extends ScalarMapStep<Element, Map<K, E>>
     protected int tokens;
     protected Traversal.Admin<Element, ? extends Property> propertyTraversal;
 
-    private Parameters parameters = new Parameters();
-    private TraversalRing<K, E> traversalRing;
+    protected Parameters parameters = new Parameters();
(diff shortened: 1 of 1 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/PropertyMapStep.java
```diff
@@ -216,7 +177,75 @@ public int getIncludedTokens() {
...
+            } else {
+                map.put(property.key(), value);
+            }
+        }
+    }
+
+    protected void applyTraversalRingToMap(Map<Object, Object> map){
+        if (!traversalRing.isEmpty()) {
+            // will cop a ConcurrentModification if a key is dropped so need this little copy here
+            final List<Object> keys = new ArrayList<>(map.keySet());
```
porunov, line 239: The only logic change was here. Changed the structure from HashSet to ArrayList as HashSet copy will take N*logN complexity but ArrayList copy will take N complexity only. Moreover, the copied underlying HashSet array is 25% bigger than the underlying array of ArrayList because HashSet needs a bigger array to have smaller collision chance. In contrariety ArrayList won't have any collisions.
I.e. ArrayList here will have smaller overhead and faster insertion. We already guaranteed keys uniqueness during this list traversal by copying data from `map.keySet()`. No need to ensure uniqueness again here for below loop operation.

PR #2025, opened 2023-04-17
 Introduce new marker interfaces to identify whether a step can perform write or delete or both on graph data and static map to capture steps associated with an operator

phanindhra876:
Currently all steps that can change graph data implements Mutating interface. However, there is no straight forward way to know whether the step can perform write or delete or both on the graph data. This PR introduces two new marker interfaces `Writing` and `Deleting` defining whether step can perform write or delete or both.

In addition to this, this PR creates a static map of Traversal steps that shall be added for a given operator. Both can be combined to assess whether a given query can perform read or write or delete or any combination of these before query execution.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -31,6 +31,8 @@ This release also includes changes from <<release-3-5-6, 3.5.6>>.
 * Fixed bug in `GroovyTranslator` that surrounded `String` keys with parenthesis for `Map` when not necessary.
 * Added support to the grammar allowing `List` and `Map` key declarations for `Map` entries.
 * Improved performance of comparison (equals) between not compatible types and nulls.
+* Introduced `Writing` and `Deleting` marker interfaces to identify whether a step can perform write or delete or both on Graph.
+* Added static map capturing possible Traversal steps that shall be added to traversal for a given operator
 
 [[release-3-6-2]]
 === TinkerPop 3.6.2 (Release Date: January 16, 2023)
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/Deleting.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/Deleting.java
@@ -1,0 +1,30 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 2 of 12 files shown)


phanindhra876 (commented):
submitting for review

kenhuuu:
Thanks for contributing this. It may also be nice to add a section to the [upgrade documentation](https://github.com/apache/tinkerpop/blob/3.6-dev/docs/src/upgrade/release-3.6.x.asciidoc) for providers to let them know about this new feature.

VOTE +1

Cole-Greer:
Thanks @phanindhra876 for these changes. Overall they look good to me. My main question is if we've actually covered all the correct steps with these marker interfaces. By the look of it, this PR is just updating all of the steps which previously implemented the Mutating interface which in theory is sufficient. I wonder though if some consideration should be given into adding these markers to `io()` and possibly `call()` as well.

Also I wanted to ask if the `findPossibleTraversalStepsForModulatorOperators()` test is necessary as it appears to me that all of those modulators should already be tested by `findPossibleTraversalSteps()`. It is possible I'm misunderstanding this though so please correct me if I'm wrong.

phanindhra876:
Intention of test `findPossibleTraversalSteps` is to check we have covered all the symbols defined in the map or not. We don't do any checks whether the symbol is associated with any Step or not. To avoid over complicating the test, i have created another test to check for emptyness of ModulatingOperators.

phanindhra876:
> I wonder though if some consideration should be given into adding these markers to io() and possibly call() as well.

For the scope of this PR, i have updated steps extending Mutating to implement Writing or Deleting or both. However, `IoStep` and `CallStep` does not implement Mutating step and i need to spend some time to understand how to update them. Maybe that can be added in a separate PR? I will work with Stephen to figure this out.

spmallette:
`Writing` and `Deleting` aren't really marker interface only, because `Mutating` isn't. `Mutating` brings with it the `Event` subsystem for `EventStrategy`. I don't think that will work well with `io()` or `call()`. I think that providers will just need to account for those steps as special cases rather than test them for `Writing` or `Deleting`. for example analyzing `call()` bytecode might require the additional check of examining the service being called for that step to determine its action. so, long story short, it was good to call those steps out, but i don't think they can implement any part of the `Mutating` interface the way it is defined. VOTE +1

Cole-Greer:
> Intention of test findPossibleTraversalSteps is to check if we have covered all the symbols defined in the map or not. We don't do any checks whether the symbol is associated with any Step or not. To avoid over complicating the test, i have created another test to check for emptyness of ModulatingOperators.

I see, I had missed that the second test is asserting an empty list. Those tests look good to me. I agree with Stephen's analysis of `io()` and `call()` as well. Thanks for submitting these changes, they look good to me. VOTE +1 (Non-binding).

kenhuuu:
Do you mind squashing your commits and force pushing? Our current policy is to have the author do this rather than the person merging. Thanks.

phanindhra876:
@kenhuuu  Wouldn't the force push close this CR?

phanindhra876:
@kenhuuu Squashed all commits into one commit.

saikiranboga (approved):
lgtm

xiazcy:
VOTE +1.
I've created a JIRA here for tacking purposes: https://issues.apache.org/jira/browse/TINKERPOP-2929.

PR #2026, opened 2023-04-18
TINKERPOP-2855 Refactored strategy application

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2855
https://issues.apache.org/jira/browse/TINKERPOP-2888

Removed some recursion. Added `Traversal.lock()` to offer a formal way to finalize a traversal. Offers a significant performance improvement. Removed expectation that Graph and related metadata for a `Traversal` is available in child traversals.

On TINKERPOP-2855 the script demonstrating the slowdown was running at 9.1 seconds on my system for 3.5.5. With this change it is executing at 1.7 seconds. 

VOTE +1

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
@@ -121,63 +121,20 @@ public class DefaultTraversal<S, E> implements Traversal.Admin<S, E> {
     @Override
     public void applyStrategies() throws IllegalStateException {
         if (this.locked) throw Traversal.Exceptions.traversalIsLocked();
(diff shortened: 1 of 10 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DefaultTraversal.java
```diff
@@ -435,4 +437,4 @@ public int hashCode() {
         }
         return result;
     }
-}
+}
```
lyndonbauto, line 440: Nit - add new line.
```suggestion
}

```

lyndonbauto:
VOTE+1

Reviewed code and tried this change locally on Aerospike Graph. I had a query that was heavy on coalesce use and found that strategies were taking ~120 ms on this specific query for my local system. This change dropped the time to apply strategies to 10 ms.

lyndonbauto (approved):

PR #2033, opened 2023-04-19
TINKERPOP-2893 Remove CountStrategy optimizations for predicates with negative numbers

kenhuuu:
Fixes https://issues.apache.org/jira/browse/TINKERPOP-2893

Predicates that contain negative numbers aren't handled correctly by the current CountStrategy implementation. In particular, the use of the highRange to determine when to use not() or to remove the count().is() doesn't account for these cases. This leads to the optimization being applied when it shouldn't be. This commit temporarily solves this by disabling the optimization in these cases. Further work could be done to selectively apply the optimization as there are some cases even with predicates that contain negative numbers where it could be applied.


vkagamlyk:
VOTE+1

xiazcy:
VOTE +1

kenhuuu:
VOTE +1

PR #2050, opened 2023-04-27
[TINKERPOP-2934] Shrink the content of ObjectWritable.toString()

ministat:
Reduce the toString() of ObjectWritable to avoid OOM for running OLAP queries on Spark.

### DIFF
--- a/hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/structure/io/ObjectWritable.java
+++ b/hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/structure/io/ObjectWritable.java
@@ -32,7 +32,9 @@ import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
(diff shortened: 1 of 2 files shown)


Cole-Greer:
The one question I have about this change is if any users might currently expect to get the Collections contents to be returned in the toString output. I'm not too familiar with the use cases here so it might not be a big concern. Just wondering if there are any potential losses that should be considered.

ministat:
@Cole-Greer Generally, the requirement to get the content through the ObjectWritable.toString() is to display it in SparkUI. Too long content caused OOM. See https://issues.apache.org/jira/browse/TINKERPOP-2934.
If users want to get that content. The correct way is ObjectWritable.get().toString().

kenhuuu:
VOTE +1

xiazcy:
VOTE +1

Cole-Greer:
> If users want to get that content. The correct way is ObjectWritable.get().toString().

Thanks that sounds good to me. Looks like nobody should be negatively impacted by this change.

VOTE +1 (Non-binding)

vkagamlyk:
VOTE +1

PR #2052, opened 2023-05-01
Disable Parallelization for Gherkin tests

vkagamlyk:
Integration and Gherkin tests for .net executed in parallel, therefore random failures are possible when Gherkin test modify graph, for example `pageRank`: https://github.com/apache/tinkerpop/actions/runs/4833390970/jobs/8613413406

### DIFF
--- a/gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/GherkinTestRunner.cs
+++ b/gremlin-dotnet/test/Gremlin.Net.IntegrationTest/Gherkin/GherkinTestRunner.cs
@@ -38,6 +38,10 @@ using Xunit.Abstractions;
 
 namespace Gremlin.Net.IntegrationTest.Gherkin
 {
+    [CollectionDefinition(nameof(GherkinTestDefinition), DisableParallelization = true)]
+    public class GherkinTestDefinition { }
+
+    [Collection(nameof(GherkinTestDefinition))]
     public class GherkinTestRunner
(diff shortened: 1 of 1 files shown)


FlorianHockmann:
Just for my own understanding: The problem here is not that some Gherkin tests are executed in parallel which shouldn't happen with xUnit as tests within the same class [are executed sequentially](https://xunit.net/docs/running-tests-in-parallel), but that other integration tests are executed in parallel with the Gherkin tests, right?
And you only add this config to the Gherkin tests and not also to those other integration tests, because it's currently enough if we execute the Gherkin tests completely sequentially?

vkagamlyk:
> Just for my own understanding: The problem here is not that some Gherkin tests are executed in parallel which shouldn't happen with xUnit as tests within the same class [are executed sequentially](https://xunit.net/docs/running-tests-in-parallel), but that other integration tests are executed in parallel with the Gherkin tests, right? And you only add this config to the Gherkin tests and not also to those other integration tests, because it's currently enough if we execute the Gherkin tests completely sequentially?

Yes, some Gherkin tests temporary add properties to Modern graph and this leads to random errors. [For example](https://github.com/apache/tinkerpop/actions/runs/4833390970/jobs/8613413406). The probability of an error on the master is much greater because tests for properties have been added there.

lyndonbauto:
VOTE+1

vkagamlyk:
VOTE +1

PR #2053, opened 2023-05-04
[TINKERPOP-2941] Stop deleting the output location if it is not empty

ministat:
In production environment, it is dangerous to delete a folder which is not empty. In order to avoid any wrong operations, it is better ask user to delete it manually if it is not empty.

### DIFF
--- a/hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/Constants.java
+++ b/hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/Constants.java
@@ -73,6 +73,7 @@ public final class Constants {
     public static final String GREMLIN_SPARK_PERSIST_STORAGE_LEVEL = "gremlin.spark.persistStorageLevel";
     public static final String GREMLIN_SPARK_SKIP_PARTITIONER = "gremlin.spark.skipPartitioner"; // don't partition the loadedGraphRDD
     public static final String GREMLIN_SPARK_SKIP_GRAPH_CACHE = "gremlin.spark.skipGraphCache";  // don't cache the loadedGraphRDD (ignores graphStorageLevel)
+    public static final String GREMLIN_SPARK_DONT_DELETE_NON_EMPTY_OUTPUT = "gremlin.spark.dontDeleteNonEmptyOutput"; // don't delete the output if it is not empty
     public static final String SPARK_SERIALIZER = "spark.serializer";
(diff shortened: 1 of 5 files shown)


hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/Constants.java
```diff
@@ -73,6 +73,7 @@ private Constants() {
     public static final String GREMLIN_SPARK_PERSIST_STORAGE_LEVEL = "gremlin.spark.persistStorageLevel";
     public static final String GREMLIN_SPARK_SKIP_PARTITIONER = "gremlin.spark.skipPartitioner"; // don't partition the loadedGraphRDD
     public static final String GREMLIN_SPARK_SKIP_GRAPH_CACHE = "gremlin.spark.skipGraphCache";  // don't cache the loadedGraphRDD (ignores graphStorageLevel)
+    public static final String GREMLIN_SPARK_DONT_DELETE_NONE_EMPTY_OUTPUT = "gremlin.spark.dontDeleteNonEmptyOutput"; // don't delete the output if it is not empty
```
kenhuuu, line 76: Nit: I think there is a typo here. It should probably be GREMLIN_SPARK_DONT_DELETE_NON_EMPTY_OUTPUT

kenhuuu:
Could you please also add a CHANGELOG entry?

VOTE +1, pending minor fixes.

hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/Constants.java
ministat: Ack

ministat:
>Could you please also add a CHANGELOG entry?
I have no idea for which CHANGELOG entry I should add. My working branch is 3.5-dev and I don't know which release will include this patch.

kenhuuu:
> I have no idea for which CHANGELOG entry I should add. My working branch is 3.5-dev and I don't know which release will include this patch.

The CHANGELOGs for each subsequent branch state that they include the items from the previous branches. Since you are merging into 3.5-dev, you can just add a short entry to the upcoming release for 3.5. Just place it here https://github.com/apache/tinkerpop/blob/3.5-dev/CHANGELOG.asciidoc#L27

ministat:
@kenhuuu Thanks. I have added a new entry.

vkagamlyk:
VOTE+1

PR #2056, opened 2023-05-06
TINKERPOP-2703 Build on JDK17

kenhuuu:
Fixes https://issues.apache.org/jira/browse/TINKERPOP-2703


docs/src/upgrade/release-3.7.x.asciidoc
```diff
@@ -146,6 +146,24 @@ See: link:https://issues.apache.org/jira/browse/TINKERPOP-2819[TINKERPOP-2819]
 distributions. Any app which makes use of both `gremlin-driver` and `gremlin-server` will now need to directly
 include both modules.
 
+==== Building and Running with JDK 17
```
xiazcy, line 149: Should we also add a section in the developer doc section for this?

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerGraph.java
```diff
@@ -103,9 +103,9 @@ public final class TinkerGraph implements Graph {
     protected TinkerIndex<TinkerVertex> vertexIndex = null;
     protected TinkerIndex<TinkerEdge> edgeIndex = null;
 
-    protected final IdManager<?> vertexIdManager;
-    protected final IdManager<?> edgeIdManager;
-    protected final IdManager<?> vertexPropertyIdManager;
+    protected IdManager<?> vertexIdManager;
+    protected IdManager<?> edgeIdManager;
+    protected IdManager<?> vertexPropertyIdManager;
```
lyndonbauto, line 108: Is the removal of `final` here required?

docs/src/upgrade/release-3.7.x.asciidoc
kenhuuu: Yeah, that is a good idea.

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerGraph.java
kenhuuu: I tried to explain this in the commit message.

> The final modifier was removed in some TinkerGraph manager fields as these were removed during testing with reflection. However, as of Java 18 there is no reasonable way to use reflection to do this. Also, the class is marked final which should reduce its impact.

Please let me know if you have further concerns about this.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/jsr223/ScriptEngineCacheTest.java
```diff
@@ -35,6 +37,9 @@ public void shouldBeUtilityClass() throws Exception {
 
     @Test
     public void shouldGetEngineFromCache() {
+        String javaVersion = System.getProperty("java.version");
```
spmallette, line 40: is there another way to test this that perhaps doesn't require us to use nashorn? that way we could avoid binding to a particular jdk version/type.

pom.xml
```diff
@@ -198,6 +198,22 @@ limitations under the License.
         <log4j-silent.properties>file:target/test-classes/log4j-silent.properties</log4j-silent.properties>
 
         <muteTestLogs>false</muteTestLogs>
+        <jdk17JvmArgs>
```
spmallette, line 201: a comment with some details on why this is here and when/if all/any of it can be removed in the future would be nice.

docs/src/upgrade/release-3.7.x.asciidoc
spmallette: i suspect there are more places in the docs that make various proclamations about JDK versions. please search around for a bit and see what might need updating.

kenhuuu: I added a small section to the developer documentation that mentioned running with various JDK versions.

> i suspect there are more places in the docs that make various proclamations about JDK versions. please search around for a bit and see what might need updating.

I looked around for this but most of the places that mention which Java runtime to use are generic and don't refer to a specific version. In the areas that specifically mention using Java 11 (like building documentation), I think it makes sense to keep them like that for now as I don't think JDK 17 will become our default choice yet.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/jsr223/ScriptEngineCacheTest.java
kenhuuu: There doesn't seem to be any default ScriptEngines that ship with the JDK anymore. To actually test the current method as is we would probably need to add another ScriptEngineFactory that the service provider could load. There isn't an easy way to mock this and still test the actual implementation of the cache. I'm not sure if it is worth adding those other pieces to make the ScriptEngineManager load a ScriptEngine just to test this.

xiazcy:
VOTE +1

vkagamlyk:
VOTE +1

PR #2058, opened 2023-05-09
TINKERPOP-2944 Dispose Cancellation callbacks to fix memory leak

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2944

This should fix the memory leak described in TINKERPOP-2944 caused by not cleaning up the cancellation token registrations which contain a reference to the `RequestMessage`.

VOTE +1

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
@@ -50,14 +50,12 @@ namespace Gremlin.Net.Driver
         private readonly string _sessionId;
         private readonly bool _sessionEnabled;
 
-        private readonly ConcurrentQueue<(RequestMessage msg, CancellationToken cancellationToken)> _writeQueue =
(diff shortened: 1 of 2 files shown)


gremlin-dotnet/src/Gremlin.Net/Driver/Connection.cs
```diff
@@ -175,6 +175,7 @@ private void HandleReceivedMessage(ResponseMessage<List<object>> receivedMsg)
             {
                 responseHandler?.Finalize(status.Attributes);
                 _callbackByRequestId.TryRemove(receivedMsg.RequestId.Value, out _);
+                _cancellationByRequestId.TryRemove(receivedMsg.RequestId.Value, out _);
```
vkagamlyk, line 178: maybe it's better to move clearing of `_cancellationByRequestId` 2 lines higher to avoid leak when `responseHandler?.Finalize` failed?

vkagamlyk:
LGTM
VOTE +1

kenhuuu:
VOTE +1

PR #2062, opened 2023-05-11
fixing licenses/NOTICE for console and server distributions

Cole-Greer:
Fixing some issues found during the release of 3.5.6/3.6.4.

Noteworthy that mockito-core is being added to the licenses folder as it is currently included in the lib directory in the distributions for both console and server. I believe this inclusion is likely a mistake and should be investigated and removed at a later date. I have opted to leave it in and correct the licensing for now as I don't believe such a change should be made right before the 3.6.4 patch release gets made.

Additionally adding static NOTICE and licenses files for gremlin-shaded for inclusion in the shaded jar. Also updates notices for console and server as their distributions include gremlin-shaded.

### DIFF
--- a/gremlin-console/pom.xml
+++ b/gremlin-console/pom.xml
@@ -249,6 +249,10 @@ limitations under the License.
                                         <resource>NOTICE.txt</resource>
                                         <resource>NOTICE</resource>
                                         <resource>licenses</resource>
+                                        <!--Exclude kryo and minilog licenses to avoid conflict as they are included in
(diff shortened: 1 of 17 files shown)


vkagamlyk:
Thanks @Cole-Greer  for sorting this out.
VOTE +1

kenhuuu:
VOTE +1

xiazcy:
VOTE +1, but this will not be merged before the 3.6.4 patch release, as we are doing licenses for each branch separately.

PR #2063, opened 2023-05-12
3.6: fixing licenses/NOTICE for console and server distributions

Cole-Greer:
Fixing some issues found during the release of 3.5.6/3.6.4.

Additionally adding static NOTICE and licenses files for gremlin-shaded for inclusion in the shaded jar. Also updates notices for console and server as their distributions include gremlin-shaded.

### DIFF
--- a/gremlin-console/src/main/static/LICENSE
+++ b/gremlin-console/src/main/static/LICENSE
@@ -207,6 +207,7 @@ BSD-style Licenses
 
 The Apache TinkerPop project bundles the following components under the BSD License:
 
+     Antlr4 (org.antlr:antlr4-runtime:4.9.1 - https://www.antlr.org) - for details, see licenses/antlr4
      jcabi-log (com.jcabi:jcabi-log:0.14 - http://www.jcabi.com/jcabi-log) - for details, see licenses/jcabi-log
      jcabi-manifests 1.1 (com.jcabi:jcabi-manifests:1.1 - http://www.jcabi.com/jcabi-manifests) - for details, see licenses/jcabi-manifests
      JLine (jline:jline:2.14.6 - https://github.com/jline/jline2) - for details, see licenses/jline2
--- a/gremlin-console/src/main/static/licenses/antlr4
+++ b/gremlin-console/src/main/static/licenses/antlr4
@@ -1,0 +1,28 @@
+Copyright (c) 2012-2022 The ANTLR Project. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+
+3. Neither name of copyright holders nor the names of its contributors
+may be used to endorse or promote products derived from this software
+without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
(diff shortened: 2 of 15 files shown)


gremlin-console/src/main/static/LICENSE
```diff
@@ -216,6 +216,10 @@ The Apache TinkerPop project bundles the following components under the BSD Lice
      minlog (com.esotericsoftware:minlog:1.3.0 - https://github.com/EsotericSoftware/minlog)
        - shaded in gremlin-shaded to org.apache.tinkerpop.shaded.minlog
        - for details, see licenses/minlog
+     Antlr4 (org.antlr:antlr4-runtime:4.9.1 - https://www.antlr.org) - for details, see licenses/antlr4
```
spmallette, line 219: the groovy jar LICENSE already references ANTLR (not the detail level we do, but i suppose that is their style). i don't think we need to do it again. it just complicates our LICENSE. simply including

gremlin-console/src/main/static/licenses/antlr4
```diff
@@ -0,0 +1,28 @@
+Copyright (c) 2012-2022 The ANTLR Project. All rights reserved.
```
spmallette, line 1: already in the groovy jar. i don't think we need to include it.

gremlin-console/src/main/static/licenses/asm
```diff
@@ -0,0 +1,27 @@
+ASM: a very small and fast Java bytecode manipulation framework
```
spmallette, line 1: already in the groovy jar. i dont' think we need to include it.

gremlin-console/src/main/static/LICENSE
Cole-Greer: I'm certainly not an expert in this domain but my understanding was that for any dependency, including transitive ones, if any binary/source is packaged within our distribution then we need to meet any NOTICE/license obligations at the root level. If it is acceptable for those to be buried within the jar that contains the transitive dependency then ASM can definitely be removed here.

On the other hand the ANTLR jar is directly included in lib directory in both the console and server distributions. In this case I don't think the license buried within the groovy jar is good enough.

gremlin-console/src/main/static/licenses/asm
Cole-Greer: Sure, I wasn't aware that inclusion in groovy was sufficient. I will remove ASM from the licenses.

gremlin-console/src/main/static/LICENSE
spmallette: oh...ok. then yeah for antlr then. 

for asm i think a less complicated NOTICE beats the buried one, but this is all an art. i tried to write "rules" as best i could in all my learnings from the ASF on how to do this, but i wouldn't say my "rules" are hard and fast. 

if you really want to dig into it, you could try to get an answer from the ASF. i'd guess the incubator would be the best place to ask this question. be prepared for 10 different answers though 🙂

Cole-Greer: I have removed ASM licenses as they are properly covered by groovy. The BSD requirements are that redistributions must reproduce the license and it is definitely reproduced within the groovy jar. This argument could be extended to remove kryo and minilog from the console and server licenses as this PR is adding proper licensing for them within the gremlin-shaded jar.

gremlin-server/src/main/static/NOTICE
```diff
@@ -64,3 +64,29 @@ HPPC borrowed code, ideas or both from:
...
 ------------------------------------------------------------------------
 Copyright 2014 The Netty Project
+
+------------------------------------------------------------------------
+Objenesis 2.4 (Included through gremlin-shaded)
+------------------------------------------------------------------------
+Copyright 2006-2016 Joe Walnes, Henri Tremblay, Leonardo Mesquita
+
+------------------------------------------------------------------------
+Jackson-Databind 2.14.0 and Jackson-Core (Included through gremlin-shaded)
```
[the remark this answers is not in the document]
Cole-Greer: I want to push for a slightly less abbreviated version as

> It was originally written by Tatu Saloranta (tatu.saloranta@iki.fi), and has been in development since 2007.
> It is currently developed by a community of developers.

The original notice has a clear intent to give attribution to the community in addition to the original author and I believe we should preserve that. I've updated the notice accordingly (and fixed the ordering). If you still think a shorter version is preferable I can update it again.

xiazcy:
Thanks Cole! VOTE +1, merging as CTR.

PR #2065, opened 2023-05-16
Added missing annotation in test

lvca:
I recently upgrade ArcadeDB to the latest 3.6.3, but there is a test that won't pass in `org.apache.tinkerpop.gremlin.process.traversal.step.map.CountTest`:

```java
        @Override
        public Traversal<Vertex, Long> get_g_VX1X_valuesXageX_countXlocalX() {
            return g.V(1).values("age").count(Scope.local);
        };
```

I think the following annotation was missing:

```java
@FeatureRequirement(featureClass = Graph.Features.VertexFeatures.class, feature = Graph.Features.VertexFeatures.FEATURE_USER_SUPPLIED_IDS)
```


vkagamlyk:
Thanks for fixing this, but it's better to make `id` a test parameter like in other tests: https://github.com/apache/tinkerpop/blob/0ecc91dd10c5e4595052e9c52c320ac0a3a2969f/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/OrderTest.java#L503 In this case the test will work without `@FeatureRequirement`

lvca:
@vkagamlyk ok for me. ETA?

lvca:
I've updated it.

vkagamlyk:
Fixed in 3.5-dev in https://github.com/apache/tinkerpop/commit/822f652916f15be311f2f7760e5722a37813e38c

lvca:
Any ETA to cherry-pick it in 3.6.x?

PR #2068, opened 2023-05-16
fixing licenses/NOTICE Master

Cole-Greer:
Fixing missing entries in console and server distributions which were found during the 3.5.6 release

Adding static NOTICE and licenses files for gremlin-shaded for inclusion in the shaded jar. Also updates notices for console and server as their distributions include gremlin-shaded.

Also fixing missing entries in the gremlin-driver shaded jar


kenhuuu:
Minor changes that were already approved in similar PR for 3.5-dev and 3.6-dev. Merging as CTR.

PR #2079, opened 2023-05-30
[TINKERPOP-2955] Support SSL in WebSocketClient

ministat:
Support SSL connection in WebSocketClient


ministat:
I'm working on building a RESTful service which provides gremlin query capability and integrated with the company's frontend. Its implementation depends on gremlin-driver and gremlin-server.

For security requirement, I have enabled SSL. Currently, the WebSocketClient is not a mandatory requirement. But I'm thinking if the frontend supports WebSocket which is on plan, I also need to consider using WebSocketClient.

spmallette:
sorry, i saw your reply and forgot to respond last week. ok, if you think you need `WebSocketClient` for your testing then this change is fine i guess. 

VOTE +1

PR #2086, opened 2023-06-05
Add gremlin-java8.bat for Java 8 Windows Console users.

kenhuuu:
The gremlin.bat file no longer works for users of Java 8 due to some changes in how pathing is determined in Groovy. This bat file finds all jars in lib and ext rather than using the * wildcard as that no longer works in Groovy. Because of the hard limit on the length of a variable for commands, relative paths had to be used instead of absolute paths. Although the use of relative paths for the -cp option is discouraged by Java, it is necessary in this case.


spmallette:
i can't test, but the reasoning makes sense to me and i'm fine with the approach. VOTE +1

Cole-Greer:
VOTE +1

PR #2090, opened 2023-06-12
TINKERPOP-2958 Added cancel for query timeout task when query completes.

lyndonbauto:
I noticed extreme memory usage in recent versions of TinkerPop. After some investigation I found that this was because the task that is scheduled to cancel the query if it doesn't complete in time is not cancelled when the query completes.

This means that if the timeout is large and you have high throughput on the graph, memory starts adding up. On my laptop with Aerospike Graph, I was seeing ~1 GB per second growth in memory until it hit about 10 GB and then my system would slow down and eventually OOME.

I added a cancel for this in this PR to fix this issue and tested it and the memory growth does not happen. (I tested with changes on 3.6-dev but then brought those to 3.5-dev since the problem is rooted there).

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Upgraded `gremlin-go` to Go 1.20.
 * Improved the python Translator class
 * Added `gremlin-java8.bat` file as a workaround to allow loading the console using Java 8 on Windows.
+* Fixed a bug in `gremlin-server` where timeout tasks were not cancelled and could cause very large memory usage when timeout is large.
 
 [[release-3-5-6]]
 === TinkerPop 3.5.6 (Release Date: May 1, 2023)
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Context.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Context.java
@@ -35,6 +35,7 @@ import org.slf4j.LoggerFactory;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
@@ -55,6 +56,9 @@ public class Context {
     private final RequestContentType requestContentType;
     private final Object gremlinArgument;
     private final AtomicBoolean startedResponse = new AtomicBoolean(false);
+    private ScheduledFuture<?> timeoutExecutor = null;
+    private boolean timeoutExecutorGrabbed = false;
+    private final Object timeoutExecutorLock = new Object();
 
     /**
      * The type of the request as determined by the contents of {@link Tokens#ARGS_GREMLIN}.
@@ -90,6 +94,25 @@ public class Context {
         this.gremlinArgument = requestMessage.getArgs().get(Tokens.ARGS_GREMLIN);
         this.requestContentType = determineRequestContents();
         this.requestTimeout = determineTimeout();
+    }
+
+    public void setTimeoutExecutor(final ScheduledFuture<?> timeoutExecutor) {
+        synchronized (timeoutExecutorLock) {
+            this.timeoutExecutor = timeoutExecutor;
+
+            // Timeout was grabbed before we got here, this means the query executed before the timeout was created.
+            if (timeoutExecutorGrabbed) {
+                // Cancel the timeout.
+                this.timeoutExecutor.cancel(true);
+            }
+        }
+    }
+
+    public ScheduledFuture<?> getTimeoutExecutor() {
+        synchronized (timeoutExecutorLock) {
+            timeoutExecutorGrabbed = true;
+            return this.timeoutExecutor;
+        }
     }
 
     /**
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/session/SessionOpProcessor.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/session/SessionOpProcessor.java
@@ -65,6 +65,7 @@ import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Future;
 import java.util.concurrent.FutureTask;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 import java.util.function.Supplier;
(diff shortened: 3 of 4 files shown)


vkagamlyk:
Can similar problem occur in `GremlinExecutor.eval`?

Are you planning to open separate PR for 3.5 and 3.6?

lyndonbauto:
> Can similar problem occur in `GremlinExecutor.eval`?
I looked and as far as I can tell, this issue is limited to the two places I fixed it, however I am not well versed with this part of the gremlin code base so I may be wrong.

> Are you planning to open separate PR for 3.5 and 3.6?
This code is identical in 3.5 and 3.6 so my plan was to merge this to 3.5-dev then merge 3.5-dev into 3.6-dev. I just noticed I was pointed at master but meant to be pointed at 3.5-dev in this PR. @spmallette is this the right merging strategy?

kenhuuu:
Hi Lyndon, is it possible for you to share what kind of testing you were running that showed this issue?

lyndonbauto:
@kenhuuu 
> Hi Lyndon, is it possible for you to share what kind of testing you were running that showed this issue?

Hey Ken, yeah sure, this isn't pretty code but here it is:

```
package com.aerospike.firefly.server;

import com.aerospike.firefly.Server;
import org.apache.tinkerpop.gremlin.driver.Cluster;
import org.apache.tinkerpop.gremlin.driver.remote.DriverRemoteConnection;
import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
import org.apache.tinkerpop.gremlin.structure.Property;
import org.junit.Test;

import java.util.ArrayList;
import java.util.List;
import java.util.Random;
import java.util.Timer;
import java.util.TimerTask;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.Future;
import java.util.stream.Collectors;

import static org.apache.tinkerpop.gremlin.driver.Tokens.ARGS_EVAL_TIMEOUT;
import static org.apache.tinkerpop.gremlin.process.traversal.AnonymousTraversalSource.traversal;

public class TestServer {

    private static final long SECOND = 1000;
    private static final long MINUTE = 60 * SECOND;
    private static final long TEST_DURATION = 4 * MINUTE;
    private long totalExecuted = 0;
    Server server;
    @Test
    public void testServer() throws Exception {
        server = Server.main(new String[] {"../conf/firefly-gremlin-server-local.yaml"});
        HEAP_REPORTING_TIMER.schedule(new HeapReport(), 0, 5000);
        exposeLeak();
    }

    private static final Timer HEAP_REPORTING_TIMER = new Timer(true);

    private class HeapReport extends TimerTask {
        @Override
        public void run() {
            System.gc();
            System.gc();
            System.gc();
            System.out.printf("Heap size: %d MB%n", Runtime.getRuntime().totalMemory() / 1024 / 1024);
            System.out.printf("Executed %d queries%n", totalExecuted);
        }
    }

    public void exposeLeak() throws Exception {
        final long startTime = System.currentTimeMillis();
        Cluster cluster = Cluster.build().addContactPoint("localhost").port(8182).create();
        DriverRemoteConnection drc = DriverRemoteConnection.using(cluster, "g");
        // final DriverRemoteConnection drc = DriverRemoteConnection.using("localhost", 8182, "g");
        final GraphTraversalSource g = traversal().withRemote(drc);
        final ExecutorService executorService = Executors.newFixedThreadPool(16);

        g.with(ARGS_EVAL_TIMEOUT, 1L).V().drop().iterate();
        for (int i = 0; i < 25000; i++) {
            g.addV("randomVertex").
                    property("last_seen", String.format("%d", System.currentTimeMillis())).
                    property("something", String.format("%d", i)).
                    property("otherThing", String.format("%d", i)).
                    property("anotherThing", String.format("%d", i)).
                    property("yetAnotherThing", String.format("%d", i)).
                    property("andAnotherThing", String.format("%d", i)).
                    property("andYetAnotherThing", String.format("%d", i)).
                    iterate();
        }

        final List<Object> ids = g.V().id().toList();
        final List<Future<List<? extends Property>>> futures = new ArrayList<>();
        while (System.currentTimeMillis() - startTime <= TEST_DURATION) {
            while (futures.size() > 500) {
                Thread.sleep(100);
                final List<Future<List<? extends Property>>> completedFutures = futures.stream().filter(Future::isDone).collect(Collectors.toList());
                completedFutures.forEach(f -> {
                    try {
                        f.get();
                    } catch (Exception e) {
                        e.printStackTrace();
                    }
                });
                totalExecuted += completedFutures.size();
                futures.removeAll(completedFutures);
            }
            final Random rand = new Random();
            final List<Object> idz = new ArrayList<>();
            for (int i = 0; i < 9; i++) {
                idz.add(ids.get(rand.nextInt(ids.size())));
            }
            futures.add(executorService.submit(() -> g.V(idz).properties("last_seen").toList()));
        }
        futures.forEach(f-> {
            try {
                f.get();
            } catch (Exception e) {
                e.printStackTrace();
            }
        });
        drc.close();
        Thread.sleep(30000);
    }
}
```

The server code is:

```
package com.aerospike.firefly;

import org.apache.tinkerpop.gremlin.server.GremlinServer;
import org.apache.tinkerpop.gremlin.server.Settings;
import org.apache.tinkerpop.gremlin.server.util.ServerGremlinExecutor;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

import java.util.concurrent.CompletableFuture;

public class Server {
    private static final Logger logger = LoggerFactory.getLogger(Server.class);
    private GremlinServer gremlinServer = null;
    private final String confPath;
    private CompletableFuture<Void> serverStarted = null;
    private CompletableFuture<Void> serverStopped = null;

    public Server(final String file) {
        confPath = file;
    }

    public static Server main(final String[] args) {
        if (args.length != 1) {
            System.err.println("Usage: Server <conf file>");
            System.exit(1);
        }
        String file = args[0];
        Server fireflyServer = new Server(file);
        fireflyServer.start().exceptionally(t -> {
            logger.error("Firefly Server was unable to start and will now begin shutdown", t);
            fireflyServer.stop().join();
            return null;
        }).join();
        return fireflyServer;
    }

    public synchronized CompletableFuture<Void> start() {
        if (serverStarted != null) {
            return serverStarted;
        }
        serverStarted = new CompletableFuture<>();
        try {
            Settings settings = Settings.read(confPath);
            gremlinServer = new GremlinServer(settings);
            serverStarted = CompletableFuture.allOf(gremlinServer.start());
        } catch (Exception ex) {
            serverStarted.completeExceptionally(ex);
        }
        return serverStarted;
    }

    public synchronized CompletableFuture<Void> stop() {
        if (gremlinServer == null) {
            return CompletableFuture.completedFuture(null);
        }
        if (serverStopped != null) {
            return serverStopped;
        }
        serverStopped = gremlinServer.stop();
        return serverStopped;
    }
}
```
which i adapted from the JanusGraphServer implementation specifically to test this since I didn't have any in Java server bootstrapping code. 

Basically you just run and watch the memory, for Aerospike Graph after a few mins we see about 1 GB total if the bug is not present, if the bug is present I see ~1 GB per minute growth and will see ~3-4 GB at the end of the test. 

This is kind of obvious, but make sure the evaluationTimeout in the server yaml is big (mine was 20 mins).

kenhuuu:
VOTE +1

The same kind of issue exists with the HttpGremlinEndpointHandler (`GremlinExecutor.eval` as Valentyn mentioned) and the UnifiedHandler but to a far lesser degree so I think it is fine to keep those as is. The `Context` object is much larger which is why it affects these two processors much more.

lyndonbauto:
Thanks for looking into that Ken, and yeah I noticed that this didn't really become an issue until the Context was added to it.

xiazcy:
LGTM, VOTE +1

PR #2094, opened 2023-06-13
TINKERPOP-2959 Allowed parsing of parameters with the grammar

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2959

This adds functionality similar to `GremlinGroovyScriptEngine` where bindings can be applied to a script as in `g.V(x)` where `x` is a variable that can be bound to a value external to the script. This functionality doesn't offer any particular performance enhancement for the grammar the way that it does for groovy but it does better allow one engine to be swapped for the other which is crucial for a migration away from the groovy engine. This change also opens up better opportunities for translator and gherkin improvements as well.

Purposefully did not add upgrade docs as I still don't think this functionality is completely ready for general usage.

This is a fairly large diff because it essentially rewrites the core of the antlr grammar and that force a rewrite of much of the parser. It might be best to focus on what's being tested and to try to determine if anything further should be tested in relation to this functionality. 

Performance testing showed that somehow, these changes actually made the parser a tiny bit faster. I basically ran all the Gremlin (minus a few troublemakers) in the documentation (about 600 queries) through the parser 10000 times. My runs with these changes tended to finish in about 65~ seconds typically. Runs on `master` prior to these changes typically ran around 72~ seconds. Here's the script which I just pasted into Gremlin Console (if you want to run be sure to replace `TINKERPOP_HOME_PATH` with the path to your TinkerPop git repo home:

```groovy
import org.apache.tinkerpop.gremlin.language.corpus.*
g = TinkerGraph.open().traversal()
b = new javax.script.SimpleBindings()
b.put('g', g)
lang = new GremlinLangScriptEngine()
home = 'TINKERPOP_HOME_PATH'
gremlins = DocumentationReader.parse(home + '/docs/src'); []
gremlins = gremlins.findAll{ !it.contains("g.union") && !it.contains("coalesce(E") && !it.contains("eq(constant('josh'))") && !it.contains("xx") && !it.contains("branch") && !it.contains("program") && !it.contains("profile") && !it.contains("explain") && !it.contains("connectedComponent") && !it.contains("fill") && !it.contains("next") && !it.contains("toList") && !it.contains("iterate") && !it.contains("withComputer") && !it.contains("{") && !it.contains("withoutStrategies") };[]
start = System.currentTimeMillis()
gremlins.each { lang.eval(it, b);[] }
(0..10000).each {
    if (it % 10 == 9) 
        println "."
    else 
        print "."
    gremlins.each { lang.eval(it, b);[] }
}
end = System.currentTimeMillis() - start
```

VOTE +1


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/ArgumentVisitor.java
```diff
@@ -0,0 +1,420 @@
...
+                int length = Array.getLength(l);
+                Object[] result = new Object[length];
+                for (int i = 0; i < length; i++) {
+                    result[i] = Array.get(l, i);
+                }
+                return result;
+            } else if (l instanceof Collection) {
+                Collection<?> collection = (Collection<?>) l;
+                return collection.toArray();
+            } else {
```
vkagamlyk, line 196: is it possible to get Vararg as `Iterable`?

xiazcy:
This feature definitely opens up improvement opportunties. I'm not too familiar with many of the technical details, but on a high level the change looks good. I also like how the grammar variables are more intuitive now. VOTE +1

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/language/grammar/ArgumentVisitorTest.java
```diff
@@ -0,0 +1,260 @@
...
+                {Float.class, "x", new VariableResolverException("x"), createAntlr(VariableResolver.NoVariableResolver.instance())},
+                {Float.class, "0.0d", 0.0, createAntlr(new VariableResolver.DefaultVariableResolver(ElementHelper.asMap("x", 1000.0)))},
+                {Float.class, "0d", 0.0, createAntlr(new VariableResolver.DefaultVariableResolver(ElementHelper.asMap("x", 1000.0)))},
+                {Float.class, "0F", 0.0F, createAntlr(new VariableResolver.DefaultVariableResolver(ElementHelper.asMap("x", 1000.0F)))},
+                {Float.class, "x", 0.0, createAntlr(new VariableResolver.DefaultVariableResolver(ElementHelper.asMap("x", 0.0)))},
+                {Float.class, "x", 0.0F, createAntlr(new VariableResolver.DefaultVariableResolver(ElementHelper.asMap("x", 0.0F)))},
+                {String.class, "x", new VariableResolverException("x"), createAntlr(VariableResolver.NoVariableResolver.instance())},
+                {String.class, "'test'", "test", createAntlr(new VariableResolver.DefaultVariableResolver(ElementHelper.asMap("x", "nope")))},
```
vkagamlyk, line 96: is it possible to pass `null` as string variable value?

gremlin-language/src/main/antlr4/Gremlin.g4
```diff
@@ -1403,7 +1521,7 @@ genericLiteralMap
     | LBRACK mapEntry (COMMA mapEntry)* RBRACK
     ;
 
-// allow builds of Map that sorta make sense in the Gremlin context
+// allow builds of Map that sorta make sense in the Gremlin context.
 mapEntry
     : NEW COLON genericLiteral  // explicit for [new: true] - if we had other keywords like that maybe we'd group them up?
     | (LPAREN stringLiteral RPAREN | stringLiteral) COLON genericLiteral
```
Cole-Greer, line 1527: I believe that for each of these, the value part of the mapEntry should be a `genericLiteralArgument` instead of just a `genericLiteral`. Just checked in console and groovy allows passing variables as the value in maps.

Cole-Greer:
Thanks for putting this all together Stephen. I'm a big fan of anything which reduces the gap between gremlin-lang and gremlin-groovy. This is a bit out of my wheelhouse but I took a good look through the new grammar and tests. Other than the one question I posted above about mapEntry, it all looks good to me.

gremlin-language/src/main/antlr4/Gremlin.g4
spmallette: ehhh - i purposefully didn't allow that. if you want a variable in a `Map` for the grammar then you gotta do the whole `Map`. if we hate that in the future we can always open that up i guess.

spmallette:
@vkagamlyk addressed your comments and pushed a new version

vkagamlyk:
> @vkagamlyk addressed your comments and pushed a new version

Thank you!

VOTE +1

Cole-Greer:
LGTM VOTE +1

PR #2099, opened 2023-06-16
String manipulation function concat()

xiazcy:
Implementation for string `concat()` with reference to on the [remove-closures proposal](https://github.com/apache/tinkerpop/blob/master/docs/src/dev/future/proposal-3-remove-closures.asciidoc#concat), and not considering the scope for list `concat()`

As discussed on dev-list, Scope has been removed from the original proposal.


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Operator.java
```diff
@@ -204,5 +204,5 @@ else if (a instanceof Collection && a instanceof Collection)
         public Object apply(final Object a, final Object b) {
             return (long) a + (long) b;
         }
-    }
+    },
```
xiazcy, line 207:
```suggestion
    }
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/DefaultGremlinBaseVisitor.java
```diff
@@ -911,6 +911,20 @@ protected void notImplemented(final ParseTree ctx) {
 	 * {@inheritDoc}
 	 */
 	@Override public T visitTraversalMethod_call_string_map_traversal(final GremlinParser.TraversalMethod_call_string_map_traversalContext ctx) { notImplemented(ctx); return null; }
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
```
spmallette, line 917: please single line these like the others?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/ConcatStep.java
```diff
@@ -0,0 +1,106 @@
...
+import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
+import org.apache.tinkerpop.gremlin.process.traversal.Traverser;
+import org.apache.tinkerpop.gremlin.process.traversal.step.TraversalParent;
+import org.apache.tinkerpop.gremlin.process.traversal.traverser.TraverserRequirement;
+import org.apache.tinkerpop.gremlin.process.traversal.util.TraversalUtil;
+
+import java.util.Collections;
+import java.util.Set;
+
+public class ConcatStep<S> extends ScalarMapStep<S, String> implements TraversalParent {
```
spmallette, line 30: some basic javadoc for the class would be nice

spmallette: how do you feel about `final` for this class?

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -609,6 +609,29 @@ link:https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/java/o
 link:https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/service/ServiceRegistry.java[ServiceRegistry],
 link:https://tinkerpop.apache.org/docs/x.y.z/reference/#call-step[reference]
 
+[[concat-step]]
```
spmallette, line 612: i think this section should address nulls.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/ConcatStepTest.java
```diff
@@ -0,0 +1,67 @@
...
+                __.__("a", "b", "c", "d").concat().toList().toArray());
+        assertArrayEquals(new String[]{"ade", "bde", "cde"},
+                __.__("a", "b", "c").concat("d", "e").toList().toArray());
+        assertArrayEquals(new String[]{"abc", "bbc"},
+                __.__("a", "b").concat(__.__("b", "c")).toList().toArray());
+
+        assertArrayEquals(new String[]{"Mr.a", "Mr.b", "Mr.c", "Mr.d"},
+                __.__("a", "b", "c", "d").as("letters").constant("Mr.").concat(__.select("letters")).toList().toArray());
+
+        String nullStr = null;
```
spmallette, line 58: is this sufficient testing for null? what if one value is null and the rest of the stream is strings for example?

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/process/graph-traversal.js
```diff
@@ -543,6 +543,16 @@ class GraphTraversal extends Traversal {
     return this;
   }
 
+  /**
```
spmallette, line 546: i don't believe this is the only change required for javascript. you need the `statics` method definition as well

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/ProcessStandardSuite.java
```diff
@@ -144,6 +145,7 @@ public class ProcessStandardSuite extends AbstractGremlinSuite {
             AddEdgeTest.Traversals.class,
             AddVertexTest.Traversals.class,
             CoalesceTest.Traversals.class,
+            ConcatTest.Traversals.class,
```
spmallette, line 148: i think TinkerPop needs to start pressing on providers to get off of the old java process tests. its time consuming to maintain both these and the gherkin tests as duplicates. not sure if it's good to start this path where all the new string functions that follow will need java and gherkin tests. thoughts?

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -609,6 +609,29 @@ link:https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/java/o
...
+
+*Description:* Concatenates the incoming String traverser with the input String arguments, and return the joined String.
+If the incoming traverser is a non-String value then an `IllegalArgumentException` will be thrown.
+
+*Syntax:* `concat()` | `concat(String...)` | `concat(Traversal)`
+
+[width="100%",options="header"]
+|=========================================================
+|Start Step |Mid Step |Modulated |Domain |Range
+|N |Y |N |`Map` |`String`
```
Cole-Greer, line 623: I believe this should be:
|N |Y |N |`String` |`String`

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -609,6 +609,29 @@ link:https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/java/o
...
+
+*Arguments:*
+
+* `String...` - If one or more String values are provided, they will be concatenated with the
+incoming traverser. If no argument is provided, the String value from the incoming traverser is returned.
+* `Traversal` - The `Traversal` value must resolve to a `String`.
+
+See: link:https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/ConcatStep.java[source],
+link:https://tinkerpop.apache.org/docs/x.y.z/reference/#concat-step[reference]
+
```
Cole-Greer, line 634: Would be good to include a brief *Exceptions* section here similar to the Call step docs. Just a one liner  regarding IllegalArgumentException would be good (also any exceptions related to nulls if applicable).

docs/src/reference/the-traversal.asciidoc
```diff
@@ -1049,6 +1049,41 @@ g.V().coin(1.0)
...
+<1> Add a new vertex with id 10 which should be labeled like an existing vertex but with some prefix attached
+<2> Attach the prefix "Mr." to all the names using the `constant()`-step
+<3> Generate a string of software names and the language they use
+<4> Generate a string description for each of marko's outgoing edges
+<5> A generic use of `concat()` to join strings together
+
+*Additional References*
+
+link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.html#concat-String++[`concat(String...)`]
+link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.html#concat-Traversal++[`concat(Traversal)`]
```
Cole-Greer, line 1085: Consider adding link to semantics docs.

link:++https://tinkerpop.apache.org/docs/x.y.z/dev/provider/#concat-step++[`Semantics`]

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/ConcatStepTest.java
xiazcy: Added additional cases.

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/ProcessStandardSuite.java
xiazcy: I would agree for string functions there is no need to keep the duality. I will remove it here.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -1049,6 +1049,44 @@ g.V().coin(1.0)
...
+    concat(' uses ').
+    concat(select('a').values('lang')) <3>
+g.V(1).outE().as("a").constant("").
+    concat(V(1).values("name")).
+    concat(" ").
+    concat(select("a").label()).
+    concat(" ").
+    concat(select("a").inV().values("name")) <4>
+g.inject("hello", "hi").concat(__.V().values('name')) <5>
+g.inject("This").concat(" ").concat("is a ", "gremlin.") <5>
```
vkagamlyk, line 1074:
```suggestion
g.inject("This").concat(" ").concat("is a ", "gremlin.") <6>
```

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs
```diff
@@ -501,6 +501,28 @@ private GraphTraversal(ICollection<ITraversalStrategy> traversalStrategies, Byte
...
+            Bytecode.AddStep("concat", concatTraversal);
+            return Wrap<TStart, TEnd>(this);
+        }
+
+        /// <summary>
+        ///     Adds the concat step to this <see cref="GraphTraversal{SType, EType}" />.
+        /// </summary>
+        public GraphTraversal<TStart, TEnd> Concat<TNewEnd> (params string?[] concatStrings)
+        {
+            // need null check?
```
vkagamlyk, line 518: yes

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/Concat.feature
```diff
@@ -0,0 +1,213 @@
...
+  Scenario: g_injectXa_bX_concat_Xinject_c_dX
+    Given the empty graph
+    And the traversal of
+      """
+      g.inject("a", "b").concat(__.inject("c", "d"))
+      """
+    When iterated to list
+    Then the result should be unordered
+      | result |
+      | acda |
```
vkagamlyk, line 70: @spmallette what is your opinion on this result?

spmallette: not what i would have expected. i think i would have expected:

```
| result |
| ac |
| bc|
```

[looking at other tests](https://github.com/apache/tinkerpop/pull/2099/files/86238327a61b7246aa182b86d7b65c135513929f#diff-3cc4e477a58382ce059167e0cede86838765da192b3b734705e68c5d1364eaceR121), it appears the behavior is to consume the whole traversal rather than `next()`. what was the thinking there? that's sorta different that what child traversals do in Gremlin. 

i guess that still doesn't explain the added "a" and "b" at the end of each result. i'm trying to understand why it does that.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/BytecodeHelper.java
```diff
@@ -196,6 +196,7 @@ public final class BytecodeHelper {
             put(GraphTraversal.Symbols.max, Arrays.asList(MaxGlobalStep.class, MaxLocalStep.class));
             put(GraphTraversal.Symbols.min, Arrays.asList(MinGlobalStep.class, MinGlobalStep.class));
             put(GraphTraversal.Symbols.mean, Arrays.asList(MeanGlobalStep.class, MeanLocalStep.class));
+            put(GraphTraversal.Symbols.concat, Collections.singletonList(ConstantStep.class));
```
kenhuuu, line 199: Typo: ConcatStep

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/Concat.feature
spmallette: The semantics of `inject()` are sometimes hard to think about and I think that's what makes this example look strange. Consider:

```
gremlin> g.inject("a","b").map(inject("c","d"))
==>a
==>b
gremlin> g.inject("a","b").map(inject("c","d").fold())
==>[a,c,d]
==>[b]
gremlin> g.inject("a","b").project("a").by(inject('c','d').fold())
==>[a:[a,c,d]]
==>[a:[b]]
gremlin> g.inject("a","b").union(inject("c","d"))
==>a
==>c
==>d
==>b
gremlin> g.inject("a","b").inject("c","d")
==>c
==>d
==>a
==>b
```

I can't seem to find context where `inject()` resets, so given that, perhaps the expectation for the `concat()` is maybe:

```
| result |
|  ac  |
|  bd |
```

???

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/Concat.feature
```diff
@@ -0,0 +1,213 @@
...
+      g.inject(null, "a").concat(null, "b")
+      """
+    When iterated to list
+    Then the result should be unordered
+      | result |
+      | b |
+      | ab |
+
+  @GraphComputerVerificationInjectionNotSupported
+  Scenario: g_injectXhello_hiX_concat_XV_valuesXnameXX
```
spmallette, line 121: as i mentioned elsewhere, this one feels off to me. i wouldn't expect the traversal given to `concat()` to be fully iterated so:

```
| result |
| hellomarko |
| himarko |
```

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/Concat.feature
xiazcy: My understanding of the spec is that concat appends the strings in the arguments together and then with the incoming traverser, with each traverser mapping to the results inside the arguments. 

I'm using TraversalUtil.applyAll to get the result of the child traversal, and In this case, "a" is appended in front of the "c" and "d" injected. Then for inject, it appends the "a" from the parent to the end. Then the map operation will move to the next incoming traverser "b" then does the same thing:
```
g.inject("a", "b").concat(__.inject("c", "d"))
==>acdb
==>bcdb
```

We thought about getting "ac" and "bd", but I wasn't sure if it'd be a scalarMap, since we'll have to keep track of matching traverser from the child traversal to the next incoming traverser for map.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/Concat.feature
spmallette: >  The reasoning for getting it fully iterated is treating the traversal results like the string inputs.

i think my main problem with this is that i can't think of any steps that work that way. no steps allow you to expand a traversal's results as multiple arguments. i'm wary of introducing that, even if it sorta makes sense and is convenient because then you have to know which step does an expansion and which just takes the first value as a single argument.

xiazcy: > i think my main problem with this is that i can't think of any steps that work that way. no steps allow you to expand a traversal's results as multiple arguments. i'm wary of introducing that, even if it sorta makes sense and is convenient because then you have to know which step does an expansion and which just takes the first value as a single argument.
Ah yes, that is true as well. I will revert to taking the first value.

xiazcy:
I believe I've addressed all existing comments so far. 

Concat with inject is now updated to the expected behaviour:
```
gremlin> g.inject('a', 'b', 'c').concat(__.inject('1', '2'))
==>a1
==>b2
==>c
gremlin> g.inject('a', 'b').concat(__.inject('1', '2', '3'))
==>a1
==>b2
```
Similarly for child traversals, reverted to iterating only the first result:
```
gremlin> g.inject('hello', 'hi').concat(__.V().values('name'))
==>hellomarko
==>himarko
```

> also, should concat(Traversal) be concat(Traversal...) to match the String form? this way you could concat(select('a'), select('b'), select('c')) which would match its String overload?

After some thought around this, I think maybe we should not implement `concat(Traversal…)` at this point, to lessen the complexity of `concat()`.
It’s also currently achievable by chaining concat():
```
gremlin> g.V().as('a').out().as('b').out().as('c').constant("").concat(select("a").label()).concat(select("b").label()).concat(select("c").label())
==>personpersonsoftware
==>personpersonsoftware
```
However, I do want to explore this avenue a bit, so if it leads somewhere I will update here or on the dev list.

vkagamlyk:
VOTE +1

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/Concat.feature
xiazcy: Updated to 
```
| result |
|  ac  |
|  bd |
```
(see latest comment).

kenhuuu:
VOTE +1

Cole-Greer:
VOTE +1

xiazcy:
Thanks for everyone's review and input! If there are no additional comments, I'll merge this end of day today.

PR #2100, opened 2023-06-16
TINKERPOP-2920 Fixed bug in SubgraphStrategy when key not present

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2920

`ProductiveByStrategy` was interacting badly with `SubgraphStrategy` and altering a child traversal in `PropertyMapStep` that it probably shouldn't have been.

VOTE +1


Cole-Greer:
LGTM VOTE +1

PR #2115, opened 2023-06-29
[TINKERPOP-2963] Introduce new mimeType to use Graphson-1.0 text serializer

phanindhra876:
Users/Providers who wish to use GraphSON-1.0 in text format should make serializer 
`GraphSONMessageSerializerV1d0` default serializer. Alternatively introduce new mimeType `application/vnd.gremlin-v1.0+json;sparse=true` to let users/providers use GraphSON-1.0 in text format without changing defaults.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ser/Serializers.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ser/Serializers.java
@@ -33,6 +33,7 @@ public enum Serializers {
      */
     GRAPHSON(SerTokens.MIME_JSON),
     GRAPHSON_V1D0(SerTokens.MIME_GRAPHSON_V1D0),
+    GRAPHSON_V1D0_SPARSE(SerTokens.MIME_GRAPHSON_V1D0_SPARSE),
     GRAPHSON_V2D0(SerTokens.MIME_GRAPHSON_V2D0),
     GRAPHSON_V3D0(SerTokens.MIME_GRAPHSON_V3D0),
     GRAPHBINARY_V1D0(SerTokens.MIME_GRAPHBINARY_V1D0);
@@ -54,6 +55,8 @@ public enum Serializers {
                 return new GraphSONMessageSerializerV3d0();
             case SerTokens.MIME_GRAPHSON_V1D0:
                 return new GraphSONMessageSerializerGremlinV1d0();
+            case SerTokens.MIME_GRAPHSON_V1D0_SPARSE:
+                return new GraphSONMessageSerializerV1d0();
             case SerTokens.MIME_GRAPHSON_V2D0:
                 return new GraphSONMessageSerializerV2d0();
             case SerTokens.MIME_GRAPHBINARY_V1D0:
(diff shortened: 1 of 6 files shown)


vkagamlyk:
Hi @phanindhra876,
Thank you for contribution!
It would be nice to add a test to confirm that this code works

phanindhra876:
Sure @vkagamlyk. I didnt find any tests for Graphson-1.0. Maybe we need a new test in `GremlinServerHttpIntegrateTest` for this change?

vkagamlyk:
Great example for new mime type can be found in [this PR](https://github.com/apache/tinkerpop/pull/2070/files#diff-8ff383b057866ad668f388dcab65b4e4b26531a82b7b9fc462aff9881ce3c51c)

phanindhra876:
Thanks for the reference. Will add necessary tests.

phanindhra876:
@vkagamlyk Added UT to verify that change works. Please review now.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ser/Serializers.java
```diff
@@ -54,6 +55,8 @@ public MessageSerializer<?> simpleInstance() {
                 return new GraphSONMessageSerializerV3d0();
             case SerTokens.MIME_GRAPHSON_V1D0:
                 return new GraphSONMessageSerializerGremlinV1d0();
+            case SerTokens.MIME_GRAPHSON_V1D0_SPARSE:
```
vkagamlyk, line 58: What serializer will be used for `MIME_JSON`?

phanindhra876: On server side it depends on the order of serializers defined in serializerSettings. Which is the first in the serializerSettings field will be picked for application/json.

I believe this snippet of code is used in gremlin driver. Not sure how it works on client side.

phanindhra876: Correction. Seems like on client side `GraphSONMessageSerializerV3d0` will be used as serializer for `MIME_JSON` as you can see on L54(old) in the same class.

xiazcy:
LGTM, VOTE +1

phanindhra876:
Please confirm if there are any concerns with this change. Else i will add an entry to capture this in 3.6.5 changelog and will squash the commit to merge the change,

saikiranboga:
What does `sparse=true` parameter represent in the mime type?

phanindhra876:
Using `application/vnd.gremlin-v1.0+json` it conflicts with GraphSONMessageSerializerGremlinV1d0. User/provider can never configure server to use both of them(same problem as `application/json`). Adding `sparse=true` distinguishes this from `application/vnd.gremlin-v1.0+json` not so much that it's a different format but very similar. We can change it to something like `application/vnd.gremlin-v1.0+json+text` or similar if sparse feels odd.

saikiranboga:
> Using `application/vnd.gremlin-v1.0+json` it conflicts with GraphSONMessageSerializerGremlinV1d0. User/provider can never configure server to use both of them(same problem as `application/json`). Adding `sparse=true` distinguishes this from `application/vnd.gremlin-v1.0+json` not so much that it's a different format but very similar. We can change it to something like `application/vnd.gremlin-v1.0+json+text` or similar if sparse feels odd.

Thanks. Something like `type=true` seem more closer to the intention.

`strict/sparse` could support more under the naming than closely linking to just type, but I am unsure if there are any plans to make the serializer support more than this.

spmallette:
working on the merge now. was ok with "sparse" but agree with using "type" instead now that i look more carefully. i'll add a follow-on commit to do that.

saikiranboga (approved):
Let's follow up with changing parameter for mime type from `sparse` to `type` as mentioned by Stephen.

spmallette:
Merged and here's the follow-on commit that renames "sparse" and does some other things  to bring a bit more consistency here: https://github.com/apache/tinkerpop/commit/14f2d31b5ebc7cedd50173f746cf8aaa95dfaa0e

PR #2124, opened 2023-07-06
TINKERPOP-2965 Fixed bug in FilterRankingStrategy

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2965

This problem showed up in 3.5.6/3.6.3 after TINKERPOP-2919 which improved performance. The changes failed to properly propogate labels up from child traversals into the TraversalParent cache it created. As a result, labels started shifting around when they shouldn't have.

VOTE +1


lvca:
Thanks@ Is it going to be in the next 3.6.5 or 3.7?

spmallette:
all of them, 3.5.7, 3.6.5 and 3.7.0

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/DepthComparator.java
```diff
@@ -0,0 +1,54 @@
...
+    public static DepthComparator instance() {
+        return instance;
+    }
+
+    @Override
+    public int compare(final Step<?, ?> step1, final Step<?, ?> step2) {
+        return getDepth(step2) - getDepth(step1);
+    }
+
+    private int getDepth(final Step<?,?> step) {
```
vkagamlyk, line 44: why is the variable `depth` needed here? can this code be simplified to 
```
private int getDepth(final Step<?,?> step) {
	return step == null || step instanceof EmptyStep ? 0 : getDepth((Step) step.getTraversal().getParent()) + 1;
}
```

spmallette: oh, yeah, that could be simplified. i think i had something else happening there where `depth` made more sense. good catch.

PR #2138, opened 2023-07-13
TINKERPOP-2967 Added untyped GraphSON 3.0

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2967

Essentialy exposes untyped GraphSON 3.0 which is basically the untyped GraphSON 1.0. Had to refactor the GraphSONModule slightly to handle this by passing TypeInfo through it to the serializers which needed to know whether they were serializing that way or not.

VOTE +1


Cole-Greer:
LGTM VOTE +1

xiazcy:
VOTE +1

PR #2150, opened 2023-07-24
TINKERPOP-2946 Detect ordering issues in Gherkin tests

Cole-Greer:
https://issues.apache.org/jira/browse/TINKERPOP-2946

Addresses TINKERPOP-2946 by introducing TinkerShuffleGraph and running feature tests against an embedded instance of this graph.

TinkerShuffleGraph is derived from TinkerGraph and intended for test purposes only. It's only differentiation from TinkerGraph is that it will returned shuffled results when getting any of the following:

vertices from the graph,
edges from the graph,
vertices from a vertex,
edges from a vertex,
vertices from an edge,
properties from a vertex,
properties from an edge, and
properties from a VertexProperty

A new feature test annotation `@InsertionOrderingRequired` has been added to flag tests which are reliant on a graph which predictably returns elements in the order of insertion (such as TinkerGraph). Such tests should be avoided whenever possible. 2 ValueMap tests require the flag due to TINKERPOP-2974, a few of the new tests for combine() and conjoin() are also order dependent which we should seek to resolve soon. There is also one test for shortestPath which I am unsure why it is dependent on ordering.


Cole-Greer:
I've updated the description with some recent changes to this. I think it is good to go now.

VOTE +1

vkagamlyk:
VOTE+1

kenhuuu:
VOTE +1

PR #2159, opened 2023-07-25
Add deprecation notice for Neo4j-Gremlin

Cole-Greer:
Based on dev list discussion, a decision was made to deprecate neo4j-gremlin. It will remain in the repo for foreseeable future and we will continue to maintain compatibility when possible. If the effort to maintain neo4j-gremlin becomes unreasonable in the future, a decision to remove it may be made at that time.

See DISCUSS thread on the dev list for more details: (https://lists.apache.org/thread/lxn4s9fs8rzggm0jlnffnphfpqnpn3h8)

VOTE +1


Cole-Greer:
I've rebased this to 3.5-dev in response to some users raising concerns about security issues with some of the transitive dependencies in neo4j-gremlin. Deprecating in all branches will give users a more accurate impression of the current level of support for the module.

vkagamlyk:
VOTE +1

PR #2157, opened 2023-07-25
TINKERPOP-2951: add translator to the Go GLV

itaywol:
Added to gremlingo the ability to translate bytecode to gremlin-groovy string.

This addresses the following [ticket](https://issues.apache.org/jira/browse/TINKERPOP-2951)

PR #2160, opened 2023-07-26
TINKERPOP-2951: add translator to the Go GLV

itaywol:
This readdresses the following PR https://github.com/apache/tinkerpop/pull/2157
But moved the commits to 3.5-dev as base branch

### DIFF
--- a/gremlin-go/driver/translator.go
+++ b/gremlin-go/driver/translator.go
@@ -1,0 +1,216 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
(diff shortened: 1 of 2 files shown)


itaywol:
@Cole-Greer @vkagamlyk Indeed more test were required, it inflated a little due to the fact i had to use reflection to open some of the struct underlying the interfaces

Cole-Greer:
Hi Itay,

Thanks for all the rapid changes to the PR. To build off of Valentyn’s comments once again, I think there are still additional types that will need to be supported in order for the translator to be considered complete. Support for the following will all be needed:

- Translation of nested structures (within lists and maps)
- Primitive types (Char, Byte/int8, Short/int16)
- More common structure types (Date, Datetime, UUID, BigInteger, BigDecimal)
- Strategies

The [python translator](https://github.com/apache/tinkerpop/blob/master/gremlin-python/src/main/python/gremlin_python/process/translator.py) is a good reference to follow for this.
There is still an opportunity for this to be merged tomorrow morning (Thursday in the Americas) if its complete although the release process is already well underway so this would be the final chance to make it into the upcoming release. If there is only minor tweaks needed tomorrow morning we may be able to assist with getting it finished up and into the release.

Thanks,

Cole

Cole-Greer:
Hi @itaywol, I just wanted to check in here as a bit of time has now passed. Do you still intend to return to this PR at some point? Is there anything we can help with to drive this work to completion?

itaywol:
Hey @Cole-Greer, I am actually working now on better implementation for this PR i will open new one today or tomorrow

itaywol:
@Cole-Greer What should be my base branch?

PR #2163, opened 2023-07-26
Remove TEST_TRANSACTIONS flag from GLVs

xiazcy:
As we are switching to TinkerTransactionGraph, we can run transaction tests by default and no longer need the environment flag `TEST_TRANSACTIONS` to run tests only when Neo4j plugin is installed. This flag and its usage is now removed from all GLVs.


vkagamlyk (approved):
LGTM
VOTE +1

xiazcy:
VOTE +1

kenhuuu:
VOTE +1

PR #2164, opened 2023-07-27
[TINKERPOP-2973] Enable vertex properties filtering for GraphComputer

li-boxuan:
GraphComputer has vertexFilter and edgeFilter, but not yet a filter for vertex properties. Many vertex programs don't need many if not all vertex properties. PageRank, for example, does not need any vertex properties at all. Having a vertex property filter can greatly reduce memory/disk usage.


spmallette:
neat PR - i think this is a duplicate issue of https://issues.apache.org/jira/browse/TINKERPOP-1403

@Cole-Greer suggest closing the newer one as a duplicate and merging under 1403. needs a CHANGELOG and a bullet in the reference documentation:

https://tinkerpop.apache.org/docs/current/reference/#graph-filter

other than that i think this is fine.

VOTE +1

xiazcy:
VOTE +1

Cole-Greer:
VOTE +1 Thanks for putting this together. I will follow up with a CTR which adds a bit of documentation.

PR #2174, opened 2023-07-28
TINKERPOP-2802: Add support for custom deserializers

j-white:
JIRA: https://issues.apache.org/jira/browse/TINKERPOP-2802

Working implementation for custom deserializers.

### DIFF
--- a/gremlin-go/driver/graphBinary.go
+++ b/gremlin-go/driver/graphBinary.go
@@ -37,6 +37,7 @@ type dataType uint8
 
 // dataType defined as constants.
 const (
+	customType            dataType = 0x00
 	intType               dataType = 0x01
 	longType              dataType = 0x02
 	stringType            dataType = 0x03
@@ -223,11 +224,12 @@ func instructionWriter(instructions []instruction, buffer *bytes.Buffer, typeSer
 
 // Format: {steps_length}{step_0}…{step_n}{sources_length}{source_0}…{source_n}
 // Where:
-//		{steps_length} is an Int value describing the amount of steps.
-//		{step_i} is composed of {name}{values_length}{value_0}…{value_n}, where:
-//      {name} is a String. This is also known as the operator.
-//		{values_length} is an Int describing the amount values.
-//		{value_i} is a fully qualified typed value composed of {type_code}{type_info}{value_flag}{value} describing the step argument.
+//
+//			{steps_length} is an Int value describing the amount of steps.
+//			{step_i} is composed of {name}{values_length}{value_0}…{value_n}, where:
+//	     {name} is a String. This is also known as the operator.
+//			{values_length} is an Int describing the amount values.
+//			{value_i} is a fully qualified typed value composed of {type_code}{type_info}{value_flag}{value} describing the step argument.
 func bytecodeWriter(value interface{}, buffer *bytes.Buffer, typeSerializer *graphBinaryTypeSerializer) ([]byte, error) {
 	var bc Bytecode
 	switch typedVal := value.(type) {
@@ -1323,6 +1325,26 @@ func readFullyQualifiedNullable(data *[]byte, i *int, nullable bool) (interface{
 	deserializer, ok := deserializers[dataTyp]
 	if !ok {
 		return nil, newError(err0408GetSerializerToReadUnknownTypeError, dataTyp)
+	}
+
+	return deserializer(data, i)
+}
+
+// {name}{type specific payload}
+func customTypeReader(data *[]byte, i *int) (interface{}, error) {
+	// we need to decrement the index by 1 to be read the 32-bit int with the size of the string
+	*i = *i - 1
+	customTypeName, err := readString(data, i)
+	if err != nil {
+		return nil, err
+	}
+
+	// grab a read lock for the map of deserializers, since these can be updated out-of-band
+	customTypeReaderLock.RLock()
+	defer customTypeReaderLock.RUnlock()
+	deserializer, ok := customDeserializers[customTypeName.(string)]
+	if !ok {
+		return nil, newError(err0409GetSerializerToReadUnknownCustomTypeError, customTypeName)
 	}
 	return deserializer(data, i)
 }
--- a/gremlin-go/driver/serializer.go
+++ b/gremlin-go/driver/serializer.go
@@ -25,6 +25,7 @@ import (
 	"math/big"
 	"reflect"
 	"strings"
+	"sync"
 
 	"github.com/google/uuid"
 )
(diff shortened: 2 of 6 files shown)


gremlin-go/driver/graphBinary.go
```diff
@@ -1282,6 +1284,78 @@ func metricsReader(data *[]byte, i *int) (interface{}, error) {
 	return metrics, nil
 }
 
+type JanusgraphRelationIdentifier struct {
```
j-white, line 1287: This will be moved outside of the driver, but including here for an example of what an implementation could look like.

gremlin-go/driver/graphBinary.go
```diff
@@ -1282,6 +1284,78 @@ func metricsReader(data *[]byte, i *int) (interface{}, error) {
...
+
+// {name}{type specific payload}
+func customTypeReader(data *[]byte, i *int) (interface{}, error) {
+	// type name
+	*i = *i - 1
+	customTypeName, err := readString(data, i)
+	if err != nil {
+		return nil, err
+	}
+	deserializer, ok := customDeserializers[customTypeName.(string)]
```
j-white, line 1352: For custom types, it looks like we write the name of the type, and then the remaining bytes are specific to the implementation. So it could make sense to key off the name, and let the custom reader handle the rest.

Does this make sense?

gremlin-go/driver/graphBinary.go
```diff
@@ -1282,6 +1284,78 @@ func metricsReader(data *[]byte, i *int) (interface{}, error) {
...
+		r.InVertexIdString = vertexId.(string)
+	}
+
+	return r, nil
+}
+
+// {name}{type specific payload}
+func customTypeReader(data *[]byte, i *int) (interface{}, error) {
+	// type name
+	*i = *i - 1
```
j-white, line 1347: Need to backup the index by 1 to be read the 32-bit int with the size of the string

gremlin-go/driver/graphBinary.go
```diff
@@ -1282,6 +1284,78 @@ func metricsReader(data *[]byte, i *int) (interface{}, error) {
...
+	customTypeName, err := readString(data, i)
+	if err != nil {
+		return nil, err
+	}
+	deserializer, ok := customDeserializers[customTypeName.(string)]
+	if !ok {
+		return nil, newError(err0408GetSerializerToReadUnknownTypeError, 0x00)
```
j-white, line 1354: Introduce a new error code so we can pass the type name as a string? (this error code expects a number)

gremlin-go/driver/serializer.go
```diff
@@ -320,5 +321,12 @@ func initDeserializers() {
 		// Metrics
 		metricsType:          metricsReader,
 		traversalMetricsType: traversalMetricsReader,
+
+		// Customer
+		customType: customTypeReader,
+	}
+
+	customDeserializers = map[string]reader{
```
j-white, line 329: We could an option to provide this map during initialization. Would it make sense to add a property to the `DriverRemoteConnectionSettings` struct?

gremlin-go/driver/graphBinary.go
```diff
@@ -224,11 +225,12 @@ func instructionWriter(instructions []instruction, buffer *bytes.Buffer, typeSer
 
 // Format: {steps_length}{step_0}…{step_n}{sources_length}{source_0}…{source_n}
 // Where:
-//		{steps_length} is an Int value describing the amount of steps.
-//		{step_i} is composed of {name}{values_length}{value_0}…{value_n}, where:
-//      {name} is a String. This is also known as the operator.
-//		{values_length} is an Int describing the amount values.
-//		{value_i} is a fully qualified typed value composed of {type_code}{type_info}{value_flag}{value} describing the step argument.
+//
```
j-white, line 228: unchanged, just go fmt

gremlin-go/driver/serializer.go
vkagamlyk: and to `ClientSettings`

gremlin-go/driver/graphBinary.go
vkagamlyk: there is example of implementation of custom serializer for Java, looks like a good starting point
https://github.com/apache/tinkerpop/blob/master/gremlin-util/src/test/java/org/apache/tinkerpop/gremlin/util/ser/binary/types/sample/SamplePersonSerializer.java

gremlin-go/driver/serializer.go
vkagamlyk: or alternatively, you may add something like `CustomTypeSerializerRegistry` to store serializers for all custom types

j-white:
thanks for the prompt feedback @vkagamlyk, I'll work on the next iteration

gremlin-go/driver/serializer.go
```diff
@@ -320,5 +328,23 @@ func initDeserializers() {
...
 		traversalMetricsType: traversalMetricsReader,
+
+		// Customer
+		customType: customTypeReader,
 	}
+	customDeserializers = map[string]CustomTypeReader{}
+}
+
+// RegisterCustomTypeReader register a reader (deserializer) for a custom type
+func RegisterCustomTypeReader(customTypeName string, reader CustomTypeReader) {
```
j-white, line 339: I opted to allow registering custom types here rather than passing them in through the `DriverRemoteConnectionSettings` or `ClientSettings` since `readFullyQualifiedNullable` does not have any of the client context and bringing this along introduces a lot of code changes.

Registering them in this fashion while being careful about concurrency on the map seems to be sufficient.

vkagamlyk:
In what version would you like to see support for custom types serializers? 3.5.x, 3.6.x or 3.7.x?

vkagamlyk:
> > In what version would you like to see support for custom types serializers? 3.5.x, 3.6.x or 3.7.x?
> 
> Would be nice to have in 3.6.x+ for our use case

Then need to rebase and open PR to branch 3.6-dev.
Also good to drop a line with short description to [changelog](https://github.com/apache/tinkerpop/blob/3.6-dev/CHANGELOG.asciidoc)

Other than that LGTM, VOTE+1

j-white:
> Then need to rebase and open PR to branch 3.6-dev. Also good to drop a line with short description to [changelog](https://github.com/apache/tinkerpop/blob/3.6-dev/CHANGELOG.asciidoc)

Done

prahaladd (commented):
Thank you for this @j-white . I was working on a similar set of changes to resolve this ticket - the approach was conceptually similar but involved using Interfaces and embedded structs. This implementation is much simpler and cleaner and I would go for this.
I am actually thinking of using this change in order to help me progress on my other project https://pkg.go.dev/github.com/prahaladd/gograph which aims to provide a simple no-frills API surface unification across different types of graph databases (with the lofty aim of unifying the gremlin and cypher worlds :) )

xiazcy:
LGTM, thanks for the contribution! VOTE +1

PR #2176, opened 2023-07-31
TINKERPOP-2976 Fix modification while iterating

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2976

In rare cases users could get an `InvalidOperationException` if the dictionary of `_serializerByType` was modified while another thread in parallel iterated over the keys of this dictionary. Just creating a copy of the supported types and then iterating over that should fix this.
The problem was reported for GraphBinary, but the GraphSONWriter had the same problem so I also fixed it there.

VOTE +1

Since we are currently in code freeze, we can simply merge this after that has been lifted.

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphBinary/TypeSerializerRegistry.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphBinary/TypeSerializerRegistry.cs
@@ -159,11 +159,11 @@ namespace Gremlin.Net.Structure.IO.GraphBinary
         /// <exception cref="InvalidOperationException">Thrown when no serializer can be found for the type.</exception>
         public ITypeSerializer GetSerializerFor(Type valueType)
         {
-            if (_serializerByType.ContainsKey(valueType))
+            if (_serializerByType.TryGetValue(valueType, out var serializerForType))
             {
-                return _serializerByType[valueType];
+                return serializerForType;
             }
-            
+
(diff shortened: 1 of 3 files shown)


gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphBinary/TypeSerializerRegistry.cs
```diff
@@ -199,7 +199,7 @@ public ITypeSerializer GetSerializerFor(Type valueType)
                 return serializer;
             }
 
-            foreach (var supportedType in _serializerByType.Keys)
+            foreach (var supportedType in new List<Type>(_serializerByType.Keys))
```
vkagamlyk, line 202: What if serializer removed when iteration already started?

FlorianHockmann: Removing a serializer isn't really supported so I don't see how that could happen.
Do you see a good use case where someone would actually want to remove a serializer?

vkagamlyk: yes, no deletion

vkagamlyk:
VOTE +1

xiazcy:
VOTE +1, should we have a changelog entry for this?

PR #2206, opened 2023-08-15
TINKERPOP-2982 Allowed gremlin-driver to work over HTTP.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2982

HTTP is just an option of experimental nature for now. Websockets is still intact. There shouldn't be any breaking changes here. While there were some server changes, compatibility should be retained there...just added some extra parameters that were formerly only needed with TraversalOpProcessor. g.tx() and sessions will not work with this configuration.

VOTE +1


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/HttpGremlinResponseDecoder.java
```diff
@@ -0,0 +1,51 @@
...
+    private final MessageSerializer<?> serializer;
+
+    public HttpGremlinResponseDecoder(final MessageSerializer<?> serializer) {
+        this.serializer = serializer;
+    }
+
+    @Override
+    protected void decode(final ChannelHandlerContext channelHandlerContext, final FullHttpResponse httpResponse, final List<Object> objects) throws Exception {
+        final String content = httpResponse.content().toString(CharsetUtil.UTF_8);
+        objects.add(((MessageTextSerializer) serializer).deserializeResponse(content));
```
phanindhra876, line 49: Wont this throw error if serializer is a non-text serializer?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpGremlinEndpointHandler.java
```diff
@@ -147,40 +136,43 @@ public void channelRead(final ChannelHandlerContext ctx, final Object msg) {
...
             } catch (IllegalArgumentException iae) {
                 HttpHandlerUtil.sendError(ctx, BAD_REQUEST, iae.getMessage(), keepAlive);
                 ReferenceCountUtil.release(msg);
                 return;
             }
 
-            final String acceptString = Optional.ofNullable(req.headers().get("Accept")).orElse("application/json");
-            final Pair<String, MessageTextSerializer<?>> serializer = chooseSerializer(acceptString);
+            final String acceptMime = Optional.ofNullable(req.headers().get(HttpHeaderNames.ACCEPT)).orElse("application/json");
+            final Pair<String, MessageTextSerializer<?>> serializer = chooseSerializer(acceptMime);
```
phanindhra876, line 149: This is not exactly related to PR. However, shouldn't we need to drop the assumption that HTTP server can support only `MessageTextSerializer`?

phanindhra876 (changes requested):

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Host.java
```diff
@@ -49,7 +49,8 @@ public final class Host {
     Host(final InetSocketAddress address, final Cluster cluster) {
         this.cluster = cluster;
         this.address = address;
-        this.hostUri = makeUriFromAddress(address, cluster.getPath(), cluster.connectionPoolSettings().enableSsl);
+        this.hostUri = makeUriFromAddress(address, cluster.getPath(), cluster.connectionPoolSettings().enableSsl,
```
rdtr, line 52: Suggestion: It'd be great if the driver can allow a user to send a request with custom path and custom request header.

It may give too much freedom but at the same time, when graph providers offer additional http endpoints an user can still use the driver to call such endpoints. For example, Neptune provides its own explain / profile endpoint 
https://docs.aws.amazon.com/neptune/latest/userguide/gremlin-profile-api.html

It'd be great if we can support it at driver level so that users don't need to switch clients.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/HttpGremlinRequestEncoder.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.function.UnaryOperator;
+
+/**
+ * Converts {@link RequestMessage} to a {@code HttpRequest}.
+ */
+@ChannelHandler.Sharable
+public final class HttpGremlinRequestEncoder extends MessageToMessageEncoder<RequestMessage> {
+    private final MessageSerializer<?> serializer;
+
+    private final ObjectMapper mapper = new ObjectMapper();
```
rdtr, line 50: Do we need to have an instance of ObjectMapper for each handler ? I believe creating a mapper is relatively expensive operation, thus if we can make it `static` it'd be better.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpGremlinEndpointHandler.java
spmallette: yes, i think my previous comment addresses that. `MessageTextSerializer` really isn't a necessary anymore.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Host.java
spmallette: yes, i often dislike that the Java driver works the way that it does, at least insofar as how it configures host/port/scheme. like you have to do:

```groovy
g = traversal().withRemote(
                DriverRemoteConnection.using("localhost", 8182))
```

but would be nice if you could do like the other drivers:

```groovy
g = traversal().withRemote(
                new DriverRemoteConnection('ws://localhost:8182/gremlin'))
```

to provide a bit more flexibility for providers and how they setup their servers. that said, imo, i do think Neptune has gone a little too far in creating its own explain/profile endpoints. it should simply implement the `explain()` / `profile()` steps.  For what it's worth, you can use the interceptor interface to add a custom request header now.

Anyway, feel free to open a JIRA for allowing users to specify a url to a "gremlin endpoint" rather than our constructing one. And if you feel like we need to somehow support other custom endpoints, I think you should start a DISCUSS thread on the dev list.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
```diff
@@ -143,18 +152,18 @@ protected void initChannel(final SocketChannel socketChannel) throws Exception {
     /**
      * WebSocket {@link Channelizer} implementation.
      */
-    public final class WebSocketChannelizer extends AbstractChannelizer {
+    final class WebSocketChannelizer extends AbstractChannelizer {
```
Cole-Greer, line 155: Small breaking change here. I'm not aware of any users which would be impacted by this but I would prefer to defer such a change to the next major release.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpGremlinEndpointHandler.java
```diff
@@ -49,33 +50,21 @@
...
 import javax.script.SimpleBindings;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.UUID;
+import java.util.*;
```
Cole-Greer, line 59: There are a couple of wildcard imports introduced throughout the PR, would prefer to avoid them.

Cole-Greer (commented):
Thanks Stephen. LGTM. I just left a few minor comments above.

I'm also wondering if the changes here warrant any additional documentation?

Am I correct in understanding that users would only need to configure the driver to use the new HTTPChannelizer in order to connect to an HTTP only gremlin server? Perhaps some reference docs which show how to use this would be worthwhile.

VOTE +1 (Pending resolution of the above)

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Host.java
rdtr: Got it, yeah for explain and profile, supporting that on Neptune side and the driver should just add `explain()` or `profile()` in a query, I agree.
It's too much to support more other endpoints like /status, bulkload for Neptune so I don't think at this point adding a custom endpoint is critical at least for Neptune.

spmallette:
> I'm also wondering if the changes here warrant any additional documentation?

not yet from my perspective. this was mostly experimental to see how it could work. i think doing the same for python might validate the idea further, but that can be a separate PR. i've adjusted the changelog/javadoc accordingly.

> Am I correct in understanding that users would only need to configure the driver to use the new HTTPChannelizer in order to connect to an HTTP only gremlin server? 

yes

PR #2212, opened 2023-08-20
TINKERPOP-2986: [StarGraph] Drop edge properties when dropping edges

li-boxuan:
SparkGraphComputer uses StarGraph to represent an in-memory graph. EdgeFilter can be configured by users to drop unneeded edges, but the edge properties aren't dropped. This fixes this bug and greatly reduces memory footprint.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/star/StarGraph.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/star/StarGraph.java
@@ -360,12 +360,20 @@ public final class StarGraph implements Graph, Serializable {
             super(id, label);
         }
 
+        private void dropEdgeProperty(Object id) {
+            if (edgeProperties != null) {
+                edgeProperties.remove(id);
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


Cole-Greer:
Thanks @li-boxuan, LGTM

VOTE +1

vkagamlyk:
are these changes only needed in 3.7?

if so then VOTE +1

li-boxuan:
> are these changes only needed in 3.7?
> 
> if so then VOTE +1

I am not sure which branch I should target, so I set master at the moment. Is 3.5 branch still being supported? If so, I could retarget this PR to 3.5 branch since it is technically a bug fix.

kenhuuu:
> > are these changes only needed in 3.7?
> > if so then VOTE +1
> 
> I am not sure which branch I should target, so I set master at the moment. Is 3.5 branch still being supported? If so, I could retarget this PR to 3.5 branch since it is technically a bug fix.

It seems like 3.5-dev is being phased out, so this can probably target 3.6-dev.

li-boxuan:
Retargeted 3.6-dev

PR #2216, opened 2023-08-22
[TINKERPOP-2972] ProjectStep and SelectStep do not allow a duplicate key

rdtr:
Having duplicate keys does not make sense and a user won't use this feature (correct me otherwise). But due to the round-robin nature, determining which traversal should be mapped to the key is complicated in terms of implementation.

So we should throw fast so that we don't have to deal with duplicate keys.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,7 +25,9 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 This release also includes changes from <<release-3-5-8, 3.5.8>>.
 
+==== Improvements
 
+* TINKERPOP-2972 Throw an Exception when a duplicate key is supplied for ProjectStep and SelectStep
 
 [[release-3-6-5]]
 === TinkerPop 3.6.5 (Release Date: July 31, 2023)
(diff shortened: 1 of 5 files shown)


rdtr:
I see this error in a build. How should I resolve it ?

```
gremlin-javascript: Error occurred while calling a method on a Groovy class from classpath. InvocationTargetException: javax.script.ScriptException: java.lang.IllegalArgumentException: keys must be unique in SelectStep -> [Help 1]
```

xiazcy:
> I see this error in a build. How should I resolve it ?
> 
> ```
> gremlin-javascript: Error occurred while calling a method on a Groovy class from classpath. InvocationTargetException: javax.script.ScriptException: java.lang.IllegalArgumentException: keys must be unique in SelectStep -> [Help 1]
> ```

This looks like the build is failing when trying to generate the Gherkin Feature test traversals file for javascript through `gremlin-javascript/build/generate.groovy`, which looks like is due to a Select Feature test case that is now failing the new error condition introduced in this PR:
https://github.com/apache/tinkerpop/blob/e0779f768289a5e436523c7c6fd7ac1b0f660c37/gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/Select.feature#L895

You might need to look into the test, which is presenting a potential use case, and see if it needs to be changed with this new behaviour you are introducing. 

This also ties into a general comment, I'd recommend adding Gherkin Feature tests for these step behaviour changes in addition to unit tests. An example for asserting errors in Feature tests is here: https://github.com/apache/tinkerpop/blob/e0779f768289a5e436523c7c6fd7ac1b0f660c37/gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/MergeVertex.feature#L969

I usually run these Feature tests through TinkerGraph locally with this test, which runs all the steps: 
`tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/TinkerGraphFeatureTest.java`

rdtr:
Thanks @xiazcy ! I will fix these tests

PR #2217, opened 2023-08-23
[TINKERPOP-2972] ProjectStep does not allow keys specified in a query

rdtr:
ProjectStep should not accept a duplicate key, anyway doing so would not make much sense while the implementation for Graph provider would be complicated because of the logic to determine what is the actual traversal to be mapped to the key.


Cole-Greer:
Thanks for putting this together. Overall the change looks good to me.

If possible it would be nice to have a feature test for this just to ensure there are no issues in the GLV's. A good example of feature tests catching errors can be found here: https://github.com/apache/tinkerpop/blob/397eff08f73ae555bd3aab87498b9ff131792741/gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/MergeVertex.feature#L769

kenhuuu:
VOTE +1

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/Project.feature
```diff
@@ -106,4 +106,13 @@ Feature: Step - project()
       | m[{"a":"d[0].l", "b":"d[29].i"}] |
       | m[{"a":"d[1].l", "b":"d[27].i"}] |
       | m[{"a":"d[1].l", "b":"d[32].i"}] |
-      | m[{"a":"d[0].l", "b":"d[35].i"}] |
\ No newline at end of file
+      | m[{"a":"d[0].l", "b":"d[35].i"}] |
+
+
+  Scenario: g_V_projectXa_aX
```
xiazcy, line 112: Looks like the Gherkin tests can't handle cases that fast fail on step construction (I believe it needs the step to be successfully constructed before translating it into the GLV tests). I don't think there is any easy work around currently, sorry @rdtr, you might need to remove this test and just have unit tests.

(As an aside, the test is missing a `When iterated to list`, but that's not really important now that this test won't really work in Gherkin).

xiazcy: Actually there is a quick workaround I just remembered. You can technically add static translation of this test to all the feature tests in the `generate.groovy` file of each GLV, but it's ultimately a hack so I'd think it's ok to not include it at this point.

Cole-Greer: Sorry I missed that this scenario is incompatible with the feature tests.

VOTE +1 (with failing feature test removed)

rdtr: Ok, I deleted the feature test. Please check when you have time, thanks !

PR #2227, opened 2023-08-29
Implement trim(), lTrim(), rTrim(), and reverse() String Functions

xiazcy:
Another set of string function implementations from the [proposal](https://github.com/apache/tinkerpop/blob/master/docs/src/dev/future/proposal-3-remove-closures.asciidoc#string-manipulation-functions-in-tinkerpop). Also not considering scopes at this point.


gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/LTrim.feature
```diff
@@ -0,0 +1,87 @@
...
+    And the traversal of
+    """
+    g.inject(xx1).lTrim()
+    """
+    When iterated to list
+    Then the traversal will raise an error with message containing text of "The lTrim() step can only take string as argument"
+
+  Scenario: g_V_valuesXnameX_ltrim_concatXleft_trimX
+    Given the empty graph
+    And the graph initializer of
```
vkagamlyk, line 60: why not to use modern graph?

vkagamlyk:
VOTE +1

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/LTrim.feature
xiazcy: I wanted to demonstrate trimming of properties with existing leading and trailing spaces, only the empty graph is reset, all other graphs are shared so we shouldn't modify them

vkagamlyk: ahh, it's modified modern graph
sorry, I missed

kenhuuu:
VOTE +1. Consider adding tests for empty string here as well.

PR #2241, opened 2023-09-15
TINKERPOP-2994 Fixed mergeV/E usage with PartitionStrategy

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2994

Neither PartitionStrategy nor mergeV/E were properly setup to work together so the partition wasn't being respected with those steps.

VOTE +1


gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/integrated/PartitionStrategy.feature
```diff
@@ -0,0 +1,458 @@
...
+    And the graph initializer of
+      """
+      g.addV("person").property("_partition","a").property("name","alice").
+        addV("person").property("_partition","b").property("name","alice")
+      """
+    And using the parameter xx1 defined as "m[{\"t[label]\": \"person\", \"name\":\"alice\"}]"
+    And using the parameter xx2 defined as "m[{\"name\":\"bob\"}]"
+    And the traversal of
+      """
+      g.withStrategies(new PartitionStrategy(partitionKey: "_partition", writePartition: "a", readPartitions: ["a"])).
```
vkagamlyk, line 413: How `mergeV` will work when readPartitions not contains writePartition? 
something like
`g.withStrategies(new PartitionStrategy(partitionKey: "_partition", writePartition: "b", readPartitions: ["a"])).
        mergeV(xx1).option(Merge.onMatch, xx2)`

spmallette: it appears that if you do not specify that it will write a null for the partition. a `null` for graphs that don't support storage of it would be effectively ignored:

```text
gremlin> g.addV().property('k',null)
==>v[0]
gremlin> g.V().elementMap()
==>[id:0,label:vertex]
```

so, the vertex would end up in no partition at all. for graphs that support null i guess it would write the partition as a `null` which i guess would then treat null as a partition value.

spmallette: oh - sorry - i misread what you wrote.  in the case you describe it's like a write only behavior. you can write stuff that you can't read. so for mergeV it would just always insert.

vkagamlyk:
VOTE +1

PR #2254, opened 2023-09-24
[TINKERPOP-2423] Do not use XOR in hashCode computation when redundant key is possible

rdtr:
We may want to do away with XOR for hashCode computation for everything, but for now I only limit those whose keys for XOR are very simple and can be duplicate.
Duplicate keys are canceled with each other when XOR is applied, that is the problem and a different Step can be considered as "equal" as consequence.

This CR fixes the issue.


vkagamlyk:
VOTE +1

xiazcy:
VOTE +1

kenhuuu:
Hi @rdtr, this PR is ready to merge but there is a small conflict with the CHANGELOG. Could you update this PR? Alternatively, if you are too busy, I can cherry-pick your change and update it.

PR #2282, opened 2023-10-12
[TINKERPOP-3004] improve GremlinValueComparator performance

vkagamlyk:
There are 3 reasons for the reduced performance of element comparison compared to 3.5
1. Relatively fast [`BulkSet.Contains`](https://github.com/apache/tinkerpop/blob/ea0bbed85a22ffefc8fa599e9d50ee68d60162fd/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Contains.java#L57) replaced with slow iteration and comparison one by one object. Most expensive operation here is building and using of Iterator over incoming `BulkSet`
2. Equality comparison uses [lots of type checking](https://github.com/apache/tinkerpop/blob/32ec7f97c0464cb015bdc7cf766a94e6fbb71210/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/GremlinValueComparator.java#L127). 
3. Equality comparison uses [reflection](https://github.com/apache/tinkerpop/blob/32ec7f97c0464cb015bdc7cf766a94e6fbb71210/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/GremlinValueComparator.java#L142). 

I used test query included in this PR.
Total test execution time for
vanilla 3.7 is 2700ms
with my change is 1000ms
vanilla 3.5 is 200ms

Main impact on execution time is 
Iterator adds ~200 ms
`Compare.eq.test(first, o)` is ~ 2000 ms

I also tried to verify all objects types and than reuse `BulkSet.contains`, execution time was ~550ms, but this doesn't work for all types


Cole-Greer:
Thanks Valentyn, it's always nice to see big gains in performance. Hopefully we can get back closer to 3.5 performance at some point but this is a big step forward.

VOTE +1

xiazcy:
VOTE +1

vkagamlyk:
VOTE+1

PR #2283, opened 2023-10-13
[TINKERPOP-3006] Enable remote connection over HTTP for gremlin python

xiazcy:
https://issues.apache.org/jira/browse/TINKERPOP-3006

Adding experimental feature for python to connect via HTTP. This is relying on the python translator, for which additional types were added. There shouldn't be any breaking changes.

Connecting to remote can be done via
`remote_conn = DriverRemoteConnection('http://localhost':8182/, 'g')`
and for traversals
`g = traversal().with_remote(remote_conn)`

One note, in addition to the integration test, I ran the feature test suite (radish tests) locally with the HTTP endpoint for 3.6 and all tests passed. One exception is that `g.inject(null, null)` test will fail due to combination of Groovy script issue that throws NPE with 2nd nulls and server HTTP handler hanging upon NPEs. 

VOTE +1

### DIFF
--- a/gremlin-python/src/main/python/gremlin_python/driver/aiohttp/transport.py
+++ b/gremlin-python/src/main/python/gremlin_python/driver/aiohttp/transport.py
@@ -138,3 +138,104 @@ class AiohttpTransport(AbstractBaseTransport):
     def closed(self):
         # Connection is closed if either the websocket or the client session is closed.
         return self._websocket.closed or self._client_session.closed
+
+
+class AiohttpHTTPTransport(AbstractBaseTransport):
+    nest_asyncio_applied = False
+
+    def __init__(self, call_from_event_loop=None, read_timeout=None, write_timeout=None, **kwargs):
+        if call_from_event_loop is not None and call_from_event_loop and not AiohttpTransport.nest_asyncio_applied:
+            """ 
+                The AiohttpTransport implementation uses the asyncio event loop. Because of this, it cannot be called 
+                within an event loop without nest_asyncio. If the code is ever refactored so that it can be called 
+                within an event loop this import and call can be removed. Without this, applications which use the 
+                event loop to call gremlin-python (such as Jupyter) will not work.
+            """
+            import nest_asyncio
+            nest_asyncio.apply()
+            AiohttpTransport.nest_asyncio_applied = True
+
+        # Start event loop and initialize client session and response to None
+        self._loop = asyncio.new_event_loop()
+        self._client_session = None
+        self._http_req_resp = None
+        self._enable_ssl = False
+
+        # Set all inner variables to parameters passed in.
+        self._aiohttp_kwargs = kwargs
+        self._write_timeout = write_timeout
+        self._read_timeout = read_timeout
+        if "ssl_options" in self._aiohttp_kwargs:
+            self._ssl_context = self._aiohttp_kwargs.pop("ssl_options")
+            self._enable_ssl = True
+
+        # if "ssl_options" in self._aiohttp_kwargs:
+        #     self._aiohttp_kwargs["ssl_context"] = self._aiohttp_kwargs.pop("ssl_options")
+
+    def __del__(self):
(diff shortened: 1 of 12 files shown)


gremlin-python/src/main/python/tests/process/test_translator.py
```diff
@@ -373,6 +374,30 @@ class SuperStr(str):
         tests.append([g.V().has("p1", None),
                      "g.V().has('p1',null)"])
 
+        # 104
+        vertex = Vertex(0, "person")
```
vkagamlyk, line 378: Could you add similar test for `vertex = Vertex(0, "'person'")` to confirm correct work with strings?

vkagamlyk (commented):
great job, I think this is enough for experimental implementation

VOTE +1

gremlin-python/src/main/python/gremlin_python/driver/aiohttp/transport.py
```diff
@@ -138,3 +138,104 @@ async def async_close():
...
+        # Set all inner variables to parameters passed in.
+        self._aiohttp_kwargs = kwargs
+        self._write_timeout = write_timeout
+        self._read_timeout = read_timeout
+        if "ssl_options" in self._aiohttp_kwargs:
+            self._ssl_context = self._aiohttp_kwargs.pop("ssl_options")
+            self._enable_ssl = True
+
+        # if "ssl_options" in self._aiohttp_kwargs:
+        #     self._aiohttp_kwargs["ssl_context"] = self._aiohttp_kwargs.pop("ssl_options")
```
kenhuuu, line 173: Intentional?

gremlin-python/src/main/python/gremlin_python/driver/client.py
```diff
@@ -45,15 +46,20 @@ def __init__(self, url, traversal_source, protocol_factory=None,
...
         self._url = url
         self._headers = headers
         self._enable_user_agent_on_connect = enable_user_agent_on_connect
         self._traversal_source = traversal_source
-        if "max_content_length" not in transport_kwargs:
+        if not self._use_http and "max_content_length" not in transport_kwargs:
             transport_kwargs["max_content_length"] = 10 * 1024 * 1024
         if message_serializer is None:
             message_serializer = serializer.GraphBinarySerializersV1()
+            # message_serializer = serializer.GraphSONMessageSerializer()
```
kenhuuu, line 62: Intentional?

kenhuuu:
client and protocol have very similar code in lots of instances. I would consider refactoring by either adding in another class in-between to share functionality or move that shared code into functions in the parent class.

gremlin-python/src/main/python/tests/process/test_translator.py
xiazcy: Hmm, ok. Is that a use case to crate a label (that's a string) with a single quote? Or it's just to test?

gremlin-python/src/main/python/gremlin_python/driver/aiohttp/transport.py
xiazcy: Ah yes, I forgot to remove this part.

gremlin-python/src/main/python/gremlin_python/driver/client.py
xiazcy: Also something I forgot to remove, thanks!

gremlin-python/src/main/python/tests/process/test_translator.py
vkagamlyk: create label with single quote to verify serialization

Cole-Greer:
Thanks Yang, LGTM. It's nice to see some additions to the Python translator included here as well.

VOTE +1

gremlin-python/src/main/python/tests/process/test_translator.py
xiazcy: Updated to using repr() for string formatting for now.

PR #2284, opened 2023-10-13
Fix Java driver HTTP connection request not encoding all args

xiazcy:
I noticed that the Java driver wasn't encoding the remaining args into the payload, such as aliases for traversal source were not registered. Not sure if it was intentional, or if the type of args encoded should be limited instead of having all of them, so I'm adding this quick fix as a PR.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/HttpGremlinRequestEncoder.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/HttpGremlinRequestEncoder.java
@@ -71,6 +71,7 @@ public final class HttpGremlinRequestEncoder extends MessageToMessageEncoder<Req
                 gremlin = gremlinStringOrBytecode.toString();
             }
             final byte[] payload = mapper.writeValueAsBytes(new HashMap<String,Object>() {{
+                this.putAll(requestMessage.getArgs());
                 put(Tokens.ARGS_GREMLIN, gremlin);
                 put(Tokens.REQUEST_ID, requestMessage.getRequestId());
                 if (usesBytecode) put("op", Tokens.OPS_BYTECODE);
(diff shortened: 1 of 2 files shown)


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/HttpGremlinRequestEncoder.java
```diff
@@ -71,6 +71,7 @@ protected void encode(final ChannelHandlerContext channelHandlerContext, final R
                 gremlin = gremlinStringOrBytecode.toString();
             }
             final byte[] payload = mapper.writeValueAsBytes(new HashMap<String,Object>() {{
+                this.putAll(requestMessage.getArgs());
                 put(Tokens.ARGS_GREMLIN, gremlin);
                 put(Tokens.REQUEST_ID, requestMessage.getRequestId());
```
Cole-Greer, line 76: Is this redundant given the above putAll()?

vkagamlyk: request id is not part of Args.

It would be nice to add a test that will confirm that the bindings work correctly

kenhuuu:
I'm not sure about the consequence of making this change but it seems to be needed to have parity with the WebSocket implementation so my VOTE is +1.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/HttpGremlinRequestEncoder.java
xiazcy: The `Tokens.ARGS_GREMLIN` here replaces the `gremlin` stuff in the original request with the processed ones above, so it'd be needed (this was the easier way)

vkagamlyk:
VOTE +1

Cole-Greer:
VOTE +1

xiazcy:
This fix may or may not be relevant anymore, will wait on the bytecode over HTTP work.

Cole-Greer:
My understanding is that this PR is no longer relevant as of https://github.com/apache/tinkerpop/pull/2336 and should now be closed. Any thoughts?

vkagamlyk:
> My understanding is that this PR is no longer relevant as of #2336 and should now be closed. Any thoughts?

yes, this PR is no longer valid

PR #2293, opened 2023-10-16
TINKERPOP-2811 Fixed bug with hasId() order for ElementIdStrategy

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2811

Since HasContainer rolls up sequential has() steps checking the first item in the HasContainer only might not catch the id that needs to be replaced.

VOTE +1


vkagamlyk:
VOTE+1

Cole-Greer:
LGTM, VOTE +1

PR #2295, opened 2023-10-17
Upgrade node to 18 in docs/gremlint

ryn5:


Cole-Greer:
LGTM thanks Ryan, merging as CTR. VOTE +1

FlorianHockmann:
The smoke test is currently failing on `master` and the error message sounds to me like it could be related to this PR:
```
Error:  The archive file /home/runner/.m2/repository/com/github/eirslett/node/18.17.0/node-18.17.0-linux-x64.tar.gz is corrupted and will be deleted. Please try the build again.
```

although the check seems to have passed for the commit from this PR. Any idea what could be the problem here?

All other commits after this PR was merged are just Dependabot .NET updates so I don't think that they could have caused this.

Cole-Greer:
I don't think this PR was the cause of this error. It's a random failure I've seen before although the frequency has grown to an alarming level. I just pulled this error out of a failed run on 3.6-dev from a few hours ago:
```
[INFO] Unpacking /home/runner/.m2/repository/com/github/eirslett/node/16.20.0/node-16.20.0-linux-x64.tar.gz into /home/runner/work/tinkerpop/tinkerpop/gremlin-javascript/src/main/javascript/gremlin-javascript/node/tmp
Error:  The archive file /home/runner/.m2/repository/com/github/eirslett/node/16.20.0/node-16.20.0-linux-x64.tar.gz is corrupted and will be deleted. Please try the build again.
```
This PR is just updating the node version for docs/gremlint, which as far as I know is not tested in our CI at the moment. It appears that this error is originating from gremlin-javascript. I think more investigation is needed here to find out how to prevent the node download from corrupting so frequently.

xiazcy:
I would also think this is just a burst of nodejs server issue. I just re-ran the actions for 3.6 and master they all past.

FlorianHockmann:
Ah ok, thanks for looking into this, @Cole-Greer and @xiazcy! It just seemed logical that there's a connection between the error message and a commit updating node right before the failing builds 😅

PR #2296, opened 2023-10-17
Change substring() to use `startIndex, endIndex` instead of `startIndex, length`

ryn5:
Substring step was changed to follow implementation of python string slicing instead of SQL's SUBSTRING(), added step and feature tests, and updated docs to reflect change.


vkagamlyk:
VOTE +1

Cole-Greer:
LGTM, VOTE +1

xiazcy:
VOTE +1, thanks for the update!

PR #2297, opened 2023-10-17
TINKERPOP-2999 3.7.0 Remote Console Sends Incomplete Queries

ryn5:
https://issues.apache.org/jira/browse/TINKERPOP-2999
Fixes issue where multiline commands would throw error on the first line, caused by upgrade to Groovy 4.  Handling pulled from Groovysh.groovy

### DIFF
--- a/gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/GremlinGroovysh.groovy
+++ b/gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/GremlinGroovysh.groovy
@@ -24,8 +24,10 @@ import org.apache.groovy.groovysh.Groovysh
 import org.apache.groovy.groovysh.ParseCode
 import org.apache.groovy.groovysh.Parser
 import org.apache.groovy.groovysh.util.CommandArgumentParser
+import org.apache.groovy.groovysh.util.ScriptVariableAnalyzer
 import org.apache.tinkerpop.gremlin.console.commands.GremlinSetCommand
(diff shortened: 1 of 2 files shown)


gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/GremlinGroovysh.groovy
```diff
@@ -140,4 +171,48 @@ class GremlinGroovysh extends Groovysh {
...
+            })
+        }
+        // Evaluate the current buffer w/imports and dummy statement
+        List<String> buff
+        if (variableBlocks) {
+            buff = [importsSpec] + ['try {', 'true'] + current + ['} finally {' + variableBlocks + '}']
+        } else {
+            buff = [importsSpec] + ['true'] + current
+        }
+        setLastResult(result = interp.evaluate(buff))
```
kenhuuu, line 198: I'm a bit concerned here that we just do setLastResult() to whatever is in the buffer here. Notice how for remote console we normally do this with something like `mediator.currentRemote().submit([script])`. Have you checked to see if this is a problem?

ryn5: Confirmed that multiline queries are fixed for both interpreter modes

Cole-Greer:
Thanks Ryan, LGTM.

VOTE +1

kenhuuu:
VOTE +1

xiazcy:
VOTE +1

PR #2299, opened 2023-10-17
TINKERPOP-2995 Create Sample Applications in each GLV 3.6

ryn5:
https://issues.apache.org/jira/browse/TINKERPOP-2995
Created sample applications for each GLV in Java, Python, C#, JS, and Go which includes connection, basic Gremlin, and simple traversal examples.


gremlin-dotnet/example/example.csproj
```diff
@@ -0,0 +1,14 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net7.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Gremlin.Net" Version="3.7.0" />
```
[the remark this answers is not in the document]
ryn5: Changed version to 3.6.5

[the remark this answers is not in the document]
Cole-Greer: There may be some trickery which can be done by configuring multiple instances of dependabot to run in each branch along with some rules to restrict the version in each branch. That would still require the committer to be careful when merging those commits up to other branches to preserve the correct version. If we consider it essential to have the example versions align with the latest release in each respective branch, this might be best left as an additional administrative task for release managers. It may be possible to use the Maven Replacer Plugin to automatically this version with every `mvn compile` but the same caution around merging between branches would apply.

Personally I'm somewhat indifferent to how/if the versions should be maintained here. The main value I see in these examples is demonstrating basic driver setup, connection, and querying. This has remained relatively constant over the last few major releases and I think that a 3.7.x example is still useful for someone attempting to use 3.6.x.

Cole-Greer:
LGTM, VOTE +1 (pending resolution to the version management discussion above)

gremlin-dotnet/example/example.csproj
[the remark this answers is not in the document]
ryn5: I'm thinking it might be fine to let Dependabot handle it since the examples are quite simple and would probably still work with future updates.  We could maybe implement tests that execute during the maven build at some point to catch breaking changes, which would avoid adding extra overhead for release managers.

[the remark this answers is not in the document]
FlorianHockmann: > Personally I'm somewhat indifferent to how/if the versions should be maintained here. The main value I see in these examples is demonstrating basic driver setup, connection, and querying. This has remained relatively constant over the last few major releases and I think that a 3.7.x example is still useful for someone attempting to use 3.6.x.

You're probably right and in addition to that, I'm not sure if it's worth any effort to maintain different examples for the different versions. Most users will just use the latest version any way. That's at least the case for .NET according to the NuGet download statistics.
So, from my side, we can stay on 3.7.0 here.

gremlin-python/src/main/python/examples/basic_gremlin.py
```diff
@@ -0,0 +1,54 @@
...
+    # iterate() does not return any data and is used to just generate side-effects (i.e. write data to the database)
+    g.V(v1).add_e('knows').to(v2).property('weight', 0.75).iterate()
+    g.V(v1).add_e('knows').to(v3).property('weight', 0.75).iterate()
+
+    # retrieve the data from the "marko" vertex
+    marko = g.V().has('person', 'name', 'marko').values('name').next()
+    print("name: " + marko)
+
+    # find the "marko" vertex and then traverse to the people he "knows" and return their data
+    people_marko_knows = g.V().has('person', 'name', 'marko').out('knows').values('name').toList()
```
joshperryman, line 46: should the terminating step be `to_list()`?

ryn5: Yes, thanks for the catch!

gremlin-driver/src/main/java/examples/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+    <build>
+        <sourceDirectory>.</sourceDirectory>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
```
kenhuuu, line 58: Why is shading necessary?

run_examples.sh
```diff
@@ -0,0 +1,105 @@
...
+#
+#Unless required by applicable law or agreed to in writing,
+#software distributed under the License is distributed on an
+#"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#KIND, either express or implied.  See the License for the
+#specific language governing permissions and limitations
+#under the License.
+
+open -a Docker
+docker pull tinkerpop/gremlin-server
```
kenhuuu, line 21: In general, you can't mix server/driver versions so this might need to explicitly specify the version.

run_examples.sh
```diff
@@ -0,0 +1,105 @@
...
+echo
+echo "Running Go examples:"
+cd gremlin-go/examples || exit
+go run connections.go
+go run basic_gremlin.go
+cd ../.. || exit
+
+echo
+echo -n "Shutting down container: "
+docker stop glv-examples
```
kenhuuu, line 67: Depending on whether or not something went wrong earlier or the user quit in the middle, this might not get called. Cleanup should be done in an EXIT trap sort like https://github.com/apache/tinkerpop/blob/master/docker/build.sh#L33

gremlin-driver/src/main/java/examples/pom.xml
ryn5: The shading is providing an uberjar to be used in running the Java examples via bash

gremlin-driver/src/main/java/examples/dependency-reduced-pom.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.apache.tinkerpop</groupId>
+  <artifactId>run-examples</artifactId>
```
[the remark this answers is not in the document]
ryn5: It's being generated by the shading process to list reduced dependencies but doesn't seem to actually be needed.  Think I'll just remove it from Git.

run_examples.sh
ryn5: I initially did specify the version, but was thinking that since this script would mainly be used before releases just to check that the examples are working, it might save some work to update the version each release if it only pulled the latest.  Do you think I should add it back?

gremlin-driver/src/main/java/examples/pom.xml
kenhuuu: I see. For some reason I thought this was tied to a maven phase rather than producing a jar to run in a script, but this way is fine too, it would just be harder to wire into GHA if we were to do that in the future.

gremlin-driver/src/main/java/examples/dependency-reduced-pom.xml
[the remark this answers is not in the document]
Cole-Greer: I had the same question on the 3.7 PR. I just remembered that I ran into this problem in the past. I believe this is what you need:
```
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-shade-plugin</artifactId>
    <version>3.4.1</version>
    <configuration>
        <createDependencyReducedPom>false</createDependencyReducedPom>
    </configuration>
</plugin>
```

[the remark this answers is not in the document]
ryn5: Thanks!

gremlin-dotnet/Examples/BasicGremlin/BasicGremlin.csproj
```diff
@@ -0,0 +1,13 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net7.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Gremlin.Net" Version="3.7.0" />
```
vkagamlyk, line 11: nit: should be 3.6.x

vkagamlyk:
VOTE +1

PR #2306, opened 2023-10-23
[TINKERPOP-3008] Update concat() to accept traversal varargs 

xiazcy:
https://issues.apache.org/jira/browse/TINKERPOP-3008

This updates `concat(Traversal)` to accept traversal varags and also corrected how it handles inject() child traversals. 

The updated signature is `concat(Traversal, Traversal...)` to disambiguate from `concat(String...)` method call.

Also a small breaking change on how inject child traversals are handled. We mistakenly added special handling, but it is the default behaviour of `inject()` to return the incoming traverser, so we've corrected that:
```
// behavior of child traversal inject as aligned with its current definition & implementation
gremlin> g.inject("a").concat(inject("b"))
==>aa
// wrongly special cased in 3.7.0
gremlin> g.inject("a").concat(inject("b"))
==>ab
// proper way to add a string in concat
gremlin> g.inject("a").concat(constant("b"))
==>ab
```

VOTE +1


docs/src/upgrade/release-3.7.x.asciidoc
```diff
@@ -30,7 +30,32 @@ complete list of all the modifications that are part of this release.
...
+in arguments. The inject() step injects the incoming traverser as the first item in the stream by default. We mistakenly
+skipped it and used its child traversal value in 3.7.0 based on false expectation of how inject behaves. The
+behavior below is the expected outcome of inject in arguments.
+[source,text]
+----
+gremlin> g.inject("a").concat(inject())
+==>aa
+gremlin> g.inject("a").concat(inject("b"))
+==>aa
+----
```
Cole-Greer, line 56: I don't think this needs to focus too much on implementation details or motivation. I would suggest shortening it and focusing on describing the differences.
```suggestion
A notable breaking change from 3.7.0 is that we have output order of `inject()` as a child of `concat()` to be consistent with other parent steps. Any 3.7.0 uses of `concat(inject(X))` should change to `concat(constant(X))` to retain the old semantics.
[source,text]
----
// 3.7.0
gremlin> g.inject("a").concat(inject("b"))
==>ab

// 3.7.1
gremlin> g.inject("a").concat(inject())
==>aa
gremlin> g.inject("a").concat(inject("b"))
==>aa
gremlin> g.inject("a").concat(constant("b"))
==>ab
----
```

Cole-Greer (approved):
Thanks Yang, just a docs nit.

LGTM VOTE +1

kenhuuu:
VOTE +1

PR #2313, opened 2023-10-27
Update Dev and Future docs for 4.x development

Cole-Greer:
Updating docs now that `master` is set for 4.x development. I updated everything to the best of my knowledge, however others may want to give some input on the changes. I didn't get into what features should or should not be listed for 4.x in the future docs, I simply removed features which have already been implemented and released. Further revision of the future docs would be warranted following community consensus.


docs/src/dev/developer/for-committers.asciidoc
```diff
@@ -141,8 +141,8 @@ TinkerPop has several release branches:
 * `3.4-dev` - 3.4.x (no longer maintained)
 * `3.5-dev` - 3.5.x (non-breaking bug fixes and enhancements)
```
vkagamlyk, line 142: May be 3.5-dev only for critical fixes and dependency updates?

vkagamlyk:
VOTE +1

docs/src/dev/developer/for-committers.asciidoc
Cole-Greer: I do agree, however I didn't think that was within scope for this change. I think that should be discussed and updated after the release of 3.5.8.

kenhuuu:
There are actually a few items from Proposal 3 that aren't done for the list functions (like length, indexOf). Its probably OK to keep them there just as a design reference rather than deleting them.

Regardless of what we decide to do for the above, I'm VOTE +1 for this change.

kenhuuu:
Just to clarify, when I said

> Its probably OK to keep them there just as a design reference rather than deleting them.

I meant keeping all the proposals as references. We could just add an update to them to say they are done or something similar.

Cole-Greer:
> There are actually a few items from Proposal 3 that aren't done for the list functions (like length, indexOf). Its probably OK to keep them there just as a design reference rather than deleting them.
> 
> Regardless of what we decide to do for the above, I'm VOTE +1 for this change.

I don't have a strong feeling either way regarding the proposals. I have added them back for now.

PR #2314, opened 2023-10-29
Option to force the message to be sent as string instead of bytecode

divad1196:
Hi, here is a small PR that would allow to use the string version instead of the bytecode one.
May reason for that is that some databases, like CosmosDB, doesn't support the bytecode format for now (it was supposed to be ready in 2018...).

This simple option makes most of the job to allow the use of the fluent API with these databases.
I would like to know if there is a place to discuss a change proposal prior to the implementation ? The issues are not open on github.
There are other changes that I would like to propose:
- Allow variant translation (e.g. CosmosDB uses `incr` and `decr` instead of `Order.asc/desc`
- In the __next__ function, add the following code:
  ```python
   if not hasattr(self.last_traverser, "object"):
      object = self.last_traverser
      self.last_traverser = None
      return object
  ```
  The reason is that is it is possible that at this point, the `self.last_traverser` is already the object itself and it would crash to try accessing `object`
- Add a tiny minimal OGM using pydantic that would allow cleaner code (Goblin-OGM exists but it is not maintained, badly documented, ...)
  ```python
  from pydantic import BaseModel

  class Person(BaseModel)
        __label__ = "person"
        name: str
        age: int

  person1 = Person("john", "42")
  person2 = Person("James", "24")
  client.save(person1).save(person2).next()
  # or simply
  # client.save([person1, person2]).next()
  ```
  Nb: We can also image the retrieval to automaticly wrap the result based on the label
- (Improve the general quality of the module)

Most of these changes are already made on my side through a custom library that inject itself into gremlin_python, but it would be better to have it the base module.

### DIFF
--- a/gremlin-python/src/main/python/gremlin_python/driver/client.py
+++ b/gremlin-python/src/main/python/gremlin_python/driver/client.py
@@ -24,6 +24,7 @@ from concurrent.futures import ThreadPoolExecutor
 
 from gremlin_python.driver import connection, protocol, request, serializer
 from gremlin_python.process import traversal
+from gremlin_python.process.translator import Translator
 
 log = logging.getLogger("gremlinpython")
 
@@ -44,7 +45,7 @@ class Client:
                  transport_factory=None, pool_size=None, max_workers=None,
                  message_serializer=None, username="", password="",
                  kerberized_service="", headers=None, session=None,
-                 enable_user_agent_on_connect=True, **transport_kwargs):
+                 enable_user_agent_on_connect=True, bytecode_allowed=True, **transport_kwargs):
         log.info("Creating Client with url '%s'", url)
 
         # check via url that we are using http protocol
@@ -55,6 +56,7 @@ class Client:
         self._headers = headers
         self._enable_user_agent_on_connect = enable_user_agent_on_connect
         self._traversal_source = traversal_source
+        self._bytecode_allowed = bytecode_allowed
         if not self._use_http and "max_content_length" not in transport_kwargs:
             transport_kwargs["max_content_length"] = 10 * 1024 * 1024
         if message_serializer is None:
@@ -165,6 +167,13 @@ class Client:
             self._transport_factory, self._executor, self._pool,
             headers=self._headers, enable_user_agent_on_connect=self._enable_user_agent_on_connect)
 
+    def allow_bytecode(self, bytecode_allowed):
+        self._bytecode_allowed = bytecode_allowed
+        return self
+
+    def get_translator(self):
+        return Translator(self._traversal_source)
+
     def submit(self, message, bindings=None, request_options=None):
         return self.submit_async(message, bindings=bindings, request_options=request_options).result()
 
@@ -184,8 +193,11 @@ class Client:
         processor = ''
         op = 'eval'
         if isinstance(message, traversal.Bytecode):
-            op = 'bytecode'
-            processor = 'traversal'
+            if not self._bytecode_allowed:
+                message = self.get_translator().translate(message)
+            else:
+                op = 'bytecode'
+                processor = 'traversal'
 
         if isinstance(message, str) and bindings:
             args['bindings'] = bindings


Cole-Greer:
Hey @divad1196, thanks for submitting this PR. I will take a closer look at your changes later today or tomorrow. I wanted to explain our process a bit first though.

The best places to discuss issues and changes before development will either be through our [JIRA](https://issues.apache.org/jira/projects/TINKERPOP/) (used for issue tracking) or through the [dev@tinkerpop.apache.org](mailto:dev@tinkerpop.apache.org) maillist. You can subscribe to the dev list if you would like by sending an email to [dev-subscribe@tinkerpop.apache.org](mailto:dev-subscribe@tinkerpop.apache.org). You may need a PMC member to help activate a JIRA account, you can request help with this here, on the dev list, or in the TinkerPop [discord server](https://discord.gg/tinkerpop).

In order for the changes to get accepted, they will need 3 approving votes from TinkerPop committers. For smaller PR's that usually does not take too long.

One additional thing to note here is that your branch is based off of master. The master branch is where we accumulate breaking changes for the next major release (in this case 4.0.0), if your changes are non-breaking and you would like them included in an older release line, please rebase the branch and retarget the PR to the oldest appropriate dev branch. For example, if you want your changes included in the upcoming 3.6.6 and 3.7.1 releases, the PR should target 3.6-dev. Please let me know if you need any help with this or have any further questions.

divad1196:
Hi and thank you for all these details.
(It would be create to have them quickly available from the readme)

I will follow your instructions as soon as possible and also rebase my branch.

divad1196:
@Cole-Greer Hi, I did the rebase on 3.6-dev and joined the discord. I am not sure to understand how editing only the 3.6-dev branch will also impact 3.6.6 and 3.7.1 at once ?

I took a look at the Jira: I am more interested in improving the clients for the moment, but I don't see a clear way to filter by client. I think I will use the discord for the rest. Thank you

Cole-Greer:
Thanks @divad1196, I agree that a brief first time contributors guide somewhere such as the README would be useful. I will try to find some time assemble something there soon. The official contributing guidelines can be found [here](https://tinkerpop.apache.org/docs/current/dev/developer/#_contributing_code_changes), although the discoverability and accessibility could likely be improved. As for your question about getting changes into 3.6.6 and 3.7.1, it comes from the way we merge PR's. When a PR gets accepted into 3.6-dev, the committer who merges it in will also merge that change up to 3.7-dev and master to ensure that it also ends up in newer branches. We always merge changes up the branches, but not down.

As for your changes here, I took a closer look today and I think there is more consideration needed before we can proceed further. The first question is if the "script-only" mode would bring benefit to any systems beyond CosmosDB. While I do have some sympathy regarding incompatibilities, if CosmosDB is the only outlier here, I would much rather see them unify behind the standard TinkerPop system rather than fracture our drivers with additional configurations. One of my long term goals for TinkerPop would be to have the number of configurations for drivers and servers reduced significantly. Additionally, I really want to reduce the differentiation between all of our drivers. We currently have too many features which are exclusive to only a few of our drivers. For that reason, if we accept this change, it would require us to prioritize adding similar capabilities to the other drivers (Java, go, .net, js) sometime in the near future.

I think this is a decision which would need a [dev-list](mailto:dev@tinkerpop.apache.org) discussion before proceeding in order to decide if this is a direction the community would like to move in. If the community does want to accept these changes, some tests would need to be written for this PR to ensure that the driver remains compatible with gremlin-server with `bytecode_allowed=False`.

Overall I think it is an interesting change, and it definitely looks valuable for CosmosDB users, but I'm not sure if extending the drivers to support outdated systems is the right direction for TinkerPop at this time.

divad1196:
Hi,

Once again, thank you for the detailed response and the explanation. I thought that you would merge up the commits but I felt like it was a lot of manual work. So thank you for the work here also. For the contribution guide, your initial message would be a good basis, or just adding the link to the readme would be a first step.


For the reasons of being able to force the string mode: debug, retrocompatibility (e.g database never upgraded) or interoperability (having a gremlin-like query system doesn't mean it has anything to do with tinkerpop. In CosmosDB, `Order.desc` is `decr`)

Even if there is only one database concerned, the marketshare must be considered. CosmosDB is Azure graph database and a lot of person what to use that. Many changed to neo4j or memgraph and use the official library instead. Also, I thought it might be something planned because I saw that the translator wasn't available in 3.4.3 but was in 3.7.0. What would the use-cases be otherwise, except maybe to write unit tests?

On another topic, even though I think that this feature would be needed in every language (I can add it to java/js if the translator exists already), I think that wanting to keep all languages with the same features is slowing down their respective evolutions. Especially, there are not as easy to improve and the same utility libraries won't be available in all languages and we cannot take advantage of the language specific features. Let me know what you think about it.

Regarding the tests, I think that at some point we should mock a server (or the request library use in the client) to see what is sent. I can add the tests once the development is accepted and once we agree on the tests to write

For the dev-list, is an email better than discord or can I use discord for that?

Thank you and have a nice day.

Cole-Greer:
Hi,

Email would be the preferred platform for the discussion. Discord is good to help build out a broader TinkerPop community, but according to Apache guidelines, any development decisions about the project need to happen on the dev-list.

As for the testing, we currently have a set of Gherkin tests which are used by all of our drivers. Each test scenario runs a traversal through the driver connected to a gremlin-server instance, and asserts that the correct results come back. I believe the best solution for testing here would be to adapt these existing tests to utilize your driver configuration. I can examples of these tests if we need them after the dev-list discussion.

One final thing to note here is that while there are translators for each GLV, they are not all fully featured. They all handle basic/core gremlin well but they are not perfect. There was a bunch of additions to the Python translator that just got merged in the last few weeks. To the best of my knowledge the python translator is now complete, however there were some substantial gaps in it prior to these recent changes.

divad1196:
Hi @Cole-Greer , I think that I will close the MR for now. I agree that we shouldn't adapt the library to CosmosDB. I still think that a solution should be proposed and having an external "monkey-patching library" would be good enough.

For the other ideas, I will come back with the correct process. Thank you again for your time and the project.

PR #2316, opened 2023-10-31
[TINKERPOP-2996] fix Golang Translator to work with list arguments

vkagamlyk:
Current implementation of Golang Translator can't handle array arguments.

For example `g.inject([[1,2],[3,4]])` translated as `g.Inject(1,2,3,4)' what is wrong.

Also in this PR `P.Within` and `P.Without` extended to accept array arguments similar to other GLV's.

https://issues.apache.org/jira/browse/TINKERPOP-2996


kenhuuu:
I don't know `gremlin-go` well, but it is important to maintain consistency among the GLVs so my VOTE is +1.

xiazcy:
LGTM VOTE +1. Just a note to whoever merges this, since we are changing the generated feature file with the translation, might need to regenerate it up the branch.

PR #2317, opened 2023-10-31
Added Scope to String Functions

xiazcy:
Added `Scope` to asString(), length(), toLower(), toUpper(), trim(), lTrim() rTrim(),replace(),split() and substring() to allow String manipulation inside incoming lists.

Added a `StringLocalStep` abstract class for all local scoped functions, except asString() because it's structured slightly differently (doesn't throw type exceptions). 

Some notes on implementation:
- `Scope.local` will accept list traverser and single string value but throw on all other types.
- Non-string lists will also throw.
- Nulls will be passed on as usual

Scope is not added to reverse() step because it overloads as a list function, and local doesn't make sense for lists. 

VOTE +1


docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -618,24 +618,26 @@ link:https://tinkerpop.apache.org/docs/x.y.z/reference/#asDate-step[reference]
 [[asString-step]]
 === asString()
 
-*Description:* Returns the value of incoming traverser as strings
+*Description:* Returns the value of incoming traverser as strings, or if `Scope.local` is specified, returns each element inside
+incoming list traverser as string.
 
-*Syntax:* `asString()`
+*Syntax:* `asString()` | `asString(local)`
```
kenhuuu, line 624: This looks a bit odd to me. I'd think it would be something like `asString(Scope)`

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -618,24 +618,26 @@ link:https://tinkerpop.apache.org/docs/x.y.z/reference/#asDate-step[reference]
...
 |=========================================================
 |Start Step |Mid Step |Modulated |Domain |Range
-|N |Y |N |`any` |`String`
+|N |Y |N |`any` |`String`/`List`
 |=========================================================
 
 *Arguments:*
 
-None
+* `scope` - Determines the type of travers it operates on. Both scope will operate on the level of individual traversers.
```
kenhuuu, line 634:
```suggestion
* `scope` - Determines the type of traversers it operates on. Both scopes will operate on the level of individual traversers.
```

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -618,24 +618,26 @@ link:https://tinkerpop.apache.org/docs/x.y.z/reference/#asDate-step[reference]
...
+The `global` scope will operate on individual traverser. The `local` scope will operate on list traverser.
 
 Null values from the incoming traverser are not processed and remain as null when returned.
 
 *Exceptions*
-* If the incoming traverser is a non-String value then an `IllegalArgumentException` will be thrown.
+None
```
kenhuuu, line 640: Maybe add some explanation about what happens for non-String traversers for Global Scope and non-list traversers for local scope.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/DefaultGremlinBaseVisitor.java
```diff
@@ -987,16 +1003,31 @@ protected void notImplemented(final ParseTree ctx) {
 	 * {@inheritDoc}
 	 */
 	@Override
```
kenhuuu, line 1005: This file contains inconsistencies on the placement of the Override annotation. In some places its inline with the method signature, in others, its on a separate line. You should keep these in the same line to match the rest of the file.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AsStringGlobalStep.java
```diff
@@ -32,15 +32,15 @@
  * @author David Bechberger (http://bechberger.com)
  * @author Yang Xia (http://github.com/xiazcy)
  */
-public final class AsStringStep<S> extends ScalarMapStep<S, String> {
+public final class AsStringGlobalStep<S, E> extends ScalarMapStep<S, E> {
```
kenhuuu, line 35: Why did you decide to change this type parameter from <S, String> to <S, E>?

xiazcy: `public default GraphTraversal<S, E> asString(final Scope scope)`
This will complain otherwise, cause the local and global scopes return different thing (string vs list of string).

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/DefaultGremlinBaseVisitor.java
xiazcy: Prob copy and past error, I'll fix them

docs/src/dev/provider/gremlin-semantics.asciidoc
xiazcy: Yea I think there were a few misses on this doc...I'll scan through and update all mistakes on this.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/StringLocalStep.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    public Set<TraverserRequirement> getRequirements() {
+        return Collections.singleton(TraverserRequirement.OBJECT);
+    }
+
+    protected abstract E applyStringOperation(final String item);
+
+    protected IllegalArgumentException newListItemException(final String className) {
+        return new IllegalArgumentException(
+                String.format("String manipulation steps with local scope can only take string or list of strings, encountered %s in list", className));
```
kenhuuu, line 77: It might be nice to get the actual name of the step in this error message. Would be helpful in cases where the query contains multiple, different string steps.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -1470,94 +1480,195 @@ public default GraphTraversal<S, String> concat(final String... concatStrings) {
...
+
+    /**
+     * Returns the value of incoming traverser as strings. Null values are returned as a string value "null".
+     *
+     * @param scope local will operate on individual strings within incoming lists, global will operate on current traversal as a single object.
+     * @return the traversal with an appended {@link AsStringGlobalStep} or {@link AsStringLocalStep} depending on the {@link Scope}.
+     * @see <a href="http://tinkerpop.apache.org/docs/${project.version}/reference/#asString-step" target="_blank">Reference Documentation - AsString Step</a>
+     * @since 3.7.1
+     */
+    public default GraphTraversal<S, E> asString(final Scope scope) {
```
kenhuuu, line 1500: I'm concerned about the use of type inference here. Given that the return value could be a String or a List and an unchecked cast is used to return E. Can you confirm that you've tested to make sure this doesn't cause casting errors at runtime?

Code like:
`Integer shouldntCast = g.V().asString(Scope.global).next()` should cause a compilation issue and not a runtime cast exception.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/StringLocalStep.java
xiazcy: I couldn't find a way to get the nice name of the step (e.g. `toUpper(local)`), you can only get it from the Bytecode and it's ugly. It will be the actual step class name (e.g. ToUpperLocalStep), which I don't know if it's something we want to expose. I'll find a simpler way to override it per step.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
xiazcy: The cast compilation warning occurs when the assignment type doesn't match the `S` type. Unfortunately I'm not sure if there is a better way to work around the casting with returning String or List, do you have any suggestions?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/StringLocalStep.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public Set<TraverserRequirement> getRequirements() {
+        return Collections.singleton(TraverserRequirement.OBJECT);
+    }
+
+    protected abstract E applyStringOperation(final String item);
+
+    protected String getStepName() {
+        return "local scope of string manipulation";
```
kenhuuu, line 78: This doesn't seem like a reasonable default. Might be better to just make this abstract to force subclasses to override.

kenhuuu:
VOTE +1.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/StringLocalStep.java
xiazcy: Makes sense, I'll update it.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
xiazcy: Update to use `E2`

docs/src/dev/provider/gremlin-semantics.asciidoc
xiazcy: I believe that is explained above under arguments

vkagamlyk:
VOTE +1

PR #2318, opened 2023-11-01
Bytecode support for HTTP requests

vkagamlyk:
Added Gremlin bytecode support for HTTP requests for Java and Python GLV's without translation to Gremlin script.
Added serialization for entire `RequestMessage`.
Payload type distinguished by `Content-Type` header .

### DIFF
--- a/docs/src/dev/provider/index.asciidoc
+++ b/docs/src/dev/provider/index.asciidoc
@@ -978,6 +978,9 @@ with a custom sub-protocol.  Under this configuration, Gremlin Server accepts re
 evaluates that script and then streams back the results.  The notion of "streaming" is depicted in the diagram to the
 right.
 
+There is also an experimental possibility to use link:http://en.wikipedia.org/wiki/HTTP[HTTP]. Currently this is only
+supported by Java and Python GLV's.
+
 The diagram shows an incoming request to process the Gremlin script of `g.V()`.  Gremlin Server evaluates that script,
 getting an `Iterator` of vertices as a result, and steps through each `Vertex` within it.  The vertices are batched
 together given the `resultIterationBatchSize` configuration.  In this case, that value must be `2` given that each
@@ -1032,6 +1035,21 @@ The above JSON represents the "body" of the request to send to Gremlin Server. W
 WebSocket, Gremlin Server can accept a packet frame using a "text" (1) or a "binary" (2) opcode. Using "text"
 is a bit more limited in that Gremlin Server will always process the body of that request as JSON.  Generally speaking
 "text" is just for testing purposes.
+
+For HTTP can be used following serializers distinguished by `Content-Type` HTTP header
+[width="100%",cols="3,10, 10",options="header"]
+|=========================================================
+|Serializer | Content-Type| Description
+|GraphBinaryMessageSerializerV1 | application/vnd.graphbinary-v1.0 | Efficient binary format recommended for default use.
+|GraphSONMessageSerializerV3 | application/vnd.gremlin-v3.0+json | Json-compatible format containing types of all data.
+|GraphSONUntypedMessageSerializerV3 | application/vnd.gremlin-v3.0+json;types=false | Json-compatible format without metadata. For this reason it cannot be used for transmission of ByteCode requests.
+|=========================================================
+
+Same options can be used for `Accept` Header. By default, same serializer will handle both request and response.
+Only HTTP Post requests are supported.
+Both Gremlin strings and Gremlin bytecode requests may be supported.
+
+NOTE: All this HTTP options designed for GLV's. For usage with browsers, curl, Postman and other tools see link:https://tinkerpop.apache.org/docs/x.y.z/reference/#connecting-via-http[Connecting via HTTP section].
 
 The preferred method for sending requests to Gremlin Server is to use the "binary" opcode.  In this case, a "header"
 will need be sent in addition to to the "body".  The "header" basically consists of a "mime type" so that Gremlin
--- a/docs/src/dev/provider/index.asciidoc
+++ b/docs/src/dev/provider/index.asciidoc
@@ -152,11 +152,11 @@ link:https://tinkerpop.apache.org/docs/x.y.z/reference/#hadoop-gremlin[HadoopGra
 implementations for ideas and patterns.
 
 . Online Transactional Processing Graph Systems (*OLTP*)
-.. Structure API: `Graph`, `Element`, `Vertex`, `Edge`, `Property` and `Transaction` (if transactions are supported).
-.. Process API: `TraversalStrategy` instances for optimizing Gremlin traversals to the provider's graph system (i.e. `TinkerGraphStepStrategy`).
+ .. Structure API: `Graph`, `Element`, `Vertex`, `Edge`, `Property` and `Transaction` (if transactions are supported).
+ .. Process API: `TraversalStrategy` instances for optimizing Gremlin traversals to the provider's graph system (i.e. `TinkerGraphStepStrategy`).
 . Online Analytics Processing Graph Systems (*OLAP*)
-.. Everything required of OLTP is required of OLAP (but not vice versa).
-.. GraphComputer API: `GraphComputer`, `Messenger`, `Memory`.
+ .. Everything required of OLTP is required of OLAP (but not vice versa).
+ .. GraphComputer API: `GraphComputer`, `Messenger`, `Memory`.
 
 Please consider the following implementation notes:
 
--- a/docs/src/dev/provider/index.asciidoc
+++ b/docs/src/dev/provider/index.asciidoc
@@ -152,11 +152,11 @@ link:https://tinkerpop.apache.org/docs/x.y.z/reference/#hadoop-gremlin[HadoopGra
 implementations for ideas and patterns.
 
 . Online Transactional Processing Graph Systems (*OLTP*)
- .. Structure API: `Graph`, `Element`, `Vertex`, `Edge`, `Property` and `Transaction` (if transactions are supported).
- .. Process API: `TraversalStrategy` instances for optimizing Gremlin traversals to the provider's graph system (i.e. `TinkerGraphStepStrategy`).
+.. Structure API: `Graph`, `Element`, `Vertex`, `Edge`, `Property` and `Transaction` (if transactions are supported).
+.. Process API: `TraversalStrategy` instances for optimizing Gremlin traversals to the provider's graph system (i.e. `TinkerGraphStepStrategy`).
 . Online Analytics Processing Graph Systems (*OLAP*)
- .. Everything required of OLTP is required of OLAP (but not vice versa).
- .. GraphComputer API: `GraphComputer`, `Messenger`, `Memory`.
+.. Everything required of OLTP is required of OLAP (but not vice versa).
+.. GraphComputer API: `GraphComputer`, `Messenger`, `Memory`.
 
 Please consider the following implementation notes:
 
@@ -978,9 +978,6 @@ with a custom sub-protocol.  Under this configuration, Gremlin Server accepts re
 evaluates that script and then streams back the results.  The notion of "streaming" is depicted in the diagram to the
 right.
 
-There is also an experimental possibility to use link:http://en.wikipedia.org/wiki/HTTP[HTTP]. Currently this is only
-supported by Java and Python GLV's.
-
 The diagram shows an incoming request to process the Gremlin script of `g.V()`.  Gremlin Server evaluates that script,
 getting an `Iterator` of vertices as a result, and steps through each `Vertex` within it.  The vertices are batched
 together given the `resultIterationBatchSize` configuration.  In this case, that value must be `2` given that each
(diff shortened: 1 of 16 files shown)


gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpHandlerUtil.java
```diff
@@ -74,6 +78,39 @@ public class HttpHandlerUtil {
     public static RequestMessage getRequestMessageFromHttpRequest(final FullHttpRequest request) {
         final String contentType = Optional.ofNullable(request.headers().get(HttpHeaderNames.CONTENT_TYPE)).orElse("application/json");
 
+        if (contentType.equals(SerTokens.MIME_GRAPHSON_V3)) {
+            final GraphSONMessageSerializerV3 serializer = new GraphSONMessageSerializerV3();
+            try {
+                final ByteBuf buffer = request.content();
+
+                final int first = buffer.readByte();
+                // !!! skip header. Why we write this to json?
```
spmallette, line 87: like every odd looking design choice, i think it goes back to an old concept. the gremlin server protocol was written such that the request message included the content type in the serialized message itself: 

https://tinkerpop.apache.org/docs/3.7.0/dev/provider/#_graph_driver_provider_requirements

in this way the server could detect the serializer used for the message and choose the right one to decode it.  For HTTP it is a bit redundant to do this because we supply `Content-Type` explicitly, but when we had a pure NIO based protocol to support we used that as the means for supplying this information. It also stuck for websockets where it is still in use, though in retrospect we probably could have passed that information once in the handshake.

Interestingly, the fact that we encode the serializer here offers a possible solution to some issues faced here with this change. The `Content-Type` for the payload on the `POST` could always stay `application/json` and preserve the old functionality because the serialized `RequestMessage` itself contains the serializer to use to decode it. There would be no need to add an custom header to tell the server what serializer to use. Serializing a whole `RequestMessage` into the JSON request structure for HTTP seems like a fairly clean solution doesn't it?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpHandlerUtil.java
```diff
@@ -71,9 +74,32 @@ public class HttpHandlerUtil {
     /**
      * Convert a http request into a {@link RequestMessage}.
      */
-    public static RequestMessage getRequestMessageFromHttpRequest(final FullHttpRequest request) {
-        final String contentType = Optional.ofNullable(request.headers().get(HttpHeaderNames.CONTENT_TYPE)).orElse("application/json");
+    public static RequestMessage getRequestMessageFromHttpRequest(final FullHttpRequest request,
+                                                                  Map<String, MessageSerializer<?>> serializers) throws SerializationException {
+        final String contentType = request.headers().get(HttpHeaderNames.CONTENT_TYPE);
+
+        if (!contentType.equals("application/json") && serializers.containsKey(contentType)) {
```
spmallette, line 81: the `!contentType.equals("application/json")` condition means that you use a naked JSON `ObjectMapper` to process the request. with the `Accept` header a user could specify "application/json" to mean Typed GraphSON 3 (for example) if the server was not configured that way:

https://github.com/apache/tinkerpop/blob/master/gremlin-util/src/main/java/org/apache/tinkerpop/gremlin/util/ser/GraphSONMessageSerializerV3.java#L85

I guess for the `Content-Type` and the HTTP specification it's expected that the client be explicit in terms of what the payload is and there is no form of content negotiation that goes on there.  I suppose that having it work this way spares the user from having to properly configure the server, such that `application/json` ties to a Untyped GraphSON serializer (which i can only presume would work as well as a naked `ObjectMapper`).  

could you test to see if that is the case? if the `GraphSONUntypedMessageSerializerV3` decodes the request in the same fashion that the naked `ObjectMapper` does, then I think it would remove doubts about any inconsistencies we'd need to warn users about. Moreover, we could technically remove the naked` ObjectMapper` and this condition and just try to decode with the serializers. 

Perhaps said another way, instead of using `!contentType.equals("application/json")` to control flow here, use the presence/absence of a header in the payload to do that, then if it is there  you just use the `MessageSerializer` and if it is not you pull the `Mapper` object from the serializer:

https://github.com/apache/tinkerpop/blob/master/gremlin-util/src/main/java/org/apache/tinkerpop/gremlin/util/ser/AbstractGraphSONMessageSerializerV2.java#L162-L165

and use that to decode the payload. I think one neat thing about this approach is that you could conceivable then send your payload with or without the header potentially (though the underlying `Mapper` object between GraphSON and GraphBinary don't have the same interfaces but theoretically it seems like it could work). 

one other point here perhaps it would be good to check the verb earlier to avoid this path: `request.method() == POST`?

vkagamlyk: what types of requests I think to support:
1. existing https://tinkerpop.apache.org/docs/current/reference/#connecting-via-http
intended to use with curl, postman, etc by users
both GET and POST
`Content-Type` header can be empty or `application/json`
`Accept`  header can be any, but I think most useful can be `application/json`, `text/plain`, `application/vnd.gremlin-v3.0+json` and `application/vnd.gremlin-v3.0+json;types=false`
Request body example: `{ "gremlin": "g.V()" }`
We can introduce some model to deserialize with object mapper or continue to use [existing code](https://github.com/apache/tinkerpop/blob/35ac4963f72699dd123fbc99022824df37988f94/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpHandlerUtil.java#L74)

2. newly introduced payload with serialized `RequestMessage`
intended for drivers/GLV's. Support both gremlin and bytecode queries.
only POST
`Content-Type` is defined by used serializer, expected type GraphSON `application/vnd.gremlin-v3.0+json` or GraphBinary `application/vnd.graphbinary-v1.0`. Untyped GraphSON is not supported, it can't deserialize bytecode
`Accept`  header can be any. I may add validation that used same serializer for request and response.
Request body contains serialized `RequestMessage`

To distinguish this types of requests I propose to use [Content-Type HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type). I think that using a payload to determine the data type is an unnecessary complication and not HTTP way.

`GraphSONUntypedMessageSerializerV3` can't deserialize bytecode request, it knows nothing about steps or parameters, everything turns into HashMap of HashMap.
image

spmallette: It seems strange to me to use the `Content-Type` for anything more than a serialization rule. If I understand correctly, by your definition of what you want to support, you can only access the old HTTP API with `application/json` but if you make it `application/vnd.gremlin-v3.0+json` then the payload must be a `RequestMessage` or else it won't work. I suppose we didn't support `application/vnd.gremlin-v3.0+json` as a `Content-Type` there before so maybe that's ok, but now that there is a path that allows you to POST that format, it wouldn't surprise me if folks tried it, especially since from the perspective `Accept` you can request `application/json` and assuming you've configured the server that way you could get back a `application/vnd.gremlin-v3.0+json` formatted payload. 

well, please think through the validation on requests here carefully and write good documentation to clarify how this works. we need to make sure folks understand the API they are calling and what `Content-Type` and `Accept` headers the need to use to get the results that they want.

docs/src/dev/provider/index.asciidoc
```diff
@@ -1033,6 +1036,21 @@ WebSocket, Gremlin Server can accept a packet frame using a "text" (1) or a "bin
...
 "text" is just for testing purposes.
 
+For HTTP can be used following serializers distinguished by `Content-Type` HTTP header
+[width="100%",cols="3,10, 10",options="header"]
+|=========================================================
+|Serializer | Content-Type| Description
+|GraphBinaryMessageSerializerV1 | application/vnd.graphbinary-v1.0 | Efficient binary format recommended for default use.
+|GraphSONMessageSerializerV3 | application/vnd.gremlin-v3.0+json | Json-compatible format containing types of all data.
+|GraphSONUntypedMessageSerializerV3 | application/vnd.gremlin-v3.0+json;types=false | Json-compatible format without metadata. For this reason it cannot be used for transmission of ByteCode requests.
+|=========================================================
```
xiazcy, line 1046: I wonder if we want to give a bit more details on the design? E.g. how the request is handled when it comes in? Why we are using Content-Type, what should the payload looks like, and some notes on what happens if you used different serializers for request/response? Also some potential issue people may encounter with this (especially if spec isn't followed)?

Also I wonder if this needs to be in a separate section, it feels a bit odd being sticked into the middle of this doc, but I don't have any good recommendations right now either.

spmallette: I'd echo @xiazcy concerns here. I sense that it's possible that some of the docs have even fallen behind here as I read the content around what's been added. This feature remains experimental so when I was thinking about documentation here I was mostly thinking about more extensive javadoc rather than official changes to asciidoc. When the experimental label is dropped we can add this sort of content.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpGremlinEndpointHandler.java
```diff
@@ -262,7 +269,7 @@ public void channelRead(final ChannelHandlerContext ctx, final Object msg) {
                             attemptCommit(requestMessage.getArg(Tokens.ARGS_ALIASES), graphManager, settings.strictTransactionManagement);
 
                             try {
-                                return Unpooled.wrappedBuffer(serializer.getValue1().serializeResponseAsString(responseMessage, ctx.alloc()).getBytes(UTF8));
+                                return Unpooled.wrappedBuffer(serializer.getValue1().serializeResponseAsBinary(responseMessage, ctx.alloc()));
```
spmallette, line 272: i don't remember why it was chosen to encode this way (and perhaps it was wrong to do) but this seems like a breaking change.  The following is from 3.7.0 and would no longer return the same response, right?:

```
$ curl -v -H Accept:application/vnd.graphbinary-v1.0 http://localhost:8182?gremlin=g.V%28%29
*   Trying 127.0.0.1:8182...
* Connected to localhost (127.0.0.1) port 8182 (#0)
> GET /?gremlin=g.V%28%29 HTTP/1.1
> Host: localhost:8182
> User-Agent: curl/7.81.0
> Accept:application/vnd.graphbinary-v1.0
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< content-type: application/vnd.graphbinary-v1.0
< content-length: 872
< 
* Connection #0 to host localhost left intact
gQDX0dKWp4pPrJQSEq6fq21cAAAAyAAAAAAAAAAAAAAAAAAJAAAAAAYRAAIAAAAAAAAAAAEAAAAGcGVyc29uCQAAAAACEgACAAAAAAAAAAAAAAAABG5hbWUDAAAAAAVtYXJrb/4BCQAAAAAAEgACAAAAAAAAAAABAAAAA2FnZQEAAAAAHf4BCQAAAAAAEQACAAAAAAAAAAACAAAABnBlcnNvbgkAAAAAAhIAAgAAAAAAAAAAAgAAAARuYW1lAwAAAAAFdmFkYXP+AQkAAAAAABIAAgAAAAAAAAAAAwAAAANhZ2UBAAAAABv+AQkAAAAAABEAAgAAAAAAAAAAAwAAAAhzb2Z0d2FyZQkAAAAAAhIAAgAAAAAAAAAABAAAAARuYW1lAwAAAAADbG9w/gEJAAAAAAASAAIAAAAAAAAAAAUAAAAEbGFuZwMAAAAABGphdmH+AQkAAAAAABEAAgAAAAAAAAAABAAAAAZwZXJzb24JAAAAAAISAAIAAAAAAAAAAAYAAAAEbmFtZQMAAAAABGpvc2j+AQkAAAAAABIAAgAAAAAAAAAABwAAAANhZ2UBAAAAACD+AQkAAAAAABEAAgAAAAAAAAAABQAAAAhzb2Z0d2FyZQkAAAAAAhIAAgAAAAAAAAAACAAAAARuYW1lAwAAAAAGcmlwcGxl/gEJAAAAAAASAAIAAAAAAAAAAAkAAAAEbGFuZwMAAAAABGphdmH+AQkAAAAAABEAAgAAAAAAAAAABgAAAAZwZXJzb24JAAAAAAISAAIAAAAAAAAAAAoAAAAEbmFtZQMAAAAABXBldGVy/gEJAAAAAAASAAIAAAAAAAAAAAsAAAADYWdlAQAAAAAj/gEJAAAAAAA=
```

vkagamlyk: $ curl -v -H Accept:application/vnd.graphbinary-v1.0 http://localhost:45940?gremlin=gmodern.V%28%29
* processing: http://localhost:45940?gremlin=gmodern.V%28%29
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying [::1]:45940...
* Connected to localhost (::1) port 45940
> GET /?gremlin=gmodern.V%28%29 HTTP/1.1
> Host: localhost:45940
> User-Agent: curl/8.2.1
> Accept:application/vnd.graphbinary-v1.0
>
< HTTP/1.1 200 OK
< content-type: application/vnd.graphbinary-v1.0
< content-length: 629
<
{ [629 bytes data]

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpHandlerUtil.java
```diff
@@ -70,10 +76,53 @@ public class HttpHandlerUtil {
...
+        final String contentType = request.headers().get(HttpHeaderNames.CONTENT_TYPE);
+
+        if (request.method() == POST && contentType != null && !contentType.equals("application/json") && serializers.containsKey(contentType)) {
+            final MessageSerializer<?> serializer = serializers.get(contentType);
+
+            final ByteBuf buffer = request.content();
+
+            // additional validation for header
+            final int first = buffer.readByte();
+            if (first != 0x7b) {
```
rdtr, line 106: can you add a comment that we are checking `'{'` ?

vkagamlyk:
re-targeted to 3.6 in https://github.com/apache/tinkerpop/pull/2336

PR #2319, opened 2023-11-01
TINKERPOP-2984 Replace Moq with NSubstitute

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-2984


vkagamlyk:
Missing changelog entry

otherwise VOTE+1

FlorianHockmann:
> Missing changelog entry

Do you think this needs one as its only a test dependency? Users shouldn't notice this change.

Cole-Greer:
VOTE +1

PR #2329, opened 2023-11-07
TINKERPOP-2988 Add requestId checking to Javascript and Python drivers.

kenhuuu:
Added client side checking instead of server side checking due to the difference in how graphson/graphbinary deserialization works. Also, this mostly affects websockets and not HTTP, so it is likely only drivers that are affected. The other GLVs already check that requestIds are valid UUIDs, so this PR just brings the Javascript/Python drivers to parity with the others.

https://issues.apache.org/jira/browse/TINKERPOP-2988

VOTE +1


vkagamlyk:
VOTE+1

gremlin-python/src/main/python/gremlin_python/driver/connection.py
```diff
@@ -61,6 +61,7 @@ def write(self, request_message):
         request_id = str(uuid.uuid4())
         if request_message.args.get("requestId"):
             request_id = request_message.args.get("requestId")
+            uuid.UUID(request_id) # The server will return an error if non-UUID sent.
```
xiazcy, line 64:
```suggestion
            uuid.UUID(request_id) # checks UUID is not malformed as the server will return an error if non-UUID sent.
```
Thought I'd add a bit more clarification that this is doing a check and not the transformation. I was initially wondering why you didn't assign it back to request id.

PR #2330, opened 2023-11-07
GremlinServer: Restrict the thread pool size to 1 when serving a TinkerTransactionGraph

danielcweber:
For sessionless requests, executing requests on different threads leads to unpredictable behaviour, even if requests are serialized on the client side (even with some reasonable wait time between requests). Wrapping each request in a transaction on the client side works just fine however, so this change might be a bit to broad for explicit transaction processing. Maybe there is a way to only restrict the sessionless processor and leave the session-processor multi-threaded?

### DIFF
--- a/gremlin-server/conf/gremlin-server-transaction.yaml
+++ b/gremlin-server/conf/gremlin-server-transaction.yaml
@@ -39,6 +39,7 @@ metrics: {
   csvReporter: {enabled: true, interval: 180000, fileName: /tmp/gremlin-server-metrics.csv},
   jmxReporter: {enabled: true},
   slf4jReporter: {enabled: true, interval: 180000}}
+gremlinPool: 1
 strictTransactionManagement: false
 idleConnectionTimeout: 0
 keepAliveInterval: 0


vkagamlyk:
Hi Daniel,
Could you share some details about unpredictable behavior? Maybe some code examples or steps to reproduce the problem

danielcweber:
>>Maybe some code examples or steps to reproduce the problem
That's my main issue...I can't reliably.

>>Could you share some details about unpredictable behavior
We got a large set of integration tests interfacing Gremlin Server with "conf/gremlin-server-transaction.yaml" setup. The tests run sequentially. Even with a pool size of 1, and even with an artificial wait time between tests, the results from Gremlin Server are randoml missing vertex properties. It seems to be only vertex properties. They are added on one unit test, and are missing on the next. It all goes away when the server is configured to run on one thread only.

The main question is here: Is it expected that TinkerTransactionGraph in sessionless mode may exhibit non durable updates, even when requests are run sequentially, even with some wait time? It seems, as it is currently, that it's only to expected to work reliably when requests are run on the same thread each. If that is the case, my PR makes sense. If it's not the case, there might be a bug in TinkerTransactionGraph.

I will continue to try to reproduce this reliably.

vkagamlyk:
PR with fix https://github.com/apache/tinkerpop/pull/2343

PR #2344, opened 2023-11-15
[TINKERPOP-2877] Added integer overflow checks

vkagamlyk:
For some integer operations the result is not checked for overflow, which can lead to unexpected results.
For example
```
gremlin> __.inject(Long.MAX_VALUE,Long.MAX_VALUE,2).sum()
==>0
```

In this PR I added throwing `ArithmeticException` for overflow cases. 

More details in https://issues.apache.org/jira/browse/TINKERPOP-2877

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
@@ -28,10 +28,22 @@ import java.util.function.BiFunction;
  */
 public final class NumberHelper {
 
+    private static byte asByte(int arg) {
+        if (arg > Byte.MAX_VALUE || arg < Byte.MIN_VALUE)
+            throw new ArithmeticException("byte overflow");
+        return (byte) arg;
+    }
+
+    private static short asShort(int arg) {
+        if (arg > Short.MAX_VALUE || arg < Short.MIN_VALUE)
+            throw new ArithmeticException("short overflow");
+        return (short) arg;
+    }
+
     static final NumberHelper BYTE_NUMBER_HELPER = new NumberHelper(
-            (a, b) -> (byte) (a.byteValue() + b.byteValue()),
-            (a, b) -> (byte) (a.byteValue() - b.byteValue()),
-            (a, b) -> (byte) (a.byteValue() * b.byteValue()),
+            (a, b) -> asByte(a.byteValue() + b.byteValue()),
+            (a, b) -> asByte(a.byteValue() - b.byteValue()),
+            (a, b) -> asByte(a.byteValue() * b.byteValue()),
             (a, b) -> (byte) (a.byteValue() / b.byteValue()),
             (a, b) -> {
                 if (isNumber(a)) {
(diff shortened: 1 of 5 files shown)


kenhuuu:
Does this account for Math Step? exp is a valid function there.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
```diff
@@ -28,10 +28,22 @@
  */
 public final class NumberHelper {
 
+    private static byte asByte(int arg) {
+        if (arg > Byte.MAX_VALUE || arg < Byte.MIN_VALUE)
+            throw new ArithmeticException("byte overflow");
```
kenhuuu, line 33: Nit: this technically doesn't qualify for an ArithmeticException but should probably be an IllegalArgumentException since it is bound checking the incoming argument.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/util/NumberHelperTest.java
```diff
@@ -96,6 +98,21 @@ public class NumberHelperTest {
                     new Quartet<>(BigDecimal.ONE, BigDecimal.ONE, BigDecimal.class, BigDecimal.class)
             );
 
+    private final static List<Triplet<Number, Number, String>> OVERFLOW_CASES = Arrays.asList(
```
kenhuuu, line 101: Very minor nit: This doesn't need to be done, but ParameterizedTest could have been used here to split each into individual test.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
vkagamlyk: `Math.addExact` for int and long throw `ArithmeticException`, I do the same for byte and short

vkagamlyk:
> Does this account for Math Step? exp is a valid function there.

there is no exp in `NumberHelper`.
Math Step evaluation handled by external library `net.objecthunter.exp4j`, it's out of scope.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
kenhuuu: Their usage is different as it's not used for argument checking, it applies checking to a result. I'm OK with not changing this due to how NumberHelper is set up as it would add complicated exception handling for little gain.

kenhuuu:
> > Does this account for Math Step? exp is a valid function there.
> 
> there is no exp in `NumberHelper`. Math Step evaluation handled by external library `net.objecthunter.exp4j`, it's out of scope.

Ok I don't think this needs any changes, but it does introduce inconsistencies if that library doesn't throw the ArithmeticException. Maybe a test case would be good.

kenhuuu:
VOTE +1

vkagamlyk:
VOTE+1

PR #2335, opened 2023-11-10
[TINKERPOP-3014] Reorder the jcl-over-slf4j dependency to avoid the dependency conflict.

HappyHacker123:
Related to https://issues.apache.org/jira/browse/TINKERPOP-3014. 
Dependencny jcl-over-slf4j in gremlin-core is declared but unused due to dependency conflict, change the order so that the jcl-over-slf4j exceeds commons-logging in classpath.

PR #2349, opened 2023-11-16
[TINKERPOP-3014] Reorder the jcl-over-slf4j dependency to avoid the dependency conflict.

HappyHacker123:
Fix #2335.

### DIFF
--- a/gremlin-core/pom.xml
+++ b/gremlin-core/pom.xml
@@ -36,6 +36,14 @@ limitations under the License.
             <version>${project.version}</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


Cole-Greer:
Thanks @HappyHacker123, this change looks good to me, VOTE +1.

One thing to note here is that as there are no tests related to this change, it's possible that this dependency ordering may be altered in the future, unknowingly affecting this change.

As this is a small change, I will merge this in as a CTR (Commit then Review), bypassing the standard RTC (Review then Commit) process which requires 3 approving votes.

PR #2351, opened 2023-11-16
Update conjoin() and asString() null handling.

kenhuuu:
The handling of null values in some recently implemented Steps were inconsistent. This addresses that issue by making conjoin() behave like concat() and asString() like the other String functions.


kenhuuu:
This is a small change so I'm merging as CTR.

PR #2361, opened 2023-11-23
Fix flaky traversal test due to duplicated key generated for project step

xiazcy:
The random strings generated for method arguments can sometimes contain duplicate, which will fail the new requirement for project step to have unique keys only (introduced in https://github.com/apache/tinkerpop/pull/2217). 

This PR adds processing to remove duplicates after the string args are generated.

(Note this is not needed for the current release).

### DIFF
--- a/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
+++ b/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
@@ -34,7 +34,6 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Random;
 import java.util.Set;
--- a/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
+++ b/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
@@ -168,7 +168,7 @@ public class GraphTraversalTest {
                     }
                     // remove duplicates in argument[1] if any
                     arguments[1] = Arrays.stream((String[]) arguments[1]).distinct().toArray(String[]::new);
-                    list.addAll(Arrays.stream((String[]) arguments[1]).collect(Collectors.toList()));
+                    list.addAll(Arrays.asList((String[]) arguments[1]));
                 } else {
                     for (int i = 0; i < stepMethod.getParameterTypes().length; i++) {
                         final Class<?> type = stepMethod.getParameterTypes()[i];
--- a/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
+++ b/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
@@ -34,6 +34,7 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Random;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
```diff
@@ -174,8 +175,13 @@ else if (boolean.class.isAssignableFrom(type))
                         else if (String[].class.isAssignableFrom(type)) {
                             arguments[i] = new String[random.nextInt(10) + 1];
                             for (int j = 0; j < ((String[]) arguments[i]).length; j++) {
-                                list.add(((String[]) arguments[i])[j] = randomString(random));
+                                list.add(((String[]) arguments[i])[j] = arguments[0] + randomString(random)); // adds the first string to all to avoid duplicates
```
kenhuuu, line 178: It's not obvious to me how adding arguments[0] avoids duplicates. At first glance it seems to just make longer duplicates.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
```diff
@@ -174,8 +175,13 @@ else if (boolean.class.isAssignableFrom(type))
...
                             }
+                            // makes sure that no duplicated random string is created by removing duplicates
+                            arguments[i] = Arrays.stream((String[]) arguments[i]).distinct().toArray(String[]::new);
+                            final Set<Object> tempSet = new LinkedHashSet<>(list);
+                            list.clear();
+                            list.addAll(tempSet);
```
kenhuuu, line 184: This logic seems overly complex. The simplest way to do this would probably be to just generate the arguments and ensure they are unique. Then, add that unique array to list after.

kenhuuu:
Duplicates for String[] should probably be allowed (just make an exception for ProjectStep) so I would have preferred to do it that way but I think this approach is likely fine as well.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
xiazcy: No, this array is strictly argument[1]. The way this is set up, the actual argument array is [`String`, `String[]`], where argument[0] is of type `String` and gets processed (turned into a random string) in the previous if case, and only argument[1] gets to this if case. As I think there is still a chance that whatever string was generated for argument[0] to be duplicated in the arrays of string generate for `String[]`, so the simplest way to avoid that is to just append argument[0] to all the string inside of argument[1].

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
xiazcy: There is actually two collections here, the argument array to be passed into the method and the list saved for the equality comparison part of the test. I didn't want to change the list adding part, but you are correct I could add them to a temp set first then add to the list. I'll update that.

xiazcy:
> Duplicates for String[] should probably be allowed (just make an exception for ProjectStep) so I would have preferred to do it that way but I think this approach is likely fine as well.

True, I'll special case project then.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
xiazcy: updated in https://github.com/apache/tinkerpop/pull/2361/commits/701245111c6d72c762d3ef9635b0e2046fcf165e

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
```diff
@@ -158,6 +159,16 @@ else if (stepMethod.getName().equals("to") || stepMethod.getName().equals("from"
...
+                } else if (stepMethod.getName().equals("project")) {
+                    // project has two arguments [String, String[]]
+                    list.add(arguments[0] = randomString(random));
+                    arguments[1] = new String[random.nextInt(10) + 1];
+                    for (int j = 0; j < ((String[]) arguments[1]).length; j++) {
+                        ((String[]) arguments[1])[j] = arguments[0] + randomString(random); // adds argument[0] to avoid getting its duplicate in argument[1]
+                    }
+                    // remove duplicates in argument[1] if any
+                    arguments[1] = Arrays.stream((String[]) arguments[1]).distinct().toArray(String[]::new);
+                    list.addAll(Arrays.stream((String[]) arguments[1]).collect(Collectors.toList()));
```
kenhuuu, line 171: Might be able to do something similar to
```suggestion
                    list.addAll(Arrays.asList(argument[1]);
```

kenhuuu:
VOTE +1

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalTest.java
```diff
@@ -38,6 +38,7 @@
 import java.util.Random;
 import java.util.Set;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
```
Cole-Greer, line 41: This looks like an unused import
```suggestion
```

Cole-Greer:
VOTE +1

PR #2366, opened 2023-11-29
TINKERPOP-3017 Added AnonymousTraversalSource.with()

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-3017

Replaces `withEmbedded()` and `withRemote()` and removes the long deprecated
`withGraph()`. A single unified with() removes the need to check the object being given to the `AnonymousTraversalSource` to decide which method to use. Can't remember why that split was used in the first place. This should make it easier to explain how to use the `AnonymousTraversalSource` in the first place, requires a bit less typing and simply streamlines the API.

VOTE +1

### DIFF
--- a/gremlin-go/driver/anonymousTraversal.go
+++ b/gremlin-go/driver/anonymousTraversal.go
@@ -21,7 +21,12 @@ package gremlingo
 
 // AnonymousTraversalSource interface for generating anonymous traversals.
 type AnonymousTraversalSource interface {
+	// With used to set the DriverRemoteConnection within the AnonymousTraversalSource
+	With(drc *DriverRemoteConnection) *GraphTraversalSource
+
 	// WithRemote used to set the DriverRemoteConnection within the AnonymousTraversalSource
+	//
+	// Deprecated: WithRemote replaced by With
 	WithRemote(drc *DriverRemoteConnection) *GraphTraversalSource
 }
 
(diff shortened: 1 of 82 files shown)


kenhuuu:
This change makes sense to me for Java as there is no clashing of parameter types between the remote and embedded versions so this is a nice change. However, it looks a bit odd for Python since it is renamed to with_.

VOTE +1

spmallette:
>  However, it looks a bit odd for Python since it is renamed to with_.

the underscore is basically part of the whole Gremlin language for python though: https://tinkerpop.apache.org/docs/current/reference/#gremlin-python-differences

gremlin-go/driver/anonymousTraversal.go
```diff
@@ -22,7 +22,7 @@ package gremlingo
 // AnonymousTraversalSource interface for generating anonymous traversals.
 type AnonymousTraversalSource interface {
 	// WithRemote used to set the DriverRemoteConnection within the AnonymousTraversalSource
-	WithRemote(drc *DriverRemoteConnection) *GraphTraversalSource
+	With(drc *DriverRemoteConnection) *GraphTraversalSource
```
xiazcy, line 25: Are we just deprecating the usage? If so, we should keep WithRemote() in addition to adding With(), as removing it will remove the method implementation from the interface. I.e.:
```suggestion
	With(drc *DriverRemoteConnection) *GraphTraversalSource
	// WithRemote used to set the DriverRemoteConnection within the AnonymousTraversalSource
	//
	// Deprecated: WithRemote replaced by With
	WithRemote(remote *DriverRemoteConnection) *GraphTraversalSource
```

spmallette: that was a mistake. good catch

xiazcy:
VOTE +1

PR #2367, opened 2023-11-29
[TINKERPOP-2128] Added ExplainStrategy

vkagamlyk:
Added `ExplainStrategy` which allows to get traversal explanation for remote traversals and GLV's.

https://issues.apache.org/jira/browse/TINKERPOP-2128


gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/finalization/ExplainStrategyTest.java
```diff
@@ -0,0 +1,50 @@
...
+public class ExplainStrategyTest {
+
+    private static final GraphTraversalSource g = traversal().withEmbedded(EmptyGraph.instance()).
+            withStrategies(CountStrategy.instance(), ReferenceElementStrategy.instance());
+
+    @Test
+    public void shouldModifyTraversal() {
+        final Traversal<Vertex, ?> t = g.withStrategies(ExplainStrategy.instance()).V().limit(1).count().is(0);
+
+        final String explanation = (String) t.next();
```
kenhuuu, line 42: The casting to String here is similar to the problem stated in https://issues.apache.org/jira/browse/TINKERPOP-2932 . Essentially, strategies that can remove steps from the end of the traversal will cause the S, E type parameters to become incorrect. This could end up being confusing for users. I'm fine with this trade-off as long as other potential implementations have even worse useability.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/integrated/ExplainStrategy.feature
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+@StepClassIntegrated
+Feature: Step - ExplainStrategy
```
kenhuuu, line 19: A test with more steps in the traversal should be added.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -5097,6 +5097,17 @@ new transaction immediately following the `commit()` that raises the events. The
 may also not behave as "snapshots" at the time of their creation as they are "live" references to actual database
 elements.
 
+=== ExplainStrategy
```
spmallette, line 5100: this change seems to create a new and additional way to do explain in Java where only this new one works in non-JVM languages? there's always going to be little differences among the various languages but this one seems really fundamental. everyone knows the `explain()` and I don't think there should be another way to do it.  

I see the complexity of the `explain()` signature in trying to make this change - it returns `TraversalExplanation` rather than `Traversal` and is effectively a form of terminal step. 

I don't know if this makes this change much more elegant, but couldn't you remove the user need to know about `ExplainStrategy` if `Traversal.explain()` added it and then self-iterated to return the `TraversalExplanation`?

In Java, I suppose you would self-iterate a clone so that you could still do:

```
gremlin> t = g.V().out();[]
gremlin> t.explain()
==>Traversal Explanation
==========================================================================================
Original Traversal                    [GraphStep(vertex,[]), VertexStep(OUT,vertex)]
ConnectiveStrategy              [D]   [GraphStep(vertex,[]), VertexStep(OUT,vertex)]
...
Final Traversal                       [TinkerGraphStep(vertex,[]), VertexStep(OUT,vertex)]
gremlin> t
==>v[3]
==>v[2]
==>v[4]
==>v[5]
==>v[3]
==>v[3]
```

On the idea of a completely different approach, I can't help wondering if "explain" isn't a form of `GraphOp`, like transactions are (i.e. a special form of `Bytecode` that trigger a particular command on the server). Did you consider that?

vkagamlyk:
Let's return to this topic in the context of TinkerPop 4

PR #2377, opened 2023-12-05
Replace `none()` with `discard()`

ryn5:
Renames none() to discard() to make room for a new list filtering none() step in https://github.com/apache/tinkerpop/pull/2385.


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/__.java
```diff
@@ -806,6 +805,13 @@ public static <A> GraphTraversal<A, Set<?>> difference(final Object values) {
         return __.<A>start().difference(values);
     }
 
+    /**
+     * @see GraphTraversal#discard()
+     */
+    public static <A> GraphTraversal<A, A> discard() {
```
Cole-Greer, line 811: I notice that there is no corresponding `none()` method being replaced here, is this needed?

Cole-Greer (commented):
Thanks Ryan, these changes LGTM.

VOTE +1 (pending 1 minor comment below)

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/__.java
ryn5: This was originally added to pass a test that looks for steps being implemented in GraphTraversal, but after looking into it again it seems it should just be ignored as none() was previously, so I'll be adding "discard" as an ignore case and removing this function.

vkagamlyk:
VOTE+1

xiazcy:
LGTM. VOTE +1

PR #2382, opened 2023-12-06
Let JavaTranslator.translateObject() throw clearer ISE

metawilm-aws:
Wrap the full original exception, because its message might not be enough to pinpoint the issue.


spmallette:
cherry picked back to 3.6-dev - 500252dcfc99b1791e949e71210c510cafcee01a

PR #2395, opened 2023-12-11
draft proposal for multi-threaded tx

vkagamlyk:


docs/src/dev/future/proposal-multi-threaded-transaction-5.asciidoc
```diff
@@ -0,0 +1,146 @@
...
+=== Introduction
+The current implementation of transactions and sessions uses approach with thread local variables that requires complex implementation and LifeCycle management. Using non-threaded sessions can improve performance in some scenarios.
+
+=== Motivation
+* Simlify GremlinExecutor. Complex LifeCycle control is now used.
+* Easier integration with HTTP because transactions can be used without being tied to a connection.
+
+=== Assumptions
+
+* Need to reuse existing code as much as possible.
```
kenhuuu, line 32: I don't think you should be restricted by needing to reuse code.

docs/src/dev/future/proposal-multi-threaded-transaction-5.asciidoc
```diff
@@ -0,0 +1,146 @@
...
+* Users should be able to work both types of transactions, thread-bound and new non-thread bound. Perhaps after performance testing only one type will be left.
```
kenhuuu, line 33: I think it makes more sense to just phase out the old transaction model. There should ultimately only be one model.

docs/src/dev/future/proposal-multi-threaded-transaction-5.asciidoc
```diff
@@ -0,0 +1,146 @@
...
+* Transaction implementation should not have additional external dependencies.
```
kenhuuu, line 34: Not clear what this means. Could you expand on it.

docs/src/dev/future/proposal-multi-threaded-transaction-5.asciidoc
```diff
@@ -0,0 +1,146 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+////
+image::apache-tinkerpop-logo.png[width=500,link="https://tinkerpop.apache.org"]
+
+*x.y.z - Proposal 4*
+
+Proposal for adding support for tx which are not thread bound.
```
kenhuuu, line 21: Based on the rest of the document, this seems to be a design of how to modify TinkerTransactionGraph. Could you expand on how the transaction model itself changes? This information is more helpful for providers to determine if the new model fits with their systems. They may not know how TinkerTransactionGraph works.

kenhuuu:
Can you describe a scenario where multiple threads are using the same transaction (e.g. multiple threads trying to add vertices to the same transaction).

docs/src/dev/future/proposal-multi-threaded-transaction-5.asciidoc
```diff
@@ -0,0 +1,146 @@
...
+
+=== Introduction
+The current implementation of transactions and sessions uses approach with thread local variables that requires complex implementation and LifeCycle management. Using non-threaded sessions can improve performance in some scenarios.
+
+=== Motivation
+* Simlify GremlinExecutor. Complex LifeCycle control is now used.
```
Cole-Greer, line 27:
```suggestion
* Simplify GremlinExecutor. Complex LifeCycle control is now used.
```

docs/src/dev/future/proposal-multi-threaded-transaction-5.asciidoc
Cole-Greer: Is this statement specifically about the transaction model in TinkerGraph or for all providers who support transactions? I agree that TinkerGraph should eventually converge on a single model, however I imagine there would be some providers who would like to retain a thread-bound model.

docs/src/dev/future/proposal-multi-threaded-transaction-5.asciidoc
```diff
@@ -0,0 +1,146 @@
...
+* Easier integration with HTTP because transactions can be used without being tied to a connection.
+
+=== Assumptions
+
+* Need to reuse existing code as much as possible.
+* Users should be able to work both types of transactions, thread-bound and new non-thread bound. Perhaps after performance testing only one type will be left.
+* Transaction implementation should not have additional external dependencies.
+* Existing drivers should work without changes.
```
Cole-Greer, line 35: In my opinion, existing driver compatibility is a strong plus, but not a strict requirement if these changes are targeted to TP 4. I would view this as a soft requirement that could be dropped with proper justification.

docs/src/dev/future/proposal-multi-threaded-transaction-5.asciidoc
```diff
@@ -0,0 +1,146 @@
...
+
+
+==== CRUD operations with transaction
+===== Create
+Add new TinkerElementTxContainer to `vertices` and `edges` in TinkerGraphTx.
+
+==== Read
+Read values from `vertices` TinkerElementTxContainer in TinkerGraphTx.
+If marked as deleted than return null;
+If not found try TinkerTransactionGraph2.vertices.
```
Cole-Greer, line 120: Would we want to add the following to ensure consistent reads?

If found in `TinkerTransactionGraph2.vertices`, copy to `vertices` TinkerElementTxContainer in TinkerGraphTx?

docs/src/dev/future/proposal-multi-threaded-transaction-5.asciidoc
```diff
@@ -0,0 +1,146 @@
...
+
+===== Update
+Add or update if exist corresponding TinkerElementTxContainer  in `vertices`.
```
Cole-Greer, line 123: Would any CRUD operation on a property essentially be processed as an Update to its parent/grandparent vertex/edge?

docs/src/dev/future/proposal-multi-threaded-transaction-5.asciidoc
xiazcy: Will this new model fully replace the current model once it's out in a major release? Or will it be more of deprecating current model, then removing it a major release after? Or are you proposing to keep both indefinitely?

docs/src/dev/future/proposal-multi-threaded-transaction-5.asciidoc
xiazcy: Agreed, this would be in a major release, for which compatibility with existing driver wouldn't be guaranteed.

PR #2397, opened 2023-12-12
TINKERPOP-2950 Add docker shutdown handling to close containers gracefully

ryn5:
JIRA: https://issues.apache.org/jira/browse/TINKERPOP-2950.  Handles SIGTERM when shutting down docker container by sending SIGINT to server processes.

### DIFF
--- a/gremlin-server/src/main/docker/docker-entrypoint.sh
+++ b/gremlin-server/src/main/docker/docker-entrypoint.sh
@@ -35,6 +35,6 @@ handler()
 }
 
 exec $GREMLIN_SERVER "$@" &
-PID=$(ps aux | grep -w $GREMLIN_SERVER | grep -v grep | awk '{print $1}')
+PID=$(ps aux | grep -w $GREMLIN_SERVER | grep -v grep | awk 'NR==1 {print $1}')
 trap 'handler $PID' SIGTERM
 wait "$PID"
--- a/gremlin-server/src/main/docker/docker-entrypoint.sh
+++ b/gremlin-server/src/main/docker/docker-entrypoint.sh
@@ -35,6 +35,6 @@ handler()
 }
 
 exec $GREMLIN_SERVER "$@" &
-PID=$(ps aux | grep -w $GREMLIN_SERVER | grep -v grep | awk 'NR==1 {print $1}')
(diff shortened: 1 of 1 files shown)


Cole-Greer:
Thanks Ryan, I checked out the branch and tested out the new image, it's working great!

VOTE +1

xiazcy:
VOTE +1

vkagamlyk:
VOTE +1

rcbyron:
I'm still having issues with graceful shutdown even with this code.

I use the default scripts/empty-sample.groovy which has:
```groovy
globals << [hook : [
        onStartUp: { LifeCycleHook.Context ctx ->
            ctx.logger.info("Executed once at startup of Gremlin Server.")
        },
        onShutDown: { LifeCycleHook.Context ctx ->
            ctx.logger.info("Executed once at shutdown of Gremlin Server.")
            graph.close()  // <-- I ADDED THIS
        }
] as LifeCycleHook]
```
And `Executed once at shutdown...` is never called I tried stopping/starting/restarting from Docker Desktop.

danielcweber:
This is an issue for me as well. It seems that docker-entrypoint.sh forwards SIGTERMS as SIGINTS to /opt/gremlin-server/bin/gremlin-server.sh but it's ignored there. What needs to be sent the signal instead is the java process that gremlin-server.sh starts.

Cole-Greer:
I'm seeing the same behaviour as well now. The change here is preventing docker from hanging due to `/opt/gremlin-server/bin/gremlin-server.sh` not terminating, but it does not properly terminate the gremlin server JVM process.

@danielcweber As you've already dug into this, would you be willing to submit a PR with a fix? If not let's reopen the JIRA for tracking and ensure it gets fixed for the next release.

danielcweber:
@Cole-Greer My naive attempt that is working for me [would be this change](https://github.com/danielcweber/tinkerpop/commit/a02d15c9ea15a4a2c98dfe6d4d9c11d41eb9c1c4). This is not even PR quality because I don't know what this will break for other users. Basically, through this change, the 'java' process will end up as PID 1 and receive all the signals.

Cole-Greer:
Thanks @danielcweber. I don't see anything that would break just from looking at the code. I will find some time to run some tests with it, if everything checks out I think we can merge that in.

danielcweber:
@Cole-Greer If everything is fine, [here's a PR](#2712).

PR #2422, opened 2023-12-28
[TINKERPOP-2872] Change JS tests to compare elements by ID for consistency with other GLVs

ryn5:
https://issues.apache.org/jira/projects/TINKERPOP/issues/TINKERPOP-2872
Passes an ID comparator for deep-eql to use when object is type Element, then overwrites chai's members() and isSubsetOf() to use this comparator.


gremlin-javascript/src/main/javascript/gremlin-javascript/test/unit/element-comparison-test.js
```diff
@@ -0,0 +1,261 @@
...
+    const e2 = new Edge(3, v1, "chases", v3, undefined);
+    const e3 = new Edge(3, v2, "chases", v3, undefined);
+
+    const vp1 = new VertexProperty(3, "size", "small", undefined);
+    const vp2 = new VertexProperty(3, "size", "large", undefined);
+    const vp3 = new VertexProperty(4, "size", "large", undefined);
+
+    it('should pass with same id, different values', function () {
+        expect(deepEqual(v1, v2, opt)).to.be.true;
+        expect(deepEqual(v3, e1, opt)).to.be.true;
```
Cole-Greer, line 64: This might be swinging the comparison from too strict to too loose. It might be better if the comparison is only true if both elements are the same type and have the same ID. I'm not sure when someone might end up with an edge and a vertex with the same ID but I don't think they should be considered equal.

gremlin-javascript/src/main/javascript/gremlin-javascript/test/cucumber/element-comparison.js
```diff
@@ -0,0 +1,122 @@
...
+ * The functions deepMembersById, flag, and isSubsetOf are adapted from
+ * Chai's source code.
+ * See licenses/chai for full license.
+ */
+
+const chai = require('chai');
+const deepEqual = require('deep-eql');
+
+function isElement(obj) {
+    return obj !== null && obj.hasOwnProperty('id') && obj.hasOwnProperty('label');
```
vkagamlyk, line 32: can we use `obj instanceof Element` here?

gremlin-javascript/src/main/javascript/gremlin-javascript/test/unit/element-comparison-test.js
```diff
@@ -0,0 +1,261 @@
...
+    it('should pass only properties that match exactly', function () {
+        expect(deepEqual(p1, p2, opt)).to.be.true;
+        expect(deepEqual(p1, p3, opt)).to.be.false;
+        expect(deepEqual(p1, p4, opt)).to.be.false;
+        expect(deepEqual(p3, p5, opt)).to.be.false;
+        expect(deepEqual(p3, p5, opt)).to.be.false;
+    });
+});
+
+describe('arrays', function () {
```
vkagamlyk, line 93: Good to have test for array of elements

xiazcy:
Missing a CHANGELOG entry

gremlin-javascript/src/main/javascript/gremlin-javascript/test/unit/element-comparison-test.js
ryn5: Updated to check for class equality and changed these tests with same id, different class to false

gremlin-javascript/src/main/javascript/gremlin-javascript/test/cucumber/element-comparison.js
ryn5: I don't think so since Element isn't an export of the graph.js module, but `obj instanceof Edge || obj instanceof Vertex || obj instanceof VertexProperty` seems to work well

gremlin-javascript/src/main/javascript/gremlin-javascript/test/unit/element-comparison-test.js
ryn5: Added `element arrays` tests similar to `arrays`

vkagamlyk:
VOTE+1

xiazcy:
VOTE +1

Cole-Greer:
Thanks Ryan! VOTE +1

PR #2424, opened 2024-01-03
TINKERPOP-3029 Fix enumeration for .NET 8

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-3029

`IEnumerable<T>.Current` has been changed in .NET 8. Before .NET 8, it simply returned `null` if `MoveNext()` wasn't called first. With .NET 8 however it throws an exception.
Since this has apparently already been undefined behavior before, we should fix this irrespective of .NET 8:
https://github.com/dotnet/runtime/issues/85243#issuecomment-1521085177

The problem can be reproduced by executing the Gherkin tests with .NET without the change in `DefaultTraversal` (by changing the `TargetFramework` in `Gremlin.Net.IntegrationTest.csproj` to `net8.0`). .NET 8 needs to be installed for this of course.

VOTE +1

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Process/Traversal/DefaultTraversal.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Process/Traversal/DefaultTraversal.cs
@@ -82,21 +82,22 @@ namespace Gremlin.Net.Process.Traversal
         }
         
         private bool MoveNextInternal()
-        {                   
+        {
             if (_fetchedNext) return _nextAvailable;
-            
-            var currentTraverser = TraverserEnumerator.Current;
-            if (currentTraverser?.Bulk > 1)
-            {
-                currentTraverser.Bulk--;
-                _nextAvailable = true;
-            }
-            else
+
+            if (!_nextAvailable || TraverserEnumerator.Current?.Bulk == 0)
             {
                 _nextAvailable = TraverserEnumerator.MoveNext();
             }
+            if (!_nextAvailable) return false;
+            
+            var currentTraverser = TraverserEnumerator.Current;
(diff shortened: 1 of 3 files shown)


gremlin-dotnet/src/Gremlin.Net/Process/Traversal/DefaultTraversal.cs
```diff
@@ -82,21 +82,22 @@ public bool MoveNext()
         }
         
         private bool MoveNextInternal()
-        {                   
+        {
             if (_fetchedNext) return _nextAvailable;
+
+            if (!_nextAvailable || _nextAvailable && TraverserEnumerator.Current?.Bulk == 0)
```
vkagamlyk, line 88: can be simplified
```suggestion
            if (!_nextAvailable || TraverserEnumerator.Current?.Bulk == 0)
```

EricSites: This code will also throw an exception if called without first starting the iterator: You will need to track that the iterator has not been started.

```.csharp
private object GetCurrent()
{
    // Use dynamic to object to prevent runtime dynamic conversion evaluation
    return TraverserEnumerator.Current?.Object;
}
```

Please review the new [.Net code here](https://source.dot.net/#PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/Projections/IEnumerable.cs,239)

vkagamlyk:
thank you @FlorianHockmann! 

VOTE+1

xiazcy:
VOTE +1

jorgebay (approved):
VOTE +1

PR #2425, opened 2024-01-04
Improved contains check for bulkset with elements

steigma:
Improved within test check for bulkset with elements (i.e., Vertex, Edge, VertexProperty) by using contains method. Due to changes w.r.t. Gremlin comparison semantics (cf. https://tinkerpop.apache.org/docs/3.7.0/dev/provider/#gremlin-semantics-concepts) this check was no longer done efficiently, which led to some regressions (see query/example below). In some cases, we can however ensure that the contains of the bulkset (using hash code and Object.equals) leads to the same results as the GremlinValueComparator.COMPARABILITY.equals. In fact, for elements, both checks are only be done with the ids of these elements.

This change re-enables an efficient check for elements (if the bulkset also contains these elements and only contains these kind of elements). This is realized via a transient attribute (allContainedElementsSameClass) in the bulkset class that represents whether all elements are of same type/class, which is checked by the within test method. Tje attribute is computed lazily when accessed to avoid overhead if the information is not required.

Pseudo code for sample data:
```
final Vertex x1 = G.addVertex(T.id, "x1", T.label, "person", "age", 27, "name", "x1");
// many friends for x1
for (int i = 1; i < 10000; ++i) {
    final Vertex x1fi = G.addVertex(T.id, "f"+i, T.label, "person", "age", 27, "name", "f"+i);
    x1.addEdge("knows", x1fi, T.id, "e-x1-f"+i, "weight", 0.5);
}
// one special friend that also has many other friends
final Vertex x1f0 = G.addVertex(T.id, "f0", T.label, "person", "age", 27, "name", "f0");
x1.addEdge("knows", x1f0, T.id, "e-x1-f0", "weight", 0.5);

// adding these many other friends, so friends of friends for x1
for (int i = 1; i < 10000; ++i) {
    final Vertex x1f0ofi = G.addVertex(T.id, "fof"+i, T.label, "person", "age", 27, "name", "fof"+i);
    x1f0.addEdge("knows", x1f0ofi, T.id, "e-f0-f"+i, "weight", 0.5);
}
```

Sample query (which is very inefficiently executed without this change):
```
g.V("x1").as("root").aggregate("directFriends")
                                .select("root").out().aggregate("directFriends")
                                .select("directFriends").limit(1).unfold().out().where(without("directFriends"))
```
The query is obviously not optimally formulated, but reproduces the issue

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Contains.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Contains.java
@@ -54,9 +54,6 @@ public enum Contains implements BiPredicate<Object, Collection> {
         public boolean test(final Object first, final Collection second) {
             if (first instanceof Element &&
                     second instanceof BulkSet<?> &&
-                    ((BulkSet<?>)second).allContainedElementsSameClass() &&
-                    ((BulkSet<?>)second).getAllContainedElementsClass() != null &&
-                    Element.class.isAssignableFrom(((BulkSet<?>)second).getAllContainedElementsClass()) &&
                     first.getClass() == ((BulkSet<?>)second).getAllContainedElementsClass()) {
                 /*
                  * For elements (i.e., vertices, edges, vertex properties) it is safe to use the contains check
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Contains.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Contains.java
@@ -18,10 +18,10 @@
  */
 package org.apache.tinkerpop.gremlin.process.traversal;
 
-import org.apache.tinkerpop.gremlin.util.iterator.IteratorUtils;
+import org.apache.tinkerpop.gremlin.process.traversal.step.util.BulkSet;
+import org.apache.tinkerpop.gremlin.structure.Element;
 
 import java.util.Collection;
-import java.util.Objects;
 import java.util.function.BiPredicate;
 
 /**
@@ -52,6 +52,21 @@ public enum Contains implements BiPredicate<Object, Collection> {
     within {
         @Override
         public boolean test(final Object first, final Collection second) {
+            if (first instanceof Element &&
+                    second instanceof BulkSet<?> &&
+                    ((BulkSet<?>)second).allContainedElementsSameClass() &&
+                    ((BulkSet<?>)second).getAllContainedElementsClass() != null &&
+                    Element.class.isAssignableFrom(((BulkSet<?>)second).getAllContainedElementsClass()) &&
+                    first.getClass() == ((BulkSet<?>)second).getAllContainedElementsClass()) {
+                /*
+                 * For elements (i.e., vertices, edges, vertex properties) it is safe to use the contains check
+                 * since the hash code computation and equals comparison give the same result as the Gremlin equality comparison
+                 * (using GremlinValueComparator.COMPARABILITY.equals) based on the Gremlin comparison semantics
+                 * (cf. <a href="https://tinkerpop.apache.org/docs/3.7.0/dev/provider/#gremlin-semantics-concepts">...</a>).
+                 * In both cases, we just compare the ids of the elements. Therefore, it is safe to use the contains check.
+                 */
+                return second.contains(first);
+            }
             GremlinTypeErrorException typeError = null;
             for (final Object o : second) {
                 try {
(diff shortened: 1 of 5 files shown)


steigma:
javascript build errors seem unrelated: 
```
Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.15.0:install-node-and-npm (install node and npm) on project gremlin-javascript: Could not download Node.js: Got error code 522 from the server. -> [Help 1]
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Contains.java
```diff
@@ -52,6 +52,21 @@ public enum Contains implements BiPredicate<Object, Collection> {
     within {
         @Override
         public boolean test(final Object first, final Collection second) {
+            if (first instanceof Element &&
+                    second instanceof BulkSet<?> &&
+                    ((BulkSet<?>)second).allContainedElementsSameClass() &&
+                    ((BulkSet<?>)second).getAllContainedElementsClass() != null &&
+                    Element.class.isAssignableFrom(((BulkSet<?>)second).getAllContainedElementsClass()) &&
```
Cole-Greer, line 59: Could these checks be removed? I believe they are redundant. If any of these checks fail, then the final `first.getClass() == ((BulkSet<?>)second).getAllContainedElementsClass()` check would also fail.
```suggestion
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/BulkSet.java
```diff
@@ -42,6 +42,68 @@
...
+            boolean hadNull = false;
+            for (final S key : this.map.keySet()) {
+                if ((key == null || key.getClass() == null)) {
+                    if (allContainedElementsClass != null) {
+                        allContainedElementsClass = null;
+                        break;
+                    }
+                    hadNull = true;
+                } else if (hadNull) {
+                    break;
```
Cole-Greer, line 95: Can be simplified. If a single null is found in the set, then we can return false.
```suggestion
            for (final S key : this.map.keySet()) {
                if ((key == null || key.getClass() == null)) {
                    allContainedElementsClass = null;
                    return false;
```

Cole-Greer (commented):
Thanks for the submission @steigma. I see this approach as a good step towards recovering some of the efficiency of the old comparison semantics. I left a few comments with minor tweaks to simplify it slightly.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/BulkSet.java
```diff
@@ -42,6 +42,68 @@
...
+     * @return true if it is guaranteed that all contained elements/objects are of the same type/class and not null (but not
+     * necessarily, i.e., they can have the same type/class, but we may return false here if it was not analysed/identified)
+     */
+    public boolean allContainedElementsSameClass() {
+        if (!allContainedElementsClassChecked) {
+            allContainedElementsClass = null;
+            allContainedElementsClassChecked = true;
+            boolean hadNull = false;
+            for (final S key : this.map.keySet()) {
+                if ((key == null || key.getClass() == null)) {
```
vkagamlyk, line 88: In what situation can `key.getClass() == null` be true?

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/BulkSetTest.java
```diff
@@ -67,6 +68,56 @@ public void shouldHaveProperCountAndNotOutOfMemoryException() {
...
+
+    @Test
+    public void shouldNotHaveSameClassWhenEmpty() {
+        final BulkSet<Object> bulkSet = new BulkSet<>();
+        assertNull(bulkSet.getAllContainedElementsClass());
+        assertFalse(bulkSet.allContainedElementsSameClass());
+    }
+
+    @Test
+    public void shouldNotHaveSameClassForNull() {
```
vkagamlyk, line 107: Missing test case for `null` in the middle.
Something like Vertex1, null, Vertex2

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/BulkSet.java
```diff
@@ -42,6 +42,68 @@
...
+                    if (allContainedElementsClass != null) {
+                        allContainedElementsClass = null;
+                        break;
+                    }
+                    hadNull = true;
```
vkagamlyk, line 93: probably `hadNull = true;` should be before `if` in line 89

xiazcy:
Thanks for opening the PR with the improvement. I don't have much to add to what's already been commented, but having a CHANGELOG entry would be helpful.

vkagamlyk:
Thanks for contribution @steigma.

VOTE+1

Cole-Greer:
Thanks @steigma, VOTE +1

PR #2432, opened 2024-01-08
[TINKERPOP-3025] Add missing l_trim() and r_trim() to python

ryn5:
https://issues.apache.org/jira/projects/TINKERPOP/issues/TINKERPOP-3025?filter=allopenissues Added missing l_trim/r_trim, deprecated ltrim/rTrim, and minor consistency fixes to python.

### DIFF
--- a/gremlin-python/src/main/python/gremlin_python/process/graph_traversal.py
+++ b/gremlin-python/src/main/python/gremlin_python/process/graph_traversal.py
@@ -1468,10 +1468,6 @@ class __(object, metaclass=MagicType):
         return cls.graph_traversal(None, None, Bytecode()).min_(*args)
 
     @classmethod
-    def none(cls, *args):
-        return cls.graph_traversal(None, None, Bytecode()).none(*args)
-
-    @classmethod
     def not_(cls, *args):
         return cls.graph_traversal(None, None, Bytecode()).not_(*args)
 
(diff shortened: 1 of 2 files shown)


spmallette:
needs a changelog entry but this was the right idea - VOTE +1 pending that change.

gremlin-python/src/main/python/gremlin_python/process/graph_traversal.py
```diff
@@ -1451,6 +1467,10 @@ def merge_v(cls, *args):
     def min_(cls, *args):
         return cls.graph_traversal(None, None, Bytecode()).min_(*args)
 
+    @classmethod
+    def none(cls, *args):
+        return cls.graph_traversal(None, None, Bytecode()).none(*args)
```
xiazcy, line 1472: If my understanding is correct, `none()` is a terminal step so there is no use to add it in the `__` class here (which is why it hasn't been in the class).

gremlin-python/src/main/python/gremlin_python/process/graph_traversal.py
```diff
@@ -2461,6 +2497,8 @@ def where(*args):
 
 statics.add_static('in_e', in_e)
 
+statics.add_static('inV', inV)
```
vkagamlyk, line 2500: probably not related to this PR

gremlin-python/src/main/python/gremlin_python/process/graph_traversal.py
ryn5: Oh right, I forgot that none hasn't been changed to a filtering step yet.  Removing this change.

vkagamlyk:
VOTE+1

Cole-Greer:
Thanks Ryan! LGTM!
VOTE +1

PR #2438, opened 2024-01-13
Bump gremlin-python dependencies

ryn5:
Merges dependabot PRs #2401 #2402 #2403 #2404 #2405, verified that the upgrades cause no issues.


Cole-Greer:
Thanks Ryan, LGTM. Merging as CTR

PR #2468, opened 2024-01-31
TINKERPOP-3030 Update to .NET 8

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-3030

I've also noticed that the examples used camel casing for some method names which I switched to pascal casing to stay idiomatic in C#.

I had to update the Docker image to Ubuntu 20.04 because .NET 8 is not available for Ubuntu 18.04. Since 18.04 is out of support since June already, this update makes sense in general [1].
I verified that the Docker build is still working with:
```sh
./docker/build.sh --dotnet
```

VOTE +1

[1]: https://wiki.ubuntu.com/Releases

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -264,10 +264,10 @@ jobs:
         with:
           java-version: '11'
           distribution: 'temurin'
-      - name: Set up .NET 6.0.x
+      - name: Set up .NET 8.0.x
         uses: actions/setup-dotnet@v4
         with:
-          dotnet-version: '6.0.x'
+          dotnet-version: '8.0.x'
       - name: Get Cached Server Base Image
         uses: actions/cache@v3
         id: gremlin-server-test-docker-image
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -15,24 +15,24 @@
 # specific language governing permissions and limitations
 # under the License.
 
-FROM ubuntu:bionic
+FROM ubuntu:focal
 
 LABEL maintainer="Daniel Kuppitz <me@gremlin.guru>"
 
 RUN apt-get update
 RUN apt-get -y install software-properties-common python3-software-properties apt-transport-https curl dpkg netcat net-tools iproute2
 RUN add-apt-repository ppa:openjdk-r/ppa
-RUN sh -c 'curl -s https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb -o packages-microsoft-prod.deb'
+RUN sh -c 'curl -s https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb -o packages-microsoft-prod.deb'
 RUN sh -c 'dpkg -i packages-microsoft-prod.deb'
 RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
 RUN apt-get install apt-transport-https gnupg ca-certificates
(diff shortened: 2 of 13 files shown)


Cole-Greer:
I have some concerns over dropping the .NET 6 tests from 3.6-dev. .NET 6 remains supported by Microsoft for another 9 months. We should continue to support it in 3.6.x until it reaches end of life. I think it makes sense to add .NET 8 tests but it should be parallel to, not replacing the .NET 6 tests.

FlorianHockmann:
Just to clarify: This PR doesn't change which frameworks we support with Gremlin.Net. It's only about which frameworks we use to build and test it.
Gremlin.Net mainly [targets .NET Standard 2.0](https://github.com/apache/tinkerpop/blob/master/gremlin-dotnet/src/Gremlin.Net/Gremlin.Net.csproj#L21) which is implemented by various .NET frameworks like .NET 5, .NET 6, .NET 7, and .NET 8. It has .NET 6 as an additional explicit target because we optionally use some APIs which were only introduced in .NET 6 to support Websocket compression. This can therefore only be used by users on .NET 6 and higher.

Before this PR, we built and tested Gremlin.Net with .NET 6. With this PR, we're using .NET 8 instead. But users can still stay on .NET 6 (or 7 or whatever) and use newer versions of Gremlin.Net with it.

Not sure if this was clear to you and you just wanted to propose that we **test** against .NET 6 **and** .NET 8. In that case, I agree with you in general that it would be good to test on both frameworks (or better to say runtimes in this case). The problem with that however is that all contributors then also need both runtimes installed. If you just have .NET 8 installed, then the tests will fail as you also need .NET 6 in that case.
That's also why we always use the same version across all release branches.

In the end, the question comes down to how many contributors are actually affected by this and how often do we expect to find version specific bugs because of it. I guess many contributors simply don't execute .NET tests on their machines and leave that up to GH Actions or they use the Docker based setup to execute the tests. So, this probably doesn't affect many contributors.

On the other hand, [TINKERPOP-3029](https://issues.apache.org/jira/browse/TINKERPOP-3029) is also the only bug that comes to my mind which we could have catched by executing tests on different runtimes. Although I think it wouldn't have helped much as we still would have to be on .NET 8 in the first place to find it. So, the better question is how often a bug occurs that is specific to an **older** framework than the latest LTS version which we want to use by default.

I don't have a strong opinion on this. If we are confident that this doesn't really affect contributors negatively, then I'm in favor of executing the tests on multiple runtimes. Otherwise, I'd stay with our current approach.

vkagamlyk:
@FlorianHockmann did you leave `src\Gremlin.Net\Gremlin.Net.csproj` unchanged on purpose?

FlorianHockmann:
> did you leave src\Gremlin.Net\Gremlin.Net.csproj unchanged on purpose?

Yes, we still only target .NET Standard 2.0 and .NET 6. .NET 6 is there also only because we're using Websocket compression which wasn't present in .NET Standard, but was only added in .NET 6.
Users on .NET 7 or 8 can of course still use Gremlin.Net. They should get the version built for .NET 6 in that case.

We could of course add another target for .NET 7 and/or .NET 8, but I think that doesn't add any benefit if we're not using any APIs specific for these versions.

Cole-Greer:
In general I like the idea of testing both .NET 6 and .NET 8 in GHA similar to how Java is currently setup (I don't see a need for contributors to run both locally). In my mind testing both runtimes gives the strongest guarantee against issues which only appear for specific versions. .NET compatibility is not something I know very well however so I will defer to @FlorianHockmann and @vkagamlyk to determine what level of testing is required.

VOTE +0

PR #2475, opened 2024-02-02
TINKERPOP-3045: EarlyLimitStrategy is too aggresive to promote Limit …

upadhyay-prashant:
…and thus causing incorrect results

https://issues.apache.org/jira/browse/TINKERPOP-3045

EarlyLimitStrategy was too aggressive to promote limit before map steps. Not all the Map step can be displaced. Specifically if a Map step has capability to filter the results, then it should be displaced.

This code change brings in a utility to statically determine the output arity of a step and accordingly allow EarlyLimitStrategy to make a decision whether a map step can be moved.

Test is added in Range.feature which ensures that this bug is fixed. This test gives wrong result without this code change.

### DIFF
--- a/gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/Range.feature
+++ b/gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/Range.feature
@@ -281,3 +281,15 @@ Feature: Step - range()
     Then the result should be unordered
       | result |
       | d[29].i |
+
+  Scenario: g_V_mapXinX_limitX2X_valuesXnameX
+    Given the modern graph
+    And the traversal of
+      """
+      g.V().map(__.in().hasId("1")).limit(2).values("name")
+      """
+    When iterated to list
+    Then the result should be unordered
+      | result |
+      | marko |
+      | marko |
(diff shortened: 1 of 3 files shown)


saikiranboga (approved):

gremlin-core/pom.xml
```diff
@@ -101,6 +101,16 @@ limitations under the License.
             <artifactId>hamcrest</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
spmallette, line 104: it looks like there is relatively minor usage of guava and lombok. could you please adjust the `StepOutputArityPredictor` to   not use these libraries so that we don't need to take the dependency?

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/Range.feature
```diff
@@ -280,4 +280,16 @@ Feature: Step - range()
...
       | result |
-      | d[29].i |
\ No newline at end of file
+      | d[29].i |
+
+  Scenario: g_V_mapXinX_limitX2X_valuesXnameX
+    Given the modern graph
+    And the traversal of
+      """
+      g.V().map(__.in()).limit(2).values('name')
```
spmallette, line 289: nit: we prefer double quotes for strings in our feature tests as a convention.

spmallette:
for other committers, i will help backfill some extra tests on this one, deal with CHANGELOG, etc. on merge

gremlin-core/pom.xml
upadhyay-prashant: done

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/Range.feature
upadhyay-prashant: done

PR #2481, opened 2024-02-05
TINKERPOP-3045: EarlyLimitStrategy is too aggresive to promote Limit …

upadhyay-prashant:
…and thus causing incorrect results

https://issues.apache.org/jira/browse/TINKERPOP-3045

EarlyLimitStrategy was too aggressive to promote limit before map steps. Not all the Map step can be displaced. Specifically if a Map step has capability to filter the results, then it should be displaced.

This code change brings in a utility to statically determine the output arity of a step and accordingly allow EarlyLimitStrategy to make a decision whether a map step can be moved.

Test is added in Range.feature which ensures that this bug is fixed. This test gives wrong result without this code change.


spmallette:
closing for #2475 which was updated to match what is here.

PR #2483, opened 2024-02-07
TINKERPOP-2862 Added withoutStrategies syntax to grammar

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2862

Made "new" keyword optional in grammar TINKERPOP-3046. Language variants still have trouble using withoutStrategies() because of TINKERPOP-3047 and TINKERPOP-3048 that will need to get completed before they will work properly.


docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -75,10 +75,43 @@ See: link:https://issues.apache.org/jira/browse/TINKERPOP-3017[TINKERPOP-3017]
...
+
+*Supports withoutStrategies()*
+
+The `withoutStrategies()` configuration step is now supported syntax for the grammar. While this option is not commonly
+used it is still a part of the Gremlin language and there are times where it is helpful to have this fine grained
+control over how a traversal works.
+
+[source,groovy]
+----
+g.V().withoutStrategies(CountStrategy)
```
vkagamlyk, line 104: does it make sense to add an overload that accepts string as argument? Might be useful for GLV's, like
`withoutStrategies(SubgraphStrategy.class)` vs `withoutStrategies("SubgraphStrategy")`

vkagamlyk: also for providers

spmallette: It doesn't need a String. The grammar should accept any `Identifier` there. If it matches a registered strategy from a provider, it will work as valid syntax. 

https://github.com/apache/tinkerpop/pull/2483/files#diff-39ed59720093773649512a14aeba8c3da9e6d21ce016c7f0cdf7a1242e3a76a1R72

I suppose it will be up to each programming language to decide what to do with that. For Java, you could have reference to the actual class file from the provider and could use that. Since it matches on the simple name of the class a dummy class could also be used. Alternatively, Java (and other languages) could be changed to allow a `String` I suppose, but I wasn't ready to make that choice. I mostly wanted to focus in on the grammar and getting that working right for standard TinkerPop strategies. Created: https://issues.apache.org/jira/browse/TINKERPOP-3055 to track.

Cole-Greer:
LGTM VOTE +1

PR #2486, opened 2024-02-13
Remove gremlin-archetype

ryn5:
Removes the gremlin-archetypes module since newer sample applications have been added to the repo (#2298) as discussed in https://lists.apache.org/thread/qr0f2cwc6bx5v0wdf1oyo4q6k58l1vf0.


spmallette:
maybe better targetted at `master` as opposed to `3.7-dev` given semver?

Cole-Greer:
There's one reference to the archetypes in the reference docs that I see. I think we can simply delete it as it is followed by a section on the new examples. https://github.com/apache/tinkerpop/blob/master/docs/src/reference/gremlin-variants.asciidoc#gremlin-archetypes

spmallette:
in removing a few modules here, it probably needs upgrade docs, but other than that, VOTE +1

Cole-Greer:
Thanks Ryan, VOTE +1

PR #2504, opened 2024-02-23
Avoiding expensive hash computation in filter ranking strategy

steigma:
Avoiding expensive hash computation for complex steps with many children in filter ranking strategy by using identity hash map instead of linked hash map. With the linked hash map, a hash computation is repeated for the traversal parent with each child when doing the grouping, which can be quite costly when there are several hundreds of children (especially since the hash computation also includes the hash of child steps).


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/FilterRankingStrategy.java
```diff
@@ -232,4 +243,64 @@ public Set<Class<? extends OptimizationStrategy>> applyPrior() {
...
+
+    /**
+     * Get steps of the specified class throughout the traversal and grouping them based on the traversal parent
+     * collecting them in a fashion that orders them from the deepest steps first
+     */
+    public static List<Pair<TraversalParent, List<Step<?,?>>>> collectStepsOfAssignableClassRecursivelyFromDepthGroupedByParent(
+            final Traversal.Admin<?, ?> traversal, final Class<?> stepClass) {
+
+        final List<Pair<TraversalParent, List<Step<?,?>>>> collectionList = new ArrayList<>();
+        final Stack<Step<?,?>> stack = new Stack<>();
```
vkagamlyk, line 255: nit: not informative variable names. may be something like `unprocessedSteps` will be better?

vkagamlyk:
Good to have some test cases, especially for traversal where there are steps with nested global and local children

Cole-Greer:
I agree with @vkagamlyk that some additional tests would be beneficial here. It is quite hard to prove that `collectStepsOfAssignableClassRecursivelyFromDepthGroupedByParent()` does not effect the order of the resulting map in a meaningful way through reasoning alone. I would also be curious to know how substantial the performance difference can be in a "worst case" scenario with hundreds of child steps.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/FilterRankingStrategyTest.java
```diff
@@ -85,44 +86,148 @@ public void doTest() {
...
-                {has("value", 0).filter(out()).dedup(), has("value", 0).filter(out()).dedup(), Collections.emptyList()},
-                {has("value", 0).or(has("name"), has("age")).has("value", 1).dedup(), has("value", 0).has("value", 1).or(has("name"), has("age")).dedup(), Collections.singletonList(InlineFilterStrategy.instance())},
-                {has("value", 0).or(out(), in()).as(Graph.Hidden.hide("x")).has("value", 1).dedup(), has("value", 0).has("value", 1).or(outE(), inE()).dedup(), TraversalStrategies.GlobalCache.getStrategies(Graph.class).toList()},
-                {has("value", 0).and(has("age"), has("name", "marko")).is(10), __.is(10).has("value", 0).has("age").has("name", "marko"), Collections.singletonList(InlineFilterStrategy.instance())},
-                {has("value", 0).filter(or(not(has("age")), has("age", 1))).has("value", 1).dedup(), has("value", 0).has("value", 1).or(not(filter(properties("age"))), has("age", 1)).dedup(), TraversalStrategies.GlobalCache.getStrategies(Graph.class).toList()},
```
steigma, line 125: I reformatted these tests slightly and added some new ones on the top

steigma:
Thanks for the feedback. I added a few tests and renamed some variables. I also found that the lambda handling was not really working since the step collection was restricted to TraversalParent subclasses. This should now also be fixed and I added some basic test coverage for this. 

In our experiments with queries of the form
```
g.V('vertex')...
coalesce(
    and(
      values('property').is('val1'),
      values('property').is('val2'),
      ...
      values('property').is('val800')
   ),
   sideEffect(properties('property').drop()).
   property(set, 'propery', 'val1').
   property(set, 'propery', 'val2').
   ...
   property(set, 'propery', 'val800')
)...
```
we have seen improvement of around 25% including all strategies and executing the queries, i.e,. the FilterRankingStrategy improved significantly (as also other strategies taking some significant amount of time).

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/FilterRankingStrategyCacheInitializationTest.java
```diff
@@ -0,0 +1,449 @@
...
+ * Testing whether the cache for the FilterRankingStrategy is correctly initialized,
+ * which includes gathering/collecting the traversal parents and their children steps
+ * in the correct deep first manner.
+ */
+@RunWith(JUnit4.class)
+public class FilterRankingStrategyCacheInitializationTest {
+
+
+    @Test
+    public void test_collection_lambda_nested() {
```
vkagamlyk, line 58: nit: test name

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/FilterRankingStrategyCacheInitializationTest.java
```diff
@@ -0,0 +1,449 @@
...
+                __.where(
+                        __.or(
+                                __.select("n").hasLabel("software"),
+                                __.select("n").map(Lambda.function("it.get().out()"))
+                        )
+                ));
+
+        // cache entries
+        final List<Pair<Boolean, Set<String>>> expectedTraversalParentCacheEntries = Arrays.asList(
+                new Pair<>(true, new HashSet<>(Collections.emptyList())),
```
vkagamlyk, line 84: Why need to pass `Collections.emptyList()` to `HashSet` constructor?

kenhuuu:
I don't completely follow the changes being made here but they seem to be well tested so my VOTE is +1.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/FilterRankingStrategyCacheInitializationTest.java
steigma: Not completely sure what you mean, but I tried to improve the names of the test cases.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/FilterRankingStrategyCacheInitializationTest.java
steigma: It is not necessary. I removed it in the new revision. Also changed some arguments to Optional to make it clear that they do not need to be provided (where we then just compare with the old cache initialization method).

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/FilterRankingStrategyCacheInitializationTest.java
vkagamlyk: test naming convention is camelCase and `shouldDescribeWhatTestDo`.
please look at the names of other tests

steigma: Thanks, I fixed the test names in the new version

vkagamlyk:
Thank you @steigma!
VOTE +1

Cole-Greer:
Thanks @steigma, the new tests LGTM. VOTE +1

PR #1070, opened 2019-02-23
TINKERPOP-2143 JavaScript GLV: Support browsers

fdominik:
https://issues.apache.org/jira/browse/TINKERPOP-2143
Summary of changes:
* Updated dependency on WS package (which was very old 3.0.0) to v 6.0.0. This is required by browsers.
* Added dependency to util and events packages which were required by the javascript module,  but were not listed in dependencies, so that it wouldn't compile.
* Added new dependency to uiid library.

* Generation of UUIDs was passed to UUID library instead of utilizing crypto package (which is not anymore available as separate library in browsers world, it is bundled, however some functions used in generating UUID were not available in the new bundled crypto). The uuid library can generate UUID of version 4, which was the version used to generate the UUIDs.
* Usage of standard WebSocket object from default packages.
 -  a new Buffer handling had to be written because the browser WS library doesnt have automatic creation of Buffer Arrays, so it had to be implemented from scratch => using Uint8Array.

Code has been tested in our UI project, where we can send requests from UI to Gremlin server. I wasn't able to run the Tinkerpop tests locally.

PR #2506, opened 2024-02-26
feat: `gremlin-javascript` browser support

tien:
- Replace node APIs that have cross runtime equivalent
- For those that can't be replaced, use `node:` prefix so that environment that does polyfill automatically can easily do so
- Remove manual ping/pong logic, this is already handled internally by browser or `ws` library + WhatWG `WebSocket` API doesn't have ability to do manual ping/pong


tien:
Hey guys, would love to get this working in an environment outside of Node.js.

I think this would boost adoption greatly, as a lot of serverless runtimes (i.e. Cloudflare Worker, Vercel, etc) now utilise standard Web APIs.

This is just something I draft quickly together with the limited knowledge that I have. Would appreciate any helps & guidances. 🙏

kenhuuu:
Thanks for you interest in contributing. There has been a lot of interest around adding this support in the past. This is the open Jira issue for reference https://issues.apache.org/jira/browse/TINKERPOP-2143. The past attempts seem to have run into some issues with tests https://github.com/apache/tinkerpop/pull/1070 and https://github.com/apache/tinkerpop/pull/1291. I'm not familiar with JavaScript so I'll try to have someone that is look into this.

Thanks again.

tien:
@kenhuuu thanks for getting back & reminding me about the integration test, I was able to run it locally and have all tests passed. For actually running in the browser API compatible environment, I tested this on my application running on Cloudflare Worker & everything so far works as expected.

tien:
Forgot that `globalThis.crypto` is not available on Node 18, so I have made a changed to use [uuid](https://github.com/uuidjs/uuid) instead.

But optionally we can change the supported node version to LTS, which is 20 right now if you guys think that is best.

tien:
Hmm, looks like the `.nvmrc` file was picked up by the license checker. `.nvmrc` files can't contain comment, so I have added it to the list of ignored files.

kenhuuu:
> But optionally we can change the supported node version to LTS, which is 20 right now if you guys think that is best.

It's against our current runtime policy (https://tinkerpop.apache.org/docs/current/dev/developer/#runtimes) to upgrade unless the supported runtime has reached end of life. If you feel that it would really be worth it to upgrade to Node 20 then it's a conversation you could start on the dev-list.

> Hmm, looks like the .nvmrc file was picked up by the license checker. .nvmrc files can't contain comment, so I have added it to the list of ignored files.

I think that makes sense to ignore it then.

tien:
@kenhuuu Regarding the Node version, whatever make sense to you guys, of course. It's not a major deal.

But I think the wording in the link that you shared could be a bit better:

> Each programming language has a runtime that TinkerPop supports. In general, TinkerPop will attempt to support the current LTS version for a particular major version for the lifetime of its minor and patch releases.

It's not clear whether that mean supporting the **current LTS** version, which is **20**. Or supporting previous version that hasn't reached **end of life yet**, which is **18**.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
```diff
@@ -230,92 +213,82 @@ class Connection extends EventEmitter {
...
-              ),
+              `Server error (no request information): ${response.status.message} (${response.status.code})`,
               response.status,
             ),
           );
         }
         // TINKERPOP-2285: keep the old server error message in case folks are parsing that - fix in a future breaking version
         return handler.callback(
-          new ResponseError(util.format('Server error (no request information): %j', response), response.status),
+          new ResponseError(`Server error (no request information): ${response}`, response.status),
```
kmcginnes, line 291: @tien One thing to keep in mind with template strings is that they do not expand objects. They typically print as `[Object object]`.

You can see an example of that here: https://stackoverflow.com/a/46146887/1558906

One solution could be to use `JSON.stringify()`, but this could potentially lead to issues with circular references, which is inherent to `JSON.stringify()`.

My best guess is that the chances of running in to a circular reference in the `response` object is quite low. In addition, this `ResponseError` is already a fallback to a more specific case. So I would recommend using `JSON.stringify()` here to ensure the error message is usable.

gremlin-javascript/src/main/javascript/gremlin-javascript/.eslintrc.js
```diff
@@ -22,18 +22,17 @@
...
   parserOptions: {
-    ecmaVersion: 2021,
+    ecmaVersion: 2022,
   },
   ignorePatterns: ['test/**/*.js', 'doc/**/*.js'],
   extends: ['eslint:recommended', 'prettier'],
   plugins: ['prettier'],
   rules: {
     'prettier/prettier': ['error', { endOfLine: 'auto' }],
-    indent: ['error', 2, { SwitchCase: 1 }],
```
kmcginnes, line 36: @tien Are the changes to the ESLint rules necessary? I'm specifically concerned about removing the `indent` rule.

tien: Yes, there was an error in this part. Where ESLint was conflicting with prettier.
```ts
    this._ws = new WebSocket(
      this.url,
      globalThis.WebSocket === undefined
        ? {
            headers: headers,
            ca: this.options.ca,
            cert: this.options.cert,
            pfx: this.options.pfx,
            rejectUnauthorized: this.options.rejectUnauthorized,
            agent: this.options.agent,
          }
        : undefined,
    );
```

kmcginnes:
I've reviewed through the code and it looks great overall. Nice work @tien.

I called out a few areas of potential concern that could use some attention. 

It would also be nice to have some automated tests that execute within the context of a browser to ensure we don't regress in the future. Perhaps [Playwright](https://playwright.dev/) can help with this. This is likely to be a large task, so we might want to do it as a separate set of work.

gremlin-javascript/src/main/javascript/gremlin-javascript/.eslintrc.js
kmcginnes: Ah, Prettier is handling things. Nice catch.

tien:
@kmcginnes Thanks for taking a look at this, regarding testing in a browser env, yes that would potentially be a much bigger change than this PR 😅, so I think it's best done as a separate task.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
tien: Yep, good catch, I've updated this part to use `JSON.stringify`

gremlin-javascript/src/main/javascript/gremlin-javascript/test/integration/client-behavior-tests.js
```diff
@@ -21,58 +21,56 @@
...
-    });
-    describe('#submit()', function () {
-        it('should reconnect after server closes connection', async function () {
-            let connectionClosed = false;
-            await client.submit('1', null, {requestId: settings.CLOSE_CONNECTION_REQUEST_ID})
-                .catch(function(error){
-                    assert.equal(error.toString(), 'Error: Connection has been closed.');
-                    connectionClosed = true;
-                });
+  before(function () {
```
kenhuuu, line 30: I noticed that for the test files a lot of the indentation was changed. I'm assuming this was done by a linter but do you mind reverting it back to what it was before. It's a bit difficult to read the Gremlin queries when they are split over multiple lines.

kenhuuu:
You are targetting this PR against the master branch which is currently for the next major release (4.0). The way our branching system works is that you target the lowest version that you want the change to go into. Then when your PR gets merged, the committer will merge your changes all the way up to the latest master branch. It's a bit difficult to explain so I'll use an example. If you want your changes to go into the next 3.6.x, 3.7.x and 4.x release, then you would target the 3.6-dev branch and it will get merged up for you. If you target master then it will only go into the next 4.x release.

I say this because I believe that currently there is no timeline for releasing 4.0 as we are doing releases for 3.6.x and 3.7.x. So if you were to target master then you may need to wait for a while for this change to get released. However, because its currently targetting an unreleased branch, you are technically allowed to upgrade the Node version and ES version.

So, do you prefer to wait for this change to arrive in 4.x? In which case, you can make Node and ES version changes. Or, if you want it released sooner then you will probably have to undo your changes to ESLint and target the 3.7-dev branch as some of those changes could be considered breaking.

gremlin-javascript/src/main/javascript/gremlin-javascript/test/integration/client-behavior-tests.js
tien: Yes, this is from prettier formatting. I've got my editor setup to automatically format on save if the project have prettier setted up, which is project have. I think it's a good idea to enforce formatting rule, even on CI.

But if an exception should be made for the test files, should I add all test files to `.prettierignore` then?

vkagamlyk:
Is there way to simply test these changes in different browsers?

@tien maybe you have a simple application that can be put into [examples](https://github.com/apache/tinkerpop/tree/master/gremlin-javascript/examples) and run to check the basic functionality of the driver? This can also be used as an example for users

gremlin-javascript/src/main/javascript/gremlin-javascript/test/integration/client-behavior-tests.js
tien: But in saying that, I've reverted all the style changes here for now.

tien:
@vkagamlyk the proper way would be to actually run the unit tests in the browser, like @kmcginnes mentioned.

But that could potentially be a huge change to the current test suite.

Another way would be to inject a browser `WebSocket` polyfill into the current test suite using `jsdom`. It's not perfect but would require a lot less changes.

tien:
@vkagamlyk also, because the API signature is the exact same, those existing examples can be run just fine in a browser environment.

tien:
@vkagamlyk or are you suggesting I should create like a simple html page that simply run all the stuffs under `examples`?

vkagamlyk:
> @vkagamlyk or are you suggesting I should create like a simple html page that simply run all the stuffs under `examples`?

I thought about something more simple, like html page which runs single query like `g.V().group().by(label).by(count())` and show results.
This will be useful for those who want to use Gremlin from the browser, and it will allow developers to relatively easily check the functionality of the driver after changes.

tien:
Okay I'll add a simple webpage. There's some flaky test that is failing randomly on GitHub Action it seems btw.

tien:
@vkagamlyk have added a simple webpage that just display your graph running on `localhost` on the screen.

tien:
Modified the README a bit to say this is also compatible with Web APIs compatible environments (browsers, WinterCG runtimes, etc)

vkagamlyk:
@tien I got an error `index.js:22 Uncaught SyntaxError: The requested module '/@fs/C:/_Projects/tinkerpop-tien/gremlin-javascript/src/main/javascript/gremlin-javascript/index.js' does not provide an export named 'default' (at index.js:22:8)` in Chrome.

tien:
> @tien I got an error `index.js:22 Uncaught SyntaxError: The requested module '/@fs/C:/_Projects/tinkerpop-tien/gremlin-javascript/src/main/javascript/gremlin-javascript/index.js' does not provide an export named 'default' (at index.js:22:8)` in Chrome.

Is this from running the example? To run the example, you need to do this at `tinkerpop/gremlin-javascript/examples/browser`:

```sh
yarn
yarn dev
```

vkagamlyk:
> > @tien I got an error `index.js:22 Uncaught SyntaxError: The requested module '/@fs/C:/_Projects/tinkerpop-tien/gremlin-javascript/src/main/javascript/gremlin-javascript/index.js' does not provide an export named 'default' (at index.js:22:8)` in Chrome.
> 
> Is this from running the example? To run the example, you need to do this at `tinkerpop/gremlin-javascript/examples/browser`:
> 
> ```shell
> yarn
> yarn dev
> ```

ahh, I tried with `npm`.
Looks nice
image

vkagamlyk:
VOTE +1

tien:
> VOTE +1

Thanks 🙏, just made a small change, updating the browser example package.json description.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/utils.js
```diff
@@ -112,8 +95,36 @@ exports.getUserAgentHeader = function getUserAgentHeader() {
   return 'User-Agent';
 };
 
-const userAgent = generateUserAgent();
+exports.getUserAgent = async () => {
```
kenhuuu, line 98: Please excuse my lack of knowledge in JavaScript, but it seems like this changes the user agent header that we will be sending to the server. If this is the case, we will want to change this to be more like how the NodeUserAgent works. Our specification for the user agent can be found at https://tinkerpop.apache.org/docs/current/dev/provider/#_graph_driver_provider_requirements . Notably the form should be of "[Application Name] [GLV Name].[Version] [Language Runtime Version] [OS].[Version] [CPU Architecture]". The reason this is important is because the server will likely start using this value in the future to enable/disable features based on the GLV's version.

tien: Yes, but unfortunately to my knowledge, Web API Web Socket doesn't allow you to set custom headers.

tien:
@kenhuuu regarding the user agent header [used](https://tinkerpop.apache.org/docs/current/dev/provider/#_graph_driver_provider_requirements) by Gremlin server

This is likely a separate discussion, but `User-Agent` header might not be the best way to identify client supported GLV version. Since browser environment for example doesn't allow you to set or modify this for web socket connection.

What might be better is using the [`Sec-WebSocket-Protocol`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism#sec-websocket-protocol) header instead. Which is designed to specify the sub-protocol requested by client & can be set in the browser by doing:

```js
new WebSocket("ws://localhost:8182", "glv3.7.0");
```

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/utils.js
kenhuuu: I see. Although it would have been best for this to work the same between the web version and the Node version, I still think its worth getting this PR in so I think this can be overlooked. By the way, in the master branch, we will be removing WebSocket support and transitioning to HTTP so this won't be an issue there anyway.

kenhuuu:
> @kenhuuu regarding the user agent header [used](https://tinkerpop.apache.org/docs/current/dev/provider/#_graph_driver_provider_requirements) by Gremlin server
> 
> This is likely a separate discussion, but `User-Agent` header might not be the best way to identify client supported GLV version. Since browser environment for example doesn't allow you to set or modify this for web socket connection.
> 
> What might be better is using the [`Sec-WebSocket-Protocol`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism#sec-websocket-protocol) header instead. Which is designed to specify the sub-protocol requested by client & can be set in the browser by doing:
> 
> ```js
> new WebSocket("ws://localhost:8182", "glv3.7");
> ```
> 
> The sub-protocol can even be registered [here](https://www.iana.org/assignments/websocket/websocket.xml#subprotocol-name) if needed :))

Thanks for the information. As I mentioned above, we'll be moving to HTTP only in 4.x so its probably best to keep the 3.x WebSocket behavior the same for now.

kenhuuu:
VOTE +1. Thanks again for your willingness to contribute. It is much appreciated.

I'm not sure if you are familiar with our policy but it takes a VOTE+1 from three different committers to allow a PR to get merged immediately. Otherwise, if you get at least one of them (you have two right now), then the PR can be merged one week after the first VOTE +1. So at the latest this PR will be eligible for merging on Monday of next week.

tien:
@kenhuuu thanks for this. I'll try to find another VOTE +1 if possible. Would be beneficial for me to get this in early because there's a work project I'm on that's currently using this.

Cole-Greer:
Thanks for putting all of this together @tien, that's awesome that gremlin-javascript will now work in browsers and I think having that simple little visualizer as an example is really cool. I do have a slight apprehension towards advertising full support for browser runtimes without a proper test suite for it. How would you feel about changing the wording in the readme to something such as `Gremlin-Javascript implements Gremlin within the JavaScript language and can be used on Node.js and has experimental support for Web APIs compatible runtimes`. I would be more comfortable listing it as full support after it has gone through a few release cycles and we have more confidence that browser compatibility is stable and we aren't likely to accidentally break it with the absence of tests.

tien:
@Cole-Greer Yeah I think that make sense, have changed the README accordingly.

Cole-Greer:
Thanks @tien, I think this is great. VOTE +1. I will merge this into both 3.7-dev and master so it will be reflected in the upcoming 3.7.2 and 4.0.0 releases. Just to set some expectations on release timelines, no discussion has yet taken place on when the 3.7.2 release will take place, but based on historical release timing, we typically see a release in the rough April-May window. I would expect that discussions around the next release will begin to take place in the [dev maillist](https://tinkerpop.apache.org/docs/current/dev/developer/#_mailing_list_participation) sometime within the next couple of weeks.

Cole-Greer:
> Just to set some expectations on release timelines, no discussion has yet taken place on when the 3.7.2 release will take place, but based on historical release timing, we typically see a release in the rough April-May window. I would expect that discussions around the next release will begin to take place in the [dev maillist](https://tinkerpop.apache.org/docs/current/dev/developer/#_mailing_list_participation) sometime within the next couple of weeks.

I hadn't checked my email just yet when I wrote this, looks like the release discussion has just started. The current proposal is for a early April release. https://lists.apache.org/thread/8vvo6f74lo1tb7pzclo1pwqbv6xrncl3

PR #2522, opened 2024-03-14
TINKERPOP-3063 Fix bug in Gremlin.Net authentication for parallel requests

FlorianHockmann:
https://issues.apache.org/jira/browse/TINKERPOP-3063

Executing multiple queries in parallel could lead to authentication failures if `MaxInProcessPerConnection` is set to a value higher than `1` as the second request could then be sent to the server while the server was still waiting for the authentication challenge response from the driver for the first query.

This made it necessary to set `MaxInProcessPerConnection=1` as a workaround for such scenarios which of course means that connection pooling is less efficient.

Simply sending a validation request on each connection right after establishing the connection and therefore before it can be used to submit queries from the user should fix this issue. The downside of this is of course that connection establishment takes slightly longer. I think this is an acceptable trade-off even for scenarios where authentication is not used since this also validates the connection irrespective of authentication and also because this delay only occurs once right at the start of an application.

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -24,6 +24,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 === TinkerPop 3.6.7 (NOT OFFICIALLY RELEASED YET)
 
 * Fixed a bug in Gremlin.Net for .NET 8 that led to exceptions: `InvalidOperationException: Enumeration has not started. Call MoveNext.`
+* Fixed a bug in the Gremlin.Net driver that could lead to failed authentication if multiple requests are executed in parallel.
 * Fixed message requestId serialization in `gremlin-python`.
 * Improved performance of `PathRetractionStrategy` for traversals that carry many children, but don't hold many labels to propogate.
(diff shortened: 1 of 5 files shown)


Cole-Greer:
@FlorianHockmann Do you have any thoughts on the server side solution by @tien in https://github.com/apache/tinkerpop/pull/2525? If the server fix fully addresses the issue, then it should make the changes here as well as equivalent GLV changes unnecessary.

FlorianHockmann:
> Do you have any thoughts on the server side solution by @tien in https://github.com/apache/tinkerpop/pull/2525? If the server fix fully addresses the issue, then it should make the changes here as well as equivalent GLV changes unnecessary.

I'm just back from vacation which is why I couldn't review the PR yet.
But I agree in general that a server fix is definitely better in general than implementing workarounds in all GLV drivers. I'll try to review #2525 today.
If it fixes the issue, then we of course don't need this PR any more.

FlorianHockmann:
Closing this as it was superseded by #2525 which fixes the problem on the server side so the workaround I implemented here for the .NET driver isn't needed any more.

PR #2523, opened 2024-03-14
Fixflaky

saivivek116:


Cole-Greer:
Hi @saivivek116, thanks for the submission. Could you give a quick description of the changes? Was the test flaky simply due to the unpredictable ordering of HashSet elements? What does the new workflow do? I'm not familiar with Non-Dex.

If this is just a case of a HashSet being used in a test should use a LinkedHashSet, I think that the coverage from https://github.com/apache/tinkerpop/blob/master/.github/workflows/build-test.yml is enough.

saivivek116:
I apologize for the insufficient details in the PR.  I am currently preparing a new pull request with a comprehensive description and all the necessary details to ensure clarity. Thank you for your patience and understanding

saivivek116:
Thank you for your feedback. The workflow I've proposed is primarily designed to automate the steps I've taken to demonstrate that the test method is flaky. However, upon reflection, I realize it may not add significant value in its current form. Therefore, I will proceed to submit a new pull request that excludes the GitHub workflow actions file, focusing solely on the update from HashSet to LinkedHashSet as discussed.

PR #2524, opened 2024-03-15
Allow aliased client to pass along options when submitting traversals

ryn5:
Refactored `DriverRemoteConnection.getRequestOptions()` to be a public static method in `RequestOptions` so that `Client.submitAsync()` can get and pass along options with it.  This fixes the issue where aliased clients submitting traversals containing `with()` configurations would ignore sending these options as part of the request in https://issues.apache.org/jira/browse/TINKERPOP-3039.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
@@ -19,8 +19,6 @@
 package org.apache.tinkerpop.gremlin.driver;
 
 import org.apache.commons.lang3.exception.ExceptionUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.apache.tinkerpop.gremlin.driver.exception.ConnectionException;
 import org.apache.tinkerpop.gremlin.driver.exception.NoHostAvailableException;
 import org.apache.tinkerpop.gremlin.driver.message.RequestMessage;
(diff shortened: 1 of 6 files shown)


Cole-Greer:
Thanks Ryan,
There are currently failures in the java 8 actions but I don't think it's related to this PR, as we are seeing the same errors in the master branch right now. That error needs separate investigation but that it not related to this PR.

VOTE +1.

vkagamlyk:
VOTE +1

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
```diff
@@ -914,6 +922,26 @@ public Settings create() {
         }
     }
 
+    public static class ClientUtil {
```
kenhuuu, line 925: Why did you make a class to hold this static method?

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/ClientWithOptionsTest.java
```diff
@@ -0,0 +1,87 @@
...
+import static org.apache.tinkerpop.gremlin.process.traversal.AnonymousTraversalSource.traversal;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.fail;
+
+/**
+ * Tests for options sent by the client using with().
+ *
+ * @author Ryan Tan
+ */
+public class ClientWithOptionsTest extends AbstractGremlinServerIntegrationTest {
```
kenhuuu, line 43: Very minor nit: This looks to be a integration test so you should probably name it as <name>IntegrateTest so that is picked up as such by maven.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
ryn5: Moved method to RequestOptions

kenhuuu:
VOTE +1

PR #2525, opened 2024-03-16
[TINKERPOP-3061] fix: failing authentication when multiple initially requests are executed concurrently

tien:
This solution try to resolve the concurrent initial unauthenticated requests problem described in [TINKERPOP-3063](https://issues.apache.org/jira/browse/TINKERPOP-3063), [TINKERPOP-2132](https://issues.apache.org/jira/browse/TINKERPOP-2132) & [TINKERPOP-3061](https://issues.apache.org/jira/browse/TINKERPOP-3061) by batching them for later processing when authentication handshake is in progress.

### DIFF
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/SaslAuthenticationHandler.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/SaslAuthenticationHandler.java
@@ -22,40 +22,44 @@ import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.util.Attribute;
-
-import java.net.InetAddress;
-import java.net.InetSocketAddress;
-import java.net.SocketAddress;
-import java.util.Base64;
-import java.util.HashMap;
-import java.util.Map;
-
-import io.netty.util.AttributeMap;
-import org.apache.tinkerpop.gremlin.util.Tokens;
-import org.apache.tinkerpop.gremlin.util.message.RequestMessage;
-import org.apache.tinkerpop.gremlin.util.message.ResponseMessage;
-import org.apache.tinkerpop.gremlin.util.message.ResponseStatusCode;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.tinkerpop.gremlin.server.GremlinServer;
 import org.apache.tinkerpop.gremlin.server.Settings;
-import org.apache.tinkerpop.gremlin.server.auth.AuthenticatedUser;
 import org.apache.tinkerpop.gremlin.server.auth.AuthenticationException;
 import org.apache.tinkerpop.gremlin.server.auth.Authenticator;
 import org.apache.tinkerpop.gremlin.server.authz.Authorizer;
 import org.apache.tinkerpop.gremlin.server.channel.WebSocketChannelizer;
+import org.apache.tinkerpop.gremlin.util.Tokens;
+import org.apache.tinkerpop.gremlin.util.message.RequestMessage;
+import org.apache.tinkerpop.gremlin.util.message.ResponseMessage;
+import org.apache.tinkerpop.gremlin.util.message.ResponseStatusCode;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
+import java.time.Duration;
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.HashMap;
+import java.util.List;
+import java.util.function.Function;
 
 /**
  * A SASL authentication handler that allows the {@link Authenticator} to be plugged into it. This handler is meant
  * to be used with protocols that process a {@link RequestMessage} such as the {@link WebSocketChannelizer}
  *
- * @author Stephen Mallette (http://stephen.genoprime.com)
+ * @author Stephen Mallette (<a href="http://stephen.genoprime.com">http://stephen.genoprime.com</a>)
  */
 @ChannelHandler.Sharable
 public class SaslAuthenticationHandler extends AbstractAuthenticationHandler {
     private static final Logger logger = LoggerFactory.getLogger(SaslAuthenticationHandler.class);
     private static final Base64.Decoder BASE64_DECODER = Base64.getDecoder();
     private static final Base64.Encoder BASE64_ENCODER = Base64.getEncoder();
+    public static final Duration MAX_REQUEST_DEFERRABLE_DURATION = Duration.ofSeconds(5);
     private static final Logger auditLogger = LoggerFactory.getLogger(GremlinServer.AUDIT_LOGGER_NAME);
 
     protected final Settings settings;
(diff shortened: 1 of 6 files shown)


vkagamlyk:
I think this is a good solution for concurrent auth issue.

Good to have test for js driver to be 100% sure.
Also missing `changelog` entry

tien:
Have added 1 changelog entry & a test to the JS driver 🙏

vkagamlyk:
> Have added 1 changelog entry & a test to the JS driver 🙏

Thank you @tien!

VOTE +1

Cole-Greer:
Thanks @tien, looks great. VOTE +1

kenhuuu:
Could you also add a test where the authentication fails and you check that all requests get the proper exception in that case?

Cole-Greer:
I agree with @kenhuuu that an additional test is warranted here to ensure that the server will always send a response to every request. We are now entering code freeze week in preparation for the 3.6.7 and 3.7.2 releases. I believe it is fair to grant an exception for a few days to give some time for such a test to be implemented and to ensure this PR can be included in the release.

tien:
I've added one test for the unhappy path.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/SaslAuthenticationHandler.java
```diff
@@ -75,106 +79,159 @@ public SaslAuthenticationHandler(final Authenticator authenticator, final Author
...
+                    requestMessage,
+                    builder -> builder.statusMessage("Authenticator is not ready to handle requests").code(ResponseStatusCode.SERVER_ERROR),
+                    ctx);
+            }
+
+            return;
         }
+
+        // If authentication negotiation is pending, store subsequent non-authentication requests for later processing
+        if (negotiator.get() != null && !requestMessage.getOp().equals(Tokens.OPS_AUTHENTICATION)) {
```
FlorianHockmann, line 120: (nitpick) Isn't `negotiator.get() != null` duplicate code since we have an `if (negotiator.get() == null)` right before that returns ?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/SaslAuthenticationHandler.java
```diff
@@ -75,106 +79,159 @@ public SaslAuthenticationHandler(final Authenticator authenticator, final Author
...
+            deferredRequests.setIfAbsent(new ImmutablePair<>(LocalDateTime.now(), new ArrayList<>()));
+            deferredRequests.get().getValue().add(requestMessage);
+
+            final Duration deferredDuration = Duration.between(deferredRequests.get().getKey(), LocalDateTime.now());
+
+            if (deferredDuration.compareTo(MAX_REQUEST_DEFERRABLE_DURATION) > 0) {
+                respondWithError(
+                    requestMessage,
+                    builder -> builder.statusMessage("Too many unauthenticated requests").code(ResponseStatusCode.TOO_MANY_REQUESTS),
```
FlorianHockmann, line 129: Is the problem here really that there are _too many unauthenticated requests_? Isn't the problem that authentication took longer than `MAX_REQUEST_DEFERRABLE_DURATION`?
I think as a user it might be good to know whether I simply submitted too many requests or whether authentication is just too slow.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerAuthIntegrateTest.java
```diff
@@ -218,9 +266,19 @@ public void shouldAuthenticateAndWorkWithVariablesOverGraphSONV1Serialization()
 
     private static void assertConnection(final Cluster cluster, final Client client) throws InterruptedException, ExecutionException {
```
FlorianHockmann, line 267: This method is used in 4 different tests, such as `shouldAuthenticateWithPlainText`. These 4 tests will now fail if submitting multiple requests initially in parallel isn't working.
I think it would be good if we could keep these tests as simple as possible so they don't include parallelization of initial requests. A test like `shouldAuthenticateWithPlainText` should really only fail if _authenticate with plain text_ isn't working, not if submitting multiple requests in parallel isn't working.

Long story short, I think it would be good if you could revert the changes to this method and instead write a new test specifically for the parallelization issue.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerAuthIntegrateTest.java
```diff
@@ -163,6 +171,46 @@ public void shouldFailAuthenticateWithPlainTextBadUsername() throws Exception {
...
+
+            List<ResponseMessage> responses = new ArrayList<>();
+
+            responses.addAll(future1.join());
+            responses.addAll(future2.join());
+            responses.addAll(future3.join());
+            responses.addAll(future4.join());
+
+            for (ResponseMessage response : responses) {
+                if (response.getStatus().getCode() != ResponseStatusCode.AUTHENTICATE) {
```
FlorianHockmann, line 207: I think it's a bit hard to understand what the expected outcome is here / what this is really asserting since it's just iterating over all received responses and then either accepting them if their status code is `AUTHENTICATE` or if it's `TOO_MANY_REQUESTS`.
Can't we explicitly assert which request should get which response status code? I guess `future4` should get `TOO_MANY_REQUESTS` and the other 3 should get `AUTHENTICATE` (?).

Not that important, but I think it would also improve readability of this test if these weren't named `future1` - `future4`, but maybe something like `futureOfRequestWithinAuthDuration` vs `futureOfRequestSubmittedTooLate` or something like that.

FlorianHockmann (commented):
Just reviewed the changes. I like the general approach taken here! It means that we don't have to implement workarounds in all GLVs for this like I initially did in #2522. Thanks a lot for tackling this, @tien!

I added some inline comments, but they are all basically only about code style issues.
I also tested this fix with the test that I wrote for the .NET driver in #2522 and it passed! :-)

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/SaslAuthenticationHandler.java
tien: I'm stumped on this 2, do you have a recommendation on what status code & message would make sense here?

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerAuthIntegrateTest.java
```diff
@@ -163,6 +171,46 @@ public void shouldFailAuthenticateWithPlainTextBadUsername() throws Exception {
         }
     }
 
+    @Test
+    public void shouldFailAuthenticateWithUnAuthenticatedRequestAfterMaxDeferrableDuration() throws Exception {
```
Cole-Greer, line 175: Hi Tien, I want to be absolutely certain that we aren’t going to lose any of these deferred requests in the case of errors. If the server fails to send a response the drivers will just be left hanging indefinitely. If I’m understanding this test right, the first 3 requests are all expected to succeed, and then after a delay the final request is submitted and fails. Would it also be possible to setup a test such that there are multiple pending requests in the server’s deferred requests queue at the time that auth fails, and then we can verify that the correct error message gets sent to all currently pending requests?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/SaslAuthenticationHandler.java
FlorianHockmann: Wouldn't `UNAUTHORIZED` be correct in this case? The description starts with:
> The server could not authenticate the request

which makes sense in my opinion if the max duration was passed for the authentication to happen.

And the status message can then contain more detailed information about the duration that was passed, maybe something like: `"authentication did not finish in the allowed duration (" + MAX_REQUEST_DEFERRABLE_DURATION + " s)"`?

tien:
Sorry, I'm away this week and don't have access to my work laptop. Will take a look at all the pending comments & resolve them next week 🙏

xiazcy:
> Sorry, I'm away this week and don't have access to my work laptop. Will take a look at all the pending comments & resolve them next week 🙏

No worries, thank you for all the contributions! 

Just a quick note. Not sure if you have gotten a chance to start looking at the comments, as we'd like to release this with 3.7.2 this week, we will likely be cherry-picking your changes into another PR for the release branch today. If we do proceed with that we'll be closing this PR, and you shouldn't need to do any further work.

Now there might still be functionality improvements we miss, so please feel free to add additional changes once the branches re-open.

xiazcy:
Closing this PR as all changes are merged via 22db8cf for the release. Please feel free to re-open if you find additional improvements and/or updates needed. Thanks!

tien:
@xiazcy nah I just got back today, thanks for making this available in the next release 💪

PR #2526, opened 2024-03-20
Add chunked transfer encoding to HTTP Channelizer.

kenhuuu:
The HttpGremlinEndpointHandler will now chunk responses for both script and bytecode traversals. Much of the implementation is taken from the StandardOpProcessor/TraversalOpProcessor/UnifiedHandler.

Note that there are test failures due to the change in error handling which is not reflected in the GLVs so certain GLV integration tests will be failing. This will be updated in the future when the GLVs are updated to handle the new error mechanism sent by the server.


kenhuuu:

PR #2528, opened 2024-03-21
Remove the UnifiedChannelizer/UnifiedHandler and associated classes.

kenhuuu:
Removing this as part of the server maintenance cleanup. See https://lists.apache.org/thread/vfs1j9ycb8voxwc00gdzfmlg2gghx3n1

Again, tests will be failing as the GLVs aren't setup to test the new HTTP API.


kenhuuu:

PR #2529, opened 2024-03-25
Temporary fix to failing JDK8 test

ryn5:
A new SelectStep test is failing in JDK8 tests due to a difference in hashing between Java 8 and 11.  In Java 8, the hash used to identify the traverser to be removed from the no-op barrier is recomputed at removal, and can't find a match due to other traversal branches having added vertices to the bulkset, thus changing its potential hash value.  Java 11 simply uses the hash field stored at creation instead of recomputing it which avoids this problem.  

This additional barrier step is a temporary fix that helps the query wait until all the vertices have been added to the bulkset before adding traversals to the no-op barrier so that hashing stays consistent.


Cole-Greer:
Thanks Ryan! LGTM VOTE +1

PR #2532, opened 2024-03-29
Request message serde

kenhuuu:


kenhuuu:
Merging into feature branch so votes not needed.

PR #2536, opened 2024-04-02
Disable LazyBarrierStrategy for failing JDK8 test

ryn5:
Temporary fix that preserves the select feature test case, updated from #2529.


Cole-Greer:
Thanks Ryan, LGTM. Merging as CTR.

PR #2546, opened 2024-04-08
[TINKERPOP-3060] Manage the versions of snappy-java and jackson-* centrally to avoid version conflict.

HappyHacker123:
Currently some modules of Tinkerpop are using the same dependency. However, some of these dependencies' versions are not centrally managed and therefore cause discrepancy. This pr defines version in parent pom.xml centrally to avoid version conflict.

### DIFF
--- a/hadoop-gremlin/pom.xml
+++ b/hadoop-gremlin/pom.xml
@@ -123,7 +123,7 @@ limitations under the License.
         <dependency>
             <groupId>org.xerial.snappy</groupId>
             <artifactId>snappy-java</artifactId>
-            <version>1.1.8.2</version>
+            <version>${snappy-java.version}</version>
         </dependency>
         <dependency>
             <groupId>com.nimbusds</groupId>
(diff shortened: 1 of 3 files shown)


HappyHacker123:
@xiazcy @Cole-Greer Could you please help me review this pr? Many thanks :)

xiazcy:
Thanks for the PR @HappyHacker123! I like the improvement for the dependency version management. I do want to take a closer look into other pom files that might be able to take advantage of this. Note that we are in the middle of a release so we'll hold off on any merging and I'll revisit once this PR the release is done.

HappyHacker123:
Thanks for your response :) @xiazcy . 
> Note that we are in the middle of a release so we'll hold off on any merging and I'll revisit this PR once the release is done.
Sure, I'm fully supportive of holding off the pr until after the release is completed :).
> I do want to take a closer look into other pom files that might be able to take advantage of this.
Actually i ran a check on the whole Tinkerpop project and found many other situation like this. If you need it, i'm happy to provide the whole list :).

xiazcy:
Hi @HappyHacker123, much apologies for the late reply. I gave it some more thoughts, and while this will be a great improvement, I don't think we are quite ready to centralize these dependencies yet. The complexity is mainly due to the way dependencies are currently managed that’s leading to conflicts and may require module-specific versions, especially in hadoop and spark modules (as you may see some of the code comments in the poms). We are looking to improve this systemically, but before a decision is reached I think we’ll hold off on this PR and close it for now. 

If you are still interested and have thoughts or recommendations related to dependency management, chime into our discussions for [removing gremlin-shaded](https://lists.apache.org/thread/m7yvvz9zc69nzn0nr6vz0tdo3ggffoy6) and [JPMS](https://lists.apache.org/thread/vkvqcxn6t2lsgpfym7v3g3qvksrh3cwc). We'd love to see more opinions on these topics. Again, thank you for taking the time submitting this PR!

PR #2569, opened 2024-04-19
Add back evaluationTimeout and materializeProperties

kenhuuu:
evaluationTimeout and materializeProperties should be allowed to be set from the request body as it affects how the server should execute the query.

### DIFF
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Context.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Context.java
@@ -200,13 +200,15 @@ public class Context {
     private long determineTimeout() {
         // timeout override - handle both deprecated and newly named configuration. earlier logic should prevent
         // both configurations from being submitted at the same time
-        // evaluationTimeout RequestMessage argument was removed starting in 4.0.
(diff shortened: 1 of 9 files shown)


gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Context.java
```diff
@@ -219,8 +221,11 @@ private String determineMaterializeProperties() {
                         : Tokens.MATERIALIZE_PROPERTIES_ALL;
         }
 
+        final String materializeProperties = requestMessage.getField(Tokens.ARGS_MATERIALIZE_PROPERTIES);
         // all options except MATERIALIZE_PROPERTIES_TOKENS treated as MATERIALIZE_PROPERTIES_ALL
-        return Tokens.MATERIALIZE_PROPERTIES_ALL;
+        return (null != materializeProperties) && materializeProperties.equals(Tokens.MATERIALIZE_PROPERTIES_TOKENS)
```
vkagamlyk, line 226: nit: can be  simplified to `Tokens.MATERIALIZE_PROPERTIES_TOKENS.equals(materializeProperties)`

PR #2599, opened 2024-05-09
feat: typescript utilities

tien:
- Improve structures type definitions
```ts
type Person = Vertex<"person", { name: string; age: number }>;
type Pet = Vertex<"pet", { name: string; age: number }>;
type Owns = Edge<Person, "owns", Pet, { since: Date }>;
```

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.ts
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.ts
@@ -30,11 +30,10 @@ import { TraversalStrategies } from '../process/traversal-strategy.js';
 export class Graph {
   /**
    * Returns the graph traversal source.
-   * @param {Function} [traversalSourceClass] The constructor to use for the {@code GraphTraversalSource} instance.
-   * @returns {GraphTraversalSource}
+   * @param TraversalSourceClass The constructor to use for the {@code GraphTraversalSource} instance.
    * @deprecated As of release 3.3.5, replaced by the traversal() anonymous function.
    */
-  traversal(TraversalSourceClass: typeof GraphTraversalSource = GraphTraversalSource) {
+  traversal(TraversalSourceClass: typeof GraphTraversalSource = GraphTraversalSource): GraphTraversalSource {
     return new TraversalSourceClass(this, new TraversalStrategies());
   }
 
@@ -43,27 +42,32 @@ export class Graph {
   }
 }
 
-class Element {
+class Element<TLabel extends string = string, TId = any> {
   constructor(
-    readonly id: string,
-    readonly label: string,
+    readonly id: TId,
+    readonly label: TLabel,
   ) {}
 
   /**
    * Compares this instance to another and determines if they can be considered as equal.
-   * @param {Element} other
-   * @returns {boolean}
    */
-  equals(other: any) {
+  equals(other: any): boolean {
     return other instanceof Element && this.id === other.id;
   }
 }
 
-export class Vertex<T extends Record<string, any> = Record<string, any>> extends Element {
+export class Vertex<
+  TLabel extends string = string,
+  TProperties extends Record<string, any> = Record<string, any>,
+  TId = number,
+  TVertexProperties = {
+    [P in keyof TProperties]: P extends string ? VertexProperties<P, TProperties[P]> : never;
+  },
+> extends Element<TLabel, TId> {
   constructor(
-    id: string,
-    label: string,
-    readonly properties?: T,
+    id: TId,
+    label: TLabel,
+    readonly properties?: TVertexProperties,
   ) {
     super(id, label);
   }
@@ -73,20 +77,26 @@ export class Vertex<T extends Record<string, any> = Record<string, any>> extends
   }
 }
 
-export class Edge<T extends Record<string, any> = Record<string, any>> extends Element {
+export class Edge<
+  TOutVertex extends Vertex = Vertex,
+  TLabel extends string = string,
+  TInVertex extends Vertex = Vertex,
+  TProperties extends Record<string, any> = Record<string, any>,
+  TId = number,
+> extends Element<TLabel, TId> {
   constructor(
-    id: string,
-    readonly outV: Element,
-    label: string,
-    readonly inV: Element,
-    readonly properties: T = {} as T,
+    id: TId,
+    readonly outV: TOutVertex,
+    readonly label: TLabel,
+    readonly inV: TInVertex,
+    readonly properties: TProperties = {} as TProperties,
   ) {
     super(id, label);
     if (properties) {
       const keys = Object.keys(properties);
       for (let i = 0; i < keys.length; i++) {
         const k = keys[i];
-        this.properties[k as keyof T] = properties[k].value;
+        this.properties![k as keyof TProperties] = properties[k].value;
       }
     }
   }
@@ -99,14 +109,19 @@ export class Edge<T extends Record<string, any> = Record<string, any>> extends E
   }
 }
 
-export class VertexProperty<T1 = any, T2 = any> extends Element {
+export class VertexProperty<
+  TLabel extends string = string,
+  TValue = any,
+  TProperties extends Record<string, any> | null | undefined = Record<string, any>,
+  TId = any,
+> extends Element<TLabel, TId> {
   readonly key: string;
 
   constructor(
-    id: string,
-    label: string,
-    readonly value: T1,
-    readonly properties: T2,
+    id: TId,
+    label: TLabel,
+    readonly value: TValue,
+    readonly properties?: TProperties,
   ) {
     super(id, label);
     this.value = value;
@@ -118,6 +133,13 @@ export class VertexProperty<T1 = any, T2 = any> extends Element {
     return `vp[${this.label}->${summarize(this.value)}]`;
   }
 }
+
+export type VertexProperties<
+  TLabel extends string = string,
+  TValue = any,
+  TProperties extends Record<string, any> = Record<string, any>,
+  TId = any,
+> = [VertexProperty<TLabel, TValue, TProperties, TId>, ...Array<VertexProperty<TLabel, TValue, TProperties, TId>>];
 
 export class Property<T = any> {
   constructor(
@@ -134,13 +156,10 @@ export class Property<T = any> {
   }
 }
 
+/**
+ * Represents a walk through a graph as defined by a traversal.
+ */
 export class Path {
-  /**
-   * Represents a walk through a graph as defined by a traversal.
-   * @param {Array} labels
-   * @param {Array} objects
-   * @constructor
-   */
   constructor(
     readonly labels: string[],
     readonly objects: any[],
--- a/gremlin-javascript/examples/browser/yarn.lock
+++ b/gremlin-javascript/examples/browser/yarn.lock
@@ -729,14 +729,14 @@ __metadata:
 
 "gremlin@file:../../src/main/javascript/gremlin-javascript::locator=browser%40workspace%3A.":
   version: 4.0.0-alpha1
-  resolution: "gremlin@file:../../src/main/javascript/gremlin-javascript#../../src/main/javascript/gremlin-javascript::hash=83bc8e&locator=browser%40workspace%3A."
+  resolution: "gremlin@file:../../src/main/javascript/gremlin-javascript#../../src/main/javascript/gremlin-javascript::hash=7358f9&locator=browser%40workspace%3A."
   dependencies:
     buffer: "npm:^6.0.3"
     eventemitter3: "npm:^5.0.1"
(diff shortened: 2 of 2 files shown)


FlorianHockmann:
If I'm not mistaken, then this PR does two things which are mostly separate from each other, right? It extends the type definitions for existing types like `Vertex`, `Edge`, and `VertexProperty` and it also adds a new utility method `parseVertex` which can be used to create a `Vertex` object where a `VertexProperty` is directly a property of the object, e.g., `vertex.name`.

I think it would be good to separate these two things from each other. Could you please create issues for them in our JIRA? Our usual development workflow is to first have an issue there which describes the bug / proposed improvement and afterwards create the PR. This also makes it possible to first discuss the issue before actually implementing it.

In this case, I'm not 100% about the added utility method. It sounds to me like it's a first step in the direction of the [_Sugar Plugin_](https://tinkerpop.apache.org/docs/current/reference/#sugar-plugin) for Groovy and I'm not sure whether we should implement such syntactic sugar for the GLVs like gremlin-javascript.
But maybe I'm just missing context why this is useful for Javascript / Typescript. You could probably better describe your motivation for this in the JIRA issue.

Sorry for the slight push back here in general. Your Typescript improvements are more than welcome, but I think they are difficult for us to review since we simply lack some experience in this area. That is why I think that following this process helps as it makes it easier for other contributors to better understand the improvements before reviewing them.

Cole-Greer:
Hi @tien, thanks for putting this together.

I agree with Florian that for this type of change, it is best to start with a JIRA or dev list post before starting implementation.

I do think that consistency between the GLV's is important, but I can also see how this could be a valuable convenience for users. I do think if we are to add this for vertices in JS, we need to do the equivalent for edges.

In terms of naming, an alternative worth considering might be `flattenVertex` but this isn't something I feel strongly about.

The overall implementation looks good to me, although I wonder if the need to provide explicit cardinality schemas for anything other than singles will limit the convenience upside here. It would be cool if there was an option to try and detect the cardinality of each property, and only force a conversion if the user explicitly specifies an override. Granted this would currently be limited to just single and list, as set is currently blocked by [TINKERPOP-3011](https://issues.apache.org/jira/browse/TINKERPOP-3011)

In general, if this is something that JS users want in the GLV, then I'm willing to put aside some of the cross-GLV consistency concerns here. I think it's worth a quick dev list thread just to ensure the design is aligned with the community.

tien:
Hey guys, thanks for the feedbacks, yes that make sense. So I've reduced the scope of the PR change to just improving the TypeScript definition for graph structures: vertex, edge, etc

tien:
I mainly included the utility function as a demonstration of what a strong TypeScript definition can enable.

As for the benefit of having an utility function like that, with TypeScript I have to do a lot of boilerplates just extract basic values from vertices, i.e:

```ts
const vertices = await g.V().hasLabel("person").toList<PersonVertex>();

const people = vertices.map(vertex => {
  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
  const age = vertex.properties.age.at(0)!.value;

  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
  const name = vertex.properties.name.at(0)!.value;
  
  const addresses = vertex.properties.address.map(x => x.value);

  return {
    age,
    name,
    addresses
  };
});
```

which with the utility function can be simplified to

```ts
const vertices = await g.V().hasLabel("person").toList<PersonVertex>();
const people = vertices.map(vertex => parseVertex(vertex, { address: "list" }));
```

tien:
I'm happy to pursue this with a Jira ticket or via the dev list as mentioned if you guys think a utility like this is valuable to be included in the core gremlin lib.

vkagamlyk:
Hi @tien,
I think this is a good direction for improvement for TinkerPop 5/6 when/if we introduce schema, but now it is important to have similar capabilities in all GLV's.

Perhaps it makes sense to make an example from this code? Or some utility class?

this is really beautiful code, but my opinion is VOTE+0

tien:
Hey @vkagamlyk, as per my comment [here](https://github.com/apache/tinkerpop/pull/2599#issuecomment-2118831749) I've removed all none GLV's feature parity changes, so this PR will just be improving the TypeScript definition. No feature changes are involved.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/io/type-serializers.ts
```diff
@@ -349,9 +349,9 @@ export class EdgeSerializer extends TypeSerializer<g.Edge> {
       [valueKey]: {
         id: this.writer.adaptObject(item.id),
         label: item.label,
-        outV: this.writer.adaptObject(item.outV.id),
+        outV: this.writer.adaptObject(item.outV?.id),
```
FlorianHockmann, line 352: Can `outV` really be `undefined`? And shouldn't users get an exception here if it is?

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.ts
```diff
@@ -73,20 +77,26 @@ export class Vertex<T extends Record<string, any> = Record<string, any>> extends
   }
 }
 
-export class Edge<T extends Record<string, any> = Record<string, any>> extends Element {
+export class Edge<
+  TOutVertex extends Vertex = Vertex,
```
FlorianHockmann, line 81: Why `TOutVertex extends Vertex`? Why not simply `Vertex`?

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.ts
```diff
@@ -73,20 +77,26 @@ export class Vertex<T extends Record<string, any> = Record<string, any>> extends
...
+  TLabel extends string = string,
+  TInVertex extends Vertex = Vertex,
+  TProperties extends Record<string, any> = Record<string, any>,
+  TId extends number | string = number,
```
FlorianHockmann, line 85: I think this is too restrictive. Providers can use any type they like for the ID. JanusGraph for example has its own type.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.ts
```diff
@@ -119,6 +134,13 @@ export class VertexProperty<T1 = any, T2 = any> extends Element {
   }
 }
 
+export type VertexProperties<
+  TLabel extends string = string,
+  TValue = any,
+  TProperties extends Record<string, any> = Record<string, any>,
+  TId = any,
+> = [VertexProperty<TLabel, TValue, TProperties, TId>, ...Array<VertexProperty<TLabel, TValue, TProperties, TId>>];
```
FlorianHockmann, line 142: Sorry, but I don't know this notation and it's a bit hard to find any information on it as I don't even know what to search for here 😅 
Would you be so kind to provide me with some documentation for this kind of type definition or maybe the name of the concept for this so I can search for it myself? :-)

FlorianHockmann (commented):
I just tried to review this, but I have to say that my TypeScript knowledge is quite limited. So most of my comments are really only questions of understanding.

In some cases, your type definitions are also more strict than what we have in other GLVs, and especially in Java which should be the reference.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.ts
```diff
@@ -43,27 +42,32 @@ export class Graph {
   }
 }
 
-class Element {
+class Element<TLabel extends string = string, TId = any> {
   constructor(
-    readonly id: string,
-    readonly label: string,
+    readonly id: TId,
+    readonly label: TLabel,
```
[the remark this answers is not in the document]
tien: In TypeScript, we have string literal, which can be a generic, example:

```ts
type Foo<T> = T;

type Bar = Foo<"string-literal">;
```

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.ts
tien: The extends part will restrict the input type to `Vertex`, while the `= Vertex` part will default that generic to a generic Vertex if no type is provided.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.ts
tien: I believe you are referring to this part?

```ts
[VertexProperty<TLabel, TValue, TProperties, TId>, ...Array<VertexProperty<TLabel, TValue, TProperties, TId>>]
```

It basically will ensure the array will have at least 1 element.

You can find more information regarding it [here](https://www.typescriptlang.org/docs/handbook/variable-declarations.html#spread) :D

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.ts
tien: Thanks, I've removed the type restriction.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/io/type-serializers.ts
tien: Oops, nice catch, I've removed the unnecessary optional chaining.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.ts
```diff
@@ -43,27 +42,32 @@ export class Graph {
...
-  equals(other: any) {
+  equals(other: any): boolean {
     return other instanceof Element && this.id === other.id;
   }
 }
 
-export class Vertex<T extends Record<string, any> = Record<string, any>> extends Element {
+export class Vertex<
+  TLabel extends string = string,
+  TProperties extends Record<string, any> = Record<string, any>,
```
joshsh, line 61: This looks like good work, but I don't know TypeScript. Quick question to get oriented: is optionality in property types supported, and if so, how? I.e. can you declare a certain vertex type to have a mandatory string-valued "name" property, and another vertex type to have an optional string-valued "name" property?

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.ts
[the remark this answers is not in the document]
joshsh: I'm not sure I follow this, either, but I want to understand. What is the difference between string-literal and string? I think of both vertex labels and edge labels as simply being strings, or being of a special vertex-label or edge-label type which contains string.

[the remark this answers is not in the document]
tien: TypeScript string literal are kinda like enum in Java or C#, so for example:

```ts
type LiteralPerson = Vertex<"person", { name: string; age: number }>;
type LoosePerson = Vertex<string, { name: string; age: number }>;

let person1: LiteralPerson // person1.label will have a const type of `"person"`
let person2: LoosePerson // person2.label will instead only be typed as `string`
```

source: https://www.typescriptlang.org/docs/handbook/literal-types.html#string-literal-types

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.ts
tien: Yes, you can do something like this:

```ts
type Human = Vertex<"person", { name: string; }>;
type NewBorn = Vertex<"person", { name?: string; }>;
```

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.ts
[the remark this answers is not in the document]
FlorianHockmann: I see, so users who don't define their own types (like `Person`) will simply have a `Vertex` where `label` is a `string`, but users can restrict the `label` to their own allowed set of label if they want to.
Thanks for the explanation! I'm marking this as resolved now. @joshsh feel free to unresolve if you more questions about this.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.ts
FlorianHockmann: Interesting, didn't know the spread operator. Thanks for the explanation :)

FlorianHockmann (approved):
Thanks a lot for your explanations and your patience here, @tien!

VOTE +1

xiazcy:
Thank you for the contribution! As lazy consensus has been reached, I'll be merging this PR today if there are no further comments. 

One small note, it does look like this PR is missing a CHANGELOG entry, which I'm happy to add it after merging as CTR.

PR #2600, opened 2024-05-11

upadhyay-prashant:
CAUSE:
 post this.

 FIX:
 Ignore the ArrayOutOfBoundException in trying to get an object from
 connections object in case of creating a poolInfo message. This way,
 the code tries to provide the accurate poolInfo best efforts basis.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
@@ -665,13 +665,22 @@ final class ConnectionPool {
                                    final List<Connection> connections) {
         final int connectionCount = connections.size();
         for (int ix = 0; ix < connectionCount; ix++) {
+            try {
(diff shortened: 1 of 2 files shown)


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
```diff
@@ -665,13 +665,22 @@ private void appendConnections(final StringBuilder sb, final Connection connecti
...
+                if (c.equals(connectionToCallout)) {
                 sb.append("==> ");
-            else
+                }
+                else {
                 sb.append("> ");
-
+                }
             sb.append(c.getConnectionInfo(false));
+            } catch (final ArrayIndexOutOfBoundsException ex) {
```
spmallette, line 677: i mentioned this to @upadhyay-prashant elsewhere but others looking at this, i think we want to try to handle this differently as missing logs in the event of failure might prove to be confusing when debugging.

PR #2601, opened 2024-05-13
sigv4 auth client implementation

vkagamlyk:


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Auth.java
```diff
@@ -18,33 +18,258 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static com.amazonaws.auth.internal.SignerConstants.AUTHORIZATION;
+import static com.amazonaws.auth.internal.SignerConstants.HOST;
+import static com.amazonaws.auth.internal.SignerConstants.X_AMZ_DATE;
+import static com.amazonaws.auth.internal.SignerConstants.X_AMZ_SECURITY_TOKEN;
 
 public abstract class Auth implements RequestInterceptor {
```
kenhuuu, line 53: I'm not sure if this class makes sense anymore. Each implementation like Basic might be better off just implementing RequestInterceptor for now and having Auth be more like a factory/builder type of class.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Auth.java
```diff
@@ -18,33 +18,258 @@
...
+                // todo: confirm is needed to replace header `Host` with `host`
+                fullHttpRequest.headers().remove(HttpHeaderNames.HOST);
+                fullHttpRequest.headers().add(HOST, awsSignableRequest.getHeaders().get(HOST));
+                fullHttpRequest.headers().add(X_AMZ_DATE, awsSignableRequest.getHeaders().get(X_AMZ_DATE));
+                fullHttpRequest.headers().add(AUTHORIZATION, awsSignableRequest.getHeaders().get(AUTHORIZATION));
+
+                if (!sessionToken.isEmpty()) {
+                    fullHttpRequest.headers().add(X_AMZ_SECURITY_TOKEN, sessionToken);
+                }
+            } catch (final Throwable t) {
```
kenhuuu, line 125: We probably need slightly better handling of errors here rather than just RuntimeException.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Auth.java
```diff
@@ -18,33 +18,258 @@
...
+            return convertToSignableRequest(
+                    request.method().name(),
+                    endpointUri,
+                    uri.getPath(),
+                    headersInternal,
+                    parametersInternal,
+                    content);
+        }
+
+        private Map<String, List<String>> extractParametersFromQueryString(final String queryStr) {
```
kenhuuu, line 207: I don't know if we need to support this right now.

PR #2609, opened 2024-05-17
Refactor mutation events registration by moving reusable code from relevant steps to EventUti

porunov:
This will allow to reuse callback registration logic for extended classes of mutation steps.

Related discussion where the conclusion was made to relax individual steps based on necessity: https://lists.apache.org/thread/vjbjh29kwjhd5lcmkqqqqrhw7rw2ynh9

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,7 +25,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 This release also includes changes from <<release-3-6-8, 3.6.8>>.
 
-
+* Refactored mutation events registration by moving reusable code from relevant steps to `EventUtil`
 
 [[release-3-7-2]]
 === TinkerPop 3.7.2 (April 8, 2024)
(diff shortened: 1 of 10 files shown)


spmallette:
do you think there should be utilities for all of these sorts of things? like [this similar sort of code](https://github.com/apache/tinkerpop/blob/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AddVertexStep.java#L95-L99) in `AddVertexStep`.

porunov:
> do you think there should be utilities for all of these sorts of things? like [this similar sort of code](https://github.com/apache/tinkerpop/blob/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AddVertexStep.java#L95-L99) in `AddVertexStep`.

I think the more we can make logic reusable the better. So, I think we can make utilities for all the relevant places where such event registration is used.

xiazcy:
I'm good with the change, but I have similar question as Stephen. Is it worth adding one extra util file just for the removed events registration? I don't quite see these method used elsewhere.

For example, when refactoring for other types of event registration, it might be too much to `AddedEventUtil` or `ChangedEventUtil`. Perhaps renaming `RemovedEventUtil` to something more general, like `EventUtil`, so we can group refactored methods together for all event types?

porunov:
> I'm good with the change, but I have similar question as Stephen. Is it worth adding one extra util file just for the removed events registration? I don't quite see these method used elsewhere.
> 
> For example, when refactoring for other types of event registration, it might be too much to `AddedEventUtil` or `ChangedEventUtil`. Perhaps renaming `RemovedEventUtil` to something more general, like `EventUtil`, so we can group refactored methods together for all event types?

Definitely makes sense 👍 I will update it tomorrow.

porunov:
@spmallette @xiazcy Pushed an update to make this PR more general. Refactored all places with events registration logic by moving reusable code to `EventUtil`.

spmallette:
thanks - added in a couple follow-on bits for formatting/javadoc: bd5836a5bcc8b12b7653c833420171b71119fb0f

PR #2610, opened 2024-05-17
Fix nullability checks in .net deserializers

vkagamlyk:


gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/EdgeDeserializer.cs
```diff
@@ -31,26 +31,26 @@ internal class EdgeDeserializer : IGraphSONDeserializer
         public dynamic Objectify(JsonElement graphsonObject, GraphSONReader reader)
         {
             var outVId = reader.ToObject(graphsonObject.GetProperty("outV"));
-            var outVLabel = graphsonObject.TryGetProperty("outVLabel", out var outVLabelProperty)
-                ? outVLabelProperty.GetString()
+            string outVLabel = graphsonObject.TryGetProperty("outVLabel", out var outVLabelProperty)
```
FlorianHockmann, line 34: (nitpick) The change from `var` to `string` shouldn't be necessary and I'd like to avoid that. With the added `!` behind `GetString()`, my IDE already shows that `outVLabel` is of type `string` if I keep `var` here.
(The same applies to the other places in this diff of course.)

FlorianHockmann (approved):
LGTM, I just left one nitpick comment inline.

Do you know why this suddenly caused our builds to fail? The code here hasn't changed in over a year if I'm not mistaken.

VOTE +1

gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/EdgeDeserializer.cs
vkagamlyk: Visual Studio 17.9.5 think `GetString()!` is not enough
image
(formatting changed so that everything would fit)

vkagamlyk:
> Do you know why this suddenly caused our builds to fail? The code here hasn't changed in over a year if I'm not mistaken.
> 

I assume that this is due to some updates in .net in build pipeline.

We have enabled `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>`  for `Gramlin.Net` project for a long time, so this is not the reason.
We can play with `<WarningsAsErrors>nullable</WarningsAsErrors>` or `<WarningsNotAsErrors>nullable</WarningsNotAsErrors>`, but now this is not in the project

xiazcy:
VOTE +1

vkagamlyk:
VOTE +1

gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/EdgeDeserializer.cs
FlorianHockmann: > Visual Studio 17.9.5 think GetString()! is not enough

OK, I see. I'm using Rider and that even complains if `string` is used instead of `var`, but then we can stick with `string` of course.

PR #2614, opened 2024-05-21

upadhyay-prashant:
CAUSE:
 post this.

 FIX:
 Iterate the connections list while creating the connectionPool Info.
 Since the list used is copyOnWrite, the iterator API creates a clone
 and uses that clone for referring the element. Thus providing thread
 safe interface.

 However the information provided by this iteration is a bit stale, but
 this doesn't matter.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
@@ -26,6 +26,7 @@ import org.apache.tinkerpop.gremlin.util.message.RequestMessage;
 import org.apache.tinkerpop.gremlin.util.TimeUtil;
 
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


kenhuuu:
Could you rebase and retarget this PR to point to 3.6-dev? This change most likely applies to version 3.6.x and above rather than 4.0.x (master).

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
```diff
@@ -663,18 +664,19 @@ public String getPoolInfo(final Connection connectionToCallout) {
 
     private void appendConnections(final StringBuilder sb, final Connection connectionToCallout,
                                    final List<Connection> connections) {
-        final int connectionCount = connections.size();
-        for (int ix = 0; ix < connectionCount; ix++) {
-            final Connection c = connections.get(ix);
-            if (c.equals(connectionToCallout))
+        final Iterator<Connection> it = connections.iterator();
```
kenhuuu, line 667: Minor nit: There aren't any tests for this PR, but this might be ok since appendConnections is only used for logging. But seeing as how we are now dependent on the CopyOnWriteArrayList's iterator behavior, we might want to add a small comment and maybe even explicitly declare connections as a CopyOnWriteArrayList<Connection> rather than just a List<Connection>.

kenhuuu:
VOTE +1 pending resolution of some nits.

upadhyay-prashant:
@kenhuuu  can you please take a look. I raised another pull request with the fix and your nits addressed.
https://github.com/apache/tinkerpop/pull/2618

PR #2616, opened 2024-05-22
[TINKERPOP-3080] AggregateGlobalStep accepts all pre-defined Operators

rdtr:
Support all Operators in AggregateGlobalStep and AggregateLocalStep.
https://issues.apache.org/jira/browse/TINKERPOP-3080

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Operator.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Operator.java
@@ -185,7 +185,7 @@ public enum Operator implements BinaryOperator<Object> {
             }
 
             if (a instanceof Map && b instanceof Map)
-                ((Map<?,?>) a).putAll((Map) b);
+                ((Map<?, ?>) a).putAll((Map) b);
             else if (a instanceof Collection && a instanceof Collection)
                 ((Collection<?>) a).addAll((Collection) b);
             else
@@ -204,5 +204,5 @@ public enum Operator implements BinaryOperator<Object> {
         public Object apply(final Object a, final Object b) {
             return (long) a + (long) b;
         }
-    }
+    };
 }
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AggregateGlobalStep.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AggregateGlobalStep.java
@@ -20,6 +20,7 @@ package org.apache.tinkerpop.gremlin.process.traversal.step.sideEffect;
 
 import org.apache.tinkerpop.gremlin.process.traversal.Operator;
 import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
+import org.apache.tinkerpop.gremlin.process.traversal.TraversalSideEffects;
 import org.apache.tinkerpop.gremlin.process.traversal.Traverser;
 import org.apache.tinkerpop.gremlin.process.traversal.step.ByModulating;
 import org.apache.tinkerpop.gremlin.process.traversal.step.LocalBarrier;
@@ -31,10 +32,8 @@ import org.apache.tinkerpop.gremlin.process.traversal.traverser.TraverserRequire
 import org.apache.tinkerpop.gremlin.process.traversal.traverser.util.TraverserSet;
 import org.apache.tinkerpop.gremlin.process.traversal.util.FastNoSuchElementException;
 import org.apache.tinkerpop.gremlin.process.traversal.util.TraversalUtil;
-import org.apache.tinkerpop.gremlin.structure.util.CloseableIterator;
 import org.apache.tinkerpop.gremlin.structure.util.StringFactory;
 import org.apache.tinkerpop.gremlin.util.function.BulkSetSupplier;
-
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
@@ -122,8 +121,17 @@ public final class AggregateGlobalStep<S> extends AbstractStep<S, S> implements 
 
     @Override
     public void processAllStarts() {
+        final TraversalSideEffects sideEffects = this.getTraversal().getSideEffects();
+
+        // addAll and assign operator collect solutions as a BulkSet. And addAll merges them with an existing
+        // sideEffect value which is expected to be Collection. assign replaces the existing value with the new BulkSet.
+        // Therefore, they need a different treatment how they gather inputs.
+        final boolean bulkReducingOperator = sideEffects.getReducer(sideEffectKey) == Operator.addAll ||
+                sideEffects.getReducer(sideEffectKey) == Operator.assign;
+
         if (this.starts.hasNext()) {
             final BulkSet<Object> bulkSet = new BulkSet<>();
+
             while (this.starts.hasNext()) {
                 final Traverser.Admin<S> traverser = this.starts.next();
                 TraversalUtil.produce(traverser, aggregateTraversal).ifProductive(p -> bulkSet.add(p, traverser.bulk()));
@@ -133,7 +141,12 @@ public final class AggregateGlobalStep<S> extends AbstractStep<S, S> implements 
                 // when barrier is reloaded, the traversers should be at the next step
                 this.barrier.add(traverser);
             }
-            this.getTraversal().getSideEffects().add(this.sideEffectKey, bulkSet);
+
+            if (bulkReducingOperator) {
+                sideEffects.add(this.sideEffectKey, bulkSet);
+            } else {
+                bulkSet.forEach(p -> sideEffects.add(sideEffectKey, p));
+            }
         }
     }
 
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AggregateLocalStep.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AggregateLocalStep.java
@@ -20,6 +20,7 @@ package org.apache.tinkerpop.gremlin.process.traversal.step.sideEffect;
 
 import org.apache.tinkerpop.gremlin.process.traversal.Operator;
 import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
+import org.apache.tinkerpop.gremlin.process.traversal.TraversalSideEffects;
 import org.apache.tinkerpop.gremlin.process.traversal.Traverser;
 import org.apache.tinkerpop.gremlin.process.traversal.step.ByModulating;
 import org.apache.tinkerpop.gremlin.process.traversal.step.SideEffectCapable;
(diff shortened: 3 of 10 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AggregateGlobalStep.java
```diff
@@ -122,8 +121,16 @@ protected Traverser.Admin<S> processNextStart() {
 
     @Override
     public void processAllStarts() {
+        final TraversalSideEffects sideEffects = this.getTraversal().getSideEffects();
+
+        // Pre-defined Operator such as addAll and assign will reduce over the whole input set, rather than
+        // applying a single input one by one.
+        final boolean isOperatorForBulkSet = sideEffects.getReducer(sideEffectKey) == Operator.addAll ||
```
saikiranboga, line 128: nit: suggest renaming this to `bulkReducingOperator` or something similar to callout observation towards the nature of reducing operators instead of limiting to BulkSet. addAll works with either maps or any collection and assign works with any type of input.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AggregateGlobalStep.java
```diff
@@ -122,8 +121,16 @@ protected Traverser.Admin<S> processNextStart() {
 
     @Override
     public void processAllStarts() {
+        final TraversalSideEffects sideEffects = this.getTraversal().getSideEffects();
+
+        // Pre-defined Operator such as addAll and assign will reduce over the whole input set, rather than
+        // applying a single input one by one.
```
saikiranboga, line 127: Suggest to also add to comment here on how both these operators work to be clear on why we focus on reduction over the whole input set. Basically, `addAll` can only work when either inputs to it are Maps or instances of Collection interface. And `assign` works for any type of input.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AggregateLocalStep.java
```diff
@@ -52,9 +53,20 @@ public AggregateLocalStep(final Traversal.Admin traversal, final String sideEffe
 
     @Override
     protected void sideEffect(final Traverser.Admin<S> traverser) {
+        final TraversalSideEffects sideEffects = this.getTraversal().getSideEffects();
+        // Pre-defined Operator such as addAll and assign will reduce over the whole input set, rather than
+        // applying a single input one by one.
+        final boolean isOperatorForBulkSet = sideEffects.getReducer(sideEffectKey) == Operator.addAll ||
+                sideEffects.getReducer(sideEffectKey) == Operator.assign;
+
```
saikiranboga, line 61: Same suggestions here as changes from AggregateGlobalStep.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/verification/ComputerVerificationStrategy.java
```diff
@@ -80,6 +78,13 @@ public void apply(final Traversal.Admin<?, ?> traversal) {
 
             if (UNSUPPORTED_STEPS.stream().filter(c -> c.isAssignableFrom(step.getClass())).findFirst().isPresent())
                 throw new VerificationException("The following step is currently not supported on GraphComputer: " + step, traversal);
+
+            if (step instanceof SideEffectCapable) {
+                final BinaryOperator<?> sideEffectOperator = traversal.getSideEffects().getReducer(((SideEffectCapable<?, ?>) step).getSideEffectKey());
+                if (sideEffectOperator instanceof Operator && (!((Operator) sideEffectOperator).isCommutative())) {
+                    throw new VerificationException("The following step has an SideEffect operator " + sideEffectOperator + " which is currently not supported on GraphComputer: " + step, traversal);
+                }
+            }
```
saikiranboga, line 87: Curious, what was the issue here?

saikiranboga (approved):
Looks good overall. Few nits.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AggregateGlobalStep.java
rdtr: thanks, renamed it

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/verification/ComputerVerificationStrategy.java
rdtr: Without this, in GraphComputer test I see that the unit tests I added failed. It seems GraphComputer works like map-reduce so when we have input like `1,2,3,4,5,6` for example, they are split into small chunk and operator is applied. 

So for example, let's say we have 3 hosts and data are split like `1, 6`, `2, 4`, and `3, 5`. With `ADD` it is fine because on each host the result would be `7`, `6`, and `8`. They are further summed up to 7 + 6 + 8 = 21 so the result is consistent.
However, for `MINUS`, a result in each host would be `-5`, `-2`, and `-2` then -5 - (-2) - (-2) = -1. This is not what we want because what we expect to happen is 1 - 2 - 3 - 4 - 5 - 6. 

I believe that we can still make it to work in distributed env with some tweak, but Stephen agreed to disable it for now.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AggregateGlobalStep.java
rdtr: Changed the comment, please take a look

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/sideEffect/Aggregate.feature
```diff
@@ -144,145 +144,490 @@ Feature: Step - aggregate()
...
+    g.withSideEffect("a", true, Operator.or).V().constant(false).aggregate(Scope.local, "a").cap("a")
+    """
+    When iterated to list
+    Then the result should be unordered
+      | result |
+      | true |
+
+  Scenario: g_withSideEffectXa_l1_addAllX_V_aggregateXaX_byXageX_capXaX
+    Given the modern graph
+    And using the parameter l1 defined as "l[d[1].i,d[2].i,d[3].i]"
```
xiazcy, line 571: `l1` is actually used to indicate lambda parameters, this and similar tests below, should be changed to `And using the parameter xx1`.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/sideEffect/Aggregate.feature
```diff
@@ -144,145 +144,490 @@ Feature: Step - aggregate()
...
+        addE("knows").from("marko").to("vadas").property("weight", 0.5d).
+        addE("knows").from("marko").to("josh").property("weight", 1.0d).
+        addE("created").from("marko").to("lop").property("weight", 0.4d).
+        addE("created").from("josh").to("ripple").property("weight", 1.0d).
+        addE("created").from("josh").to("lop").property("weight", 0.4d).
+        addE("created").from("peter").to("lop").property("weight", 0.2d)
       """
+    And the traversal of
+    """
+    g.withSideEffect("a", 1L, Operator.sumLong).V().aggregate(Scope.local, "a").by("age").cap("a")
```
xiazcy, line 360: For the test traversals, `1L` is lost in translation for GLVs, we'll need to specific this as a parameter with `d[1].l`

xiazcy:
Thanks @rdtr for the PR! There is a couple of fixes needed in the feature tests for GLVs. 
Another note, since you've updated the feature tests, please also push the correspondingly updated GLV tests files, ie the `gremlin.cs`, `gremlin.js`, `gremlin.py`, `gremlin.go` files (should be auto generated after doing a local `mvn clean install`).

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/sideEffect/Aggregate.feature
rdtr: Changed all `l1` to `xx1`. Please check

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/sideEffect/Aggregate.feature
```diff
@@ -144,145 +144,490 @@ Feature: Step - aggregate()
...
+    Then the result should be unordered
+      | result   |
+      | d[124].l |
+
+  Scenario: g_withSideEffectXa_1_sumLongX_V_aggregateXlocal_aX_byXageX_capXaX
+    Given the empty graph
+    And the graph initializer of
+    # This graph is the same as ModernGraph, but ages are stored as Long.
+      """
+      g.addV("person").property("name", "marko").property("age", 29L).as("marko").
```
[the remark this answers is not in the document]
rdtr: OK I  think it doesn't matter because when using `sumLong` operator, values are casted to `long` anyway. I will just use the regular Modern graph and see if the test passes.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/sideEffect/Aggregate.feature
rdtr: Same as above.

xiazcy:
VOTE +1. Thanks @rdtr! 
Since you weren't able to generate the feature test file for the GLVs, we'll have whoever's merging CTR them in.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Operator.java
```diff
@@ -113,6 +133,11 @@ public Object apply(final Object a, final Object b) {
         public Object apply(final Object a, final Object b) {
             return b;
         }
+
+        @Override
+        public boolean isCommutative() {
```
Cole-Greer, line 138: I believe assignment should be non-commutative.

Cole-Greer (commented):
Thanks @rdtr, I added 2 small questions, but other than that everything looks good to me.

VOTE +1 (pending resolution of below questions)

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Operator.java
rdtr: isCommutative is a confusing name, `assign` may be also the one but it is already used in GraphComputer's code internally. I just want to exclude `div` and `minus` at the moment, let's create a static list to disable execution for them.

Cole-Greer:
Thanks @rdtr, LGTM VOTE +1

kenhuuu:
VOTE +1

PR #2618, opened 2024-05-24

upadhyay-prashant:
CAUSE:
 post this.

 FIX:
 Iterate the connections list while creating the connectionPool Info.
 Since the list used is copyOnWrite, the iterator API creates a clone
 and uses that clone for referring the element. Thus providing thread
 safe interface.

 However the information provided by this iteration is a bit stale, but
 this doesn't matter.


upadhyay-prashant:
Previous pull request for this same fix. 
https://github.com/apache/tinkerpop/pull/2614

The previous pull request got closed because the branch got deleted.

kenhuuu:
This PR should target the 3.6-dev branch. That way it will go into the next 3.6.x/3.7.x release. If you target it to the master branch (like it is now), the change will only go into 4.0.x which likely isn't what you want.

upadhyay-prashant:
> This PR should target the 3.6-dev branch. That way it will go into the next 3.6.x/3.7.x release. If you target it to the master branch (like it is now), the change will only go into 4.0.x which likely isn't what you want.

Done.

PR #2652, opened 2024-06-13
Adding getter for fail step member

upadhyay-prashant:


kenhuuu:
Very small change to add getter so merging as CTR.

PR #2657, opened 2024-06-17
Python driver basic connectivity to HTTP server

xiazcy:
Updated python HTTP connection to use the new 4.0 request/response message formats. Reformatted existing tests to use HTTP server (tests passed with local HTTP server), but will need to further modify docker set up for GH actions. There remains a couple of client tests that need investigation. All old WS sockets are currently disabled, and will be removed at a later point along with related implementations. 

Next step is to enable chunked transfer, tests for new message formats will be added then.

### DIFF
--- a/gremlin-python/src/main/python/gremlin_python/driver/client.py
+++ b/gremlin-python/src/main/python/gremlin_python/driver/client.py
@@ -180,23 +180,19 @@ class Client:
             raise Exception("Client is closed")
 
         log.debug("message '%s'", str(message))
-        args = {'gremlin': message, 'aliases': {'g': self._traversal_source}}
-        processor = ''
-        op = 'eval'
+        # args = {'gremlin': message, 'fields': {'g': self._traversal_source}}
+        fields = {'g': self._traversal_source}
         if isinstance(message, traversal.Bytecode):
-            op = 'bytecode'
-            processor = 'traversal'
+            fields['gremlinType'] = 'bytecode'
+        elif isinstance(message, str):
+            fields['gremlinType'] = 'eval'
 
         if isinstance(message, str) and bindings:
-            args['bindings'] = bindings
-
-        if self._session_enabled:
-            args['session'] = str(self._session)
-            processor = 'session'
+            fields['bindings'] = bindings
 
         if isinstance(message, traversal.Bytecode) or isinstance(message, str):
-            log.debug("processor='%s', op='%s', args='%s'", str(processor), str(op), str(args))
-            message = request.RequestMessage(processor=processor, op=op, args=args)
+            log.debug("fields='%s', gremlin='%s'", str(fields), str(message))
+            message = request.RequestMessageV4(fields=fields, gremlin=message)
 
         conn = self._pool.get(True)
         if request_options:
--- a/gremlin-python/src/main/python/gremlin_python/driver/client.py
+++ b/gremlin-python/src/main/python/gremlin_python/driver/client.py
@@ -38,6 +38,7 @@ except ImportError:
 __author__ = 'David M. Brown (davebshow@gmail.com), Lyndon Bauto (lyndonb@bitquilltech.com)'
 
 
+# TODO: remove session, update connection pooling, etc.
 class Client:
 
     def __init__(self, url, traversal_source, protocol_factory=None,
@@ -58,7 +59,7 @@ class Client:
         if not self._use_http and "max_content_length" not in transport_kwargs:
             transport_kwargs["max_content_length"] = 10 * 1024 * 1024
         if message_serializer is None:
-            message_serializer = serializer.GraphBinarySerializersV1()
+            message_serializer = serializer.GraphBinarySerializersV4()
 
         self._message_serializer = message_serializer
         self._username = username
@@ -180,7 +181,6 @@ class Client:
             raise Exception("Client is closed")
 
         log.debug("message '%s'", str(message))
-        # args = {'gremlin': message, 'fields': {'g': self._traversal_source}}
         fields = {'g': self._traversal_source}
         if isinstance(message, traversal.Bytecode):
             fields['gremlinType'] = 'bytecode'
@@ -196,5 +196,6 @@ class Client:
 
         conn = self._pool.get(True)
         if request_options:
-            message.args.update(request_options)
+            message.fields.update(request_options)
+
         return conn.write(message)
--- a/gremlin-python/src/main/python/gremlin_python/driver/connection.py
+++ b/gremlin-python/src/main/python/gremlin_python/driver/connection.py
@@ -35,7 +35,7 @@ class Connection:
         self._executor = executor
         self._transport = None
         self._pool = pool
-        self._results = {}
+        self._result_set = None
         self._inited = False
         self._enable_user_agent_on_connect = enable_user_agent_on_connect
         if self._enable_user_agent_on_connect:
@@ -56,19 +56,15 @@ class Connection:
             self._transport.close()
 
     def write(self, request_message):
+        print('\n===conn===')
+        print(request_message)
         if not self._inited:
             self.connect()
-        if request_message.args.get("requestId"):
-            request_id = str(request_message.args.get("requestId"))
-            uuid.UUID(request_id)  # Checks for proper UUID or else server will return an error.
-        else:
-            request_id = str(uuid.uuid4())
-        result_set = resultset.ResultSet(queue.Queue(), request_id)
-        self._results[request_id] = result_set
+        self._result_set = resultset.ResultSet(queue.Queue())
         # Create write task
         future = Future()
         future_write = self._executor.submit(
-            self._protocol.write, request_id, request_message)
+            self._protocol.write, request_message)
 
         def cb(f):
             try:
@@ -79,8 +75,8 @@ class Connection:
             else:
                 # Start receive task
                 done = self._executor.submit(self._receive)
-                result_set.done = done
-                future.set_result(result_set)
+                self._result_set.done = done
+                future.set_result(self._result_set)
 
         future_write.add_done_callback(cb)
         return future
@@ -89,7 +85,7 @@ class Connection:
         try:
             while True:
                 data = self._transport.read()
-                status_code = self._protocol.data_received(data, self._results)
+                status_code = self._protocol.data_received(data, self._result_set)
                 if status_code != 206:
                     break
         finally:
--- a/gremlin-python/src/main/python/gremlin_python/driver/connection.py
+++ b/gremlin-python/src/main/python/gremlin_python/driver/connection.py
@@ -56,8 +56,6 @@ class Connection:
             self._transport.close()
 
     def write(self, request_message):
-        print('\n===conn===')
-        print(request_message)
         if not self._inited:
             self.connect()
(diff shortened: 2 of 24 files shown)


gremlin-server/conf/gremlin-server-modern.yaml
```diff
@@ -28,9 +29,9 @@ scriptEngines: {
...
-  - { className: org.apache.tinkerpop.gremlin.util.ser.GraphBinaryMessageSerializerV1 }                                                                                                           # application/vnd.graphbinary-v1.0
-  - { className: org.apache.tinkerpop.gremlin.util.ser.GraphBinaryMessageSerializerV1, config: { serializeResultToString: true }}                                                                 # application/vnd.graphbinary-v1.0-stringd
+  - { className: org.apache.tinkerpop.gremlin.util.ser.GraphSONMessageSerializerV4, config: { ioRegistries: [org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerIoRegistryV3] }}            # application/json
+  - { className: org.apache.tinkerpop.gremlin.util.ser.GraphBinaryMessageSerializerV4 }                                                                                                           # application/vnd.graphbinary-v1.0
+  - { className: org.apache.tinkerpop.gremlin.util.ser.GraphBinaryMessageSerializerV4, config: { serializeResultToString: true }}                                                                 # application/vnd.graphbinary-v1.0-stringd
```
vkagamlyk, line 34: need to keep only single GraphBinaryMessageSerializerV4

gremlin-python/src/main/python/gremlin_python/driver/serializer.py
```diff
@@ -292,3 +295,96 @@ def deserialize_message(self, message):
...
+        return message
+
+    def build_message(self, fields, gremlin):
+        message = {
+            'fields': fields,
+            'gremlin': gremlin
+        }
+        return self.finalize_message(message, 0x20, self.version)
+
+    def finalize_message(self, message, mime_len, mime_type):
```
vkagamlyk, line 341: nit: mime_len and mime_type are no longer part of request

gremlin-python/src/main/python/gremlin_python/structure/io/util.py
```diff
@@ -58,3 +61,38 @@ def of(cls, o):
...
+
+    @staticmethod
+    def end_of_stream():
+        return Marker(SingleByte(0))
+
+    def get_value(self):
+        return self._value
+
+    @classmethod
+    def of(cls, val):
```
vkagamlyk, line 94: GLV should be able to only read Marker, not send

gremlin-python/src/main/python/gremlin_python/driver/client.py
```diff
@@ -180,25 +181,21 @@ def submit_async(self, message, bindings=None, request_options=None):
...
 
         log.debug("message '%s'", str(message))
-        args = {'gremlin': message, 'aliases': {'g': self._traversal_source}}
-        processor = ''
-        op = 'eval'
+        fields = {'g': self._traversal_source}
         if isinstance(message, traversal.Bytecode):
-            op = 'bytecode'
-            processor = 'traversal'
+            fields['gremlinType'] = 'bytecode'
```
kenhuuu, line 186: Shouldn't need to set gremlinType as server should be able to determine this from the type of the serialized query.

gremlin-python/src/main/python/gremlin_python/driver/request.py
```diff
@@ -20,6 +20,5 @@
 
 __author__ = 'David M. Brown (davebshow@gmail.com)'
 
-
-RequestMessage = collections.namedtuple(
-    'RequestMessage', ['processor', 'op', 'args'])
+RequestMessageV4 = collections.namedtuple(
+    'RequestMessageV4', ['fields', 'gremlin'])
```
kenhuuu, line 24: FYI, you don't necessarily need to make the RequestMessageV4 like this. In Java, 'gremlin' was split away from 'fields' because 'gremlin' is the only field that will always exist, whereas the other items in 'fields' might change. This is to ensure that the request will always be compatible. You could just put 'gremlin' as just another key inside the map if it is easy for you do so in Python.

gremlin-python/src/main/python/gremlin_python/driver/connection.py
```diff
@@ -56,19 +56,15 @@ def close(self):
             self._transport.close()
 
     def write(self, request_message):
+        print('\n===conn===')
```
kenhuuu, line 59: Nit: clean up.

gremlin-python/src/main/python/gremlin_python/driver/protocol.py
```diff
@@ -232,28 +233,27 @@ def data_received(self, response, results_dict):
...
+                                          'message': message['status']['message'],
+                                          'attributes': {}})
         else:
-            # This message is going to be huge and kind of hard to read, but in the event of an error,
-            # it can provide invaluable info, so space it out appropriately.
-            log.error("\r\nReceived error message '%s'\r\n\r\nWith results dictionary '%s'",
-                      str(response['content']), str(results_dict))
+            log.error("\r\nReceived error message '%s'\r\n\r\nWith result set '%s'",
+                      str(response['content']), str(result_set))
             body = json.loads(response['content'])
```
kenhuuu, line 258: You could get an error that is serialized in GraphBinaryV4 back here. Should check the Content-Type instead of relying on it being JSON.

gremlin-python/src/main/python/gremlin_python/driver/resultset.py
```diff
@@ -23,9 +23,8 @@
 
 class ResultSet:
 
-    def __init__(self, stream, request_id):
+    def __init__(self, stream):
         self._stream = stream
-        self._request_id = request_id
         self._done = None
         self._aggregate_to = None
         self._status_attributes = {}
```
kenhuuu, line 30: Nit: You should be able to safely ignore status attributes so this part could be removed.

gremlin-python/src/main/python/gremlin_python/driver/serializer.py
```diff
@@ -72,10 +75,10 @@ def close(self, args):
     def bytecode(self, args):
         gremlin = args['gremlin']
         args['gremlin'] = self._writer.to_dict(gremlin)
-        aliases = args.get('aliases', '')
-        if not aliases:
-            aliases = {'g': 'g'}
-        args['aliases'] = aliases
+        bindings = args.get('bindings', '')
+        if not bindings:
+            bindings = {'g': 'g'}
```
kenhuuu, line 80: There's a "g" in the request now that is separate from bindings. The new "g" takes precedence over anything that was set in bindings (overwrites any 'g' key in bindings).

gremlin-python/src/main/python/gremlin_python/driver/client.py
xiazcy: It doesn't look like server reads it at this point, so I'll keep these in for now. It will need to be updated once we remove bytecode from python anyways.

gremlin-python/src/main/python/gremlin_python/driver/serializer.py
xiazcy: Ah, this was a left over change from me playing with the code at the beginning. This is part of the session op processor code that will be removed.

gremlin-python/src/main/python/gremlin_python/driver/request.py
xiazcy: It's actually ok, but I'll keep this in mind for when we need to do refactoring.

gremlin-python/src/main/python/gremlin_python/driver/protocol.py
xiazcy: Does the server serialize errors to GraphBinary? Even if I only have GraphBinaryV4 serializer enabled in server yaml, error responses are still sent as `application/json` (ok responses are in `application/vnd.graphbinary-v4.0`).

kenhuuu: It actually depends on the error and when the error occurs. Some are always serialized to JSON and other times it will serialize depending on the "Accept" header that was sent with the request. It really just depends on where the failure occurred in the server. For instance, if you sent an accept header that contains a MIME type the server can't handle, then obviously the server would have to send back JSON. But if you sent an error that occurred after processing has started, then it can return back serialized. In either case, the server should send back the correct content type for you to use.

xiazcy: Ok, I dug a bit into this, so I think this might come from a combination of aiohttp defines "ok":
`Boolean representation of HTTP status code ([bool](https://docs.python.org/3/library/functions.html#bool)). True if status is less than 400; otherwise, False.` and how chunking isn't being handled right now.

Basically for now, I think due to not implementing chunking and allowing the library to handle it natively, we are only passing the first status we get from the server, which will be a 200, and that will make the final failing request fall into the `if response['ok']` case, and errors will be properly serialized. 

And the not `ok` cases look to be returned with `'Content-Type': 'application/vnd.graphbinary-v4.0'`, which the `else` case can handle.

Note that my request headers are:
```
CONTENT-TYPE: application/vnd.graphbinary-v4.0
ACCEPT: application/vnd.graphbinary-v4.0
```

For example (printing response headers and error output):
1. Tweaked serializer to send wrong number of bytes to server:
```
<CIMultiDictProxy('Content-Type': 'application/json', 'Content-Length': '81', 'Gremlin-RequestId': 'cf6793cd-316e-4f88-bdde-6d1d82b9a094')>
2024-06-21 12:38:15,918 [   ERROR] [    protocol.py:254 - data_received()] - 
Received error message 'b'{"message":"Unable to deserialize request using: GraphBinaryMessageSerializerV4"}''
```
2. Sent a math() step dividing by 0:
```
 <CIMultiDictProxy('Transfer-Encoding': 'chunked', 'Content-Type': 'application/vnd.graphbinary-v4.0', 'Gremlin-RequestId': '42fad392-12d0-4978-a085-7338897751d3', 'code': '500')>
Received error message '{'status': {'code': 500, 'message': 'Division by zero!', 'exception': 'ServerErrorException'}, 'result': {'meta': {}, 'data': []}}'
With result set '<gremlin_python.driver.resultset.ResultSet object at 0x10a42b6d0>'
```
3. Sent long request that caused server timeouts during processing:
```
<CIMultiDictProxy('Transfer-Encoding': 'chunked', 'Content-Type': 'application/vnd.graphbinary-v4.0', 'Gremlin-RequestId': '84d848f7-3747-4da6-83d1-93f90bc106f3', 'code': '500')>
2024-06-21 12:32:22,415 [   ERROR] [    protocol.py:248 - data_received()] - 
Received error message '{'status': {'code': 500, 'message': 'A timeout occurred during traversal evaluation of [RequestMessageV4{, fields={gremlinType=bytecode, g=g, bindings={}}, gremlin=[[], [V(), limit(10000)]]}]
```

So I think what I'll do is to add a `TODO` for checking on this when doing proper chunking to make sure errors are still properly serialized.

kenhuuu:
VOTE +1

vkagamlyk:
VOTE+1

xiazcy:
VOTE +1. Thanks everyone for the review!

PR #2658, opened 2024-06-19
Update log lever from error to warn when server closes connection

gautam-nutalapati:
- AWS Neptune closes connections after 10 days when IAM is enabled.
  - Refer https://docs.aws.amazon.com/neptune/latest/userguide/limits.html#limits-websockets 
  `When IAM authentication is enabled, a WebSocket connection is always disconnected a few minutes more than 10 days after it was established, if it hasn't already been closed by then.`
  - This creates lot of error logs when using connection pools.
- Unless server closing connection is an error, logging the respective log as WARN may be ideal to prevent noise in logs.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java
@@ -129,7 +129,7 @@ final class Connection {
                 // close did not come from the client side. it means the server closed the channel for some reason.
                 // it's important to distinguish that difference in debugging
                 if (thisConnection.closeFuture.get() == null) {
-                    logger.error(String.format(
+                    logger.warn(String.format(
                             "Server closed the Connection on channel %s - scheduling removal from %s",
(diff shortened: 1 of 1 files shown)


xiazcy:
Thank you for the PR @gautam-nutalapati! A couple of notes on this:
- This connection logging level might be something that can affect other user's debugging workflows, so a devlist discussion post might be needed to let the community know of the proposed change
- The master branch will be for the next major release, you would need to rebase and retarget your change to 3.6-dev if you'd like to include this change in the next maintenance releases for 3.6 and 3.7. 
- Please include a CHANGELOG entry

kenhuuu:
I feel as though the solution you are suggesting here may affect others if they are depending on that information to be at the ERROR level. It could be considered breaking for some users. For this reason, I feel like a more complete change would be to update the behavior of how the driver responds to a close frame. I've suggested a way of doing this in [TINKERPOP-3094](https://issues.apache.org/jira/browse/TINKERPOP-3094).

Would you be OK with closing this PR and waiting or working on a fix like the one suggested in TINKERPOP-3094?

xiazcy:
Closing as per comment above. Please feel free to open if you have additional suggestions or concerns.

PR #2661, opened 2024-06-21
Bytecode removal from server

vkagamlyk:
The server will no longer process Bytecode requests.

Old translator replaced by new ones introduced in TINKERPOP-3028.

Feature tests will run for both gremlin-lang and gremlin-groovy script engines.


kenhuuu:
I took a look at driver and server changes (mostly to handling of RequestMessageV4 and HttpGremlinEndpointHandler) and they LGTM.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/GroovyTranslateVisitor.java
```diff
@@ -113,4 +118,63 @@ public Void visitFloatLiteral(final GremlinParser.FloatLiteralContext ctx) {
...
+    public Void visitTraversalSourceSpawnMethod_inject(final GremlinParser.TraversalSourceSpawnMethod_injectContext ctx) {
+        return handleInject(ctx);
+    }
+
+    @Override
+    public Void visitTraversalMethod_inject(final GremlinParser.TraversalMethod_injectContext ctx) {
+        return handleInject(ctx);
+    }
+
+    private Void handleInject(final ParserRuleContext ctx) {
```
spmallette, line 153: please expand the comment a bit to include "why" it can't work correctly - `inject()` ends up being ambiguous with groovy's jdk extension of `inject(Object initialValue, Closure closure)`.  not sure why `inject(Closure)` doesn't seem to conflict with inject(null)` - a groovy mystery? i wonder if that is just coincidental? should we just account for it to avoid future issues?

i think it would be good to add a specific tests for these cases in the `GremlinTranslatorTest`.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/remote/EmbeddedRemoteConnection.java
```diff
@@ -59,7 +62,11 @@ public <E> CompletableFuture<RemoteTraversal<?, E>> submitAsync(final GremlinLan
         // the new submit() in 3.3.x when the deprecation is removed
         final CompletableFuture<RemoteTraversal<?, E>> promise = new CompletableFuture<>();
         try {
-            promise.complete(new EmbeddedRemoteTraversal(JavaTranslator.of(g).translate(gremlinLang)));
+            final GremlinLangScriptEngine scriptEngine = new GremlinLangScriptEngine();
```
spmallette, line 65: do we need a new instance each time here? i assume this could be created per instance?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/remote/EmbeddedRemoteConnection.java
```diff
@@ -59,7 +62,11 @@ public <E> CompletableFuture<RemoteTraversal<?, E>> submitAsync(final GremlinLan
...
+            final Bindings b = new SimpleBindings();
```
spmallette, line 66: better to prefer `scriptEngine.createBindings()` which does the same thing right now but if we ever made lang specific bindings it would spare us some change.

spmallette:
i think there is an extensive set of changelog and upgrade docs to write up here with major removals to keep track of and new things introduced.  lots and lots of change to account for that is going to a branch with an equal amount of change. probably a good time to take stock of what was done here and document it as part of the PR.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GremlinLang.java
```diff
@@ -59,9 +56,6 @@ public class GremlinLang implements Cloneable, Serializable {
```
spmallette, line 56: please add some class level javadoc here as `GremlnLang`

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalStrategies.java
```diff
@@ -247,10 +248,13 @@ public static final class GlobalCache {
             put(SubgraphStrategy.class.getSimpleName(), SubgraphStrategy.class);
 
             // finalization
+            put(MatchAlgorithmStrategy.class.getSimpleName(), MatchAlgorithmStrategy.class);
```
spmallette, line 251: the strategies you added here were not originally part of the grammar because this list only consisted of strategies that we had explicitly included there. i guess these were added to allow tests to pass. it might not be wrong but if you add them here then they should be tested in the feature tests (and i dont think they are). if you do that then each GLV will likely need changes to support them. and if you do these, i don't see why we wouldn't do all of them, which seems like an inevitability. that inevitability leads to (1) is there a better way for GLVs to support new strategies so that there isn't a huge implementation cost of adding proxy strategy classes and (2) a wider question of what happens when a strategy is deprecated or removed...it dynamically affects the grammar.....

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -2291,10 +2291,7 @@ public default GraphTraversal<S, E> has(final T accessor, final Traversal<?, ?>
      * @since 3.0.0-incubating
      */
     public default GraphTraversal<S, E> has(final String propertyKey, final Traversal<?, ?> propertyTraversal) {
-        // the translation here of null to has(String, Object) is likely what was intended. a null Traversal doesn't
-        // really make much sense. this should resolve issues with JavaTranslator grabbing this method when bytecode
-        // uses null as the second argument. we've taken this tactic for other overloads of has() as well, so just
-        // continuing with that pattern.
+        // todo: check if this magic is needed for gremlin-lang
```
spmallette, line 2294: clean up todo....

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/DataType.java
```diff
@@ -44,8 +44,8 @@ public enum DataType {
     VERTEX(0X11),
     VERTEXPROPERTY(0X12),
     BARRIER(0X13),
-    BINDING(0X14),
```
spmallette, line 47: this change begs an update to the IO docs and what will be the start of the upgrade docs. i don't know if this doc work is being tracked separately but i think subtle things will be missed if it all waits until the master-http branch merge.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoVersion.java
```diff
@@ -321,12 +320,10 @@ public static List<TypeRegistration<?>> initV3Registrations() {
             add(GryoTypeReg.of(ProjectedTraverser.class, 168));
             add(GryoTypeReg.of(DefaultRemoteTraverser.class, 123, new GryoSerializersV3.DefaultRemoteTraverserSerializer()));
 
-            add(GryoTypeReg.of(GremlinLang.class, 122, new GryoSerializersV3.BytecodeSerializer()));
```
spmallette, line 324: rather than remove these lines wholly, please leave a comment in each space like:
```
// 122 was Bytecode until it was removed for 4.x
// 126 was Bytecode.Binding until it was removed 4.x
```

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
```diff
@@ -105,31 +102,6 @@ public Client alias(final String graphOrTraversalSource) {
         return new AliasClusteredClient(this, graphOrTraversalSource);
     }
 
-    /**
```
spmallette, line 108: another key change to document in changelog/upgrade docs.

gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GremlinGroovyScriptTestEngine.java
```diff
@@ -0,0 +1,50 @@
...
+import javax.script.ScriptContext;
+import javax.script.ScriptException;
+
+/**
+ * Designed to be used in tests.
+ * Converts incoming script from gremlin-lang to gremlin-groovy compatible form and executes with
+ * {@code GremlinGroovyScriptEngine}.
+ */
+
+public class GremlinGroovyScriptTestEngine extends GremlinGroovyScriptEngine {
```
spmallette, line 34: This class is for tests but only by convention because it is in main `gremlin-groovy` code. Can it be actually moved to test code somehow so that by design it is not usable any other way?

gremlin-groovy/src/test/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GremlinGroovyScriptEngineTest.java
```diff
@@ -474,122 +467,6 @@ public void customizerShouldOverrideCoreImports() throws Exception {
...
-        final PythonTranslator pyTranslator = PythonTranslator.of("g");
-        final String gremlinAsPy = pyTranslator.translate(bytecode).getScript();
-        assertEquals("g.inject(float('-inf'),float('nan'),xx1).is_(P.eq(float('inf')).or_(P.eq(float('nan'))).or_(P.eq(float('inf'))))", gremlinAsPy);
-
-        final DotNetTranslator dnTranslator = DotNetTranslator.of("g");
-        final String gremlinAsDn = dnTranslator.translate(bytecode).getScript();
-        assertEquals("g.Inject(Double.NegativeInfinity,Double.NaN,xx1).Is(P.Eq(Double.PositiveInfinity).Or(P.Eq(Double.NaN)).Or(P.Eq(Double.PositiveInfinity)))", gremlinAsDn);
-    }
-
-    @Test
```
spmallette, line 551: it occurs to me that we don't have this kind of coverage anywhere for the new translators? are we losing some coverage by removing this test? should we add some coverage somewhere else to make up fo rit?

gremlin-groovy/src/test/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GroovyTranslatorProvider.java
```diff
@@ -222,7 +212,6 @@
 public class GroovyTranslatorProvider extends TinkerGraphProvider {
     @Override
     public GraphTraversalSource traversal(final Graph graph) {
-        final GraphTraversalSource g = graph.traversal();
-        return g.withStrategies(new TranslationStrategy(g, GroovyTranslator.of("g"), true));
+        return graph.traversal();
```
spmallette, line 215: given this change, this test and `ParameterizedGroovyTranslatorProvider` really aren't doing anything anymore? where are we testing the new Groovy translator with this rigor? if we are, then may be don't need this anymore?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Context.java
```diff
@@ -187,15 +164,6 @@ public GremlinExecutor getGremlinExecutor() {
      */
     public void setStartedResponse() { startedResponse.set(true); }
 
-    private RequestContentType determineRequestContents() {
```
spmallette, line 190: so nice......

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/remote/AbstractRemoteGraphProvider.java
```diff
@@ -221,16 +221,22 @@ public abstract class AbstractRemoteGraphProvider extends AbstractGraphProvider
...
     private final boolean useComputer;
+    private final String queryLanguage;
 
 
     public AbstractRemoteGraphProvider(final Cluster cluster) {
         this(cluster, false);
     }
 
     public AbstractRemoteGraphProvider(final Cluster cluster, final boolean useComputer) {
+        this(cluster, useComputer, "gremlin-groovy");
```
spmallette, line 232: i guess the `AbstractRemoteGraphProvider` tests using "gremlin-groovy" will test the groovy path for remote execution, i.e. taking `GremlinLang` produced Gremlin, passing it through the translator to groovy so that it can be executed in the `GremlinGroovyScriptEngine`. i think it's worth clarifying that in the class javadocs. worth tracking a better mechanism for testing `GremlinGroovyScriptEngine` that doesn't rely on the java tests and instead uses the feature tests.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/remote/GraphBinaryGroovyRemoteGraphProvider.java
```diff
@@ -16,23 +16,20 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.tinkerpop.gremlin.jsr223;
```
spmallette, line 19: interesting that this class is being tracked by git as a rename........ 🤷

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/remote/GraphBinaryRemoteGraphComputerProvider.java
```diff
@@ -143,10 +143,14 @@
         test = "org.apache.tinkerpop.gremlin.process.traversal.step.map.PathTest",
         method = "g_injectX1_null_nullX_path_dedup",
         reason = "The inject() step is not supported by GraphComputer")
+@Graph.OptOut(
```
spmallette, line 146: do all the tests need to opted out? only the `GreedyMatchTraversals` use that strategy syntax, no?

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/remote/GraphBinaryGroovyRemoteGraphProvider.java
```diff
@@ -16,23 +16,20 @@
...
  */
-public interface TranslatorCustomizer extends Customizer {
-
-    /**
-     * Construct a {@link Translator.ScriptTranslator.TypeTranslator} that will be used by a
-     * {@link Translator.ScriptTranslator} instance within the {@link GremlinScriptEngine} to translate
-     * {@link GremlinLang} to a script.
-     */
-    public Translator.ScriptTranslator.TypeTranslator createTypeTranslator();
+@Graph.OptOut(
```
spmallette, line 27: do all the tests need to opted out? only the `GreedyMatchTraversals` use that strategy syntax, no?

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/remote/RemoteWorld.java
```diff
@@ -159,19 +159,35 @@ public GraphTraversalSource getGraphTraversalSource(final LoadGraphWith.GraphDat
...
-    public static class GraphBinaryRemoteWorld extends RemoteWorld {
-        public GraphBinaryRemoteWorld() { super(createTestCluster(SerializersV4.GRAPHBINARY_V4)); }
+    public static class GraphBinaryLangRemoteWorld extends RemoteWorld {
+        public GraphBinaryLangRemoteWorld() { super(createTestCluster(SerializersV4.GRAPHBINARY_V4)); }
 
         @Override
         public GraphTraversalSource getGraphTraversalSource(final LoadGraphWith.GraphData graphData) {
             final GraphTraversalSource g = super.getGraphTraversalSource(graphData);
-            return g.with("language", "gremlin-lang");
+             return g.with("language", "gremlin-lang");
```
spmallette, line 168: `with("language", "gremlin-lang")` is being used to rig up this test. i assume this isn't meant to be something users can do though? like, if they did `.with("language", "gremlin-groovy")` they would have a problem because you may or may not get groovy compliant output. maybe this is one of the request options we shouldn't allow with the language variants through `with()`? i dont' know what that means for testing through.......

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerAuthzIntegrateTest.java
```diff
@@ -158,6 +158,7 @@ public void shouldAuthorizeGremlinLangRequestWithLambda() {
         }
     }
 
+    @Ignore("not implemented for string requests")
```
spmallette, line 161: you could probably wholly remove the "lambda" test since GLVs can't send those anymore. as mentioned elsewhere in the PR, it would be good to get a JIRA up for tracking these things and have it referenced here in the ignores so it doesn't get lost.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/HttpDriverIntegrateTest.java
```diff
@@ -71,6 +69,21 @@ public Settings overrideSettings(final Settings settings) {
         return settings;
     }
 
+    @Test
```
spmallette, line 72: is this a real test? can it be removed? if not, better name for it?

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/VertexTest.java
```diff
@@ -276,6 +277,7 @@ public void g_EXe7_e11X() {
     }
 
     @Test
+    @Ignore("g.E(Arrays.asList(e7, e11)) is not valid for gremlin-lang")
```
spmallette, line 280: ignoring at this level will ignore for embedded tests. the change should be to ignore for gremlin-lang only, right?

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/InjectTest.java
```diff
@@ -114,6 +115,7 @@ public void g_injectXnull_1_3_nullX() {
     }
 
     @Test
+    @Ignore
```
spmallette, line 118: the issue with these tests was with groovy execution right? i thought that was solved earlier with the changes to the translator? can these be enabled again?

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/structure/io/script/script-input.groovy
```diff
@@ -1,47 +0,0 @@
-import org.apache.tinkerpop.gremlin.structure.VertexProperty
```
spmallette, line 1: any reason this was removed? still relevant to hadoop testing, no?

gremlin-util/src/test/java/org/apache/tinkerpop/gremlin/util/ser/binary/TypeSerializerFailureTests.java
```diff
@@ -75,11 +73,9 @@ public static Collection input() {
 
         // Provide instances that are malformed for serialization to fail
         return Arrays.asList(
-                b,
                 vertex,
-                Collections.singletonMap("one", b),
                 bulkSet,
-                bytecode,
+                gremlinLang,
```
spmallette, line 78: what does graphbinary do with a `GremlinLang` object?

hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/structure/HadoopGraph.java
```diff
@@ -60,26 +57,26 @@
...
-        test = "org.apache.tinkerpop.gremlin.process.traversal.step.map.MatchTest$Traversals",
-        method = "g_V_matchXa_0sungBy_b__a_0writtenBy_c__b_writtenBy_d__c_sungBy_d__d_hasXname_GarciaXX",
-        reason = "Hadoop-Gremlin is OLAP-oriented and for OLTP operations, linear-scan joins are required. This particular tests takes many minutes to execute.",
-        computers = {"ALL"})
-@Graph.OptOut(
-        test = "org.apache.tinkerpop.gremlin.process.traversal.step.map.MatchTest$Traversals",
-        method = "g_V_matchXa_0sungBy_b__a_0writtenBy_c__b_writtenBy_dX_whereXc_sungBy_dX_whereXd_hasXname_GarciaXX",
-        reason = "Hadoop-Gremlin is OLAP-oriented and for OLTP operations, linear-scan joins are required. This particular tests takes many minutes to execute.",
-        computers = {"ALL"})
+//@Graph.OptOut(
```
spmallette, line 60: what was the reason for commenting these out?

tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/io/graphson/AbstractTinkerGraphGraphSONTranslatorProvider.java
```diff
@@ -238,8 +228,7 @@ public abstract class AbstractTinkerGraphGraphSONTranslatorProvider extends Tink
 
     @Override
     public GraphTraversalSource traversal(final Graph graph) {
-        final GraphTraversalSource g = graph.traversal();
-        return g.withStrategies(new TranslationStrategy(g, new GraphSONTranslator<>(JavaTranslator.of(g), version), true));
+        return graph.traversal();
```
spmallette, line 231: similar question here as asked earlier - what are we really testing here after this change?

tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/io/gryo/TinkerGraphGryoTranslatorProvider.java
```diff
@@ -224,7 +222,6 @@ public class TinkerGraphGryoTranslatorProvider extends TinkerGraphProvider {
 
     @Override
     public GraphTraversalSource traversal(final Graph graph) {
-        final GraphTraversalSource g = graph.traversal();
-        return g.withStrategies(new TranslationStrategy(g, new GryoTranslator<>(JavaTranslator.of(g)), true));
+        return graph.traversal();
```
spmallette, line 225: similar question here as asked earlier - what are we really testing here after this change?

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/remote/GraphBinaryGroovyRemoteGraphProvider.java
vkagamlyk: is was 'GraphBinaryRemoteGraphProvider' before I made the tests to run for both `gremlin-lang` and `groovy-test`

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/HttpDriverIntegrateTest.java
vkagamlyk: removed

gremlin-util/src/test/java/org/apache/tinkerpop/gremlin/util/ser/binary/TypeSerializerFailureTests.java
vkagamlyk: removed

gremlin-groovy/src/test/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GroovyTranslatorProvider.java
vkagamlyk: removed. 
I think we have enough coverage for new Groovy translator with `groovy-test` engine.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/remote/GraphBinaryRemoteGraphComputerProvider.java
vkagamlyk: right, OptOut reduced to `MatchTest$GreedyMatchTraversals`

gremlin-groovy/src/test/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GremlinGroovyScriptEngineTest.java
spmallette: i meant particularly for `call()`

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/authz/Authorizer.java
```diff
@@ -38,18 +36,10 @@ public interface Authorizer {
...
-     * the request.
-     *
-     * @param user {@link AuthenticatedUser} that needs authorization.
-     * @param bytecode The gremlin {@link GremlinLang} request to authorize the user for.
-     * @param aliases A {@link Map} with a single key/value pair that maps the name of the {@link TraversalSource} in the
-     *                {@link GremlinLang} request to name of one configured in Gremlin Server.
-     * @return The original or modified {@link GremlinLang} to be used for further processing.
-     */
-    public GremlinLang authorize(final AuthenticatedUser user, final GremlinLang bytecode, final Map<String, String> aliases) throws AuthorizationException;
+    // todo: implement auth for gremlin-lang
```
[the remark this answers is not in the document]
vkagamlyk: [Discussion thread in dev list](https://lists.apache.org/thread/kb834lvttpy7ntos5q4skypmr447j1qm)

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/VertexTest.java
vkagamlyk: This is not valid for Grammar, probably need to clean up this a bit.

in Grammar we allow `g.V(vertex)` and `g.E(vertex)`, but not `g.V(edge)` and `g.E(edge)`.
gremlin-lang and gremlin-groovy can handle `g.V(vertex)` and `g.E(edge)`.

For gremlin-lang `g.E(edge)` translated as `g.E(_1)` because no Edge defined in Grammar. But this approach not valid for `g.E([edge1, edge2])`, gremlin-lang can't handle Parameters inside array.

gremlin-groovy/src/main/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GremlinGroovyScriptTestEngine.java
vkagamlyk: moved to gremlin-server tests

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/VertexTest.java
spmallette: > in Grammar we allow g.V(vertex) and g.E(vertex), but not g.V(edge) and g.E(edge).

what is `g.E(vertex)` and `g.V(edge)`? 

> For gremlin-lang g.E(edge) translated as g.E(_1) because no Edge defined in Grammar.

i think think that perhaps the problem here is that the grammar was adapted to allow the tests to work. from a grammar perspective, i'm not sure that anyone needs to do `g.V(Vertex(1,'person'))` (or the edge equivalent). You would simply do `g.V(1)`. i think `g.V(Vertex)` though is useful at the API level because you can do:

```groovy
v = g.V(1).next()
g.addE('self').from(v).to(v)
```

but again, that's from a programming language perspective not the query perspective. might be smart to take something away here. 

> But this approach not valid for g.E([edge1, edge2]), gremlin-lang can't handle Parameters inside array.

i think i know why we allowed that `List` syntax:

```groovy
l = g.V(1,2,3).toList()
g.V(l)
```

i wonder if that's a good idea though. maybe we're doing too much with this sort of overload.

vkagamlyk: > > in Grammar we allow g.V(vertex) and g.E(vertex), but not g.V(edge) and g.E(edge).
> 
> what is `g.E(vertex)` and `g.V(edge)`?
> 
`vertex` is some Vertex object, so `g.E(Vertex(1,'person'))` is valid for Grammar

> 
> i wonder if that's a good idea though. maybe we're doing too much with this sort of overload.

I assume we can limit possible arguments for `V()` step aka traversalSourceSpawnMethod_V, because now there are too many options that make no sense, for example `g.V(Merge.outV)` and `g.V(datetime(...))` is kind of valid for Grammar.
Same for `E()` step

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/remote/AbstractRemoteGraphProvider.java
vkagamlyk: I added javadoc to constructor

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalStrategies.java
vkagamlyk: I don't have strong opinion about adding strategies to Grammar.
May be we can add something similar to `call()` step, but for strategies?

gremlin-groovy/src/test/java/org/apache/tinkerpop/gremlin/groovy/jsr223/GremlinGroovyScriptEngineTest.java
vkagamlyk: I added test case for translator and [ticket](https://issues.apache.org/jira/browse/TINKERPOP-3091) to track feature tests

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/structure/io/script/script-input.groovy
vkagamlyk: reverted

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalStrategies.java
spmallette: after some thought, i feel more inclined to say that i think that they should all be registered, but i guess you don't need to do that for that PR. for now, maybe just symmetrically add the tests for the strategies you added to the feature tests so that this isn't missed and then create a JIRA for how GLVs might have a more dynamic way to instantiate strategy classes. i don't think it's by way of a grammar change, like `call()` but perhaps something similar to: 

https://github.com/apache/tinkerpop/blob/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/TraversalStrategyProxy.java

where there is just one construct in a GLV to construct any strategy. then we just deprecate any existing concrete classes that construct strategies. this also solves the problem for how users can work with their own strategies they might create and register by way of Gremlin Server and for provider strategies that get registered per graph.

Cole-Greer:
Thanks Valentyn, this is a huge piece of work and a massive step towards simplifying TinkerPop.
LGTM VOTE +1

xiazcy:
VOTE +1. Thanks for this great body of work!

vkagamlyk:
VOTE +1

PR #2664, opened 2024-06-25
Add PR template.

kenhuuu:
Adding a template based on the dev list DISCUSS thread found here https://lists.apache.org/thread/63r2lzv69yf0odvwbz111yckvx7q10of

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,14 @@
+<!--
+Thanks for contributing! Reminders:
++ TARGET the earliest branch where you want the change
+    3.6-dev -> 3.6.8 (bugs only)
+    3.7-dev -> 3.7.3 (non-breaking)
+    master  -> 4.0.0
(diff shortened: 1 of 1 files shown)


Cole-Greer:
Thanks Ken, LGTM
VOTE +1

kenhuuu:
> I don't mean to be overly focused on number of words, but if this takes too long to read, it might not have the desired effect.

I agree with this and have tried to explain the target branch in as few words as possible, but couldn't quite shorten it down anymore. I feel the need to explain our branch management strategy as it seems to be slightly different than other projects and a common source of confusion for new contributors.

> (2) more formatting to draw the eye to stuff that matters.

Also agree, but I'm limited on what I can do here based on this being a hidden comment that won't render formatting.

vkagamlyk:
VOTE +1

kenhuuu:
VOTE +1

PR #2665, opened 2024-06-26
TINKERPOP-3089 Not throwing NoSuchElementException when encountering an empty iterator input

rdtr:
https://issues.apache.org/jira/browse/TINKERPOP-3089

```
gremlin> g.V().local(union(values("age"), outE().values("weight")).fold())
==>[29,0.4,0.5,1.0]
==>[27]
==>[]
==>[32,1.0,0.4]
==>[]
==>[35,0.2]
```

```
gremlin> g.V().local(union(values("age"), outE().values("weight")).fold()).max(local)
==>29.0
==>27
```
This is wrong. TinkerPop needs to filter `[]` and handle non empty List instead of throwing `NoSuchElementException` immediately.


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MaxLocalStep.java
```diff
@@ -34,23 +33,24 @@
  * @author Marko A. Rodriguez (http://markorodriguez.com)
  * @author Daniel Kuppitz (http://gremlin.guru)
  */
-public final class MaxLocalStep<E extends Comparable, S extends Iterable<E>> extends ScalarMapStep<S, E> {
+public final class MaxLocalStep<E extends Comparable, S extends Iterable<E>> extends FlatMapStep<S, E> {
```
saikiranboga, line 36: This would extend just abstract MapStep class too right? We wont ever emit more than single solution out which is usually when FlatMapStep is used for? (Same for other local step in the PR).

saikiranboga (approved):
Mostly looks good. One comment on extending FlatMapStep.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MaxLocalStep.java
rdtr: Introduced `OptionalScalarMap` to support no output case

saikiranboga (approved):

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MaxLocalStep.java
```diff
@@ -34,23 +34,23 @@
...
         if (iterator.hasNext()) {
-            Comparable result = iterator.next();
+            result = iterator.next();
             while (iterator.hasNext()) {
                 result = max(iterator.next(), result);
             }
-            return (E) result;
         }
-        throw FastNoSuchElementException.instance();
+        return Optional.of((E) result);
```
saikiranboga, line 53: Seems this is throwing NPE. You need to use `ofNullable` instead.

spmallette:
VOTE +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MaxLocalStep.java
rdtr: Unfortunately in TinkerPop, there is some case that returns `null` so I could not always treat like this and skip when the result is `null`. I reverted the change and extended just `MapStep`

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MaxLocalStep.java
```diff
@@ -34,23 +34,26 @@
...
 
     @Override
-    protected E map(final Traverser.Admin<S> traverser) {
-        final Iterator<E> iterator = IteratorUtils.asIterator(traverser.get());
-        if (iterator.hasNext()) {
-            Comparable result = iterator.next();
-            while (iterator.hasNext()) {
-                result = max(iterator.next(), result);
+    protected Traverser.Admin<E> processNextStart() throws NoSuchElementException {
+        while (true) {
```
vkagamlyk, line 45: I don’t quite understand how this loop will end with an empty traverser iterator, when `iterator.hasNext()` is false. 
Good to have tests for corner cases like `g.inject(null,[1]).max(local)` and `g.inject([],[1]).max(local)`.
Same for other steps.

rdtr: @vkagamlyk 
```
this.starts.next();
```
is expected to throw `NoSuchElementException` at the end.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/MaxLocalStep.java
rdtr: I reverted `OptionalScalarMap`, in some case TinkerPop still wants to emit `null` so we can't always filter the input.

PR #2676, opened 2024-07-04
remove traverser serialization

vkagamlyk:
traverser serialization previously used only with Bytecode, so now it can be removed


vkagamlyk:
replaced with https://github.com/apache/tinkerpop/pull/2677

PR #2679, opened 2024-07-08
enable bulking for HTTP requests

vkagamlyk:
in TP3 bulking works only for Bytecode requests, implemented same functionality for DriverRemoteConnection queries.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/RequestOptions.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/RequestOptions.java
@@ -48,6 +48,7 @@ public final class RequestOptions {
     private final Long timeout;
     private final String language;
     private final String materializeProperties;
+    private final boolean bulking;
 
     private RequestOptions(final Builder builder) {
         this.graphOrTraversalSource = builder.graphOrTraversalSource;
@@ -56,6 +57,7 @@ public final class RequestOptions {
         this.timeout = builder.timeout;
         this.language = builder.language;
         this.materializeProperties = builder.materializeProperties;
+        this.bulking = builder.bulking;
     }
 
     public Optional<String> getG() {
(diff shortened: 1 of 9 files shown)


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/RequestOptions.java
```diff
@@ -152,6 +157,11 @@ public Builder batchSize(final int batchSize) {
             return this;
         }
 
+        public Builder withBulking(final boolean bulking) {
```
spmallette, line 160: nit: javadoc please

spmallette:
is the bulking feature optional for the server? it feels like it should be an optimization that the server can choose to implement or not rather than something required since it is now being treated as a parameter. i also wonder if the driver should be required to get that setting right for it to work. it was different when there was bytecode. the user (and driver provider) had no control over these options, but this is different now under http. 

also, needs changelog entry

vkagamlyk:
> is the bulking feature optional for the server? it feels like it should be an optimization that the server can choose to implement or not rather than something required since it is now being treated as a parameter. i also wonder if the driver should be required to get that setting right for it to work. it was different when there was bytecode. the user (and driver provider) had no control over these options, but this is different now under http.

I changed implementation that the client can consume the server's results both in the form of `RemoteTraverser` and flat data.
So server can return any result from now, providers may decide what is better is there use cases.

Cole-Greer:
Hey @vkagamlyk, there is a recent [dev-list post](https://lists.apache.org/thread/q7h5yzd2r064lv82njbmt6lmty4s24y7) which among other things proposes removing both traverser and bulkset from the serializers, and instead adds a "bulking" flag to GraphBinary's List type which could be used to retain the performance benefit of bulked results. Such an approach would make this PR obsolete. Do you have any thoughts on if we should proceed with the proposed changes and close out this PR, or if the proposed serializer changes should be amended to retain Traverser?

vkagamlyk:
> Hey @vkagamlyk, there is a recent [dev-list post](https://lists.apache.org/thread/q7h5yzd2r064lv82njbmt6lmty4s24y7) which among other things proposes removing both traverser and bulkset from the serializers, and instead adds a "bulking" flag to GraphBinary's List type which could be used to retain the performance benefit of bulked results. Such an approach would make this PR obsolete. Do you have any thoughts on if we should proceed with the proposed changes and close out this PR, or if the proposed serializer changes should be amended to retain Traverser?

I agree that the optional flag is the best solution that will allow to not add overhead and get a performance boost when needed

PR #2686, opened 2024-07-15
Add support for HTTP chunked transfer in gremlin-python

xiazcy:
Added support for http chunked transfer, removed websocket and replaced old serializers with graphbinaryv4. 

Tests pass in old HTTP server with bytecode, but will not pass current one until bytecode removal is completed in python and Docker testing with new http server is set up.


gremlin-python/src/main/python/gremlin_python/driver/serializer.py
```diff
@@ -386,10 +129,15 @@ def deserialize_message(self, message):
 
     def read_payload(self, buffer):
         results = []
-        while buffer.readable():
+        readable = True
+        while buffer.readable():  # find method or way to access readable bytes without using buffer.getvalue()
+            # print(buffer.tell())
```
vkagamlyk, line 134: nit

vkagamlyk (commented):
is it possible to fix or turn off invalid tests?

xiazcy:
> is it possible to fix or turn off invalid tests?

Yes, there is a flag to skip them. However, right now tests won't pass anyways because we are sending bytecode, and Docker isn't set up for it. I plan to remove bytecode next so we can re-enable all integration tests.

vkagamlyk:
VOTE+1

gremlin-python/src/main/python/gremlin_python/driver/protocol.py
```diff
@@ -203,6 +60,12 @@ def __init__(self,
         self._message_serializer = message_serializer
         self._username = username
         self._password = password
+        self._response_msg = {'status': {'code': 0,
```
kenhuuu, line 63: 0 is probably not the right default for code unless you are explicitly checking for 0 as a sentinel value.

kenhuuu:
VOTE +1

gremlin-python/src/main/python/gremlin_python/driver/protocol.py
xiazcy: I think I'll leave it for now, these values do get replaced in the chunking methods as soon as we read the response.

xiazcy:
VOTE +1

PR #2692, opened 2024-07-20
Remove redundant six dependency

hugovk:
The Python package no longer uses the six library, but still installs it as a dependency. Let's remove it to make installs a bit quicker and smaller.

Also universal wheels are for supporting both Python 2 and 3, so is no longer needed.

Do you still support Python 3.4? I don't think so, but couldn't find the min version documented. Well, this can be also removed if not:

```python
if sys.version_info < (3, 5):
    install_requires += ['pyparsing>=2.4.7,<3.0.0']
```

### DIFF
--- a/gremlin-python/docker-compose.yml
+++ b/gremlin-python/docker-compose.yml
@@ -67,7 +67,7 @@ services:
       bash -c "apt-get update && apt-get -y install libkrb5-dev krb5-user
       && echo 'password' | kinit stephen
       && klist
-      && pip install wheel radish-bdd PyHamcrest aenum isodate kerberos six
+      && pip install wheel radish-bdd PyHamcrest aenum isodate kerberos
       && python3 ./setup.py build
       && python3 ./setup.py test
       && python3 ./setup.py install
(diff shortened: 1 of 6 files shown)


Cole-Greer:
Hi @hugovk, Thanks for submitting this PR. Due to our branching structure, it would be best for this change to go into `3.6-dev` in order for the change to be incorporated in the existing 3.6.x and 3.7.x release lines, as well as future major releases.
Would you retargetting this PR to the `3.6-dev` branch?

> Do you still support Python 3.4? I don't think so, but couldn't find the min version documented. Well, this can be also removed if not:
Thanks for catching this. We do not support Python 3.4 anymore. Our current min version is python 3.8, however this will soon be raised to something more modern as 3.8 falls out of support.

hugovk:
Sure, coming up!

I didn't fully expect changing the base branch on the PR would work, but I'll rebase locally 😅

hugovk:
Updated! Also removed the mock backport dependency, we can use the stdlib version instead.

xiazcy:
Thank you for the update! VOTE +1

Cole-Greer:
Thanks @hugovk, looks great! Just so you know, our process requires 3 approving votes and then we can merge the PR.
My vote is VOTE +1

vkagamlyk:
VOTE+1

PR #2702, opened 2024-08-01
Bytecode removal in gremlin-python

xiazcy:
Replaced bytecode with gremlin lang scripts in gremlin-python, similar to what's been implemented in gremlin-java. 

Groovy translators are no longer needed, as well as some other code left over from WebSocket; will be cleaning them up along with bytecode codes in a separate PR after this is merged. 

Some TODOs are pending core changes, e.g. bulking, transactions. 

Local unit and integration tests pass, next step is to implement authentication interface, and modifying docker-compose set up to enable tests during build/GitHub actions.

~~Also added set in gremlin-java gremlin lang as set is now supported in grammar (can separate that into different PR/CTR if preferred).~~ Will open set changes for gremlin-java in separate PR due to additional changes needed.

### DIFF
--- a/gremlin-python/src/main/python/gremlin_python/driver/client.py
+++ b/gremlin-python/src/main/python/gremlin_python/driver/client.py
@@ -156,7 +156,8 @@ class Client:
         log.debug("message '%s'", str(message))
         fields = {'g': self._traversal_source}
 
-        # TODO: binding is part of request_options, evaluate the need to keep it separate in python
+        # TODO: bindings is now part of request_options, evaluate the need to keep it separate in python.
+        #  Note this bindings parameter only applies to string script submissions
         if isinstance(message, str) and bindings:
             fields['bindings'] = bindings
 
@@ -169,8 +170,9 @@ class Client:
             message.fields.update({token: request_options[token] for token in TokensV4
                                    if token in request_options})
             if 'params' in request_options:
-                if 'bindings' not in message.fields:
-                    message.fields['bindings'] = bindings if bindings else {}
-                message.fields['bindings'].update(request_options['params'])
+                if 'bindings' in message.fields:
+                    message.fields['bindings'].update(request_options['params'])
+                else:
+                    message.fields['bindings'] = request_options['params']
 
         return conn.write(message)
--- a/gremlin-python/src/main/python/gremlin_python/driver/client.py
+++ b/gremlin-python/src/main/python/gremlin_python/driver/client.py
@@ -24,6 +24,7 @@ from concurrent.futures import ThreadPoolExecutor
 
 from gremlin_python.driver import connection, protocol, request, serializer
 from gremlin_python.process import traversal
+from gremlin_python.driver.request import TokensV4
 
 log = logging.getLogger("gremlinpython")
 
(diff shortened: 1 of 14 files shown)


gremlin-python/src/main/python/gremlin_python/driver/driver_remote_connection.py
```diff
@@ -106,22 +108,26 @@ def cb(f):
...
+    #
+    # def rollback(self):
+    #     log.info("Submitting rollback graph operation.")
+    #     return self._client.submit(Bytecode.GraphOp.rollback())
 
     @staticmethod
-    def _extract_request_options(bytecode):
-        options_strategy = next((x for x in bytecode.source_instructions
-                                 if x[0] == "withStrategies" and type(x[1]) is OptionsStrategy), None)
+    def extract_request_options(gremlin_lang):
```
vkagamlyk, line 121: Could you add test for `g.with("evaluationTimeout", 1000).with("batchSize", 100).V()` to verify that more than one options_strategy is processed correctly?

gremlin-python/src/main/python/gremlin_python/driver/driver_remote_connection.py
```diff
@@ -106,22 +108,26 @@ def cb(f):
...
         if options_strategy:
-            allowed_keys = ['evaluationTimeout', 'scriptEvaluationTimeout', 'batchSize', 'requestId', 'userAgent',
-                            'materializeProperties']
-            request_options = {allowed: options_strategy[1].configuration[allowed] for allowed in allowed_keys
-                               if allowed in options_strategy[1].configuration}
+            request_options = {token: options_strategy.configuration[token] for token in TokensV4
+                               if token in options_strategy.configuration}
+        if gremlin_lang.parameters is not None and len(gremlin_lang.parameters) > 0:
+            if request_options is None:
+                request_options = gremlin_lang.parameters
```
vkagamlyk, line 129: good to have test coverage

gremlin-python/src/main/python/gremlin_python/driver/request.py
```diff
@@ -22,3 +22,5 @@
 
 RequestMessageV4 = collections.namedtuple(
     'RequestMessageV4', ['fields', 'gremlin'])
+
+TokensV4 = ['evaluationTimeout', 'batchSize', 'language', 'materializeProperties']
```
vkagamlyk, line 26: Should "g" be in this list?

vkagamlyk (changes requested):
thank you, great job! with a couple of small nits

gremlin-python/src/main/python/gremlin_python/driver/request.py
xiazcy: Actually yes, I'll add it.

gremlin-python/src/main/python/gremlin_python/driver/client.py
```diff
@@ -154,20 +155,19 @@ def submit_async(self, message, bindings=None, request_options=None):
...
+        if isinstance(message, str):
             log.debug("fields='%s', gremlin='%s'", str(fields), str(message))
             message = request.RequestMessageV4(fields=fields, gremlin=message)
 
         conn = self._pool.get(True)
         if request_options:
-            message.fields.update(request_options)
+            message.fields.update({token: request_options[token] for token in TokensV4
+                                   if token in request_options})
+            if 'params' in request_options:
```
Cole-Greer, line 170: Is there a distinction between `bindings` and `request_options['params']`? Do we need to have both or can we simplify this interface a bit?

If we do keep both, would it be better to merge both dicts together instead of taking either `bindings` and `request_options['params']`?

gremlin-python/src/main/python/gremlin_python/driver/driver_remote_connection.py
xiazcy: Added in drc test

gremlin-python/src/main/python/gremlin_python/driver/client.py
xiazcy: I'm fairly certain that bindings should be sent within request_options, but python has been having it separated. I'm not sure if we should change the API with this PR yet, so I've updated the `request_options['params']` to add to the `bindings` dict instead.

vkagamlyk: good to have test for case when both `bindings` and request_options are present

vkagamlyk (approved):

vkagamlyk:
python GLV build is broken, but it seems not because of changes in this PR, the same problem is in other branches.

VOTE+1

Cole-Greer:
VOTE +1

xiazcy:
Thank you for all the reviews! VOTE +1

PR #2712, opened 2024-08-09
Regression: Graceful shutdowns for GremlinServer

danielcweber:
Fixes what seems to be a regression of [TINKERPOP-2950.](https://issues.apache.org/jira/browse/TINKERPOP-2950). Originally fixed in #2397.

With this change, the gremlin-server process will end up as PID 1 and receive all the signals, specifically SIGTERM, and allow the server to shutdown gracefully (and e.g. write a graph to a volume).


kenhuuu:
VOTE +1

Cole-Greer:
Thanks @danielcweber, sorry it took a little while to get a review in.
VOTE +1

PR #2731, opened 2024-08-19
Add pluggable auth in gremlin-python with HTTP

xiazcy:
Referencing the Java driver, added an interface for authentication with HTTP, with basic and SigV4 auth as reference implementations. e.g. `username='stephen', password='password'` is replaced by `auth=basic('stephen', 'password')`.

Local tests pass, will be updating the docker compose set-up as next step for integration tests.

### DIFF
--- a/gremlin-python/src/main/python/gremlin_python/driver/auth.py
+++ b/gremlin-python/src/main/python/gremlin_python/driver/auth.py
@@ -1,0 +1,55 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+
+def basic(username, password):
+    from aiohttp import BasicAuth as aiohttpBasicAuth
+
+    def apply(request):
+        return request['headers'].update({'authorization': aiohttpBasicAuth(username, password).encode()})
+
+    return apply
+
+
+def sigv4(region, service):
+    import os
+    from boto3 import Session
+    from botocore.auth import SigV4Auth
+    from botocore.awsrequest import AWSRequest
+
+    def apply(request):
+        access_key = os.environ.get('AWS_ACCESS_KEY_ID', '')
+        secret_key = os.environ.get('AWS_SECRET_ACCESS_KEY', '')
+        session_token = os.environ.get('AWS_SESSION_TOKEN', '')
+
+        session = Session(
+            aws_access_key_id=access_key,
+            aws_secret_access_key=secret_key,
+            aws_session_token=session_token,
+            region_name=region
+        )
+
+        sigv4_request = AWSRequest(method="POST", url=request['url'], data=request['payload'])
+        SigV4Auth(session.get_credentials(), service, region).add_auth(sigv4_request)
+        request['headers'].update(sigv4_request.headers)
+        request['payload'] = sigv4_request.data
+        return request
+
+    return apply
+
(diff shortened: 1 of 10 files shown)


gremlin-python/src/main/python/gremlin_python/driver/auth.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class SigV4Auth(Auth):
+
+    def __init__(self, region_name, aws_access_key_id='', aws_secret_access_key='', session_token='',
+                 service_name=''):
+        import os
+
+        self._region_name = region_name
+        self._aws_access_key_id = aws_access_key_id if aws_access_key_id else os.environ.get('AWS_ACCESS_KEY_ID')
```
vkagamlyk, line 57: may be better to use boto to get credentials?
something like
`session = botocore.Session()
credentials = session.get_credentials()`

gremlin-python/src/main/python/gremlin_python/driver/auth.py
```diff
@@ -0,0 +1,81 @@
...
+            else os.environ.get('AWS_SECRET_ACCESS_KEY')
+        self._session_token = session_token if session_token else os.environ.get('AWS_SESSION_TOKEN')
+        self._service_name = service_name if service_name else "neptune-db"
+
+    def apply(self, request):
+        from botocore.auth import SigV4Auth as botocoreSigV4Auth
+        from botocore.awsrequest import AWSRequest
+        from types import SimpleNamespace
+
+        assert ((self._aws_access_key_id is not None and self._aws_secret_access_key is not None)
```
vkagamlyk, line 68: will it work with session token?

gremlin-python/src/main/python/gremlin_python/driver/auth.py
```diff
@@ -0,0 +1,81 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+import abc
+
+
+class Auth(metaclass=abc.ABCMeta):
```
kenhuuu, line 22: Following the Java implementation here seems a bit odd to me. The Auth interface only has one function apply() which makes it like a functional interface. This is needed in Java because functions aren't first class. The same isn't true for Python. In gremlin-python, you should just accept a list of functions that take in one parameter for the request. This function would be the interceptor and should be allowed to change the request as needed. The auth implementation should use that same underlying interceptor mechanism to update the rquest.

gremlin-python/src/main/python/gremlin_python/driver/auth.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @abc.abstractmethod
+    def apply(self, request):
+        """Applies the necessary authentication operations to the request and returns the modified request."""
+        pass
+
+    @staticmethod
+    def basic(username, password):
```
kenhuuu, line 30: `basic` and `sigv4` could probably just be part of the auth module rather than part of the abstract base class.

gremlin-python/src/main/python/gremlin_python/driver/auth.py
```diff
@@ -0,0 +1,81 @@
...
+        return request['headers'].update({'authorization': aiohttpBasicAuth(self._username, self._password).encode()})
+
+
+class SigV4Auth(Auth):
+
+    def __init__(self, region_name, aws_access_key_id='', aws_secret_access_key='', session_token='',
+                 service_name=''):
+        import os
+
+        self._region_name = region_name
```
Cole-Greer, line 56: Adding to Valentyn's comment below, `region_name` could be made optional and default to `session.region_name`

gremlin-python/src/main/python/gremlin_python/driver/auth.py
xiazcy: Updated to using boto3.Session() based on [sigV4 examples](https://github.com/aws-samples/sigv4-signing-examples/blob/main/sdk/python/main.py)

gremlin-python/src/main/python/gremlin_python/driver/auth.py
xiazcy: Updated

gremlin-python/src/main/python/gremlin_python/driver/auth.py
xiazcy: Region is still a required input for sigV4 signing, even with session.region_name, it looks like it needs to be a user input (or env variable). I'm leaning towards required user input (as with Java driver).

kenhuuu:
VOTE +1

Cole-Greer (approved):
Thanks @xiazcy. LGTM

VOTE +1

xiazcy:
VOTE +1. Thanks for all the reviews! I will merge this now.

PR #2736, opened 2024-08-22
Adding a public getter for Repeat Loop Name

upadhyay-prashant:


saikiranboga (approved):

kenhuuu:
Merging this change as CTR since it's just adding a public getter.

PR #2738, opened 2024-08-22
Fix issue with DotNet 8 when a traverser is empty

MavroudisMant:
I found an issue with the driver when upgrading to .NET 8.0, caused from changes on how the Enumerator works.
Current is always in a state of throwing an exception if the Traversers are empty. Before MoveNext() is called it throws "Enumeration has not started. Call MoveNext." and after the call it throws "System.InvalidOperationException: Enumeration already finished."

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Process/Traversal/DefaultTraversal.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Process/Traversal/DefaultTraversal.cs
@@ -198,9 +198,10 @@ namespace Gremlin.Net.Process.Traversal
         /// </summary>
         /// <returns>The result.</returns>
         public E Next()
-        {
-            MoveNext();
-            return Current;
+        { 
+            return MoveNext() 
+                ? Current 
+                : default;
         }
 
         /// <summary>
--- a/gremlin-dotnet/src/Gremlin.Net/Process/Traversal/DefaultTraversal.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Process/Traversal/DefaultTraversal.cs
@@ -199,9 +199,7 @@ namespace Gremlin.Net.Process.Traversal
         /// <returns>The result.</returns>
         public E Next()
         { 
-            return MoveNext() 
-                ? Current 
-                : default;
+            return MoveNext() ? Current : default;
         }
 
         /// <summary>
--- a/gremlin-dotnet/test/Gremlin.Net.UnitTest/Process/Traversal/TraversalTests.cs
+++ b/gremlin-dotnet/test/Gremlin.Net.UnitTest/Process/Traversal/TraversalTests.cs
@@ -98,6 +98,17 @@ namespace Gremlin.Net.UnitTest.Process.Traversal
 
             Assert.Null(drainedTraversal.Next());
         }
+        
+        [Fact]
+        public void ShouldReturnNullWhenTraverserIsEmpty()
+        {
+            var someObjs = new List<object?>();
+            var traversal = new TestTraversal(someObjs);
+
+            var emptyTraversal = traversal.Iterate();
+
+            Assert.Null(emptyTraversal.Next());
+        }
 
         [Fact]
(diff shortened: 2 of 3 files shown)


gremlin-dotnet/src/Gremlin.Net/Process/Traversal/DefaultTraversal.cs
```diff
@@ -198,9 +198,10 @@ public bool HasNext()
         /// </summary>
         /// <returns>The result.</returns>
         public E Next()
-        {
-            MoveNext();
-            return Current;
+        { 
+            return MoveNext()
```
kenhuuu, line 202: Small nit: this should probably be one line to follow the style that exists in our cs files.

gremlin-dotnet/test/Gremlin.Net.UnitTest/Process/Traversal/TraversalTests.cs
```diff
@@ -98,6 +98,17 @@ public void ShouldDrainAllTraversersWhenIterateIsCalled()
...
+        
+        [Fact]
+        public void ShouldReturnNullWhenTraverserIsEmpty()
+        {
+            var someObjs = new List<object?>();
+            var traversal = new TestTraversal(someObjs);
+
+            var emptyTraversal = traversal.Iterate();
+
+            Assert.Null(emptyTraversal.Next());
```
kenhuuu, line 110: I'm not sure if returning `null` is the more proper behavior. `Iterate()` is called on the Traversal which will move to the end of the enumerator. Then `Next()` is called which I think would be reasonable to throw an exception.

kenhuuu:
Hi and thank you for contributing.

As I mentioned in a comment in the test file, I'm not sure if this change produces the right behavior. @FlorianHockmann do you have an opinion on this?

MavroudisMant:
Hello and thanks for taking the time to review this. 

I run the above test on dotnet 6 without my change on MoveNext() and it produced `null`, so I tried to make it have the same behaviour on dotnet 8.
In our case when we upgraded our first issue was on a query like this one `await g.V(id).Promise(t => t.Next());` when there wasn't a vertex with the given id.

kenhuuu:
> I run the above test on dotnet 6 without my change on MoveNext() and it produced `null`, so I tried to make it have the same behaviour on dotnet 8.

Yea, generally, this makes sense to me since our policy is to prevent breaking changes. However, in this specific case, that old behavior was kind of undefined behavior and we've been OK with changing it in 3.6-dev recently (see https://github.com/apache/tinkerpop/pull/2424#issue-2064310991)

I don't have a strong opinion on this topic because I'm not that familiar with C# so if no one else that is experienced in C# opposes it then I'm fine with supporting this PR as is.

kenhuuu:
VOTE +1

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/DefaultTraversal.cs
MavroudisMant: Updated this to follow the style on the rest of the files

FlorianHockmann:
Sorry for the late response @MavroudisMant and thanks for reporting this and offering a PR!

However, I think that we shouldn't change this behavior. I tried to describe the situation in this JIRA issue: [TINKERPOP-3072](https://issues.apache.org/jira/browse/TINKERPOP-3072).

The short version is that Gremlin also throws an exception in other languages like Java and Python if `next()` is called on a traversal that doesn't return any elements. Java throws a nice `NoSuchElementException` which is of course more descriptive than `System.InvalidOperationException: Enumeration already finished`, but it doesn't just silently return `null`.

So, I think we can improve the situation in .NET by also throwing an exception with a clear error message to inform the user that there are no elements to return, but I don't think that we should return `null` here.

`null` could also be the valid result of a traversal which is something different than having a traversal which doesn't return anything but the user still calls `Next()` on it and therefore expects a result.

kenhuuu:
Based on Florian's explanation of the situation, I think we should close this PR as we no longer want the Traversal to behave this way moving forward.

PR #2741, opened 2024-08-22
TINKERPOP-3083: The split() step should provide a way to split an entire string

andreachild:
https://issues.apache.org/jira/browse/TINKERPOP-3083

Changed split() local and global steps to parse a string into individual characters if the separator is an empty string - previous implementation left the string intact. Introduced StringUtil utility class to reuse logic between the global and local step classes. Feature tests were executed using `-DskipIntegrationTests=false` and manual testing was executed using gremlin console with and without remote context.


docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -1517,7 +1517,7 @@ link:https://tinkerpop.apache.org/docs/x.y.z/reference/#rTrim-step[reference]
 *Arguments:*
 
 * `separator` - The string character(s) used as delimiter to split the input string. Nullable, a null separator will split on
-whitespaces.
+whitespaces. An empty separator will split on each character.
```
kenhuuu, line 1520: Minor nit: I think it should probably be "empty string separator"

kenhuuu:
Requires a CHANGELOG entry.

VOTE +1 pending minor nits. Thanks for contributing.

docs/src/dev/provider/gremlin-semantics.asciidoc
andreachild: Changed to 'empty string separator' for this file and also the-traversal.asciidoc.

Cole-Greer:
Thanks @andreachild, looks great!

VOTE +1

xiazcy:
VOTE +1, thanks @andreachild!

PR #2742, opened 2024-08-23
[TINKERPOP-3105] Running 3.6.x python-driver with 3.7.x server leads to deserialization errors

xiazcy:
Currently, when skipping properties either via `tokens` or `ReferenceElementStrategy` with `GraphBinaryV1`, properties on element will return the default empty list, however drivers of older versions are expecting `null` properties for Reference Elements, as such serialization errors can occur in GLVs that does strict checking (i.e. `python` and `go`).

This PR updated the check to ReferenceVertex/Edge as `value.properties() == null` won't be reached, and will write `null` for properties of Reference Elements in the GraphBinary serializer. 

Also updated docs to make it more clear.


kenhuuu:
Did you do the same check for VertexPropertySerializer or are metaproperties expected to be empty list instead of null.

xiazcy:
> Did you do the same check for VertexPropertySerializer or are metaproperties expected to be empty list instead of null.

Yes, that's now also updated to write `null` instead of empty list as expected by older drivers.

spmallette:
Does GraphSON have these problems with 3.6.x drivers connecting to 3.7x servers?

vkagamlyk:
VOTE+1

xiazcy:
> Does GraphSON have these problems with 3.6.x drivers connecting to 3.7x servers?

No, GraphSON works fine, I've only seen errors in GraphBinary when I tested them. I'll update the docs to specify that.

docs/src/reference/gremlin-applications.asciidoc
```diff
@@ -2336,6 +2336,44 @@ List<Vertex> results = g.V().hasLabel("person").elementMap('name').toList();
 
 Both of the above requests return a list of `Map` instances that contain the `id`, `label` and the "name" property.
 
+*Compatibility*
+
+*It is not recommended to use older driver versions with 3.7.x Gremlin Server*, as some older drivers do not construct
```
Cole-Greer, line 2341:
```suggestion
*It is not recommended to use 3.6.x or below driver versions with 3.7.x or above Gremlin Server*, as some older drivers do not construct
```
This might be a bit more future proof as new releases go out.

docs/src/reference/gremlin-applications.asciidoc
```diff
@@ -2336,6 +2336,44 @@ List<Vertex> results = g.V().hasLabel("person").elementMap('name').toList();
...
+|Deserialization error
+|Returns null
+
+|`gremlin-python`
+|Deserialization error
+|Skips properties in Elements
+
+|`gremlin-go`
+|Deserialization error
+|Skips properties in Elements
```
Cole-Greer, line 2374: Some of these may still be confusing for some users. It's not clear what the result of skipping properties will be. It might be best to focus on what the actual de-serialized vertex will look like (when there's no deserialization errors). Something like `vertex.properties = []` and `vertex.properties = null`

xiazcy: By skip, I mean those GLVs don't have `properties` in their `Vertex` object, so actually the `vertex.properties` are non-existent keys/attributes. I can make that part more clear.

Cole-Greer:
VOTE +1

PR #2749, opened 2024-08-28
Java driver and gremlin server maxContentLength changes for http

andreachild:
Java driver:
* renamed maxContentLength to maxResponseContentLength to clarify that the check is done against the response and not the request
* increased default value to Integer.MAX_VALUE as the size check is now done against the total http response size instead of previous web socket frame size
* changed int to long to allow for higher max value
* changed zero to be an acceptable value which disables the response size check

Gremlin server:
* renamed maxContentLength to maxRequestContentLength to clarify that the check is done against the request and not the response

Feature tests were executed for java driver and pass (other GLV feature tests currently do not work for master-http branch). Manual testing of the changes were done using local gremlin-console with remote connection configured to local gremlin server.


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/HttpGremlinResponseStreamDecoder.java
```diff
@@ -79,8 +79,8 @@ protected void decode(ChannelHandlerContext ctx, DefaultHttpObject msg, List<Obj
             ByteBuf content = ((HttpContent) msg).content();
             Attribute<Integer> bytesRead = ctx.channel().attr(BYTES_READ);
             bytesRead.set(bytesRead.get() + content.readableBytes());
-            if (bytesRead.get() > maxContentLength) {
-                throw new TooLongFrameException("Response exceeded " + maxContentLength + " bytes.");
+            if (maxResponseContentLength > 0 && bytesRead.get() > maxResponseContentLength) {
```
Cole-Greer, line 82: We are mixing types here which could cause problems. `bytesRead.get()` is an `Integer` while `maxResponseContentLength` is a `double`. Any `maxResponseContentLength` greater than `Integer.MAX_VALUE` will never be hit.

andreachild: Whoops, thanks for catching.

Cole-Greer:
Thanks @andreachild, LGTM
VOTE +1

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/HttpGremlinResponseStreamDecoder.java
andreachild: Changed to store the bytes read as Double instead of Integer.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
```diff
@@ -328,10 +328,10 @@ public int getMaxWaitForClose() {
     }
 
     /**
-     * Gets the maximum size in bytes of any request sent to the server.
+     * Gets the maximum size in bytes of any request received from the server.
      */
-    public int getMaxContentLength() {
-        return manager.connectionPoolSettings.maxContentLength;
+    public double getMaxResponseContentLength() {
```
vkagamlyk, line 333: Why not `Long`?

andreachild: Hello @vkagamlyk thanks for taking a look at my PR!
It was suggested by @kenhuuu that using double would provide more flexibility to users that need a maximum that is larger than the Long.MAX_VALUE.

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -797,7 +797,7 @@ The following table describes the various configuration options for the Gremlin
 |connectionPool.keyStore |The private key in JKS or PKCS#12 format. |_none_
 |connectionPool.keyStorePassword |The password of the `keyStore` if it is password-protected. |_none_
 |connectionPool.keyStoreType |`JKS` (Java 8 default) or `PKCS12` (Java 9+ default)|_none_
-|connectionPool.maxContentLength |The maximum length in bytes that a message can be sent to the server. This number can be no greater than the setting of the same name in the server configuration. |65536
+|connectionPool.maxResponseContentLength |The maximum length in bytes that a message can be received from the server. |`Integer.MAX_VALUE`
```
kenhuuu, line 800: Might be nice to just write out 2147483647 instead of Integer.MAX_VALUE. Saves people a step from having to look it up.

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/handler/HttpGremlinResponseStreamDecoderTest.java
```diff
@@ -0,0 +1,102 @@
...
+        final FullHttpResponse httpResponse = createResponse(largeResponse);
+        final EmbeddedChannel testChannel = initializeChannel(0);
+
+        testChannel.writeInbound(httpResponse);
+        final ResponseMessage inbound = testChannel.readInbound();
+        assertEquals(largeResponse, inbound.getResult().getData().get(0));
+    }
+
+    @Test
+    public void shouldSucceedIfResponseSizeLargerThanMaxResponseContentLength() throws SerializationException {
```
kenhuuu, line 78: The name of this test is bit misleading. `shouldSucceed` when the assert is that an exception occurs.

kenhuuu:
VOTE +1

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
vkagamlyk: Long.MAX_VALUE is 9 223 372 terabytes. Do users really need that kind of range?

andreachild: Discussed with @kenhuuu and will change to Long.

andreachild: Changed to long

docs/src/reference/gremlin-variants.asciidoc
andreachild: Changed to 2147483647

gremlin-driver/src/test/java/org/apache/tinkerpop/gremlin/driver/handler/HttpGremlinResponseStreamDecoderTest.java
andreachild: Changed to 'shouldThrow'

vkagamlyk:
VOTE+1

PR #2753, opened 2024-08-30
Reinstate handshake error handling on channel connect

andreachild:
Fixed failing GremlinServerSslIntegrateTest.shouldEnableSslAndFailIfCiphersDontMatch by reinstating the previous WebSocket channelizer logic that waited for the handshake to complete after the channel connects. If the handshake fails then a ConnectionException is thrown.

Tested by executing feature tests for gremlin-server using -DskipIntegrationTests=false and manually by starting up a gremlin-server with ssl enabled and gremlin-console with ssl enabled (with limited sslCipherSuites) and then executing a remote query to view the correct exception is thrown.


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
```diff
@@ -87,6 +88,7 @@ default String getScheme(final boolean sslEnabled) {
     abstract class AbstractChannelizer extends ChannelInitializer<SocketChannel> implements Channelizer {
         protected Connection connection;
         protected Cluster cluster;
+        protected SslHandler sslHandler;
```
kenhuuu, line 91: Is this safe to do? The SslHandler isn't annotated with "Shareable".

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
```diff
@@ -149,6 +155,18 @@ protected void initChannel(final SocketChannel socketChannel) {
...
         }
+
+        @Override
+        public void connected() {
+            if (supportsSsl()) {
+                try {
+                    // Block until the handshake is complete either successfully or with an error.
+                    sslHandler.handshakeFuture().sync();
+                } catch (Exception ex) {
+                    throw new ConnectionException(connection.getUri(), HANDSHAKE_ERROR, ex);
```
kenhuuu, line 166: This might leak the channel. Can you confirm if something actually closes the channel if an exception is thrown here?

andreachild: When there is a handshake failure, the SslHandler closes the channel via call chain: SslUtils.handleHandshakeFailure-->ChannelHandlerContext.close().

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
andreachild: There should not be multiple threads setting the SslHandler as it is initialized as part of the Connection constructor which calls `new Channelizer.HttpChannelizer()` and eventually `ChannelInitializer.initChannel` which only executes `initChannel(Channel)` once per `ChannelHandlerContext`.

kenhuuu:
VOTE +1

Cole-Greer:
VOTE +1

PR #2758, opened 2024-09-04
Update IO Model to reflect v4 serialization types

kenhuuu:
The model is up-to-date for the current implementation. Will need to have further updates as it doesn't match GraphBinaryV4 yet.

VOTE +1


Cole-Greer (approved):
Thanks @kenhuuu, this looks great. As you mentioned, these tests will need updates some of the serializer changes agreed in the dev-list get implemented. I think these tests are ready to merge now, and we can update them as development progresses.

Once all TP4 changes to the serializers are implemented, I'd like to revisit the total set of models and assess if there are any cases missing.

VOTE +1

xiazcy:
VOTE +1. Thanks @kenhuuu!

PR #2776, opened 2024-09-12
Update `DateTime` serializer according to GraphBinaryV4 IO specs 

xiazcy:
In light of the recently discussed [updates](https://lists.apache.org/thread/q7h5yzd2r064lv82njbmt6lmty4s24y7) to V4 IO spec, `DateTime` is the new serialized type for `Date`, and is based off of the previous `OffsetDateTime`. This PR updates the Java and Python serializers to align with the new spec, as well as the Gremlin Core components that used to use the Java `Date` objet and is now the `OffsetDateTime`.

This PR also includes removal of serializers that will no longer be in the TP4 I/O spec.

PR #2779, opened 2024-09-17
Update Map serializer according to GraphBinaryV4 IO specs

xiazcy:
Changes:
* Add ordered `value_flag` `0x02` to Map serializers in Java and Python drivers.

Note this branch was based off of [graphbinary-add-datetime](https://github.com/apache/tinkerpop/tree/graphbinary-add-datetime) feature branch, as such the PR is to that branch to minimize file diff and facilitate review. Target branch will be changed to `master-http` once the feature branch is merged in https://github.com/apache/tinkerpop/pull/2776.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/types/MapSerializer.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/types/MapSerializer.java
@@ -24,13 +24,31 @@ import org.apache.tinkerpop.gremlin.structure.io.binary.GraphBinaryWriter;
 import org.apache.tinkerpop.gremlin.structure.io.Buffer;
 
 import java.io.IOException;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
 
 public class MapSerializer extends SimpleTypeSerializer<Map> {
+
     public MapSerializer() {
         super(DataType.MAP);
+    }
+
+    @Override
+    public Map readValue(final Buffer buffer, final GraphBinaryReader context, final boolean nullable) throws IOException {
+        if (nullable) {
+            final byte valueFlag = buffer.readByte();
+            if ((valueFlag & 1) == 1) {
+                return null;
+            }
+            if ((valueFlag & 2) == 2) {
+                return readValue(buffer, context);
+            }
+
+        }
+
+        return readMap(buffer, context);
     }
 
     @Override
@@ -38,6 +56,17 @@ public class MapSerializer extends SimpleTypeSerializer<Map> {
         final int length = buffer.readInt();
 
         final Map<Object,Object> result = new LinkedHashMap<>(length);
+        for (int i = 0; i < length; i++) {
+            result.put(context.read(buffer), context.read(buffer));
+        }
+
+        return result;
+    }
+
+    private Map readMap(final Buffer buffer, final GraphBinaryReader context) throws IOException {
+        final int length = buffer.readInt();
+
+        final Map<Object,Object> result = new HashMap<>(length);
         for (int i = 0; i < length; i++) {
             result.put(context.read(buffer), context.read(buffer));
         }
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/GraphBinaryWriter.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/GraphBinaryWriter.java
@@ -45,6 +45,7 @@ public class GraphBinaryWriter {
     private final TypeSerializerRegistry registry;
     private final static byte VALUE_FLAG_NULL = 1;
     private final static byte VALUE_FLAG_NONE = 0;
+    private final static byte VALUE_FLAG_ORDERED = 2;
     public final static byte VERSION_BYTE = (byte)0x81;
     private final static byte[] unspecifiedNullBytes = new byte[] { DataType.UNSPECIFIED_NULL.getCodeByte(), 0x01};
     private final static byte[] customTypeCodeBytes = new byte[] { DataType.CUSTOM.getCodeByte() };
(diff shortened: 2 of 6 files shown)


vkagamlyk:
Have you considered the option of adding new types and serializers for OrderedMap and OrderedList (or set)?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/types/BigDecimalSerializer.java
```diff
@@ -36,7 +36,7 @@ public BigDecimalSerializer() {
     }
 
     @Override
-    protected BigDecimal readValue(final Buffer buffer, final GraphBinaryReader context) throws IOException {
+    protected BigDecimal readValue(final Buffer buffer, final GraphBinaryReader context, byte valueFlag) throws IOException {
```
spmallette, line 39: `valueFlag`  should be `final` - please review all use of `final` in this PR

xiazcy:
> Have you considered the option of adding new types and serializers for OrderedMap and OrderedList (or set)?

Actually my wording might have been misleading, only Map will have the ordered flag. List is ordered by definition and Set not ordered, and we aren't changing those. 
Lists however will be using the flag 0x02 for a different purpose (bulking), so this flag change was needed anyways. (see [devlist](https://lists.apache.org/thread/zsd7o7t6hqncc8hzkm47hkz3omdrfm9g) and [updated IO](https://github.com/apache/tinkerpop/blob/master-http/docs/src/dev/io/graphbinary.asciidoc))

xiazcy:
VOTE +1

kenhuuu:
I think you can reduce the number of files changed in this PR. It makes sense to do one of:

- Add a readValueFlag(), writeValueFlag() that uses the current implementation as the default implementation
- Make MapSerializer override `public T readValue(final Buffer buffer, final GraphBinaryReader context, final boolean nullable)` and `public void writeValue(final T value, final Buffer buffer, final GraphBinaryWriter context, final boolean nullable)`

Either of those options makes more sense than changing the signature to the current abstract method as we already know that in v4, only map and list will need to modify those values.

xiazcy:
> * Make MapSerializer override `public T readValue(final Buffer buffer, final GraphBinaryReader context, final boolean nullable)` and `public void writeValue(final T value, final Buffer buffer, final GraphBinaryWriter context, final boolean nullable)`

Makes sense, I updated to overriding only the readValue in map serializer, since writeValue just needed an if/else to write the flag I don't think it needs a specific override.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/types/MapSerializer.java
```diff
@@ -24,20 +24,37 @@
...
 
 import java.io.IOException;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
 
 public class MapSerializer extends SimpleTypeSerializer<Map> {
+
+    private byte valueFlag = (byte)0x00;
```
kenhuuu, line 34: Although not explicitly stated, I think the serializers are thread-safe because they don't make use of member variables. So don't use a member variable here as it would break that.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/types/MapSerializer.java
```diff
@@ -24,20 +24,37 @@
...
+
+        return readValue(buffer, context);
+    }
+
     @Override
     protected Map readValue(final Buffer buffer, final GraphBinaryReader context) throws IOException {
         final int length = buffer.readInt();
 
-        final Map<Object,Object> result = new LinkedHashMap<>(length);
+        // original read value outputs LinkedHashMap, do we want to keep all result as LinkedHashMap anyway?
```
Cole-Greer, line 56: This question does have me wondering if we've landed on the right design with this ordered map flag. In theory the only downside of using LinkedHashMap all the time is a slight increase in overhead to maintain the links. Considering Java has always converted all maps to LinkedHashMap upon deserialization, and that JS and Python both implicitly deserialize to ordered maps, I'm wondering if the best answer is to update our spec to state "All gremlin language drivers should attempt to deserialize to an ordered map wherever possible".

As far as the official GLVs go, this would require a switch in .Net to using `OrderedDictionary`, whereas Go would need to have lack of ordering added as a documented limitation.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/types/MapSerializer.java
```diff
@@ -24,19 +24,50 @@
...
+
     public MapSerializer() {
         super(DataType.MAP);
     }
 
+    @Override
+    public Map readValue(final Buffer buffer, final GraphBinaryReader context, final boolean nullable) throws IOException {
+        if (nullable) {
+            // save the reader index when we check if value is null or not
+            final int readerIdx = buffer.readerIndex();
```
kenhuuu, line 42: Nit: unused

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/types/MapSerializer.java
```diff
@@ -24,19 +24,50 @@
...
+        }
+
+        return readValue(buffer, context);
+    }
+
     @Override
     protected Map readValue(final Buffer buffer, final GraphBinaryReader context) throws IOException {
         final int length = buffer.readInt();
 
+        final Map<Object,Object> result = new HashMap<>(length);
```
kenhuuu, line 60: Nit: Probably makes sense to swap this around as the previous behavior defaulted to LinkedHashMap.

kenhuuu:
VOTE +1, pending changes to nits.

PR #2790, opened 2024-09-26
Update GraphSONv4 labels, extended type names and edge properties.

kenhuuu:
These updates bring the implementation closer to the format specified in https://github.com/apache/tinkerpop/blob/master-http/docs/src/dev/io/graphson.asciidoc

VOTE +1

### DIFF
--- a/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONMapperTest.java
+++ b/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONMapperTest.java
@@ -88,10 +88,12 @@ public class GraphSONMapperTest {
         // v2 untyped seems to serialize the VertexProperty label. not changing that since it's already been
         // introduced a long while back. v3 dips back to v1 style - since it has never existed prior to 3.7.0
         // we can make this change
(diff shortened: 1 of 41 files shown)


xiazcy:
VOTE +1

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONMapperTest.java
```diff
@@ -150,8 +162,10 @@ public void shouldHandlePath() throws Exception {
                 extend("alice", new HashSet<>(Collections.singletonList("c")));
         final String json = mapper.writeValueAsString(p);
 
-        if (version.startsWith("v1") || version.startsWith("v3") || version.startsWith("v4"))
+        if (version.startsWith("v1") || version.startsWith("v3"))
```
vkagamlyk, line 165: do we need tests for v1 and v3?

kenhuuu: Probably. The old GraphSONMappers should stay as they can be used with io(). It would be best if users updated their storage format to v4, but all versions should be provided so that the user can read in an older version then write into v4.

xiazcy:
VOTE +1

Cole-Greer:
VOTE +1

PR #2799, opened 2024-09-27
Update RequestInterceptor to better work with HTTP.

kenhuuu:
The interceptor is updated to run for every request to allow modifying each HTTP request. It works on a newly created abstraction of a request called HttpRequest to prevent breaking changes in the event the underlying library ever changes.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
@@ -29,7 +29,9 @@ import io.netty.handler.ssl.util.InsecureTrustManagerFactory;
 import io.netty.util.concurrent.Future;
 import org.apache.commons.configuration2.Configuration;
 import org.apache.commons.lang3.concurrent.BasicThreadFactory;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.tinkerpop.gremlin.driver.auth.Auth;
+import org.apache.tinkerpop.gremlin.driver.interceptor.GraphBinarySerializationInterceptor;
 import org.apache.tinkerpop.gremlin.util.MessageSerializer;
 import org.apache.tinkerpop.gremlin.util.message.RequestMessage;
 import org.apache.tinkerpop.gremlin.util.ser.Serializers;
@@ -57,6 +59,7 @@ import java.security.cert.CertificateException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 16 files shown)


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
```diff
@@ -732,16 +739,82 @@ public Builder loadBalancingStrategy(final LoadBalancingStrategy loadBalancingSt
...
+            return INTERCEPTOR_NOT_FOUND;
+        }
+
+        /**
+         * Adds an Auth {@link RequestInterceptor} to the end of list of interceptors.
+         */
         public Builder auth(final Auth auth) {
-            interceptors.add(auth);
+            interceptors.add(Pair.of(auth.getClass().getSimpleName().toLowerCase() + "-auth", auth));
             return this;
```
Cole-Greer, line 818: Nit: Utilize `addInterceptor` for the slight benefit of the name collision safety check.
```suggestion
            return addInterceptor(auth.getClass().getSimpleName().toLowerCase() + "-auth", auth);
```

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
```diff
@@ -732,16 +739,82 @@ public Builder loadBalancingStrategy(final LoadBalancingStrategy loadBalancingSt
...
+
+            return INTERCEPTOR_NOT_FOUND;
+        }
+
+        /**
+         * Adds an Auth {@link RequestInterceptor} to the end of list of interceptors.
+         */
         public Builder auth(final Auth auth) {
-            interceptors.add(auth);
+            interceptors.add(Pair.of(auth.getClass().getSimpleName().toLowerCase() + "-auth", auth));
```
Cole-Greer, line 817: Quick question, since interceptors are considered an advanced feature, is it assumed that anyone using interceptors needs to understand the impact that interceptor ordering may have on auth?

Cole-Greer:
Thanks Ken!

VOTE +1 pending response to minor questions.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
kenhuuu: Yes, that is the intended usage. Essentially, if you are going to use the interceptors, you'll have to become intimately aware of exactly what other interceptors are used and how they interact with one another.

The driver's default interceptors, however, should be able to be used without the user understanding the ordering of interceptors. For example, if you are using basic auth, it should just work as the auth() method should know where to place that interceptor in relation to the default "serializer" interceptor.

kenhuuu:
VOTE +1. Merging with lazy consensus from previous +1.

PR #2800, opened 2024-09-27
Handle TLS 1.3 client certification auth failure

andreachild:
Fix NullPointerException being thrown by outbound encoder due to null remote host which can happen with TLS 1.3 client certification auth failure. Client cert auth failure can happen after a successful handshake, causing an inbound SSLException. Changed the inbound handler to detect and store the SSLException in the channel attributes and changed the outbound encoder to detect if remote address is not available and there was an SSLException. Changed the integration test to log the stack trace of the caught exception for easier troubleshooting as these SSL tests have been flaky.

Tested the fix by running GremlinServerSslIntegrateTest and gremlin-console remote connection to gremlin-server with ssl enabled and client auth required.


xiazcy:
VOTE +1

Cole-Greer:
Thanks, it's great to know why that test has been so troublesome and to finally get it pinned down.

VOTE +1

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/HttpGremlinRequestEncoder.java
```diff
@@ -68,14 +71,15 @@ protected void encode(final ChannelHandlerContext channelHandlerContext, final R
                     requestMessage));
         }
 
+        InetSocketAddress remoteAddress = getRemoteAddress(channelHandlerContext.channel());
```
kenhuuu, line 74: Nit: add `final`. Please check over other parts that may need similar update.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/GremlinResponseHandler.java
```diff
@@ -44,6 +46,7 @@
 public class GremlinResponseHandler extends SimpleChannelInboundHandler<ResponseMessage> {
     private static final Logger logger = LoggerFactory.getLogger(GremlinResponseHandler.class);
     private static final AttributeKey<ResponseException> CAUGHT_EXCEPTION = AttributeKey.valueOf("caughtException");
+    private static final AttributeKey<Throwable> SSL_EXCEPTION = AttributeKey.valueOf("sslException");
```
kenhuuu, line 49: This is duplicated in two classes. Consider moving to `StateKey`.

kenhuuu:
VOTE +1, pending fixes to minor nits.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/HttpGremlinRequestEncoder.java
andreachild: Added a couple finals

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/GremlinResponseHandler.java
andreachild: Removed duplication by changing the AttributeKey in the GremlinResponseHandler to public (as it is the class that adds the attribute to the context) and changing HttpGremlinRequestEncoder to reuse the AttributeKey from GremlinResponseHandler.

PR #2806, opened 2024-10-01
Custom Strategy Construction

Cole-Greer:
This PR represents the second and final part of the changes to TraversalStrategies from [a recent dev list discussion](https://lists.apache.org/thread/xv235syvzgl06qcsqr0rrx8boch7ozjp).

This PR adapts `TraversalStrategyProxy` in Java to work based off of strategy names instead of strategy classes. This change allows the proxy to be used in traversals in order to pass custom/provider defined strategies and configurations into the gremlin script. The advantage of this approach is that the user does not need any concrete strategy class for custom strategies, and the only actions needed from a provider to support a custom strategy is to either register it for use with `gremlin-lang`, or add it to an import customizer for `gremlin-groovy`. The `TraversalStrategy` base class in python is similarly adapted to pass custom strategies into gremlin scripts.

My vote is VOTE +1

NOTE: This branch builds off of the work previously done in https://github.com/apache/tinkerpop/pull/2781, and thus this PR is targeting that branch to present a clean diff. Once that PR is approved and merged, this PR will be re-targeted to `master-http`.


xiazcy:
VOTE +1

kenhuuu:
VOTE +1

PR #2808, opened 2024-10-03
[TINKERPOP-3086] Upgrade Python Version to >=3.9

Cole-Greer:
https://issues.apache.org/jira/browse/TINKERPOP-3086

As python 3.8 will be reaching end of life at the end of the month, bumping requirements to min 3.9 as per our [policy on runtimes](https://tinkerpop.apache.org/docs/current/dev/developer/#runtimes).

NOTE: I've left the console test image using the bullseye (debian 11) based image. This is the same base image from `python:3.8.12`. The default python 3.9 images have switched to debian 12 (bookworm) as a base image, which has dropped openjdk-11 from it's stable repository.

My vote is VOTE +1


kenhuuu:
VOTE +1, though I'm curious why the bump is only to 3.9 since this is for the master branch. Seems like it will need to be bumped again before any release from this branch so you might want to make a follow-up Jira for that.

Cole-Greer:
> VOTE +1, though I'm curious why the bump is only to 3.9 since this is for the master branch. Seems like it will need to be bumped again before any release from this branch so you might want to make a follow-up Jira for that.

Good catch, this was intended for 3.6 and 3.7 (only permitted to update to 3.9 according to our policies). I've updated the PR to target 3.6-dev. I intend to follow up later with a python upgrade (likely to 3.10) in master-http.

PR #2823, opened 2024-10-11
Add data list back to result object in GraphSONv4.

kenhuuu:
There might be potential additions to the `result` object in the response so adding back the `data` list just in case.


Cole-Greer:
VOTE +1

vkagamlyk:
VOTE +1

kenhuuu:
VOTE +1

PR #2827, opened 2024-10-11
TINKERPOP-3088 Retain lower common type for sum() global

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-3088

The multiplication of the traverser value by the bulk was promoting the type to long so that even if the values were all int and fitting well within that space you'd still end up with a long value. Old behavior:

```
gremlin> a = g.V().values('age').sum().next()
==>123
gremlin> a.class
==>class java.lang.Long
```

new behavior after change:

```
gremlin> a = g.V().values('age').sum().next()
==>123
gremlin> a.class
==>class java.lang.Integer
```

VOTE +1


Cole-Greer:
Looks like there's a few tests in server and spark which need corresponding updates from Long to Integer. Other than that LGTM.

VOTE +1 pending test fixes.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/SumGlobalStep.java
```diff
@@ -68,7 +69,48 @@ public void processAllStarts() {
 
     @Override
     public S projectTraverser(final Traverser.Admin<S> traverser) {
-        return (S) mul(traverser.get(), traverser.bulk());
+        final S value = traverser.get();
+        final long bulk = traverser.bulk();
+        final Number projected = mul(value, bulk);
+
+        // the following behavior allows global sum() to behave the same as sum(local)
+        if (value instanceof Long || value instanceof Double ||
```
kenhuuu, line 77: Float should probably be here too? I think that Long gets promoted to Float during addition, so Float is considered the "wider" type?

xiazcy:
VOTE +1

PR #2832, opened 2024-10-15
[TINKERPOP 2700] Make WS Compression Configurable in Remaining GLVs

Cole-Greer:
https://issues.apache.org/jira/browse/TINKERPOP-2700

Use of compression with sensitive data may make users vulnerable to attacks such as CRIME/BREACH. .Net and Go already had options which allowed users to configure the use of compression. This PR is adding such configuration to the remaining drivers.

Note: There is a corresponding PR to 3.7 which adds additional tests making use of GremlinSocketServer (https://github.com/apache/tinkerpop/pull/2833)

### DIFF
--- a/gremlin-python/src/main/python/gremlin_python/driver/client.py
+++ b/gremlin-python/src/main/python/gremlin_python/driver/client.py
@@ -44,7 +44,7 @@ class Client:
                  transport_factory=None, pool_size=None, max_workers=None,
                  message_serializer=None, username="", password="",
                  kerberized_service="", headers=None, session=None,
-                 enable_user_agent_on_connect=True, use_compression=False,
+                 enable_user_agent_on_connect=True, enable_compression=False,
                  **transport_kwargs):
         log.info("Creating Client with url '%s'", url)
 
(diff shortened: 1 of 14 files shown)


gremlin-python/src/main/python/gremlin_python/driver/client.py
```diff
@@ -55,8 +56,11 @@ def __init__(self, url, traversal_source, protocol_factory=None,
         self._headers = headers
         self._enable_user_agent_on_connect = enable_user_agent_on_connect
         self._traversal_source = traversal_source
+        self._enable_compression = enable_compression
         if not self._use_http and "max_content_length" not in transport_kwargs:
             transport_kwargs["max_content_length"] = 10 * 1024 * 1024
+        if self._enable_compression and "compress" not in transport_kwargs:
+            transport_kwargs["compress"] = 15  # enable per-message deflate compression with max 32k sliding window size
```
kenhuuu, line 63: This seems oddly specific. Is it an AioHttp value? It probably shouldn't work like this since this should work for any "transport".

Cole-Greer: Yes it's an AioHttp value. I'll make a change to propagate `enable_compression` down to the transport layer and push this code into AiohttpTransport.

Cole-Greer:
VOTE +1

kenhuuu:
VOTE +1

xiazcy:
VOTE +1

PR #2833, opened 2024-10-15
[TINKERPOP 2700] Make WS Compression Configurable in Remaining GLVs 3.7

Cole-Greer:
https://issues.apache.org/jira/browse/TINKERPOP-2700

Use of compression with sensitive data may make users vulnerable to attacks such as CRIME/BREACH. .Net and Go already had options which allowed users to configure the use of compression. This PR is adding such configuration to the remaining drivers.

Note: There is a corresponding PR to 3.6 which is missing the GremlinSocketServer tests (https://github.com/apache/tinkerpop/pull/2832)

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/test/integration/client-behavior-tests.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/test/integration/client-behavior-tests.js
@@ -64,28 +64,28 @@ describe('Client', function () {
             await noUserAgentClient.close();
         });
         it('should not request permessage deflate compression by default', async function () {
-            let result = await client.submit('1', null, {requestId: settings.SEC_WEBSOCKET_EXTENSIONS});
-            let returnedExtensions = result.first()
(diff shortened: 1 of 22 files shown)


gremlin-javascript/src/main/javascript/gremlin-javascript/test/integration/client-behavior-tests.js
```diff
@@ -63,6 +63,33 @@ describe('Client', function () {
 
             await noUserAgentClient.close();
         });
+        it('should not request permessage deflate compression by default', async function () {
+            let result = await client.submit('1', null, {requestId: settings.SEC_WEBSOCKET_EXTENSIONS});
```
vkagamlyk, line 67: for immutable variables better to use [const](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const)
same below

Cole-Greer:
VOTE +1

kenhuuu:
VOTE +1

xiazcy:
VOTE +1

PR #2834, opened 2024-10-16
[TINKERPOP-3076] Throw better error if request exceeds WriteBufferSize in Go

Cole-Greer:
https://issues.apache.org/jira/browse/TINKERPOP-3076

Gorilla has an implicit frame size limit imposed by WriteBufferSize. Requests which exceed this limit cannot be parsed by gremlin server and currently lead to cryptic errors in the server. This PR simply adds a client side check to throw a more descriptive error to the user. The resulting behaviour is very similar to how `maxContentLength` functions in the java driver.

VOTE +1

### DIFF
--- a/gremlin-go/driver/gorillaTransporter.go
+++ b/gremlin-go/driver/gorillaTransporter.go
@@ -20,7 +20,6 @@ under the License.
 package gremlingo
 
 import (
-	"errors"
 	"net/http"
 	"net/url"
 	"sync"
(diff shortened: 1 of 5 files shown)


gremlin-go/driver/gorillaTransporter.go
```diff
@@ -100,6 +101,9 @@ func (transporter *gorillaTransporter) Write(data []byte) error {
 			return err
 		}
 	}
+	if len(data) > transporter.connSettings.writeBufferSize {
+		return errors.New("request too large, consider increasing 'WriteBufferSize' configuration or split into smaller requests")
```
vkagamlyk, line 105: all errors should be defined in `error_codes.go`
and used like `return newError(err0101ConnectionCloseError)`

kenhuuu:
You should add an actual English error message to `en.json`

vkagamlyk:
VOTE+1

kenhuuu:
VOTE +1

xiazcy:
VOTE +1

PR #2849, opened 2024-10-24
Added request interceptor to gremlin-python.

kenhuuu:
The message_serializer is split into the request_serializer and the response_serializer to allow for setting the request_serializer to None which prevents payload serialization. This makes it easier for the user to modify the payload. Auth is not combined with interceptor as it requires refactoring to provide the URL in protocol and simplifies usage compared to the Java driver albeit at the cost of flexibility.

VOTE+1

### DIFF
--- a/docs/src/upgrade/release-4.x.x.asciidoc
+++ b/docs/src/upgrade/release-4.x.x.asciidoc
@@ -85,7 +85,7 @@ The properties file in the above example can either point to a remote configurat
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-3017[TINKERPOP-3017]
 
-==== Changes to RequestInterceptor
+==== Changes to Java RequestInterceptor
 
 Because the underlying transport has been changed from WebSockets to HTTP, the usage of the `RequestInterceptor` has
 changed as well. The `RequestInterceptor` will now be run per request and will allow you to completely modify the HTTP
@@ -105,6 +105,19 @@ modified. After each `RequestInterceptor` is run, the request is updated with th
 that is sent to the endpoint. There is a default interceptor added to every `Cluster` called "serializer". This
 interceptor is responsible for serializing the request body is which what the server normally expects. This is intended
 to be an advanced customization technique that should only be used when needed.
+
+==== Addition of Python interceptor
+
+HTTP interceptors have been added to `gremlin-python` to enable capability similar to that of Java GLV. These
+interceptors can be passed into either a `DriverRemoteConnection` or a `Client` using the interceptors parameter. An
+interceptor is a `Callable` that accepts one argument which is the HTTP request (dictionary containing header, payload
+and auth) or a list/tuple of these functions. The interceptors will run after the request serializer has run but before
+any auth functions run so the HTTP request may still get modified after your interceptors are run. In situations where
+you don't want the payload to be serialized, the `message_serializer` has been split into a `request_serializer` and a
+`response_serializer`. Simply set the `request_serializer` to `None` and this will prevent the `RequestMessage` from
+being serialized. Again, this is expected to be an advanced feature so some knowledge of implementation details will be
+required to make this work. For example, you'll need to know what payload formats are accepted by `aiohttp` for the
+request to be sent.
 
 ==== Changes to deserialization for gremlin-javascript
 
--- a/gremlin-python/src/main/python/gremlin_python/driver/client.py
+++ b/gremlin-python/src/main/python/gremlin_python/driver/client.py
@@ -41,8 +41,10 @@ class Client:
 
     def __init__(self, url, traversal_source, protocol_factory=None,
                  transport_factory=None, pool_size=None, max_workers=None,
-                 message_serializer=None, auth=None, headers=None,
-                 enable_user_agent_on_connect=True, enable_bulked_result=False, **transport_kwargs):
+                 request_serializer=serializer.GraphBinarySerializersV4(),
+                 response_serializer=None, interceptors=None, auth=None,
+                 headers=None, enable_user_agent_on_connect=True,
+                 enable_bulked_result=False, **transport_kwargs):
         log.info("Creating Client with url '%s'", url)
 
         self._closed = False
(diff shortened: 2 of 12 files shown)


docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -106,6 +106,19 @@ that is sent to the endpoint. There is a default interceptor added to every `Clu
...
 
+==== Addition of Python interceptor
+
+HTTP interceptors have been added to `gremlin-python` to enable capability similar to that of Java GLV. These
+interceptors can be passed into either a `DriverRemoteConnection` or a `Client` using the interceptors parameter. An
+interceptor is a `Callable` that accepts one argument which is the HTTP request (dictionary containing header, payload
+and auth) or a list/tuple of these functions. The interceptors will run after the request serializer has run but before
+any auth functions run so the HTTP request may still get modified after your interceptors are run. In situations where
+you don't want the payload to be serialized, the `message_serializer` has been split into a `request_serializer` and a
+`response_serializer`. Simply set the `request_serializer` to `None` and this will prevent the `RequestMessage` from
```
andreachild, line 117: What happens if they set they mix up the request and response serializers ie graphSON vs graph binary? Is that supported?

gremlin-python/src/main/python/gremlin_python/driver/protocol.py
```diff
@@ -67,15 +77,22 @@ def connection_made(self, transport):
...
-            'payload': self._message_serializer.serialize_message(request_message),
+            'headers': {'accept': accept},
+            'payload': self._request_serializer.serialize_message(request_message)
+                if self._request_serializer is not None else request_message,
             'auth': self._auth
         }
 
+        # The user may not want the payload to be serialized if they are using an interceptor.
+        if self._request_serializer is not None:
+            content_type = str(self._request_serializer.version, encoding='utf-8')
```
andreachild, line 90: What should the content-type be if there is no serialization?

gremlin-python/src/main/python/tests/driver/test_driver_remote_connection.py
```diff
@@ -40,8 +40,7 @@ class TestDriverRemoteConnection(object):
     # this is a temporary test for basic graphSONV4 connectivity, once all types are implemented, enable graphSON testing
     # in conftest.py and remove this
     def test_graphSONV4_temp(self):
-        remote_conn = DriverRemoteConnection(test_no_auth_url, 'gmodern',
-                                             message_serializer=serializer.GraphSONSerializerV4())
+        remote_conn = DriverRemoteConnection(test_no_auth_url, 'gmodern')
```
andreachild, line 43: Does removing the serializer param change the test serializer from graphSON to graph binary since graph binary is the default?

docs/src/upgrade/release-4.x.x.asciidoc
kenhuuu: This is supported. These effectively set the `content-type` and `accept` headers which TinkerPop 4 HTTP servers should handle.

gremlin-python/src/main/python/gremlin_python/driver/client.py
```diff
@@ -53,11 +55,11 @@ def __init__(self, url, traversal_source, protocol_factory=None,
         self._traversal_source = traversal_source
         if "max_content_length" not in transport_kwargs:
             transport_kwargs["max_content_length"] = 10 * 1024 * 1024
-        if message_serializer is None:
-            message_serializer = serializer.GraphBinarySerializersV4()
+        if response_serializer is None:
+            response_serializer = serializer.GraphBinarySerializersV4()
```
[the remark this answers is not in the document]
kenhuuu: No, this value must always be set as the response must be deserialized. The server always serializes the response. If the value were defaulted the user could still accidentally set the value to `None`.

gremlin-python/src/main/python/tests/driver/test_driver_remote_connection.py
kenhuuu: Thanks for catching this, I'll update this to set both the request and response serializers to GraphSON.

gremlin-python/src/main/python/gremlin_python/driver/protocol.py
andreachild: Should there be a check after calling the interceptors that content-type is set?

gremlin-python/src/main/python/tests/driver/test_driver_remote_connection.py
andreachild: I wonder why the test didn't fail when the serializer was changed to graph binary?

gremlin-python/src/main/python/gremlin_python/driver/protocol.py
kenhuuu: I think that is reasonable. I opted not to do it because a lot of this depends on the underlying HTTP client. In gremlin-python, you could technically use whatever client you want since the protocol and transport are abstracted. There are some clients that would probably add that value in for you (I believe aiohttp has a default as well). This is why I noted in the upgrade documentation that this is an advanced feature. The user needs to know the capabilities of the underlying client and test it out themselves. This is the same reason we don't check for the type of the payload.

gremlin-python/src/main/python/tests/driver/test_driver_remote_connection.py
kenhuuu: Those queries work for either graphson or graphbinary so there wouldn't be a failure, just a gap in testing.

andreachild: Makes sense and I don't see an easy way for the test to verify that the given serializer was used.

andreachild:
LGTM VOTE +1

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -106,6 +106,19 @@ that is sent to the endpoint. There is a default interceptor added to every `Clu
...
+being serialized. Again, this is expected to be an advanced feature so there is some knowledge of implementation details
+will be required to make this work. For example, you'll need to know what payload formats are accepted by `aiohttp` for
```
Cole-Greer, line 119:
```suggestion
being serialized. Again, this is expected to be an advanced feature so some knowledge of implementation details
will be required to make this work. For example, you'll need to know what payload formats are accepted by `aiohttp` for
```

Cole-Greer (approved):
LGTM, just a minor docs nit.

VOTE +1

xiazcy:
VOTE +1

PR #2855, opened 2024-10-28
Add log entry in WsAndHttpChannelizerHandler

dh-cloud:
The log entry here is necessary because when an exception thrown by
Netty reaches exceptionCaught(), if the error is not printed, the
server side will not be able to know what has happened.

For example, if the maxContentLength in JanusGraph's Gremlin-Server
configuration is set to 65536, and the client sends data over this
size via WebSocket, the JanusGraph server logs will appear normal,
but the client will experience a disconnection, making it difficult
to immediately identify the cause of the issue.

### DIFF
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/WsAndHttpChannelizerHandler.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/WsAndHttpChannelizerHandler.java
@@ -18,6 +18,9 @@
  */
 package org.apache.tinkerpop.gremlin.server.handler;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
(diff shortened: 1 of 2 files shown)


gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/WsAndHttpChannelizerHandler.java
```diff
@@ -100,6 +105,12 @@ public void channelRead(final ChannelHandlerContext ctx, final Object obj) {
 
     @Override
     public void exceptionCaught(final ChannelHandlerContext ctx, final Throwable cause) {
+        // The log printing here is necessary because when an exception thrown by Netty reaches this method,
+        // if the error is not printed, the server side will not be able to perceive what has happened.
+        // For example, if the maxContentLength in JanusGraph's Gremlin-Server configuration is set to 65536,
+        // and the client sends data over this size via WebSocket, the JanusGraph server logs will appear normal,
+        // but the client will experience a disconnection, making it difficult to immediately identify the cause of the issue.
+        logger.error("Error processing request from netty", cause);
```
kenhuuu, line 113:
```suggestion
        logger.error("Could not process the request", cause);
```
Nit: the error should be slightly more generic to match what is done in `gremlin-driver`

kenhuuu:
Thanks, I just have one small nit regarding the logged message.

VOTE +1, pending change.

xiazcy:
VOTE +1

Cole-Greer:
VOTE +1

PR #2862, opened 2024-10-28
Remove Traverser serializers for GraphSON/GraphBinary V4

kenhuuu:
Traverser isn't used anymore as the server will put the "bulked" results into a list for GraphBinary. GraphSON won't support bulking or returning traversers at all and instead will return expanded results.

VOTE +1


vkagamlyk:
VOTE+1

Cole-Greer:
VOTE +1

PR #2875, opened 2024-10-31
Update BigDecimal in gremlin-python statics to be consistent with server definition

xiazcy:
Added a value property to BigDecimal so it can be properly serialized in Gremlin Lang to the server for GraphBinary. 

Added a `to_bigdecimal` function to help creating BigDecimal in python and for GraphSON. 

Currently targeting the test branch, will rebased once that PR is merged.

### DIFF
--- a/gremlin-python/src/main/python/gremlin_python/statics.py
+++ b/gremlin-python/src/main/python/gremlin_python/statics.py
@@ -16,7 +16,7 @@
 # specific language governing permissions and limitations
 # under the License.
 #
-
+import decimal
 from types import FunctionType
 from aenum import Enum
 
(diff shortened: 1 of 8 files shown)


Cole-Greer:
VOTE +1

xiazcy:
VOTE +1

gremlin-python/src/main/python/gremlin_python/statics.py
```diff
@@ -74,12 +75,35 @@ def __init__(self, gremlin_type):
...
 
+    @property
+    def value(self):
+        self._as_decimal = Decimal(self.unscaled_value)
+        precision = len(self._as_decimal.as_tuple().digits)
+        decimal.getcontext().prec = precision
+        return self._as_decimal.scaleb(-self.scale)
+
+"""
+Create a BigDecimal from a number that can be converted to a Decimal. Note precision may be lost in the convertion.
```
kenhuuu, line 94:
```suggestion
Create a BigDecimal from a number that can be converted to a Decimal. Note precision may be lost during the conversion.
```
Nit

gremlin-python/src/main/python/gremlin_python/statics.py
```diff
@@ -74,12 +75,35 @@ def __init__(self, gremlin_type):
...
+    """
     def __init__(self, scale, unscaled_value):
         self.scale = scale
         self.unscaled_value = unscaled_value
 
+    @property
+    def value(self):
+        self._as_decimal = Decimal(self.unscaled_value)
+        precision = len(self._as_decimal.as_tuple().digits)
+        decimal.getcontext().prec = precision
```
kenhuuu, line 90: Does this cause the precision on a global level?

xiazcy: Actually it does, I'll update to either reset or use local context.

xiazcy: Updated with using a local context.

kenhuuu:
VOTE +1

PR #2883, opened 2024-11-04
Update SigV4 interceptor to use latest aws sdk version

andreachild:
Updated SigV4 interceptor to use the latest version of `software.amazon.awssdk` and removed old dependency on `aws-java-sdk-core`. Since latest version of `SignedRequest` has headers that support multiple values for the same header via Map<String,List> I added validation checks that the expected signed headers only have 1 value.


andreachild:
Closing in favour of https://github.com/apache/tinkerpop/pull/2884

PR #2888, opened 2024-11-08
Changed structure feature test parameter name

DKZed:
Removed the second variable in the parameter name since it made it so that the tests couldn't be opted out of.

For example, they would require one specify `specific = supportsIntegerArrayValues([I@1d2b8249)`.


lyndonbauto:
VOTE+1

vkagamlyk:
VOTE+1

PR #2894, opened 2024-11-09
Fixes TINKERPOP-3120

unautre:
Hello,
This PR fixes TINKERPOP-3120 that I just created.

Session support seems to have been dropped from the main branch, so this is for 3.7-dev only.

Best regards,


kenhuuu:
Hi, thanks for contributing and providing a detailed explanation of the problem in the Jira.

Could you add a test to `GremlinServerAuthIntegrateTest` for this case?

kenhuuu:
Hi again, just checking in to see if you are able to add that test.

unautre:
> Hi again, just checking in to see if you are able to add that test.

Just did ! Sorry it took so long.
The exception is only logged and not thrown - so I have to resort using the `LogCaptor`  to make the test work.

kenhuuu:
Thanks for including the test.

VOTE +1

Cole-Greer:
Thanks for reporting and fixing this bug.

VOTE +1

PR #2906, opened 2024-11-18
TINKERPOP-2959 Enums can't be variables in grammar

spmallette:
Enums dont quite make sense as variables as you likely wouldn't do change something like Scope dynamically as it alters the whole intent of the traversal. In addition, technically enums can't be parameters because serialization doesn't support them - they are not supported types (T/Direction are the only exceptions but only because they are returned in query results from valueMap()/elementMap()).

VOTE +1


Cole-Greer:
VOTE +1

xiazcy:
VOTE +1

PR #2935, opened 2024-12-02
Add unexpected-response handler to gremlin-javascript connection websocket

KyleBoyer:
This PR helps provide more information to the end user when a websocket `unexpected-response` is encountered. This is particularly helpful for using AWS Neptune, which sometimes encounters an "unexpected-response" status code, with a body describing the issue.


kenhuuu:
Hi and thank you for contributing.

The `master` branch which is for 4.0 won't support WebSockets so your changes will need to target 3.7-dev instead. The code you are modifying now will be replaced with code that supports HTTP.

Have you noticed the same `unexpected-response` coming back from Gremlin Server?

KyleBoyer:
Sounds good, I'll open a new PR to target the correct branch and close this PR. We have noticed unexpected responses for 3 different situations on AWS Neptune so far. One was a random 400 error status code bug that AWS fixed. Another situation with AWS Neptune was using invalid AWS credentials. Third one we noticed recently is a throttling 500 status code. These situations were difficult to debug before we started reading the body when the web socket hit these unexpected responses.

KyleBoyer:
New PR targeting the correct base branch opened here: https://github.com/apache/tinkerpop/pull/2936

PR #2936, opened 2024-12-03
Add unexpected-response handler to gremlin-javascript connection websocket

KyleBoyer:
This PR helps provide more information to the end user when a websocket unexpected-response is encountered. This is particularly helpful for using AWS Neptune, which sometimes encounters an "unexpected-response" status code, with a body describing the issue.

This is the same PR as https://github.com/apache/tinkerpop/pull/2935 just targeting the correct base branch.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Add log entry in `WsAndHttpChannelizerHandler`.
 * Changed `IdentityRemovalStrategy` to omit `IdentityStep` if only with `RepeatEndStep` under `RepeatStep`.
 * Changed Gremlin grammar to make use of `g` to spawn child traversals a syntax error.
+* Added `unexpected-response` handler to `ws` for `gremlin-javascript`
 
 [[release-3-7-3]]
 === TinkerPop 3.7.3 (October 23, 2024)
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
@@ -260,7 +260,8 @@ class Connection extends EventEmitter {
       });
       res.on('error', reject);
     });
-    const errorMessage = `Unexpected server response code ${res.statusCode} with body:\n${body.toString()}`;
+    const statusCodeErrorMessage = `Unexpected server response code ${res.statusCode}`;
+    const errorMessage = body ? `${statusCodeErrorMessage} with body:\n${body.toString()}` : statusCodeErrorMessage;
     const error = new Error(errorMessage);
     this.#handleError({
       error,
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
@@ -256,7 +256,7 @@ class Connection extends EventEmitter {
         chunks.push(data instanceof Buffer ? data : Buffer.from(data));
       });
       res.on('end', () => {
-        resolve(Buffer.concat(chunks));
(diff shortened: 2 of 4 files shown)


kenhuuu:
Thanks again for contributing.

I think I'm OK with the implementation you added but could you add an integration test as well? You can probably trigger a 401 that should be handled by your new function in a test in ./gremlin-javascript/src/main/javascript/gremlin-javascript/test/integration/sasl-authentication-tests.js

KyleBoyer:
@kenhuuu I have updated the PR to include a test for this unexpected-response handler. Please let me know if this is the correct way to implement this test. Thank you!

KyleBoyer:
Happy new year! Any updates about/for this PR?

Thank you!

kenhuuu:
Happy new year. Sorry for the delay in following up on this. I'll take a look at this today.

kenhuuu:
The test isn't quite what I had in mind. Does your handler only trigger for 500 responses? I thought maybe it could be triggered with a 401 or 403 which is why I was suggesting that the test be placed there. If this isn't the case, and it requires a 500, then it should probably be moved to a different test file as it starts its own server rather than relying on the Gremlin Server that runs for integration tests.

KyleBoyer:
Understood! It was a few weeks ago and I remember trying to figure out how to get an immediate 4XX status code response from gremlin with a body, but I was unable to invoke that situation at the time. I’ll give it another go, and if it doesn’t work out using the integration test Gremlin Server, I’ll break it out into its own test file.Do you have any ideas how to immediately get the Gremlin Server to respond with a 4XX or 5XX status code?Thank you!On Jan 6, 2025, at 5:03 PM, kenhuuu ***@***.***> wrote:﻿
The test isn't quite what I had in mind. Does your handler only trigger for 500 responses? I thought maybe it could be triggered with a 401 or 403 which is why I was suggesting that the test be placed there. If this isn't the case, and it requires a 500, then it should probably be moved to a different test file as it starts its own server rather than relying on the Gremlin Server that runs for integration tests.

Cole-Greer:
Hi @KyleBoyer, one easy way to force a 401 response is to adapt a simple test such as [this one](https://github.com/apache/tinkerpop/blob/169f4a26bfa63752f1c1a425e14beca5c40b226e/gremlin-javascript/src/main/javascript/gremlin-javascript/test/integration/sasl-authentication-tests.js#L46-L55) and just replace the `'password'` in the first line with some invalid junk.

You can also force a 599 status code by putting some junk in the bytecode you are submitting such as 
```
connection.submit(new Bytecode().addStep('notARealStep', []).addStep('tail', []))
```

KyleBoyer:
Thanks for the suggestions - I did try those methods and the Connection.js -> handleMessage is called with the "websocket message", albeit, the error, but still considered a valid websocket response according to the `ws` package. I believe the case I'm trying to cover is invalid responses that are not part of the websocket upgraded connection. I have tried a variety of methods to invoke the unexpected-response handler, but all the following methods all get returned a "valid" websocket message:

- Invalid credentials for sasl authenticator
- Invalid bytecode steps
- Incorrect path `/gremlin` -> `/something-unexistant` (causes a socket hang up error, so no "unexpected response")

The way we originally encountered this error was by utilizing AWS Neptune, providing invalid credentials, the websocket connection would not be upgraded, and instead we would immediately get a standard 401 reply.

Therefore, I think I will need to break this test out into its own file.

KyleBoyer:
@kenhuuu / @Cole-Greer 

Alright, I have removed the changes to the `sasl-authentication-tests.js` and introduced a new `socket-connection-tests.js`. I have included two tests in this new file, to test for an invalid response with no body, and with a body.

Please let me know how this is looking now, and if any other tweaks should be included.

Thanks again!

kenhuuu:
I think your latest changes are good enough.

VOTE +1

Cole-Greer:
Looks good to me as well, thanks for the contribution!

VOTE +1

xiazcy:
VOTE +1

PR #2966, opened 2025-01-06
Promote gremlin-lang as default over gremlin-groovy

Cole-Greer:
As previously discussed in the [devlist](https://lists.apache.org/thread/qh8h810s4tn9d6xltc4v0zp9n8fwt4qt), TinkerPop 4 should default to using gremlin-lang over gremlin-groovy in cases where users do not explicitly specify a language for a given script. This PR switches the default engine and updates any integration tests which were submitting incompatible scripts. For the most part I tried to replace with some an equivalent gremlin-lang traversal (eg `1+1` -> `g.inject(2)`). Some tests have been left with groovy as the behaviour cannot easily be replicated in gremlin-lang (mostly tests which use `Thread.sleep()` or use groovy to generate unserializable responses on the server)

My vote is VOTE +1


gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
```diff
@@ -306,11 +310,29 @@ public void shouldProcessTraversalInterruption() {
...
+            final ResponseException re = (ResponseException) ex.getCause();
+            assertEquals(HttpResponseStatus.INTERNAL_SERVER_ERROR, re.getResponseStatusCode());
+            assertEquals("ServerTimeoutExceededException", re.getRemoteException());
+        } finally {
+            cluster.close();
+        }
+    }
+
+    @Test
+    public void shouldProcessTraversalInterruptionGroovy() {
```
andreachild, line 325: What is the difference between this test and `shouldProcessTraversalInterruption()`?

Cole-Greer: Thanks for catching that, this was a remnant of a partially-reverted experiment to replace the groovy sleep-based tests with slow gremlin-lang queries. I opted to only use the groovy sleep variants as the slow queries proved unreliable. I will clean this up and remove the duplicate.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
```diff
@@ -898,15 +891,17 @@ public void shouldExecuteSessionlessScriptOnTransactionalGraph() throws Exceptio
         final Client client = cluster.connect();
 
         // this line is important because it tests GraphTraversal which has a certain transactional path
-        final Vertex vertexRequest1 = client.submit("g.addV().property(\"name\",\"stephen\")").all().get().get(0).getVertex();
+        final Vertex vertexRequest1 = client.submit("g.addV().property(\"name\",\"stephen\")", groovyRequestOptions)
```
andreachild, line 894: Curious why `groovyRequestOptions` is required here?

Cole-Greer: Groovy isn't required for this script, but it is required for both `graph.vertices().next()` and `graph.addVertex(\"name\",\"marko\")`. I opted to use gremlin-groovy for the entire test, although it should work the exact same if the first and last scripts were submitted as gremlin-lang instead.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Context.java
```diff
@@ -197,7 +197,7 @@ public void handleDetachment(final List<Object> aggregate) {
 
             if (firstElement instanceof Element) {
                 for (int i = 0; i < aggregate.size(); i++)
-                    aggregate.set(i, ReferenceFactory.detach((Element) aggregate.get(i)));
```
andreachild, line 200: Was this changed because it was unnecessary casting? Or did switch to gremlin-lang cause an error with this code?

gremlin-python/src/main/python/tests/driver/test_driver_remote_connection.py
```diff
@@ -116,12 +116,12 @@ def test_extract_request_options_with_params(self, remote_connection):
         g = traversal().with_(remote_connection)
         t = g.with_("evaluationTimeout",
                     1000).with_("batchSize", 100).with_("userAgent",
-                                                        "test").V(GValue('ids', [1, 2, 3])).count()
```
andreachild, line 119: To decrease the chance of forgetting to investigate why this fails with gremlin-lang can we use `@unittest.skip("investigate why 'ids' fails to resolve")` for this test and create a new one which uses `x` instead?

xiazcy:
VOTE +1

andreachild:
VOTE +1 pending minor nits

kenhuuu:
Would be nice to have a public final String constant for "gremlin-groovy" and "gremlin-lang" somewhere in core so that isn't in so many files.

VOTE +1

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -402,6 +402,12 @@ g.V().has_label(GValue('label', 'person'))
 Use of `GValue` in traversals with repeated patterns may lead to improved performance in certain graphs. Consult the
 documentation for your specific graph provider for recommendations on how to best utilize `GValue` in traversals.
 
+==== Change Default Script Language in Gremlin-Server
```
spmallette, line 405: the `language` field is one way, but seems worth mentioning that they could also set it from the server-side as well by switching the configuration back to gremlin-groovy in the yaml. that's another possible way, right?

spmallette:
> This PR switches the default engine
if this is a switch of the default engine, then why are all the yaml files not modified? https://github.com/apache/tinkerpop/tree/master/gremlin-server/conf

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Context.java
Cole-Greer: I did switch it in response to an error, but it's only indirectly related switching to gremlin-lang. This is the code which converts elements to ReferenceElements (no properties) when `with('materializeProperties', 'tokens')` is used. The bug was that this code failed for any traversals with bulking enabled. Iterating a bulked traversal produces results such as `[V[1] //value, 2 //bulk, V[2] //value, 1 //bulk]`. This caused a casting exception as this code would try to cast the Long's (bulk values) into Elements. Simply removing the cast does not create an issue here as ReferenceFactory.detach() is overloaded for Object and non-detacheable objects are simply returned as is.

I believe there may still be more potential bugs in this code if a traversal produces a mix of Element and non-Element results, but I need to investigate that a bit further. If I find more bugs in this code I will either CTR fixes or open new JIRAs.

I'm considering bugs in this code to be out of scope of this PR anyways as the bugs are pre-existing, and we just happened to find them while converting some tests to gremlin-lang. I included this fix here as it was simple and quick and needed to get some tests to pass.

Cole-Greer:
@spmallette 
> > This PR switches the default engine
> 
> if this is a switch of the default engine, then why are all the yaml files not modified? https://github.com/apache/tinkerpop/tree/master/gremlin-server/conf

I've gone ahead and added gremlin-lang without any special configuration in all of the yaml files. It's arguably redundant as gremlin-lang is configured by default in the server `Settings` class in this branch (in place of groovy), however it may be more clear for users to have it explicitly listed in the yaml files.

I left the groovy configuration in all of the yaml files as it's still needed setup `g` and load data.

> i think there's a lot more documentation changes required as well - is this more of a partial switch in preparation for more work?

My goal with this PR is not necessarily about getting rid of groovy as much as it is about making gremlin-lang more prominent. The main driver here is that with the removal of bytecode, the GLVs are generating and sending gremlin-lang scripts. The scripts should be mostly mutually compatible with lang and groovy, however there are potential incompatibilities there and it should not be recommended to execute GraphTraversal generated scripts as groovy.

There is definitely some outstanding future work related to this such as dropping the dependence on groovy to initialize the server This PR captures what I view as the critical piece for the first milestone release which is to prevent users from inadvertently running GraphTraversal generated scripts as groovy.

docs/src/upgrade/release-4.x.x.asciidoc
Cole-Greer: As it's currently implemented, if a request comes in with no specified language, the server assumes a language of `gremlin-lang` (see changes to [HttpGremlinEndpointHandler](https://github.com/apache/tinkerpop/pull/2966/files#diff-ddd50660d38c3c8ba46087caed5ee89f059a3f760e951081e75d6f5864616208) and [HttpRequestMessageDecoder](https://github.com/apache/tinkerpop/pull/2966/files#diff-ee912e4dd22cabd2776846b569646a94ebfbd99aa073eed2360d53ddef046724)).

It's currently not smart enough to switch this based on available script engines in the server.

I will reconsider the messaging in the upgrade docs and see if I can include some more nuance in it.

spmallette: since you have both configured explicitly now, i guess its more clear that gremlin-groovy is still there. i think the context is more clear now as you have it written.

minor thing, please consider more concise section title as longer ones never look nice in the toc navigation on the left. mabye "Gremlin Server Default Language" would be better?

vkagamlyk:
VOTE +1

PR #2969, opened 2025-01-10
Use .on and .off for unexpected-response listener

KyleBoyer:
The `unexpected-response` listener needs to use `.on` and `.off` because the `ws` package's `addEventListener` only accepts certain event types
Ref: https://github.com/websockets/ws/blob/8.16.0/lib/event-target.js#L202-L241

Additionally PR #2936 had slightly different logic for the body chunk return, so I have updated that here too.


kenhuuu:
Thanks again for contributing and for catching these issues from the merge.

VOTE +1

KyleBoyer:
@kenhuuu is this good to merge? 😁

PR #2982, opened 2025-01-16
Update gremlin console to display version 4

flora-jin:
Updated gremlin console ASCII art to show version 4 instead of 3.

Screenshot 2025-01-16 at 11 23 33 AM

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/Gremlin.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/Gremlin.java
@@ -37,6 +37,11 @@ public final class Gremlin {
         return gremlinVersion;
     }
 
+    /**
+     * Gets the current major version of tinkerpop.
+     */
+    public static String majorVersion() { return version().split("\\.")[0]; }
+
     public static void main(final String[] arguments) throws IOException {
         System.out.println("gremlin " + version());
     }
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GremlinServer.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GremlinServer.java
@@ -390,7 +390,7 @@ public class GremlinServer {
         builder.append(Gremlin.version() + "\r\n");
         builder.append("         \\,,,/\r\n");
         builder.append("         (o o)\r\n");
-        builder.append("-----oOOo-(3)-oOOo-----\r\n");
+        builder.append("-----oOOo-(" + Gremlin.majorVersion() + ")-oOOo-----\r\n");
         return builder.toString();
     }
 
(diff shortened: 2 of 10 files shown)


spmallette:
thanks - crucial! there is similar need in Gremlin Server

andreachild:
Hi @flora-jin I found some more places where the ascii art version needs to be changed in the documentation:

Screenshot 2025-01-16 at 11 38 11 AM

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GremlinServer.java
```diff
@@ -390,7 +390,7 @@ public static String getHeader() {
         builder.append(Gremlin.version() + "\r\n");
         builder.append("         \\,,,/\r\n");
         builder.append("         (o o)\r\n");
-        builder.append("-----oOOo-(3)-oOOo-----\r\n");
```
andreachild, line 393: We could make this dynamic by using `Gremlin.version().split("\\.")[0]` so that we don't have to keep changing it when the version changes.

andreachild: You could create a new method in `org.apache.tinkerpop.gremlin.util.Gremlin` like

```
public static String majorVersion() {
        return version().split("\\.")[0];
    }
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/Gremlin.java
```diff
@@ -37,6 +37,8 @@ public static String version() {
         return gremlinVersion;
     }
 
+    public static String majorVersion() { return version().split("\\.")[0]; }
```
andreachild, line 40: Nit: Would be nice to have javadoc for this new utility method.

andreachild:
VOTE +1

andreachild:
For the failing `ImportGremlinPluginTest` it could be changed as such:

```
@Test
    public void shouldImportWildcardMethod() throws Exception {
        final Method version = Gremlin.class.getMethod("version");
        final Method majorVersion = Gremlin.class.getMethod("majorVersion");
        final ImportGremlinPlugin module = ImportGremlinPlugin.build()
                .methodImports(Collections.singletonList(Gremlin.class.getCanonicalName() + "#*")).create();

        final DefaultImportCustomizer customizer = (DefaultImportCustomizer) module.getCustomizers().get()[0];
        assertEquals(1, module.getCustomizers().get().length);
        assertThat(customizer.getMethodImports(), hasItems(version, majorVersion));

        // will also have the static main() method
        assertEquals(3, customizer.getMethodImports().size());
    }
```

Cole-Greer:
Thanks @flora-jin, LGTM.
VOTE +1 (I will merge this as a CTR)

PR #2983, opened 2025-01-16
Change server response to 400 for malformed queries

Cole-Greer:
Alters the servers response to bad queries from `500 Internal Server Error` to `400 Bad Request`. Note that malformed groovy scripts may still produce `500 Internal Server Error` as groovy is unable to statically verify correct gremlin syntax.

VOTE +1

### DIFF
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpGremlinEndpointHandler.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpGremlinEndpointHandler.java
@@ -35,6 +35,7 @@ import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.tinkerpop.gremlin.groovy.engine.GremlinExecutor;
 import org.apache.tinkerpop.gremlin.groovy.jsr223.TimedInterruptTimeoutException;
 import org.apache.tinkerpop.gremlin.jsr223.GremlinScriptEngine;
+import org.apache.tinkerpop.gremlin.language.grammar.GremlinParserException;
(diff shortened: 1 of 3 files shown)


gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpGremlinEndpointHandler.java
```diff
@@ -295,6 +296,9 @@ private GremlinError formErrorResponseMessage(Throwable t, RequestMessage reques
             logger.warn(error.getMessage());
             return error;
         }
+        if (t instanceof GremlinParserException) {
+            return GremlinError.parsing(requestMessage.getGremlin());
```
kenhuuu, line 300: GremlinParserException has a useful message that probably should get included in the response message.

Cole-Greer: I've converted this to simply preserve the original exception message.

kenhuuu:
VOTE +1

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/util/GremlinError.java
```diff
@@ -86,6 +87,11 @@ public static GremlinError binding(final String aliased) {
         return new GremlinError(HttpResponseStatus.BAD_REQUEST, message, "InvalidRequestException");
     }
 
+    public static GremlinError parsing(final GremlinParserException error) {
+        final String message = String.format(error.getMessage());
```
kenhuuu, line 91: Nit: this line probably not needed.

PR #2992, opened 2025-01-25
Fix Bug in TinkerTransactionGraph transaction used counts

Cole-Greer:
This is based on a [reported issue in discord](https://discord.com/channels/838910279550238720/1260355937604997240/1332331746418491475) where users were observing a `TransactionException` due to concurrent modification in an unexpected case. The test added in this case is a simplified reproducer for the reported issue.

The root cause of the issue is that a read only traversal will increment an elements [usedInTransactions](https://github.com/apache/tinkerpop/blob/077e3eb22c463545c3fc974944d4bf1d5d6a32fd/tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerElementContainer.java#L65) count as the read operation must call [getWithClone()](https://github.com/apache/tinkerpop/blob/077e3eb22c463545c3fc974944d4bf1d5d6a32fd/tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerElementContainer.java#L90), however in 3.7.3, the `usedInTransactions` counter is never decremented for read-only elements when the transaction is committed.

In the example used in the test in this PR, vertex 1 will have an incorrect `usedInTransactions` of 1 after the read-only transaction has been committed. This causes the subsequent "delete" query not to remove the vertex from the graph as it is still "in use". This then leads to a TransactionException due to concurrent modification when the final query attempts to create a new vertex with the same ID.

### DIFF
--- a/tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransaction.java
+++ b/tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransaction.java
@@ -203,7 +203,7 @@ final class TinkerTransaction extends AbstractThreadLocalTransaction {
 
             final Set<TinkerElementContainer> readElements = txReadElements.get();
             if (readElements != null)
-                readElements.stream().forEach(e -> e.reset());
+                readElements.stream().forEach(e -> e.commit(txVersion));
 
             txChangedVertices.remove();
             txChangedEdges.remove();
--- a/tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerElementContainer.java
+++ b/tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerElementContainer.java
@@ -197,7 +197,7 @@ final class TinkerElementContainer<T extends TinkerElement> {
                 element.removed = true;
             element = null;
             isDeleted = true;
-        } else {
+        } else if (isModifiedInTx.get()){
             element = transactionUpdatedValue.get();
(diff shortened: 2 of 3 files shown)


tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransaction.java
```diff
@@ -203,7 +203,7 @@ protected void doCommit() throws TransactionException {
 
             final Set<TinkerElementContainer> readElements = txReadElements.get();
             if (readElements != null)
-                readElements.stream().forEach(e -> e.reset());
+                readElements.stream().forEach(e -> e.commit(txVersion));
```
vkagamlyk, line 206: Thanks Cole for figuring this out! 

I think it would be good to add a test to verify case when element is in 
both `changedVertices` and `readElements`, that `usedInTransactions` counter will be reduced exactly 2 times

Cole-Greer: Thanks for the suggestion, I've added such a test

Cole-Greer:
VOTE +1

vkagamlyk:
Is these fix for 3.7.x or 4.x?

anyway VOTE +1

spmallette:
seems like its a fix for 3.7-dev, no? and changelog needed...otherwise VOTE +1

Cole-Greer:
Thanks, yes this was intended for 3.7. I've fixed the target branch and added a changelog entry. Will merge once actions pass.

danielcweber:
I'd like to propose a 3.7.4 release containing this fix.

PR #2998, opened 2025-01-30
Console plugins file can't accept empty lines

flora-jin:
As reported in [3.7.0 gremlin-groovy console with multi-line statements](https://groups.google.com/g/gremlin-users/c/U2gIxmRcrI8/m/WzX3eyhTAAAJ)

An empty line in the plugins.txt file will cause a NPE to occur.

Fixed by adding if else statement to check for null pluggedIn variable.
If null pluggedIn variable detected, will output invalid plugin on console.

### DIFF
--- a/gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/Console.groovy
+++ b/gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/Console.groovy
@@ -143,10 +143,16 @@ class Console {
         // if there are active plugins then initialize them in the order that they are listed
         activePlugins.each { pluginName ->
             def pluggedIn = mediator.availablePlugins[pluginName]
-            pluggedIn.activate()
 
-            if (!io.quiet)
-                io.out.println(Colorizer.render(Preferences.infoColor, "plugin activated: " + pluggedIn.getPlugin().getName()))
+            if (pluggedIn != null) {
+                pluggedIn.activate()
+
+                if (!io.quiet)
+                    io.out.println(Colorizer.render(Preferences.infoColor, "plugin activated: " + pluggedIn.getPlugin().getName()))
+            } else {
+                if (!io.quiet)
+                    io.out.println(Colorizer.render(Preferences.infoColor, "invalid plugin: " + pluginName))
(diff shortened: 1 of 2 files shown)


gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/Console.groovy
```diff
@@ -143,10 +143,16 @@ class Console {
         // if there are active plugins then initialize them in the order that they are listed
         activePlugins.each { pluginName ->
             def pluggedIn = mediator.availablePlugins[pluginName]
-            pluggedIn.activate()
 
-            if (!io.quiet)
-                io.out.println(Colorizer.render(Preferences.infoColor, "plugin activated: " + pluggedIn.getPlugin().getName()))
+            if (pluggedIn != null) {
```
andreachild, line 147: An alternative solution could be to change the `Mediator.groovy.readPluginState()` method to filter out the blank lines, which would be preferable IMO because of the separation of concerns of the code. The `Mediator` is responsible for reading the file and this code is responsible for loading plugins. The code which is reading the file should be responsible for skipping the blank lines.

gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/Console.groovy
```diff
@@ -143,10 +143,16 @@ class Console {
...
+                pluggedIn.activate()
+
+                if (!io.quiet)
+                    io.out.println(Colorizer.render(Preferences.infoColor, "plugin activated: " + pluggedIn.getPlugin().getName()))
+            } else {
+                if (!io.quiet)
+                    io.out.println(Colorizer.render(Preferences.infoColor, "invalid plugin: " + pluginName))
```
andreachild, line 154: Nit: I don't feel that the console needs to call out that there are blank lines in the plugin file, just skip them silently.

flora-jin: Yes Cole mentioned yesterday and will fix !

PR #3012, opened 2025-02-03
TINKERPOP-3070 Cannot run console if working directory contains spaces

flora-jin:
https://issues.apache.org/jira/browse/TINKERPOP-3070
If the current working directory contains spaces, gremlin.sh does not handle it properly when launching the console, and the JVM will search for an incorrect main class and fail.

For example if I simply rename the standard console distribution directory to `foo bar`:

foo bar/ $ bin/gremlin.sh                                                                                                                                                                                                                                            
Error: Could not find or load main class bar
Caused by: java.lang.ClassNotFoundException: bar

Added quotations around directory path in order to ensure spaces in directory name are not considered two separate strings

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -31,7 +31,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Changed Gremlin grammar to make use of `g` to spawn child traversals a syntax error.
 * Added `unexpected-response` handler to `ws` for `gremlin-javascript`
 * Fixed bug in `TinkerTransactionGraph` where a read-only transaction may leave elements trapped in a "zombie transaction".
-* Added quotations in 'gremlin.sh' to accept directories with spaces in the name
+* Fixed bug in `gremlin.sh` where it couldn't accept a directory name containing spaces.
 
 [[release-3-7-3]]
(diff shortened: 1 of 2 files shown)


kenhuuu:
VOTE +1, pending CHANGELOG entry.

Cole-Greer:
Thanks for tracking down that fix @flora-jin, VOTE +1 (pending CHANGELOG entry)

andreachild:
VOTE +1

CHANGELOG.asciidoc
```diff
@@ -31,6 +31,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Changed Gremlin grammar to make use of `g` to spawn child traversals a syntax error.
 * Added `unexpected-response` handler to `ws` for `gremlin-javascript`
 * Fixed bug in `TinkerTransactionGraph` where a read-only transaction may leave elements trapped in a "zombie transaction".
+* Added quotations in 'gremlin.sh' to accept directories with spaces in the name
```
kenhuuu, line 34:
```suggestion
* Fixed bug in `gremlin.sh` where it couldn't accept a directory name containing spaces.
```
In general, you want to explain what you fixed, not what you did.

xiazcy:
VOTE +1

PR #3020, opened 2025-02-06
[TINKERPOP-2489] Server doesn't start if folder has spaces

flora-jin:
TINKERPOP: https://issues.apache.org/jira/browse/TINKERPOP-2489

Bug: If working directory contained spaces running gremlin-server.sh resulted in file not found exception

Fix: added quotations to appropriate areas and fixed code that replaced spaces with colons to not change any spaces directory names.


gremlin-server/src/main/bin/gremlin-server.sh
```diff
@@ -96,7 +96,7 @@ fi
 
 # Build Java CLASSPATH
 CP="$GREMLIN_HOME/conf/"
-CP="$CP":$( echo $GREMLIN_HOME/lib/*.jar . | sed 's/ /:/g')
```
andreachild, line 99: This logic is also in gremlin-console `gremlin.sh` which had a very similar bug that you fixed as part of https://github.com/apache/tinkerpop/pull/3012. Does this mean that the console's `gremlin.sh` also needs this change?

flora-jin: The logic in gremlin-console doesn't include any working directory names and only includes lib/*.jar. which is why it was okay in gremlin-console/bin/gremlin.sh I believe.

andreachild: Right, that makes sense

gremlin-server/src/main/bin/gremlin-server.sh
```diff
@@ -166,7 +166,7 @@ start() {
       exit 1
     fi
 
-    $JAVA -Dlogback.configurationFile=$LOGBACK_CONF $JAVA_OPTIONS -cp $CLASSPATH $GREMLIN_SERVER_CMD "$GREMLIN_YAML" >> "$LOG_FILE" 2>&1 &
+    $JAVA -Dlogback.configurationFile="$LOGBACK_CONF" $JAVA_OPTIONS -cp "$CLASSPATH" "$GREMLIN_SERVER_CMD" "$GREMLIN_YAML" >> "$LOG_FILE" 2>&1 &
```
andreachild, line 169: Should `JAVA_OPTIONS` also be quoted?

flora-jin: JAVA_OPTIONS being quoted threw an exception regarding memory.

andreachild:
VOTE +1

gremlin-server/src/main/bin/gremlin-server.sh
```diff
@@ -96,7 +96,7 @@ fi
 
 # Build Java CLASSPATH
 CP="$GREMLIN_HOME/conf/"
-CP="$CP":$( echo $GREMLIN_HOME/lib/*.jar . | sed 's/ /:/g')
+CP="$CP":$( find "$GREMLIN_HOME/lib" -name "*.jar" -print | tr '\n' ':' )
```
Cole-Greer, line 99: There's a slight difference in the output of these 2 commands.

`echo $GREMLIN_HOME/lib/*.jar . | sed 's/ /:/g'` produces something like this:
`path_to_x.jar:path_to_y.jar:path_to_z.jar:.`

whereas `find "$GREMLIN_HOME/lib" -name "*.jar" -print | tr '\n' ':'` produces something like this:
`path_to_x.jar:path_to_y.jar:path_to_z.jar:`

The new classpath is missing the current directory (`.`).

Also just a small nit, the new classpath is ending with a trailing `:`, I don't think this causes any issues but it is a bit atypical.

Cole-Greer:
VOTE +1

PR #3023, opened 2025-02-07
TINKERPOP-3130 Allowed none() as anonymous and chained

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-3130

Should be a non-breaking change, in that it just expands what `none()` can do to be more consistent in its usage.

VOTE +1


gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/translator/PythonTranslatorTest.java
```diff
@@ -61,6 +60,13 @@ public void shouldTranslateOption() {
         assertEquals("g.V().has('person','name','marko')", gremlinAsPython);
     }
 
+    @Test
+    public void shouldTranslateValues() {
```
andreachild, line 64: Did you mean to also add a test for `none()` translation?

```
@Test
public void shouldTranslateNone() {
    final String gremlinAsPython = translator.translate(
            g.V().none().asAdmin().getBytecode()).getScript();
    assertEquals("g.V().none()", gremlinAsPython);
}
```

spmallette: no, it was for `values("age")` though it was related to figuring out a problem with `none()`. i guess i didn't  bother to delete it. off the top of my head, i don't think it's too redundant, so probably ok to leave it there.

andreachild: Agree doesn't hurt to keep the `values` test. Should the `none()` translation test be added?

andreachild:
VOTE +1

kenhuuu:
I think this is a behavior change but can be classified as a bug fix.

VOTE +1

xiazcy:
VOTE +1

PR #3026, opened 2025-02-09
[TINKERPOP-3133] Allow customize the output partition

ministat:
set the repartition number by, for example, 'gremlin.spark.outputRepartition=500'. only integer values larger than 0 is valid input. Otherwise the repartition will be skipped silently.


spark-gremlin/src/main/java/org/apache/tinkerpop/gremlin/spark/structure/io/OutputFormatRDD.java
```diff
@@ -46,9 +46,15 @@ public final class OutputFormatRDD implements OutputRDD {
     public void writeGraphRDD(final Configuration configuration, final JavaPairRDD<Object, VertexWritable> graphRDD) {
         final org.apache.hadoop.conf.Configuration hadoopConfiguration = ConfUtil.makeHadoopConfiguration(configuration);
         final String outputLocation = hadoopConfiguration.get(Constants.GREMLIN_HADOOP_OUTPUT_LOCATION);
+        final String repartitionString = hadoopConfiguration.get(Constants.GREMLIN_SPARK_OUTPUT_REPARTITION);
```
andreachild, line 49: Nit: this int parsing can be moved inside the if block as it doesn't need to be done unless the `outputLocation` is non-null.

spark-gremlin/src/main/java/org/apache/tinkerpop/gremlin/spark/structure/io/OutputFormatRDD.java
```diff
@@ -46,9 +46,15 @@ public final class OutputFormatRDD implements OutputRDD {
...
+        final int repartition = null == repartitionString ? -1 : Integer.parseInt(repartitionString);
         if (null != outputLocation) {
             // map back to a <nullwritable,vertexwritable> stream for output
-            graphRDD.mapToPair(tuple -> new Tuple2<>(NullWritable.get(), tuple._2()))
+            JavaPairRDD<Object, VertexWritable> javaPairRDD = graphRDD;
+            if (repartition > 0) {
+                javaPairRDD = javaPairRDD.repartition(repartition);
```
andreachild, line 55: Should this configuration also apply to the `writeMemoryRDD` method? If not should the configuration name be more specific to exclude memory?

ministat: Yes, it should be applied to both GraphRDD and MemoryRDD.

spark-gremlin/src/main/java/org/apache/tinkerpop/gremlin/spark/structure/io/OutputFormatRDD.java
ministat: Accept

spark-gremlin/src/main/java/org/apache/tinkerpop/gremlin/spark/structure/io/OutputFormatRDD.java
```diff
@@ -75,4 +77,17 @@ public <K, V> Iterator<KeyValue<K, V>> writeMemoryRDD(final Configuration config
...
     }
-}
\ No newline at end of file
+
+    /**
+     * Allow users to customize the RDD partitions to reduce HDFS small files
+     */
+    private <K, V> JavaPairRDD<K, V> repartitionJavaPairRDD(final org.apache.hadoop.conf.Configuration hadoopConfiguration, JavaPairRDD<K, V> graphRDD) {
+        JavaPairRDD<K, V> javaPairRDD = graphRDD;
+        final String repartitionString = hadoopConfiguration.get(Constants.GREMLIN_SPARK_OUTPUT_REPARTITION);
```
andreachild, line 86: Does `PersistedOutputRDD` also need to be aware of this configuration?

andreachild:
VOTE +1

Cole-Greer:
What's the impact of this change if the users do not explicitly configure the spark output partitioning? Does this change impact the default behaviour in any meaningful way?

Also is this intended to be targeting the master branch or is it intended for 3.7-dev?

spark-gremlin/src/main/java/org/apache/tinkerpop/gremlin/spark/structure/io/PersistedOutputRDD.java
```diff
@@ -73,15 +74,29 @@ public <K, V> Iterator<KeyValue<K, V>> writeMemoryRDD(final Configuration config
...
 
     @Override
     public boolean supportsResultGraphPersistCombination(final GraphComputer.ResultGraph resultGraph, final GraphComputer.Persist persist) {
         return persist.equals(GraphComputer.Persist.NOTHING) || resultGraph.equals(GraphComputer.ResultGraph.NEW);
     }
+
+    /**
+     * Allow users to customize the RDD partitions to reduce HDFS small files
+     */
+    private static <K, V> JavaPairRDD<K, V> repartitionJavaPairRDD(final Configuration configuration, JavaPairRDD<K, V> graphRDD) {
```
andreachild, line 93: Nit: you can pull this common code into the shared interface `OutputRDD` as a default method.

ministat:
> What's the impact of this change if the users do not explicitly configure the spark output partitioning? Does this change impact the default behaviour in any meaningful way?
If not explicitly configure this option, the output partitions number is determined by the input dataset. If the input data contains many partitions based on the default partition policy in Spark, that will cause small HDFS files problem. That is why I create this PR. But if user can tolerate small file problem, it is ok to not configure this option.
> 
> Also is this intended to be targeting the master branch or is it intended for 3.7-dev?
In my previous PR, I targeted 3.7-dev, so I follow it here. Shall I change to target for master?
> 
> Could you also add a quick changelog entry and document the new configuration in https://github.com/apache/tinkerpop/blob/master/docs/src/reference/implementations-spark.asciidoc?
Sure

spark-gremlin/src/main/java/org/apache/tinkerpop/gremlin/spark/structure/io/PersistedOutputRDD.java
ministat: ok

Cole-Greer:
>> Also is this intended to be targeting the master branch or is it intended for 3.7-dev?
> 
> In my previous PR, I targeted 3.7-dev, so I follow it here. Shall I change to target for master?

I'm not sure which previous PR you're referencing here. Our branching strategy is such that any change to an older development branch will be merged up into newer dev branches such that it gets included in all upcoming releases. A PR which targets 3.7-dev will also get merged up into master, but a PR which targets master will only be merged there. In other words, a PR which targets 3.7-dev will be included in the upcoming 3.7.4 and 4.0.0 releases, where a PR targeting master will only be included in the 4.0.0 release.

This is a non-breaking change so it is ok to target 3.7-dev if you would like it included in 3.7.4.

VOTE +1 (pending confirmation of desired target branch)

ministat:
Thanks for explanation. I hope this change can be included in the upcoming release.

PR #3030, opened 2025-02-10
[TINKERPOP-3024] Deprecate has_key_() for has_key() in python

flora-jin:
Tinkerpop: https://issues.apache.org/jira/browse/TINKERPOP-3024

has_key_() was added to python because it was reserved on Dictionary but that was removed in python 3. since TinkerPop doesn't support python 2 anymore there is little need for this difference anymore. deprecate the underscore suffixed version and add the property has_key() version.

### DIFF
--- a/gremlin-python/src/main/python/gremlin_python/process/graph_traversal.py
+++ b/gremlin-python/src/main/python/gremlin_python/process/graph_traversal.py
@@ -542,9 +542,16 @@ class GraphTraversal(Traversal):
             "gremlin_python.process.GraphTraversalSource.hasKey will be replaced by "
             "gremlin_python.process.GraphTraversalSource.has_key.",
             DeprecationWarning)
-        return self.has_key_(*args)
+        return self.has_key(*args)
 
     def has_key_(self, *args):
+        warnings.warn(
+            "gremlin_python.process.GraphTraversalSource.has_key_ will be replaced by "
+            "gremlin_python.process.GraphTraversalSource.has_key.",
+            DeprecationWarning)
+        return self.has_key(*args)
+    
+    def has_key(self, *args):
         self.bytecode.add_step("hasKey", *args)
         return self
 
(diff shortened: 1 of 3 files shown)


gremlin-python/src/main/python/gremlin_python/process/graph_traversal.py
```diff
@@ -2485,7 +2502,9 @@ def where(*args):
 
 statics.add_static('hasKey', hasKey)
 
-statics.add_static('has_key', has_key_)
+statics.add_static('has_key_', has_key_)
```
andreachild, line 2505: I don't think you need to add this static import to the deprecated `has_key_`, you can just change the existing static `has_key` to point to the new `has_key` instead of the deprecated `has_key_`.

andreachild:
I see a reference to `has_key_` in gremlin-variants.asciidoc` that should be removed:

```
In situations where Python reserved words and global functions overlap with standard Gremlin steps and tokens, those
bits of conflicting Gremlin get an underscore appended as a suffix:

*Steps* - <<all-step,all_()>>, <<and-step,and_()>>, <<any-step,any_()>>, <<as-step,as_()>>, <<filter-step,filter_()>>, <<from-step,from_()>>,
<<has-step,has_key_>>, <<id-step,id_()>>, <<is-step,is_()>>, <<in-step,in_()>>, <<max-step,max_()>>,
<<min-step,min_()>>, <<not-step,not_()>>, <<or-step,or_()>>, <<range-step,range_()>>, <<sum-step,sum_()>>,
<<with-step,with_()>>
```

kenhuuu:
VOTE +1

andreachild:
VOTE +1

andreachild:
VOTE +1

Cole-Greer:
VOTE +1

xiazcy:
VOTE +1

PR #3034, opened 2025-02-12
Update Graph Binary serializer to 4.0.0 spec for Go 

xiazcy:
First set of changes for type serializers in Go. Updated element serializers, datetime, list and map serializers, and removed old types no longer in 4.0 specs. 

For ordered map, Go doesn't have native support for ordered map, so an open source library was used. Alternative option is to not support ordered map in Go and just return regular map if it's a concern to users, and note in documentation.

### DIFF
--- a/gremlin-go/driver/graphBinary.go
+++ b/gremlin-go/driver/graphBinary.go
@@ -23,6 +23,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
+	"github.com/wk8/go-ordered-map/v2"
 	"math"
 	"math/big"
 	"reflect"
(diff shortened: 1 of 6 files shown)


gremlin-go/driver/graphBinary.go
```diff
@@ -952,15 +761,28 @@ func getDefaultValue(dataType dataType) interface{} {
...
+			val, err := readFullyQualifiedNullable(data, i, true)
+			if err != nil {
+				return nil, err
+			}
+			bulk := readIntSafe(data, i)
+			for k := int32(0); k < bulk; k++ {
+				valList = append(valList, val)
+			}
+		}
+	} else {
```
andreachild, line 778: Does this work if the flag is the null value flag `0x01`?

gremlin-go/driver/graphBinary.go
```diff
@@ -1028,7 +877,8 @@ func readMapUnqualified(data *[]byte, i *int) (interface{}, error) {
 }
 
 func readSet(data *[]byte, i *int) (interface{}, error) {
-	list, err := readList(data, i)
+	// TODO placeholder flag
```
andreachild, line 880: Why was this left as TODO?

gremlin-go/driver/graphBinary.go
```diff
@@ -1351,9 +1073,17 @@ func readFullyQualifiedNullable(data *[]byte, i *int, nullable bool) (interface{
 		}
 		return nil, nil
 	} else if nullable {
-		if readByteSafe(data, i) == valueFlagNull {
+		flag := readByteSafe(data, i)
+		fmt.Println(flag)
```
andreachild, line 1077: Should be removed?

gremlin-go/driver/graphBinary.go
```diff
@@ -168,92 +148,45 @@ func mapWriter(value interface{}, buffer *bytes.Buffer, typeSerializer *graphBin
...
-	keys := v.MapKeys()
-	err := binary.Write(buffer, binary.BigEndian, int32(len(keys)))
-	if err != nil {
-		return nil, err
-	}
-	for _, k := range keys {
-		convKey := k.Convert(v.Type().Key())
-		// serialize k
-		_, err := typeSerializer.write(k.Interface(), buffer)
+	if val, ok := value.(*orderedmap.OrderedMap[interface{}, interface{}]); ok {
```
andreachild, line 151: I can't seem to find where the {value_flag} equal to 0x02 for an ordered map is set, can you point me to it?

gremlin-go/driver/graphBinary.go
```diff
@@ -952,15 +761,28 @@ func getDefaultValue(dataType dataType) interface{} {
...
-		if err != nil {
-			return nil, err
+	if flag == 0x02 {
+		for j := int32(0); j < sz; j++ {
+			val, err := readFullyQualifiedNullable(data, i, true)
+			if err != nil {
+				return nil, err
+			}
+			bulk := readIntSafe(data, i)
+			for k := int32(0); k < bulk; k++ {
```
andreachild, line 774: I can't seem to find where bulk list value flag and value are written. Can you point me to it?

gremlin-go/driver/graphBinary.go
andreachild, line 1: I can't find any usage of `readFullyQualifiedNullable` where `nullable` isn't hardcoded to `true`. Can that function param just be removed?

gremlin-go/driver/graphBinary.go
```diff
@@ -1351,9 +1073,17 @@ func readFullyQualifiedNullable(data *[]byte, i *int, nullable bool) (interface{
 		}
 		return nil, nil
 	} else if nullable {
-		if readByteSafe(data, i) == valueFlagNull {
+		flag := readByteSafe(data, i)
+		fmt.Println(flag)
+		if flag == valueFlagNull {
 			return getDefaultValue(dataTyp), nil
 		}
+		if dataTyp == listType {
```
andreachild, line 1081: From what I can tell the `readList` and `readMap`  functions are treated differently here because they accept the `flag` as a param. Since the `flag` value is always read before calling the deserializer can the deserializer's `reader` function be changed to accept a `flag` param and the implementations can either use it or ignore it depending on the type?

gremlin-go/driver/graphBinary.go
xiazcy: So both ordered dict and bulk list are result formats that's sent by the server, nothing (that I'm aware of, I could be wrong) the GLV sends is bulked or ordered.

gremlin-go/driver/graphBinary.go
xiazcy: If I recall correctly, to do that we'll have to update implementations for all and I wanted the change to be relatively isolated to these 2 types, but I'll take a closer look.

gremlin-go/driver/graphBinary.go
xiazcy: I'll double check, but I think that'll be a separate set of changes from this PR

gremlin-go/driver/graphBinary.go
```diff
@@ -975,32 +797,59 @@ func readByteBuffer(data *[]byte, i *int) (interface{}, error) {
...
+			if k == nil {
+				mapData.Set(nil, v)
+			} else if reflect.TypeOf(k).Comparable() {
+				mapData.Set(k, v)
+			} else {
+				switch reflect.TypeOf(k).Kind() {
+				case reflect.Map:
+					mapData.Set(&k, v)
+				default:
+					mapData.Set(fmt.Sprint(k), v)
```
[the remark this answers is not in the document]
xiazcy: Yes, slices themselves cannot be set as keys (non-comparable type) in go, this turns it into a string to be used as the key.

gremlin-go/driver/graphBinary.go
xiazcy: I meant to update set as well. Will do that.

xiazcy:
Merging this now to feature branch.

PR #3057, opened 2025-02-25
Replace websockets with http

andreachild:
Changed the go driver to use http instead of web sockets, using go's out of the box http client. Connection management is delegated to the http client. Minimal changes were made to serialization/deserialization - just enough to be able to send a g.V().count() request and receive a response. POC was manually tested using the Test client.submit() test in connection_test.go. Docker deletion of containers as part of test cleanup disabled for troubleshooting purposes (after tests start passing, can be re-enabled). Still TODO: Full connection config, response chunking, removal of session, error response handling.


xiazcy:
LGTM. Just going to merge into feature branch, so no need for full review process.

PR #3066, opened 2025-03-03
Fix serializer_test.go for feature branch

andreachild:
Fixes for the Golang driver feature branch related to failures with the `serializer_test.go` test.
* Removed tests for custom type serialization. 
* Removed response id as it is no longer used. 
* Adjusted expected serialized request bytes according to TP4 format. 
* Adjusted bytes for testing deserialization according to TP4 format. 
* Removed serializer unwrapping of single item results as it was causing serialization test to fail.

### DIFF
--- a/gremlin-go/driver/response.go
+++ b/gremlin-go/driver/response.go
@@ -19,8 +19,6 @@ under the License.
 
 package gremlingo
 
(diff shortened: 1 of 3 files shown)


xiazcy:
LGTM. Merging to feature branch.

PR #3068, opened 2025-03-07
TINKERPOP-3015 Use wildcard instead of Object for hasId/hasValue predicates

Cole-Greer:
https://issues.apache.org/jira/browse/TINKERPOP-3015

Update hasId(P) signature to accept P<?> instead of P<Object>. This allows for types such as P<String> which are not subtypes of P<Object> to be passed directly to this method, instead of going through the hasId(Object) overload and then casting to a P raw type. Note that the P special case handling still exists in hasId(Object) to cover any cases in which the correct overload is not identifiable at compile-time.

All of the above applies equivalently for hasValue().

This is non-breaking and targeting 3.7-dev as all it is doing is allowing more predicates to go into hasId(P) directly, and skipping the cast to the P raw-type.

My vote is VOTE +1


vkagamlyk:
VOTE +1

PR #3075, opened 2025-03-19
TINKERPOP-3144 Close traversal on failure

langrp:
https://issues.apache.org/jira/browse/TINKERPOP-3144

Close traversal on execution failure or interruption in order to release all its resources. Currently the traversal gets closed only when its consumed entirely, but it will not get closed on failures.

`StandardOpProcessor` is kinda in question still as it requires to read bindings and close all traversals.


spmallette:
i'm a bit surprised we've not had a call to close the traversal on exceptions in Gremlin Server. Not sure how we've gotten away with that.  Is there not similar need for this with script executions (e.g. `AbstractEvalOpProcessor`)  where a `Traversal` is produce?

langrp:
Yes, indeed there is a need for it also in script execution. I missed that part.

xiazcy:
Thanks for the contribution! VOTE +1

Cole-Greer:
VOTE +1

kenhuuu:
VOTE +1

PR #3076, opened 2025-03-25
TINKERPOP-2971 Fixed bug in value traversal to group()

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-2971

Ensured that reducing/supplying barriers that follow collecting barriers get some respect. Basically, `group()` used to do this:

```
gremlin> g.V().has("person","name",P.within("vadas","peter")).group().by().by(__.out().fold())
==>[v[2]:[],v[6]:[v[3]]]
gremlin> g.V().has("person","name",P.within("vadas","peter")).group().by().by(__.out().order().fold())
==>[v[6]:[v[3]]]
```

after this change, the second traversal has the same result as the first as one would expect. Had to add an entry for gryo. I guess we never had tests that pushed an adjacent vertex into the traverser (if we had I guess we'd have found this bug earlier) and as a result gryo needs to know about it.

VOTE +1


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/util/detached/DetachedFactory.java
```diff
@@ -26,6 +27,7 @@
 import org.apache.tinkerpop.gremlin.structure.Property;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.apache.tinkerpop.gremlin.structure.VertexProperty;
+import org.apache.tinkerpop.gremlin.structure.util.star.StarGraph;
```
Cole-Greer, line 30: Nit: unused?

Cole-Greer:
VOTE +1

xiazcy:
VOTE +1

PR #3079, opened 2025-04-01
Fix connection pool bug which can cause delays in obtaining a connection

andreachild:
It has been observed in some environments that integration tests such as `GraphBinaryGroovyRemoteFeatureTest` can encounter 16 second delays between test executions. Investigation shows this can occur if one thread is attempting to obtain a connection and another thread is creating or releasing a connection at the same time. It is possible for the thread releasing or creating a connection to signal the `hasAvailableConnection` condition after the other thread has determined there are no available connections but before calling `await` on the condition. Thus the thread obtaining the connection will wait for default timeout of 16 seconds before attempting to retrieve a connection again, causing the observed delays.

This fix involves locking the `waitLock` before obtaining and potentially calling await for an available connection which prevents other threads from signalling the `hasAvailableConnection` before other threads call `await`. The fix was tested by observing integration test execution on environments which previously produced the 16 second delays and also by executing `ProfilingApplication`. The results of `ProfilingApplication` did not demonstrate any degradation of average requests per second.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
@@ -157,6 +157,9 @@ final class ConnectionPool {
         // Get valid connection
         Connection availableConnection;
         try {
+            // acquire lock before attempting to obtain an available connection and potentially waiting for one
+            // to prevent race conditions with other threads that are releasing or creating new connections
+            // otherwise signal may be called before await which would result in timeouts
             waitLock.lock();
             availableConnection = getAvailableConnection();
             if (availableConnection == null) {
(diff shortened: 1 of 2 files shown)


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
```diff
@@ -526,20 +537,20 @@ private Connection getAvailableConnection() {
 
     private void awaitAvailableConnection(long timeout, TimeUnit unit) throws InterruptedException {
```
Cole-Greer, line 538: Nit: I'm a bit worried of the potential for future mistakes now that the caller is responsible for obtaining `waitLock` before calling this method. A javadoc comment here, and on `waitForConnection()` might be helpful.

kenhuuu:
VOTE +1

xiazcy:
VOTE +1

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/ConnectionPool.java
andreachild: Technically calling this method on its own does not mandate acquiring the lock. The issue was the combination of calling `getAvailableConnection` and then subsequently `waitForConnection` without the lock. I have added a comment inside of the `borrowConnection` to clarify.

PR #3090, opened 2025-04-04
Backport OffsetDateTime changes from 4.0

xiazcy:
OffsetDateTime is now the default Date type in Gremlin. 
Changes were back-ported from Java/Python. For Go/JS/.NET, the OffsetDateTime serializers were added. 
Made OffsetDateTime the default serializer for dates in GLVs, the Date type will only deserialize responses.

VOTE +1.


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/DotNetTranslateVisitor.java
```diff
@@ -155,10 +156,10 @@ public Void visitFloatLiteral(final GremlinParser.FloatLiteralContext ctx) {
     public Void visitDateLiteral(final GremlinParser.DateLiteralContext ctx) {
         // child at 2 is the date argument to datetime() and comes enclosed in quotes
         final String dtString = ctx.getChild(2).getText();
-        final Date dt = DatetimeHelper.parse(removeFirstAndLastCharacters(dtString));
-        sb.append("DateTimeOffset.FromUnixTimeMilliseconds(");
-        sb.append(dt.getTime());
-        sb.append(")");
+        final OffsetDateTime dt = DatetimeHelper.parse(removeFirstAndLastCharacters(dtString));
```
Cole-Greer, line 159: I think this is the wrong type name for .net. If it is indeed the wrong type, I'm wondering if we have sufficient test coverage for these translators.
```suggestion
        final DateTimeOffset dt = DatetimeHelper.parse(removeFirstAndLastCharacters(dtString));
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/PythonTranslateVisitor.java
```diff
@@ -161,8 +162,8 @@ public Void visitMapEntry(final GremlinParser.MapEntryContext ctx) {
     public Void visitDateLiteral(final GremlinParser.DateLiteralContext ctx) {
         // child at 2 is the date argument to datetime() and comes enclosed in quotes
         final String dtString = ctx.getChild(2).getText();
-        final Date dt = DatetimeHelper.parse(removeFirstAndLastCharacters(dtString));
-        sb.append("datetime.datetime.utcfromtimestamp(" + dt.getTime() + " / 1000.0)");
+        final OffsetDateTime dt = DatetimeHelper.parse(removeFirstAndLastCharacters(dtString));
+        sb.append("datetime.datetime.fromtimestamp(" + dt.toEpochSecond() + ").astimezone(datetime.timezone.utc)");
```
Cole-Greer, line 166: Are we unnecessarily losing precision by using `dt.toEpochSecond()`? Python's datetime.datetime supports microsecond precision: https://docs.python.org/3/library/datetime.html#datetime.datetime

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/GoTranslateVisitor.java
```diff
@@ -57,8 +58,8 @@ public GoTranslateVisitor(final String graphTraversalSourceName) {
     public Void visitDateLiteral(final GremlinParser.DateLiteralContext ctx) {
         // child at 2 is the date argument to datetime() and comes enclosed in quotes
         final String dtString = ctx.getChild(2).getText();
-        final Date dt = DatetimeHelper.parse(removeFirstAndLastCharacters(dtString));
-        sb.append("time.UnixMilli(" + dt.getTime() + ")");
+        final OffsetDateTime dt = DatetimeHelper.parse(removeFirstAndLastCharacters(dtString));
+        sb.append("time.UnixMilli(" +  dt.toInstant().toEpochMilli() + ")");
```
Cole-Greer, line 62: We may be able to preserve more precision as well as the timezone offset via https://pkg.go.dev/time#Date

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/translator/GolangTranslator.java
```diff
@@ -138,6 +139,11 @@ protected String getSyntax(final Date o) {
             return "time.UnixMilli(" + o.getTime() + ")";
         }
 
+        @Override
+        protected String getSyntax(final OffsetDateTime o) {
+            return "time.UnixMilli(" + o.toInstant().toEpochMilli() + ")";
```
Cole-Greer, line 144: We may be able to preserve more precision as well as the timezone offset via https://pkg.go.dev/time#Date

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/DatetimeHelper.java
```diff
@@ -93,45 +97,41 @@ public static String format(final Instant d) {
      * </ul>>
      *
      */
-    public static Date parse(final String d) {
+    public static OffsetDateTime parse(final String d) {
```
Cole-Greer, line 100: There's a lot going on in this method. Based on the javadoc comment, could it simply be replaced with [OffsetDateTime.parse(String)](https://docs.oracle.com/javase/8/docs/api/java/time/OffsetDateTime.html#parse-java.lang.CharSequence-) or [OffsetDateTime.from(TemporalAccessor)](https://docs.oracle.com/javase/8/docs/api/java/time/OffsetDateTime.html#from-java.time.temporal.TemporalAccessor-)?

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/language/grammar/GeneralLiteralVisitorTest.java
```diff
@@ -500,8 +501,8 @@ public void shouldParseCurrentDate() {
             final GremlinParser parser = new GremlinParser(new CommonTokenStream(lexer));
             final GremlinParser.DateLiteralContext ctx = parser.dateLiteral();
 
-            final Date dt = (Date) new GenericLiteralVisitor(new GremlinAntlrToJava()).visitDateLiteral(ctx);
-            assertTrue(new Date().getTime() - dt.getTime() < 1000);
+            final OffsetDateTime dt = (OffsetDateTime) new GenericLiteralVisitor(new GremlinAntlrToJava()).visitDateLiteral(ctx);
+            assertTrue(OffsetDateTime.now(UTC).toEpochSecond() - dt.toEpochSecond() < 1000);
```
Cole-Greer, line 505: The < 1000 is too broad here as the units have changed from millis to seconds.
```suggestion
            assertTrue(OffsetDateTime.now(UTC).toInstant().toEpochMilli() - dt.toInstant().toEpochMilli() < 1000);
```

gremlin-go/driver/graphBinary.go
```diff
@@ -1044,6 +1078,18 @@ func timeReader(data *[]byte, i *int) (interface{}, error) {
...
 
+func offsetDateTimeReader(data *[]byte, i *int) (interface{}, error) {
+	year := readIntSafe(data, i)
+	month := readByteSafe(data, i)
+	day := readByteSafe(data, i)
+	ns := readLongSafe(data, i)
+	offset := readIntSafe(data, i)
+	// only way to pass offset info, timezone display is fixed to UTC as consequence (offset is calculated properly)
+	loc := time.FixedZone("UTC", int(offset))
+	datetime := time.Date(int(year), time.Month(month), int(day), 0, 0, 0, int(ns), loc)
```
Cole-Greer, line 1089: int(ns) may be an issue for times which are late in the day. According to the io docs, the "time" value can be from 0 to 86,399,999,999,999. The max size of `int` is only guaranteed to be 2,147,483,647, although on most 64-bit machines it will be a 64 bit integer in practice.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/JavascriptTranslateVisitor.java
```diff
@@ -133,9 +134,9 @@ public Void visitMapEntry(final GremlinParser.MapEntryContext ctx) {
     public Void visitDateLiteral(final GremlinParser.DateLiteralContext ctx) {
         // child at 2 is the date argument to datetime() and comes enclosed in quotes
         final String dtString = ctx.getChild(2).getText();
-        final Date dt = DatetimeHelper.parse(removeFirstAndLastCharacters(dtString));
+        final OffsetDateTime dt = DatetimeHelper.parse(removeFirstAndLastCharacters(dtString));
         sb.append("new Date(");
-        sb.append(dt.getTime());
+        sb.append(dt.toInstant().toEpochMilli());
```
kmcginnes, line 139: This should work, and it seems equivalent to what was there before. As long as you've tested this with various time zones it should be fine.

However, I'd wager that if someone is using the translator to translate a query, I doubt they started with a date value as the millisecond since unix epoch. More likely they started with a string version of the date time.

Perhaps a better approach would be to use the ISO string value as the input to `new Date()`. The precision should be the same (milliseconds), and it supports time zone offsets or UTC. It's just a more human friendly translation.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/DateAddStep.java
```diff
@@ -57,16 +46,32 @@ public DateAddStep(final Traversal.Admin traversal, final DT dateToken, final in
...
         final Object object = traverser.get();
 
-        if (!(object instanceof Date)) throw new IllegalArgumentException("dateAdd accept only Date.");
-
-        final Calendar cal = Calendar.getInstance();
-        cal.setTime((Date) object);
-        cal.add(DTtoCalendar.get(dateToken), value);
-
-        return cal.getTime();
+        if (!(object instanceof OffsetDateTime)) throw new IllegalArgumentException("dateAdd accept only DateTime.");
```
kmcginnes, line 52: Should the error be:

> dateAdd accept only OffsetDateTime.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/DateAddStep.java
kmcginnes, line 1: Is it intentional for this step to only have second precision?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/DateAddStep.java
xiazcy: We changed it a general DateTime type in 4.0, but you are right it prob make more sense to be OffsetDateTime here. I will also make it accept Date as well for compatibility reasons.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/DateDiffStep.java
```diff
@@ -53,17 +54,17 @@ public DateDiffStep(final Traversal.Admin traversal, final Traversal<?, Date> da
...
-        final Date otherDate = value != null ? value :
+        final OffsetDateTime otherDate = value != null ? value :
                 dateTraversal != null ? TraversalUtil.apply(traverser, dateTraversal) : null;
 
         // let's not throw exception and assume null date == 0
-        final long otherDateMs = otherDate == null ? 0 : otherDate.getTime();
+        final long otherDateMs = otherDate == null ? 0 : otherDate.toEpochSecond();
 
-        return (((Date) object).getTime() - otherDateMs) / 1000;
+        return (((OffsetDateTime) object).toEpochSecond() - otherDateMs);
```
kmcginnes, line 67: Did you mean to use `.toEpochMilli()`?

If you did intend to use seconds instead of milliseconds, then you should change the variable name `otherDateMs` to be `otherDateSeconds` or something.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/DateAddStep.java
xiazcy: The original design of the step uses the `dateToken` enums, which only goes to second precision. I suppose we could add support for additional precisions, but that would be outside the scope of this PR.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/translator/JavascriptTranslator.java
```diff
@@ -144,6 +145,11 @@ protected String getSyntax(final Date o) {
             return "new Date(" + o.getTime() + ")";
         }
 
+        @Override
+        protected String getSyntax(final OffsetDateTime o) {
+            return "new Date(" + o.toInstant().toEpochMilli() + ")";
```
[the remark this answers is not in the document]
xiazcy: Not quite, new Date() construction will create a date with local time offset regardless. The JS Date type is...messy.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/language/grammar/GeneralLiteralVisitorTest.java
kmcginnes, line 505: It looks like we are dropping from millisecond to second precision here.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Translator.java
```diff
@@ -310,6 +316,9 @@ protected Script convertToScript(final Object object) {
                 } else if (object instanceof Date) {
                     final Object objectOrWrapper = withParameters ? object : getSyntax((Date) object);
                     return script.getBoundKeyOrAssign(withParameters, objectOrWrapper);
+                } else if (object instanceof OffsetDateTime) {
+                    final Object objectOrWrapper = withParameters ? object : getSyntax((OffsetDateTime) object);
+                    return script.getBoundKeyOrAssign(withParameters, objectOrWrapper);
```
kmcginnes, line 321: Now that this exists, should the `else if (object instanceof Date)` above be removed?

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/DateDiffStepTest.java
```diff
@@ -36,64 +35,49 @@ public class DateDiffStepTest extends StepTest {
...
         assertEquals(-604800L, (long) __.__(now).dateDiff(__.constant(other)).next());
     }
 
     @Test
     public void shouldHandleNullTraversalParam() {
-        final Date now = new Date();
+        final OffsetDateTime now = OffsetDateTime.now(UTC);
 
-        assertEquals(now.getTime() / 1000, (long) __.__(now).dateDiff(__.constant(null)).next());
+        assertEquals(now.toEpochSecond(), (long) __.__(now).dateDiff(__.constant(null)).next());
```
kmcginnes, line 69: Drops precision from milliseconds to seconds.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/AnonymizedTranslatorVisitor.java
```diff
@@ -23,6 +23,7 @@
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.time.OffsetDateTime;
 import java.util.Date;
```
andreachild, line 27: This import is now unused

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/DotNetTranslateVisitor.java
```diff
@@ -27,6 +27,7 @@
 import org.apache.tinkerpop.gremlin.structure.VertexProperty;
 import org.apache.tinkerpop.gremlin.util.DatetimeHelper;
 
+import java.time.OffsetDateTime;
 import java.util.Date;
```
andreachild, line 31: This import is now unused

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/GoTranslateVisitor.java
```diff
@@ -27,6 +27,7 @@
 import org.apache.tinkerpop.gremlin.structure.VertexProperty;
 import org.apache.tinkerpop.gremlin.util.DatetimeHelper;
 
+import java.time.OffsetDateTime;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Date;
```
andreachild, line 33: This import is now unused

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/JavaTranslateVisitor.java
```diff
@@ -25,6 +25,7 @@
 import org.apache.tinkerpop.gremlin.structure.util.reference.ReferenceVertex;
 import org.apache.tinkerpop.gremlin.util.DatetimeHelper;
 
+import java.time.OffsetDateTime;
 import java.util.Date;
```
andreachild, line 29: This import is now unused

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/Translator.java
andreachild: I believe it should stay because both `Date` and `OffsetDateTime` types are still supported, just now the default is `OffsetDateTime`.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -204,6 +204,7 @@
 import org.apache.tinkerpop.gremlin.structure.VertexProperty;
 import org.apache.tinkerpop.gremlin.util.function.ConstantSupplier;
 
+import java.time.OffsetDateTime;
```
andreachild, line 207: Date import is now unused

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/__.java
```diff
@@ -37,6 +37,7 @@
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.apache.tinkerpop.gremlin.structure.VertexProperty;
 
+import java.time.OffsetDateTime;
 import java.util.Collection;
 import java.util.Date;
```
andreachild, line 42: Unused

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AsDateStep.java
```diff
@@ -20,46 +20,44 @@
...
         final Object object = traverser.get();
         if (object == null)
-            throw new IllegalArgumentException("Can't parse null as Date.");
-        if (object instanceof Date)
-            return (Date) object;
+            throw new IllegalArgumentException("Can't parse null as OffsetDateTime.");
+        if (object instanceof OffsetDateTime)
+            return (OffsetDateTime) object;
         if (object instanceof Byte || object instanceof Short || object instanceof Integer || object instanceof Long)
             // numbers handled as milliseconds since January 1, 1970, 00:00:00 GMT.
```
andreachild, line 54: Comment should be updated to `UTC` instead of `GMT` which are slightly different.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AsDateStep.java
```diff
@@ -20,46 +20,44 @@
...
         super(traversal);
     }
 
     @Override
-    protected Date map(final Traverser.Admin<S> traverser) {
+    protected OffsetDateTime map(final Traverser.Admin<S> traverser) {
         final Object object = traverser.get();
         if (object == null)
-            throw new IllegalArgumentException("Can't parse null as Date.");
-        if (object instanceof Date)
```
andreachild, line 53: Technically since `Date` types are supported isn't it possible to use the `asDate` step on a traversal that produces `Date`? In that case should the `Date` be converted to `OffsetDateTime`?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/DateDiffStep.java
```diff
@@ -25,6 +25,7 @@
 import org.apache.tinkerpop.gremlin.process.traversal.util.TraversalUtil;
 import org.apache.tinkerpop.gremlin.structure.util.StringFactory;
 
+import java.time.OffsetDateTime;
 import java.util.Collections;
 import java.util.Date;
```
andreachild, line 30: Unused

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/DateAddStep.java
Cole-Greer: I agree that the available precision here is odd, although this is really out of scope for this PR and should be taken as a separate discussion. I think there should be come consideration of taking a Duration as input instead of an int and DT enum, although some consideration is needed around how to make Duration's constructable in gremlin-lang.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/DateDiffStep.java
andreachild, line 67: Suggestion:
```
return otherDate == null ? 0L : Duration.between(otherDate, (OffsetDateTime) object).getSeconds();
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/DatetimeHelper.java
```diff
@@ -22,7 +22,11 @@
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
+import java.time.Month;
+import java.time.OffsetDateTime;
+import java.time.Year;
 import java.time.YearMonth;
```
andreachild, line 28: Unused imports

gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphBinary/DataType.cs
```diff
@@ -78,6 +78,7 @@ public class DataType : IEquatable<DataType>
         // TODO: Support metrics and traversal metrics
         public static readonly DataType Char = new DataType(0x80);
         public static readonly DataType Duration = new DataType(0x81);
+        public static readonly DataType OffsetDateTime = new DataType(0x88);
```
andreachild, line 81: Just wondering why `OffsetDateTime` was not previously supported as a type for .NET? This PR does not add the type to TInkerPop, only changes the default date type.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/DateDiffStep.java
xiazcy: No this step meant to return epoch time in seconds, so I'm just directly using it.

xiazcy: I like the simplification, not quite the right logic but I can use it, i.e. if `otherDate` is null we consider it as 0L and return the epoch seconds of `date`, instead of returning `0L`.

gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphBinary/DataType.cs
xiazcy: Not just .NET, it actually wasn't supported in any of the GLVs because it's an extended serialization type.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AsDateStep.java
xiazcy: Yes, this is part of the change I'll revert along with other changes to make it compatible with Date as type.

gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphBinary/Types/OffsetDateTimeSerializer.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            await stream.WriteIntAsync(value.Year, cancellationToken).ConfigureAwait(false);
+            await stream.WriteByteAsync(Convert.ToByte(value.Month), cancellationToken).ConfigureAwait(false);
+            await stream.WriteByteAsync(Convert.ToByte(value.Day), cancellationToken).ConfigureAwait(false);
+            var h = value.Hour;
+            var m = value.Minute;
+            var s = value.Second;
+            var ms = value.Millisecond; 
+            // Note there will be precision loss as microsecond and nanosecond access was added after .net 7
+            var ns = h * 60 * 60 * 1e9 + m * 60 * 1e9 + s * 1e9 + ms * 1e6;
```
andreachild, line 57: Q says obtaining nanoseconds is possible using `Ticks`:

```
// Convert the time portion of DateTimeOffset to nanoseconds since midnight
  static long TimeToNanosecondsSinceMidnight(DateTimeOffset dto)
  {
      // Get the time of day as TimeSpan
      TimeSpan timeOfDay = dto.TimeOfDay;
      
      // Convert ticks to nanoseconds (1 tick = 100 nanoseconds in .NET)
      return timeOfDay.Ticks * 100;
  }
```

gremlin-go/driver/graphBinary.go
xiazcy: Unfortunately the Date constructor requires `int`, it's more like a caveat of go, not sure if it's avoidable.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/DatetimeHelper.java
xiazcy: Not quite, dates without time or time without dates cannot be transformed into `OffsetDateTime`, so we'll still need the manual parsing. But you have a point that I can combine the last two cases.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AsDateStepTest.java
```diff
@@ -45,13 +46,12 @@ protected List<Traversal> getTraversals() {
...
     public void shouldParseDate() {
 
-        final Instant testInstant = ZonedDateTime.of(2023, 8, 2, 0, 0, 0, 0, UTC).toInstant();
-        final Date testDate = new Date(testInstant.getEpochSecond() * 1000);
+        final OffsetDateTime testDate = OffsetDateTime.of(LocalDateTime.of(2023, 8, 2, 0, 0, 0, 0), UTC);
 
-        assertEquals(new Date(1), __.__(1).asDate().next());
-        assertEquals(new Date(3), __.__(3L).asDate().next());
-        assertEquals(new Date(6), __.__(new BigInteger("6")).asDate().next());
-        assertEquals(testDate, __.__(testDate.getTime()).asDate().next());
```
kenhuuu, line 54: You should add tests where Date is the input traverser. Same goes for the DateAddStepTest and DateDiffStepTest

xiazcy: Yup I can do that for asDate and dateAdd (tests in dateDiff were added already)

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/PythonTranslateVisitor.java
```diff
@@ -27,6 +27,7 @@
 import org.apache.tinkerpop.gremlin.util.DatetimeHelper;
 
 import java.math.BigInteger;
+import java.time.OffsetDateTime;
```
andreachild, line 30: Unused imports

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/GoTranslateVisitor.java
```diff
@@ -57,8 +57,16 @@ public GoTranslateVisitor(final String graphTraversalSourceName) {
...
-        sb.append("time.UnixMilli(" + dt.getTime() + ")");
+        final OffsetDateTime dt = DatetimeHelper.parse(removeFirstAndLastCharacters(dtString));
+        sb.append("time.Date(").append(dt.getYear()).
+                append(", ").append(dt.getMonthValue()).
+                append(", ").append(dt.getDayOfMonth()).
+                append(", ").append(dt.getHour()).
+                append(", ").append(dt.getMinute()).
+                append(", ").append(dt.getSecond()).
+                append(", ").append(dt.getNano()).
+                append(", time.FixedZone(\"UTC\", ").append(dt.getOffset().getTotalSeconds()).append(")").
```
andreachild, line 68: Looking at documentation https://pkg.go.dev/time#FixedZone it looks like the zone name should include the offset, ie. `UTC-8`.

Also can a test be added to `GremlinTranslatorTest.java` for non zero offset scenario?

xiazcy: I can add the tests. 

Technically the fixed zone name in go for this is a string only for display purposes...I.e. you can pass in any random string and it just appears at the end of your time display (doesn't affect the date calculation at all). For simplicity I just left the "UTC" string in, but can get the zoneId from OffsetDateTime instead.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/DateAddStep.java
```diff
@@ -57,16 +48,40 @@ public DateAddStep(final Traversal.Admin traversal, final DT dateToken, final in
...
+                newDate = date.plus(Duration.ofMinutes(value));
+                break;
+            case hour:
+                newDate = date.plus(Duration.ofHours(value));
+                break;
+            case day:
+                newDate = date.plus(Duration.ofDays(value));
+                break;
+            default:
+                throw new IllegalArgumentException("DT tokens should only be second, minute, hour, or day.");
```
andreachild, line 81: Should this validation be moved to the constructor instead?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/DateDiffStep.java
```diff
@@ -36,34 +39,61 @@
...
+            } else if (traversalDate instanceof OffsetDateTime) {
+                otherDate = (OffsetDateTime) traversalDate;
+            } else {
+                throw new IllegalArgumentException(String.format("DateDiff can only take OffsetDateTime or Date (deprecated) as argument, encountered %s", object.getClass()));
+            }
+        } else {
+            otherDate = value;
+        }
 
         // let's not throw exception and assume null date == 0
```
andreachild, line 95: Nit: this comment should be moved above to else statement:

```
else {
    // let's not throw exception and assume null date should result in zero difference
    otherDate = value
}
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/DateDiffStep.java
```diff
@@ -36,34 +39,61 @@
...
+
+    public DateDiffStep(final Traversal.Admin traversal, final OffsetDateTime value) {
         super(traversal);
         this.value = value;
     }
 
-    public DateDiffStep(final Traversal.Admin traversal, final Traversal<?, Date> dateTraversal) {
+    public DateDiffStep(final Traversal.Admin traversal, final Traversal<?, ?> dateTraversal) {
         super(traversal);
         this.dateTraversal = this.integrateChild(dateTraversal.asAdmin());
```
andreachild, line 57: If the incoming `Traversal` is of type `?` then shouldn't `this.dateTraversal` be of type `?` as well since there's no guarantee the traversal resolves to an `OffsetDateTime`?

gremlin-go/driver/graphBinary.go
```diff
@@ -509,6 +510,39 @@ func timeWriter(value interface{}, buffer *bytes.Buffer, _ *graphBinaryTypeSeria
...
+	err = binary.Write(buffer, binary.BigEndian, byte(t.Month()))
+	if err != nil {
+		return nil, err
+	}
+	err = binary.Write(buffer, binary.BigEndian, byte(t.Day()))
+	if err != nil {
+		return nil, err
+	}
+	// construct time of day in nanoseconds
+	h := t.Hour()
```
andreachild, line 529: Q recommends type conversion to `int64` for these values to prevent issues with overflow on 32 bit systems.

gremlin-go/driver/graphBinary.go
Cole-Greer: I think we need to break the 64bit `ns` into `hours`, `minutes`, `seconds`, and `nanos` to ensure none of them will overflow on systems with 32bit ints. We can't allow those datetimes to overload on deserialization.

gremlin-go/driver/graphBinary.go
xiazcy: Hmmm technically all date methods return/accept `int`, so that will remain an issue for 32 bit system regardless, but I can make it `int64` to make the problem a bit less.

Cole-Greer:
VOTE +1 pending fix to Go GraphBinary overflow issue and upgrade docs.

kenhuuu:
VOTE +1

PR #3091, opened 2025-04-07
TINKERPOP-3047 Keywords are allowed as Map keys in grammar.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-3047

Major refactoring to the grammar itself, translators. Removed a few gherkin tests that tested lambdas which weren't really necessary given they will be reserved for embedded cases. Changed infinity/nan to be able to parse to float.

VOTE +1


gremlin-language/src/main/java/org/apache/tinkerpop/gremlin/language/corpus/GrammarReader.java
```diff
@@ -0,0 +1,63 @@
...
+    private static final Pattern tokenPattern = Pattern.compile("T[A-Z0-9_]+:\\s*'([^']+)'\\s*;");
+
+    /**
+     * Parse a grammar file to extract all keyword string representations.
+     *
+     * @param grammarFilePath The path to the ANTLR grammar file
+     * @return A set of keyword string representations
+     * @throws IOException If there's an error reading the file
+     */
+    public static Set<String> parse(final String grammarFilePath) throws IOException {
```
Cole-Greer, line 47: I'm not entirely sure on the purpose of this method yet, but I'm worried that the regex-based approach may be fragile if there are any keyword definitions in the grammar which are split over multiple lines. If the goal is simply to obtain a list of all keyword tokens from the grammar, it may be more robust to parse this out of the gremlin-language generated sources such as `gremlin-language/target/generated-sources/antlr4/org/apache/tinkerpop/gremlin/language/grammar/Gremlin.tokens`.

Cole-Greer: Just something like this:
```
K_ALL
    : 'all'
    ;
```
It's a bit odd to break such a short definition over multiple lines but it is completely valid antlr syntax.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/Order.feature
```diff
@@ -34,24 +34,6 @@ Feature: Step - order()
       | ripple |
       | vadas  |
 
-  Scenario: g_V_name_order_byXa1_b1X_byXb2_a2X
```
Cole-Greer, line 37: Were these scenarios removed because of an issue with the lambda? Was there something about the changes in this PR which causes them to fail where they previously passed?

gremlin-language/src/main/antlr4/Gremlin.g4
```diff
@@ -22,6 +22,12 @@ grammar Gremlin;
     PARSER RULES
 **********************************************/
 
+gremlinValue
+    : (numericLiteral | booleanLiteral | stringLiteral | dateLiteral | nullLiteral | nanLiteral | infLiteral |
```
Cole-Greer, line 26: `nanLiteral` and `infLiteral` are now captured by `numericLiteral`
```suggestion
    : (numericLiteral | booleanLiteral | stringLiteral | dateLiteral | nullLiteral |
```

gremlin-language/src/main/java/org/apache/tinkerpop/gremlin/language/corpus/GrammarReader.java
spmallette: oh....we shouldn't do that 🙂  i was shooting for just some basic good faith testing here rather than complete coverage. i cant say i feel like we need the latter for this.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/Order.feature
spmallette: removed for lambdas. given https://issues.apache.org/jira/browse/TINKERPOP-3136 it was more trouble to figure out the problem than to just get a head start on that issue. these tests already exist in embedded.

gremlin-language/src/test/java/org/apache/tinkerpop/gremlin/language/grammar/ReferenceGrammarTest.java
```diff
@@ -48,6 +52,7 @@
 public class ReferenceGrammarTest extends AbstractGrammarTest {
     private static final String featureDir = Paths.get("..", "gremlin-test", "src", "main", "resources", "org", "apache", "tinkerpop", "gremlin", "test", "features").toString();
     private static final String docsDir = Paths.get("..", "docs", "src").toString();
+    private static final String grammarFile = Paths.get("src", "main", "antlr4","Gremlin.g4").toString();
```
Cole-Greer, line 55: This appears to be unused

gremlin-language/src/test/java/org/apache/tinkerpop/gremlin/language/grammar/ReferenceGrammarTest.java
```diff
@@ -89,29 +94,73 @@ public class ReferenceGrammarTest extends AbstractGrammarTest {
...
+    public static Iterable<Pair<String, ParserRule>> parseTestItems() throws IOException {
+        // gremlin scripts from docs
+        final Set<Pair<String, ParserRule>> scripts = DocumentationReader.parse(docsDir).
+                stream().map(g -> Pair.with(g, ParserRule.QUERY_LIST)).
+                collect(Collectors.toCollection(LinkedHashSet::new));
+
+        // helps make sure we're actually adding test scripts for each load. there are well over 500 gremlin
+        // examples in the docs so we should have at least that much
+        int size = scripts.size();
+        assert size > 500;
```
Cole-Greer, line 106: Nit: Is there a reason to prefer Java assertions over JUnit assertions?

gremlin-language/src/test/java/org/apache/tinkerpop/gremlin/language/grammar/ReferenceGrammarTest.java
```diff
@@ -89,29 +94,73 @@ public class ReferenceGrammarTest extends AbstractGrammarTest {
...
+
+        // gremlin scripts from feature tests
+        scripts.addAll(FeatureReader.parseGrouped(featureDir, stringMatcherConverters).values().stream().
+                flatMap(Collection::stream).
+                map(g -> Pair.with(g, ParserRule.QUERY_LIST)).
+                collect(Collectors.toList()));
+
+        // there are well over 1000 tests so make sure we parsed something sensible
+        assert size < scripts.size() + 1000;
```
Cole-Greer, line 115:
```suggestion
        assert size + 1000 < scripts.size();
```

gremlin-language/src/test/java/org/apache/tinkerpop/gremlin/language/grammar/ReferenceGrammarTest.java
spmallette: yep - making sure scripts increment in size is not what this test is about.  the java asserts simply validate that our test is setup as we expect it to be.

gremlin-language/src/test/resources/gremlin-values.txt
```diff
@@ -0,0 +1,419 @@
...
+1e2
+1.0e2
+1e+2
+1e-2
+1.23e45
++1.23e45
+-1.23e45
++1.23e+45
+-1.23e-45
+1.0f
```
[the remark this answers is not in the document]
spmallette: probably should have a JIRA for that i guess.............

gremlin-language/src/test/resources/gremlin-values.txt
```diff
@@ -0,0 +1,419 @@
...
+'hello\u0041world'
+"hello\uAAAAworld"
+'hello\uAAAAworld'
+"!@#$%^&*()"
+'!@#$%^&*()'
+"symbols + special chars: <>,.?/"
+'symbols + special chars: <>,.?/'
+"mixed 123 abc !@#"
+'mixed 123 abc !@#'
+"line1\
```
Cole-Greer, line 129: I'm a bit surprised that this works at all considering ReferenceGrammarTest only parses a single line at a time. This might not be testing as much as it appears.

spmallette: found a bug. there's some bad data in gremlin-values.txt. cleaning that up.

Cole-Greer:
It might be nice to have some (feature?) tests which `T` and `Direction` remain valid map keys and have precedence over string keys. Other enums remain disallowed as map keys which is why those keywords can be interpreted as Strings.

Eg:
```
{label: "foo"} -> {T.label: "foo"}
{local: "foo"} -> {"local": "foo"}
```

gremlin-language/src/test/resources/gremlin-values.txt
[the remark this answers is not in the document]
Cole-Greer: Opened one up. Should be easy to knock out leading up to the next release. https://issues.apache.org/jira/browse/TINKERPOP-3151

gremlin-language/src/main/antlr4/Gremlin.g4
```diff
@@ -22,6 +22,12 @@ grammar Gremlin;
...
+       traversalT | traversalCardinality | traversalDirection | traversalMerge |
+       structureVertex | genericLiteralSet | genericLiteralCollection | genericLiteralMap) EOF
```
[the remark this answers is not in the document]
andreachild: What was the deciding factor when choosing what is considered a `gremlinValue`?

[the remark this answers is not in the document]
spmallette: anything that could be conceived of as being passed as a parameter. `Traversal` is not one of those things.

spmallette:
> It might be nice to have some (feature?) tests which T and Direction remain valid map keys and have precedence over string keys. 

not sure feature tests will work for this. bytecode and the translator will convert/normalize them to the long form so it won't actually get tested.  might need a new class of pure "script" tests? either that or make a configuration of the translators that will use the short form? JIRA?

gremlin-language/src/main/antlr4/Gremlin.g4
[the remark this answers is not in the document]
andreachild: To exclude traversals and other non-paramterizables would it make sense to create separate `gremlinValueSet`, `gremlinValueCollection`, `gremlinValueMap` instead of using the `genericLiteralX` rules?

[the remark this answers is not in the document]
spmallette: yeah, that is the path under consideration. just would like to do it while maximizing reuse. still thinking...

gremlin-language/src/main/antlr4/Gremlin.g4
```diff
@@ -67,34 +73,34 @@ traversalSourceSelfMethod
...
-    | 'withSack' LPAREN genericLiteralArgument COMMA traversalBiFunctionArgument RPAREN
+    : K_WITHSACK LPAREN genericLiteralArgument RPAREN
+    | K_WITHSACK LPAREN genericLiteralArgument COMMA traversalBiFunction RPAREN
     ;
 
 traversalSourceSelfMethod_withSideEffect
-    : 'withSideEffect' LPAREN stringArgument COMMA genericLiteralArgument RPAREN
-    | 'withSideEffect' LPAREN stringArgument COMMA genericLiteralArgument COMMA traversalBiFunctionArgument RPAREN
+    : K_WITHSIDEEFFECT LPAREN stringArgument COMMA genericLiteralArgument RPAREN
+    | K_WITHSIDEEFFECT LPAREN stringArgument COMMA genericLiteralArgument COMMA traversalBiFunction RPAREN
```
andreachild, line 90: Is the replacement of `traversalBiFunctionArgument` with `traversalBiFunction` backwards compatible? The `traversalBiFunctionArgument` (prior to its deletion in this PR) used to allow `variable` but `traversalBiFunction` does not.

```
traversalBiFunctionArgument
    : traversalBiFunction
    | variable
    ;
```

```
traversalBiFunction
    : traversalOperator
    ;
```

Cole-Greer:
> > It might be nice to have some (feature?) tests which T and Direction remain valid map keys and have precedence over string keys.
> 
> not sure feature tests will work for this. bytecode and the translator will convert/normalize them to the long form so it won't actually get tested. might need a new class of pure "script" tests? either that or make a configuration of the translators that will use the short form? JIRA?

I didn't consider that, it can be left out of scope here. I'm not sure that such a test case is enough to motivate a new class of tests or translator config.

gremlin-language/src/main/antlr4/Gremlin.g4
spmallette: the intention is to break this (and other places) as we go about using variables more judiciously and just where they make sense. arguments that are enums aren't one of those places where variables will be used. they don't really make sense there. the only reason Direction and T are in that case is because of `mergeE/V` and their use inside the `Map` argument as a special situation.

andreachild: Should there be a changelog and/or upgrade doc entry to list the scenarios that will no longer accept variables? I also see some scenarios where variable support was added which didn't have it previously (ex. `traversalMethod_both`). These could also use a changelog entry.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AddPropertyStep.java
```diff
@@ -34,6 +34,7 @@
 import org.apache.tinkerpop.gremlin.structure.T;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.apache.tinkerpop.gremlin.structure.VertexProperty;
+import org.apache.tinkerpop.gremlin.structure.io.gryo.kryoshim.shaded.ShadedOutputAdapter;
```
kenhuuu, line 37: Is this import actually used?

kenhuuu:
Something that I find a bit unintuitive with how map keys are defined in the grammar is that you can't use fully qualified Enums as keys. If that is the intention then it would be nice to add some tests to let others know this was intentional.

gremlin-language/src/main/antlr4/Gremlin.g4
spmallette: there are many more changes to come for this, so in this case, i'd say its not necessary because a more general body of changes will sum all of this up before release. i think we just want one CHANGELOG entry to reflect that "variable usage has been narrowed" otherwise we'll have scores of entries for all the changes to come. i just needed to make some things consistent for the `Map` keys change to work.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/GenericLiteralVisitor.java
```diff
@@ -331,33 +331,39 @@ public Object visitGenericLiteralMap(final GremlinParser.GenericLiteralMapContex
             // [key : value] - index 0 is key, index 1 is COLON, index 2 is value
             // also note that a child count of 5 indicates the key is an expression wrapped
```
andreachild, line 332: This comment should be updated to reference the updated logic for determining if the key is an expression (presence of left parenthesis) or removed.

Cole-Greer:
Thanks Stephen, everything looks good.
VOTE +1

kenhuuu:
VOTE +1

PR #3094, opened 2025-04-14
Update feature tests to assert traversal construction exceptions.

kenhuuu:
It wasn't possible to assert for failures that occurred during traversal construction in the feature tests which made them unsuitable for testing steps that threw exceptions.

Go Traversal needed to be updated as well since HasNext and Next didn't properly surface those exceptions.

### DIFF
--- a/docs/src/upgrade/release-3.8.x.asciidoc
+++ b/docs/src/upgrade/release-3.8.x.asciidoc
@@ -253,14 +253,6 @@ The `GremlinLangScriptEngine` has been modified to treat float literals without 
 or 'd') as Double by default. Users who need BigDecimal precision can still use the 'm' suffix (e.g., 1.0m). 
 `GremlinGroovyScriptEngine` will still default to BigDecimal for float literals.
 
-==== Gremlin Go API Changes
-
-The `IsEmpty` function in `resultSet` has been changed to return `(bool, error)`. If an error is detected in the
-channel then `true` will be returned along with the error. This is in contrast to the previous behavior where true
-was returned and the error was ignored which lead to error information being missing from the result. This may have
-lead you to believe that there were simply no results when there was actually an issue with the query. This change also
-means the `HasNext` and `Next` functions for `Traversal` will correctly return the error.
-
 ==== group() Value Traversal Semantics
 
 The `group()` step takes two `by()` modulators. The first defines the key for the grouping and the second acts upon the
--- a/docs/src/upgrade/release-3.8.x.asciidoc
+++ b/docs/src/upgrade/release-3.8.x.asciidoc
@@ -253,6 +253,14 @@ The `GremlinLangScriptEngine` has been modified to treat float literals without 
 or 'd') as Double by default. Users who need BigDecimal precision can still use the 'm' suffix (e.g., 1.0m). 
 `GremlinGroovyScriptEngine` will still default to BigDecimal for float literals.
 
+==== Gremlin Go API Changes
+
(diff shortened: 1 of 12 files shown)


docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -253,6 +253,14 @@ The `GremlinLangScriptEngine` has been modified to treat float literals without
 or 'd') as Double by default. Users who need BigDecimal precision can still use the 'm' suffix (e.g., 1.0m). 
 `GremlinGroovyScriptEngine` will still default to BigDecimal for float literals.
 
+==== Gremlin Go API Changes
+
+The `IsEmpty` function in `resultSet` has been changed to return `(bool, error)`. If an error is detected in the
+channel then `false` will be returned along with the error. This is in contrast to the previous behavior where false
```
Cole-Greer, line 259: Should this be
```suggestion
channel then `true` will be returned along with the error. This is in contrast to the previous behavior where true
```

gremlin-go/driver/resultSet.go
```diff
@@ -70,25 +70,33 @@ func (channelResultSet *channelResultSet) sendSignal() {
...
 	// If our channel is empty and we have no data in it, wait for signal that the state has been updated.
 	if len(channelResultSet.channel) != 0 {
 		// Channel is not empty.
 		channelResultSet.channelMutex.Unlock()
-		return false
+		return false, nil
+	} else if channelResultSet.err != nil {
+		// Channel has an error.
+		channelResultSet.channelMutex.Unlock()
+		return true, channelResultSet.err
```
andreachild, line 95: Is it possible for the result set to be non-empty AND have an error? ie. maybe some results were obtained but then something unexpected happened and the full result set wasn't able to be obtained

kenhuuu: This can happen with either an evaluation timeout or an OOM from the server. What is your suggestion for this case?

andreachild: Then instead of returning `true, channelResultSet.err` can you return `len(channelResultSet.channel) != 0, channelResultSet.err`?

gremlin-go/driver/traversal.go
```diff
@@ -102,7 +102,13 @@ func (t *Traversal) HasNext() (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	return !results.IsEmpty(), nil
+
+	isEmpty, emptyErr := results.IsEmpty()
+	if emptyErr != nil {
+		return false, emptyErr
```
andreachild, line 108: Could this prevent caller from obtaining partial results? Similar to https://github.com/apache/tinkerpop/pull/3094/files#r2045029111

gremlin-go/driver/resultSet.go
kenhuuu: I think this makes sense, but it will cause a slight behavior difference with other GLVs. For example, the Java driver may throw away unread results if an error comes in. Is it important for the user to get all of the "partial results" if they ultimately can't get all the results due to an exception from the server?

andreachild: Good point, do you know how the other GLVs handle this scenario? Are there cases when the Java driver may not throw away unread results if an error occurs?

gremlin-go/driver/traversal.go
kenhuuu: No longer relevant as this change was reverted.

xiazcy:
VOTE +1

Cole-Greer:
VOTE +1

kenhuuu:
VOTE +1

PR #3095, opened 2025-04-16
TINKERPOP-3155 Fixed typo for Operator.addAll to check that all args are collections instead of just the first

andreachild:
[TINKERPOP-3155](https://issues.apache.org/jira/browse/TINKERPOP-3155) 

Fixed typo for Operator.addAll to check that all args are collections instead of just the first and added unit test.


Cole-Greer:
VOTE +1

xiazcy:
VOTE +1

PR #3097, opened 2025-04-17
TINKERPOP-3136 Complete Gremlin test suite move to Gherkin

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-3136

Gherkin is now the official way to test Gremlin behavior and is preferred over the old Java test suite. There is still a JVM-specific test suite focused on things like Lambdas which aren't supported by the grammar. Cleaned up test ignores to be more current. Newest ignores are for the new `subgraph()` and `tree()` tests which don't have support outside of Java, though those features are coming, so will be enabled soon. 

This PR is only for the `3.8-dev` portion of this work - will squash to a single commit for merge. 

VOTE +1


gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/ProcessLimitedStandardSuite.java
```diff
@@ -18,105 +18,20 @@
...
-            SubgraphTest.class,
-            TreeTest.class,
-    };
-
-    /**
-     * This constructor is used by JUnit and will run this suite with its concrete implementations of the
-     * {@code testsToEnforce}.
-     */
+@Deprecated
+public class ProcessLimitedStandardSuite extends ProcessEmbeddedStandardSuite {
```
Cole-Greer, line 29: Do these old "limited" suites have any utility left after stripping out `allTests` and `testsToEnforce`? I'm wondering if we are better off completely removing the old suites instead of leaving them stripped bare and deprecated.

neo4j-gremlin/pom.xml
```diff
@@ -50,6 +50,11 @@ limitations under the License.
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.apache.tinkerpop</groupId>
+            <artifactId>tinkergraph-gremlin</artifactId>
+            <version>${project.version}</version>
```
Cole-Greer, line 56: Nit: Can this be limited to `<scope>test</scope>`?

Cole-Greer:
VOTE +1

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/ProcessLimitedStandardSuite.java
spmallette: the utility is that it gives providers a chance to upgrade without breaks. they can make the change at their discretion rather than face a compile error. i dont think it hurts to leave them in 3.x and remove them in 4.x.

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/features/StepDefinition.java
```diff
@@ -637,4 +723,70 @@ private static <T> List<org.hamcrest.Matcher<? super T>> getMatchers(final T[] i
...
+        return roots;
+    }
+
+    private static int countLeadingCharacters(final String line) {
+        int count = 0;
+        for (char c : line.toCharArray()) {
+            if (c == ' ') count++;
+            else break;
+        }
+        return count / 3; // Assuming 4 spaces per level
```
xiazcy, line 759:
```suggestion
        return count / 3; // Assuming 3 spaces per level
```

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/features/StepDefinition.java
```diff
@@ -374,6 +375,91 @@ else if (result instanceof Map)
...
+
+            // expected vertices match the count in the graph
+            assertEquals(expectedVertices.size(), sg.V().count().next().intValue());
+
+            // assert the structure of the subgraph. there should be no references here as this is serialized as
+            // a full TinkerGraph. also, ids should be the same as they are in the source graph so we can use all
+            // of this to do a complete assertion. there is only support for the modern graph right now but it
+            // wouldn't be hard to add others.
+            for (Vertex vertex : expectedVertices) {
+                final String variableKey = vertex.label().equals("person") ? "age" : "lang";
```
kenhuuu, line 405: This seems very specific. Is this limitation documented somewhere?

neo4j-gremlin/pom.xml
spmallette: oops - that's more than just a nit....thanks for noticing that.

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/features/StepDefinition.java
spmallette: added note to dev docs

xiazcy:
VOTE +1

spmallette:
merged with: 45bf1ba3693eba9308bc4a5d978bb1fea9c247f7

PR #3098, opened 2025-04-21
TINKERPOP-2974 Change PropertyMap to only allow single by() if order not guaranteed

kenhuuu:
https://issues.apache.org/jira/browse/TINKERPOP-2974

Not all implementations of Gremlin are going to be able to iterate the map produced by PropertyMap step in order. This means that multiple by()s can't be applied in order. Therefore, for those implementations, an exception should be thrown instead.


docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -1615,3 +1615,41 @@ Null values from the incoming traverser are not processed and remain as null whe
 See: link:https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/TrimGlobalStep.java[source],
 link:https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/TrimLocalStep.java[source (local)],
 link:https://tinkerpop.apache.org/docs/x.y.z/reference/#trim-step[reference]
+
+[[valueMap-step]]
+=== valueMap()
+
+*Description:* Removes repeatedly seen results from the Traversal Stream.
```
Cole-Greer, line 1622: Wrong description

xiazcy:
VOTE +1

PR #3099, opened 2025-04-23
TINKERPOP-3055 Fixed issues with with(out)Strategies for bytecode

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-3055

Changes to strategy/class serialization in GraphSON v2/v3 in support of this feature. Required changes to python, .net, javascript. Go basically worked because it only supported GraphBinary and didn't have GraphSON support.

VOTE +1


docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1396,6 +1396,29 @@ can be passed in the constructor of a new `Client` or `DriverRemoteConnection` :
...
+// use an instance for withStrategies()
+g.withStrategies(new InlineFilterStrategy);
+
+// if the strategies has arguments, specify them as named arguments
+g.withStrategies(new SeedStrategy({seed: 99999}))
+----
+
+NOTE: Many of the `TraversalStrategy` classes in Gremlin-Javascript are proxies to the respective strategy on
+Apache TinkerPop's JVM-based Gremlin traversal machine. As such, their `apply(Traversal)` method does nothing. However,
+the strategy is encoded in the Gremlin-Python bytecode and transmitted to the Gremlin traversal machine for
```
andreachild, line 1419: Do you mean the Gremlin-Javascript bytecode?

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1396,6 +1396,29 @@ can be passed in the constructor of a new `Client` or `DriverRemoteConnection` :
...
+[[gremlin-javascript-strategies]]
+=== Traversal Strategies
+
+In order to add and remove <<traversalstrategy,traversal strategies>> from a traversal source, Gremlin-Javascript has a
+`TraversalStrategy` class along with a collection of subclasses that mirror the standard Gremlin-Java strategies.
+
+[source,javascript]
+----
+// use the class name for the strategy directly for withoutStrategies()
+g.withoutStrategies(InlineFilterStrategy);
```
andreachild, line 1408: For consistency would it be better to also require an instance for `withoutStrategies` but then downstream propagate just the class name of the provided instance?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/decoration/HaltedTraverserStrategy.java
```diff
@@ -19,6 +19,7 @@
 
 package org.apache.tinkerpop.gremlin.process.traversal.strategy.decoration;
 
+import org.apache.commons.configuration2.BaseConfiguration;
 import org.apache.commons.configuration2.Configuration;
 import org.apache.commons.configuration2.MapConfiguration;
```
andreachild, line 24: There's unused imports after the configuration change.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/decoration/OptionsStrategy.java
```diff
@@ -63,7 +63,7 @@ public Configuration getConfiguration() {
 
     @Override
     public void apply(final Traversal.Admin<?, ?> traversal) {
-        // has not effect on the traversal itself - simply carries a options with it that individual steps
+        // has not affect on the traversal itself - simply carries a options with it that individual steps
```
andreachild, line 66: Nit: should say 'has no effect'

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/decoration/HaltedTraverserStrategy.java
```diff
@@ -75,10 +77,10 @@ public static HaltedTraverserStrategy create(final Configuration configuration)
 
     @Override
     public Configuration getConfiguration() {
-        final Map<String, Object> map = new HashMap<>();
-        map.put(STRATEGY, HaltedTraverserStrategy.class.getCanonicalName());
-        map.put(HALTED_TRAVERSER_FACTORY, this.haltedTraverserFactory.getCanonicalName());
-        return new MapConfiguration(map);
```
andreachild, line 81: Was the usage of `MapConfiguration` causing serialization errors? If so there's still other strategies using `MapConfiguration` - do those need to also be changed to use `BaseConfiguration`?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/finalization/MatchAlgorithmStrategy.java
```diff
@@ -18,6 +18,7 @@
  */
 package org.apache.tinkerpop.gremlin.process.traversal.strategy.finalization;
 
+import org.apache.commons.configuration2.BaseConfiguration;
 import org.apache.commons.configuration2.Configuration;
 import org.apache.commons.configuration2.MapConfiguration;
```
andreachild, line 23: There are now unused imports

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/decoration/HaltedTraverserStrategy.java
```diff
@@ -75,10 +77,10 @@ public static HaltedTraverserStrategy create(final Configuration configuration)
...
+        final BaseConfiguration conf = new BaseConfiguration();
+        conf.setListDelimiterHandler(GremlinDisabledListDelimiterHandler.instance());
+        conf.setProperty(HALTED_TRAVERSER_FACTORY, this.haltedTraverserFactory.getCanonicalName());
```
andreachild, line 82: Javadoc for this method indicates this config is used to pass to create method. Could be useful to have a unit test to ensure round trip of configuration passed to create works as expected:

```
HaltedTraverserStrategyTest.java

@Test
public void shouldRoundTripGetConfigurationAndCreate() {
    final HaltedTraverserStrategy originalStrategy = HaltedTraverserStrategy.create(new MapConfiguration(new HashMap<String, Object>() {{
        put(HaltedTraverserStrategy.HALTED_TRAVERSER_FACTORY, DetachedFactory.class.getCanonicalName());
    }}));
    assertEquals(originalStrategy, HaltedTraverserStrategy.create(originalStrategy.getConfiguration()));
}
```

xiazcy:
VOTE +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/decoration/HaltedTraverserStrategy.java
andreachild, line 82: Nit: it's not a problem since both factory types are not inner classes but it would be best to use `this.haltedTraverserFactory.getName()` instead of `getCanonicalName()` as the canonical name is the human readable class name while the name is the JVM representation of the class name, which can differ for inner classes. So if the factory classes are moved to inner classes or another factory class is created which is an inner class, Class.forName() will not work with the canonical name.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/TraversalSerializersV2.java
```diff
@@ -19,6 +19,7 @@
 
 package org.apache.tinkerpop.gremlin.structure.io.graphson;
 
+import org.apache.commons.configuration2.BaseConfiguration;
 import org.apache.commons.configuration2.ConfigurationConverter;
 import org.apache.commons.configuration2.MapConfiguration;
```
andreachild, line 24: Now unused

docs/src/reference/gremlin-variants.asciidoc
spmallette: not sure. i'm not sure how new users think about strategies. like to me, knowing how TinkerPop works, i wouldn't find that intuitive at all. but...maybe some folks would would wonder why in one case you need the type but others the instance. anyway, i'd prefer not to tackle that question here. it could be introduced as a non-breaking change at any point technically so if you'd like to propose it as something separate from this change, it could be discussed.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/decoration/HaltedTraverserStrategy.java
spmallette: yeah, that's more than a nit actually. i missed fixing that here after fixing it other places.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/decoration/HaltedTraverserStrategy.java
spmallette: At first I thought there might have been some issue there, but I don't think there is. There's really no need for `MapConfiguration` that i can think of. No idea why we first created a `Map` just to throw it away into a `MapConfiguration`. Since it just seemed better, I left it. I guess I can change them all to be consistent.

docs/src/reference/gremlin-variants.asciidoc
andreachild: I agree it makes more sense to just pass the strategy name to `withoutStrategies`. Ideally there would be an overload allowing `withStrategies` to be passed just the name for consistency. But that's out of scope for this PR.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/decoration/HaltedTraverserStrategy.java
andreachild: Consistency would be nice but I was more concerned that if MapConfiguration was causing serialization errors then others would need to be changed as well. Since this is not an issue, this comment becomes just a nit.

gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphSON/TraversalStrategySerializer.cs
```diff
@@ -31,7 +31,12 @@ internal class TraversalStrategySerializer : IGraphSONSerializer
         public Dictionary<string, dynamic> Dictify(dynamic objectData, GraphSONWriter writer)
         {
             AbstractTraversalStrategy strategy = objectData;
-            return GraphSONUtil.ToTypedValue(strategy.StrategyName, writer.ToDict(strategy.Configuration));
+            var stratData = new Dictionary<string, dynamic>
```
andreachild, line 34: I would expect a similar change to be required for deserialization of a `TraversalStrategy` but I can't find it in this PR. Am I missing something?

xiazcy: Strategies are serialized by GLV to be applied on the sever, and only results are returned, so GLVs don't need to deserialize strategies.

andreachild: Ah yes that makes sense, thanks

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/process/traversal-strategy.js
```diff
@@ -239,6 +276,16 @@ class PathRetractionStrategy extends TraversalStrategy {
   }
 }
 
+class ProductiveByStrategy extends TraversalStrategy {
+  /**
+   * @param {Object} [options]
+   * @param {Array<String>} [options.productiveKeys] set of keys that will always be productive
+   */
+  constructor(options) {
+    super('org.apache.tinkerpop.gremlin.process.traversal.strategy.optimization.ProductiveByStrategy', options);
```
andreachild, line 285: I'm curious why in this case the full `options` are passed through to the super constructor but for other strategies like `SeedStrategy` only specific options are passed through?

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1396,6 +1396,29 @@ can be passed in the constructor of a new `Client` or `DriverRemoteConnection` :
...
+
+// use an instance for withStrategies()
+g.withStrategies(new InlineFilterStrategy);
+
+// if the strategies has arguments, specify them as named arguments
+g.withStrategies(new SeedStrategy({seed: 99999}))
+----
+
+NOTE: Many of the `TraversalStrategy` classes in Gremlin-Javascript are proxies to the respective strategy on
+Apache TinkerPop's JVM-based Gremlin traversal machine. As such, their `apply(Traversal)` method does nothing. However,
```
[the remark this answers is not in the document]
spmallette: no - doesn't need to exist right now. could probably be removed: https://issues.apache.org/jira/browse/TINKERPOP-3157

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONModule.java
```diff
@@ -177,6 +178,7 @@ static final class GraphSONModuleV3 extends GraphSONModule {
                     put(OrP.class, "P");
                     put(P.class, "P");
                     put(TextP.class, "TextP");
+                    put(TraversalStrategyProxy.class, "TraversalStrategy");
```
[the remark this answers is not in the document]
spmallette: yes, i agree with all that. i figured that 4.x will wholly resolve this mess, so 3.x may just have to be a little ugly.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/process/traversal-strategy.js
spmallette: missed changing that one. good find.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/io/binary/internals/ClassSerializer.js
```diff
@@ -0,0 +1,63 @@
...
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+/**
+ * @author Igor Ostapenko
```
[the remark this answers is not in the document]
spmallette: shouldn't use author tags in general for new code - dropped it.

andreachild:
VOTE +1

Cole-Greer:
VOTE +1

PR #3100, opened 2025-04-24
TINKERPOP-3023 - Update Vertex structure in grammar

xiazcy:
Modified Vertex in grammar to only require an `id` with `label` as optional, i.e. `Vertex(id)`, removed the support for `new` and `ReferenceVertex`. Part of TINKERPOP-3023.

VOTE +1


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/StructureElementVisitor.java
```diff
@@ -32,7 +32,11 @@ public StructureElementVisitor(final GremlinAntlrToJava antlr) {
 
     @Override
     public Vertex visitStructureVertex(final GremlinParser.StructureVertexContext ctx) {
-        return new ReferenceVertex(antlr.argumentVisitor.visitGenericLiteralArgument(ctx.genericLiteralArgument()),
-                (String) antlr.argumentVisitor.visitStringArgument(ctx.stringArgument()));
+        if (ctx.getChildCount() == 4) {
```
Cole-Greer, line 35: Nit: Prefer to use `if (ctx.stringArgument() == null)` instead of checking the child count.

xiazcy:
Closing this PR based on [devlist post](https://lists.apache.org/thread/ptq81160z4y4vddg4lxz2nv244x2yzhq), we'll be removing Vertex structure itself from the Grammar.

PR #3101, opened 2025-04-25
[TINKERPOP-3143] Updated gremlin-javascript to default to GraphBinary instead of GraphSONv3

flora-jin:
Fixed gremlin-javascript to default to GraphBinary serializer instead of GraphSONv3.

https://issues.apache.org/jira/browse/TINKERPOP-3143


gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
```diff
@@ -39,7 +39,7 @@ const responseStatusCode = {
   authenticationChallenge: 407,
 };
 
-const defaultMimeType = 'application/vnd.gremlin-v3.0+json';
+const defaultMimeType = 'application/vnd.graphbinary-v1.0';
```
andreachild, line 42: I feel like this could be unit tested, maybe in the `client-test.js` unit test?

andreachild: It can be done like this:
```
it('should use default mimeType', function () {
  const customClient = new Client('ws://localhost:9321', {traversalSource: 'g', connectOnStartup: false});
  assert.strictEqual(customClient._connection.mimeType, 'application/vnd.graphbinary-v1.0')
});

it('should use given mimeType', function () {
  const customClient = new Client('ws://localhost:9321', {traversalSource: 'g', connectOnStartup: false, mimeType: 'application/vnd.gremlin-v2.0+json'});
  assert.strictEqual(customClient._connection.mimeType, 'application/vnd.gremlin-v2.0+json')
});
```

andreachild:
VOTE +1

Cole-Greer:
VOTE +1, merging as CTR

PR #3102, opened 2025-04-25
Prevent multiple by modulators for sack step

andreachild:
https://issues.apache.org/jira/browse/TINKERPOP-3149

Prevent multiple by operators from being used for the sack count step by throwing IllegalStateException if modulateBy is called more than once.


xiazcy:
VOTE +1

Cole-Greer:
VOTE +1

PR #3104, opened 2025-04-29
TINKERPOP-3023  (Part of) Implement UUID which should have two forms: UUID() and UUID(“uuid”)

EchoNullify:
TARGET 3.8-dev

https://issues.apache.org/jira/browse/TINKERPOP-3023

Part of TINKERPOP-3023. Implement UUID which should have two forms: UUID() and UUID(“uuid”).


gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/language/translator/GremlinTranslatorTest.java
```diff
@@ -151,6 +151,15 @@ public static Collection<Object[]> data() {
                             null,
                             "g.with_(\"x\")",
                             "g.with_('x')"},
+                    {"g.inject(UUID(\"f47ac10b-58cc-4372-a567-0e02b2c3d479\"))",
```
Cole-Greer, line 154: It would be good to add equivalent translation tests for something like `"g.inject(UUID())"` to ensure it properly translates to a new random UUID in each of the GLVs (something like `"g.Inject<object>(Guid.NewGuid())"` in .NET)

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/language/translator/GremlinTranslatorTest.java
```diff
@@ -151,6 +151,15 @@ public static Collection<Object[]> data() {
...
+                            null,
+                            "g.inject(string0)",
+                            "g.Inject<object>(Guid.Parse(\"f47ac10b-58cc-4372-a567-0e02b2c3d479\"))",
+                            "g.Inject(uuid.Parse(\"f47ac10b-58cc-4372-a567-0e02b2c3d479\"))",
+                            null,
+                            null,
+                            "g.inject(\"f47ac10b-58cc-4372-a567-0e02b2c3d479\")",
```
Cole-Greer, line 161: I don't think we want to translate to a simple string for JS. It should either use [utils.js](https://github.com/apache/tinkerpop/blob/5b7eea59a4264f06cc05a76383db2eda480379c4/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/utils.js#L41-L44) (might need an extension there) or directly construct a uuid using the same uuid library.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/sideEffect/Uuid.feature
```diff
@@ -0,0 +1,30 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+@StepClassSideEffect @StepUuid
+Feature: Step - uuid()
```
xiazcy, line 19: Given uuid() isn't an actual step, please add this test into the `Inject.feature` file instead of creating this separate one that doesn't make sense.

PR #3105, opened 2025-04-29
TINKERPOP-3148 Prevent multiple by modulators from being used for dedup step

andreachild:
https://issues.apache.org/jira/browse/TINKERPOP-3148

Prevent dedup step from having multiple by modulators by throwing `IllegalStateException` if `modulateBy` is called more than once.


xiazcy:
VOTE +1

Cole-Greer:
VOTE +1

PR #3114, opened 2025-05-02
TINKERPOP-3150 Prevent sample step from having multiple by modulators…

EchoNullify:
TARGET 3.8-dev

https://issues.apache.org/jira/browse/TINKERPOP-3150

Prevent multiple by operators from being used for sample step by throwing IllegalStateException if modulateBy is called more than once.


Cole-Greer:
Reminder: Update PR target to 3.8-dev

Cole-Greer:
Looks like something in `ByModulatorOptimizationStrategyTest` is failing. It appears something in that test is using `sample().by().by()` and needs to be corrected or removed. That's probably an issue with the outdated test and not any issue with your change itself.

VOTE +1 pending test fix.

xiazcy:
VOTE +1

PR #3117, opened 2025-05-07
Promote to next highest number type to prevent sum step overflow

EchoNullify:
Promote to next highest number type to prevent sum step from overflowing


tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/TinkerGraphFeatureTest.java
```diff
@@ -30,7 +30,7 @@
 
 @RunWith(Cucumber.class)
 @CucumberOptions(
-        tags = "not @GraphComputerOnly and not @AllowNullPropertyValues",
```
andreachild: Why was this change necessary?

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/Sum.feature
```diff
@@ -18,6 +18,17 @@
 @StepClassMap @StepSum
 Feature: Step - sum()
 
+  Scenario: g_V_sum_overflow_int
+    Given the modern graph
+    And the traversal of
+      """
+      g.inject(2147483647, 1).sum()
```
andreachild, line 25: Please add more tests for different combinations of input types

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
```diff
@@ -361,8 +377,18 @@ public static Class<? extends Number> getHighestCommonNumberClass(final boolean
      */
     public static Number add(final Number a, final Number b) {
         if (null == a || null == b) return a;
-        final Class<? extends Number> clazz = getHighestCommonNumberClass(a, b);
-        return getHelper(clazz).add.apply(a, b);
+        NumberInfo numberInfo = getHighestCommonNumberInfo(false, a, b);
```
andreachild, line 380: Please add unit tests for this new behaviour in `NumberHelperTest.java`

tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/TinkerGraphFeatureTest.java
EchoNullify: it wasn't and you was too quick to see my rollback :)

andreachild:
Please add changelog entry

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
```diff
@@ -361,8 +377,18 @@ public static Class<? extends Number> getHighestCommonNumberClass(final boolean
...
+        while (true) {
+            try {
+                final Class<? extends Number> clazz = determineNumberClass(numberInfo.bits, numberInfo.fp);
```
andreachild, line 383: Javadoc for this method should be updated to describe type promotion on overflow

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
```diff
@@ -361,8 +377,18 @@ public static Class<? extends Number> getHighestCommonNumberClass(final boolean
...
+                return getHelper(clazz).add.apply(a, b);
+            } catch (ArithmeticException exception) {
+                if (numberInfo.bits == 128) {
```
andreachild, line 386: Nit: consider changing the check to `>=128` in case there is a bug introduced somewhere in the numberInfo.bits calculation, to avoid potential endless loop.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
```diff
@@ -361,8 +377,18 @@ public static Class<? extends Number> getHighestCommonNumberClass(final boolean
...
+                    throw exception;
+                }
+                numberInfo.bits <<= 1;
```
andreachild, line 389: Would this result in the following type promotions?
* Long -> BigInteger
* Double -> BigDecimal

From previous discussions we had thought to avoid auto promotion to BigInteger or BigDecimal as those types are 'extended' types and GLVs do not necessarily support them.

https://tinkerpop.apache.org/docs/3.7.2/dev/io/#_extended

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
```diff
@@ -28,6 +28,17 @@
  */
 public final class NumberHelper {
 
+    private static class NumberInfo {
+
+        int bits;
```
andreachild, line 33: Nit: there's code convention to add final and private where possible so this class itself can be final and the fields can be private final (will need getters as well).

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
```diff
@@ -361,8 +386,26 @@ public static Class<? extends Number> getHighestCommonNumberClass(final boolean
...
-        return getHelper(clazz).add.apply(a, b);
+        NumberInfo numberInfo = getHighestCommonNumberInfo(false, a, b);
+        Number result = 0;
+        while (true) {
+            try {
+                final Class<? extends Number> clazz = determineNumberClass(numberInfo.bits, numberInfo.fp);
+                result = getHelper(clazz).add.apply(a, b);
+                if (Double.isInfinite(result.doubleValue()))
+                {
+                    throw new ArithmeticException("Floating point overflow detected");
```
Cole-Greer, line 397: I don't think we want an exception in this case, infinity is an acceptable sum result for floating point numbers.

andreachild, line 397: We should not be throwing exception on double infinity

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/util/NumberHelperTest.java
```diff
@@ -460,22 +484,68 @@ public void testMinMaxCompare() {
...
+        Number value = mul(Double.MAX_VALUE, 2F);
+        assertTrue(value instanceof Double);
+        assertTrue(Double.isInfinite(value.doubleValue()));
+    }
+
     @Test
     public void shouldThrowArithmeticExceptionOnOverflow() {
-        for (final Triplet<Number, Number, String> q : OVERFLOW_CASES) {
+        for (final Quartet<Number, Number, String, Boolean> q : OVERFLOW_CASES) {
+            if (!q.getValue3()) {
```
andreachild, line 518: From what I can tell, having a `q=false` in the test data set essentially is the same as removing the entire quartet value from the data set as this check will just skip execution of the test case.

If my understanding is correct then I would recommend separating the test cases that have `q=false` from the `OVERFLOW_CASES` data set into their own unit test as those entries are not actually executed.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/util/NumberHelperTest.java
```diff
@@ -460,22 +484,68 @@ public void testMinMaxCompare() {
...
+    @Test
+    public void shouldNotThrowArithmeticExceptionOnOverflow() {
+        for (final Quartet<Number, Number, String, Boolean> q : OVERFLOW_CASES) {
+            if (q.getValue3()) {
+                continue;
+            }
             try {
                 switch (q.getValue2()) {
                     case "add":
                         add(q.getValue0(), q.getValue1());
```
andreachild, line 542: We should be verifying the expected result type as well.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
```diff
@@ -312,43 +356,80 @@ private NumberHelper(final BiFunction<Number, Number, Number> add,
...
+        while (true) {
+            try {
+                result = mathOperation.apply(numberInfo, a, b);
+                if (Double.isInfinite(result.doubleValue()))
+                {
+                    throw new ArithmeticException("Floating point overflow detected");
+                }
+                return result;
+            } catch (ArithmeticException exception) {
+                if (!numberInfo.fp && numberInfo.bits >= 64) {
```
vkagamlyk, line 411: can the result of multiplying two Long values be promoted to BigInteger? It would be good to add tests like this

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -331,6 +335,10 @@ link:https://tinkerpop.apache.org/docs/3.8.0/dev/provider/#gherkin-tests-suite[P
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-3136[TINKERPOP-3136]
 
+==== Auto promotion of number types
+
+Any math operations such as Add, Sub, Mul & Div will auto matically promote to next number type if an overflow is detected. For integers the promotion sequence is byte->short->integer->long->overflow. For floating point the sequence is float->double->ininity.
```
Cole-Greer, line 340: Nit: 
```suggestion
Any math operations such as Add, Sub, Mul & Div will automatically promote to next number type if an overflow is detected. For integers the promotion sequence is byte->short->integer->long->overflow. For floating point the sequence is float->double->infinity.
```

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -30,6 +30,10 @@ complete list of all the modifications that are part of this release.
 
 === Upgrading for Users
 
+==== Auto promotion of number types
+
+Any math operations such as Add, Sub, Mul & Div will auto matically promote to next number type if an overflow is detected. For integers the promotion sequence is byte->short->integer->long->overflow. For floating point the sequence is float->double->ininity.
```
Cole-Greer, line 35: Nit:
```suggestion
Any math operations such as Add, Sub, Mul & Div will automatically promote to next number type if an overflow is detected. For integers the promotion sequence is byte->short->integer->long->overflow. For floating point the sequence is float->double->infinity.
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
```diff
@@ -312,43 +356,84 @@ private NumberHelper(final BiFunction<Number, Number, Number> add,
...
     }
 
     /**
      * Adds two numbers returning the highest common number class between them.
      *
+     * <p>
+     * This method returns a result using the highest common number class between the two inputs.
+     * If an overflow occurs (either integer or floating-point), the method promotes the precision
+     * by increasing the bit width or switching to floating-point arithmetic, until a suitable type is found.
+     * If no suitable type exists (e.g., for very large integers beyond 128-bit),
```
Cole-Greer, line 432:
```suggestion
     * by increasing the bit width until a suitable type is found.
     * If no suitable type exists (e.g., for very large integers beyond 64-bit),
```
This PR no longer promotes from integers to floating point, or promotes past `long`. Same applies to the following 3 equivalent comments.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/util/NumberHelperTest.java
```diff
@@ -99,20 +114,36 @@ public class NumberHelperTest {
...
+            new Quartet<>(Byte.MAX_VALUE, 0.5f, "div", "f"),
+            new Quartet<>(Short.MAX_VALUE, (short)100, "add", "i"),
+            new Quartet<>(Short.MIN_VALUE, (short)100, "sub", "i"),
+            new Quartet<>(Short.MAX_VALUE, (short)100, "mul", "i"),
+            new Quartet<>(Short.MAX_VALUE, 0.5f, "div", "f"),
+            new Quartet<>(Integer.MAX_VALUE, 1, "add", "l"),
+            new Quartet<>(Integer.MIN_VALUE, 1, "sub", "l"),
+            new Quartet<>(Integer.MAX_VALUE, Integer.MAX_VALUE, "mul", "l"),
+            new Quartet<>(Integer.MAX_VALUE, 0.5f, "div", "f"),
+            new Quartet<>(Long.MAX_VALUE, 0.5f, "div", "d")
```
Cole-Greer, line 135: It's interesting that a combination of a float and a long is promoted to double. I would've expected a float as this is how Java handles something such as `50L + 5.5f`. I see this is pre-existing behaviour which is already asserted in the previous tests, so I suppose it is ok to keep it as is here.

Cole-Greer (commented):
VOTE +1 (Pending docs and comment fixes)

PR #3118, opened 2025-05-09
getScopingInfo utility added

vaibhavm99:
Added a utility function `getScopingInfo()` in all classes implementing `Scoping` and a method `getScopingInfoForTraversal` to get the labels needed by a traversal, which returns a Set of `ScopingInfo` class which contains the label and Pop info.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -30,6 +30,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 [[release-4-0-0-beta-1]]
 === TinkerPop 4.0.0-beta.1 (January 17, 2025)
 
+* Added the `PopContaining` interface designed to get label and `Pop` combinations held in a `PopInstruction` object.
 * Added support for deserialization of `Set` for `gremlin-javascript`.
 * Added grammar-based `Translator` for all languages including explicit ones for Java and anonymization.
 * Removed old `Translator` infrastructure.
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/PopContaining.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/PopContaining.java
@@ -1,0 +1,103 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.apache.tinkerpop.gremlin.process.traversal.step;
+
+import java.util.Objects;
+import java.util.Set;
+
+import org.apache.tinkerpop.gremlin.process.traversal.Pop;
+
+/**
+ * The {@code PopContaining} interface is implemented by traversal steps that maintain Pop instructions
+ * for label access. It provides a mechanism to track and manage how labeled elements should
+ * be accessed using {@link Pop} semantics (first, last, all, or mixed).
+ *
+ * In Gremlin traversals, various elements can be labeled and later accessed via these labels.
+ * The {@link Pop} enum determines how to access these labeled elements when there are multiple
+ * values associated with the same label.
+ *
+ * <pre>
+ * {@code
+ * // Simple example with default Pop.last behavior
+ * gremlin> g.V().as("a").out().as("a").select("a")
+ * ==>[v[2]]  // returns the last element labeled "a"
+ *
+ * // Using Pop.first to get the first labeled element
+ * gremlin> g.V().as("a").out().as("a").select(first, "a")
+ * ==>[v[1]]  // returns the first element labeled "a"
+ *
+ * // Using Pop.all to get all labeled elements
+ * gremlin> g.V().as("a").out().as("a").select(all, "a")
+ * ==>[v[1], v[2]]  // returns all elements labeled "a"
+ * }
+ * </pre>
+ *
+ * Steps implementing this interface maintain a collection of {@link PopInstruction} objects, each containing
+ * a label and a {@link Pop} value that specifies how to access elements with that label.
+ *
+ */
+public interface PopContaining {
+    public Set<PopInstruction> getPopInstructions();
+    /**
+     * A class for storing the Scope Context. It has two elements:
+     * - label: String
+     * - pop: Pop value
+     */
+    class PopInstruction {
+        private final Pop pop;
+        private final String label;
+
+        public PopInstruction(String label, Pop pop) {
+            this.pop = pop;
+            this.label = label;
+        }
+
+        public PopInstruction(Pop pop, String label) {
+            this.pop = pop;
+            this.label = label;
+        }
+
+        public PopInstruction() {
+           this.pop = null;
+           this.label = "";
+        }
+
+    public String getLabel() {
+            return label;
+        }
+
+        public Pop getPop() {
+            return pop;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+            final PopInstruction that = (PopInstruction) o;
+            return getPop() == that.getPop() && Objects.equals(getLabel(), that.getLabel());
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(getPop(), getLabel());
+        }
+    }
+}
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/PopContaining.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/PopContaining.java
@@ -68,7 +68,7 @@ public interface PopContaining {
             this.label = label;
         }
 
-        public PopInstruction(Pop pop, String label) {
+        public PopInstruction(final Pop pop, final String label) {
             this.pop = pop;
             this.label = label;
         }
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/Scoping.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/Scoping.java
@@ -23,6 +23,7 @@ import org.apache.tinkerpop.gremlin.process.traversal.Pop;
 import org.apache.tinkerpop.gremlin.process.traversal.Step;
 import org.apache.tinkerpop.gremlin.process.traversal.Traverser;
 
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
@@ -105,7 +106,7 @@ import java.util.Set;
  * @author Marko A. Rodriguez (http://markorodriguez.com)
  * @author Stephen Mallette (http://stephen.genoprime.com)
  */
-public interface Scoping {
+public interface Scoping extends PopContaining {
 
     public enum Variable {START, END}
 
@@ -165,6 +166,23 @@ public interface Scoping {
      * @return the accessed labels of the scoping step
      */
     public Set<String> getScopeKeys();
+
+    /**
+     * Used to get PopInstruction of a Step that implements Scoping. PopInstruction includes the labels it needs, and the
+     * pop type for each label.
+     *
+     * @return A Set of {@link PopInstruction} values which contain the label and Pop value
+     */
+    @Override
+    public default HashSet<PopInstruction> getPopInstructions() {
+        final Set<String> labels = this.getScopeKeys();
+        final HashSet<PopInstruction> scopingInfoSet = new HashSet<>();
+        for (final String label : labels) {
+            final PopInstruction scopingInfo = new PopInstruction(Pop.last, label);
+            scopingInfoSet.add(scopingInfo);
+        }
+        return scopingInfoSet;
+    }
 
     public static class KeyNotFoundException extends Exception {
 
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/FormatStep.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/FormatStep.java
@@ -34,6 +34,7 @@ import org.apache.tinkerpop.gremlin.structure.Element;
 import org.apache.tinkerpop.gremlin.structure.Property;
 import org.apache.tinkerpop.gremlin.structure.util.StringFactory;
 
+import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Objects;
@@ -166,6 +167,14 @@ public final class FormatStep<S> extends MapStep<S, String> implements ByModulat
     @Override
     public Set<String> getScopeKeys() {
         return variables;
+    }
+
+    @Override
+    public HashSet<PopInstruction> getPopInstructions() {
+        final HashSet<PopInstruction> popInstructions = new HashSet<>();
+        popInstructions.addAll(Scoping.super.getPopInstructions());
+        popInstructions.addAll(TraversalParent.super.getPopInstructions());
+        return popInstructions;
     }
 
     @Override
--- a/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/TestDataBuilder.java
+++ b/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/TestDataBuilder.java
@@ -31,7 +31,7 @@ import org.apache.tinkerpop.gremlin.process.traversal.step.PopContaining;
 public class TestDataBuilder {
 
     // Helper function to create a Set of `PopInstruction` values
-    public static HashSet<PopContaining.PopInstruction> createPopInstructionSet(Object[]... pairs) {
+    public static HashSet<PopContaining.PopInstruction> createPopInstructionSet(final Object[]... pairs) {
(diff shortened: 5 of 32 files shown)


Cole-Greer:
Hi @vaibhavm99, thanks for opening a PR! I should have time to give it a full review by Monday at the latest. Could I ask a bit about the motivation for the new `getScopingInfo()` method? The label functionality is mostly covered by `getScopeKeys()`, is the main objective to give some common interface for finding the Pop value of "select" type steps? Is this change in response to any difficulties with the current interface?

vkagamlyk:
Hi @vaibhavm99, will this change allow to save not the whole Path, but only the parts that are actually used?

vaibhavm99:
> Hi @vaibhavm99, thanks for opening a PR! I should have time to give it a full review by Monday at the latest. Could I ask a bit about the motivation for the new `getScopingInfo()` method? The label functionality is mostly covered by `getScopeKeys()`, is the main objective to give some common interface for finding the Pop value of "select" type steps? Is this change in response to any difficulties with the current interface?

`getScopeKeys` only gives us a List of labels needed, but does not contain the `Pop` info. This is a common interface to get both `label` and the `Pop` value for all the labels needed. This is to provide label information inside the repeat step during the Neptune FastTranslation.

vaibhavm99:
> Hi @vaibhavm99, will this change allow to save not the whole Path, but only the parts that are actually used?

Hi @vkagamlyk , I'm sorry I don't understand what you mean by "the whole Path." This change should not have any breaking problems with older code, as it is just adding a new API. This would take in a traversal and provide us with the `label` and `Pop` info for the labels that are needed in that traversal.

vkagamlyk:
> > Hi @vaibhavm99, will this change allow to save not the whole Path, but only the parts that are actually used?
> 
> Hi @vkagamlyk , I'm sorry I don't understand what you mean by "the whole Path." This change should not have any breaking problems with older code, as it is just adding a new API. This would take in a traversal and provide us with the `label` and `Pop` info for the labels that are needed in that traversal.

Hi @vaibhavm99!
In some cases it is necessary to transfer Traversers between servers. Part of Traverser can be Path object that is larger than other Traverser parts. If we know in advance which parts of the path are used, then we can transfer only them.
Path is list of pairs <label, value>, `getScopeKeys` can be used to get list of used labels.

Cole-Greer:
Hi @vaibhavm99 I'm happy with extending convenience interfaces for providers. I am curious since the `SelectStep`, `SelectOneStep`, and `TraversalSelectStep` are the only steps which are not hardcoded to `Pop.last`, if you also considered unifying these steps with some sort of `Selecting` interface which could extend `Scoping` and add `public Pop getPop();`

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/FormatStep.java
```diff
@@ -168,6 +169,19 @@ public Set<String> getScopeKeys() {
         return variables;
     }
 
+    @Override
+    public Set<ScopingInfo> getScopingInfo() {
```
Cole-Greer, line 173: As each step contains a Set of labels, but only one Pop, would it be easier for ScopingInfo to be defined as
```
public static class ScopingInfo {
    public Set<String> labels;
    public Pop pop;
```
and for this method to return `ScopingInfo` instead of `Set<ScopingInfo>`

Cole-Greer:
I see this PR is currently targeting the master branch, this branch is currently accumulating changes for TinkerPop 4.0.0-beta.2, and ultimately TinkerPop 4. I would recommend targeting this change to `3.7-dev`, which is accumulating changes for 3.7.4 (non-breaking only), or `3.8-dev` which is for TinkerPop 3.8.0 (breaking changes permitted). Any changes introduced to an older development branch will also get introduced to the newer branches when the PR is merged.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/Scoping.java
```diff
@@ -185,4 +193,39 @@ public Scoping getStep() {
...
         }
     }
+
+
+    /**
+     * A class for storing the Scoping information. It has two elements:
+     * - label: String
+     * - pop: Pop value
+     */
+    public static class ScopingInfo {
```
spmallette, line 203: Just a comment on naming - we have `getScopeKeys` so I feel like we should leave the "ing" suffix for the interface and the non-active "Scope" for the things the interface exposes. Perhaps we call this `ScopeInfo` or `ScopeContext` (I prefer the latter i think because we don't use "Info" anywhere in our code, but we do use "Context" in places) and have a `Scoping.getScopeContext()`?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/Scoping.java
```diff
@@ -166,6 +167,13 @@ public default <S> S getNullableScopeValue(final Pop pop, final String key, fina
      */
     public Set<String> getScopeKeys();
 
+    /**
+     * Used to get ScopingInfo of a Step that implements Scoping. Scoping info includes the labels it needs, and the pop type for each label.
+     *
+     * @return A Set of {@link ScopingInfo} values which contain the label and Pop value
+     */
+    public default Set<ScopingInfo> getScopingInfo() {return null; }
```
spmallette, line 175: Should there be a default implementation here? If something implements `Scoping` shouldn't it have a proper implementation? or are there cases where you would return `getScopeKeys` but not `getScopingInfo`? and if so, would `return null` be the right default?

vaibhavm99:
> > > Hi @vaibhavm99, will this change allow to save not the whole Path, but only the parts that are actually used?
> > 
> > 
> > Hi @vkagamlyk , I'm sorry I don't understand what you mean by "the whole Path." This change should not have any breaking problems with older code, as it is just adding a new API. This would take in a traversal and provide us with the `label` and `Pop` info for the labels that are needed in that traversal.
> 
> Hi @vaibhavm99! In some cases it is necessary to transfer Traversers between servers. Part of Traverser can be Path object that is larger than other Traverser parts. If we know in advance which parts of the path are used, then we can transfer only them. Path is list of pairs <label, value>, `getScopeKeys` can be used to get list of used labels.

This API would not cause any breaking change in `getScopeKeys`, this would give us the `Pop` info for those labels as well if you'd need it. When some labels are overloaded, `Pop` can be used to distinguish which mapping of that label is needed. For example, if only `first` is needed we would just need to have the <label, value1> instead of all the <label, value> pairs. This API gives additional info along with what `getScopeKeys` provides.

vaibhavm99:
> Hi @vaibhavm99 I'm happy with extending convenience interfaces for providers. I am curious since the `SelectStep`, `SelectOneStep`, and `TraversalSelectStep` are the only steps which are not hardcoded to `Pop.last`, if you also considered unifying these steps with some sort of `Selecting` interface which could extend `Scoping` and add `public Pop getPop();`

Yes, `getScopingInfo()` is overloaded in every class that implements `Scoping`, if that class hardcodes Pop to `Pop.last` that is given out, and if it gives the user option to select `Pop`, I used that value using the `getPop()` function.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/Scoping.java
spmallette: actually, building on `getScopeKeys`, i think that the primary accessor of `getScopePop` is missing here. that feels like the most natural implementation to me - a simple getter for the `Pop`, rather than a new composite object. cc/ @Cole-Greer

vaibhavm99: Understood, I'll make that change. Will change `getScopingInfo()` to `getScopeContext()`

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/Scoping.java
vaibhavm99: Yes, there are a few classes that do not implement `Pop` at all, so for them we do not need `getScopingInfo()` but they do have `getScopeKeys()`. For example, `AddEdgeStep` and `AddPropertyStep`. For those I added a default. Would changing it to return an empty Set be better?

upadhyay-prashant:
@vkagamlyk / @Cole-Greer ,

Thank you for the detailed review. You got it right, that we're working on optimizing label identification in prefix traversals, with the goal of only injecting necessary labels rather than the entire path. This approach could indeed be valuable for Traverser injection in TinkerPop.

You've made a good point about the current limitations of the Scoping interface, particularly regarding complete label information requirements. Let me explain why the current POJO structure might be more suitable than the proposed alternative:

Consider this example:
```
gremlin> g.V().as('a').
......1>   out().as('b', 'a').
......2>   in().as('b').
......3>   where(select(first, 'a').where(eq('b')))

==>v[1]
==>v[1]
==>v[1]
==>v[4]
==>v[4]
```
In this case, the outermost where step requires [['a',Pop.first], ['b',Pop.last]], which wouldn't fit well with the proposed structure:

```
public static class ScopingInfo {
    public Set<String> labels;
    public Pop pop;
}
```
@spmallette : I hope above example explains why an simple getPop API wont be enough. Let me know if you have better suggestions. 

Regarding the implementation strategy, our idea was to merge to master first and then cherry-pick to 3.7 and 3.8. Would you like to suggest any specific concerns about this approach?

Looking forward to your thoughts on this.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/Scoping.java
Cole-Greer: > actually, building on getScopeKeys, i think that the primary accessor of getScopePop is missing here.
I tend to agree that if a simple accessor for Pop is sufficient, that if preferable to adding a new structure. If we go down that path though, I would suggest `getPop()` instead of `getScopePop()` as that method already exists in the steps which take a user-defined Pop. If we prefer `getScopePop()`, then we will need to deprecate the old `getPop()` methods.

Cole-Greer:
> Regarding the implementation strategy, our idea was to merge to master first and then cherry-pick to 3.7 and 3.8. Would you like to suggest any specific concerns about this approach?

The typical approach for this in TinkerPop would be to target the initial PR to 3.7-dev. Whichever committer ends up merging the PR will then manually merge the changes up to 3.8-dev and master on your behalf. Typically the only time we ask contributors to open separate PRs for each branch is if each branch requires substantially different solutions for some reason.

xiazcy:
LGTM, thanks for the change! VOTE +1

Cole-Greer (commented):
Overall it looks good to me. Left a couple of comments for small tweaks. There are a couple of steps which give their own implementation for `getPopInstructions()` which do not have tests in this PR, which would be nice to have (TraversalSelectStep, AddVertexStep, AddVertexStartStep, AddEdgeStep, AddEdgeStartStep). It would also be nice to have some testing for steps which directly use the default implementations in Scpoing or TraversalParent.

For all of the steps which are TraversalParents, it would be nice if the tests also validated that `getPopInstructions()` is successfully producing the PopInstruction's from child traversals, as well as from the step itself.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/PopContaining.java
```diff
@@ -0,0 +1,104 @@
...
+ * // Using Pop.all to get all labeled elements
+ * gremlin> g.V().as("a").out().as("a").select(all, "a")
+ * ==>[v[1], v[2]]  // returns all elements labeled "a"
+ * }
+ * </pre>
+ *
+ * Steps implementing this interface maintain a collection of {@link PopInstruction} objects, each containing
+ * a label and a {@link Pop} value that specifies how to access elements with that label.
+ *
+ * @author Vaibhav Malhotra
```
spmallette: please don't "author" tags - any author tags currently in the code base are there mostly for legacy reasons. going forward new work doesn't include them. git is the best source of authorship.

CHANGELOG.asciidoc
```diff
@@ -30,6 +30,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 [[release-4-0-0-beta-1]]
 === TinkerPop 4.0.0-beta.1 (January 17, 2025)
 
+* Added an interface `PopContaining` which is implemented by `Scoping` and `TraversalParent`. Used to get label, and Pop info for a step/traversal. The information is wrapped in a `PopInstruction` object
```
spmallette: we tend to like to have a single sentence here. suggesting:

> Added the `PopContaining` interface designed to get label and `Pop` combinations held in a `PopInstruction` object.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/PopContaining.java
```diff
@@ -0,0 +1,104 @@
...
+ */
+public interface PopContaining {
+    public HashSet<PopInstruction> getPopInstructions();
+    /**
+     * A class for storing the Scope Context. It has two elements:
+     * - label: String
+     * - pop: Pop value
+     */
+    class PopInstruction {
+        public Pop pop;
```
spmallette: shouldn't these be `private final` access member variables?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/PopContaining.java
```diff
@@ -0,0 +1,104 @@
...
+        public String label;
+
+        public PopInstruction(String label, Pop pop) {
```
spmallette, line 66: nit: the form in TinkerPop code is to make all variables `final` where possible. please give a look over your changes here to see if you can catch any other places where this form might be used.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/Scoping.java
```diff
@@ -166,6 +167,23 @@ public default <S> S getNullableScopeValue(final Pop pop, final String key, fina
      */
     public Set<String> getScopeKeys();
 
+    /**
+     * Used to get PopInstruction of a Step that implements Scoping. PopInstruction includes the labels it needs, and the
+     * pop type for each label.
+     *
+     * @return A Set of {@link org.apache.tinkerpop.gremlin.process.traversal.step.PopContaining.PopInstruction} values which contain the label and Pop value
```
spmallette: nit: TinkerPop code format general prefers this reference being imported than being long form in the javadoc.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/PopContaining.java
```diff
@@ -0,0 +1,104 @@
...
+ * }
+ * </pre>
+ *
+ * Steps implementing this interface maintain a collection of {@link PopInstruction} objects, each containing
+ * a label and a {@link Pop} value that specifies how to access elements with that label.
+ *
+ * @author Vaibhav Malhotra
+ */
+public interface PopContaining {
+    public HashSet<PopInstruction> getPopInstructions();
```
spmallette: do we explicitly need to use the implementation `HashSet` here? if not, I would think `Set` is preferred.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/TraversalSelectStepTest.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.HashSet;
+
+import org.apache.tinkerpop.gremlin.TestDataBuilder;
+import org.apache.tinkerpop.gremlin.process.traversal.Pop;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.__;
+import org.apache.tinkerpop.gremlin.process.traversal.step.PopContaining;
+
+import junit.framework.TestCase;
+
+public class TraversalSelectStepTest extends TestCase {
```
spmallette: please prefer JUnit4 syntax with `@Test` annotation

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/SelectOneStepTest.java
```diff
@@ -61,4 +64,57 @@ public void shouldRequirePathsAccordingly() {
             assertEquals(traversalPath[0], ((Traversal.Admin<?, ?>) traversalPath[1]).getTraverserRequirements().contains(TraverserRequirement.LABELED_PATH));
         }
     }
+
+    @Test
+    public void testPopInstruction() {
```
spmallette: nit: note that TinkerPop tests prefer "should" prefix over "test" for test name, like "shouldObtainPopInstructions" would be a good name here. please take a look at other tests in the PR and make changes accordingly.

Cole-Greer:
Thanks for coming through and making all of these changes @vaibhavm99. My vote is VOTE +1.

Note that I intend to cherry pick this back to `3.7-dev` when merging as this is completely non-breaking and can probably deliver some value sooner from that branch.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/PopContaining.java
```diff
@@ -0,0 +1,103 @@
...
+    class PopInstruction {
+        private final Pop pop;
+        private final String label;
+
+        public PopInstruction(String label, Pop pop) {
+            this.pop = pop;
+            this.label = label;
+        }
+
+        public PopInstruction(Pop pop, String label) {
```
Cole-Greer:
```suggestion
        public PopInstruction(final Pop pop, final String label) {
```

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/TestDataBuilder.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * This class is responsible for building test data for `PopInstruction` Unit tests.
+ * It provides methods to create a set of `PopInstruction` test data objects.
+ *
+ */
+public class TestDataBuilder {
+
+    // Helper function to create a Set of `PopInstruction` values
+    public static HashSet<PopContaining.PopInstruction> createPopInstructionSet(Object[]... pairs) {
```
Cole-Greer:
```suggestion
    public static HashSet<PopContaining.PopInstruction> createPopInstructionSet(final Object[]... pairs) {
```

PR #3119, opened 2025-05-13
Upgrade to Go 1.24 and Node 22

Cole-Greer:
Go 1.20 and Node 20 have already past their end of life dates. Upgrading to Go 1.22 and Node 22.

VOTE +1


vkagamlyk:
VOTE +1

xiazcy:
VOTE +1 pending updates to the docker-compose files as well

Cole-Greer:
The scope of this PR has changed quite drastically since it was originally opened. The Go upgrade is no longer a part of this, as that's already been handled separately. I've updated the PR name and description to reflect the current scope which is expanding node support and testing. I am considering the old votes from Valentyn and Yang as dismissed, as the branch has changed drastically since then.

My vote is VOTE +1

.github/workflows/build-test.yml
```diff
@@ -198,16 +198,67 @@ jobs:
...
+            ~/.m2/repository/org/apache/tinkerpop/*
+          key: ${{ github.sha }}
+      - name: Load Docker Image
+        working-directory: ./gremlin-server
+        run: docker load --input gremlin-server.tar
+      - name: Build with Maven
+        run: |
+          EXCLUDE="-:gremlin-dotnet-source,-:gremlin-dotnet-tests,-:gremlin-go,-:gremlin-python,$EXCLUDE_FOR_GLV"
+          mvn clean install -pl $EXCLUDE -q -DskipTests -Dci
+          mvn verify -pl :gremlin-javascript,:gremlint,:gremlin-mcp -Dnode.test.version=20
```
andreachild, line 228: Are gremlint and gremlin-mcp needed here?

```suggestion
          mvn verify -pl :gremlin-javascript -Dnode.test.version=20
```

CHANGELOG.asciidoc
```diff
@@ -105,6 +105,7 @@ This release also includes changes from <<release-3-7-XXX, 3.7.XXX>>.
 * Introduced `NotP` class for negation of `P`.
 * Add barriers to `group(String)`, `groupCount(String)`, `tree(String)` and `subgraph(String)` steps.
 * Increase minimum Java version from 1.8 to 11 for building and running.
+* Add support for Node 22 and 24
```
andreachild: Nit: might want to mention Node 20 is still acceptable?

.github/workflows/build-test.yml
Cole-Greer: Historically we've always tested gremlint (and now gremlin-mcp) as well through the `javascript` job in github actions. I've retained that here with testing on the current minimum version (node 20), which is what the "all" in the name `javascript-all-node20` is referring to.

Ideally I'd like to be testing gremlint and gremlin-mcp on the newer node versions as well, but those builds are not currently dockerized so they are in need of a different solution (which is why the node 22 and 24 jobs are currently glv-only).

andreachild:
VOTE +1

pom.xml
```diff
@@ -226,6 +226,7 @@ limitations under the License.
 
         <runtime.npm.version>10.8.2</runtime.npm.version>
         <runtime.node.version>v20.19.4</runtime.node.version>
+        <node.test.version>20</node.test.version>
```
kenhuuu, line 229: Why can't you just re-use `runtime.node.version`?

Cole-Greer: `runtime.node.version` needs to be in its exact format (`vMAJOR.MINOR.PATCH`) to work with `frontend-maven-plugin`. I wanted to continue using the docker major version tag, without pinning minor and patch versions, also at minimum the `v` would need to be parsed out of the `runtime.node.version`. This version extraction would need to happen either in the pom file when the version is injected as an environment variable, or within `docker.compose.yml` when the node image is fetched. I didn't find a nice way to get this working so I went with the simpler (albeit redundant) approach.

I admittedly didn't spend much time on this, there likely is a better way to extract the version which I haven't found yet.

kenhuuu:
VOTE +1

PR #3121, opened 2025-05-17
TINKERPOP-3115 Better handle overflows with sum()

EchoNullify:
https://issues.apache.org/jira/browse/TINKERPOP-3115

Promote to next highest number type to prevent sum step from overflowing


docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -30,6 +30,10 @@ complete list of all the modifications that are part of this release.
 
 === Upgrading for Users
 
+==== Auto promotion of number types
+
+Any math operations such as Add, Sub, Mul & Div will automatically promote to next number type if an overflow is detected. For integers the promotion sequence is byte->short->integer->long->overflow. For floating point the sequence is float->double->infinity.
```
andreachild, line 35: Would be worth mentioning the previous behaviour would wrap around to min/max value for the type in overflow scenario and now this is the new behaviour.

Would also be good to mention that the promotion sequence will throw exception on long overflow.

CHANGELOG.asciidoc
```diff
@@ -25,6 +25,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 This release also includes changes from <<release-3-7-XXX, 3.7.XXX>>.
 
+* Added promote to next highest number type to prevent sum step overflow
```
andreachild, line 28: Since other numeric operations were also modified, this affects more than just sum step.

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -334,6 +338,10 @@ link:https://tinkerpop.apache.org/docs/3.8.0/dev/provider/#gherkin-tests-suite[P
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-3136[TINKERPOP-3136]
 
+==== Auto promotion of number types
+
+Any math operations such as Add, Sub, Mul & Div will automatically promote to next number type if an overflow is detected. For integers the promotion sequence is byte->short->integer->long->overflow. For floating point the sequence is float->double->infinity.
```
andreachild, line 343: Worth mentioning that if providers can change overflow handling logic but it's not recommended to handle overflow by wrapping to min/max value which can produce unexpected results for users.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
```diff
@@ -28,6 +28,50 @@
...
+        int bits;
+        boolean fp;
+
+        NumberInfo(int bits, boolean fp) {
+            this.bits = bits;
+            this.fp = fp;
+        }
+    }
+
+    enum MathOperation {
```
andreachild, line 42: This enum can be avoided and code duplication reduced by changing the `mathOperationWithPromote` method signature to accept a `Function` param which takes a `NumberHelper` and returns a `BiFunction`.

Something like this:

```
private static Number mathOperationWithPromote(final Function<NumberHelper, BiFunction<Number, Number, Number>> mathFunction, final boolean forceFloatingPoint, final Number a, final Number b) {
...
  final Class<? extends Number> clazz = determineNumberClass(numberInfo.bits, numberInfo.fp);
  NumberHelper helper = getHelper(clazz);
  result = mathFunction.apply(helper).apply(a, b);
...
```
}

And then it can be called like this:

```
public static Number add(final Number a, final Number b) {
    return mathOperationWithPromote(numberHelper -> numberHelper.add, false, a, b);
}

public static Number sub(final Number a, final Number b) {
    return mathOperationWithPromote(numberHelper -> numberHelper.sub, false, a, b);
}
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
```diff
@@ -312,43 +356,84 @@ private NumberHelper(final BiFunction<Number, Number, Number> add,
...
+    }
+
+    private static Number mathOperationWithPromote(final MathOperation mathOperation, final boolean forceFloatingPoint, final Number a, final Number b) {
+        if (null == a || null == b) return a;
+        NumberInfo numberInfo = getHighestCommonNumberInfo(forceFloatingPoint, a, b);
+        Number result = 0;
+        while (true) {
+            try {
+                result = mathOperation.apply(numberInfo, a, b);
+                if (result.getClass() == BigInteger.class || result.getClass() == BigDecimal.class)
```
andreachild, line 405: It would be preferable to use `result instanceof BigInteger` instead of `getClass() ==` etc for a few reasons:
* handles null while `getClass()` would throw NPE
* handles check of potential custom classes that extend the number classes
* `instanceof` is conventionally used throughout this class and the entire project

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
```diff
@@ -312,43 +356,84 @@ private NumberHelper(final BiFunction<Number, Number, Number> add,
...
+                {
+                    return result;
+                }
+                if (Double.isInfinite(result.doubleValue()))
+                {
+                    throw new ArithmeticException("Floating point overflow detected");
```
andreachild, line 411: I don't think the `BigInteger` and `BigDecimal` checks or throwing this exception are necessary. You can change to only check for Float infinity:

```
try {
    result = mathOperation.apply(numberInfo, a, b);
    if (!(result instanceof Float) || !Float.isInfinite(result.floatValue())) {
        return result;
    }
    numberInfo.bits <<= 1;
}
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
```diff
@@ -312,43 +356,84 @@ private NumberHelper(final BiFunction<Number, Number, Number> add,
...
+        }
     }
 
     /**
      * Adds two numbers returning the highest common number class between them.
      *
+     * <p>
+     * This method returns a result using the highest common number class between the two inputs.
+     * If an overflow occurs (either integer or floating-point), the method promotes the precision
+     * by increasing the bit width or switching to floating-point arithmetic, until a suitable type is found.
```
andreachild, line 431: Logic does not switch to floating point, perhaps this comment wasn't updated after a change in logic?

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/util/NumberHelperTest.java
```diff
@@ -99,20 +114,36 @@ public class NumberHelperTest {
...
             new Triplet<>(Long.MAX_VALUE, 1L, "add"),
             new Triplet<>(Long.MIN_VALUE, 1L, "sub"),
-            new Triplet<>(Long.MAX_VALUE,  Integer.MAX_VALUE, "mul"),
-            new Triplet<>(Byte.MAX_VALUE, (byte)100, "add"),
-            new Triplet<>(Byte.MIN_VALUE, (byte)100, "sub"),
-            new Triplet<>((byte)100, (byte)100, "mul"),
-            new Triplet<>(Short.MAX_VALUE, (short)100, "add"),
-            new Triplet<>(Short.MIN_VALUE, (short)100, "sub"),
-            new Triplet<>(Short.MAX_VALUE, (short)100, "mul")
+            new Triplet<>(Long.MAX_VALUE,  Integer.MAX_VALUE, "mul")
```
andreachild, line 119: These two test cases should be added to overflow scenarios:

```
new Triplet<>(Long.MIN_VALUE, -1L, "mul"),
new Triplet<>(Long.MIN_VALUE, -1L, "div")
```

and actually the `div` case fails because `Math.divideExact` was not added until Java 18 or 19, which means we have to add extra handling around the single scenario in which division causes overflow.

See https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/lang/Math.html#divideExact(int,int)

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/util/NumberHelperTest.java
```diff
@@ -99,20 +114,36 @@ public class NumberHelperTest {
...
+            new Quartet<>(Integer.MAX_VALUE, 1, "add", "l"),
+            new Quartet<>(Integer.MIN_VALUE, 1, "sub", "l"),
+            new Quartet<>(Integer.MAX_VALUE, Integer.MAX_VALUE, "mul", "l"),
+            new Quartet<>(Integer.MAX_VALUE, 0.5f, "div", "f"),
+            new Quartet<>(Long.MAX_VALUE, 0.5f, "div", "d")
     );
 
+    @Test
+    public void shouldReturnHighestCommonNumberNumberInfo() {
+        for (final Triplet<Number, Integer, Boolean> q : NUMBER_INFO_CASES) {
```
andreachild, line 140: Nit: it would be better to use junit's `Parameterized` instead of a for loop inside a single test because it would run each execution separately, making it easier to pinpoint which test cases passed and which failed. Using a for loop will just fail the entire test and then you'll need to debug which scenario failed.

See `GeneralLiteralVisitorTest` as an example of how `Parameterized` can be used.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
EchoNullify: Im using casting of any type of integer value being casted as double to cause the exception handler to follow the promotion rules

docs/src/upgrade/release-3.8.x.asciidoc
EchoNullify: Done

docs/src/upgrade/release-3.8.x.asciidoc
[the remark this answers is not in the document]
EchoNullify: Done

docs/src/upgrade/release-3.8.x.asciidoc
EchoNullify: Done

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
EchoNullify: Updated

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/util/NumberHelperTest.java
EchoNullify: All previous tests use same way, ideally id update all of them, but won't for now

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
EchoNullify: Done

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
```diff
@@ -312,43 +356,84 @@ private NumberHelper(final BiFunction<Number, Number, Number> add,
...
+    }
+
+    public static Class<? extends Number> getHighestCommonNumberClass(final boolean forceFloatingPoint, final Number... numbers) {
+        NumberInfo numberInfo = getHighestCommonNumberInfo(forceFloatingPoint, numbers);
+        return determineNumberClass(numberInfo.bits, numberInfo.fp);
+    }
+
+    private static Number mathOperationWithPromote(final MathOperation mathOperation, final boolean forceFloatingPoint, final Number a, final Number b) {
+        if (null == a || null == b) return a;
+        NumberInfo numberInfo = getHighestCommonNumberInfo(forceFloatingPoint, a, b);
```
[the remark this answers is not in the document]
EchoNullify: Done

gremlin-python/src/main/python/radish/gremlin.py
```diff
@@ -701,24 +701,15 @@
...
-    'g_injectXdatetimeXstr1XX_dateDiffXinjectXdatetimeXstr2XXX': [(lambda g:g.inject(datetime.datetime.fromisoformat('2023-08-08T00:00+00:00')).date_diff(__.inject(datetime.datetime.fromisoformat('2023-10-11T00:00+00:00'))))], 
-    'g_injectXDateTimeXstr1XX_dateDiffXDateTimeXstr2XX': [(lambda g:g.inject(datetime.datetime.fromisoformat('2023-08-02T00:00+00:00')).date_diff(datetime.datetime.fromisoformat('2023-08-09T00:00+00:00')))], 
-    'g_injectXDateTimeXstr1XX_dateDiffXconstantXDateTimeXstr2XXX': [(lambda g:g.inject(datetime.datetime.fromisoformat('2023-08-08T00:00+00:00')).date_diff(__.constant(datetime.datetime.fromisoformat('2023-08-01T00:00+00:00'))))], 
-    'g_injectXDateTimeXstr1XX_dateDiffXinjectXDateTimeXstr2XXX': [(lambda g:g.inject(datetime.datetime.fromisoformat('2023-08-08T00:00+00:00')).date_diff(__.inject(datetime.datetime.fromisoformat('2023-10-11T00:00+00:00'))))], 
+    'g_injectXdatetimeXstrXX_dateAddXDT_hour_2X': [(lambda g:g.inject(datetime.datetime.fromisoformat('2023-08-02T00:00+00:00'), datetime.datetime.fromisoformat('2023-08-02T00:00+00:00')).date_add(DT.hour, 2))],
```
andreachild, line 704: I might be missing something obvious but it is unexpected to me that there are changes to the `datetime` related tests and no changes to the `sum` tests?

EchoNullify: I think it was something missing from the auto generated file as I've done no date time changes

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/util/NumberHelperTest.java
EchoNullify: Done. For the long divide added a check for the div case throwing arithmatic exception

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/Sum.feature
```diff
@@ -18,6 +18,42 @@
 @StepClassMap @StepSum
 Feature: Step - sum()
 
+  @GraphComputerVerificationInjectionNotSupported
+  Scenario: g_V_sum_overflow_byte
```
andreachild, line 22: These tests don't follow the syntax specified by https://tinkerpop.apache.org/docs/current/dev/developer/#gremlin-language-test-cases

EchoNullify: Fixed

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/sideEffect/Sack.feature
andreachild, line 1: Should there be tests for exception cases as well? ie. long overflow

gremlin-dotnet/build/generate.groovy
```diff
@@ -94,7 +94,7 @@ radishGremlinFile.withWriter('UTF-8') { Writer writer ->
 
 
     // some traversals may require a static translation if the translator can't handle them for some reason
-    def staticTranslate = [:]
+    def staticTranslate = ["g_withSackX_128bX_injectX1bX_sackXminusX_sack" : "               {\"g_withSackX_128bX_injectX1bX_sackXminusX_sack\", new List<Func<GraphTraversalSource, IDictionary<string, object>, ITraversal>> {(g,p) =>g.WithSack((short) -128).Inject<object>((short) 1).Sack(Operator.Minus).Sack<object>()}}, "]
```
andreachild, line 97: Curious why manual translation was required for this test but not `g_withSackX127bX_injectX1bX_sackXsumX_sack`?

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/Sum.feature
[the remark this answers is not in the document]
EchoNullify: Ive added underflow tests, however I can't do long overflow test in a feature as javascript doesn't exception, which is why covered that in the number helper tests

gremlin-dotnet/build/generate.groovy
EchoNullify: because for cornet a byte can be 0 to 255 and is not signed

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/features/StepDefinition.java
```diff
@@ -723,7 +723,7 @@ private static <T> List<org.hamcrest.Matcher<? super T>> getMatchers(final T[] i
                 matchers.add((org.hamcrest.Matcher<? super T>) new IsPathEqualToMatcher((Path) item));
             } else if (item instanceof Double && Double.isFinite((Double) item)) {
                // Allow for minor rounding errors with Double
-               matchers.add((org.hamcrest.Matcher<? super T>) closeTo((Double) item, 0.000000000000001));
+               matchers.add((org.hamcrest.Matcher<? super T>) closeTo((Double) item, 1e+30));
```
andreachild, line 726: Why was this change needed, and did you mean to change it to `1e-30` since `1e+30` is a very large number?

CHANGELOG.asciidoc
```diff
@@ -25,6 +25,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 This release also includes changes from <<release-3-7-XXX, 3.7.XXX>>.
 
+* Added promote to next highest number type to prevent steps such as sum, sack from overflowing.
```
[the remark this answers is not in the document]
EchoNullify: Updated thanks

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/features/StepDefinition.java
EchoNullify: Reverted

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -30,6 +30,14 @@ complete list of all the modifications that are part of this release.
 
 === Upgrading for Users
 
+==== Auto promotion of number types
+
+Previously, operations like sum or sack that involved mathematical calculations did not automatically promote the result to a larger numeric type (e.g., from int to long) when needed. As a result, values could wrap around within their current type, leading to unexpected behavior. This issue has now been resolved by enabling automatic type promotion for results.
+
+As a result, values would previously wrap around within their current type—overflowing at their minimum or maximum boundaries—which often led to incorrect, unintuitive, or unexpected behavior. This issue has now been resolved by introducing automatic type promotion for results.
```
andreachild, line 37: It seems that this paragraph is repeating some of the sentence from the previous paragraph?

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/util/NumberHelperTest.java
```diff
@@ -99,20 +114,38 @@ public class NumberHelperTest {
...
+            new Quartet<>(Byte.MAX_VALUE, 0.5f, "div", "f"),
+            new Quartet<>(Short.MAX_VALUE, (short)100, "add", "i"),
+            new Quartet<>(Short.MIN_VALUE, (short)100, "sub", "i"),
+            new Quartet<>(Short.MAX_VALUE, (short)100, "mul", "i"),
+            new Quartet<>(Short.MAX_VALUE, 0.5f, "div", "f"),
+            new Quartet<>(Integer.MAX_VALUE, 1, "add", "l"),
+            new Quartet<>(Integer.MIN_VALUE, 1, "sub", "l"),
+            new Quartet<>(Integer.MAX_VALUE, Integer.MAX_VALUE, "mul", "l"),
+            new Quartet<>(Integer.MAX_VALUE, 0.5f, "div", "f"),
+            new Quartet<>(Long.MAX_VALUE, 0.5f, "div", "d")
```
Cole-Greer, line 137: Would be good to add cases for
```
new Quartet<>(Byte.MIN_VALUE, (byte)-1, "div", "s"),
new Quartet<>(Short.MIN_VALUE, (short)-1, "div", "i"),
new Quartet<>(Integer.MIN_VALUE, -1, "div", "l"),
```

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -104,6 +104,16 @@ convert all of them to floating point.
 BigDecimal and BigInteger may not be supported depending on the language and Storage, therefore the behavior of type
 casting for these two types can vary depending on a Graph provider.
 
+**Number Promotion on Mathematical Operations**
```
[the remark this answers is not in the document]
EchoNullify: Broke up sections with appropriate subject to make more clearer

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
```diff
@@ -122,7 +132,12 @@ private static short asShort(int arg) {
             (a, b) -> Math.addExact(a.intValue(), b.intValue()),
             (a, b) -> Math.subtractExact(a.intValue(), b.intValue()),
             (a, b) -> Math.multiplyExact(a.intValue(), b.intValue()),
-            (a, b) -> a.intValue() / b.intValue(),
+            (a, b) -> {
+                if (a.longValue() == Integer.MIN_VALUE && b.intValue() == -1) {
```
andreachild, line 136: Should be `a.intValue()` not `a.longValue()`?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
EchoNullify: I found it was needed for cases where you start with a big intereger / decimal

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/NumberHelper.java
EchoNullify: Fixed cut and paste typo

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -104,6 +104,16 @@ convert all of them to floating point.
...
+
+On math operations (when used in steps such as sum/sack) the highest common number class between
+the two inputs us used. Any number being floating point forces a floating point result.
```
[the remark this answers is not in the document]
EchoNullify: Typo fixed

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/util/NumberHelperTest.java
EchoNullify: Added: these cases

gremlin-dotnet/build/generate.groovy
Cole-Greer: For reference: https://issues.apache.org/jira/browse/TINKERPOP-3161

Cole-Greer:
VOTE +1

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -30,6 +30,12 @@ complete list of all the modifications that are part of this release.
 
 === Upgrading for Users
 
+==== Auto promotion of number types
```
spmallette, line 33: When writing documentation, It often helps to show the before and after behavior. Like, provide a Gremlin Console session showing how it behaved in prior versions and then how the output looks in  the current version. This section would benefit from that. Here's an example: https://tinkerpop.apache.org/docs/current/upgrade/#_string_manipulation_steps

also, it doesn't look like these lines are hard wrapped we tend to add line breaks with the docs.

EchoNullify: Comments should now all be addressed

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -30,6 +30,26 @@ complete list of all the modifications that are part of this release.
...
+
+As a example the following query...
+
+"""
+g.withSack(32767s).inject(1s).sack(sum).sack()
+"""
+
+Before would return a short overflow exception or wrap to -1 depending on language, but now returns 32769i.
+
+See https://issues.apache.org/jira/browse/TINKERPOP-3115 for more details.
```
Cole-Greer: Nit: 
```suggestion
See link:https://issues.apache.org/jira/browse/TINKERPOP-3115[TINKERPOP-3115] for more details.
```

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -334,6 +349,26 @@ link:https://tinkerpop.apache.org/docs/3.8.0/dev/provider/#gherkin-tests-suite[P
...
```
Cole-Greer: Nit:
```suggestion
See link:https://issues.apache.org/jira/browse/TINKERPOP-3115[TINKERPOP-3115] for more details.
```

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -30,6 +30,26 @@ complete list of all the modifications that are part of this release.
...
+if an overflow is detected. For integers, the promotion sequence is: byte → short → int → long → overflow exception. For
+floating-point numbers, the sequence is: float → double → infinity.
+
+As a example the following query...
+
+"""
+g.withSack(32767s).inject(1s).sack(sum).sack()
+"""
+
+Before would return a short overflow exception or wrap to -1 depending on language, but now returns 32769i.
```
```suggestion
The following example showcases the change in overflow behavior between 3.7.3 and 3.8.0

[source,groovy]
----
// 3.7.3
gremlin> g.inject([Byte.MAX_VALUE, (byte) 1], [Short.MAX_VALUE, (short) 1], [Integer.MAX_VALUE,1], [Long.MAX_VALUE, 1l]).sum(local)
==>-128 // byte
==>-32768 // short
==>-2147483648 // int
==>-9223372036854775808 // long

gremlin> g.inject([Float.MAX_VALUE, Float.MAX_VALUE], [Double.MAX_VALUE, Double.MAX_VALUE]).sum(local)
==>Infinity // float
==>Infinity // double

// 3.8.0
gremlin> g.inject([Byte.MAX_VALUE, (byte) 1], [Short.MAX_VALUE, (short) 1], [Integer.MAX_VALUE,1]).sum(local)
==>128 // short
==>32768 // int
==>2147483648 // long

gremlin> g.inject([Long.MAX_VALUE, 1l]).sum(local)
// throws java.lang.ArithmeticException: long overflow

gremlin> g.inject([Float.MAX_VALUE, Float.MAX_VALUE], [Double.MAX_VALUE, Double.MAX_VALUE]).sum(local)
==>6.805646932770577E38 // double
==>Infinity // double
----

PR #3123, opened 2025-05-29
Only include STJ on net standard

thompson-tomo:
This PR adjusts the inclusion of System.Text.Json so that it is only included on net standard and not for the newer frameworks ie net 6.

The benefit is that it reduces the risk of conflicts which has occurred with other libraries, enables quicker remediation to CVE'S in STJ when end application deployed being dependent on framework as windows update can update the framework.


Cole-Greer:
Hi @thompson-tomo, I'm curious what is the motivation for this change. I'm not aware of any compatibility issues here, but .Net is also outside of my area of expertise.

thompson-tomo:
@Cole-Greer i have run in to issues with other libraries. The motivation is to proactively reduce the chance of issues/conflicts while also improving the ability to quickly remediate CVE'S in the field as the responsibility for managing the affected library falls to the framework rather than the consuming application

andreachild:
VOTE +1

EchoNullify:
VOTE +1

Cole-Greer:
VOTE +1

xiazcy:
VOTE +1

Cole-Greer:
Note: I've also cherry-picked this change back to 3.8-dev: https://github.com/apache/tinkerpop/commit/00622c5d02d0e7bae653cee1c4ee9d1a437f7297

PR #3133, opened 2025-06-11
TINKERPOP-3023 Expand type syntax in grammar in 3.8 - Remove Vertex

EchoNullify:
Feat: Remove Vertex support from grammar, and for places where a vertex is required, pass in the vertex id instead at traversal construction. 

- The grammar currently includes inconsistencies, particularly around "type constructors" like ReferenceVertex(), which are being re-evaluated.

- ReferenceVertex should be renamed simply to Vertex, as the distinction (e.g., between ReferenceVertex and DetachedVertex) is no longer meaningful or guaranteed.

- The keyword new should be removed from the grammar rule for structureVertex, as it's syntactically unnecessary.

- The Vertex constructor should only require an id, not a label. Labels are not utilized in practice (e.g., in g.V(Vertex)), and can default to an empty string.

Resolves: [3023](https://issues.apache.org/jira/browse/TINKERPOP-3023),

For reference...

[https://lists.apache.org/thread/ptq81160z4y4vddg4lxz2nv244x2yzhq](https://lists.apache.org/thread/ptq81160z4y4vddg4lxz2nv244x2yzhq)


gremlin-language/src/test/resources/incorrect-traversals.txt
```diff
@@ -23,7 +23,6 @@ g.V().horder().by(id)
 g.mergeE(["weight": 0.5, Direction.out: 1])
 g.mergeE(["weight": 0.5, Direction.in: 1])
 g.addE('knows').from(g.V(1))
-g.addE('knows').from(g.V(1).next())
```
andreachild, line 26: Is this now a valid traversal?

EchoNullify: Previously was invalid but now works

docs/src/recipes/between-vertices.asciidoc
```diff
@@ -22,14 +22,14 @@ traversal on the paths found between them. Consider the following examples using
...
-g.V(1).bothE().where(otherV().hasId(2))          <2>
+g.V(1).bothE()                                          <1>
+g.V(1).bothE().where(otherV().hasId(2))                 <2>
 v1 = g.V(1).next();[]
 v2 = g.V(2).next();[]
-g.V(v1).bothE().where(otherV().is(v2))           <3>
-g.V(v1).outE().where(inV().is(v2))               <4>
-g.V(1).outE().where(inV().has(id, within(2,3)))  <5>
-g.V(1).out().where(__.in().hasId(6))             <6>
+g.V(v1.id()).bothE().where(otherV().id().is(v2.id()))   <3>
```
Cole-Greer: I'm not sure if we want to change these examples in the docs. My expectation is that most users will continue to use the sugar syntax of directly passing vertices in both the GLVs and console. It's simpler to read and understand if we leave it with the sugar syntax.

In order to keep it with the sugared syntax, we will need to update the [DocumentationReader](https://github.com/apache/tinkerpop/blob/3.8-dev/gremlin-language/src/main/java/org/apache/tinkerpop/gremlin/language/corpus/DocumentationReader.java) to ignore these cases. I think we can either add these examples to the [`throwAway` ignore set at the top of the class](https://github.com/apache/tinkerpop/blob/6c0459863758f5f4966631786985306ce8478532/gremlin-language/src/main/java/org/apache/tinkerpop/gremlin/language/corpus/DocumentationReader.java#L42), or we can change the [replace patterns](https://github.com/apache/tinkerpop/blob/6c0459863758f5f4966631786985306ce8478532/gremlin-language/src/main/java/org/apache/tinkerpop/gremlin/language/corpus/DocumentationReader.java#L113-L114) to translate v1 and v2 to something safe like `"1"` and `"2"` similarly to how `vBob` is handled above.

Thoughts?

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -30,6 +30,41 @@ complete list of all the modifications that are part of this release.
 
 === Upgrading for Users
 
+==== Removal of Vertex/ReferenceVertex from gtammar
```
```suggestion
==== Removal of Vertex/ReferenceVertex from grammar
```

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -30,6 +30,41 @@ complete list of all the modifications that are part of this release.
...
+
+Previously the use of Vertex/ReferenceVertex existed in the grammar, these have now been removed. Going forward, vertex ID should 
+be used in traversals.
+
+// 3.7.3
+gremlin> graph = TinkerFactory.createModern()
+==>tinkergraph[vertices:6 edges:6]
+gremlin> g = graph.traversal()
+==>graphtraversalsource[tinkergraph[vertices:6 edges:6], standard]
```
Cole-Greer: Nit: For brevity
```suggestion
gremlin> g = TinkerFactory.createModern().traversal()
==>graphtraversalsource[tinkergraph[vertices:6 edges:6], standard]
```

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -30,6 +30,41 @@ complete list of all the modifications that are part of this release.
...
+gremlin> v2 = g.V(2).next();
+==>v[2]
+gremlin> g.V(v1).outE().where(inV().is(v2))
+==>e[7][1-knows->2]
+
+// 3.8.0
+gremlin> graph = TinkerFactory.createModern()
+==>tinkergraph[vertices:6 edges:6]
+gremlin> g = graph.traversal()
+==>graphtraversalsource[tinkergraph[vertices:6 edges:6], standard]
```
Cole-Greer: Nit:
```suggestion
gremlin> g = TinkerFactory.createModern().traversal()
==>graphtraversalsource[tinkergraph[vertices:6 edges:6], standard]
```

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -30,6 +30,41 @@ complete list of all the modifications that are part of this release.
 
 === Upgrading for Users
 
+==== Removal of Vertex/ReferenceVertex from gtammar
+
+Previously the use of Vertex/ReferenceVertex existed in the grammar, these have now been removed. Going forward, vertex ID should
```
Cole-Greer: I think it's important to make clear that the removal is only in the grammar, and thus only applies to users directly writing gremlin-lang scripts. Vertex as an argument to these steps has been retained as sugared syntax in `GraphTraversal` for all of the GLVs, as well as in gremlin-groovy. This means that this change should not be apparent when for GLV or console users.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AddEdgeStep.java
```diff
@@ -98,43 +99,61 @@ public void addFrom(final Traversal.Admin<?, ?> fromObject) {
...
                     "addE(%s) failed because the from() traversal (which should give a Vertex) failed with: %s",
                     edgeLabel, e.getMessage()));
         }
 
-        if (!(theFrom instanceof Vertex))
+        if (theFrom == null)
             throw new IllegalStateException(String.format(
-                    "The value given to addE(%s).to() must resolve to a Vertex but %s was specified instead", edgeLabel,
-                    null == theFrom ? "null" : theFrom.getClass().getSimpleName()));
+                    "The value given to addE(%s).to() must resolve to a Vertex but null was specified instead", edgeLabel));
```
Cole-Greer:
```suggestion
                    "The value given to addE(%s).from() must resolve to a Vertex or the ID of a Vertex present in the graph, but null was specified instead", edgeLabel));
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AddEdgeStep.java
```diff
@@ -98,43 +99,61 @@ public void addFrom(final Traversal.Admin<?, ?> fromObject) {
...
                     "addE(%s) failed because the to() traversal (which should give a Vertex) failed with: %s",
                     edgeLabel, e.getMessage()));
         }
 
-        if (!(theTo instanceof Vertex))
+        if (theTo == null)
             throw new IllegalStateException(String.format(
-                    "The value given to addE(%s).to() must resolve to a Vertex but %s was specified instead", edgeLabel,
-                    null == theTo ? "null" : theTo.getClass().getSimpleName()));
+                    "The value given to addE(%s).to() must resolve to a Vertex but null was specified instead", edgeLabel));
```
Cole-Greer:
```suggestion
                    "The value given to addE(%s).to() must resolve to a Vertex or the ID of a Vertex present in the graph, but null was specified instead", edgeLabel));
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AddEdgeStartStep.java
```diff
@@ -110,27 +111,45 @@ protected Traverser.Admin<Edge> processNextStart() {
...
-            if (!(theFrom instanceof Vertex))
+            Object theFrom = this.parameters.get(traverser, FROM, () -> null).get(0);
+            if (theFrom != null && !(theFrom instanceof Vertex)) {
+                theFrom = new ReferenceVertex(theFrom);
+            }
+            if (theFrom == null)
                 throw new IllegalStateException(String.format(
-                        "The value given to addE(%s).from() must resolve to a Vertex but %s was specified instead", edgeLabel,
-                        null == theFrom ? "null" : theFrom.getClass().getSimpleName()));
+                        "The value given to addE(%s).from() must resolve to a Vertex but null was specified instead", edgeLabel));
```
Cole-Greer:
```suggestion
                        "The value given to addE(%s).from() must resolve to a Vertex or the ID of a Vertex present in the graph, but null was specified instead", edgeLabel));
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AddEdgeStartStep.java
```diff
@@ -110,27 +111,45 @@ protected Traverser.Admin<Edge> processNextStart() {
...
+            Object theTo = this.parameters.get(traverser, TO, () -> null).get(0);
+            if (theTo != null && !(theTo instanceof Vertex)) {
+                theTo = new ReferenceVertex(theTo);
+            }
+
+            if (theTo == null)
                 throw new IllegalStateException(String.format(
-                        "The value given to addE(%s).to() must resolve to a Vertex but %s was specified instead", edgeLabel,
-                        null == theTo ? "null" : theTo.getClass().getSimpleName()));
+                        "The value given to addE(%s).to() must resolve to a Vertex but null was specified instead", edgeLabel));
```
Cole-Greer:
```suggestion
                        "The value given to addE(%s).to() must resolve to a Vertex or the ID of a Vertex present in the graph, but null was specified instead", edgeLabel));
```

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs
```diff
@@ -862,6 +862,22 @@ public GraphTraversal<TStart, string> Format(string format)
...
+            if (fromStepLabel is string fromStepLabelString)
+            {
+                return From(fromStepLabelString);
+            }
+            if (fromStepLabel is int fromStepLabelSInteger)
+            {
+                return From(new Vertex(fromStepLabelSInteger));
+            }
+            throw new ArgumentException($"{fromStepLabel} is not a valid step label");
+        }
```
Cole-Greer: I don't believe it's necessary to split into the more specific overloads here, and I don't think the `ArgumentException` is right. `object? fromStepLabel` could be any arbitrary vertex id, which could be of any type. The one special case which should be handled is that if `object? fromStepLabel` is a `vertex`, the GLV should handle that pop out the ID and pass that along instead. Other than that one case, it should be fine to simply pass the id as an `object` into the bytecode for this step.

```suggestion
        public GraphTraversal<TStart, TEnd> From (object? fromStepVertexIdOrLabel)
        {
            if (fromStepVertexIdOrLabel is Vertex fromStepVertex)
            {
                fromStepVertexIdOrLabel = fromStepVertex.Id;
            }
            Bytecode.AddStep("from", fromStepVertexIdOrLabel);
            return Wrap<TStart, TEnd>(this);
        }
```

Similar should apply for the To() step below

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs
```diff
@@ -2193,6 +2209,22 @@ public GraphTraversal<TStart, Vertex> To (Direction? direction, params string?[]
             Bytecode.AddStep("to", args.ToArray());
             return Wrap<TStart, Vertex>(this);
         }
+        
+        /// <summary>
+        ///     Adds the to step to this <see cref="GraphTraversal{SType, EType}" />.
+        /// </summary>
+        public GraphTraversal<TStart, TEnd> To (object? toStepLabel)
+        {
```
Cole-Greer, line 2210: See above suggestion for From()

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs
Cole-Greer: This same sort of "pop the id out of vertices" approach should apply to the equivalent GraphTraversal methods in all of the GLVs.

gremlin-language/src/test/resources/incorrect-traversals.txt
Cole-Greer: This is interesting. from() in the grammar previously only allowed a nestedTraversal, but not a terminatedTraversal. By putting `genericArgument` in there to accept any arbitrary ID type, we've now allowed terminatedTraversal to leak in. I don't think this is a big concern.
@spmallette any thoughts on this?

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/MergeEdge.feature
```diff
@@ -151,7 +151,7 @@ Feature: Step - mergeE()
       g.addV("person").property("name", "marko").
         addV("person").property("name", "vadas")
       """
-    And using the parameter xx1 defined as "m[{\"t[label]\": \"knows\", \"D[OUT]\":\"v[marko]\", \"D[IN]\":\"v[vadas]\"}]"
+    And using the parameter xx1 defined as "m[{\"t[label]\": \"knows\", \"D[OUT]\":\"v[marko].id\", \"D[IN]\":\"v[vadas].id\"}]"
```
Cole-Greer, line 154: These mergeE cases worry me. Once this gets merged up to TinkerPop 4 and GraphTraversal starts generating GremlinLang instead of Bytecode, Vertex will no longer be allowed in the mergeE merge map. This would be an unintended breaking change.
It's a bit messy but I wonder if we need to add that same sort of `Vertex -> Id` syntax sugar in the mergeE map at the GraphTraversal level in each GLV similar to what has been done in from() and to().
Thoughts?

docs/src/recipes/between-vertices.asciidoc
spmallette: +1 to not removing the sugar syntax. at some point we'll need to better acknowledge sugar syntax vs what the grammar supports.  it would be nice to come up with some foundations for that here.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/MergeEdge.feature
Cole-Greer: After further consideration, I think it's best that this vertex -> id sugar be excluded from 3.8 dev as it's not necessary for bytecode. I think a seperate PR which includes such sugar for master might be warranted here.

gremlin-dotnet/src/Gremlin.Net/Process/Traversal/GraphTraversal.cs
Cole-Greer: After further consideration I think this vertex -> id sugar is best left exclusively to the master branch

Cole-Greer:
I've pushed some changes into the branch which updates the docs a bit and strips out the vertex -> id sugar in GraphTraversal as it's not required for bytecode. (to be added to master in a separate PR)

My vote is VOTE +1

Cole-Greer:
Note: Followup PR for master: https://github.com/apache/tinkerpop/pull/3138

PR #3135, opened 2025-06-12
Proposal for asBool() Step

xiazcy:
As another addition to the existing set of conversion steps (i.e. `asString()`, `asDate()`), this proposal aims to define a new `asBool()` step for Boolean parsing.

VOTE +1


Cole-Greer:
VOTE +1

PR #3137, opened 2025-06-19
Passing of authenticated user in query lifecycle callbacks in GraphManager.

andrii0lomakin:
Our database (YouTrackDB) works with records only on a session basis and requires information about the authenticated user for real-time authorisation of user operations for each operation on a specific resource: schema, record, sequence ... and so on. 

So I have introduced one additional lifecycle method and added an authenticated user parameter to the one that is called before the request.

An additional method is called at the end of request processing once the transaction is rolled back if it is managed by session or traversal.

### DIFF
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
@@ -18,6 +18,7 @@
  */
 package org.apache.tinkerpop.gremlin.server;
 
+import org.apache.tinkerpop.gremlin.server.auth.AuthenticatedUser;
 import org.apache.tinkerpop.gremlin.util.message.RequestMessage;
 import org.apache.tinkerpop.gremlin.process.traversal.TraversalSource;
 import org.apache.tinkerpop.gremlin.structure.Graph;
@@ -125,7 +126,7 @@ public interface GraphManager {
             return graph.features().graph().supportsTransactions() && graph.tx().isOpen();
         });
     }
- 
+
     /**
      * This method will be called before a script or query is processed by the
      * gremlin-server.
@@ -133,18 +134,40 @@ public interface GraphManager {
      * @param msg the {@link RequestMessage} received by the gremlin-server.
      */
     default void beforeQueryStart(final RequestMessage msg) {
-
     }
 
     /**
      * This method will be called before a script or query is processed by the
      * gremlin-server.
+     * <p>
+     * This method delegates call to the {@link #beforeQueryStart(RequestMessage)} by default, this functionality
+     * should be preserved by overriding methods if call of the former method is still needed.
+     *
+     * @param msg  the {@link RequestMessage} received by the gremlin-server.
+     * @param user User authenticated in channel processing request
+     */
+    default void beforeQueryStart(final RequestMessage msg, AuthenticatedUser user) {
+        beforeQueryStart(msg);
+    }
+
+    /**
+     * This method is called after the commit / rollback of transaction is called at the end
+     * of query processing. While {@link #onQuerySuccess(RequestMessage)} and {@link #onQueryError(RequestMessage, Throwable)}
+     * methods are called <b>before</b> transaction will be completed.
      *
      * @param msg the {@link RequestMessage} received by the gremlin-server.
+     */
+    default void afterQueryEnd(final RequestMessage msg) {
+    }
+
+    /**
+     * This method will be called if a script or query is processed by the
+     * gremlin-server throws an error.
+     *
+     * @param msg   the {@link RequestMessage} received by the gremlin-server.
      * @param error the exception encountered during processing from the gremlin-server.
      */
     default void onQueryError(final RequestMessage msg, final Throwable error) {
-
     }
 
     /**
@@ -153,6 +176,5 @@ public interface GraphManager {
      * @param msg the {@link RequestMessage} received by the gremlin-server.
      */
     default void onQuerySuccess(final RequestMessage msg) {
-
     }
 }
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
@@ -151,16 +151,6 @@ public interface GraphManager {
     }
 
     /**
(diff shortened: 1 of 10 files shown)


gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
```diff
@@ -125,26 +126,48 @@ public default boolean hasAnyOpenTransactions() {
...
      * This method will be called before a script or query is processed by the
      * gremlin-server.
+     * <p>
+     * This method delegates call to the {@link #beforeQueryStart(RequestMessage)} by default, this functionality
+     * should be preserved by overriding methods if call of the former method is still needed.
+     *
+     * @param msg  the {@link RequestMessage} received by the gremlin-server.
+     * @param user User authenticated in channel processing request
+     */
+    default void beforeQueryStart(final RequestMessage msg, AuthenticatedUser user) {
```
andreachild, line 149: In order to account for evolving lifecycle requirements it would be more flexible for the overloaded method to accept a single query context parameter that can be adapted to contain more information as needed in the future. This way if more data is needed, it need only be added to the query context and no additional overloaded method is needed.

Something like:
```
 /**
   * This method will be called before a script or query is processed by the
   * gremlin-server.
   *
   * @param msg the {@link RequestMessage} received by the gremlin-server.
   * @deprecated replaced by {@link #beforeQueryStart(QueryContext)}
   */
  default void beforeQueryStart(final RequestMessage msg) {
      beforeQueryStart(new QueryContext(msg, null));
  }

  default void beforeQueryStart(final QueryContext context) {
  }
  
  public static class QueryContext {
      private RequestMessage msg;
      private AuthenticatedUser user;
      // other fields can be added in the future without breaking backwards compatibility
      
      public QueryContext(final RequestMessage msg, final AuthenticatedUser user) {
          this.msg = msg;
          this.user = user;
      }
      public RequestMessage getMessage() {
          return msg;
      }
      public AuthenticatedUser getUser() {
          return user;
      }
  }
```

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractEvalOpProcessor.java
```diff
@@ -233,15 +236,23 @@ protected void evalOpInternal(final Context ctx, final Supplier<GremlinExecutor>
...
+                    graphManager.afterQueryEnd(msg);
                 })
                 .beforeEval(b -> {
-                    graphManager.beforeQueryStart(msg);
+                    AuthenticatedUser user = ctx.getChannelHandlerContext().channel().attr(StateKey.AUTHENTICATED_USER).get();
+                    if (null == user) {
+                        // This is expected when using the AllowAllAuthenticator
+                        user = AuthenticatedUser.ANONYMOUS_USER;
+                    }
+                    graphManager.beforeQueryStart(msg, user);
```
andreachild, line 253: I am wondering where the best place to unit or integration test the availability of the 'user' to the lifecycle. Perhaps `AbstractEvalOpProcessorTest.java` (requires a lot of mocking and perhaps stubbing).

andrii0lomakin: I am not sure that I understood your observation.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/GraphManager.java
andrii0lomakin: Makes sense, will update code soon.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/op/AbstractEvalOpProcessor.java
xiazcy: I would also suggest to include some form of testing for this addition, and mocking is probably the best way, though it can get rather complex. I believe Andrea is suggesting to place a set of tests inside of `AbstractEvalOpProcessorTest.java`?

andrii0lomakin: Do you suggest adding a test that checks that the processor redirected the context parameter correctly?

andreachild: I was imagining the test would check that the `AuthenticatedUser` can be accessed by the `beforeQueryStart` method as part of the lifecycle.

andrii0lomakin: @andrii0lomakin Yes, but that is very high level in what scenario? In a scenario, when does it already exist in the channel as an attribute? Or should this test cover all cases of authentication? I believe that such a test case should either exist or be added as a separate change, as that is not the point of this PR to test the correctness of authentication mechanics. As that is a very resource-consuming task, it should be handled separately. I do not mind providing such testing, as security is paramount for us, but if such tests already exist in this case, I will only add checking of the parameters of this callback there. Otherwise, I will test authentication mechanics separately, as we for sure will not go in prod with authentication holes. Do you have any objections?

andreachild: @andrii0lomakin

> that is not the point of this PR to test the correctness of authentication mechanics.

 I agree testing authentication mechanics are not in scope of this PR. It was not my intent to suggest that.

> I will only add checking of the parameters of this callback there

Agree, that is all you should have to do for this change. My initial comment was that I could not immediately identify the best existing test to add this check to.

Cole-Greer:
Hi @andrii0lomakin, do you expect to return to this PR sometime soon? I don't have any concerns other than the comments previously raised by @andreachild.

We are preparing for the 3.7.4 release soon and it would be nice if this change could be included there.

My vote is VOTE +1 (pending resolution of above comments)

andrii0lomakin:
Hi @Cole-Greer , we will return to this PR after migrating to TinkerPop as our frontend. The issue is that the tests you ask for will take a noticeable amount of time, and now we are super busy migrating both the DB and our applications to TinkerPop.

PR #3138, opened 2025-06-20
TINKERPOP-3023 Expand type syntax in grammar - Remove Vertex - TP4

Cole-Greer:
This is a followup to https://github.com/apache/tinkerpop/pull/3133 with changes exclusively for TP4. The primary difference in TP4 is that bytecode has been replaced with GremlinLang, which means that the use of `Vertex` in `V()`, `from()`, and `to()` is now sugar added by GraphTraversal. These vertices must be converted to their ids when passed into GremlinLang.


gremlin-python/src/main/python/gremlin_python/process/traversal.py
```diff
@@ -843,7 +843,7 @@ def _arg_as_string(self, arg):
                 return tmp
 
         if isinstance(arg, Vertex):
-            return f'new ReferenceVertex({self._arg_as_string(arg.id)},\'{arg.label}\')'
+            return f'{self._arg_as_string(arg.id)}'
```
Cole-Greer, line 846: Python implements the Vertex to id sugar at the GremlinLang level instead of special cases for V(), from(), to(), and mergeE(). Would we want to copy this same approach for other GLVs as they adopt GremlinLang or is it best to focus the sugar to specific steps?

Cole-Greer:
VOTE +1

gremlin-python/src/main/python/gremlin_python/process/traversal.py
Cole-Greer: Opted to implement on an individual step bases

PR #3139, opened 2025-06-20
TINKERPOP-3167 Fixed NPE when borrowing a connection

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-3167

There is a chance that the Client.hostConnectionPools can land in a state where it has zero hosts present which leads to the returned ConnectionPool itself to be null. Borrowing a Connection then leads to a NullPointerException. Specifically, the `pool` ends up null [here](https://github.com/apache/tinkerpop/blob/3a48401d83656a31c459f541550b62be67084970/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java#L520) and then the NPE happens next line on `borrowConnection`.  

By preferring a check during init of actual connection pool creation the Client avoids a state where it could grab a null pool. This change doesn't address some of the synchronization issues that likely contributed to allowing the NPE in the first place, but there doesn't seem to be evidence that having an inconsistent state here can lead to issues outside of init().

VOTE +1

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
@@ -433,7 +433,7 @@ public abstract class Client {
      */
     public final static class ClusteredClient extends Client {
 
-        ConcurrentMap<Host, ConnectionPool> hostConnectionPools = new ConcurrentHashMap<>();
+        final ConcurrentMap<Host, ConnectionPool> hostConnectionPools = new ConcurrentHashMap<>();
         private final AtomicReference<CompletableFuture<Void>> closing = new AtomicReference<>(null);
         private Throwable initializationFailure = null;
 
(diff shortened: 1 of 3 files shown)


gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
```diff
@@ -543,9 +543,20 @@ protected void initializeImplementation() {
...
+            // the connection pool seems a bit more fireproof. if we look at initializeConnectionSetupForHost
+            // we can see that a successful initialization of the host/connection pool pair is followed by
+            // marking the host available and notifying the load balancing strategy. by relying directly on
+            // the state of hostConnectionPools we ensure that there is actually a concrete
+            // host/connection pool pair. even if the connection pool has immediate problems, it can fallback
+            // to its normal reconnection operation and won't put chooseConnection in a state where it can
+            // get a NPE if hostConnectionPools ends up being empty. it seems as if the safest minimum
+            // requirement for leaving this method is to ensure that at least one ConnectionPool constructor
+            // completed for at least one Host.
+            if (hostConnectionPools.isEmpty()) {
```
andreachild, line 557: This change seems OK but I am wondering if there is a deeper issue in that `hostConnectionPools.isEmpty()` has somehow become inconsistent with `cluster.availableHosts().isEmpty()`?

xiazcy:
VOTE +1

andreachild:
VOTE +1

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
Cole-Greer: From what I gather, there is no difference here if there is only 1 client per cluster as the connection is marked available in the cluster at the same time that the connection pool is added to `hostConnectionPools` for the client: https://github.com/apache/tinkerpop/blob/3a48401d83656a31c459f541550b62be67084970/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java#L589C1-L592C38

If this change is fixing some NPE error, to me that's implying that something in the ConnectionPool constructor is failing to mark a host as unavailable if the initialization was unsuccessful.

spmallette: they never was consistent as best i can tell. the `Host` availability is `Cluster` wide so it's quite possible that you could have had one `Client` with an active `Host` then try to spin up a new one and it fail to connect because the server is down but not mark them unavailable. We'd wait for the original `Client` to mark it - note the test i modified.

Cole-Greer: The mechanics of that sound a bit strange to me, but I recall that has all been subject to considerable deliberation and testing in the past so I won't weigh in on it here. I agree that given those mechanics, it makes the most sense to use the client's `hostConnectionPools` to verify availability to avoid any influence from other clients on the same cluster.

Cole-Greer:
VOTE +1

PR #3140, opened 2025-06-23
TINKERPOP-3163 fixed CallStep serialization

vkagamlyk:
Fixed serialization of `CallStep`.

https://issues.apache.org/jira/browse/TINKERPOP-3163

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/CallStep.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/CallStep.java
@@ -291,6 +291,7 @@ public final class CallStep<S, E> extends AbstractStep<S, E> implements Traversa
         if (mapTraversal != null)
             this.integrateChild(mapTraversal);
         parameters.getTraversals().forEach(this::integrateChild);
+        ctx = null;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/CallStep.java
```diff
@@ -288,7 +291,6 @@ public void setTraversal(final Traversal.Admin<?, ?> parentTraversal) {
         if (mapTraversal != null)
             this.integrateChild(mapTraversal);
         parameters.getTraversals().forEach(this::integrateChild);
-        ctx = new ServiceCallContext(parentTraversal, this);
```
Cole-Greer, line 291: I think ctx should be set to null here so that it is forcibly regenerated on the next call to `ctx()`. I'm worried about this sequence of events:

```
Traversal.Admin a = __.inject("a").asAdmin();
Traversal.Admin b = __.inject("b").asAdmin();

CallStep step = new CallStep(a, false, "service");
step.ctx(); // generates new ServiceCallContext with Traversal `a`
step.setTraversal(b);
Service.ServiceCallContext context = step.ctx();
context.getTraversal().next(); // returns "a"
```

vkagamlyk: good catch! fixed in https://github.com/apache/tinkerpop/pull/3140/commits/73c183e30bcec48c3eff4a35431dea331f5d5cdd

Cole-Greer:
VOTE +1

xiazcy:
VOTE +1

vkagamlyk:
VOTE +1

PR #3142, opened 2025-06-27
Increase default `max_content_length`/`max_msg_size` in python to 10MB

Cole-Greer:
Back in the [tornado transport days we had set the max content length in python to 10MB](https://github.com/apache/tinkerpop/blob/3.4-dev/gremlin-python/src/main/jython/gremlin_python/driver/tornado/transport.py#L29). 10MB is consistent with the default max frame length in gremlin-server. It seems the 10MB default was lost in the migration to AiohttpTransport, and we've inherited a reduced 4MB limit from AioHTTP.  Bumping it back to 10MB for consistency with gremlin-server.

VOTE +1


xiazcy:
VOTE +1

PR #3156, opened 2025-07-11
Added a tip block to "getting started" tutorial

nnmva:
I would like to suggest this small change to add a tip in the Getting Started Tutorial. I was working through it, as a beginner to gremlin console, and I got stuck. I learned about the `:clear` command, and that helped me to complete it. I think it would be beneficial to other beginners to add this tip in case they find themselves in a similar situation I was in.


andreachild:
VOTE +1

Cole-Greer:
Thanks for the suggestion @nnmva, I think that's a great addition.
VOTE +1

Cole-Greer:
Merging as CTR

PR #2919, opened 2024-11-26
[TINKERPOP-2959] Introduce GValue for Improved Parameterization

Cole-Greer:
https://issues.apache.org/jira/browse/TINKERPOP-2959

The parameterization which was added to gremlin-lang in 3.7 has a key limitation as a result of all variables directly resolving to literals as the script is parsed. As such, the existing parameterization does not allow for any optimizations from caching recurring query plans, as may be seen with gremlin-groovy. The primary benefits of the existing parameterization is to offer users additional flexibility and to increase similarity between gremlin-lang and gremlin-groovy.

This work extends parameterization beyond gremlin-lang to enable optimizations in cases of users executing repetitive query patterns. To accomplish this, a new class `GValue` is introduced which wraps a parameter name and value. Steps which will benefit from parameterization have been given overloads to accept `GValue`'s. Users can pass GValue's into their traversals to inject parameters. When using the new `DefaultVariableResolver`, the grammar will now resolve variables into GValue which are passed to the constructed traversal, instead of directly resolving them to literals. The old variable resolution behavior can still be obtained via the `DirectVariableResolver`.

As GValue has only been explicitly added in places where parameterization is expected to be useful, a corresponding change has been made to the grammar only allow variables in places which will take a GValue.

PR #3157, opened 2025-07-15
GValue and Antlr Parsing Cache

Cole-Greer:
This PR is a redesign and rework of the [previous GValue work](https://github.com/apache/tinkerpop/pull/2919) aimed at TinkerPop 4. If this PR is accepted, it is intended to replace the current GValue implementation in the master branch.

This change introduces GValue, which is essentially a wrapper around a query parameter name and value. The primary goal of GValues is to persist the parameterized state of a traversal through strategy execution. This enables post-strategy Traversals to cached, and for parameterized values to remain mutable after a Traversal is retrieved from the cache.

This PR introduces new step interfaces and "GValue Placeholder" steps for parameterizable steps. A GValue placeholder is a non-executable placeholder step which may store its arguments as GValues. There is a new finalization strategy `GValueReductionStrategy` which is intended to be applied after a traversal has been stored in the cache which "reduces" all placeholder steps to their concrete non-parameterized form. The primary intention of this structure is to prevent GValues from unintentionally leaking into step semantics or traversal results.

Some optimization strategies such as EarlyLimitStrategy may alter a Traversal based on a specific argument value in such a way that is not generalizable for all substitutions of that parameter. To guard against this, each traversal is given a `GValueManager` which tracks if a certain variable has been "pinned". A traversal with a pinned variable is only valid for the current values of that variable, and does not allow for substitution of the pinned variable.

This PR utilizes the GValue improvements to build an optional new cache into GremlinLangScriptEngine which can be used to save on antlr parsing time.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/CallStepContract.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/CallStepContract.java
@@ -31,6 +31,9 @@ public interface CallStepContract<S, E> extends Step<S, E>, Configuring, Travers
 
     String getServiceName();
 
+    /**
+     * TODO: re-evaluate if we should change this to return the static params only
+     */
     Map getMergedParams();
 
     ServiceRegistry getServiceRegistry();
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/CallStepPlaceholder.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/CallStepPlaceholder.java
@@ -18,7 +18,15 @@
  */
 package org.apache.tinkerpop.gremlin.process.traversal.step.map;
 
-import org.apache.tinkerpop.gremlin.process.traversal.Step;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Consumer;
 import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
 import org.apache.tinkerpop.gremlin.process.traversal.Traverser;
 import org.apache.tinkerpop.gremlin.process.traversal.step.GValue;
@@ -27,22 +35,11 @@ import org.apache.tinkerpop.gremlin.process.traversal.step.util.AbstractStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.util.Parameters;
 import org.apache.tinkerpop.gremlin.process.traversal.traverser.TraverserRequirement;
 import org.apache.tinkerpop.gremlin.process.traversal.traverser.util.DummyTraverser;
-import org.apache.tinkerpop.gremlin.process.traversal.traverser.util.TraverserSet;
 import org.apache.tinkerpop.gremlin.process.traversal.util.TraversalUtil;
 import org.apache.tinkerpop.gremlin.structure.Graph;
 import org.apache.tinkerpop.gremlin.structure.service.Service;
 import org.apache.tinkerpop.gremlin.structure.service.ServiceRegistry;
 import org.apache.tinkerpop.gremlin.structure.util.StringFactory;
-
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
 
 /**
  * Reference implementation for service calls via the {@link ServiceRegistry} and {@link Service} APIs. This step
@@ -53,7 +50,6 @@ import java.util.Set;
 public final class CallStepPlaceholder<S, E> extends AbstractStep<S, E> implements CallStepContract<S, E>, GValueHolder<S, E> {
 
     private final boolean isStart;
-    private boolean first = true;
 
     private final String serviceName;
     private GValue<Map<?,?>> staticParams;
@@ -96,6 +92,11 @@ public final class CallStepPlaceholder<S, E> extends AbstractStep<S, E> implemen
         return getServiceRegistry().get(serviceName, isStart, staticParams.get());
     }
 
+    public Service<S, E> serviceGValueSafe() {
+        // throws exception for unrecognized service
+        return getServiceRegistry().get(serviceName, isStart, staticParams.get());
+    }
+
     @Override
     public String getServiceName() {
         return serviceName;
@@ -110,27 +111,36 @@ public final class CallStepPlaceholder<S, E> extends AbstractStep<S, E> implemen
     public Map getMergedParams() {
         if (mapTraversal == null && parameters.isEmpty()) {
             // static params only
-            if(staticParams.isVariable()) {
+            if (staticParams.isVariable()) {
                 traversal.getGValueManager().pinVariable(staticParams.getName());
             }
             return Collections.unmodifiableMap(this.staticParams.get());
         }
 
-        return getMergedParams(new DummyTraverser(this.traversal.getTraverserGenerator()));
+        return getMergedParams(new DummyTraverser<>(this.traversal.getTraverserGenerator()), 
+                gValue -> traversal.getGValueManager().pinVariable(gValue.getName()));
     }
 
-    protected Map getMergedParams(final Traverser.Admin<S> traverser) {
+    public Map getMergedParamsGValueSafe() {
         if (mapTraversal == null && parameters.isEmpty()) {
             // static params only
-            if(staticParams.isVariable()) {
-                traversal.getGValueManager().pinVariable(staticParams.getName());
+            return Collections.unmodifiableMap(this.staticParams.get());
+        }
+        return getMergedParams(new DummyTraverser<>(this.traversal.getTraverserGenerator()), null);
+    }
+
+    protected Map getMergedParams(final Traverser.Admin<S> traverser, Consumer<GValue> gValueConsumer) {
+        if (mapTraversal == null && parameters.isEmpty()) {
+            // static params only
+            if (gValueConsumer != null && staticParams.isVariable()) {
+                gValueConsumer.accept(staticParams);
             }
             return Collections.unmodifiableMap(this.staticParams.get());
         }
 
         // merge dynamic with static params
-        if(staticParams.isVariable()) {
-            traversal.getGValueManager().pinVariable(staticParams.getName());
+        if(gValueConsumer != null && staticParams.isVariable()) {
+            gValueConsumer.accept(staticParams);
         }
         final Map params = new LinkedHashMap(this.staticParams.get());
         if (mapTraversal != null) params.putAll(TraversalUtil.apply(traverser, mapTraversal));
@@ -142,29 +152,6 @@ public final class CallStepPlaceholder<S, E> extends AbstractStep<S, E> implemen
         return params;
     }
 
-    protected Map getMergedParams(final TraverserSet<S> traverserSet) {
-        if (mapTraversal == null && parameters.isEmpty()) {
-            // static params only
-            if(staticParams.isVariable()) {
-                traversal.getGValueManager().pinVariable(staticParams.getName());
-            }
-            return Collections.unmodifiableMap(this.staticParams.get());
-        }
-
-        /*
-         * Dynamic params with a barrier service. We need to reduce to one set of params. For now just disallow
-         * multiple property sets. Also could be sensible to group traversers by parameter set.
-         */
-        final Set<Map> paramsSet = new HashSet<>();
-        for (final Traverser.Admin<S> traverser : traverserSet) {
-            paramsSet.add(getMergedParams(traverser));
-        }
-        if (paramsSet.size() > 1) {
-            throw new UnsupportedOperationException("Cannot use multiple dynamic parameter sets with a barrier service call.");
-        }
-        return paramsSet.iterator().next();
-    }
-
     @Override
     public ServiceRegistry getServiceRegistry() {
         final Graph graph = (Graph) this.traversal.getGraph().get();
@@ -174,7 +161,6 @@ public final class CallStepPlaceholder<S, E> extends AbstractStep<S, E> implemen
     @Override
     public void reset() {
         super.reset();
-        first = true;
         if (mapTraversal != null)
             mapTraversal.reset();
         parameters.getTraversals().forEach(Traversal.Admin::reset);
@@ -247,7 +233,7 @@ public final class CallStepPlaceholder<S, E> extends AbstractStep<S, E> implemen
     }
 
     @Override
-    public Step<S, E> asConcreteStep() {
+    public CallStep<S, E> asConcreteStep() {
         CallStep<S, E> step = new CallStep<>(traversal, isStart, serviceName, staticParams.get(), mapTraversal);
         for (Map.Entry<Object, List<Object>> entry : parameters.getRaw().entrySet()) {
             for (Object value : entry.getValue()) {
--- a/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/CallStepTest.java
+++ b/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/CallStepTest.java
@@ -18,28 +18,50 @@
  */
 package org.apache.tinkerpop.gremlin.process.traversal.step.map;
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.apache.commons.configuration2.MapConfiguration;
 import org.apache.commons.lang3.SerializationUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.__;
 import org.apache.tinkerpop.gremlin.process.traversal.step.GValue;
 import org.apache.tinkerpop.gremlin.process.traversal.step.GValueStepTest;
 import org.apache.tinkerpop.gremlin.process.traversal.util.DefaultTraversal;
 import org.apache.tinkerpop.gremlin.structure.service.Service;
+import org.apache.tinkerpop.gremlin.structure.service.ServiceRegistry;
+import org.apache.tinkerpop.gremlin.structure.util.GraphFactoryTest;
+import org.junit.Ignore;
+import org.junit.Rule;
 import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
 
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
+import static org.apache.tinkerpop.gremlin.process.traversal.AnonymousTraversalSource.traversal;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.when;
 
 public class CallStepTest extends GValueStepTest {
 
+    private static final String TEST_SERVICE = "test-service";
+    private static final Map<String, String> STATIC_PARAMS = Map.of("foo", "bar", "fizz", "fuzz");
+    @Mock
+    public ServiceRegistry mockedRegistry;
+    @Mock
+    public Service<?, ?> mockedService;
+    @Rule
+    public MockitoRule mockitoRule = MockitoJUnit.rule();
+    
     @Override
     protected List<Traversal> getTraversals() {
         return Arrays.asList(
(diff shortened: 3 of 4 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -37,10 +43,27 @@ public class P<V> implements Predicate<V>, Serializable, Cloneable {
...
     protected V value;
     protected V originalValue;
+    protected GValue<V> gValue;
+    protected GValue<V>[] gValues;
+
+    public P(final PBiPredicate<V, V> biPredicate, V value) {
+        if (value instanceof GValue) {
+            this.gValue = (GValue<V>) value;
+            this.value = ((GValue<V>) value).get();
+        } else if (value instanceof List && ((List) value).stream().anyMatch(v -> v instanceof GValue)) {
```
spmallette: Is `List` too restrictive here - perhaps `Collection` is more appropriate?:

```
gremlin> g.V().has('name', within(['josh','vadas'] as Set))
==>v[2]
==>v[4]
```

or maybe we need to retain the original `Collection` implementation to some degree? not sure if that is needed by some providers or not, but might be better than reconstructing every `Collection` to a `List`.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -269,4 +292,45 @@ public static P test(final PBiPredicate biPredicate, final Object value) {
     public static <V> P<V> not(final P<V> predicate) {
         return predicate.negate();
     }
+
+    public boolean isParameterized() {
+        if (gValue != null && gValue.isVariable()) {
+            return true;
+        }
+        if (gValues != null) {
+            for (final GValue<V> gValue : gValues) {
```
spmallette: I'm wary of loops in places where I know strategies will tread. Envisioning a large `P.within(...)` creating a bottleneck. Does this have to be determined dynamically? Could parameterization be determined once at construction?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/TraversalStrategy.java
```diff
@@ -31,21 +33,37 @@
...
  * <p/>
- * A traversal strategy should be a final class as various internal operations on a strategy are based on its ability to be assigned to more general classes.
- * A traversal strategy should typically be stateless with a public static <code>instance()</code> method.
- * However, at limit, a traversal strategy can have a state defining constructor (typically via a "builder"), but that state can not mutate once instantiated.
+ * A traversal strategy should be a final class as various internal operations on a strategy are based on its ability to
+ * be assigned to more general classes. A traversal strategy should typically be stateless with a public static
+ * <code>instance()</code> method. However, at limit, a traversal strategy can have a state-defining constructor
+ * (typically via a "builder"), but that state cannot mutate once instantiated.
+ * <p/>
+ * Given that a traversal strategy can completely rewrite a traversal, it must take into account the
```
spmallette, line 49: I think this javadoc is not completely accurate anymore and needs update.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -41,13 +41,23 @@
...
 import org.apache.tinkerpop.gremlin.process.traversal.lambda.PredicateTraverser;
 import org.apache.tinkerpop.gremlin.process.traversal.lambda.TrueTraversal;
 import org.apache.tinkerpop.gremlin.process.traversal.step.ByModulating;
 import org.apache.tinkerpop.gremlin.process.traversal.step.Configuring;
+import org.apache.tinkerpop.gremlin.process.traversal.step.filter.IsStepPlaceholder;
+import org.apache.tinkerpop.gremlin.process.traversal.step.filter.RangeStepPlaceholder.RangeGlobalStepPlaceholder;
+import org.apache.tinkerpop.gremlin.process.traversal.step.filter.RangeStepPlaceholder.RangeLocalStepPlaceholder;
+import org.apache.tinkerpop.gremlin.process.traversal.step.filter.TailGlobalStepPlaceholder;
+import org.apache.tinkerpop.gremlin.process.traversal.step.filter.TailLocalStepPlaceholder;
+import org.apache.tinkerpop.gremlin.process.traversal.step.map.*;
```
spmallette: fix asterisk import

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -2852,6 +3410,9 @@ public default GraphTraversal<S, E> drop() {
      * @since 3.7.1
      */
     public default <S2> GraphTraversal<S, E> all(final P<S2> predicate) {
+        if (predicate != null && predicate.isParameterized()) {
+            throw new IllegalArgumentException("Parameterized predicates are not supported by all()"); //TODO:: should this be supported?
```
spmallette: noticed a few TODOs with this question around use of `GValue` with `P` in these cases and I'd say: no, shouldn't be supported right now. can always be added later. i think you could remove the TODOs.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -3991,6 +4574,61 @@ public default <E2> GraphTraversal<S, E> option(final Traversal<?, E2> traversal
...
+     * @param token       the token that would trigger this option which may be a {@link Pick}, {@link Merge},
+     *                    a {@link Traversal}, {@link Predicate}, or object depending on the step being modulated.
+     * @param traversalOption the option as a traversal
+     * @return the traversal with the modulated step
+     * @see <a href="http://tinkerpop.apache.org/docs/${project.version}/reference/#choose-step" target="_blank">Reference Documentation - Choose Step</a>
+     * @see <a href="http://tinkerpop.apache.org/docs/${project.version}/reference/#mergev-step" target="_blank">Reference Documentation - MergeV Step</a>
+     * @see <a href="http://tinkerpop.apache.org/docs/${project.version}/reference/#mergee-step" target="_blank">Reference Documentation - MergeE Step</a>
+     * @since 3.8.0
+     */
+    public default <M, E2> GraphTraversal<S, E> option(final GValue<M> token, final Traversal<?, E2> traversalOption) {
```
spmallette: Do we need this overload? In what context is ` GValue<M> token` used? can this be removed?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversalSource.java
```diff
@@ -342,7 +347,21 @@ public GraphTraversal<Vertex, Vertex> addV() {
...
         final GraphTraversal.Admin<Vertex, Vertex> traversal = new DefaultGraphTraversal<>(clone);
-        return traversal.addStep(new AddVertexStartStep(traversal, (String) null));
+        return traversal.addStep(new AddVertexStartStepPlaceholder(traversal, (String) null));
+    }
+
+    /**
+     * Spawns a {@link GraphTraversal} by adding a vertex with the specified label. If the {@code label} is
+     * {@code null} then it will default to {@link Vertex#DEFAULT_LABEL}.
+     *
+     * @since 4.0.0
```
spmallette: check for other `@since` annotations using 4.0.0 - should be 3.8.0.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -5233,6 +5233,60 @@ location. Please see the <<io-step,documentation>> for `io()`-step for more comp
...
+not have such a query caching mechanism, and thus will not see any performance improvements through parameterization. Users
+should consult the documentation of their specific graph system details of potential performance benefits via parameterization.
+
+NOTE: There are unique considerations regarding parameters when using `gremlin-groovy` scripts. Groovy allows for parameterization
+at arbitrary points in the query in addition to the subset of parameterizable steps documented here. Groovy is also bound by
+a comparatively slow script compilation, which makes parameterization essential for performant execution of `gremlin-groovy` scripts.
+
+[cols="1,1"]
+|===
+|Step | Parameterizable arguments
```
spmallette, line 5387: lots of steps listed here. is this really the step list for 3.8.0? i guess i'm questioning: `coin`, `combine`, `conjoin`, `constant`, `difference`, `disjunct`, `intersect`, `merge`, `product`. are those really working with the `GValue` style the rest of the steps have and is there testing for all that? Then, why that particular set but not also `all`, `any`, etc.? Need to remove some steps for this initial body of work?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/Mutating.java
```diff
@@ -28,7 +28,7 @@
  * @author Stephen Mallette (http://stephen.genoprime.com)
  * @author Matt Frantz (http://github.com/mhfrantz)
  */
-public interface Mutating<E extends Event> extends Configuring {
+public interface Mutating<E extends Event> {
```
spmallette, line 31: This change should be in the CHANGELOG and Upgrade Docs (providers section) as it might break strategies.

Cole-Greer: This one is already captured in the changelog. The upgrade docs definitely need a do-over before this can be merged. My intention is to accumulate all significant changes in the changelog throughout the review process, and then use that as a reminder/guide for which changes will need upgrade docs as well.

docs/src/reference/the-traversal.asciidoc
Cole-Greer: Yes this is an old list from the master backport, will need to be revised to the current set of steps.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
Cole-Greer: This may have been a holdover from when we intended to put GValue in more places. I agree it doesn't fit into the current design.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
Cole-Greer: `P.within` and `P.without` do accept `Collection` so I agree this handling needs to be broadened to account for that. I can certainly retain the original collection type as well, however some properties such as element order may be lost as GValues are replaced with scalars.

Cole-Greer: Retaining the original `Collection` ran into more trouble than I anticipated. This PR has `P` maintain parallel lists for both the GValue and scalar forms of `value`. When `updateVariable()` is called, the index of a matched GValue is used to determine which value in the scalar list also needs updating. This approach falls apart if we want to store non-indexed collections.

This, along with some fragility concerns regarding the parallel arrays approach led me to consider introducing a `PInterface` and `GValueP` which can be reduced to a regular `P` similarly to how the step placeholders work. I've built that out to be about 95% right and pushed it to a [separate branch](https://github.com/apache/tinkerpop/tree/GValue-3.8-PInterface), although I'm not sure it's the right approach as it brings so much more change.

I'm going to put a pin in this for the moment and focus on refinements elsewhere in this PR, and I will come back and reconsider the P solution afterwards.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
Cole-Greer: I've moved this calculation to the constructor, and anytime setValue() is called.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -5254,6 +5254,60 @@ location. Please see the <<io-step,documentation>> for `io()`-step for more comp
 
 link:++https://tinkerpop.apache.org/javadocs/x.y.z/full/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.html#write()++[`write()`]
 
+[[traversal-parameterization]]
+== Traversal Parameterization
+
+A subset of gremlin steps are able to accept parameterized arguments also known as GValues. GValues can be used to
+provide protection against gremlin-injection attacks in cases where untrusted and unsanitized inputs must be passed as
+step arguments. Additionally, use of GValues may offer performance benefits in certain environments by making use of
+some query caching capabilities. Note that the reference implementation of the gremlin language and `gremlin-server` do
```
Cole-Greer, line 5377: Reminder: Document the new script cache in GremlinLangScriptEngine and update this section with an appropriate reference.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java
```diff
@@ -0,0 +1,169 @@
...
+     */
+    public Set<String> getUnpinnedVariableNames() {
+        Set<String> variableNames = new HashSet<>();
+        variableNames.addAll(gValueRegistry.keySet());
+        variableNames.removeAll(pinnedGValues.keySet());
+        return Collections.unmodifiableSet(variableNames);
+    }
+
+    /**
+     * Gets the set of variable names used in this traversal which have not been pinned to specific values.
```
andreachild:
```suggestion
     * Gets the set of variable names used in this traversal which have been pinned to specific values.
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java
```diff
@@ -0,0 +1,169 @@
...
+     * Determines whether the manager has unpinned GValues.
+     */
+    public boolean hasUnpinnedVariables() {
+        return !getUnpinnedVariableNames().isEmpty();
+    }
+
+    /**
+     * Pins any GValue with the provided name. This should be called anytime an optimization
+     * alters the Traversal based on the current value of a GValue. This indicates that the resulting traversal is only
+     * valid for this specific value of a GValue, and is not generalizable to any parameter value.
```
andreachild:
```suggestion
     * valid for the current value of this GValue, and is not generalizable to any parameter value.
```

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/GValueTest.java
```diff
@@ -0,0 +1,418 @@
...
+        final Object returnedGValue = GValue.of(gValue);
+        assertEquals(gValue, returnedGValue);
+        assertSame(gValue, returnedGValue);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void shouldReturnAnExistInTypedGValue() {
+        final Object gValue = GValue.of("x", 123);
+        GValue.of("x", gValue);
+    }
```
xiazcy, line 60: Good to add exception cases where "_" and nested GValue are used. Also case where getName() is null.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/GValueTest.java
```diff
@@ -0,0 +1,418 @@
...
+    public void valueInstanceOfShouldReturnFalseForNonGValueObject() {
+        String nonGValue = "test";
+        assertThat(GValue.valueInstanceOf(nonGValue, String.class), is(false));
+    }
+
+    @Test
+    public void valueInstanceOfShouldReturnFalseForNullObject() {
+        assertThat(GValue.valueInstanceOf(null, String.class), is(false));
+    }
+
```
xiazcy, line 294: Potentially also test that the `gValue.get()` preserves type.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManagerTest.java
```diff
@@ -0,0 +1,172 @@
...
+import org.junit.Test;
+
+import java.util.Set;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+
+public class GValueManagerTest {
+
```
xiazcy, line 36: Is there a test for accessing a locked manager? Also potentially add concurrent access tests, e.g. race conditions.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManagerTest.java
```diff
@@ -0,0 +1,172 @@
...
+    @Test
+    public void shouldMergeIntoAnotherManager() {
+        final GValueManager sourceManager = new GValueManager();
+        final GValueManager targetManager = new GValueManager();
```
xiazcy, line 40: Should add the case where merging two GValueManagers have overlapping variable names.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/ArgumentVisitor.java
```diff
@@ -37,60 +33,16 @@ public ArgumentVisitor(final VariableResolver resolver, final GremlinAntlrToJava
...
-    public Map parseMap(final GremlinParser.GenericMapArgumentContext ctx) {
-        return (Map) visitGenericMapArgument(ctx);
-    }
-
     /**
      * Wrapper for visit function for {@code Map} types.
      */
     public Map parseMap(final GremlinParser.GenericMapNullableArgumentContext ctx) {
-        return (Map) visitGenericMapNullableArgument(ctx);
+        Object literalOrVar = visitGenericMapNullableArgument(ctx);
```
xiazcy:
```suggestion
        final Object literalOrVar = visitGenericMapNullableArgument(ctx);
```
Note to check missing `final` for rest of these visitor methods.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/DefaultGremlinBaseVisitor.java
```diff
@@ -410,11 +410,11 @@ protected void notImplemented(final ParseTree ctx) {
 	/**
 	 * {@inheritDoc}
 	 */
-	@Override public T visitTraversalMethod_from_Traversal(final GremlinParser.TraversalMethod_from_TraversalContext ctx) { notImplemented(ctx); return null; }
+	@Override public T visitTraversalMethod_from_GenricArgument(GremlinParser.TraversalMethod_from_GenricArgumentContext ctx) { return null; }
```
xiazcy:
```suggestion
	@Override public T visitTraversalMethod_from_GenricArgument(final GremlinParser.TraversalMethod_from_GenricArgumentContext ctx) { return null; }
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -35,28 +46,41 @@
...
-    protected V value;
-    protected V originalValue;
+    protected Map<String, V> variables = new HashMap<>();
+    protected Collection<V> literals = Collections.EMPTY_LIST;
+    private boolean isCollection = false;
 
-    public P(final PBiPredicate<V, V> biPredicate, final V value) {
-        this.value = value;
-        this.originalValue = value;
+    public P(final PBiPredicate<V, V> biPredicate, V value) {
```
xiazcy:
```suggestion
    public P(final PBiPredicate<V, V> biPredicate, final V value) {
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -35,28 +46,41 @@
...
+        setValue(value);
         this.biPredicate = biPredicate;
     }
 
+    public P(final PBiPredicate<V, V> biPredicate, GValue<V> value) {
```
xiazcy:
```suggestion
    public P(final PBiPredicate<V, V> biPredicate, final GValue<V> value) {
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -90,17 +156,18 @@ public boolean equals(final Object other) {
...
                 ((P) other).getBiPredicate().equals(this.biPredicate) &&
-                ((((P) other).getOriginalValue() == null && this.originalValue == null) || ((P) other).getOriginalValue().equals(this.originalValue));
+                ((((P) other).variables == null && this.variables == null) || (((P) other).variables != null && ((P) other).variables.equals(this.variables))) &&
+                ((((P) other).literals == null && this.literals == null) || (((P) other).literals != null && CollectionUtils.isEqualCollection(((P) other).literals, this.literals)));
     }
 
     @Override
     public String toString() {
-        return null == this.originalValue ? this.biPredicate.toString() : this.biPredicate.toString() + "(" + this.originalValue + ")";
+        return literals.isEmpty() && variables.isEmpty() ? this.biPredicate.toString() : this.biPredicate.toString() + "(" + getValue() + ")";
```
xiazcy: Not sure if there's any actual impact, but just to note that given we aren't using `null` as a condition, an edge case of consequence is that previously`P.eq(null).toString()` would output `eq`, and now it will be `eq(null)`.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -384,7 +416,27 @@ public default GraphTraversal<S, Edge> E(final Object... edgeIdsOrElements) {
...
+        return this.asAdmin().addStep(step);
+    }
+
+    /**
+     * Map the {@link Vertex} to its adjacent vertices given a direction.
+     *
+     * @param direction  the direction to traverse from the current vertex
+     * @return the traversal with an appended {@link VertexStep}.
+     * @see <a href="http://tinkerpop.apache.org/docs/${project.version}/reference/#vertex-steps" target="_blank">Reference Documentation - Vertex Step</a>
+     * @since 3.0.0-incubating
```
xiazcy, line 437:
```suggestion
     * @since 3.8.0
```
Nit: this overload (along with some others below) looks to be newly added?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java
```diff
@@ -0,0 +1,169 @@
...
+ * Note that the manager can be locked, at which point it becomes immutable, and any attempt to modify it will result
+ * in an exception.
+ */
+public final class GValueManager implements Serializable, Cloneable {
+
+    private final Map<String, GValue<?>> gValueRegistry = new HashMap<>();
+    private final Map<String, GValue<?>> pinnedGValues = new HashMap<>();
+
+    public GValueManager() {
+        this(Collections.EMPTY_MAP, Collections.EMPTY_MAP);
```
andreachild: Nit: to avoid type safety warnings

```suggestion
        this(Collections.emptyMap(), Collections.emptyMap());
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java
```diff
@@ -0,0 +1,169 @@
...
+    public Set<String> getVariableNames() {
+        return Collections.unmodifiableSet(gValueRegistry.keySet());
+    }
+
+    /**
+     * Gets the set of variable names used in this traversal which have not been pinned to specific values.
+     */
+    public Set<String> getUnpinnedVariableNames() {
+        Set<String> variableNames = new HashSet<>();
+        variableNames.addAll(gValueRegistry.keySet());
```
andreachild: Nit:
```suggestion
        Set<String> variableNames = new HashSet<>(gValueRegistry.keySet());
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java
```diff
@@ -0,0 +1,169 @@
...
+        variableNames.removeAll(pinnedGValues.keySet());
+        return Collections.unmodifiableSet(variableNames);
+    }
+
+    /**
+     * Gets the set of variable names used in this traversal which have not been pinned to specific values.
+     */
+    public Set<String> getPinnedVariableNames() {
```
andreachild, line 89: This method seems unused - is it intended to be used by providers? If yes then we can add unit tests.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java
```diff
@@ -0,0 +1,169 @@
...
+     * Gets the set of {@link GValue} objects used in this traversal.
+     */
+    public Set<GValue<?>> getGValues() {
+        return Collections.unmodifiableSet(new HashSet(gValueRegistry.values()));
+    }
+
+    /**
+     * Gets the set of pinned GValues used in this traversal.
+     */
+    public Set<GValue<?>> getPinnedGValues() {
```
andreachild, line 103: This method seems unused - is it intended to be used by providers? If yes then we can add unit tests.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java
```diff
@@ -0,0 +1,169 @@
...
+     */
+    public Set<String> getPinnedVariableNames() {
+        return Collections.unmodifiableSet(pinnedGValues.keySet());
+    }
+
+    /**
+     * Gets the set of {@link GValue} objects used in this traversal.
+     */
+    public Set<GValue<?>> getGValues() {
+        return Collections.unmodifiableSet(new HashSet(gValueRegistry.values()));
```
andreachild, line 93: Nit:
```suggestion
        return Set.copyOf(gValueRegistry.values());
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java
```diff
@@ -0,0 +1,169 @@
...
+        //TODO needs test
+        other.register(gValueRegistry.values());
+        other.pinGValues(pinnedGValues.values());
+    }
+
+    /**
+     * Gets the set of variable names used in this traversal.
+     */
+    public Set<String> getVariableNames() {
+        return Collections.unmodifiableSet(gValueRegistry.keySet());
```
andreachild, line 69: Nit:
```suggestion
        return Set.copyOf(gValueRegistry.keySet());
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /**
+     * Gets the set of variable names used in this traversal which have not been pinned to specific values.
+     */
+    public Set<String> getUnpinnedVariableNames() {
+        Set<String> variableNames = new HashSet<>();
+        variableNames.addAll(gValueRegistry.keySet());
+        variableNames.removeAll(pinnedGValues.keySet());
+        return Collections.unmodifiableSet(variableNames);
```
andreachild, line 79: Nit: 
```suggestion
        return Set.copyOf(variableNames);
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java
```diff
@@ -0,0 +1,169 @@
...
+ * The {@code GValueManager} class is responsible for managing the state of {@link GValue} instances and their
+ * associations with `Step` objects in a traversal. This class ensures that `GValue` instances are properly extracted
+ * and stored in a registry, allowing for dynamic query optimizations and state management during traversal execution.
+ * Note that the manager can be locked, at which point it becomes immutable, and any attempt to modify it will result
+ * in an exception.
+ */
+public final class GValueManager implements Serializable, Cloneable {
+
+    private final Map<String, GValue<?>> gValueRegistry = new HashMap<>();
+    private final Map<String, GValue<?>> pinnedGValues = new HashMap<>();
```
andreachild: It doesn't look like there's a need to have a reference to the `GValue` here as we just need to track which ones have been pinned.
```suggestion
    /**
     * Tracks pinned values for faster lookup.
     */
    private final Set<String> pinnedGValues = new HashSet<>();
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/HasContainerHolder.java
```diff
@@ -34,4 +40,30 @@ public interface HasContainerHolder {
     public default void removeHasContainer(final HasContainer hasContainer) {
         throw new UnsupportedOperationException("The holder does not support container removal: " + this.getClass().getCanonicalName());
     }
+
+    public default Collection<P<?>> getPredicates() {
+        return getHasContainers().stream().map(p -> p.getPredicate()).collect(Collectors.toList());
```
Cole-Greer, line 54: Should pin variables for all predicates

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/GraphStepPlaceholder.java
```diff
@@ -0,0 +1,158 @@
...
+        return this.returnClass.equals(Vertex.class);
+    }
+
+    @Override
+    public boolean returnsEdge() {
+        return this.returnClass.equals(Edge.class);
+    }
+
+    @Override
+    public Object[] getIds() {
```
Cole-Greer, line 82: Missing `getIdsGValueSafe()`

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManagerTest.java
Cole-Greer: Locking the manager has been removed now so no tests are necessary there. I don't believe concurrent access tests are relevant here as `GraphTraversal` itself is not tested for concurrent access and should not be considered thread safe. There is one manager per traversal, and historically it has not been considered necessary to multiple threads interacting with a single traversal.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
Cole-Greer: I considered that earlier. I think `eq(null)` is more clear than the previous behaviour although it is a slight break. I don't mind changing this to something like this to avoid the breaking change though:
```
return null == this.getValue() ? this.biPredicate.toString() : this.biPredicate.toString() + "(" + this.getValue() + ")";

File: gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java (hunk 25 above)
Comment (Cole-Greer) [reply gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java:89]: Yes it is intended for providers, it will need unit tests

File: gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java (hunk 26 above)
Comment (Cole-Greer) [reply gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java:103]: Yes it is intended for providers, it will need unit tests

File: gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java (hunk 21 above)
Comment (Cole-Greer) [reply gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java]: I opted to revert this back to the old behaviour via:
```
return null == this.getValue() ? this.biPredicate.toString() : this.biPredicate.toString() + "(" + this.getValue() + ")";
```
I still think `eq(null)` may be more clear, but there is enough change already contained in this PR that I'd like to avoid unnecessary changes where possible.

Cole-Greer:
I think this work will still benefit from some extra testing and refinement, although I think the core of it is very solid now. Given the size of this PR, I'd like to move towards merging this rather soon, with the understanding that there will be some follow up PRs which will add additional testing and minor improvements. I believe it will be much easier to manage these subsequent smaller PRs instead of continuing to add to this giant one.

VOTE +1

andreachild:
VOTE +1 given there will be some additional testing and minor improvement follow ups

xiazcy:
VOTE +1

Cole-Greer:
I'm not going to pull any of these changes into master on the initial merge here due to many anticipated conflicts. I will followup with a future PR which cherry-picks these changes in and resolves all conflicts.

Cole-Greer:
Merged to master in https://github.com/apache/tinkerpop/commit/3a142ae52c2a60b678ab81612fdaf2a1276e23ee. There were plenty of conflicts to resolve in the merge but no significant departures from the 3.8 implementation.

PR #3159, opened 2025-07-17
TINKERPOP-3168 Deprecated UnifiedChannelizer

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-3168

VOTE +1

### DIFF
--- a/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java
+++ b/gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java
@@ -218,7 +218,9 @@ public class Settings {
      * configuration only applies to the {@link UnifiedChannelizer}. By default this value is set to 4096.
      *
      * @see #maxWorkQueueSize
+     * @deprecated As of release 3.8.0, not replaced.
      */
+    @Deprecated
     public int maxSessionTaskQueueSize = 4096;
 
     /**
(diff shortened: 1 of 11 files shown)


gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/Settings.java
```diff
@@ -210,21 +210,29 @@ public Settings() {
      * created {@link Session} will queue separately given that setting per session.
      * <p/>
      * By default this value is set to 8192.
+     *
+     * @deprecated As of release 3.8.0, not replaced.
      */
+    @Deprecated
```
andreachild: Should this be deprecated since I still see this setting in master branch?

spmallette: hmm, not sure...looks like its been repurposed for HTTP. not sure i've ever had that before. in a way it's deprecated in the context of how it is being used here so in that way it makes sense, but the fact that as an API element it's not going to be removed makes it not deprecated. 

i guess i'd leave it deprecated but if you/others feel it should not be, i can change it.

andreachild: I'm in favour of not deprecating it as its usage in `ServerGremlinExecutor.java` has not changed from 3.8 to master. What has changed is only its relevance to sessions, hence the removal of the javadoc section below:

> This value should be taken in account with the {@link #maxSessionTaskQueueSize} which is related in some respects. A request that starts a new {@link Session} is handled by this queue, but additional requests to a created {@link Session} will queue separately given that setting per session.

spmallette: fixed it

andreachild:
VOTE +1

vkagamlyk:
VOTE +1

Cole-Greer:
VOTE +1

PR #3161, opened 2025-07-18
TINKERPOP-3141 Allow drop add vertex in same transaction

andreachild:
https://issues.apache.org/jira/browse/TINKERPOP-3141

Account for the scenario of dropping and re-adding a vertex within the same transaction by adding a new `unmarkDeleted` method to the `TinkerElementContainer` and calling `unmarkDeleted` from `addVertex` if it is detected that the vertex had been previously deleted. Added logic to reference the previously deleted vertex version when adding the vertex back to avoid transaction conflict error.

### DIFF
--- a/tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraph.java
+++ b/tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraph.java
@@ -152,11 +152,19 @@ public final class TinkerTransactionGraph extends AbstractTinkerGraph {
         if (container != null && container.get() != null)
             throw Exceptions.vertexWithIdAlreadyExists(idValue);
 
+        long version = txNumber;
+        if (container != null && container.isDeleted() && container.getModified() != null) {
+            // vertex being added was previously deleted 
+            // we need to reference the version from the deleted state when adding the vertex back
+            version = container.getModified().version();
+            container.unmarkDeleted((TinkerTransaction) tx());
+        }
+        
         // no existing container, let's use new one
         if (container == null)
             container = newContainer;
 
-        final TinkerVertex vertex = new TinkerVertex(idValue, label, this, txNumber);
+        final TinkerVertex vertex = new TinkerVertex(idValue, label, this, version);
         ElementHelper.attachProperties(vertex, VertexProperty.Cardinality.list, keyValues);
         container.setDraft(vertex, (TinkerTransaction) tx());
 
--- a/tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraphTest.java
+++ b/tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraphTest.java
@@ -18,10 +18,12 @@
  */
 package org.apache.tinkerpop.gremlin.tinkergraph.structure;
 
+import java.util.List;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
 import org.apache.tinkerpop.gremlin.structure.Edge;
 import org.apache.tinkerpop.gremlin.structure.T;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
+import org.apache.tinkerpop.gremlin.structure.VertexProperty;
 import org.apache.tinkerpop.gremlin.structure.util.TransactionException;
 import org.junit.Test;
 
@@ -33,6 +35,7 @@ import java.util.concurrent.atomic.AtomicLong;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
@@ -40,6 +43,7 @@ import static org.junit.Assert.fail;
 
 public class TinkerTransactionGraphTest {
 
+    private static final String PROPERTY_NAME = "name";
     final Object vid = 100;
 
     ///// vertex tests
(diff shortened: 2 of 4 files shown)


andreachild:
@vkagamlyk would appreciate your feedback on this PR if you have the time 🙂

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraph.java
```diff
@@ -152,11 +152,19 @@ public Vertex addVertex(final Object... keyValues) {
         if (container != null && container.get() != null)
             throw Exceptions.vertexWithIdAlreadyExists(idValue);
 
+        long version = txNumber;
+        if (container != null && container.isDeleted() && container.getModified() != null) {
```
vkagamlyk, line 156: I think this block should be a few lines higher, before throwing the error, something like
```
if (container != null && container.isDeleted())
    container.unmarkDeleted((TinkerTransaction) tx());
else if (container != null && container.get() != null)
    throw Exceptions.vertexWithIdAlreadyExists(idValue);
```

vkagamlyk:
> @vkagamlyk would appreciate your feedback on this PR if you have the time 🙂

Hi Andrea,
thanks for trying to fix my bug :)

tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraphTest.java
```diff
@@ -1503,4 +1504,53 @@ public void shouldHandleCorrectlyHandleCountForChangedAndReadElement() {
...
+        gtx2.V().drop().iterate();
+        gtx2.addV().property(T.id, vertexId).next();
+        gtx2.tx().commit();
+
+        verifyCommittedSingleVertexWithId(g, vertexId);
+        final TinkerElementContainer<TinkerVertex> container = g.getVertices().get(vertexId);
+        final TinkerVertex updatedVertex = container.getUnmodified();
+        assertNull(updatedVertex.properties);
+        // version should have been updated
+        assertNotEquals(originalVersion, updatedVertex.version());
```
Cole-Greer: Would be nice to throw in an edge on the vertex and ensure the edge remains dropped after gtx2 is committed
```suggestion
        gtx.addV().property(T.id, vertexId).property("name", "test").next();
        gtx.addE().from(V(vertexId)).to(V(vertexId))
        gtx.tx().commit();

        verifyCommittedSingleVertexWithId(g, vertexId);
        final TinkerVertex originalVertex = g.getVertices().get(vertexId).getUnmodified();
        assertEquals("test", originalVertex.properties.get("name").get(0).value());
        assertEquals(1, g.getEdgesCount());
        final long originalVersion = originalVertex.version();

        // drop the vertex and re-create it without any properties in the same transaction
        final GraphTraversalSource gtx2 = tx.begin();
        gtx2.V().drop().iterate();
        gtx2.addV().property(T.id, vertexId).next();
        gtx2.tx().commit();

        verifyCommittedSingleVertexWithId(g, vertexId);
        final TinkerElementContainer<TinkerVertex> container = g.getVertices().get(vertexId);
        final TinkerVertex updatedVertex = container.getUnmodified();
        assertNull(updatedVertex.properties);
        assertEquals(0, g.getEdgesCount());
        // version should have been updated
        assertNotEquals(originalVersion, updatedVertex.version());
```

Cole-Greer:
LGTM VOTE +1

tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraphTest.java
```diff
@@ -1503,4 +1504,53 @@ public void shouldHandleCorrectlyHandleCountForChangedAndReadElement() {
 
         assertFalse(vertex.inUse());
     }
+
+    @Test
+    public void shouldAllowDropThenInsertInSameTransaction() {
```
vkagamlyk: test scenarios that can be considered for reliability:
- add and remove vertex with same ID in single transaction. It can be 4 tests, combinations like add-remove-add, add-remove-add-remove,  remove-add-remove-add, remove-add-remove. Existing tests covers only cases like add-remove, for example `shouldDeleteCreatedVertexOnCommit`
- some multi-threaded test case, like remove-add in main thread (without commit) and remove-add-remove in another thread. Example of similar test is `shouldDeleteUnusedVertexContainerOnConcurrentVertexDelete`

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraph.java
```diff
@@ -152,11 +152,19 @@ public Vertex addVertex(final Object... keyValues) {
...
+            version = container.getModified().version();
+            container.unmarkDeleted((TinkerTransaction) tx());
+        }
+        
         // no existing container, let's use new one
         if (container == null)
             container = newContainer;
 
-        final TinkerVertex vertex = new TinkerVertex(idValue, label, this, txNumber);
+        final TinkerVertex vertex = new TinkerVertex(idValue, label, this, version);
```
vkagamlyk, line 167: If the element is changed in a transaction, the version must match the transaction number.

andreachild: From my understanding the transaction number is specifically not set on the vertex until commit is called and until then, the vertex should retain the original version (except when adding a brand new vertex which will by default have a version set to the transaction number).

```
public void commit(final long txVersion) {
    updateUsesCount();
    if (isDeletedInTx.get()) {
        // created and deleted in same tx
        if (null != element)
            element.removed = true;
        element = null;
        isDeleted = true;
    } else if (isModifiedInTx.get()){
        element = transactionUpdatedValue.get();
        element.currentVersion = txVersion;
    }
    reset();
}
```

Did I misunderstand?

vkagamlyk: I think you're right, for scenario remove-add in same transaction better to preserve version, so `updatedOutsideTransaction()` will be correct. 
and the test confirms it

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraph.java
andreachild: Moving the call to `unmarkDeleted` before the check for `vertexWithIdAlreadyExists` error would actually cause the error to be thrown because `unmarkDeleted` resets the `isDeletedInTx` value to `false` which causes `container.get()` to return non-null.

```
/**
 * Get transaction specific value of stored element.
 */
public T get() {
    if (isDeletedInTx.get()) return null;
    if (transactionUpdatedValue.get() != null) return transactionUpdatedValue.get();
    if (isDeleted) return null;
    return element;
}
```

tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraphTest.java
andreachild: Added check that original edge is dropped and remains dropped after adding the vertex back.

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraph.java
vkagamlyk: `vertexWithIdAlreadyExists` is in `else if` block, so can't be thrown when `unmarkDeleted` called

It's ok to leave it as you did if it works, it's too minor

xiazcy:
VOTE +1

tinkergraph-gremlin/src/main/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraph.java
andreachild: Ah yes I understand now

tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraphTest.java
andreachild: Will work on adding these new test cases.

andreachild: Added more comprehensive test cases including multi-threaded.

vkagamlyk:
thank you!

VOTE +1

PR #3163, opened 2025-07-24
Transaction reads should not interfere with add/update/delete

andreachild:
https://issues.apache.org/jira/browse/TINKERPOP-3142

Changed `TinkerElementContainer` to no longer consider element reads as being 'used in a transaction' which should be reserved for element add/update/delete only. This is appropriate for `TinkerTransactionGraph` because the isolation level is 'read committed'. With this isolation level, we do not need to protect against 'unrepeatable reads' or 'phantom reads'.

Prior to this change, a read-only thread could cause a 'Conflict: element modified in another transaction' error in a separate thread which was attempting add/drop/update.

### DIFF
--- a/tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraphTest.java
+++ b/tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraphTest.java
@@ -19,6 +19,7 @@
 package org.apache.tinkerpop.gremlin.tinkergraph.structure;
 
 import java.util.List;
+import java.util.concurrent.CountDownLatch;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
 import org.apache.tinkerpop.gremlin.structure.Edge;
 import org.apache.tinkerpop.gremlin.structure.T;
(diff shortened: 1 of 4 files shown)


andreachild:
@vkagamlyk would appreciate a review of this fix, thanks!

tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraphTest.java
```diff
@@ -1705,6 +1706,92 @@ public void shouldPreventDropUpdateVertexInConcurrentTransactions() {
...
+
+        // commit the initial vertex
+        gtx.addV().property(T.id, vertexId).next();
+        gtx.tx().commit();
+
+        // reader thread which is constantly reading the vertex ids in a separate transaction
+        CountDownLatch signal = new CountDownLatch(1);
+        Thread reader = new Thread(() -> {
+            while (signal.getCount() > 0) {
+                g.V().id().toList();
```
vkagamlyk: good to have assert that result is empty or contains 1 id vertexId

tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraphTest.java
```diff
@@ -1705,6 +1706,92 @@ public void shouldPreventDropUpdateVertexInConcurrentTransactions() {
...
+            Thread.sleep(50);
+            g.V(vertexId).drop().iterate();
+            g.tx().commit();
+            Thread.sleep(50);
+            g.addV().property(T.id, vertexId).next();
+            g.tx().commit();
+            g.V(vertexId).property(VertexProperty.Cardinality.single, "name", "test").next();
+            g.tx().commit();
+            g.V(vertexId).property(VertexProperty.Cardinality.single, "name", "updated").next();
+            g.tx().commit();
```
vkagamlyk, line 1751: could you add assert that vertex has correct property value?

vkagamlyk: same for second test

tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraphTest.java
```diff
@@ -1705,6 +1706,92 @@ public void shouldPreventDropUpdateVertexInConcurrentTransactions() {
...
+        final GraphTraversalSource gtx = g.tx().begin();
+        final int vertexId = 1;
+
+        // commit the initial vertex
+        gtx.addV().property(T.id, vertexId).next();
+        gtx.tx().commit();
+
+        // reader thread which is constantly reading the vertex ids in a separate transaction
+        CountDownLatch signal = new CountDownLatch(1);
+        Thread reader = new Thread(() -> {
```
vkagamlyk: could you add one more reading thread? to have 2 threads reading same modifying vertex.

vkagamlyk:
Thank you!

VOTE + 1 after adding assertions to tests

tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraphTest.java
andreachild: Added verification to check expected vertex ids (empty or single value of vertexId), same for the edge test.

tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraphTest.java
andreachild: Added property value check for vertex and edge test.

tinkergraph-gremlin/src/test/java/org/apache/tinkerpop/gremlin/tinkergraph/structure/TinkerTransactionGraphTest.java
andreachild: Increased reader threads to 2

xiazcy:
VOTE +1 pending CHANGELOG entry, and ideally some updates to reference doc to emphasize that non-repeatable reads and phantom reads are not protected at our isolation level.

andreachild:
> VOTE +1 pending CHANGELOG entry, and ideally some updates to reference doc to emphasize that non-repeatable reads and phantom reads are not protected at our isolation level.

Thanks, added changelog and updated semantics docs.

Cole-Greer:
VOTE +1. In my opinion the [documentation](https://tinkerpop.apache.org/docs/3.7.3/reference/#_transaction_semantics) has always been explicit regarding what isolation level we guarantee. I'm not opposed to adding an extra sentence or 2 stating that non-repeatable reads and phantom reads are not covered, but I don't believe it is necessary.

PR #3165, opened 2025-07-25
Upgraded node from version 18 to 20.

andreachild:
Upgraded node from version 18 to 20 since node 18 is past end of life.


Cole-Greer:
VOTE +1. Merging as CTR

PR #3170, opened 2025-07-30
TINKERPOP-3178 choose semantics consistency improvements

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-3178

Only match on first option for switch semantics. Introduced Pick.unproductive when the choice ends up being unproductive. Pass through unproductive choices for switch semantics so that they are like if-else. Introduce ChooseSemantics enum to inform on whether the step is in if-else or switch form. Lots of documentation improvements. Please check the documentation for more details on exactly what changed as that was fairly detailed. 

VOTE +1


docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -1145,6 +1146,77 @@ link:https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/java/o
...
+
+1. *if-then form*: `choose(predicate, trueChoice, falseChoice)` - If the predicate traversal or `P` returns a value
+(has a next element), the traverser is routed to the trueChoice traversal. Otherwise, it is routed to the falseChoice
+traversal. If the predicate is unproductive or if the falseChoice is not specified, then the traverser passes through.
+
+2. *switch form*: `choose(choice).option(pickValue, resultTraversal)` - The choice which may be a `Traversal` or
+`T` produces a value that is matched against the pickValue of each option. If a match is found, the traverser is routed
+to the corresponding resultTraversal and no further matches are attempted. If no match is found then the traverser is
+filtered, but, for these cases, the `Pick.none` option can be provided to match those cases. If the choiceTraversal is
+unproductive, then the traverser passes through by default or can be matched on `Pick.unproductive`.
```
Cole-Greer: There seems to be a bit of an inconsistency where unmatched traversers are filtered by default but unproductive cases are passed by default. Is there any motivation behind this?

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -1145,6 +1146,77 @@ link:https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/java/o
...
+* `Pick.unproductive` - Used when the choice traversal produces no value.
+
+`choose` does not allow more than one traversal to be assigned to a single `Pick`. The last `Pick` assigned via `option`
+is the one that will be used.
+
+The `choose()` step ensures that only one option is selected for each traverser, unlike other branch steps like
+`union()` that can route a traverser to multiple paths. As it is like `union()`, note that each `option` stream will
+behave like one:
+
+[source,text]
```
Cole-Greer: We don't yet have a precedent for traversal examples in the step semantics docs. Might be preferable to execute these examples via `[gremlin-groovy,modern]`

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -1145,6 +1146,77 @@ link:https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/java/o
...
+==>[marko,lop,josh,ripple,peter]
+gremlin> g.V().choose(__.has("name", "vadas"), __.values('age').fold(), __.values('name').fold())
+==>[27]
+==>[marko,lop,josh,ripple,peter]
+----
+
+*Exceptions*
+
+* `IllegalArgumentException` - If `Pick.any` is used as an option token, as only one option per traverser is allowed.
+* `IllegalArgumentException` - If the same pick token is used more than once.
```
Cole-Greer: This does not appear to be the case at the moment:
```
gremlin> g.V().choose(values("age"))
    .option(29, constant("match"))
    .option(none, constant("none"))
    .option(unproductive, constant("unproductive"))
    .option(29, constant("match2"))
    .option(none, constant("none2"))
    .option(unproductive, constant("unproductive2"))
==>match
==>none
==>unproductive
==>none
==>unproductive
==>none
```

docs/src/dev/provider/gremlin-semantics.asciidoc
spmallette: well, (1) it was always like that 🙂 but (2) i suppose i might say that if you were unproductive it means the `choose` wasn't even meant for that traverser, but if it was productive you had matches in mind and if those weren't satisfied then you'd expect it to fall through to `none` and would actively capture that. 

i'll admit that i thoroughly backed my way into (2) but doesn't sound too bad??? haha

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -1145,6 +1146,77 @@ link:https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/java/o
...
+to the corresponding resultTraversal and no further matches are attempted. If no match is found then the traverser is
+filtered, but, for these cases, the `Pick.none` option can be provided to match those cases. If the choiceTraversal is
+unproductive, then the traverser passes through by default or can be matched on `Pick.unproductive`.
+
+The `choose()` step can be used with the following special `Pick` tokens:
+
+* `Pick.none` - Used to specify a default option when no other options match.
+* `Pick.unproductive` - Used when the choice traversal produces no value.
+
+`choose` does not allow more than one traversal to be assigned to a single `Pick`. The last `Pick` assigned via `option`
```
Cole-Greer: Current implementation appears to favor the first match, not the last match.

```suggestion
`choose` does not allow more than one traversal to be assigned to a single `Pick`. The first `Pick` assigned via `option`
```

docs/src/dev/provider/gremlin-semantics.asciidoc
spmallette: that changed and i missed that update to the docs. removed that line

docs/src/dev/provider/gremlin-semantics.asciidoc
spmallette: i wasn't sure what made more sense and then moved on to other things. after some thought, i'd agree it should be an executed example.

docs/src/dev/provider/gremlin-semantics.asciidoc
spmallette: yep, good catch - that changed at some point and i must have missed that detail here in the docs.

docs/src/dev/provider/gremlin-semantics.asciidoc
Cole-Greer: My preference would be for both cases to filter by default for consistency, however I don't think this is a blocking concern.

Cole-Greer:
VOTE +1

docs/src/dev/provider/gremlin-semantics.asciidoc
spmallette: ended up agreeing with your call for consistency, but i'd rather not make `choose` seem like a filter step. i made it so that both `none` and `unproductive` cases pass through:

```
gremlin> g.V().choose(T.label).
......1>        option("person", __.out("knows").values("name")).
......2>        option("bleep", __.out("created").values("name"))
==>vadas
==>josh
==>v[3]
==>v[5]
gremlin> g.V().choose(T.label).
......1>        option("person", __.out("knows").values("name")).
......2>        option("bleep", __.out("created").values("name")).
......3>        option(none, none())
==>vadas
==>josh
```

i think i like the direct use of `none()` there to be explicit about what happens to those unmatched guys. You have to consciously think of that case now and the step hides nothing.

xiazcy:
VOTE +1

PR #3183, opened 2025-08-12
Remove P.getOriginalValue()

Cole-Greer:
`P.getOriginalValue()` is not providing much value and it often confused with `P.getValue()`. I would argue that all existing usages of originalValue in P and TextP are incorrect as they lead to very strange behaviour after `P.setValue()` is called:
```
gremlin> predicate = P.eq(5)
==>eq(5)
gremlin> predicate.equals(P.eq(5))
==>true
gremlin> predicate.setValue(10)
==>null
gremlin> predicate.equals(P.eq(5))
==>true
gremlin> predicate.equals(P.eq(10))
==>false
gremlin> predicate.test(5)
==>false
gremlin> predicate.test(10)
==>true
gremlin> predicate.toString()
==>eq(5)
```

The only other usage in our code is in MatchStepTest which is unaffected by switching to `P.getValue()`. My vote is to get rid of `originalValue` altogether as it is not offering any value and just leads to unnecessary confusion.

VOTE +1


xiazcy:
LGTM as long as there's no impact to the removal. Is there a way to test that the `P.setValue()` behaves correctly with the `originalValue` removed? 

VOTE +1

andreachild:
> LGTM as long as there's no impact to the removal. Is there a way to test that the `P.setValue()` behaves correctly with the `originalValue` removed?
> 
> VOTE +1

I've added unit tests to `PTest.java`.

andreachild:
VOTE +1

PR #3184, opened 2025-08-14
TINKERPOP-2491 Remove auto unfold of single item results for range, limit, tail local scope steps

andreachild:
[TINKERPOP-2491](https://issues.apache.org/jira/browse/TINKERPOP-2491)

Fixed inconsistent behaviour where `range()`, `limit()`, and `tail()` would return single elements instead of collections for single-item results with local scope:
- modified `RangeLocalStep.applyRangeIterable()` to always returns collections for iterable inputs regardless of number of results
- modified documentation and sample queries to reflect changed logic
- added more feature tests and unit tests

VOTE + 1


gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/Range.feature
```diff
@@ -191,7 +191,7 @@ Feature: Step - range()
       """
       g.V().fold().range(Scope.local, 6, 7)
       """
-    When iterated to list
+    When iterated next
```
xiazcy: Result of empty list can be represented as:
```
    When iterated to list
    Then the result should be unordered
      | result |
      | l[] |
```
I'd suggest either change this one or at least add another test for the empty list return.

xiazcy:
VOTE+1 pending comment addressed.

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -490,6 +490,52 @@ The `GremlinLangScriptEngine` has been modified to treat float literals without
 or 'd') as Double by default. Users who need `BigDecimal` precision can still use the 'm' suffix (e.g., 1.0m).
 `GremlinGroovyScriptEngine` will still default to `BigDecimal` for `float` literals.
 
+==== Consistent Collection Output for range(), limit(), and tail() Local Steps
```
spmallette: nit: needs a shorter section title or else it wraps terribly in the toc.

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -490,6 +490,52 @@ The `GremlinLangScriptEngine` has been modified to treat float literals without
...
+
+gremlin> g.inject([1, 2, 3]).limit(local, 2)
+==>[1,2]  // collection returned
+
+// Map behavior unchanged in both versions
+gremlin> g.inject([a: 1, b: 2, c: 3]).limit(local, 1)
+==>[a:1]  // Map entry returned (behavior unchanged)
+----
+
+If you need the old behavior of extracting single elements from collections, you can add `.unfold()` after the local step:
```
spmallette, line 551: This semantic change is a troublesome one because folks won't see it as failing when the upgrade - their queries will just return new results. I think that warrants a much stronger statement here, where there is a marching order to actively review existing Gremlin for this pattern to include `unfold`.

Cole-Greer:
VOTE +1 (pending resolution of above comments)

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/Range.feature
andreachild: Changed as suggested

docs/src/upgrade/release-3.8.x.asciidoc
andreachild: Shortened to 'Consistent Output for range(), limit(), tail()'

docs/src/upgrade/release-3.8.x.asciidoc
andreachild: Added warning text area to specifically call out migration of existing queries that use the affected steps otherwise a class cast exception could occur or unpredictable results.

PR #3185, opened 2025-08-14
GValueManager modifications

andreachild:
Modified `GValueManager` to track pinned gvalues by name only. Added deep cloning to clone method. Added javadoc. Fixed some code style warnings.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java
@@ -18,6 +18,8 @@
  */
 package org.apache.tinkerpop.gremlin.process.traversal;
 
+import java.util.Objects;
+import java.util.stream.Collectors;
 import org.apache.tinkerpop.gremlin.process.traversal.step.GValue;
 
 import java.io.Serializable;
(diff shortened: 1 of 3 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GValueManager.java
```diff
@@ -32,21 +34,25 @@
...
  */
-public final class GValueManager implements Serializable, Cloneable {
+public class GValueManager implements Serializable, Cloneable {
 
     private final Map<String, GValue<?>> gValueRegistry = new HashMap<>();
-    private final Map<String, GValue<?>> pinnedGValues = new HashMap<>();
+    /**
+     * Tracks pinned values for faster lookup.
+     */
+    private final Set<String> pinnedGValues = new HashSet<>();
```
Cole-Greer: Nit: I think something like `pinnedVariables` might be a better name if we're just storing variable names now.

andreachild: Changed to `pinnedVariables`

Cole-Greer:
Thanks for these fixes and improvements @andreachild. It all looks good to me.

VOTE +1

PR #3186, opened 2025-08-15
Deparameterize feature tests

Cole-Greer:
This PR inlines the majority of parameters in feature test scenarios for steps such as inject() and withSideEffect(). This is largely in service of the [larger work on GValue parameterization](https://github.com/apache/tinkerpop/pull/3157).

Parameters are largely unnecessary in the feature tests since they have been migrated from gremlin-groovy to gremlin-lang. A new bit of gherkin syntax is being added to directly set side effects without relying on query parameters which enables adding dynamic values such as vertex ids into side effects.

This change also exposed a bug in the Go translator, which has been resolved as well.

VOTE +1


gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/Where.feature
```diff
@@ -92,11 +92,10 @@ Feature: Step - where()
 
   Scenario: g_withSideEffectXa_josh_peterX_VX1X_outXcreatedX_inXcreatedX_name_whereXwithinXaXX
     Given the modern graph
-    And using the parameter xx1 defined as "l[josh,peter]"
     And using the parameter vid1 defined as "v[marko].id"
     And the traversal of
       """
-      g.withSideEffect("a", xx1).V(vid1).out("created").in("created").values("name").where(P.within("a"))
+      g.withSideEffect("a", ["josh","peter"]).V(vid1).out("created").in("created").values("name").where(P.within("a"))
```
xiazcy, line 98: Is the intent to replace all instances of `g.withSideEffect` with `using the side effect`? I see another one in `Math.feature` that still had it in the traversal definition.

xiazcy:
VOTE +1

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/Where.feature
Cole-Greer: There are no strict rules in cases like this one. Either inlining the side effect or using the new syntax is fine.
Passing a parameter to `withSideEffect()` will be disallowed after the GValue work is complete. The only time a side effect cannot be inlined is if it contains a dynamic value such as `v[marko].id` which needs to be processed through the test framework.

andreachild:
VOTE +1

PR #3187, opened 2025-08-15
TINKERPOP-3181 Improved host availability on initialization.

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-3181

If the `ConnectionPool` creates at least one good connection on initialization, allow the `Client` to have an available `Host` even if it's less than the `minPoolSize`. Allow normal recovery options to attempt reconnects and fall into unavailability the standard ways.

VOTE +1


vkagamlyk:
VOTE +1

Cole-Greer:
VOTE +1

PR #3193, opened 2025-08-21
Change tail feature test to use valueMap instead of inject as to not be affected by GLV map ordering

andreachild:
https://issues.apache.org/jira/browse/TINKERPOP-2491

Change tail feature test to use `valueMap()` instead of `inject(map)` as to not be affected by go's lack of ordered map data structure. Added a note to go limitations about the lack of native support for ordered map.

VOTE +1


Cole-Greer:
VOTE +1. It's ok to CTR small test fixes like this instead of opening a PR if you'd like.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/Tail.feature
```diff
@@ -239,11 +239,12 @@ Feature: Step - tail()
       | l[d[3].i] |
 
   # Test that Map local tail 1 - should still return single entry
+  # note that this test injects a map with a single entry as to not enforce an expected ordering
   Scenario: g_injectXa1_b2_c3X_tailXlocal_1X
     Given the empty graph
     And the traversal of
       """
-      g.inject(["a": 1, "b": 2, "c": 3]).tail(Scope.local, 1)
+      g.inject(["c": 3]).tail(Scope.local, 1)
```
spmallette: don't we use LinkedHashMap internally by default? 

or i guess you could `unfold().order().fold()` to force it to actually test something, no?

andreachild: I believe the ordering issue is not in the step but in the GLV feature test step definition and result parsing. I didn't get a chance to troubleshoot it thoroughly.

andreachild: > unfold().order().fold()

I can try this

spmallette: ok, well, then i think you should CTR this through so that stuff is building with stability. 

but seems like there should be some investigation into the test suite then (in part because this test really no longer tests `tail` well). is it possible the Go serializer isn't preserving the concept of a LinkedHashMap in serialization? that seems like a reasonable place where this problem would poke its head because the result parsing is getting back `{a:1}` and we're only expecting on result so the Map must not be getting there right.

andreachild: I've changed the test to use `valueMap()` instead which is a more realistic test than using `inject` anyways.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/Tail.feature
```diff
@@ -239,16 +239,16 @@ Feature: Step - tail()
...
 
   # Test that Map local tail 1 - should still return single entry
-  Scenario: g_injectXa1_b2_c3X_tailXlocal_1X
-    Given the empty graph
+  Scenario: g_VX1X_valueMapXnameX_tailXlocal_1X
+    Given the modern graph
     And the traversal of
       """
-      g.inject(["a": 1, "b": 2, "c": 3]).tail(Scope.local, 1)
+      g.V(1).valueMap('name').tail(Scope.local, 1)
```
Cole-Greer: This should use `And using the parameter vid1 defined as "v[marko].id"` and `g.V(vid1)` instead. Direct use of V(1) will break for providers who do not use int ids.

andreachild: Changed to use param

xiazcy:
VOTE +1

PR #3194, opened 2025-08-29
TINKERPOP-3182 Fixed inconsistency in Path GraphSON serialization

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-3182

`Path` objects were being forcibly detached by reference in GraphSON v1/v2 which put it at odds with v1 untyped. Inclusion or removal of properties should be controlled via `materializeProperties` only.

There's still a bit of weirdness here for Javascript particularly if you note the assertions in our tests. That should be resolved separately with a breaking change: [TINKERPOP-3186](https://issues.apache.org/jira/browse/TINKERPOP-3186).

VOTE +1

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.js
@@ -79,10 +79,12 @@ class Edge extends Element {
     this.inV = inV;
     this.properties = {};
     if (properties) {
-      const keys = Object.keys(properties);
-      for (let i = 0; i < keys.length; i++) {
-        const k = keys[i];
-        this.properties[k] = properties[k].value;
+      if (Array.isArray(properties)) {
+        // Handle array of Property objects
+        properties.forEach((prop) => (this.properties[prop.key] = prop.value));
+      } else {
+        // Handle object format as before
+        Object.keys(properties).forEach((k) => (this.properties[k] = properties[k].value));
       }
     }
   }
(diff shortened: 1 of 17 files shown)


gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.js
```diff
@@ -79,10 +79,19 @@ class Edge extends Element {
     this.inV = inV;
     this.properties = {};
     if (properties) {
-      const keys = Object.keys(properties);
-      for (let i = 0; i < keys.length; i++) {
-        const k = keys[i];
-        this.properties[k] = properties[k].value;
+      if (Array.isArray(properties)) {
```
andreachild, line 82: Should this same logic apply for `Vertex`?

spmallette: no. doing this here in `Edge` preserves the map structure for `properties` that was there before irrespective of what the serializer decides to send in the constructor. `Vertex` and `VertexProperty` just had a direct assignment which just accepted what the serializer gave it (with all the weirdness that comes with that), so i left that.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/structure/graph.js
```diff
@@ -79,10 +79,19 @@ class Edge extends Element {
...
+          const prop = properties[i];
+          this.properties[prop.key] = prop.value;
+        }
+      } else {
+        // Handle object format as before
+        const keys = Object.keys(properties);
+        for (let i = 0; i < keys.length; i++) {
+          const k = keys[i];
+          this.properties[k] = properties[k].value;
+        }
```
andreachild: Nit: can simplify
```suggestion
      if (Array.isArray(properties)) {
        // Handle array of Property objects
        properties.forEach((prop) => (this.properties[prop.key] = prop.value));
      } else {
        // Handle object format as before
        Object.keys(properties).forEach((k) => (this.properties[k] = properties[k].value));
```

andreachild:
VOTE +1

xiazcy:
VOTE +1

Cole-Greer:
VOTE +1

PR #3195, opened 2025-08-29
[TINKERPOP-3173] Simplify comparability semantics

Cole-Greer:
https://issues.apache.org/jira/browse/TINKERPOP-3173

Replaces the old system of ternary boolean semantics with simple binary semantics. The triggers for "ERROR" states from illegal comparisons are unchanged (typically comparisons with NaN or between incomparable types such as String and int). The difference now is that instead of the ERROR being propagated according to ternary logic semantics until a reduction point is reached, the error now immediately returns a value of FALSE.

This will be most visible in expressions which include negations. Prior to this change, `g.inject(NaN).not(is(1))` would produce no results as !(NaN == 1) -> !(ERROR) -> ERROR -> traverser is filtered out. After this change, the same traversal will return NaN as the same expression now evaluates as !(NaN == 1) -> !(FALSE) -> TRUE -> traverser is not filtered.

This commit also introduces NotP to model a negated P, as the old method of getting the complement of the PBiPredicate may handle edge cases incorrectly:

!(1 < NaN) != (1 >= NaN) -> !(FALSE) != (FALSE)


vkagamlyk:
is this change for 3.8 or 4.0?

Cole-Greer:
> is this change for 3.8 or 4.0?

3.8-dev 🤦

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/language/grammar/TraversalPredicateVisitorTest.java
```diff
@@ -42,26 +42,26 @@ public class TraversalPredicateVisitorTest {
...
-                {"without([1, 2])", P.without(Arrays.asList(1, 2))},
-                {"without([1, 2])", P.without(1, 2)},
-                {"without(1, 2)", P.without(1, 2)},
-                {"without(1)", P.without(1)},
-                {"without([1])", P.without(1)},
-                {"without()", P.without()},
-                {"without(1, [1, 2])", P.without(1, Arrays.asList(1, 2))},
-                {"within([1, 2, 3], 1, [1, 2])", P.within(Arrays.asList(1, 2, 3), 1, Arrays.asList(1, 2))},
-                {"not(without(1, 2))", P.not(P.without(1, 2))},
+//                {"eq(10)", P.eq(10)},
```
xiazcy: Note to uncomment these tests given they should be passing.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/GremlinValueComparator.java
```diff
@@ -331,17 +307,74 @@ private static boolean naturallyComparable(final Object f, final Object s) {
...
+            return contentsComparable(l1.iterator(), l2.iterator());
+        }
+        else if (ft == Type.MapEntry && st == Type.MapEntry) {
+            return comparable(((Map.Entry) f).getKey(), ((Map.Entry) s).getKey()) &&
+                    comparable(((Map.Entry) f).getValue(), ((Map.Entry) s).getValue());
+        }
+        else if (ft == Type.MapEntry && st == Type.MapEntry) {
+            return comparable(((Map.Entry) f).getKey(), ((Map.Entry) s).getKey()) &&
+                    comparable(((Map.Entry) f).getValue(), ((Map.Entry) s).getValue());
+        }
```
xiazcy:
```suggestion
```
Looks like a duplicated case of `MapEntry`. Or did you mean to compare a different type?

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/PTest.java
```diff
@@ -22,6 +22,7 @@
 import org.apache.tinkerpop.gremlin.process.traversal.step.GValue;
 import org.apache.tinkerpop.gremlin.process.traversal.util.AndP;
 import org.apache.tinkerpop.gremlin.process.traversal.util.OrP;
+import org.apache.tinkerpop.gremlin.util.GremlinValueComparator;
```
xiazcy: nit: unused

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/PTest.java
```diff
@@ -395,6 +397,7 @@ public static class SetValueTest {
         private static final String NEQ_FORMAT = "neq(%d)";
```
xiazcy: nit: `NEQ_FORMAT` is now unused.

xiazcy:
VOTE +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/NotP.java
```diff
@@ -0,0 +1,110 @@
...
+    public P<V> negate() {
+        return originalP;
+    }
+
+    public P<V> clone() {
+        return new NotP<>(this.originalP.clone());
+    }
+
+    public static class NotPBiPredicate<T, U> implements PBiPredicate<T, U>, Serializable {
+        PBiPredicate<T, U> original;
```
kenhuuu, line 77: Nit: final

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/NotP.java
```diff
@@ -0,0 +1,110 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.tinkerpop.gremlin.process.traversal;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+public class NotP<V> extends P<V> {
+    private P<V> originalP;
```
kenhuuu, line 29: Something feels a bit off about this class. It both wraps P and extends P but it doesn't feel like either is completely needed in this case. It would probably be easier to understand if some changes were made to P itself. Other things that are confusing is that it leads to situations where negate() returns originalP which isn't immediately clear why that would be the case. If this isn't something that you want to address for this PR then just make a short javadoc for this class that has a quick explanation about its design and make a Jira for future refactoring.

kenhuuu:
Feels like tests with nested operations like NOT (A OR B) might be missing. Essentially tests that show why the ternary logic didn't quite work in the first place.

Cole-Greer:
> Feels like tests with nested operations like NOT (A OR B) might be missing. Essentially tests that show why the ternary logic didn't quite work in the first place.

The updated tests in the renamed ComparabilitySemanticsTests provided all of the coverage I wanted to see for this change. I have not looked at special test cases for AndStep, OrStep, P.and, or P,or as none of these have any special handling regarding the new comparability semantics, and are not able to demonstrate the changes.

There were a few existing steps using the NotStep which demonstrate the new semantics, and have been updated accordingly. It's occurred to me however that we don't have a dedicated `Not.feature` file at the moment. I'm not sure why that is, but I can add such a file with basic scenarios as part of this PR.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/NotP.java
Cole-Greer: Both wrapping and extending P is a bit weird. This was primarily modeled after ConnectiveP which has a similar structure, although perhaps less odd there given it wraps a collection of P's. P should arguably be refactored as an interface, with ConnectiveP (AndP and OrP), NotP, TextP, and DefaultP as implementations. I will open up a JIRA for such a change.

kenhuuu:
VOTE +1

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -144,67 +144,10 @@ Key differences include handling of numeric types and NaN.
...
-|===
-|Argument | Result
-
-|`TRUE` | `FALSE`
-|`FALSE` | `TRUE`
-|`ERROR` | `ERROR`
-|===
+Comparability semantics are not complete with respect to traditional binary boolean semantics, as certain comparisons
+cannot be proved as either `TRUE` or `FALSE`. Common examples of such cases are Comparability against `NaN`, cross-type
+Comparability (e.g. `String` vs `Numeric`). For the purposes of Comparability within Gremlin, any such incomputable
```
vkagamlyk, line 149: Probably it's worth adding a section on special cases when typecasting (not number type promotion) is performed for comparison, like for element id's 
`g.V("1")` vs `g.V().id().is("1")`

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/types/PSerializer.java
```diff
@@ -153,7 +154,14 @@ protected void writeValue(final T value, final Buffer buffer, final GraphBinaryW
         // the predicate name is either a static method of P or an instance method when a type ConnectiveP
         final boolean isConnectedP = value instanceof ConnectiveP;
         final String predicateName = value.getPredicateName();
-        final Object args = isConnectedP ? ((ConnectiveP<?>) value).getPredicates() : value.getValue();
+        final Object args;
+        if (value instanceof ConnectiveP) {
```
vkagamlyk, line 157: nit: variable `isConnectedP` can be used here or removed

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/GremlinValueComparator.java
```diff
@@ -331,17 +307,70 @@ private static boolean naturallyComparable(final Object f, final Object s) {
...
             return f == s; // true iff both in the null space
 
+        if (eitherAreNaN(f, s))
+            return false;
+
         final Type ft = Type.type(f);
         final Type st = Type.type(s);
 
+        // if objects are collections or composites, their contents must be mutually comparable
+        if (ft == Type.List && st == Type.List) {
```
vkagamlyk, line 326: is same for arrays?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/NotP.java
Cole-Greer: New JIRA for this: https://issues.apache.org/jira/browse/TINKERPOP-3190

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/types/PSerializer.java
Cole-Greer: I've dropped the variable

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/util/GremlinValueComparator.java
Cole-Greer: Great catch. Under the original TBL semantics arrays and Lists were mutually comparable. That was missing here. I've updated Type.List to now encompass arrays, and updated the associated comparator to handle iterators from either List or Arrays.

docs/src/dev/provider/gremlin-semantics.asciidoc
Cole-Greer: I'd prefer to leave this out of this section of the docs. The parsing of Strings is really a special implementation detail of the Integer IdManager in TinkerGraph. As far as gremlin itself is concerned, `g.V("1")` is looking for a vertex with a String id `"1"`. If this needs documentation I would rather it be listed as some special sugar in TinkerGraph, rather than an exception outlined here in the comparability semantics docs.

vkagamlyk:
VOTE +1

PR #3197, opened 2025-09-03
Upgrade netty to 4.1.125.Final (GHSA-3p8m-j85q-pgmj)

smholsen:
Please see https://github.com/netty/netty/security/advisories/GHSA-3p8m-j85q-pgmj

PR #3203, opened 2025-09-09
Address Test Flakiness (overLimitOperationsShouldDelegateToSingleNewConnection)

smholsen:
This PR is spun up as a follow up from https://github.com/apache/tinkerpop/pull/3197#issuecomment-3268246328. 

It updates the overLimitOperationsShouldDelegateToSingleNewConnection test and adds a descrptive comment as to why we do not check the number of connections directly. 

The test flakiness only occurs on the GH runners (unable to reproduce locally).

### DIFF
--- a/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/ClientConnectionIntegrateTest.java
+++ b/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/ClientConnectionIntegrateTest.java
@@ -227,10 +227,17 @@ public class ClientConnectionIntegrateTest extends AbstractGremlinServerIntegrat
                 }
             });
 
-            assertEquals(2, connectionBorrowCount.size());
-            for (int finalBorrowCount : connectionBorrowCount.values()) {
-                assertEquals(usagePerConnection, finalBorrowCount);
-            }
+        // The intent of this test is to ensure that the driver grows from 1 connection to only as many as needed to satisfy maxSimultaneousUsagePerConnection limits
+        // (i.e. 2 fully utilized connections at 3 operations each for a total of 6 operations).
+        // On CI / Github runners, ephemeral third connections can appear.
+        // We can therefore instead of checking the connectionBorrowCount size directly, we check only the number of actually fully utilized connections (as per usagePerConnection),
+        // and the exact number of operations accounted for (allowing the existence of a third connection).
+
+        final long fullyUtilized = connectionBorrowCount.values().stream().filter(c -> c == usagePerConnection).count();
+        final long totalBorrows = connectionBorrowCount.values().stream().mapToInt(Integer::intValue).sum();
+
+        assertEquals("Expected total operations accounted for", operations, totalBorrows);
+        assertEquals("Expected exactly two fully utilized connections", 2, fullyUtilized);
 
         } finally {
             executorServiceForTesting.shutdown();


gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/ClientConnectionIntegrateTest.java
```diff
@@ -227,10 +227,17 @@ public void overLimitOperationsShouldDelegateToSingleNewConnection() throws Inte
                 }
             });
 
-            assertEquals(2, connectionBorrowCount.size());
-            for (int finalBorrowCount : connectionBorrowCount.values()) {
-                assertEquals(usagePerConnection, finalBorrowCount);
-            }
+        // The intent of this test is to ensure that the driver grows from 1 connection to only as many as needed to satisfy maxSimultaneousUsagePerConnection limits
+        // (i.e. 2 fully utilized connections at 3 operations each for a total of 6 operations).
+        // On CI / Github runners, ephemeral third connections can appear.
```
andreachild, line 232: Would be nice to understand what's causing the ephemeral third connection.

In the past I was able to reproduce flaky github connection test failures that I could not reproduce locally but only on a slower EC2 instance. I should still have that particular EC2 available and can try to reproduce the test failure.

smholsen: Makes sense!
First time looking into this repo at all, so apologies if my investigation is a bit lacking here. 

Looking a bit further here I have an assumption which would be great if you could validate:

We have `.maxConnectionPoolSize(operations)` (which I think is not correct for this test? I assume this should actually be 2, since we have 6 operations, and 3 `usagePerConnection`. ) 

When `invokeAll` is called, it calls `chooseConnection` on all 6. In turn, this will cal `pool.borrowConnection(...)`, which again calls sets its connection with `final Connection leastUsedConn = getLeastUsedValidConnection();`. 

`getLeastUsedValidConnection` actually has this comment:
```
// Increment borrow count and consider making a new connection if least used connection hits usage maximum
```

So for the last one of this, we will have 2 full connections (each filled with 3 operations). Since we have a `maxPoolSize` of > 2, it will create the third. 

So perhaps we can just leave the test as is, and set `.maxConnectionPoolSize(2)`?

andreachild: > So perhaps we can just leave the test as is, and set .maxConnectionPoolSize(2)?

You might have something there, I will take a closer look at the test when I can. So far I haven't been able to reproduce the test flakiness using your branch on the EC2 I had created specifically to mimic the github runner.

smholsen: I updated the other PR with this suggestion, would be interesting to see if that resolves the issue on the workflows for that PR. 
(https://github.com/apache/tinkerpop/pull/3197#issuecomment-3293741555)

andreachild: The change you added seemed to have worked for the other PR, do you want to close this one?

smholsen: Sounds good! Thanks for following up!

PR #3204, opened 2025-09-09
Fix WebSocket API inconsistency in Gremlin connection

xhuberty:
Fix WebSocket API inconsistency causing "this._ws.on is not a function" error

  The Gremlin JavaScript driver was mixing browser WebSocket API (.addEventListener)
  with Node.js WebSocket API (.on) in the same connection logic. This caused runtime
  errors when globalThis.WebSocket was detected but the browser WebSocket implementation
  doesn't support the .on() method used for 'unexpected-response' events.

  Changes:
  - Conditionally attach 'unexpected-response' listener only for Node.js WebSocket (ws package)
  - Preserve existing useGlobalWebSocket logic for legitimate browser/polyfill scenarios
  - Update cleanup method to safely remove listeners based on WebSocket type

  This maintains backward compatibility while fixing the cross-platform API incompatibility
  that prevented connections in environments with WebSocket polyfills or global WebSocket objects.

This is related to https://issues.apache.org/jira/browse/TINKERPOP-3160

### DIFF
--- a/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
+++ b/gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
@@ -163,10 +163,11 @@ class Connection extends EventEmitter {
 
     this._ws.addEventListener('open', this.#handleOpen);
     this._ws.addEventListener('error', this.#handleError);
-    this._ws.on('unexpected-response', this.#handleUnexpectedResponse);
+    if (!useGlobalWebSocket) {
+      this._ws.on('unexpected-response', this.#handleUnexpectedResponse);
+    }
     this._ws.addEventListener('message', this.#handleMessage);
     this._ws.addEventListener('close', this.#handleClose);
-
     return await this._openPromise;
   }
 
@@ -415,7 +416,11 @@ class Connection extends EventEmitter {
     });
     this._ws.removeEventListener('open', this.#handleOpen);
     this._ws.removeEventListener('error', this.#handleError);
-    this._ws.off('unexpected-response', this.#handleUnexpectedResponse);
+    // Only remove unexpected-response listener for Node.js WebSocket (ws package)
+    // Browser WebSocket does not have this event and .off() method
+    if (this._ws && 'off' in this._ws) {
+      this._ws.off('unexpected-response', this.#handleUnexpectedResponse);
+    }
     this._ws.removeEventListener('message', this.#handleMessage);
     this._ws.removeEventListener('close', this.#handleClose);
     this._openPromise = null;


xhuberty:
The test that fails is due to existing code that I have kept. Do you want to keep it or should I remove it?

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.ts
```diff
@@ -465,10 +469,14 @@ export default class Connection extends EventEmitter {
     this._ws?.removeEventListener('open', this.#handleOpen);
     // @ts-expect-error
     this._ws?.removeEventListener('error', this.#handleError);
-    // The following listener needs to use `.on` and `.off` because the `ws` package's addEventListener only accepts certain event types
-    // Ref: https://github.com/websockets/ws/blob/8.16.0/lib/event-target.js#L202-L241
-    // @ts-expect-error
-    this._ws?.off('unexpected-response', this.#handleUnexpectedResponse);
+    // Only remove unexpected-response listener for Node.js WebSocket (ws package)
+    // Browser WebSocket does not have this event and .off() method
+    if (this._ws && 'off' in this._ws) {
```
andreachild, line 421: I am wondering if the same check as above `!useGlobalWebSocket` would also work here or is there a particular reason to use different if conditions for `.on` vs `.off`?

andreachild:
> The test that fails is due to existing code that I have kept. Do you want to keep it or should I remove it?

Which test is failing exactly? I can't tell from the logs. It seems like the change is backwards compatible so I am wondering what is failing.

xhuberty:
> > The test that fails is due to existing code that I have kept. Do you want to keep it or should I remove it?
> 
> Which test is failing exactly? I can't tell from the logs. It seems like the change is backwards compatible so I am wondering what is failing.

lib/driver/connection.ts(477,7): error TS2578: Unused '@ts-expect-error' directive.

andreachild:
Hi @xhuberty can you please change the PR to rebase and target 3.7-dev branch?

andreachild:
> > > The test that fails is due to existing code that I have kept. Do you want to keep it or should I remove it?
> > 
> > 
> > Which test is failing exactly? I can't tell from the logs. It seems like the change is backwards compatible so I am wondering what is failing.
> 
> lib/driver/connection.ts(477,7): error TS2578: Unused '@ts-expect-error' directive.

Yes that should be removed as @ts-expect-error is unnecessary after your changes

xhuberty:
I will update the PR next week.

andreachild:
> I will update the PR next week.

Hi @xhuberty there is another open PR which has incorporated your changes with modifications https://github.com/apache/tinkerpop/pull/3213. I'd like to be able to merge yours first so that you are acknowledged as the commit author of the fix. Do you know when you'll be able to get back to this one?

kenhuuu:
Hi @xhuberty 

Any update on this?

As mentioned before, this PR needs to target 3.7-dev and not master. Will you have some time in the near future to address the remaining issues? As mentioned in the [devlist](https://lists.apache.org/thread/1obm5zkl7o2rpzxwr74zjd9xmr45rqzk), there is a proposed release coming soon. If there is no imminent activity on this PR, then we will have to merge the other PR (#3213) that fixes this same issue, in order to have it included in the next release. In that case, if possible, we will attempt to cherry-pick your commit into master during the merge so that you will get credit for your contribution.

xhuberty:
I will do it this morning, wanted to do it now but I'm in a train the connection is to laggy.

xhuberty:
Done

kenhuuu:
Thanks for your quick response and update to this.

VOTE +1.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
```diff
@@ -415,7 +417,9 @@ class Connection extends EventEmitter {
     });
     this._ws.removeEventListener('open', this.#handleOpen);
     this._ws.removeEventListener('error', this.#handleError);
-    this._ws.off('unexpected-response', this.#handleUnexpectedResponse);
+    if (!useGlobalWebSocket) {
```
andreachild: This fails as `useGlobalWebSocket` is not defined here so you'll have to define it or go back to your original fix
```suggestion
    // Only remove unexpected-response listener for Node.js WebSocket (ws package)
    // Browser WebSocket does not have this event and .off() method
    if (this._ws && 'off' in this._ws) {
```

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
```diff
@@ -163,10 +163,12 @@ class Connection extends EventEmitter {
...
     this._ws.addEventListener('open', this.#handleOpen);
     this._ws.addEventListener('error', this.#handleError);
-    this._ws.on('unexpected-response', this.#handleUnexpectedResponse);
+    if (!useGlobalWebSocket) {
+      this._ws.on('unexpected-response', this.#handleUnexpectedResponse);
+    }
     this._ws.addEventListener('message', this.#handleMessage);
     this._ws.addEventListener('close', this.#handleClose);
-
+
```
andreachild: The linter doesn't like these spaces
```suggestion
```

andreachild:
VOTE +1

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
xhuberty: Done

xiazcy:
VOTE +1, thanks for the change!

PR #3205, opened 2025-09-10
Refinements to GValue

Cole-Greer:
This is a followup to https://github.com/apache/tinkerpop/pull/3157 with some fixes and refinements now that these changes have had more time to marinate.

Changes are as follows:

- Refactors
  - Rename PropertyAdding interface to PropertiesHolder
  - Introduce AbstractAddEdgeStepPlaceholder, AbstractAddVertexStepPlaceholder, and AbstractMergeElementStepPlaceholder to deduplicate some of the repetitive code in some of the placeholders
- Strategy Fixes
  - Fix strategy tests not correctly verifying that GValues are unpinned which previously led to false positives
  - Fix excessive variable pinning in PartitionStrategy, SubgraphStrategy, IncidentToAdjacentStrategy, and AdjacentToIncidentStrategy
- StepContract/Placeholder Refinement
  - Get rid of placeholder exclusive methods as these add unnecessary branching and casts to strategies. All placeholder methods now must override from the contract
  - Replace most "GValueSafe" getters with "AsGValue" or "WithGValue" getter that directly produces GValues without pinning variables. This is as the main use case for accessing step arguments without pinning variables is to blindly copy all arguments to a new step, in which case its preferred to retain the GValue structure. It's arguably invalid to copy a value on its own to a new step without pinning the variable or retaining the GValue boxing as it would break future calls to `updateVariable()`
  
Here is a complete listing of the changes to the Placeholder/Contract interfaces:

- CallStep
  - Replace getMergedParamsGValueSafe() with getStaticParamsAsGValue, note switch from mergedParams to static only.
  - Stuck with public Service<S, E> serviceGValueSafe()
- GraphStep
  - Replace getIdsGValueSafe with getIdsAsGValues
- VertexStep
  - Replace getEdgeLabelsGValueSafe with getEdgeLabelsAsGValues
- MergeV/EStep
  - Replace getMergeTraversalGValueSafe with getMergeMapWithGValue
  - Replace getOnCreateTraversalGValueSafe with getOnCreateMapWithGValue
  - Replace getOnMatchTraversalGValueSafe with getOnMatchMapWithGValue
  - Return types are complicated, see javadocs in MergeStepContract
- IsStep
  - Retained getPredicateGValueSafe()
- AddElement
  - Replace getLabelGValueSafe with getLabelAsGValue
  - Replace getElementIdGValueSafe with getElementIdAsGValue
  - Replace getPropertiesGValueSafe with getPropertiesWithGValues
- AddVertex
  - Covered by AddElement changes.
- AddEdge
  - Replace getFrom/ToGValueSafe with getFrom/ToWithGValue
  - Return types are complicated, see javadocs in AddEdgeStepContract
- AddProperty
  - Replace getValueGValueSafe with getValueAsGValue
- RangeGlobal/Local
  - Replace getLowRangeGValueSafe with getLowRangeAsGValue
  - Replace getHighRangeGValueSafe with getHighRangeAsGValue
  - Remove getLowName and getHighName
- TailGlobal/Local
  - Replace getLimitGValueSafe with getLimitAsGValue
- HasContainerHolder
  - Retains getPredicatesGValueSafe()

Lingering thoughts: Don’t like interface on methods which might return GValues or Traversals (getFrom()/getTo() in AddE, getMergeMap… in mergeV/E), these cases need more consideration.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/PropertiesHolder.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/PropertiesHolder.java
@@ -16,12 +16,12 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.tinkerpop.gremlin.process.traversal.step; //TODO:: relocate class
+package org.apache.tinkerpop.gremlin.process.traversal.step;
 
 import java.util.List;
 import java.util.Map;
 
-public interface PropertyAdding {
+public interface PropertiesHolder {
     void addProperty(Object key, Object value);
     Map<Object, List<Object>> getProperties();
     boolean removeProperty(Object k);
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
@@ -62,7 +62,7 @@ import org.apache.tinkerpop.gremlin.process.traversal.step.map.MergeEdgeStepPlac
 import org.apache.tinkerpop.gremlin.process.traversal.step.map.MergeStep;
 import org.apache.tinkerpop.gremlin.process.traversal.step.map.MergeStepContract;
 import org.apache.tinkerpop.gremlin.process.traversal.step.map.MergeVertexStepPlaceholder;
-import org.apache.tinkerpop.gremlin.process.traversal.step.PropertyAdding;
+import org.apache.tinkerpop.gremlin.process.traversal.step.PropertiesHolder;
 import org.apache.tinkerpop.gremlin.process.traversal.step.map.VertexStepPlaceholder;
 import org.apache.tinkerpop.gremlin.process.traversal.step.sideEffect.AddPropertyStepPlaceholder;
 import org.apache.tinkerpop.gremlin.process.traversal.step.map.AddEdgeStepContract;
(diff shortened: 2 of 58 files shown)


Cole-Greer:
I still want to followup further on interfaces with methods which ambiguously return GValue or Traversal (getFrom()/getTo() in AddE, getMergeMap… in mergeV/E), however there is already a lot in this PR so I'd rather spin that off in a followup PR.

VOTE +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/PropertiesHolder.java
```diff
@@ -16,13 +16,17 @@
...
 import java.util.List;
 import java.util.Map;
 
-public interface PropertyAdding {
+public interface PropertiesHolder {
     void addProperty(Object key, Object value);
     Map<Object, List<Object>> getProperties();
     boolean removeProperty(Object k);
+
+    default public Map<Object, List<Object>> getPropertiesWithGValues() {
```
andreachild, line 29: Can you clarify the usage of this method? From what I can see the overrides of this method basically return the same properties but as immutable collection?

Cole-Greer: For a simple query such as `g.addV().property("name", "marko").property("age", GValue.of("a", 29))`, the regular `getProperties()` method will resolve any GValues and pin those variables in the GValueManger, where `getPropertiesWithGValues()` will preserve GValues.

```
addVertexStep.getProperties()
==> {
            "name": "marko",
            "age": 29 // GValue is reduced to literal and variable "a" is pinned
        }

addVertexStep.getPropertiesWithGValues()
==> {
            "name": "marko",
            "age": GValue.of("a", 29) // GValue is retained, variable is not pinned.
        }
```

The default implementation is only overloaded in places where the plain `getProperties()` does too much (reduce GValues and pin variables).

 The main intended use case for this method is for strategies which wish to create a new step placeholder which copies all properties from an existing step without losing GValues.

andreachild: Got it, thanks

PR #3208, opened 2025-09-17
Added Air Routes to the distribution as a sample dataset

spmallette:
This is mostly there for end use at this point as it's not tied into the test framework yet. That required a bit more work because air-routes doesn't use a "name" property unfortunately so we would have to special case assertions around "code" i guess. That issue can be saved for another day in my mind. 

VOTE +1


Cole-Greer:
This is definitely a candidate for PR of the year. Air routes is a great dataset for examples, learning and testing (once we get that wired up). VOTE +1

xiazcy:
Great dataset addition. VOTE +1

PR #3209, opened 2025-09-17
Fix TraversalParent implementations for Placeholder steps

Cole-Greer:
The TraversalParent methods were not consistently implemented in the new placeholder steps. This commit adds proper implementations in all TraversalParent placeholder steps, and introduces TraversalParentTest.

TraversalParentTest takes traversals with TraversalParents and verifies that getGlobalChildren() and getLocalChildren() are producing the expected children, and that those children have correctly set their parents. These checks are done before and after strategy application. This is primarily intended for placeholder steps to ensure that TraversalParent remains correctly implemented in both placeholder and concrete steps, however it is also useful to verify against interference from strategies.

The new tests identified an existing deficiency DateDiffStep which was also addressed.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/lambda/GValueConstantTraversal.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/lambda/GValueConstantTraversal.java
@@ -57,6 +57,17 @@ public final class GValueConstantTraversal<S, E> extends AbstractLambdaTraversal
         return constantTraversal.equals(other);
     }
 
+    @Override
+    public GValueConstantTraversal<S, E> clone() {
+        GValueConstantTraversal<S, E> clone = new GValueConstantTraversal<>(GValue.of(this.end.getName(), this.end.get()));
+        try {
+            clone.setGValueManager(this.getGValueManager().clone());
+        } catch (CloneNotSupportedException e) { //TODO:: handle properly
+            throw new RuntimeException(e);
+        }
+        return clone;
+    }
+
     public ConstantTraversal<S, E> getConstantTraversal() {
         return constantTraversal;
     }
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/TraversalParent.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/TraversalParent.java
@@ -21,6 +21,8 @@ package org.apache.tinkerpop.gremlin.process.traversal.step;
 import org.apache.tinkerpop.gremlin.process.traversal.GValueManager;
 import org.apache.tinkerpop.gremlin.process.traversal.Step;
 import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
+import org.apache.tinkerpop.gremlin.process.traversal.step.branch.BranchStep;
+import org.apache.tinkerpop.gremlin.process.traversal.step.branch.LocalStep;
 import org.apache.tinkerpop.gremlin.process.traversal.traverser.TraverserRequirement;
 import org.apache.tinkerpop.gremlin.process.traversal.util.TraversalHelper;
 
@@ -31,14 +33,30 @@ import java.util.List;
 import java.util.Set;
 
 /**
+ * Indicates that a step can contain child Traversals. Any step which implements this interface should override at least
+ * one of getGlobalChildren() or getLocalChildren(). All TraversalParents should call integrateChild() for all child
+ * traversals as they are added to the step, and TraversalParents should override {@link Step::setTraversal()}, and
+ * call integrateChild() again for all child traversals.
+ *
  * @author Marko A. Rodriguez (http://markorodriguez.com)
  */
 public interface TraversalParent extends PopContaining, AutoCloseable {
 
+    /**
+     * Gets a list of all "global" child traversals for this step. A "global" traversal is one which evaluates across
+     * all of its parents incoming traversers at once. This is typically used in cases where the child traversal
+     * represents a branch of traversal flow. See {@link BranchStep} as an example.
+     */
     public default <S, E> List<Traversal.Admin<S, E>> getGlobalChildren() {
         return Collections.emptyList();
     }
 
+    /**
+     * Gets a list of all "local" child traversals for this step. A "local" traversal is one which is evaluated
+     * independently for each incoming traverser to the parent step. This is typically used in cases where the child
+     * is used to process or augment each traverser individually. See {@link LocalStep} or {@link ByModulating} as
+     * examples.
+     */
     public default <S, E> List<Traversal.Admin<S, E>> getLocalChildren() {
         return Collections.emptyList();
     }
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractAddEdgeStepPlaceholder.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractAddEdgeStepPlaceholder.java
@@ -66,20 +66,6 @@ public abstract class AbstractAddEdgeStepPlaceholder<S> extends AbstractAddEleme
     }
 
     @Override
-    public List<Traversal.Admin<S, Edge>> getLocalChildren() {
-        final List<Traversal.Admin<S, Edge>> childTraversals = super.getLocalChildren();
-        if (from != null) childTraversals.add((Traversal.Admin) from);
-        if (to != null) childTraversals.add((Traversal.Admin) to);
-        return childTraversals;
-    }
-
-    @Override
-    public void setTraversal(final Traversal.Admin<?, ?> parentTraversal) {
-        super.setTraversal(parentTraversal);
-        this.getLocalChildren().forEach(this::integrateChild);
-    }
-
-    @Override
     public int hashCode() {
         int hash = super.hashCode();
         if (from != null) {
@@ -99,6 +85,18 @@ public abstract class AbstractAddEdgeStepPlaceholder<S> extends AbstractAddEleme
             return true;
         }
         return false;
+    }
+
+    @Override
+    public List<Traversal.Admin<?, ?>> getLocalChildren() {
+        List<Traversal.Admin<?, ?>> childTraversals = super.getLocalChildren();
+        if (from != null) {
+            childTraversals.add(from instanceof GValueConstantTraversal ? ((GValueConstantTraversal<?, ?>) from).getConstantTraversal() : from);
+        }
+        if (to != null) {
+            childTraversals.add(to instanceof GValueConstantTraversal ? ((GValueConstantTraversal<?, ?>) to).getConstantTraversal() : to);
+        }
+        return childTraversals;
     }
 
     @Override
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractAddElementStepPlaceholder.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractAddElementStepPlaceholder.java
@@ -81,20 +81,33 @@ public abstract class AbstractAddElementStepPlaceholder<S, E extends Element, X 
     }
 
     protected void addTraversal(final Traversal.Admin<?, ?> traversal) {
+        integrateChild(traversal);
         TraversalHelper.getStepsOfAssignableClassRecursively(Scoping.class, traversal).forEach(s -> scopeKeys.addAll(s.getScopeKeys()));
     }
 
     @Override
-    public List<Traversal.Admin<S, E>> getLocalChildren() {
-        List<Traversal.Admin<S, E>> childTraversals = new ArrayList<>();
-        for (List<Object> values : properties.values()) {
-            for (Object value : values) {
+    public List<Traversal.Admin<?, ?>> getLocalChildren() {
+        List<Traversal.Admin<?, ?>> childTraversals = new ArrayList<>();
+        for (Map.Entry<Object, List<Object>> entry : properties.entrySet()) {
+            if (entry.getKey() instanceof Traversal) {
+                childTraversals.add((Traversal.Admin<?, ?>) ((Traversal) entry.getKey()).asAdmin());
+            }
+            for (Object value : entry.getValue()) {
                 if (value instanceof Traversal) {
-                    childTraversals.add((Traversal.Admin<S, E>) ((Traversal) value).asAdmin());
+                    childTraversals.add((Traversal.Admin<?, ?>) ((Traversal) value).asAdmin());
                 }
             }
         }
+        if (label != null) {
+            childTraversals.add(label);
+        }
         return childTraversals;
+    }
+
+    @Override
+    public void setTraversal(final Traversal.Admin<?, ?> parentTraversal) {
+        super.setTraversal(parentTraversal);
+        this.getLocalChildren().forEach(this::integrateChild);
     }
 
     @Override
@@ -174,6 +187,9 @@ public abstract class AbstractAddElementStepPlaceholder<S, E extends Element, X 
     public void addProperty(Object key, Object value) {
         if (key instanceof GValue) {
             throw new IllegalArgumentException("GValue cannot be used as a property key");
+        }
+        if (key instanceof Traversal) {
+            this.integrateChild(((Traversal<?, ?>) key).asAdmin());
         }
         if (value instanceof GValue) {
             traversal.getGValueManager().register((GValue<?>) value);
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractMergeElementStepPlaceholder.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractMergeElementStepPlaceholder.java
@@ -33,6 +33,7 @@ import org.apache.tinkerpop.gremlin.process.traversal.strategy.decoration.Partit
 import org.apache.tinkerpop.gremlin.process.traversal.traverser.TraverserRequirement;
 import org.apache.tinkerpop.gremlin.structure.T;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
@@ -55,6 +56,7 @@ public abstract class AbstractMergeElementStepPlaceholder<S, E> extends Abstract
     public AbstractMergeElementStepPlaceholder(Traversal.Admin traversal, final Traversal.Admin<?, Map<Object, Object>> mergeTraversal, final boolean isStart) {
         super(traversal);
         this.mergeTraversal = mergeTraversal;
+        this.integrateChild(mergeTraversal);
         this.isStart = isStart;
     }
 
@@ -71,6 +73,27 @@ public abstract class AbstractMergeElementStepPlaceholder<S, E> extends Abstract
     @Override
     public Set<TraverserRequirement> getRequirements() {
         return super.getRequirements();
+    }
+
+    @Override
+    public void setTraversal(final Traversal.Admin<?, ?> parentTraversal) {
+        super.setTraversal(parentTraversal);
+        this.getLocalChildren().forEach(this::integrateChild);
+    }
+    
+    @Override
+    public List<Traversal.Admin<?, ?>> getLocalChildren() {
+        List<Traversal.Admin<?, ?>> localChildren = new ArrayList<>();
+        if (mergeTraversal != null) {
+            localChildren.add(mergeTraversal);
+        }
+        if (onCreateTraversal != null) {
+            localChildren.add(onCreateTraversal);
+        }
+        if (onMatchTraversal != null) {
+            localChildren.add(onMatchTraversal);
+        }
+        return localChildren;
     }
 
     @Override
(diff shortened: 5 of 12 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/lambda/GValueConstantTraversal.java
```diff
@@ -57,6 +57,17 @@ public boolean equals(final Object other) {
         return constantTraversal.equals(other);
     }
 
+    @Override
+    public GValueConstantTraversal<S, E> clone() {
+        GValueConstantTraversal<S, E> clone = new GValueConstantTraversal<>(GValue.of(this.end.getName(), this.end.get()));
+        try {
+            clone.setGValueManager(this.getGValueManager().clone());
+        } catch (CloneNotSupportedException e) { //TODO:: handle properly
```
Cole-Greer, line 67: There needs to be a revisit for cloning of GValues, GValueManager, and StepPlaceholders. I consider this out of scope for this PR and intend to address this soon.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/TraversalParent.java
```diff
@@ -31,14 +33,30 @@
 import java.util.Set;
 
 /**
+ * Indicates that a step can contain child Traversals. Any step which implements this interface should override at least
+ * one of getGlobalChildren() or getLocalChildren(). All TraversalParents should call integrateChild() for all child
```
spmallette: nit: `{@link ...` the method references.

spmallette:
taking some build errors it seems - also, i'd think you need to backport changes for `DateDiffStep` and `TraversalSelectStep`.

Cole-Greer:
> taking some build errors it seems - also, i'd think you need to backport changes for `DateDiffStep` and `TraversalSelectStep`.

Ahh, new issues snuck their in with a last minute tweak to merge. Should be good now.

I'll CTR the fixes to `DateDiffStep` and `TraversalSelectStep` back to 3.7-dev separately.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/lambda/GValueConstantTraversal.java
andreachild: Note that there are fields in the parent class `AbstractLambdaTraversal` which are not cloned by this current logic such as `bypassTraversal`.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractAddEdgeStepPlaceholder.java
```diff
@@ -101,6 +87,18 @@ public boolean isParameterized() {
         return false;
     }
 
+    @Override
+    public List<Traversal.Admin<?, ?>> getLocalChildren() {
```
andreachild, line 91: Is the parent interface generic contract which is defined as `List<Traversal.Admin<S, E>> getLocalChildren` too narrow since it needs to be modified here?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/TraversalParent.java
```diff
@@ -31,14 +33,31 @@
 import java.util.Set;
 
 /**
+ * Indicates that a step can contain child Traversals. Any step which implements this interface should override at least
```
andreachild, line 36: Nit: Likely out of scope for this PR but I feel it would be helpful in general for TinkerPop to have more custom annotations. In this case there could be a custom annotation like `@ChildTraversal(scope=Type.LOCAL)` to mark step fields that store a child traversal. Then there can be a single common method which handles the presence of child traversals in `AbstractStep.java` like (pseudocode):

```
@Override
public void setTraversal(final Traversal.Admin<?, ?> traversal) {
    this.traversal = traversal;
    // retrieve any step fields annotated with @ChildTraversal
    // for each child traversal, call integrateChild
}

```

And then there could also be an `AbstractTravesalParent.java` which has:

```
<S, E> List<Traversal.Admin<S, E>> getGlobalChildren() {
    // use reflection to retrieve fields annotated with @ChildTraversal(type=GLOBAL)
}

<S, E> List<Traversal.Admin<S, E>> getLocalChildren() {
    // use reflection to retrieve fields annotated with @ChildTraversal(type=LOCAL)
}
```

Then when adding new steps or adding child traversals to an existing step, all you would need to do is add the custom annotation and you wouldn't have to 'remember' to do the boilerplate logic.

If you agree with this suggestion it's possible to use an annotation for these affected classes only and then slowly migrate towards 100% usage of the annotation in future PRs.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractAddElementStepPlaceholder.java
```diff
@@ -175,6 +188,9 @@ public void addProperty(Object key, Object value) {
         if (key instanceof GValue) {
             throw new IllegalArgumentException("GValue cannot be used as a property key");
         }
+        if (key instanceof Traversal) {
+            this.integrateChild(((Traversal<?, ?>) key).asAdmin());
```
andreachild, line 211: Should `integrateChild` also be called by `setLabel` method if the label is a traversal?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractMergeElementStepPlaceholder.java
andreachild, line 1: Looking at the `clone` logic, it appears that `properties` can contain traversals. If that is the case then `addProperty` should also call `integrateChild`.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractAddElementStepPlaceholder.java
andreachild, line 211: Does anything need to be done to 'de-integrate' a child traversal if a property is removed?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AddPropertyStepContract.java
```diff
@@ -47,14 +48,17 @@ default HashSet<PopInstruction> getPopInstructions() {
     Object getKey();
 
     /**
-     * Get the property value
+     * Gets the property value. If the value was originally passed as a {@link GValue<?>}, {@link ConstantTraversal <?>},
+     * or a literal value, then the literal value is returned. Otherwise, the MergeMap is returned in Traversal form.
```
andreachild: What is a `MergeMap` in this context?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AddPropertyStepPlaceholder.java
```diff
@@ -66,11 +69,19 @@ public AddPropertyStepPlaceholder(final Traversal.Admin traversal, final VertexP
...
+            this.integrateChild(((Traversal<?, ?>) this.key).asAdmin());
+        }
         if (valueObject instanceof GValue) {
             traversal.getGValueManager().register((GValue<?>) valueObject);
+            this.value = new GValueConstantTraversal<>((GValue<?>) valueObject);
+        } else if (valueObject instanceof Traversal) {
+            this.value = ((Traversal<?, ?>) valueObject).asAdmin();
+            this.integrateChild(((Traversal<?, ?>) valueObject).asAdmin());
+        } else {
+            this.value = new ConstantTraversal<>(valueObject);
```
andreachild, line 81: Do `ConstantTraversal`s not need to be integrated as a child traversal?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractMergeElementStepPlaceholder.java
```diff
@@ -65,14 +67,67 @@ protected Traverser.Admin<E> processNextStart() throws NoSuchElementException {
...
 
     @Override
     public Set<TraverserRequirement> getRequirements() {
         return super.getRequirements();
     }
 
+    @Override
+    public void setTraversal(final Traversal.Admin<?, ?> parentTraversal) {
+        super.setTraversal(parentTraversal);
+        this.getLocalChildren().forEach(this::integrateChild);
```
andreachild, line 109: Can you clarify why `setTraversal` needs to call `integrateChild` but the constructor which calls `super(traversal)` does not need to call `integrateChild`?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractAddEdgeStepPlaceholder.java
Cole-Greer: The `TraversalParent` interface itself does not use generics. The `S` and `E` type variables are only bound on the `getLocalChildren` method itself. That means that the `S` and `E` defined in `public default <S, E> List<Traversal.Admin<S, E>> getLocalChildren()` are not the same `S` and `E` defined in this implementing step classes.

They should be independent, the addEdge step here uses `<S, Edge>` (can start with anything and always produces an Edge), but it's child traversals do not produce edges, the from/to traversals will produce a Vertex or some ID type, the label traversal will produce a String...

I think it would be most clear if TraversalParent was updated to just return `List<Traversal.Admin<?, ?>>` instead, however I considered that out of scope for this PR.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/TraversalParent.java
Cole-Greer: That's definitely a compelling idea and would be an improvement over the status quo. I'm not sure about half starting it here in this PR though. I would rather such a change be made all at once, as I'm worried there are a few steps which will be difficult to adapt to the annotation model, and may require tweaking the design. The steps which come to mind which may be difficult are the ones like AddEdge which currently shove all of their arguments into a `Parameters` object and use that to manage child Traversals.

I think this might make for a good JIRA. I don't think it's prohibitively large to tackle all at once as a standalone task. There's only about 60 implementations of TraversalParent when you factor out abstract classes and such, and there's really only a handful of unique patterns within those, they are mostly simple and repetitive.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractAddElementStepPlaceholder.java
Cole-Greer: Yes it should, same goes for `setElementId`. That appears to be a gap in the testing. I will make that fix and add those cases.

Cole-Greer: Historically speaking, no. The main purpose of integrateChild is to ensure the child is consistent with the parent (has the correct parent set, both traversals merge their side effects, and now they also merge their GValueManagers). Historically speaking there hasn't been much concern about the state of the child after it has been removed, as it is no longer a concern for the TraversalParent. One could argue that the removed child should be given an EmptyTraversal as a new parent, however this has never been enforced in the past.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractMergeElementStepPlaceholder.java
Cole-Greer: In the context of mergeV/E, we currently never call `integrateChild` for the `properties`, and we we also never return any of the properties from `getLocalChildren`. The mergeV/E steps do not really consider the properties to be child traversals. The properties really only exist in mergeV/E to support `PartitionStrategy`, and it's arguably not the right way to do this. PartitionStrategy only ever stores String values in the properties too so there really should never be a Traversal here in practice.

The proper solution here is to introduce a dedicated interface such as `Partitioning` to support `PartitionStrategy`, and then get properties out of mergeV/E. I consider this out of scope for this PR, and for the meantime I'm happy with the current state here where although Traversal's may theoretically be possible in properties, they are not considered child traversals of this step.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractMergeElementStepPlaceholder.java
Cole-Greer: Every child traversal needs to be integrated when it's added to the step. Within the constructor, the only child traversal present is `mergeTraversal`, which gets integrated in the constructor. Any calls to `setOnMatch()`, `setOnCreate()`, or `setMerge()` result in integrating the newly added child traversal.

When `setTraversal()` is called, `this` step is being assigned a new parent, and we need to re-integrate all children such that their side effects and GValueManagers are synced with the new parent.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AddPropertyStepContract.java
Cole-Greer: That's a copy-paste error, it should be `value`.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AddPropertyStepPlaceholder.java
Cole-Greer: That's an interesting point. integrating a `ConstantTraversal` as a child won't really do much, as a ConstantTraversal does not make use of GValues, and unless a `bypassTraversal` is set, it can't really have a parent or side effects either.

For the sake of consistency though, it would make sense to also integrate the ConstantTraversal here. I will make the change.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractAddElementStepPlaceholder.java
Cole-Greer: Once the child is dropped from the parent step, the parent traversal has no references of the child. As long as the user isn't externally storing a reference to the child, it should be get picked up by garbage collection.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/lambda/GValueConstantTraversal.java
Cole-Greer: I have updated this method to include a call to `super.clone()`. I will do a further followup of cloning of GValues/StepPlaceholders in a followup PR.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/AbstractAddElementStepPlaceholder.java
Cole-Greer: I have updated both `setLabel` and `setElementId` to properly integrate child traversals now.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/lambda/GValueConstantTraversal.java
andreachild: Nit: the call to `super.clone` should go inside the try block

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/TraversalParentTest.java
```diff
@@ -0,0 +1,1028 @@
...
+
+import org.apache.tinkerpop.gremlin.structure.util.GraphFactoryTest;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import static org.apache.tinkerpop.gremlin.process.traversal.AnonymousTraversalSource.traversal;
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.junit.Assert.assertEquals;
```
andreachild: Nit: unused

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/TraversalParentTest.java
```diff
@@ -0,0 +1,1028 @@
...
+    private static GraphTraversalSource g = traversal().with(GraphFactoryTest.MockGraph.open(new MapConfiguration(Map.of("service-registry", mockedRegistry))));
+
+    @BeforeClass
+    public static void setupClass() {
+        when(mockedRegistry.get(any(), anyBoolean(), any())).thenReturn(mockedService);
+        when(mockedService.getRequirements()).thenReturn(Set.of());
+    }
+
+    @Parameterized.Parameter(value = 0)
+    public Class stepClass;
```
andreachild:
```suggestion
    public Class<Step<?,?>> stepClass;
```

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/TraversalParentTest.java
```diff
@@ -0,0 +1,1028 @@
...
+
+    @Parameterized.Parameter(value = 1)
+    public Traversal.Admin traversal;
```
andreachild:
```suggestion
    public Traversal.Admin<?, ?> traversal;
```

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/TraversalParentTest.java
```diff
@@ -0,0 +1,1028 @@
...
+
+        traversal = traversal.clone();
+
+        verifyExpectedParents(postStrategyExpectedGlobalChildren == null ? expectedGlobalChildren : postStrategyExpectedGlobalChildren,
+                postStrategyExpectedLocalChildren == null ? expectedLocalChildren : postStrategyExpectedLocalChildren,
+                "(cloned traversal, pre-strategy application)");
+    }
+
+    private void verifyExpectedParents(List<Traversal.Admin<?,?>> expectedGlobalChildren, List<Traversal.Admin<?,?>> expectedLocalChildren, String messageSuffix) {
+        List<Step> steps = TraversalHelper.getStepsOfAssignableClass(stepClass, traversal);
```
andreachild:
```suggestion
        List<Step<?, ?>> steps = TraversalHelper.getStepsOfAssignableClass(stepClass, traversal);
```

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/TraversalParentTest.java
```diff
@@ -0,0 +1,1028 @@
...
+        assertTrue(String.format("Expected a single step of class %s to test, found %d %s", stepClass.getName(), steps.size(), messageSuffix), steps.size() == 1);
+        assertTrue(String.format("Expected step to implement TraversalParent %s", messageSuffix), steps.get(0) instanceof TraversalParent);
+
+        TraversalParent testStep = (TraversalParent) steps.get(0);
+        assertThat(String.format("getGlobalChildren() did not produce the expected results %s", messageSuffix), testStep.getGlobalChildren(), containsInAnyOrder(expectedGlobalChildren.toArray()));
```
andreachild, line 989: Nit: this `assertThat` method is deprecated

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/TraversalParentTest.java
```diff
@@ -0,0 +1,1028 @@
...
+        for (Traversal.Admin<?, ?> child : testStep.getLocalChildren()) {
+            if (!(child instanceof AbstractLambdaTraversal)) {
+                // Use reference equality to ensure parent points to the correct object. It's not enough that the parents contents match.
+                assertTrue(String.format("Expected child traversal %s to have correctly assigned parent %s", child.toString(), messageSuffix), testStep == child.getParent());
+            }
+        }
+    }
+
+    private static Traversal<?, ?> appendComputerAwareEndStep(final Traversal<?, ?> traversal) {
+        traversal.asAdmin().addStep(new ComputerAwareStep.EndStep(traversal.asAdmin()));
```
andreachild:
```suggestion
        traversal.asAdmin().addStep(new ComputerAwareStep.EndStep<>(traversal.asAdmin()));
```

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/TraversalParentTest.java
andreachild, line 1: Can you pls do a quick find replace for:

`ConstantTraversal(` replace with `ConstantTraversal<>(`

`ValueTraversal(` replace with `ValueTraversal<>(`

andreachild:
VOTE +1 pending minor fixes to tests

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/lambda/GValueConstantTraversal.java
Cole-Greer: I'm still leaving this scope to a followup on cloning, although `AbstractLambdaTraversal.clone()` does not throw `CloneNotSupportedException` so adding it to the try does not have any effect. There's actually no way for `GValue` or `GValueManager` to throw `CloneNotSupportedException` either given the current implementation, so in the followup this will be corrected and the try-catch will go away altogether.

PR #3213, opened 2025-09-19
Improve GLV examples reliability and documentation

kirill-stepanishin:
This PR improves the reliability and developer experience of Gremlin Language Variant (GLV) examples across all supported languages.

## Issues Fixed
- Go WriteBufferSize=0 causing E1201 buffer errors and nil pointer crashes
- Node.js v22+ WebSocket API incompatibility
- Python dependency conflicts preventing example builds
- Improved documentation for running examples

## Changes Made
- Fix WriteBufferSize=0 validation in Go gorillaTransporter
- Add conditional WebSocket .on() method usage for Node.js compatibility
- Resolve dependency conflicts in Python examples
- Add comprehensive documentation with prerequisites and build steps
- Configure all examples to consistently use local development code


gremlin-dotnet/Examples/Connections/Connections.csproj
```diff
@@ -25,6 +25,6 @@ limitations under the License.
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Gremlin.Net" Version="3.7.0" />
+    <ProjectReference Include="../../src/Gremlin.Net/Gremlin.Net.csproj" />
```
vkagamlyk, line 28: The examples should work without the need to download the driver source code.

gremlin-go/examples/go.mod
```diff
@@ -21,9 +21,11 @@ go 1.24
 
 require github.com/apache/tinkerpop/gremlin-go/v3 v3.7.3
 
+replace github.com/apache/tinkerpop/gremlin-go/v3 => ../
```
vkagamlyk, line 24: Why is this necessary?

gremlin-javascript/examples/node/package.json
```diff
@@ -9,6 +9,6 @@
   "author": "Ryan Tan",
   "license": "Apache-2.0",
   "dependencies": {
-    "gremlin": "^3.7.3"
+    "gremlin": "file:../../src/main/javascript/gremlin-javascript"
```
vkagamlyk, line 12: same

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1687,37 +1743,54 @@ no `Graph` instance to deserialize a result into on the client-side. A workaroun
...
+Modern toy graph server:
+
 [source,shell]
 ----
 docker run -d -p 8182:8182 tinkerpop/gremlin-server conf/gremlin-server-modern.yaml
 ----
 
-Make sure to install all necessary packages:
+==== Prerequisites
+- Node.js v20 installed
```
Cole-Greer: Nit: Should this be node 22 now?

gremlin-dotnet/Examples/BasicGremlin/BasicGremlin.csproj
```diff
@@ -25,6 +25,6 @@ limitations under the License.
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Gremlin.Net" Version="3.7.0" />
+    <PackageReference Include="Gremlin.Net" Version="3.7.4" />
```
Cole-Greer: Nit: There is an inconsistency where the `BasicGremlin` and `ModernTraversals` examples are configured to use the latest released driver, and the `Connections` example is using the local dev build.

gremlin-python/src/main/python/setup.py
```diff
@@ -35,7 +35,7 @@
     fd.write("'''")
     fd.write(__doc__)
     fd.write("'''\n")
-    fd.write('version   = %r\n' % os.getenv('VERSION', '?').replace('-SNAPSHOT', '.dev-%d' % timestamp))
+    fd.write('version   = %r\n' % os.getenv('VERSION', '999.999.999.dev0').replace('-SNAPSHOT', '.dev-%d' % timestamp))
```
Cole-Greer, line 38: Was this change necessary to get the python example running against the local dev build? It might be worth looking at how the[ Maven Replacer plugin is currently used](https://github.com/apache/tinkerpop/blob/e39d795177c6d05deb1a463c2d7a2124d4fa56c7/gremlin-python/pom.xml#L70-L82) to inject the gremlin version into [useragent.py](https://github.com/apache/tinkerpop/blob/e39d795177c6d05deb1a463c2d7a2124d4fa56c7/gremlin-python/src/main/python/gremlin_python/driver/useragent.py#L21). Perhaps that approach can be adapted to inject the correct version here.

gremlin-dotnet/Examples/Connections/Connections.csproj
andreachild: @vkagamlyk we were thinking of the examples in each GLV module to be executed as part of the build against the current state of the driver code so that future breakages do not happen again (the examples are currently broken). The thought is to create a project level 'examples' folder which contains examples targeting latest released version which will not be executed as part of the build. Having a project-level examples folder also makes the examples more 'discoverable'.

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -525,27 +525,42 @@ can be found in GitHub link:https://github.com/apache/tinkerpop/tree/x.y.z/greml
...
+
+- *Running locally* – see the <<gremlin-server,Gremlin Server>> documentation.
+- *Running with Docker*:
+
 [source,shell]
 ----
 docker pull tinkerpop/gremlin-server
 ----
 
-The remote connection and basic Gremlin examples can be run on a clean server, which uses the default configuration file
```
andreachild, line 534: Was this wording incorrect? Curious why it was removed.

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -525,27 +525,42 @@ can be found in GitHub link:https://github.com/apache/tinkerpop/tree/x.y.z/greml
 and are designed to connect to a running <<gremlin-server,Gremlin Server>> configured with the
 `conf/gremlin-server.yaml` and `conf/gremlin-server-modern.yaml` files as included with the standard release packaging.
 
-To run the examples, first download an image of Gremlin Server from Docker Hub:
+There are two ways to run Gremlin Server for these examples:
+
+- *Running locally* – see the <<gremlin-server,Gremlin Server>> documentation.
+- *Running with Docker*:
+
```
andreachild: Since this guide assumes Docker will be used, suggest to restructure as such (same suggestion for all GLV sections):

```suggestion
This guide assumes Gremlin Server will be executed using Docker. Alternatively, Gremlin Server can run locally (see the <<gremlin-server,Gremlin Server>> documentation for this option).

To start Gremlin Server using Docker, first download an image of Gremlin Server from Docker Hub:
```

gremlin-dotnet/Examples/Connections/Connections.csproj
vkagamlyk: Replacing `PackageReference` with `ProjectReference` will make the examples less useful because they won't run without loading and building driver source code.

To build examples with different driver versions, solution is to add 2 `.csproj` files with different setting.

Also compare with other .net examples.

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1263,19 +1278,60 @@ execute it.
...
-All examples can then be run using your IDE of choice.
+Modern toy graph server:
+
+[source,shell]
+----
+docker run -d -p 8182:8182 tinkerpop/gremlin-server conf/gremlin-server-modern.yaml
+----
+
+==== Prerequisites
+- Java 11 (JDK) installed
```
andreachild: Same suggestion for all GLV sections:

```suggestion
- Compatible JDK installed (see <<development-environment,Development Environment>> for supported versions
```

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -2092,29 +2165,54 @@ can be found in GitHub link:https://github.com/apache/tinkerpop/tree/x.y.z/greml
...
+- .NET 8 SDK installed
+
+Navigate to the examples directory:
+
+[source,shell]
+----
+cd gremlin-dotnet/Examples
+----
+
+For each example project, navigate to its folder, then build and run:
```
andreachild: Q suggested the examples can be built and executed as such (without having to navigate to each folder):
```
# Build all projects
dotnet build Examples.sln

# Run specific examples
dotnet run --project BasicGremlin
dotnet run --project Connections  
dotnet run --project ModernTraversals
```

Does this work?

gremlin-dotnet/Examples/BasicGremlin/BasicGremlin.csproj
andreachild: This is not really a nit, inconsistency should be addressed.

gremlin-dotnet/Examples/Connections/Connections.csproj
andreachild: As an alternative to multiple `csproj` files how about single file with optional build property for example `dotnet build -p:UseLatestRelease` where default for `UseLatestRelease` is `true` but when running the examples as part of the build the value is set to `false`.

Or opposite scenario like `dotnet build -p:UseLocalReference` for example.

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
```diff
@@ -163,7 +163,13 @@ class Connection extends EventEmitter {
 
     this._ws.addEventListener('open', this.#handleOpen);
     this._ws.addEventListener('error', this.#handleError);
-    this._ws.on('unexpected-response', this.#handleUnexpectedResponse);
+    // Only attach unexpected-response listener for Node.js WebSocket (ws package)
+    // Browser WebSocket does not have this event and .on() method
+    if (!useGlobalWebSocket) {
```
andreachild: Can the check be changed to be consistent with the`off` condition?
```suggestion
    if (this._ws && 'on' in this._ws) {
```

gremlin-python/src/main/python/setup.py
kirill-stepanishin: Without this change pip fails with `InvalidVersion: Invalid version: '-'` when installing the local package. 
Maven Replacer would definitely be cleaner but would require a build step first, right now the examples run directly from source

docs/src/reference/gremlin-variants.asciidoc
kirill-stepanishin: Replaced specific versions with a link to development environment documentation in 44abbe0

gremlin-dotnet/Examples/BasicGremlin/BasicGremlin.csproj
kirill-stepanishin: Fixed in 44abbe0. Thank you!

docs/src/reference/gremlin-variants.asciidoc
kirill-stepanishin: That was an oversight, added a note about server configurations in 44abbe0. Thank you

docs/src/reference/gremlin-variants.asciidoc
kirill-stepanishin: Applied this suggestion, thanks!

docs/src/reference/gremlin-variants.asciidoc
kirill-stepanishin: Applied to all GLV prerequisites, thank you

docs/src/reference/gremlin-variants.asciidoc
kirill-stepanishin: It works, applied this simplification to the instructions. Thank you

gremlin-javascript/src/main/javascript/gremlin-javascript/lib/driver/connection.js
kirill-stepanishin: Applied, thank you

gremlin-dotnet/Examples/Connections/Connections.csproj
kirill-stepanishin: Thanks for the feedback! This is part of an ongoing discussion, will update based on the outcome.

gremlin-go/examples/go.mod
kirill-stepanishin: Makes the examples use local source code instead of published packages for immediate feedback and to apply fixes instantly. Broader strategy regarding examples is in discussion.

andreachild:
VOTE +1 with the assumption that there will be a quick follow up PR to add GLV examples to the root module which are pinned to the latest released version or some other solution which would give new users access to examples that work out of the box without config changes with the latest released GLV versions. It would be helpful to prep that follow up PR while this one is waiting for votes so that they could be merged in quicker succession.

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -525,27 +525,43 @@ can be found in GitHub link:https://github.com/apache/tinkerpop/tree/x.y.z/greml
...
-To start a server with the Modern graph preloaded, launch a new container with `docker run`:
+Modern toy graph server:
+
 [source,shell]
 ----
 docker run -d -p 8182:8182 tinkerpop/gremlin-server conf/gremlin-server-modern.yaml
 ----
 
-Each example can now be run with the following commands:
+The remote connection and basic Gremlin examples can be run on a clean server, while traversal examples should be run on a server with the Modern graph preloaded.
```
kenhuuu, line 551: Nit: Try to maintain consistency in line length with the rest of the asciidoc files. This should break at 120 characters.

kenhuuu:
VOTE +1, pending nits.

docs/src/reference/gremlin-variants.asciidoc
kirill-stepanishin: applied in a follow-up PR (https://github.com/apache/tinkerpop/pull/3218). Thank you!

gremlin-dotnet/Examples/Connections/Connections.csproj
kirill-stepanishin: Addressed in a follow-up PR https://github.com/apache/tinkerpop/pull/3218

PR #3214, opened 2025-09-20
Consolidate interfaces into step contracts

Cole-Greer:
This PR migrates all of a step's interfaces (`Mutating`, `Bypassing`, `FilteringBarrier`...) to the step contract and ensures that placeholder and concrete steps share common base classes (`ScalarMapStep`, `FilterStep`...). This ensures that all methods of a step are now accessible from the StepContract type, without any need to cast to concrete types. This also enforces complete alignment between the placeholder and concrete steps. After this change, steps with contract interfaces should only implement the single contract interface. Any additional interfaces must be added to the contract itself.

The only exception currently is `Configuring`, which has been omitted from `AddVertexStepContract`, `AddEdgeStepContract`, `AddPropertyStepContract`, `VertexStepContract`, and `GraphStepContract`. These are all steps which don't make use of with()-modulation in the reference implementations, and I'm reluctant to introduce configuring into these placeholders unnecessarily due to the complexity of keeping GValues properly isolated inside Parameters. I consider a revisit of Configuring as a followup task to be addressed separately from this change.

One additional change of note is the loss of the abstract `RangeStepPlaceholder` class. The shared class structure is unfortunately no longer feasible as RangeGlobal and RangeLocal extend from distinct base classes and Java does not allow for multiple inheritance. The size and complexity of the range placeholder classes are minimal, so there isn't much value in developing a new solution to share code between them.

VOTE +1

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/IsStepPlaceholder.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/IsStepPlaceholder.java
@@ -28,7 +28,6 @@ import org.apache.tinkerpop.gremlin.process.traversal.util.TraversalHelper;
 
 import java.util.Collection;
 import java.util.EnumSet;
-import java.util.NoSuchElementException;
 import java.util.Set;
 
 public final class IsStepPlaceholder<S> extends FilterStep<S> implements GValueHolder<S, S>, IsStepContract<S> {
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/RangeGlobalStepPlaceholder.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/RangeGlobalStepPlaceholder.java
@@ -41,6 +41,9 @@ public class RangeGlobalStepPlaceholder<S> extends FilterStep<S> implements Rang
 
     public RangeGlobalStepPlaceholder(final Traversal.Admin traversal, final GValue<Long> low, final GValue<Long> high) {
         super(traversal);
+        if (null == low || null == high) {
+            throw new IllegalArgumentException("RangeGlobalStepPlaceholder requires both low and high values to be non-null");
+        }
         this.low = low;
         this.high = high;
         traversal.getGValueManager().register(low);
@@ -70,9 +73,6 @@ public class RangeGlobalStepPlaceholder<S> extends FilterStep<S> implements Rang
     }
 
     public Long getLowRange() {
-        if (low == null) {
-            return null;
-        }
         if (low.isVariable()) {
             this.traversal.getGValueManager().pinVariable(low.getName());
         }
@@ -80,9 +80,6 @@ public class RangeGlobalStepPlaceholder<S> extends FilterStep<S> implements Rang
     }
 
     public Long getHighRange() {
-        if (high == null) {
-            return null;
-        }
         if (high.isVariable()) {
             this.traversal.getGValueManager().pinVariable(high.getName());
         }
@@ -139,30 +136,6 @@ public class RangeGlobalStepPlaceholder<S> extends FilterStep<S> implements Rang
     }
 
     @Override
-    public boolean hasNextBarrier() {
-        return this.starts.hasNext();
-    }
-
-    @Override
-    public TraverserSet<S> nextBarrier() throws NoSuchElementException {
-        if(!this.starts.hasNext())
-            throw FastNoSuchElementException.instance();
-        final TraverserSet<S> barrier = (TraverserSet<S>) this.traversal.getTraverserSetSupplier().get();
-        while (this.starts.hasNext()) {
-            barrier.add(this.starts.next());
-        }
-        return barrier;
-    }
-
-    @Override
-    public void addBarrier(final TraverserSet<S> barrier) {
-        IteratorUtils.removeOnNext(barrier.iterator()).forEachRemaining(traverser -> {
-            traverser.setSideEffects(this.getTraversal().getSideEffects());
-            this.addStart(traverser);
-        });
-    }
-
-    @Override
     public Collection<GValue<?>> getGValues() {
         Set<GValue<?>> gValues = new HashSet<>();
         if (low.isVariable()) {
@@ -184,7 +157,15 @@ public class RangeGlobalStepPlaceholder<S> extends FilterStep<S> implements Rang
 
     @Override
     public RangeGlobalStepPlaceholder<S> clone() {
-        return new RangeGlobalStepPlaceholder<>(traversal, low, high);
+        RangeGlobalStepPlaceholder<S> clone = (RangeGlobalStepPlaceholder<S>) super.clone();
+        clone.bypass = this.bypass;
+        try {
+            clone.low = this.low.clone(); //TODO:: cleanup unnecessary try-catch
+            clone.high = this.high.clone();
+        } catch (CloneNotSupportedException e) {
+            throw new RuntimeException(e);
+        }
+        return clone;
     }
 
 }
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/RangeLocalStepPlaceholder.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/RangeLocalStepPlaceholder.java
@@ -38,6 +38,9 @@ public class RangeLocalStepPlaceholder<S> extends ScalarMapStep<S,S> implements 
 
     public RangeLocalStepPlaceholder(final Traversal.Admin traversal, final GValue<Long> low, final GValue<Long> high) {
         super(traversal);
+        if (null == low || null == high) {
+            throw new IllegalArgumentException("RangeLocalStepPlaceholder requires both low and high values to be non-null");
+        }
         this.low = low;
         this.high = high;
         traversal.getGValueManager().register(low);
@@ -67,9 +70,6 @@ public class RangeLocalStepPlaceholder<S> extends ScalarMapStep<S,S> implements 
     }
 
     public Long getLowRange() {
-        if (low == null) {
-            return null;
-        }
         if (low.isVariable()) {
             this.traversal.getGValueManager().pinVariable(low.getName());
         }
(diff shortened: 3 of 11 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/IsStepPlaceholder.java
```diff
@@ -32,7 +31,7 @@
 import java.util.NoSuchElementException;
```
andreachild: Not used

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/RangeLocalStepPlaceholder.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    @Override
+    public boolean isParameterized() {
+        return low.isVariable() || high.isVariable();
+    }
+
+    @Override
+    public void updateVariable(String name, Object value) {
+        if (name.equals(low.getName())) {
```
andreachild, line 57: Isn't it possible for `GValue.name` to be null for literals? Would this throw NPE in that case?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/RangeLocalStepPlaceholder.java
```diff
@@ -0,0 +1,151 @@
...
+    @Override
+    public RangeLocalStep<S> asConcreteStep() {
+        RangeLocalStep<S> step = new RangeLocalStep<>(traversal, low.get(), high.get());
+        TraversalHelper.copyLabels(this, step, false);
+        return step;
+    }
+
+    @Override
+    public RangeLocalStepPlaceholder<S> clone() {
+        return new RangeLocalStepPlaceholder<>(traversal, low, high);
```
andreachild: Shouldn't `clone` be called on the traversal, high, low?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/TailGlobalStep.java
andreachild, line 1: Unused imports

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/TailGlobalStepPlaceholder.java
```diff
@@ -50,6 +56,11 @@ public Traverser.Admin<S> processNextStart() {
...
     }
 
+    @Override
+    public void setBypass(final boolean bypass) {
+        this.bypass = bypass;
+    }
+
     @Override
     public TailGlobalStepPlaceholder<S> clone() {
         final TailGlobalStepPlaceholder<S> clone = (TailGlobalStepPlaceholder<S>) super.clone();
```
andreachild, line 69: Missing cloning of `limit` and `bypass`?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/TailGlobalStepPlaceholder.java
```diff
@@ -74,10 +85,50 @@ public GValue<Long> getLimitAsGValue() {
...
     }
 
+    @Override
+    public MemoryComputeKey<TraverserSet<S>> getMemoryComputeKey() {
+        return MemoryComputeKey.of(this.getId(), new RangeGlobalStep.RangeBiOperator<>(this.limit.get()), false, true);
+    }
+
+    @Override
+    public void processAllStarts() {
+
```
andreachild:
```suggestion
throw new IllegalStateException("TailGlobalStepPlaceholder is not executable");
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/RangeGlobalStepPlaceholder.java
```diff
@@ -126,8 +124,42 @@ public GValue<Long> getHighRangeAsGValue() {
...
+        this.bypass = bypass;
+    }
+
+    @Override
+    public void processAllStarts() {
+        throw new IllegalStateException("RangeGlobalStepPlaceholder is not executable");
+    }
+
+    @Override
+    public boolean hasNextBarrier() {
```
andreachild: The barrier related methods are duplicated between placeholder and step and can be moved to shared contract interface. For example in the contract interface:

```
ExpandableStepIterator<S> getStarts();

@Override
default boolean hasNextBarrier() {
    return getStarts().hasNext();
}

@Override
default TraverserSet<S> nextBarrier() throws NoSuchElementException {
    if(!getStarts().hasNext())
        throw FastNoSuchElementException.instance();
    final TraverserSet<S> barrier = (TraverserSet<S>) this.getTraversal().getTraverserSetSupplier().get();
    while (getStarts().hasNext()) {
        barrier.add(getStarts().next());
    }
    return barrier;
}

@Override
default void addBarrier(final TraverserSet<S> barrier) {
    IteratorUtils.removeOnNext(barrier.iterator()).forEachRemaining(traverser -> {
        traverser.setSideEffects(this.getTraversal().getSideEffects());
        this.addStart(traverser);
    });
}
```

This is also the same for the `TailGlobalStep` and `TailGlobalStepPlaceholder`

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/GraphStep.java
andreachild, line 1: Unused import

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/VertexStepPlaceholder.java
```diff
@@ -180,4 +180,9 @@ public void updateVariable(String name, Object value) {
     public Collection<GValue<?>> getGValues() {
         return Arrays.asList(edgeLabels);
     }
+
+    @Override
+    public void close() throws Exception {
```
andreachild, line 185: Can be moved to contract interface as default method as it's shared with concrete step

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AddPropertyStep.java
andreachild, line 1: Unused imports

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/RangeLocalStepPlaceholder.java
Cole-Greer: I've noticed `clone()` issues in a handful of the placeholder steps. I'm accumulating fixes in a local branch which I plan to rebase and push once all of these outstanding PRs are merged. I'll pull the fix for `RangeLocalStepPlaceholder` and `RangeGlobalStepPlaceholder` up into this PR as I did technically touch these methods already (although they aren't modified from the old abstract class in this PR).

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/TailGlobalStepPlaceholder.java
Cole-Greer: This is definitely a bug, I'm already working on a followup on cloning in the placeholder steps. I will call this fix out of scope for this PR.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/RangeGlobalStepPlaceholder.java
Cole-Greer: This is a great intermediate suggestion. I had avoided default implementations for these methods as `getStarts()` was only available in AbstractStep, and I didn't think it was within the scope of this PR to migrate that to Step (although such a change would be worth pursuing). This is a good fix for now as it prevents the unnecessary code duplication, without expanding this PR with the added risk of modifying the Step interface.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/RangeLocalStepPlaceholder.java
Cole-Greer: It's ok if `low = GValue.of(null, 1)` as this works out to `name.equals(null)` which is just false. We could get NPEs if any of `name`, `low`, or `high` were null. Those all fall into the category of things which should never happen. I will add explicit checks to block nulls and throw IllegalArgumentExceptions if someone ever tries to pass a null to this class. The concrete RangeLocalStep stores primitive `long` values here, and the placeholders really shouldn't need to figure out how to properly map a null to a primitive argument.

As for `name` being null, I'm leaning towards ignoring it. There's no reason a user should ever call `updateVariable()` with a null name, and I think it's fine to just let that surface as an NPE to them.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/VertexStepPlaceholder.java
Cole-Greer: Doing this would require pulling `protected void closeIterator()` from `FlatMapStep` up to `VertexStepContract` which forces it to be made public as well. The preferable solution would probably be making FlatMapStep implement AutoClosable, and pushing `public void close() throws Exception {closeIterator();}` down into FlatMapStep.

Cole-Greer: My preference would be to leave this change as out of scope for this PR, however if you'd like to push for it I will look into any possible concerns from pushing AutoClosable down to FlatMapStep

PR #3216, opened 2025-09-25
Add splitByElement() method for parsing nested list or set

yunlingTao:
Added `splitByElement()` method in `gremlin-test` which splits strings without breaking elements containing commas inside brackets or braces. 
It replaces `split(",")` in list and set parsing to handle structures like `s[l[1,2,3], l[2,3,4]]`.

### DIFF
--- a/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/features/StepDefinition.java
+++ b/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/features/StepDefinition.java
@@ -184,7 +184,7 @@ public final class StepDefinition {
     private List<Pair<Pattern, Function<String,Object>>> objectMatcherConverters = new ArrayList<Pair<Pattern, Function<String,Object>>>() {{
         // expects json so that should port to the Gremlin script form - replace curly json braces with square ones
         // for Gremlin sake.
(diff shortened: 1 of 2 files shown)


gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/features/StepDefinition.java
```diff
@@ -642,6 +640,27 @@ private String convertToString(final String pkey, final String pvalue) {
         return String.format("\"%s\"", pvalue);
     }
 
+    private List<String> splitByElement(String s) {
```
andreachild, line 643: Would be good to add a scenario to a feature test would would execute this new logic. Perhaps adding a scenario to the `Fold.feature` which does something like `g.V().fold().fold()` which will result in a list of lists?

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/features/StepDefinition.java
andreachild, line 1: Do other GLVs need a similar change? For example the parsing logic for lists and sets in `feature-steps.js` and `feature_steps.py`.

yunlingTao:
Closing in favor of #3379 which extends this to all GLVs and adds Gherkin scenarios.

PR #3217, opened 2025-09-26
Modify SampleGlobalStep memory key to merge traverser sets, keeping o…

andreachild:
…nly traversers that have been through the most loops.


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/SampleGlobalStep.java
```diff
@@ -154,7 +219,7 @@ private Optional<ProjectedTraverser<S, Number>> createProjectedTraverser(final T
 
     @Override
     public Set<TraverserRequirement> getRequirements() {
-        return this.getSelfAndChildRequirements(TraverserRequirement.BULK);
+        return this.getSelfAndChildRequirements(TraverserRequirement.BULK, TraverserRequirement.NESTED_LOOP);
```
Cole-Greer, line 222: Note: I don't think it's necessary to add a nested loop requirement to this step, as in theory the only time this will be needed is if its inside a nested repeat step, in which case `RepeatStep` itself will already be providing the nested loop requirement. A similar argument would apply to any other steps which may be given "loop awareness" such as `RangeGlobalStep`. Note how even `LoopsStep` does not explicitly set a single loop or nested loop requirement.

PR #3218, opened 2025-09-26
Add root-level GLV examples and fix existing example issues

kirill-stepanishin:
## Summary

This PR improves GLV examples reliability and adds user-friendly root-level examples that work out-of-the-box with published drivers.

This is a follow-up to PR #3213 which fixed GLV examples reliability issues but required switching examples to local code so that fixes could be applied immediately. However, that approach meant new users had to build TinkerPop from source to run examples. This PR addresses that limitation by adding a separate set of root-level examples using published drivers.

## Issues Fixed

**GLV Examples Reliability (from PR #3213):**
- Go WriteBufferSize=0 causing E1201 buffer errors and nil pointer crashes
- Node.js v22+ WebSocket API incompatibility (fix originally from PR #3204 by @xhuberty)
- Python dependency conflicts preventing example builds
- .NET outdated dependencies
- Inconsistent documentation for running examples

**User Experience (New):**
- No easy way for new users to run examples without building TinkerPop from source
- Examples scattered across different locations with varying setup requirements

## Changes Made

**GLV Examples Fixes (from PR #3213):**
- Fix WriteBufferSize=0 validation in Go gorillaTransporter
- Add conditional WebSocket .on() method usage for Node.js compatibility (credit: @xhuberty's fix from PR #3204)
- Resolve dependency conflicts in Python examples
- Configure all module-level examples to consistently use local development code
- Add comprehensive documentation with prerequisites and build steps

**Root-Level Examples (New):**
- Create `glv-examples/` directory with examples for all 5 GLVs using published drivers (3.7.4)
- Add comprehensive README with server setup and usage instructions
- Update main documentation to reference root-level examples for users
- Update release process documentation to include version update requirements
- Add developer documentation explaining dual example structure

## Approach

- **Module-level examples** (`gremlin-*/examples/`): Use local development code for contributors and testing
- **Root-level examples** (`glv-examples/`): Use published drivers for end users, manually updated during releases

## Known Limitations

- **JavaScript root-level examples**: Only work with Node.js v20, fail with v22+ until next release incorporates WebSocket compatibility fixes (originally from PR #3204)
- **Go root-level examples**: Connections example has nil pointer issue until next release incorporates buffer validation fixes
- **Manual version updates**: Root-level example versions must be manually updated as part of the release process (automatic updates are out of scope)

## Credits

JavaScript WebSocket compatibility fix originally contributed by @xhuberty in PR #3204.

PR #3231, opened 2025-10-03
Integrate Python driver examples into automated build process

kirill-stepanishin:
### Summary
Integrates Python Gremlin examples into CI to ensure they remain functional in future releases. Examples now execute automatically after integration tests and fail the build if broken.

### Changes Made
- Added example execution to gremlin-python-integration-tests container in docker-compose.yml
- Made server URLs configurable via environment variables (replaces hardcoded localhost:8182)
- Added cleanup to mutating examples
- Added graph binding detection - uses gmodern in CI, g locally for proper environment handling


gremlin-python/src/main/python/examples/basic_gremlin.py
```diff
@@ -25,9 +26,13 @@
 
 
 def main():
-    rc = DriverRemoteConnection('ws://localhost:8182/gremlin', 'g')
+    server_url = os.getenv('GREMLIN_SERVER_URL', 'ws://localhost:8182/gremlin').format(45940)
```
andreachild, line 31: Nit: add comment with context where this weird port is coming from
```suggestion
    # if there is a port placeholder in the env var then we are running with docker so set appropriate port 
    server_url = os.getenv('GREMLIN_SERVER_URL', 'ws://localhost:8182/gremlin').format(45940)
```

gremlin-python/src/main/python/examples/basic_gremlin.py
```diff
@@ -47,6 +52,9 @@ def main():
     for person in people_marko_knows:
         print("marko knows " + person)
 
+    # clean added data
+    g.V().drop().iterate()
+
```
andreachild: It is a little bit risky to drop all data at the beginning of this example since it will be running with other tests and examples. We can isolate the data for this test by using a specialized label (.ie `person-py-ex` instead of `person`).

```suggestion
# basic Gremlin: adding and retrieving data
v1 = g.add_v('person-py-ex').property('name', 'marko').next()
v2 = g.add_v('person-py-ex').property('name', 'stephen').next()
v3 = g.add_v('person-py-ex').property('name', 'vadas').next()

# be sure to use a terminating step like next() or iterate() so that the traversal "executes"
# iterate() does not return any data and is used to just generate side-effects (i.e. write data to the database)
g.V(v1).add_e('knows').to(v2).property('weight', 0.75).iterate()
g.V(v1).add_e('knows').to(v3).property('weight', 0.75).iterate()

# retrieve the data from the "marko" vertex
marko = g.V().has('person-py-ex', 'name', 'marko').values('name').next()
print("name: " + marko)

# find the "marko" vertex and then traverse to the people he "knows" and return their data
people_marko_knows = g.V().has('person-py-ex', 'name', 'marko').out('knows').values('name').to_list()
for person in people_marko_knows:
print("marko knows " + person)

    # clean added data
    g.V().has_label('person-py-ex').drop().iterate()

```

gremlin-python/src/main/python/examples/connections.py
```diff
@@ -40,7 +43,8 @@ def with_remote():
     #
     # which starts it in "console" mode with an empty in-memory TinkerGraph ready to go bound to a
     # variable named "g" as referenced in the following line.
-    rc = DriverRemoteConnection('ws://localhost:8182/gremlin', 'g')
+    server_url = os.getenv('GREMLIN_SERVER_URL', 'ws://localhost:8182/gremlin').format(45940)
```
andreachild, line 49: Nit: add comment with context where this weird port is coming from - same comment for all references to env vars in this example
```suggestion
    # if there is a port placeholder in the env var then we are running with docker so set appropriate port 
    server_url = os.getenv('GREMLIN_SERVER_URL', 'ws://localhost:8182/gremlin').format(45940)
```

gremlin-python/src/main/python/examples/connections.py
andreachild, line 1: Suggest isolating the data added and removed for this example using a specialized label such as `py-conn-ex` and also cleaning up the data at the end of each function instead of only once at the beginning of the entire example.

gremlin-python/src/main/python/examples/connections.py
```diff
@@ -57,32 +61,57 @@ def with_remote():
 
 # connecting with plain text authentication
 def with_auth():
-    rc = DriverRemoteConnection('ws://localhost:8182/gremlin', 'g', username='stephen', password='password')
+    server_url = os.getenv('GREMLIN_SERVER_BASIC_AUTH_URL', 'ws://localhost:8182/gremlin').format(45941)
+    # turn off certificate verification for testing purposes only
+    ssl_opts = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+    ssl_opts.check_hostname = False
+    ssl_opts.verify_mode = ssl.CERT_NONE
```
andreachild: I'm assuming this is only needed if the host is not localhost? If yes then can this be surrounded by a conditional check to only disable cert verification if not localhost?

gremlin-python/src/main/python/examples/connections.py
```diff
@@ -57,32 +61,57 @@ def with_remote():
...
-    rc.close()
+        v = g.add_v().iterate()
+        count = g.V().count().next()
+        print("Vertex count: " + str(count))
+
+        # clean added data
+        g.V().drop().iterate()
+        rc.close()
+    except Exception as e:
+        print(f"Kerberos authentication failed (expected in test environment): {e}")
```
andreachild: It looks like `conftest.py` has kerberos auth test case which is expected to pass with docker, perhaps this example needs to be configured similarly to `conftest.py` to succeed?

gremlin-python/src/main/python/examples/basic_gremlin.py
kirill-stepanishin: Applied in 6729d40b1b21b54e078b50637de55293aebc3455. Thank you

gremlin-python/src/main/python/examples/basic_gremlin.py
kirill-stepanishin: Applied in 6729d40b1b21b54e078b50637de55293aebc3455. Thank you

gremlin-python/src/main/python/examples/connections.py
kirill-stepanishin: Works if using `socket.gethostname()` instead of `gremlin-server-test` as in `conftest.py`. Applied in 6729d40b1b21b54e078b50637de55293aebc3455

gremlin-python/src/main/python/examples/connections.py
kirill-stepanishin: added conditional check in 6729d40b1b21b54e078b50637de55293aebc3455

gremlin-python/src/main/python/examples/connections.py
kirill-stepanishin: Added custom labels to all mutating examples (6729d40b1b21b54e078b50637de55293aebc3455). Thank you

gremlin-python/src/main/python/examples/connections.py
kirill-stepanishin: Added this comment to all url configurations

gremlin-python/docker-compose.yml
```diff
@@ -75,7 +75,12 @@ services:
       && python3 ./setup.py test
       && python3 ./setup.py install
       && radish -f dots -e -t -b ./radish ./gremlin-test --user-data='serializer=application/vnd.gremlin-v3.0+json'
-      && radish -f dots -e -t -b ./radish ./gremlin-test --user-data='serializer=application/vnd.graphbinary-v1.0';
+      && radish -f dots -e -t -b ./radish ./gremlin-test --user-data='serializer=application/vnd.graphbinary-v1.0'
+      && echo 'Running examples...'
```
kenhuuu, line 77: Very small nit: This command is getting a bit unwieldy. Consider splitting the examples off into another service that depends on the integration tests.

gremlin-python/src/main/python/examples/connections.py
```diff
@@ -40,60 +44,87 @@ def with_remote():
...
     rc.close()
 
 
 # connecting with Kerberos SASL authentication
 def with_kerberos():
-    rc = DriverRemoteConnection('ws://localhost:8182/gremlin', 'g', kerberized_service='gremlin@hostname.your.org')
+    # if there is a port placeholder in the env var then we are running with docker so set appropriate port 
+    server_url = os.getenv('GREMLIN_SERVER_URL', 'ws://localhost:8182/gremlin').format(45942)
+    kerberos_hostname = os.getenv('KRB_HOSTNAME', socket.gethostname())
+    kerberized_service = f'test-service@{kerberos_hostname}'
```
kenhuuu, line 91: This Kerberos example is too specific to the integration test Gremlin Server setup. It should be generalized or at the very least commented to say that it likely won't run for the user.

gremlin-python/src/main/python/examples/modern_traversals.py
```diff
@@ -31,7 +32,16 @@ def main():
     # This example requires the Modern toy graph to be preloaded upon launching the Gremlin server.
     # For details, see https://tinkerpop.apache.org/docs/current/reference/#gremlin-server-docker-image and use
     # conf/gremlin-server-modern.yaml.
-    rc = DriverRemoteConnection('ws://localhost:8182/gremlin', 'g')
+    # if there is a port placeholder in the env var then we are running with docker so set appropriate port 
+    server_url = os.getenv('GREMLIN_SERVER_URL', 'ws://localhost:8182/gremlin').format(45940)
+    
+    # CI uses port 45940 with gmodern binding, local uses 8182 with g binding
```
kenhuuu, line 38: If these examples are meant to be user facing, then I'd argue comments like these about usage with the internal CI system are confusing.

kirill-stepanishin: The user-facing examples are going to be in root after https://github.com/apache/tinkerpop/pull/3218 gets merged. The current set of examples will be set to use local development code instead of published packages, and will be tested as part of CI

andreachild: The port 45940 is the one used by docker integration tests so not user facing

gremlin-python/src/main/python/examples/basic_gremlin.py
```diff
@@ -25,28 +26,33 @@
...
+    # if there is a port placeholder in the env var then we are running with docker so set appropriate port 
+    server_url = os.getenv('GREMLIN_SERVER_URL', 'ws://localhost:8182/gremlin').format(45940)
+    rc = DriverRemoteConnection(server_url, 'g')
     g = traversal().with_remote(rc)
 
     # basic Gremlin: adding and retrieving data
-    v1 = g.add_v('person').property('name', 'marko').next()
-    v2 = g.add_v('person').property('name', 'stephen').next()
-    v3 = g.add_v('person').property('name', 'vadas').next()
+    v1 = g.add_v('person-py-ex').property('name', 'marko').next()
```
andreachild: Suggest to extract the label name to a constant so that it can be easily changed in one line instead of multiple. This would make it easier to keep in sync between user-facing examples and development examples. Same comment for all python examples.

gremlin-python/docker-compose.yml
kirill-stepanishin: Good suggestion! I tried and the main challenge is that the Gremlin server stops when integration tests complete, so a separate examples service can't connect to the same server instance. This would require either launching separate server instances for examples or reworking the docker compose orchestration to keep the server alive between test phases. I don't know if it's worth it

gremlin-python/src/main/python/examples/basic_gremlin.py
andreachild: I had another thought - ideally these examples stay in sync with the root level examples as much as possible so to achieve this we could:
1. extract the label used for the examples into a variable which by default is `person` but can be overridden by environment variable set in the docker config (similar to how the URL is set)
2. do not add dropping of the added nodes at the end of the example (unless the original example had done so) - since we are using an isolated node label for the development examples and the examples are executed after the integration tests and do not assert any counts it should be OK to not clean up the data

gremlin-python/docker-compose.yml
kirill-stepanishin: After further investigation, splitting the examples into a separate service introduces more complexity than it solves. I think we'll just leave it as is.

gremlin-python/src/main/python/examples/connections.py
kirill-stepanishin: The GLV-level examples aren’t meant to be user-facing. The root-level GLV examples won’t include those adaptations since they won’t be run as part of CI.

gremlin-python/src/main/python/examples/connections.py
```diff
@@ -40,15 +44,15 @@ def with_remote():
     #
     # which starts it in "console" mode with an empty in-memory TinkerGraph ready to go bound to a
     # variable named "g" as referenced in the following line.
-    rc = DriverRemoteConnection('ws://localhost:8182/gremlin', 'g')
+    # if there is a port placeholder in the env var then we are running with docker so set appropriate port 
+    server_url = os.getenv('GREMLIN_SERVER_URL', 'ws://localhost:8182/gremlin').format(45940)
+    vertex_label = os.getenv('VERTEX_LABEL', 'connection')
```
andreachild: Can these be moved to global variables so they don't need to be redefined for each function?

gremlin-python/src/main/python/examples/connections.py
andreachild, line 1: Can the simple changes to this file be also copied to the root level examples with the goal of keeping the root level and dev examples as in-sync as possible? Perhaps not the disabling of cert auth or kerberos specific changes but the others could be copied.

gremlin-python/src/main/python/examples/basic_gremlin.py
kirill-stepanishin: Implemented in 7f3eadd3157c7393b318ff3c444f18ee4e1173c4, thank you for the suggestion!

gremlin-python/src/main/python/examples/connections.py
kirill-stepanishin: Added in 0872c13e4a2602cfebfe58c700f8b7a28107f1a3, thank you

gremlin-python/src/main/python/examples/connections.py
kirill-stepanishin: root-level level `connections.py` already has new vertex labels. What other changes did you have in mind?

andreachild: root level examples can have constants for label as well (just don't need to retrieve the env variable override)

kirill-stepanishin: Implemented in bd2367b3c62f83acec1bfcfe5c7c937682c1ff1c, Thank you.

andreachild:
VOTE +1 thanks!

Cole-Greer:
VOTE +1

kenhuuu:
VOTE +1

andreachild:
Hoping to merge this today, I also forgot to ask for a changelog entry

kirill-stepanishin:
> Hoping to merge this today, I also forgot to ask for a changelog entry

Added a generic changelog entry for all GLV examples in 7c3bc33fb5a38ab2baf4417ef1d37bf0440ef547

andreachild:
Unfortunately I need to ask you to make a separate PR to target these same changes but against `master` branch. It seems there are a few non-trivial changes that are needed to get the dev-level examples to work against `master` code. Once the `master` PR is approved I will merge both together.

PR #3233, opened 2025-10-07
Removed `AggregateLocalStep`, `aggregate(Scope, String)`, and `store()`

xiazcy:
As follow-up to the Lazy vs Eager evaluation [proposal](https://github.com/apache/tinkerpop/blob/master/docs/src/dev/future/proposal-scoping-5.asciidoc), removed `Scope` from `aggregate('x')`, which removes `AggregateLocalStep`, as well as the deprecated equivalent step `store()`. `AggregateGlobalStep` is now `AggregateStep` with default globally scoped (eager aggregation). To achieve lazy aggregation, use the existing step `local()`, which is designed for flow control.

Updated relevant docs and added additional tests to ensure `local(aggregate('x'))` is semantically equivalent to `(aggregate(local, 'x'))`.


andreachild:
There's mention of `aggregate(local)` in `gremlin-variants.asciidoc` - should it be modified?

> * The `subgraph()`-step is not supported by any variant that is not running on the Java Virtual Machine as there is
no `Graph` instance to deserialize a result into on the client-side. A workaround is to replace the step with
`aggregate(local)` and then convert those results to something the client can use locally.

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -275,6 +275,38 @@ The properties file in the above example can either point to a remote configurat
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-3017[TINKERPOP-3017]
 
+==== Removal of Scopes in aggregate() Step
+
+Scores are not removed from the `aggregate()` step. The `AggregateStep` is now globally scoped by default.
```
andreachild: Is 'scores' a typo?

```suggestion
The scope parameter has been removed from the `aggregate()` step. The `AggregateStep` is now globally scoped by default.
```

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -275,6 +275,38 @@ The properties file in the above example can either point to a remote configurat
...
+The only way to achieve lazy evaluation with `aggregate()` is using `local()`.
```
andreachild:
```suggestion
Lazy evaluation with `aggregate()` can still be achieved by wrapping the step in `local()`.
```

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -275,6 +275,38 @@ The properties file in the above example can either point to a remote configurat
...
+
+[source,text]
+----
+// 3.7.x - scopes are still allowed (though deprecated starting in 3.7.5+)
```
andreachild:
```suggestion
// 3.7.x - scope is still supported (though deprecated starting in 3.7.5+)
```

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/EarlyLimitStrategyTest.java
```diff
@@ -91,10 +92,7 @@ public static Iterable<Object[]> generateTestParameters() {
                     {__.out().map(__.identity()).map(__.identity()).range(50, 100).map(__.identity()).map(__.identity()).range(50, -1), __.out().discard(), Collections.emptyList()},
                     {__.out().map(__.identity()).map(__.identity()).range(50, 100).map(__.identity()).map(__.identity()).range(60, -1), __.out().discard(), Collections.emptyList()},
                     {__.out().map(__.identity()).map(__.identity()).range(50, 100).as("a").map(__.identity()).map(__.identity()).range(60, -1).as("b"), __.out().discard(), Collections.emptyList()},
-                    {__.out().range(50, 100).store("a").range(50, -1), __.out().range(50, 100).store("a").discard(), Collections.emptyList()},
```
andreachild, line 94: Should these be replaced with `local(aggregate())`?

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/ProductiveByStrategyTest.java
```diff
@@ -84,12 +84,6 @@ public static Iterable<Object[]> generateTestParameters() {
                 {__.aggregate("a").by("name"),
                         __.aggregate("a").by(new ValueTraversal<>("name", coalesce(nameValueTraversal, nullTraversal).asAdmin())),
                         Collections.emptySet()},
-                {__.aggregate(Scope.local,"a").by("name"),
```
andreachild, line 87: Should these be replaced with `local(aggregate())`?

gremlin-dotnet/test/Gremlin.Net.UnitTest/Process/Traversal/Translator/GroovyTranslatorTests.cs
andreachild, line 1: I'm not sure why this entire file seems to have been changed but it looks like only this was removed:

```
{
    _g.V("3").Values<object>("runways").Store("x").V('4').Values<object>("runways").Store("x")
        .By(__.Constant(1)).V('6').Store("x").By(__.Constant(1)).Select<object>("x").Unfold<object>()
        .Sum<object>(),
    "g.V('3').values('runways').store('x').V('4').values('runways').store('x').by(__.constant(1)).V('6').store('x').by(__.constant(1)).select('x').unfold().sum()"
},
```

Should this instead be changed to reference `local(aggregate())`?

gremlin-go/driver/translator_test.go
```diff
@@ -234,12 +234,6 @@ func Test_translator_Translate(t *testing.T) {
 			},
 			equals: "g.withSack(0).V('3','5').sack(sum).by('runways').sack()",
 		},
-		{
-			assert: func(g *GraphTraversalSource) *GraphTraversal {
-				return g.V("3").Values("runways").Store("x").V("4").Values("runways").Store("x").By(T__.Constant(1)).V("6").Store("x").By(T__.Constant(1)).Select("x").Unfold().Sum()
```
andreachild, line 239: Should this be changed to use `local(aggregate())`?

gremlin-python/src/main/python/tests/process/test_translator.py
```diff
@@ -145,10 +145,7 @@ def test_translations(self):
         # 36
         tests.append([g.withSack(0).V('3', '5').sack(Operator.sum_).by('runways').sack(),
                       "g.withSack(0).V('3','5').sack(Operator.sum).by('runways').sack()"])
-        # 37
-        tests.append([g.V('3').values('runways').store('x').V('4').values('runways').store('x').by(__.constant(1)).V(
```
andreachild, line 149: Should this be changed to use `local(aggregate())`?

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AggregateTest.java
```diff
@@ -58,10 +58,6 @@
 @RunWith(GremlinProcessRunner.class)
 public abstract class AggregateTest extends AbstractGremlinProcessTest {
 
-    ////// global
-
-    public abstract Traversal<Vertex, List<String>> get_g_V_name_aggregateXglobal_xX_capXxX();
```
andreachild, line 63: Should the removed code be modified to use `local(aggregate())` instead?

docs/src/upgrade/release-3.8.x.asciidoc
andreachild: Can you add reasoning why this was done, for example what challenges were caused when it was allowed? Can reference the proposal?

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/EarlyLimitStrategyTest.java
xiazcy: No, EarlyLimitStrategy doesn't apply to aggregate global (even if it's wrapped inside of a local). This is the one of the thing that's a strategy limitation and not a semantics difference.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/ProductiveByStrategyTest.java
xiazcy: This seems to be a miss. I'll replace.

gremlin-dotnet/test/Gremlin.Net.UnitTest/Process/Traversal/Translator/GroovyTranslatorTests.cs
xiazcy: It was changed from CRLF to LF. These are translations tests for store(), which I don't quite see the need in changing given local() wrapping steps is tested elsewhere too. I could still add them.

gremlin-go/driver/translator_test.go
xiazcy: Similar to above comment in .NET.

gremlin-python/src/main/python/tests/process/test_translator.py
xiazcy: Similar to above comment in .NET.

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/AggregateTest.java
xiazcy: I didn't think there is the need. These process tests duplicated the feature tests in Java (I think the feature tests were based off of this), which are now obsolete, especially given we've added remote execution of feature tests for Java. There is plan to deprecate and remove it completely at some point, and it hasn't been updated for the past few versions.

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -275,6 +275,38 @@ The properties file in the above example can either point to a remote configurat
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-3017[TINKERPOP-3017]
 
+==== Removal of Scopes in aggregate() Step
```
spmallette: "Scopes" - plural. how about "aggregate with Scope Removed"?

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -275,6 +275,38 @@ The properties file in the above example can either point to a remote configurat
...
+
+The scope parameter has been removed from the `aggregate()` step. The `AggregateStep` is now globally scoped by default.
```
spmallette: This concepts behind `store ` and `aggregate` have been around since 3.0 inception. Might be nice to include a little more explanation here as to why this change was deemed necessary.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -635,31 +631,23 @@ g.V().out('knows').aggregate('x').by('age').cap('x')  <1>
...
 [gremlin-groovy,modern]
 ----
-g.V().aggregate(global, 'x').limit(1).cap('x')
-g.V().aggregate(local, 'x').limit(1).cap('x')
-g.withoutStrategies(EarlyLimitStrategy).V().aggregate(local,'x').limit(1).cap('x')
+g.V().aggregate('x').limit(1).cap('x')
+g.V().local(aggregate('x')).limit(1).cap('x')
 ----
 
-It is important to note that `EarlyLimitStrategy` introduced in 3.3.5 alters the behavior of `aggregate(local)`.
```
spmallette, line 649: hmm, that's a nice side-effect of removal. something to call out in Upgrade Docs to help flesh those out more? 

Just a check but have you confirmed that all these examples do in fact return the same results as the ones with `aggregate(local)`?

xiazcy: I've manually checked that the examples in the docs run, but I can do a doc build to make sure.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -454,8 +454,8 @@ image:side-effect-lambda.png[width=175,float=right]
 [gremlin-groovy,modern]
 ----
 g.V().hasLabel('person').sideEffect(System.out.&println) <1>
-g.V().sideEffect(outE().count().aggregate(local,"o")).
-      sideEffect(inE().count().aggregate(local,"i")).cap("o","i") <2>
+g.V().sideEffect(outE().count().local(aggregate("o"))).
```
Cole-Greer, line 457: Nit: It might be nice to pull all of these docs changes (aside from the semantics docs I suppose) back into the 3.7 deprecation PR. The existing examples in 3.7-dev will still run, however it is preferable to point users towards the preferred `local(aggregate())` usage instead of the deprecated `aggregate(local)`

Cole-Greer:
VOTE +1

docs/src/recipes/appendix.asciidoc
```diff
@@ -142,7 +142,7 @@ The following example assumes that the edges point in the `OUT` direction. Assum
 ----
 g.V().where(without("x")).as("a").
   outE().as("e").inV().as("b").
-  filter(bothE().where(neq("e")).otherV().where(eq("a"))).aggregate(local, "x").
+  filter(bothE().where(neq("e")).otherV().where(eq("a"))).local(aggregate("x")).
```
spmallette, line 145: i think i've asked this elsewhere, but again, have we tested these examples manually to be sure our results are the same just by adding the `local` wrapping?

docs/src/recipes/collections.asciidoc
```diff
@@ -190,9 +190,9 @@ of "java"). Note however that the `by()` modulator overrides that traverser comp
 the list of vertices in "v". Those vertices are unfolded to retrieve the name property from each and then are reduced
 with `fold()` back into a list to be stored in the side-effected named "b".
 <5> A similar use of `aggregate()` as the previous step, though this one turns "e" into a stream of edges to calculate
-the `mean()` to store in a `List` called "c". Note that `aggregate()` (short form for `aggregate(global)`) was used
-here instead of `aggregate(local)`, as the former is an eager collection of the elements in the stream
-(`aggregate(local)` is lazy) and will force the traversal to be iterated up to that point before moving forward.
+the `mean()` to store in a `List` called "c". Note that `aggregate()` with `local()` was used
```
spmallette, line 193: i think some of this content is in the wrong place. 

we deprecated in 3.7.x, so we want folks to start using the new form there. if we only deprecate and don't present the new pattern along with it, how will they understand the new form as they face deprecation? i think the 3.7.x docs should reflect the changes in the reference docs, recipes, etc. in that version, UNLESS, we've changed something in the semantics that only allow these changes in 3.8.0 (in which case there really isn't a deprecation path and maybe the #3234 doesn't make sense? thoughts?

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -293,6 +293,89 @@ The properties file in the above example can either point to a remote configurat
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-3017[TINKERPOP-3017]
 
+==== `aggregate()` with `Scope` Removed
+
+The `Scope` parameter is being removed from `aggregate()` to fix inconsistency between two different use cases: flow
+control vs. per-element application. This change aligns all side effect steps (none of the others have scope arguments)
+and reserves the `Scope` parameter exclusively for "traverser-local" application patterns, eliminating confusion about
+its contextual meanings.
```
Cole-Greer: I think this becomes a bit clearer if we start by establishing more context:
```suggestion
The meaning of `Scope` parameters in `aggregate()` have always been unique compared to all other "scopable" steps.
`aggregate(global)` is a `Barrier`, which blocks the traversal until all traversers have been aggregated into the side
effect, where `aggregate(local)` is non-blocking, and will allow traversers to pass before the side effect has been
fully aggregated. This is inconsistent with the semantics of `Scope` in all other steps. For example `dedup(global)`
filters duplicates across the entire traversal stream, while `dedup(local)` filters duplicates within individual `List`
traversers.

The `Scope` parameter is being removed from `aggregate()` to fix inconsistency between the two different use cases: flow
control vs. per-element application. This change aligns all side effect steps (none of the others have scope arguments)
and reserves the `Scope` parameter exclusively for "traverser-local" application patterns, eliminating confusion about
its contextual meanings.
```

docs/src/upgrade/release-3.8.x.asciidoc
```diff
@@ -293,6 +293,89 @@ The properties file in the above example can either point to a remote configurat
...
+==>[29,27,32]
+==>[29,27,32]
+==>[29,27,32,35]
+----
+
+An important behavioral difference exists between the removed `aggregate(local)` and its replacement `local(aggregate())`.
+In 3.8.0, `local()` changed from traverser-based to object-based processing, always debulking incoming traversers into
+individual objects. This causes `local(aggregate())` to behave differently from the original `aggregate(local)`, which
+preserved bulked traversers under the old `local()` semantics. This creates an irreplaceable use case, making this a
+_breaking change_.
```
```suggestion
An slight behavioral difference exists between the removed `aggregate(local)` and its replacement `local(aggregate())`
with respect to handling of bulked traversers. In 3.8.0, `local()` changed from traverser-local to object-local processing,
always debulking incoming traversers into individual objects. This causes `local(aggregate())` to show true lazy, 1 object
at a time aggregation, differing from the original `aggregate(local)`, which always consumed bulked traversers atomically.
There is no workaround to preserve the old "traverser-local" semantics.
```

xiazcy:
VOTE +1

PR #3274, opened 2025-11-06
Fix JavaTranslator's `has` handling for null arguments

andrii0lomakin:
- Updated `has()` special case logic to address potential issues when first or third arguments are `null`. Example : `g.E().has("knows","weight",null)` this query can lead to NPE on server execution as JavaTranslator can choose GraphTravers#has(String, Sring, P) method instead of GraphTravers#has(String, Sring, Object)

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/JavaTranslator.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/JavaTranslator.java
@@ -22,10 +22,7 @@ package org.apache.tinkerpop.gremlin.jsr223;
 import org.apache.commons.configuration2.BaseConfiguration;
 import org.apache.commons.configuration2.Configuration;
 import org.apache.commons.configuration2.MapConfiguration;
-import org.apache.tinkerpop.gremlin.process.traversal.Bytecode;
-import org.apache.tinkerpop.gremlin.process.traversal.Translator;
-import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
-import org.apache.tinkerpop.gremlin.process.traversal.TraversalSource;
+import org.apache.tinkerpop.gremlin.process.traversal.*;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal;
 import org.apache.tinkerpop.gremlin.process.traversal.lambda.CardinalityValueTraversal;
 import org.apache.tinkerpop.gremlin.process.traversal.step.util.BulkSet;
@@ -60,7 +57,7 @@ public final class JavaTranslator<S extends TraversalSource, T extends Traversal
     private final S traversalSource;
     private final Class<?> anonymousTraversal;
     private static final Map<Class<?>, Map<String, List<ReflectedMethod>>> GLOBAL_METHOD_CACHE = new ConcurrentHashMap<>();
-    private final Map<Class<?>, Map<String,Method>> localMethodCache = new ConcurrentHashMap<>();
+    private final Map<Class<?>, Map<String, Method>> localMethodCache = new ConcurrentHashMap<>();
     private final Method anonymousTraversalStart;
 
     private JavaTranslator(final S traversalSource) {
@@ -109,7 +106,7 @@ public final class JavaTranslator<S extends TraversalSource, T extends Traversal
         return StringFactory.translatorString(this);
     }
 
-    ////
+    /// /
 
     private Object translateObject(final Object object) {
         if (object instanceof Bytecode.Binding)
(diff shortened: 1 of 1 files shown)


andreachild:
It would be good to add a unit test for this. Here is a suggested one since there are no existing (direct) unit tests for the `JavaTranslator`. Note that if you execute the test below without the code fix it will randomly fail/pass depending on the order of methods loaded into the `JavaTranslator.GLOBAL_METHOD_CACHE`.

```
package org.apache.tinkerpop.gremlin.jsr223;

import org.apache.tinkerpop.gremlin.process.traversal.Bytecode;
import org.apache.tinkerpop.gremlin.process.traversal.P;
import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
import org.apache.tinkerpop.gremlin.structure.util.empty.EmptyGraph;
import org.junit.Test;

import static org.junit.Assert.assertEquals;

public class JavaTranslatorTest {
    private GraphTraversalSource g = EmptyGraph.instance().traversal();
    private JavaTranslator<GraphTraversalSource, Traversal.Admin<?, ?>> translator = JavaTranslator.of(EmptyGraph.instance().traversal());

    @Test
    public void shouldTranslateHasWithObjectArgValue() {
        final Bytecode bytecode = new Bytecode();
        bytecode.addStep("E");
        bytecode.addStep("has", "knows", "weight", 1.0);
        final Traversal.Admin<?, ?> translation = translator.translate(bytecode);
        assertEquals(g.E().has("knows", "weight", 1.0).asAdmin(), translation);
    }

    @Test
    public void shouldTranslateHasWithPredicateArgValue() {
        final Bytecode bytecode = new Bytecode();
        bytecode.addStep("E");
        bytecode.addStep("has", "knows", "weight", P.eq(1.0));
        final Traversal.Admin<?, ?> translation = translator.translate(bytecode);
        assertEquals(g.E().has("knows", "weight", P.eq(1.0)).asAdmin(), translation);
    }

    @Test
    public void shouldTranslateHasWithNullThirdArgValue() {
        final Bytecode bytecode = new Bytecode();
        bytecode.addStep("E");
        bytecode.addStep("has", "knows", "weight", null);
        final Traversal.Admin<?, ?> translation = translator.translate(bytecode);
        assertEquals(g.E().has("knows", "weight", (String) null).asAdmin(), translation);
    }

}
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/JavaTranslator.java
```diff
@@ -22,10 +22,7 @@
 import org.apache.commons.configuration2.BaseConfiguration;
 import org.apache.commons.configuration2.Configuration;
 import org.apache.commons.configuration2.MapConfiguration;
-import org.apache.tinkerpop.gremlin.process.traversal.Bytecode;
-import org.apache.tinkerpop.gremlin.process.traversal.Translator;
-import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
-import org.apache.tinkerpop.gremlin.process.traversal.TraversalSource;
+import org.apache.tinkerpop.gremlin.process.traversal.*;
```
andreachild, line 25: It looks like your IDE has auto style formatting which differs from the conventions used in the tinkerpop project and thus the wildcard import has been added and various whitespace changes as well. According to https://tinkerpop.apache.org/docs/current/dev/developer/#_code_style we try to keep the style consistent with the existing conventions.

> Contributors should examine the current code base to determine what the code style patterns are and should match their style to what is already present. Of specific note however, TinkerPop does not use "import wildcards" - IDEs should be adjusted accordingly to not auto-wildcard the imports.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/JavaTranslator.java
```diff
@@ -301,20 +298,30 @@ private Object invokeMethod(final Object delegate, final Class<?> returnType, fi
...
+                            //first case has((T)null, ...) instead of has((String)null, ...)
+                            if (null == newArguments[0] &&
+                                    parametersTypes[0].isAssignableFrom(org.apache.tinkerpop.gremlin.structure.T.class)) {
+                                found = false;
+                            } else if (newArguments.length == 3 && newArguments[2] == null && parametersTypes[0] == String.class &&
+                                    parametersTypes[1] == String.class &&
+                                    parametersTypes[2] == P.class) {
+                                //the second case has(String, String, (P)(null)) instead of has(String,String, (Object)null)
+                                found = false;
+                            }
```
andreachild, line 324: I'm surprised the has overloads with 2 method args have not also surfaced also a problem (it could just be by luck depending on the order of the methods loaded in the global cache). You could change the logic to address both the 2 and 3 arg method overloads:

```suggestion
                            } else if (newArguments[newArguments.length - 1] == null && parametersTypes[newArguments.length - 1] == P.class) {
                                //the second case has(String, String, (P)(null)) instead of has(String,String, (Object)null)
                                //or has(String, (P)(null)) instead of has(String, (Object)null)
                                found = false;
                            }
```

andrii0lomakin:
Hi.
I will reimplement this fix.

The correct fix is to sort the list of methods during the building method cache by isAssigned from the relationship. That will fix all current and future problems.

andreachild:
Hi @andrii0lomakin I opened https://github.com/apache/tinkerpop/pull/3278 because I wanted a fix for the issue you found in 3.8.0 for which the release vote is happening today which unfortunately doesn't give you enough time to refactor the change as you mentioned. Please do continue with this PR and it can be added to 3.8.1 or 4.0. Thanks!

PR #3297, opened 2026-01-23
Integrate Java driver examples into automated build process

kirill-stepanishin:
Integrates Java driver examples into CI to ensure they remain functional and catch breaking changes in the driver. Examples now execute automatically during integration tests and fail the build if broken.

## Changes Made

- Moved examples from `gremlin-driver/src/main/java/examples` to `gremlin-server/src/test/java/examples` to add example execution into existing integration test infrastucture
- Added `DriverExamplesIntegrateTest` to run examples against test server
- Removed compiler exclusion for examples in `gremlin-driver/pom.xml`

Root-level Java driver examples remain available in `gremlin-examples/gremlin-java`


gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/DriverExamplesIntegrateTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @BeforeClass
+    public static void setUpEnvironment() {
+        // Set environment variables for examples to connect to test server
+        System.setProperty("GREMLIN_SERVER_HOST", "localhost");
+        System.setProperty("GREMLIN_SERVER_PORT", "45940");
+    }
+
+    @AfterClass
+    public static void tearDownAfterClass() {
```
kenhuuu: Nit: This static method's name doesn't match the setup method's name.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/DriverExamplesIntegrateTest.java
kenhuuu, line 1: It's a bit odd to me that this file runs the main function of other classes. I know those classes are there to mimic the ones from gremlin-examples, but seeing as how this is going to be tied to integration test, you might as well just make each of those classes a test in this file instead? Just a thought, doesn't really need to be changed if you feel that it makes more sense this way.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/DriverExamplesIntegrateTest.java
kirill-stepanishin: Renamed to tearDownEnvironment to match the setup method.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/driver/DriverExamplesIntegrateTest.java
kirill-stepanishin: I agree that making the examples complete tests makes more sense in this isolated context. It's done more to mimic the rest of the GLVs, where currently the examples are also simply executed as part of the integration tests. Perhaps adding output validation and converting examples into proper tests could be a future piece of work.
As of now, I think it makes sense to keep the current approach to stay consistent with the other GLVs, and the main goal of catching the breaking changes that affect examples is still achieved.

Cole-Greer:
VOTE +1

kenhuuu:
VOTE +1

PR #3381, opened 2026-04-08
TINKERPOP-3233 Go support for escaped characters

GumpacG:
https://issues.apache.org/jira/browse/TINKERPOP-3233
This is the first part of resolving the ticket.

This change fixes a bug in Go in which characters that should be escaped are being replaced by a dot.
Bug:
The traversal `result, err := g.V().Has(vertexLabel, "name", "m   m").Next()` would generate the GremlinLang `g.V().has(vertexLabel, "name", "m.m").next()`.
New behaviour:
The traversal `result, err := g.V().Has(vertexLabel, "name", "m   m").Next()` would generate the GremlinLang `g.V().has(vertexLabel, "name", "m\tm").next()`.

A test has been added to `gremlinlang_test.go` for testing all cases that should be escaped.

### DIFF
--- a/gremlin-go/driver/gremlinlang_test.go
+++ b/gremlin-go/driver/gremlinlang_test.go
@@ -674,6 +674,12 @@ func Test_GremlinLang(t *testing.T) {
 			},
 			equals: "g.inject(NaN).is(eq(NaN))",
 		},
+		{
+			assert: func(g *GraphTraversalSource) *GraphTraversal {
+				return g.V().Has("name", "\"marko\n\r\t\b\f\"")
+			},
+			equals: "g.V().has(\"name\",\"\\\"marko\\n\\r\\t\\b\\f\\\"\")",
+		},
 	}
 
 	var testsToRun []test
--- a/gremlin-go/driver/gremlinlang.go
+++ b/gremlin-go/driver/gremlinlang.go
@@ -110,16 +110,40 @@ func (gl *GremlinLang) addToGremlin(name string, args ...interface{}) error {
 	return nil
 }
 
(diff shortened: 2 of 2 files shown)


gremlin-go/driver/gremlinlang_test.go
```diff
@@ -674,6 +674,12 @@ func Test_GremlinLang(t *testing.T) {
 			},
 			equals: "g.inject(NaN).is(eq(NaN))",
 		},
+		{
+			assert: func(g *GraphTraversalSource) *GraphTraversal {
+				return g.V().Has("name", "\"marko\n\r\t\b\f\"")
+			},
+			equals: "g.V().has(\"name\",\"\\\"marko\\n\\r\\t\\b\\f\\\"\")",
```
Cole-Greer, line 681: Nit: Throw a single quote in there too
```suggestion
				return g.V().Has("name", "\"marko\n\r\t\b\f\'")
			},
			equals: "g.V().has(\"name\",\"\\\"marko\\n\\r\\t\\b\\f\\\'\")",
```

Cole-Greer:
Thanks @GumpacG, LGTM. VOTE +1

gremlin-go/driver/gremlinlang_test.go
GumpacG: This will fail because single quotes are not escaped similar to Java since the string arguments are always wrapped in double quotes. It doesn't need to be added as it is not an escaped character.

xiazcy:
VOTE +1

kenhuuu:
VOTE +1

PR #3395, opened 2026-04-21
Response Streaming in JS

Cole-Greer:
Add streaming HTTP response handling to gremlin-javascript, enabling incremental result consumption via fetch response.body ReadableStream.

- Add StreamReader abstraction for async byte reading over streaming and buffered sources
- Refactor all ~20 GraphBinary serializers from sync deserialize(buffer) to async deserialize(reader)
- Refactor GraphBinaryReader.readResponse() to use StreamReader
- Add Connection.stream() using fetch response.body ReadableStream
- Add Client.stream() returning AsyncGenerator
- Wire Traversal API (next(), hasNext(), toList()) to streaming for incremental consumption, matching Go GLV behavior
- submit() remains non-streaming, buffers full response
- Remove dead readable-stream dependency and Readable imports

VOTE +1

### DIFF
--- a/docs/src/upgrade/release-4.x.x.asciidoc
+++ b/docs/src/upgrade/release-4.x.x.asciidoc
@@ -51,6 +51,39 @@ deserialization in GraphBinary is unchanged. Applications that depend on the `uu
 `gremlin-javascript` brought it in as a transitive dependency should add it directly to their own `package.json`
 if they still need it.
 
+==== JS HTTP Streaming Response Support
+
+The JavaScript driver now supports incremental HTTP streaming. Results are deserialized from the server response as
+they arrive, rather than buffering the entire response before processing.
+
+Traversal API terminal steps (`next()`, `toList()`, `hasNext()`) are now truly incremental. `next()`
+returns the first result as soon as it is deserialized from the wire, without waiting for the full response.
+In 3.x, `next()` waited for all WebSocket frames before returning.
+
+`Client.stream()` now returns an `AsyncGenerator` for direct incremental consumption. This is a breaking change
+from 3.x where `stream()` returned a Node.js `Readable`. The new return type works in both Node.js and browsers:
+
+[source,javascript]
+----
+// 3.x — Readable stream (no longer supported)
+// const stream = client.stream('g.V()');
+// stream.on('data', (resultSet) => { ... });
+
+// 4.0 — AsyncGenerator
+for await (const item of client.stream('g.V()', null)) {
+  console.log(item);
+  if (someCondition) break; // stops reading from the HTTP stream
+}
+----
+
+`Client.submit()` remains unchanged. It still buffers the full response and returns `Promise<ResultSet>`.
+
+=== Upgrading for Providers
+
+==== Graph System Providers
+
+==== Graph Driver Providers
+
 == TinkerPop 4.0.0-beta.2
 
 *Release Date: April 1, 2026*
--- a/gremlin-js/gremlin-javascript/lib/driver/connection.ts
+++ b/gremlin-js/gremlin-javascript/lib/driver/connection.ts
@@ -25,21 +25,19 @@ import { Buffer } from 'buffer';
 import { EventEmitter } from 'eventemitter3';
 import type { Agent } from 'node:http';
 import ioc from '../structure/io/binary/GraphBinary.js';
+import StreamReader from '../structure/io/binary/internals/StreamReader.js';
 import * as utils from '../utils.js';
 import ResultSet from './result-set.js';
 import {RequestMessage} from "./request-message.js";
-import {Readable} from "stream";
 import ResponseError from './response-error.js';
 import { Traverser } from '../process/traversal.js';
 
-const { DeferredPromise } = utils;
 const { graphBinaryReader, graphBinaryWriter } = ioc;
 
 const responseStatusCode = {
   success: 200,
   noContent: 204,
   partialContent: 206,
-  authenticationChallenge: 407,
 };
 
 export type HttpRequest = {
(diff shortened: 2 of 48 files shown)


gremlin-js/gremlin-javascript/lib/driver/connection.ts
```diff
@@ -115,20 +113,87 @@ export default class Connection extends EventEmitter {
...
+      // For error responses, buffer and parse the error body
+      const buffer = Buffer.from(await response.arrayBuffer());
+      const errorMessage = `Server returned HTTP ${response.status}: ${response.statusText}`;
+      const reader = this.#getReaderForContentType(response.headers.get("Content-Type"));
+
+      if (reader) {
+        try {
+          const deserialized = await reader.readResponse(buffer);
+          const attributes = new Map();
+          if (deserialized.status.exception) {
```
kenhuuu: I believe this is not the correct format of the response for the HTTP API. There are no longer any status attributes.

gremlin-js/gremlin-javascript/lib/structure/io/binary/internals/GraphBinaryReader.js
```diff
@@ -35,7 +46,13 @@ export default class GraphBinaryReader {
...
   }
 
-  readResponse(buffer) {
+  /**
+   * Read a complete response from a Buffer. Used by the non-streaming submit() path.
+   * Returns the full { status, result } object after reading all data.
+   * @param {Buffer} buffer
+   * @returns {Promise<{status: {code, message, exception}, result: {data: any[], bulked: boolean}}>}
+   */
+  async readResponse(buffer) {
```
kenhuuu, line 55: Just curious, is there any benefit to making GraphBinary deserialization async? You need to use the value almost immediately after calling await which is effectively synchronous?

kenhuuu:
Is there a test for checking that the GraphBinaryReader handles values that are input across chunk boundaries? This is technically covered by the new tests for the StreamReader, but it isn't quite the integration test that I'm expecting as it doesn't go from test input through the GraphBinary deserializers.

gremlin-js/gremlin-javascript/test/unit/graphbinary/error-cases-test.js
```diff
@@ -26,365 +26,380 @@
 import { assert } from 'chai';
 import { Buffer } from 'buffer';
 import ioc from '../../../lib/structure/io/binary/GraphBinary.js';
+import StreamReader from '../../../lib/structure/io/binary/internals/StreamReader.js';
 import { P, TextP, Traverser } from '../../../lib/process/traversal.js';
 import { OptionsStrategy } from '../../../lib/process/traversal-strategy.js';
 
 const { anySerializer, intSerializer, longSerializer, stringSerializer, listSerializer, mapSerializer, uuidSerializer, dateTimeSerializer, floatSerializer, shortSerializer, byteSerializer, bigIntegerSerializer, binarySerializer, setSerializer, enumSerializer } = ioc;
```
GumpacG: Should this test file cover all serializer errors? For example, there is a missing UnspecifiedNullSerializer error.

GumpacG (commented):
Could you also add an integration test that would verify early result consumption before the full response completes?

Otherwise, LGTM. VOTE +1

gremlin-js/gremlin-javascript/lib/structure/io/binary/internals/GraphBinaryReader.js
Cole-Greer: The deserializers all need to call something like `await reader.readInt32BE()` to block on reads if they are waiting for the next chunk. The deserialization itself needs to be async in order to call `await`.

gremlin-js/gremlin-javascript/lib/driver/connection.ts
Cole-Greer: I've updated to the correct format.

gremlin-js/gremlin-javascript/test/unit/graphbinary/error-cases-test.js
Cole-Greer: It would be nice if it covers all errors, in general I would argue that existing gaps are out of scope of this PR. I've pushed in a handful of extra cases including the UnspecifiedNullSerializer error for an unexpected value flag. It's possible there are other missing cases, I've added the ones I've found so far.

Cole-Greer:
@kenhuuu 
> Is there a test for checking that the GraphBinaryReader handles values that are input across chunk boundaries? This is technically covered by the new tests for the StreamReader, but it isn't quite the integration test that I'm expecting as it doesn't go from test input through the GraphBinary deserializers.

I've added some cases at the end of [GraphBinaryReader-test.js](https://github.com/apache/tinkerpop/pull/3395/files#diff-20fdc9b9a769c017fbab412994fc9f21e09e524a6207be28240c98d8b6e1c11e) to give more coverage here. They validate that GraphBinaryReader can handle items split across chunk boundaries, as well as validating it can incrementally produce the first result prior to all data being available in to the stream. It doesn't quite reach the level of a full integration test as it's not wired into DRC. Such a test would be useful but needs dedicated server infrastructure, which I consider out of scope here.

@GumpacG 
> Could you also add an integration test that would verify early result consumption before the full response completes?

Piggy-backing on my above response, I believe the new unit tests added to GraphBinaryReader are the best we can reasonably do for now, without dedicated server infrastructure to control size and timing of response chunks.

gremlin-js/gremlin-javascript/lib/driver/connection.ts
```diff
@@ -115,20 +113,81 @@ export default class Connection extends EventEmitter {
...
+            exception: deserialized.status.exception,
+          });
+        } catch (err) {
+          if (err instanceof ResponseError) throw err;
+        }
+      }
+
+      throw new ResponseError(errorMessage, {
+        code: response.status,
+        message: response.statusText,
```
xiazcy, line 177: Still trying to understand how it works here, but are we missing the path for `application/json` errors for the `stream()` case?

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -51,6 +51,39 @@ deserialization in GraphBinary is unchanged. Applications that depend on the `uu
...
+==== JS HTTP Streaming Response Support
+
+The JavaScript driver now supports incremental HTTP streaming. Results are deserialized from the server response as
+they arrive, rather than buffering the entire response before processing.
+
+Traversal API terminal steps (`next()`, `toList()`, `hasNext()`) are now truly incremental. `next()`
+returns the first result as soon as it is deserialized from the wire, without waiting for the full response.
+In 3.x, `next()` waited for all WebSocket frames before returning.
+
+`Client.stream()` now returns an `AsyncGenerator` for direct incremental consumption. This is a breaking change
```
xiazcy, line 63: Should anything inside the JS reference doc section be updated as well around this?

gremlin-js/gremlin-javascript/lib/driver/connection.ts
Cole-Greer: You're right, that path is missing. I've added it back in and added some tests which cover plain `application/json` error responses in both the submit and streaming paths.

docs/src/upgrade/release-4.x.x.asciidoc
Cole-Greer: Yes, I've gone ahead and updated the reference docs to explain the streaming api in JS.

xiazcy:
VOTE +1

kenhuuu:
VOTE +1

PR #3396, opened 2026-04-21
Streaming support for Python

GumpacG:
Adds streaming GraphBinary deserialization to the Python driver. Results are now deserialized directly from the HTTP response stream and pushed to the ResultSet individually, rather than buffering the entire response before processing.

Implements https://lists.apache.org/thread/qyxb845gy7fbhg87pmtcqs5zf0q33zm8

Changes:                                                                                                                                                                                                                                                              
  - Added AiohttpSyncStream wrapper and get_stream() to transport, replacing read()                                                     
  - Rewrote Connection._receive() to stream GB objects one-at-a-time via GraphBinaryReader.to_object()                                  
  - Updated ResultSet.one()/all() for individual item queue entries; added _EXHAUSTED sentinel to handle None as a valid Gremlin result 
  - Added content-type check before GB deserialization for non-GB error responses                
  - Removed dead code: buffered read(), data_received_aggregate, deserialize_message/read_payload, stream_chunk, max_content_length     
  - Removed GraphSON parameterized test fixtures                                          
  - Updated AbstractBaseTransport: read() → get_stream()                                                                            
  - 44 new unit tests covering streaming, bulking, error handling, content-type checks, and incremental delivery timing                 
  - Updated CHANGELOG and upgrade docs                           
  - Removed AbstractBaseTransport (gremlin_python.driver.transport)
  - Removed AbstractBaseProtocol and GremlinServerHTTPProtocol (gremlin_python.driver.protocol) 
  - Removed protocol_factory parameter from Client, DriverRemoteConnection, and Connection
  - Removed transport_factory parameter from Client, DriverRemoteConnection, and Connection 
  - GremlinServerError moved from gremlin_python.driver.protocol to gremlin_python.driver.connection
  - Connection constructor signature changed — protocol and transport_factory positional args removed; serializer, auth, and interceptor options are now keyword arguments


Breaking changes:
- ResultSet iteration now yields individual items instead of lists. Code using results += result must change to                       
  results.append(result).
- read() replaced with get_stream()
- Connection constructor signature changed, protocol and transport_factory positional args removed; serializer, auth, and interceptor options are now keyword arguments
- Custom transport implementations are no longer supported — the driver uses AiohttpHTTPTransport directly


docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -51,6 +51,34 @@ deserialization in GraphBinary is unchanged. Applications that depend on the `uu
 `gremlin-javascript` brought it in as a transitive dependency should add it directly to their own `package.json`
 if they still need it.
 
+==== Python HTTP Streaming Response Support
+
+The Python driver now streams GraphBinary results directly from the HTTP response body, matching the Go driver's
```
Cole-Greer: Nit: I don't think it makes sense to compare to the Go driver here, technically go got HTTP streaming first in the TP4 betas, but these upgrade docs should focus on the perspective of TP3.8.x users upgrading to TP4. We will arguably want to rewrite all of these sections once all drivers are done, to either group all streaming updates together, or to better organize sections per-GLV.

For the purposes of this PR, I think we should just ensure we capture all the right information from that TP3 upgrade perspective, and then a larger restructuring can follow if/when it makes sense.

kenhuuu:
Should the `transport.py` file be removed as well?

gremlin-python/src/main/python/gremlin_python/driver/connection.py
```diff
@@ -82,19 +119,58 @@ def cb(f):
...
+            flags = stream.read(1)[0]
+            bulked = flags == 0x01
+
+            # Deserialize results one at a time into the ResultSet
+            while True:
+                obj = reader.to_object(stream)
+                if obj == Marker.end_of_stream():
+                    break
+                if bulked:
+                    bulk = reader.to_object(stream)
```
xiazcy: I don't think we want to de-bulk here? All the other GLVs, and the old Python code all preserve the bulk count as a Traverser object and let the traversal iteration layer expand lazily. The current code would put all de-bulked objects into the queue, which kinda defeat the purpose of server-side bulking. This should prob be:
```
if bulked:
      bulk = reader.to_object(stream)
      self._result_set.stream.put_nowait(Traverser(obj, bulk))
else:
      self._result_set.stream.put_nowait(obj)
```

GumpacG:
> Should the `transport.py` file be removed as well?

@kenhuuu it was removed. Git just interprets it as a rename rather that a deletion for some reason. 
Screenshot 2026-05-04 at 2 26 09 PM

gremlin-python/src/main/python/gremlin_python/driver/connection.py
```diff
@@ -82,19 +119,58 @@ def cb(f):
...
+                    self._result_set.stream.put_nowait(obj)
+
+            # Read status after EndOfStream
+            status_code = int32_unpack(stream.read(4))
+            msg_is_null = stream.read(1)[0] == 0x01
+            status_message = '' if msg_is_null else reader.to_object(stream, DataType.string, False)
+            exc_is_null = stream.read(1)[0] == 0x01
+            status_exception = '' if exc_is_null else reader.to_object(stream, DataType.string, False)
+
+            if status_code not in (0, 200, 204):
```
xiazcy: Is there a case we get a status code of 0?

GumpacG: Good catch! No there isn't. Removed

gremlin-python/src/main/python/gremlin_python/driver/connection.py
GumpacG: That makes sense. Updated

xiazcy:
VOTE +1

kenhuuu:
VOTE +1

gremlin-python/src/main/python/gremlin_python/driver/aiohttp/transport.py
```diff
@@ -26,14 +24,27 @@
...
 else:
     import async_timeout
-from aiohttp import ClientPayloadError
-from gremlin_python.driver.protocol import GremlinServerError
-from gremlin_python.driver.transport import AbstractBaseTransport
 
 __author__ = 'Lyndon Bauto (lyndonb@bitquilltech.com)'
 
 
-class AiohttpHTTPTransport(AbstractBaseTransport):
```
Cole-Greer, line 36: Are we disallowing custom transports after this work as well? Or is that a separate task? We should remove `AbstractBaseTransport` if it is no longer relevant.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -138,6 +138,81 @@ g.V().has("blob",P.eq(Binary("AQID")))
...
+caller as they arrive from the server, rather than waiting for the full response to be buffered in memory. This improves
+memory efficiency for large result sets and reduces time-to-first-result.
+
+Traversal API terminal steps (`next()`, `to_list()`, `has_next()`) are now truly incremental. `next()` returns the
+first result as soon as it is deserialized from the wire, without waiting for the full response. In 3.x, `next()`
+waited for all data to be buffered before returning.
+
+The following internal APIs have changed:
+
+- The transport method `read()` has been replaced with `get_stream()`. Custom transport implementations must
```
Cole-Greer: This line doesn't fit well with the next section about removing custom transport implementations

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -138,6 +138,81 @@ g.V().has("blob",P.eq(Binary("AQID")))
...
+  update accordingly.
+- `max_content_length` is no longer supported. Response size limits should be configured server-side via
+  `maxRequestContentLength`.
+- The response path is GraphBinary only
```
Cole-Greer: This might need some expansion. Does this simply allude to us dropping our GraphSON serializers, or is it stating that users are completely prohibited from providing custom a response_serializer for their Client? My expectation is that the driver continues to support custom response serializers, even if we didn't put in the work to adapt our GraphSON serializers to the new streaming paradigm. If this is not correct, this may need further discussion.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -138,6 +138,81 @@ g.V().has("blob",P.eq(Binary("AQID")))
...
+- `ResultSet` now yields individual items instead of lists when iterated. Code that uses `results += result`
+  to accumulate results from a `ResultSet` iterator must change to `results.append(result)`. The `to_list()`
+  and `all()` APIs are unaffected and remain the recommended way to collect results.
+
+The `AbstractBaseTransport` abstract base class and the `transport_factory` extension point have been removed. Custom
```
Cole-Greer: This item probably fits better in the Python Driver Simplification section. It may be fully redundant with that section.

gremlin-python/src/main/python/gremlin_python/driver/serializer.py
```diff
@@ -32,7 +27,6 @@
     import json
 
 from gremlin_python.structure.io import graphbinaryV4, graphsonV4
```
Cole-Greer: Should we be deleting `gremlin_python.structure.io.graphsonV4` as well as the `GraphSONSerializersV4` class in this file since we are breaking their functionality and removing graphSON tests?

gremlin-python/src/main/python/gremlin_python/driver/connection.py
```diff
@@ -82,19 +120,57 @@ def cb(f):
...
+                        'message': body,
+                        'exception': ''
+                    })
+
+            # 204 No Content
+            if status == 204:
+                return
+
+            stream = self._transport.get_stream()
+            reader = GraphBinaryReader()
```
Cole-Greer: This class has `self._response_serializer`, but it isn't used in favour of a hardcoded `GraphBinaryReader`. This is somewhat required due to the following block which reads the GB header. I think we might need a bit more consideration of what the intended behaviour should be here.

gremlin-python/src/main/python/gremlin_python/driver/aiohttp/transport.py
GumpacG: It's been removed. See https://github.com/apache/tinkerpop/pull/3396#issuecomment-4374613532 as it's the same odd GitHub rendering.

docs/src/upgrade/release-4.x.x.asciidoc
GumpacG: Removed. Thanks!

docs/src/upgrade/release-4.x.x.asciidoc
GumpacG: Updated to match the old behaviour. Thanks!

docs/src/upgrade/release-4.x.x.asciidoc
GumpacG: Removed. Thanks!

gremlin-python/src/main/python/gremlin_python/driver/serializer.py
GumpacG: Removed GraphSON. Thanks!

gremlin-python/src/main/python/gremlin_python/driver/connection.py
GumpacG: Hardcoding has been removed and the old behaviour of supporting custom serializers have been restored. Thanks!

gremlin-python/src/main/python/gremlin_python/driver/aiohttp/transport.py
Cole-Greer: Ahh I missed that. Thanks!

Cole-Greer:
VOTE +1

PR #3397, opened 2026-04-23
Fix bulking behavior in gremlin-go for TP 4.0

xiazcy:
Fixed gremlin-go in TP 4.0 to properly handle bulked result with `bulkResults` request option. Bulk counts is now preserved inside Traverser objects, and lazily unroll during traversal iteration, matching Java/Python/.NET/JS behavior.

Also updated the relevant docs, and captured the fact that all GLVs will set "bulkResults" to true through DRC, overriding the server's default to not bulk. 

VOTE +1

### DIFF
--- a/gremlin-go/driver/client_test.go
+++ b/gremlin-go/driver/client_test.go
@@ -158,6 +158,37 @@ func TestClient(t *testing.T) {
 		AssertMarkoVertexWithoutProperties(t, result)
 	})
 
+	t.Run("Test client.SubmitWithOptions() with bulkResults", func(t *testing.T) {
+		skipTestsIfNotEnabled(t, integrationTestSuiteName, testNoAuthEnable)
+		client, err := NewClient(testNoAuthUrl,
+			func(settings *ClientSettings) {
+				settings.TlsConfig = testNoAuthTlsConfig
+			})
+		assert.NoError(t, err)
+		assert.NotNil(t, client)
+		defer client.Close()
+
+		resultSet, err := client.SubmitWithOptions("g.inject(1,2,3,2,1)",
+			new(RequestOptionsBuilder).SetBulkResults(true).Create())
+		assert.NoError(t, err)
+		assert.NotNil(t, resultSet)
+		results, err := resultSet.All()
+		assert.NoError(t, err)
+		// With bulkResults=true, the ResultSet contains Traverser objects (one per unique value).
+		// This is consistent with Java, Python, .NET, and JS which all expose Traverser objects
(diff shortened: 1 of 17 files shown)


gremlin-go/driver/client_test.go
```diff
@@ -158,6 +158,37 @@ func TestClient(t *testing.T) {
...
+		totalBulk := int64(0)
+		for _, r := range results {
+			tr, err := r.GetTraverser()
+			if err == nil {
+				totalBulk += tr.Bulk
+			} else {
+				totalBulk++
+			}
+		}
+		assert.Equal(t, int64(5), totalBulk)
```
GumpacG, line 188: Can you add a check for the amount of traversers as that would confirm that it is getting bulked properly.

Should there be a test for when `SetBulkResults(false)` where an error should be returned since there are no traversers?

GumpacG (commented):
VOTE +1

gremlin-go/driver/requestOptions.go
```diff
@@ -30,6 +32,7 @@ type RequestOptions struct {
 	userAgent             string
 	bindings              map[string]interface{}
 	materializeProperties string
+	bulkResults           string
```
kenhuuu: Why string not boolean?

gremlin-go/driver/traversal_test.go
```diff
@@ -542,6 +542,151 @@ func TestTraversal(t *testing.T) {
...
+		assert.Nil(t, err)
+		assert.True(t, ok)
+		assert.Equal(t, int32(99), val)
+
+		_, ok, err = trav.nextValue()
+		assert.Nil(t, err)
+		assert.False(t, ok)
+	})
+
+	t.Run("handles mixed Traverser and raw results", func(t *testing.T) {
```
kenhuuu: Is this test valuable? In what situation would this happen? The reason I'm asking this question is, if someone were to make a change that somehow broke this test, they would assume this is the appropriate behavior. But in my mind, results should either be bulked or not bulked, but not a mix of the two?

Cole-Greer:
Other than the existing comments from Guian and Ken, everything looks good to me.

gremlin-go/driver/requestOptions.go
xiazcy: No particular reason for either. This was to match types in Java's RequestOption.

gremlin-go/driver/traversal_test.go
xiazcy: Makes sense. I'll remove.

kenhuuu:
VOTE +1

PR #3398, opened 2026-04-23
Aligned AGENTS.md better with ASF AI contribution guidelines

spmallette:
Focus of this PR is more on AGENTS.md itself, but cleaned up some skill related items I noticed weren't working well. Dropped less used content that an agent likely wouldn't use. 

Inspiration for ASF alignment came from Groovy who are still discussing these changes on their dev list - but i think the component I added is the key one: 

https://github.com/apache/groovy/blob/master/AGENTS.md#licensing-and-provenance-read-first

Used attribution style from linux that I imagine went through much discussion and appears quite flexible: 

https://github.com/torvalds/linux/blob/master/Documentation/process/coding-assistants.rst#attribution

VOTE +1

### DIFF
--- a/.skills/tinkerpop-dev/references/build-java.md
+++ b/.skills/tinkerpop-dev/references/build-java.md
@@ -19,9 +19,18 @@ under the License.
 
 # Java / Core Module Builds
 
-## Building Specific Modules
+## Full Build (skip tests)
 
-Build a single module:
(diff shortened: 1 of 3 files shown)


Cole-Greer:
VOTE +1

xiazcy:
VOTE +1

PR #3400, opened 2026-04-24
Add getOldChar() and getNewChar() getters to ReplaceGlobalStep and ReplaceLocalStep

abethini-web:
ReplaceGlobalStep and ReplaceLocalStep currently expose oldChar and newChar only as constructor parameters with no public accessors. Graph database implementations that perform ahead-of-time query translation need to inspect these parameters during the translation phase to generate optimized execution plans — for example, to determine whether the replace operation can be pushed down to a native query engine or whether null arguments should be treated as a no-op at translation time. Without getters, translators have no way to access these values from the step object.

This PR adds getOldChar() and getNewChar() public getters to both steps. The change is non-breaking and follows the same pattern used by other parameterized steps such as SubstringGlobalStep and SplitGlobalStep.

TARGET: 3.8-dev — should be merged forward to master (4.0.0).


Cole-Greer:
Hi @abethini-web, thanks for submitting this PR. I don't see any issues with making this data available to strategies and such.

VOTE +1

kenhuuu:
VOTE +1.

xiazcy:
VOTE +1

PR #3401, opened 2026-04-27
TINKERPOP-3153 Add Char, Duration, Binary support to gremlin-lang.

kenhuuu:
https://issues.apache.org/jira/browse/TINKERPOP-3153

Duration is based on GraphBinary's seconds and nanos representation instead of Java Duration's ISO-8601 string. Despite ISO-8601 being more human-readable, the lack of support for it from the languages meant that lots of extra parsing had to be added to make it work. Also, this implementation doesn't preclude a future overload to Duration() from being added to the grammar.

A 's' suffix was added in addition to the 'c' suffix for users to explicitly choose a string over a character but string is already the default so it shouldn't cause any changes.

Support was only added to GLVs that already had support for Char, Duration, and Binary in GraphBinary.


gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/language/grammar/GeneralLiteralVisitorTest.java
```diff
@@ -238,6 +241,11 @@ public static Iterable<Object[]> generateTestParameters() {
                     {"'abc\\u2300def'", "abc\u2300def"},
                     {"'\u2300'", "\u2300"},
                     {"'abc\u2300def'", "abc\u2300def"},
+                    // explicit 's' suffix for string literals
+                    {"\"hello\"s", "hello"},
+                    {"'hello's", "hello"},
+                    {"\"\"s", "Empty"},
```
GumpacG, line 247: Shouldn't `"\"\"s"` be `""`?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/PythonTranslateVisitor.java
```diff
@@ -289,6 +289,42 @@ public Void visitUuidLiteral(final GremlinParser.UuidLiteralContext ctx) {
         return null;
     }
 
+    @Override
+    public Void visitCharacterLiteral(final GremlinParser.CharacterLiteralContext ctx) {
+        final String text = ctx.getText();
+        final String withoutSuffix = text.substring(0, text.length() - 1);
+        final String inner = removeFirstAndLastCharacters(withoutSuffix);
+        sb.append("SingleChar('").append(inner).append("')");
```
GumpacG, line 297: Does this work well with repr()? I'm wondering if repr() could potentially mess this up as repr() switches the wrapping quotes to `"` if the string is a `'`

gremlin-language/src/main/antlr4/Gremlin.g4
```diff
@@ -2500,6 +2520,26 @@ SignedInfLiteral
...
+CharacterLiteral
+    : '"' DoubleQuotedStringCharacter '"' 'c'
+    | '\'' SingleQuotedStringCharacter '\'' 'c'
+    ;
+
+// String literal with explicit 's' suffix. The suffix is optional (no suffix defaults to string),
+// but can be used to force string interpretation for single-character values.
+// Must appear before NonEmptyStringLiteral for longest-match priority.
+StringSuffixLiteral
+    : '"' DoubleQuotedStringCharacters '"' 's'
```
GumpacG, line 2534: Should this be for `DoubleQuotedStringCharacter` instead of `DoubleQuotedStringCharacters` as it's only for a single character?

gremlin-language/src/main/antlr4/Gremlin.g4
```diff
@@ -2500,6 +2520,26 @@ SignedInfLiteral
 
 // String Literals
 
+// Character literal is a single character string with a 'c' suffix, modeled after printf's %c.
+// Must appear before NonEmptyStringLiteral for longest-match priority.
+CharacterLiteral
+    : '"' DoubleQuotedStringCharacter '"' 'c'
```
GumpacG, line 2526: Any thoughts about switching the grammar to `CHARACTER("a")` instead of `"a"c`? We have `datetime()`, `uuid()`, etc. to specify types. Would it be worth following this format?

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -51,6 +51,65 @@ deserialization in GraphBinary is unchanged. Applications that depend on the `uu
...
+===== Character
+
+A character literal is a single-character string with a `c` suffix:
+
+[source,text]
+----
+g.V().has("initial","M"c)
+----
+
+NOTE: Character is not supported in gremlin-go or gremlin-javascript.
```
Cole-Greer, line 96: Can we map Gremlin characters to the builtin `Rune` type in Go? 

https://go.dev/blog/strings#code-points-characters-and-runes

gremlin-language/src/main/antlr4/Gremlin.g4
```diff
@@ -2500,6 +2520,26 @@ SignedInfLiteral
...
+    | '\'' SingleQuotedStringCharacter '\'' 'c'
```
Cole-Greer, line 2527: Should we also allow `C` as a suffix (and `s` and `S` for Strings). That would be more consistent with `FloatTypeSuffix` and `IntegerTypeSuffix`

docs/src/dev/developer/for-committers.asciidoc
```diff
@@ -385,7 +385,12 @@ appropriate types expected by the tests.
 The syntax of the type notation involves a prefix character to help denote the type, a value between two square
 brackets, optionally suffixed with some additional notation depending on the primary type.
 
+* Binary - *bin[_xxx_]* - The "xxx" should be replaced with a base64-encoded string.
+* Character - *char[_xxx_]* - The "xxx" should be replaced with a single UTF-8 character.
 * Date - *dt[_xxx_]* - The "xxx" should be ISO 8601 string.
+* Duration - *dur[_xxx_,_yyy_,_zzz_]* - The "xxx" should be replaced a Long representing seconds, the "yyy" should be
+replaced an Int representing nanoseconds, and the optional "zzz" should be replaced by a boolean which is true if the
+value is positive and false if its negative.
 * Edge - *e[_xxx_]* - The "xxx" should be replaced with a representation of an edge in the form of the
```
Cole-Greer, line 394: Can we also document the new `@DataChar`, `@DataDuration`, and `@DataBinary` tags in the below `[[gherkin-tags]]` section?

Cole-Greer (commented):
Overall looks good to me. My main concern is support for Runes in Go.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/language/grammar/GeneralLiteralVisitorTest.java
kenhuuu: Yes, but its part of the test definition.
```
if (expected.equals("Empty")) {
    // handle special case for Empty string
    assertEquals("", new GenericLiteralVisitor(new GremlinAntlrToJava()).visitStringLiteral(ctx));
```

gremlin-language/src/main/antlr4/Gremlin.g4
kenhuuu: This is the newly added "s" suffix that is for strings which can be multicharacter.

gremlin-language/src/main/antlr4/Gremlin.g4
kenhuuu: This is actually to follow the integer literal format. The other "constructor" type formats are used for classes.

docs/src/upgrade/release-4.x.x.asciidoc
kenhuuu: I think so. I just didn't do this because there is no support for Character in gremlin-go's GraphBinary implementation. I think this is a separate Jira for character support in gremlin-go.

gremlin-language/src/main/antlr4/Gremlin.g4
kenhuuu: Maybe? That's probably a good idea. The way I have it set up right now follows printf where c/C and s/S are different things where wide-characters are a thing. But I guess since we are UTF-8 only, it does make sense to have it more consistent within the grammar. Let me think about this some more.

docs/src/dev/developer/for-committers.asciidoc
kenhuuu: We didn't seem to do that for any of the other `@Data...` test files. Any particular reason you think it would be valuable to do this here? I'm assuming that at some point these data types will be used in other feature files so having these tags mentioned would be kind of pointless. You would probably want a `@SupportsChar` or something later?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/PythonTranslateVisitor.java
kenhuuu: Yes this works. But you are actually pointing out an inconsistency that I didn't mention. Java/.NET/Go output the value wrapped in "" by default, but JavaScript and Python wrap in ''. So there is a bit of an inconsistency there. The grammar, however, accepts both, so even when repr() swaps from ' to ", it still works.

GumpacG:
LGTM. VOTE +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/GoTranslateVisitor.java
```diff
@@ -346,6 +347,35 @@ public Void visitUuidLiteral(final GremlinParser.UuidLiteralContext ctx) {
...
+        throw new TranslatorException("Character literals are not supported in Go");
+    }
+
+    @Override
+    public Void visitDurationLiteral(final GremlinParser.DurationLiteralContext ctx) {
+        final long seconds = Long.parseLong(ctx.integerLiteral(0).getText());
+        final int nanos = Integer.parseInt(ctx.integerLiteral(1).getText());
+        final boolean isPositive = ctx.booleanLiteral() == null ||
+                Boolean.parseBoolean(ctx.booleanLiteral().getText());
+        final long totalNanos = seconds * 1_000_000_000L + nanos;
```
xiazcy, line 361: Looks like you used BigInt in the JS translator for Go, i.e.:

> // Use BigInt to avoid precision loss for durations over ~104 days
>         const totalNanos = BigInt(seconds) * 1_000_000_000n + BigInt(nanos);

Should the Java translator here also use BigInt for the same reason?

xiazcy:

docs/src/upgrade/release-4.x.x.asciidoc
Cole-Greer: I'm happy to defer Rune support as long as it's tracked somewhere.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/GoTranslateVisitor.java
kenhuuu: This shouldn't be necessary as Long is sufficient for Java. The reason this is needed for TypeScript is because numbers are stored as floats so there is a `Number.MAX_SAFE_INTEGER` which will likely get exceeded in this multiplication. That makes BigInt safer for the TypeScript version.

docs/src/upgrade/release-4.x.x.asciidoc
kenhuuu: https://issues.apache.org/jira/browse/TINKERPOP-3248

gremlin-language/src/main/antlr4/Gremlin.g4
kenhuuu: I'll make a Jira for this after merging.

docs/src/dev/developer/for-committers.asciidoc
kenhuuu: Added a `@SupportChar`,  `@SupportDuration`,  `@SupportBinary` to the feature files and this documentation

kenhuuu:
VOTE +1

PR #3402, opened 2026-04-27
TINKERPOP-3247 Convert request bindings to gremlin-lang string format

kenhuuu:
https://issues.apache.org/jira/browse/TINKERPOP-3247

Moving parameters from binary-serialized maps to string representations
makes the request side pure text, decoupling Gremlin language evolution
from GraphBinary versioning. New types can be introduced in minor/patch
versions without touching GraphBinary, eliminating the need for a major
version bump across the ecosystem for every new request-side type.

The asParameter() fallback is replaced with an unsupportedType flag that
records the class name and falls back to toString(). A flag is used
rather than throwing because embedded Traversals build GremlinLang as a
side effect but never send it, so unknown types must not break
execution. All other GLVs throw immediately since they have no embedded
mode and the early throw gives better errors.

Client APIs accept both map and string bindings (but not both at the
same time) because users who use the Client directly with raw Gremlin
strings shouldn't need to hand-craft gremlin-lang map literals. Mixing
both throws immediately to prevent silent loss where one set of bindings
would be discarded.

Edge and VertexProperty tests that relied on the old asParameter
fallback were removed because they aren't supported in gremlin-lang.

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/GremlinQueryParser.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/GremlinQueryParser.java
@@ -18,13 +18,16 @@
  */
 package org.apache.tinkerpop.gremlin.language.grammar;
 
-import org.antlr.v4.runtime.CharStream;
 import org.antlr.v4.runtime.CharStreams;
 import org.antlr.v4.runtime.CommonTokenStream;
 import org.antlr.v4.runtime.atn.PredictionMode;
 import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.util.Collection;
+import java.util.Map;
+import javax.lang.model.SourceVersion;
 
 /**
  * Parses Gremlin strings to an {@code Object}, typically to a {@link Traversal}.
@@ -44,21 +47,15 @@ public class GremlinQueryParser {
      * Parse Gremlin string using a specified {@link GremlinAntlrToJava} object.
      */
     public static Object parse(final String query, final GremlinVisitor<Object> visitor)  {
-        final CharStream in = CharStreams.fromString(query);
-        final GremlinLexer lexer = new GremlinLexer(in);
-        lexer.removeErrorListeners();
-        lexer.addErrorListener(errorListener);
-
+        final GremlinLexer lexer = createLexer(query);
         final CommonTokenStream tokens = new CommonTokenStream(lexer);
 
         // Setup error handler on parser
-        final GremlinParser parser = new GremlinParser(tokens);
+        final GremlinParser parser = createParser(tokens);
         // SLL prediction mode is faster than the LL prediction mode when parsing the grammar,
         // but it does not cover parsing all types of input.  We use the SLL by default, and fallback
         // to LL mode if fails to parse the query.
         parser.getInterpreter().setPredictionMode(PredictionMode.SLL);
-        parser.removeErrorListeners();
-        parser.addErrorListener(errorListener);
 
         GremlinParser.QueryListContext queryContext;
         try {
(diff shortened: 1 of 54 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/GremlinQueryParser.java
```diff
@@ -91,4 +88,94 @@ public static Object parse(final String query, final GremlinVisitor<Object> visi
...
+    public static Map<String, Object> parseParameters(final String parameterMapString) {
+        if (parameterMapString == null || parameterMapString.isEmpty()) {
+            return Map.of();
+        }
+
+        final GremlinParser parser = createParser(parameterMapString);
+        final GremlinParser.GenericMapLiteralContext mapCtx = parser.genericMapLiteral();
+
+        final ParameterMapVisitor visitor = new ParameterMapVisitor(new GremlinAntlrToJava());
+        final Map<Object, Object> rawMap = (Map<Object, Object>) visitor.visitGenericMapLiteral(mapCtx);
```
Cole-Greer, line 111: Do we need some error handling here? Am I right in assuming `mapCtx` will be null if the `parameterMapString` is not actually a gremlin-lang map? How would such an error propagate if the user provides a bad parameter string?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/GremlinQueryParser.java
```diff
@@ -91,4 +88,94 @@ public static Object parse(final String query, final GremlinVisitor<Object> visi
...
+    /**
+     * Recursively validates that a parameter value does not contain a {@link Traversal}. Nested validation is needed
+     * because steps like mergeV iterate map values, so a Traversal hiding inside a nested map or collection would still
+     * be dangerous.
+     */
+    private static void validateParameterValue(final Object value) {
+        if (value instanceof Traversal) {
+            throw new GremlinParserException("Traversals are not allowed as parameter values");
+        }
+        if (value instanceof Map) {
```
Cole-Greer, line 143: Should we also recurse through map keys?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/GremlinQueryParser.java
```diff
@@ -91,4 +88,94 @@ public static Object parse(final String query, final GremlinVisitor<Object> visi
...
+    }
+
+    /**
+     * Recursively validates that a parameter value does not contain a {@link Traversal}. Nested validation is needed
+     * because steps like mergeV iterate map values, so a Traversal hiding inside a nested map or collection would still
+     * be dangerous.
+     */
+    private static void validateParameterValue(final Object value) {
+        if (value instanceof Traversal) {
+            throw new GremlinParserException("Traversals are not allowed as parameter values");
```
Cole-Greer, line 141: I think this should be kept out of scope from this work, but I'm curious if there may be demand to parameterize something like `CardinalityValueTraversal` in the future. That's a funky case which generally acts as a static value but is technically a Traversal.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/ParameterMapVisitor.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Override
+    public Object visitGenericCollectionLiteral(final GremlinParser.GenericCollectionLiteralContext ctx) {
+        currentNestingDepth++;
+        if (currentNestingDepth > maxNestingDepth) {
+            throw new GremlinParserException("Parameter map nesting depth exceeds maximum of " + maxNestingDepth);
+        }
+        try {
+            return super.visitGenericCollectionLiteral(ctx);
```
Cole-Greer, line 84: Am I right in understanding that this will result in all objects contained in the collection to be parsed by `GenericLiteralVisitor.visitGenericLiteral()` instead of `ParameterMapVisitor.visitGenericLiteral()`? Do we need more careful handling of collections and any composite types to ensure that they are recursively parsed through this class and not handed off to the unguarded `GenericLiteralVisitor`?

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/language/grammar/ParameterMapVisitorTest.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    @Test(expected = GremlinParserException.class)
+    public void shouldRejectNestedTraversalInValue() {
+        GremlinQueryParser.parseParameters("[\"x\":__.out()]");
+    }
+
+    @Test(expected = GremlinParserException.class)
+    public void shouldRejectTerminatedTraversalInValue() {
+        GremlinQueryParser.parseParameters("[\"x\":g.V().drop().iterate()]");
```
Cole-Greer, line 176: I think it might be worth a few more cases which try to bury a terminatedTraversal deeper inside other collections/composite types.

gremlin-js/gremlin-javascript/lib/driver/request-message.ts
```diff
@@ -128,16 +131,24 @@ export class Builder {
...
   }
 
   addBindings(otherBindings: object): Builder {
     if (!otherBindings) throw new Error('bindings argument cannot be null.');
+    if (this.bindingsString) throw new Error('Cannot mix addBindings() with addBindingsString().');
     Object.assign(this.bindings, otherBindings)
     return this;
   }
 
+  addBindingsString(bindingsString: string): Builder {
```
Cole-Greer, line 161: Nit: `addBindings` acts as a merge, but `addBindingsString` is a replacement. I think that's ok to an extent, (we definitely don't want to try merging strings), but might be worth a quick jsdoc. Perhaps it would be also reasonable to only allow `addBindingsString` to be called once.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/GremlinQueryParser.java
kenhuuu: Error handling is done via the GremlinErrorListener that throws a GremlinParserException if its not a map so there won't be a null. See the `shouldReturnUserFriendlyErrorMessageForMalformedParameterStrings` test for more info.

gremlin-js/gremlin-javascript/lib/driver/request-message.ts
kenhuuu: Yea I think that should get commented so I'll update that. Regarding allowing it to only be called once, my opinion is that its not necessary for now. The documentation will say "last one wins".

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/GremlinQueryParser.java
kenhuuu: This is just extra checking so neither technically needs to be exist, but yes, I'll add it for completeness.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/GremlinQueryParser.java
kenhuuu: That's a good point and we might have to relax some of these rules in the future.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/ParameterMapVisitor.java
kenhuuu: Yes, thanks for pointing this out, something is definitely not right here. I modified the test in ParameterMapVisitorTest slightly to assert the exception message and it confirms that it isn't dispatching to the right visitor. I'll have to think about how to make a simple change to fix this.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/language/grammar/ParameterMapVisitorTest.java
kenhuuu: Agreed, will add.

xiazcy:
VOTE +1 pending resolution of existing comments

kenhuuu:
VOTE +1

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerIntegrateTest.java
```diff
@@ -1155,4 +1130,56 @@ public void ensureScriptEngineDefaultsToGremlinLang() {
...
+                {"[\"x\":,\"y\":1]",            "could not be converted into a Map. Query parsing failed at"},
+                {"[\"x\":__.V().drop()]",       "Traversals are not allowed"},
+                {"[\"~id\":1]",                 "must be a valid identifier"}
+        };
+
+        for (final String[] testCase : cases) {
+            final ResultSet result = client.submit(
+                    "g.V(x)", RequestOptions.build().addParametersString(testCase[0]).create());
+            try {
+                result.one();
```
Cole-Greer, line 1174: Nit:
```suggestion
                result.one();
                fail("Expect to throw exception")
```

Cole-Greer:
VOTE +1

PR #3418, opened 2026-05-03
Replace AWK/shell docs preprocessing with AsciidoctorJ extension

Cole-Greer:
### Overview

This PR introduces the gremlin-docs module — a custom AsciidoctorJ TreeProcessor extension that replaces the entire AWK/shell preprocessing pipeline (10 AWK scripts, 5 shell scripts) used to build TinkerPop documentation.

The new system delegates Gremlin code execution to a real Gremlin Console process via stdin/stdout, then generates language variant tabs using the ANTLR-based GremlinTranslator. This provides the full console environment (correct result formatting, Sugar plugin, SPARQL, Neo4j, remote connections) while 
eliminating the AWK pipeline and external Hadoop dependency.

This PR is currently in draft state to collect early feedback. There are several known issues which need to be resolved, and I suspect more issues may be found upon further inspections of the docs.

### Architecture

- **GremlinTreeProcessor** — AsciidoctorJ TreeProcessor that walks the document AST, finds [gremlin-groovy] blocks, orchestrates execution and translation, and builds tabbed HTML output
- **ConsoleExecutor** — Launches bin/gremlin.sh as a long-running subprocess, communicates via stdin/stdout with per-statement sentinel boundaries. Handles the console's "Display stack trace?" error prompt robustly via double-sentinel protocol
- **VariantTranslator** — Wraps GremlinTranslator to produce Java, Python, JavaScript, C#, and Go tabs for every translatable Gremlin example
- **process-docs.sh** — Build entry point that installs console plugins, starts Gremlin Server (for remote examples) and Gephi mock (for visualization examples), and passes configuration to the AsciidoctorJ extension

### What it does

- **Executes [gremlin-groovy] code blocks** via the real Gremlin Console, producing live console output with correct gremlin> prompts, ==> results, and proper datatype formatting (Path, Map, etc.)
- **Auto-generates language variant tabs** (Java, Python, JavaScript, C#, Go) for every translatable Gremlin example using the ANTLR-based GremlinTranslator
- **Supports standalone tab groups** for manually-authored [source,LANG,tab] blocks
- **Syntax highlighting via Rouge** (server-side, build-time) — replaces both CodeRay (no C# support) and highlight.js (destroyed callout markers)
- **Callout numbers render correctly** as black circled numbers in code blocks

### How to use

bash
# Prerequisites: build console and server distributions
mvn clean install -pl :gremlin-server,:gremlin-console -am -DskipTests

# Full build with live gremlin execution
bin/process-docs.sh

# Dry-run: skip execution (fast, for layout checks — no console/server needed)
bin/process-docs.sh --dry-run

### Changes

- New gremlin-docs/ module (not in the Maven reactor — built separately by the build script)
- New ConsoleExecutor.java — per-statement stdin/stdout protocol with the Gremlin Console
- New VariantTranslator.java — ANTLR-based translation to all GLV languages
- Updated bin/process-docs.sh — builds extension, installs plugins, starts server/Gephi mock
- Restored bin/gephi-mock.py — simple HTTP mock for Gephi visualization examples
- Root pom.xml: switched syntax highlighter to Rouge, added SnakeYAML 1.33 override for Rouge/Psych compatibility
- Removed old AWK/shell pipeline (10 AWK scripts, 5 shell scripts, gephi-mock)
- Updated development-environment.asciidoc to document new build requirements

### Known issues

Neo4j examples fall back to dry-run output. The neo4j-gremlin plugin is excluded from the default plugin list because its Spark jars conflict with spark-gremlin on the classpath. The old system handled this by swapping plugins per-document file; the new system processes the entire reference book as one 
document, making per-file swapping impractical. Neo4j is deprecated in TinkerPop.

A few Hadoop hdfs.head() calls may fail. The hdfs.head('output/~g') and hdfs.head('output', GryoInputFormat) calls may produce different output in local-mode Spark vs a real cluster. This was also an issue in the old system.

### Not changed

- All AsciiDoc source files are unchanged (no doc content modifications)
- Document structure, sections, anchors, and static content are preserved


gremlin-docs/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
```
spmallette, line 21: If this is a standalone utility/tool, should it use TinkerPop as a parent? Could we tuck gremlin-docs away in the `docs/` directory? Perhaps even better would be to call it "tinkerpop-docs" to not confuse it with our standard "gremlin" prefixing/suffixing? 

bigger thinking - i think there's going to be more and more of these types of little tools and things for various bits of automation and such. not sure where those things should live. i think `docs/upgrade` was one of those to start it. maybe the examples too. they all sorta build outside of the primary maven reactor. Do we need something else that ties these together better? almost need a top-level command system of sorts to keep all these tools and scripts unified somehow.

Cole-Greer:
Superseded by https://github.com/apache/tinkerpop/pull/3455

PR #3420, opened 2026-05-05
Supported python version matrix for CI/CD

GumpacG:
## Summary
Run gremlin-python tests across all [supported Python versions](https://tinkerpop.apache.org/docs/current/reference/#Gremlin-Python) in parallel

Converts the single-version python job in build-test.yml into a fail-fast: false matrix covering Python 3.10-3.14, and
parameterizes gremlin-python/docker-compose.yml (image: python:${PYTHON_VERSION:-3.10}) so the matrix value actually reaches theinterpreter inside Docker. Default behavior for local mvn verify -pl gremlin-python is unchanged. Per-version JUnit reports areuploaded as artifacts.

## Design
The python build-test for multiple versions are running in parallel and should take around the same time to complete (~25 min.) as before this change.
Screenshot 2026-05-05 at 9 30 23 AM

## Follow up tasks

- This is a CI/CD change only and should be backported to 3.8-dev and 3.7-dev. Python 3.9 should be added to the matrix for 3.7-dev. 
- Other GLVs should adopt this change as well for documented supported versions

 
## TODO
There is an error from testing regarding `asyncio.get_event_loop` for Python 3.14 that is in progress.

## Note

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -301,13 +301,6 @@ jobs:
           EXCLUDE="${EXCLUDE_MODULES/,-:gremlin-python},$EXCLUDE_FOR_GLV"
           mvn clean install -pl $EXCLUDE -q -DskipTests -Dci
           mvn verify -pl gremlin-python
-      - name: Upload test reports
-        if: always()
-        uses: actions/upload-artifact@v4
-        with:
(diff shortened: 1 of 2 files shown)


GumpacG:
Closing as https://github.com/apache/tinkerpop/pull/3421 is merging into 3.7-dev

PR #3422, opened 2026-05-12
Add byte buffer to improve gremlin-python performance

GumpacG:
### Description
The streaming-support change in #3396 regressed the python CI job from ~5-6 min to 10-19 min and started hitting the 20-min timeout. The root cause is that AiohttpSyncStream.read(n) was running a full asyncio event-loop turn (loop.run_until_complete(...)) for every byte-level read the GraphBinary deserializer performs. A typical response of N results triggers at minimum 3N + 8 reads - for a 5,000-result response that's ~500 ms of pure event-loop overhead vs ~0.5 ms before streaming.

This PR keeps the streaming design intent but pays the asyncio cost at chunk granularity (~64 KB) instead of per byte.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -32,7 +32,6 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Removed `readable-stream` dependency from `gremlin-javascript`.
 * Added `Character`, `Binary`, `Duration` to the `gremlin-lang` grammar, GLVs, and Translators.
 * Refactored `gremlin-python` receive path to stream results directly from the HTTP response through the configured `response_serializer`.
-* Fixed slow GraphBinary deserialization in `gremlin-python` by chunk-buffering the HTTP response in `AiohttpSyncStream` instead of running the asyncio event loop on every byte-level read.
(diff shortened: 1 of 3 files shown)


CHANGELOG.asciidoc
```diff
@@ -32,6 +32,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Removed `readable-stream` dependency from `gremlin-javascript`.
 * Added `Character`, `Binary`, `Duration` to the `gremlin-lang` grammar, GLVs, and Translators.
 * Refactored `gremlin-python` receive path to stream results directly from the HTTP response through the configured `response_serializer`.
+* Fixed slow GraphBinary deserialization in `gremlin-python` by chunk-buffering the HTTP response in `AiohttpSyncStream` instead of running the asyncio event loop on every byte-level read.
```
Cole-Greer: Nit: This probably doesn't need a changelog entry as it's just augmenting the yet unreleased python streaming change.

Cole-Greer:
VOTE +1

kenhuuu:
VOTE +1

xiazcy:
VOTE +1

PR #3423, opened 2026-05-13
Enable autocommit in GremlinServer

kenhuuu:
This behaves the same as the TraversalOpProcessor would have in the 3.x line. All traversals are now transactional if the underlying Graph supports transactions. Traversals that aren't explicitly in a transaction are now wrapped into their own implicit transaction and the server will autocommit on sucess and rollback on failure.

Assisted-by: Kiro:claude-opus-4-6

### DIFF
--- a/docs/src/dev/provider/index.asciidoc
+++ b/docs/src/dev/provider/index.asciidoc
@@ -1260,25 +1260,32 @@ Graph transactions are typically `ThreadLocal`-bound. The server maintains a sin
 to ensure all operations execute on the same thread. This is an important implementation detail for graph system
 providers whose `Transaction` implementation relies on thread-local state.
 
-==== Auto-Commit for Non-Transactional Requests
+==== Implicit Transaction Management
 
-Non-transactional requests (those without a `transactionId`) are not affected by any of the transaction-specific
-behavior described above. However, if the underlying graph supports transactions, the server must automatically manage
-transactions for these requests:
+Implicit transactions (requests without a `transactionId`) are not affected by any of the explicit transaction-specific
+behavior described above. All graphs participate in implicit transactions regardless of whether they support explicit
+transactions. The difference lies in what the server does at the boundaries:
+
+When the graph supports explicit transactions, the server auto-commits and rolls back implicit transactions:
 
 * Before processing: roll back any stale open transaction to prevent state leakage between requests.
 * On success: commit the transaction after the traversal has been fully iterated and serialized.
 * On error: roll back the transaction so that partial mutations are not persisted.
 
-This auto-commit behavior ensures that users who do not use explicit transactions still get durable writes on success
-and clean rollback on failure. Graph system providers implementing their own server or HTTP endpoint must replicate
-this behavior if their graph implementation supports transactions. The commit occurs after serialization is complete
-but before the final response is flushed to the client, so the client only receives a success response after the
-data is committed.
+When the graph does not support explicit transactions, writes are immediately durable. If a failure occurs
+mid-traversal, the graph may be left in a partially mutated state since rollback is not possible.
 
-IMPORTANT: This auto-commit/rollback logic must only apply to non-transactional requests. Requests that carry a
-`transactionId` are part of a client-managed transaction and must not be auto-committed or auto-rolled-back by the
-server.
+Graph system providers implementing their own server or HTTP endpoint should replicate the auto-commit/rollback behavior
+when their graph implementation supports explicit transactions. The commit occurs after serialization is complete but
+before the final response is flushed to the client, so the client only receives a success response after the data is
+committed. Note that because HTTP responses use chunked transfer encoding, the initial HTTP status is sent as 200 OK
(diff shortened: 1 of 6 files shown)


docs/src/dev/provider/index.asciidoc
```diff
@@ -1260,8 +1260,25 @@ Graph transactions are typically `ThreadLocal`-bound. The server maintains a sin
...
+Non-transactional requests (those without a `transactionId`) are not affected by any of the transaction-specific
+behavior described above. However, if the underlying graph supports transactions, the server must automatically manage
+transactions for these requests:
+
+* Before processing: roll back any stale open transaction to prevent state leakage between requests.
+* On success: commit the transaction after the traversal has been fully iterated and serialized.
+* On error: roll back the transaction so that partial mutations are not persisted.
+
+This auto-commit behavior ensures that users who do not use explicit transactions still get durable writes on success
+and clean rollback on failure. Graph system providers implementing their own server or HTTP endpoint must replicate
```
GumpacG: Since we are using "must", is there a consequence if they don't (Exception)? Can we document that?

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerHttpIntegrateTest.java
```diff
@@ -489,36 +491,97 @@ public void should200OnPOSTWithGremlinJsonEndcodedBodyForJavaTime() throws Excep
...
+
+        try (final CloseableHttpResponse response = httpclient.execute(countPost2)) {
+            assertEquals(200, response.getStatusLine().getStatusCode());
+            final String json = EntityUtils.toString(response.getEntity());
+            final JsonNode node = mapper.readTree(json);
+            assertEquals(1, node.get("result").get(TOKEN_DATA)
+                    .get(GraphSONTokens.VALUEPROP).get(0)
+                    .get(GraphSONTokens.VALUEPROP).intValue());
+        }
+    }
```
GumpacG, line 584: Is there also a way to test stale open transactions?

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpGremlinEndpointHandler.java
GumpacG: Could any of these lines throw an error and should be in the rollback try catch in this case?

Cole-Greer:
VOTE +1

docs/src/reference/gremlin-applications.asciidoc
```diff
@@ -2242,10 +2242,14 @@ above with the use of the `maximumSize`.
 [[considering-transactions]]
 ==== Considering Transactions
 
-Non-transactional requests (those without a `transactionId`) behave as self-contained units of work where the graph's
-own transaction semantics apply. Each traversal executes within its own transaction as managed by the graph
-implementation itself. Transactional requests participate in a transaction opened via `g.tx().begin()`, where the
-client explicitly controls the lifecycle through `g.tx().commit()` and `g.tx().rollback()`.
+Non-transactional requests (those without a `transactionId`) behave as self-contained units of work. If the underlying
```
spmallette: In the provider docs we wrote about "graphs that don't support transactions" and here we talk about "non-transactional". I wonder if we should avoid that language and just better introduce the notion of implicit and explicit transactions and talk only in those terms. In this way, all graphs have a transaction and they all support implicit (i.e. auto-commit, sessionless, non-transactional, etc. in the old parlance) and may support explicit (i.e. manual commit, session, transactional, etc. in the old parlance).  In this way we don't drag back any of the older terminology.

xiazcy:
VOTE +1 pending comment resolution.

docs/src/dev/provider/index.asciidoc
kenhuuu: I'll change the wording to `should`, but no there are no consequences as to the user it would just look like it worked even though it failed.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinServerHttpIntegrateTest.java
kenhuuu: This would be hard to do as you can't make this happen in the normal flow since implicit transactions always attempt to rollback or commit and explicit transactions happen on a separate pool completely. We'd need some sort of mock server that has this flaw built into it so I'm going to leave this out for now if that is ok with you.

gremlin-server/src/main/java/org/apache/tinkerpop/gremlin/server/handler/HttpGremlinEndpointHandler.java
kenhuuu: This is a good point. I'll move the try block up to include the script eval in case gremlin-groovy is used.

GumpacG:
VOTE +1

kenhuuu:
Merging as lazy consensus has been reached.

PR #3427, opened 2026-05-15
Add typed numeric wrappers and precise number mode to gremlin-javascript

kirill-stepanishin:
## Summary

JavaScript's single `Number` type (IEEE 754 double) loses the distinction between Gremlin numeric types like `int`, `float`, `long`, and `double`. This PR gives `gremlin-javascript` users explicit control over numeric type fidelity on both sides of the wire.

**Serialization** — New wrapper classes and factory functions (`toInt`, `toFloat`, `toDouble`, `toLong`, `toShort`, `toByte`) control the exact GraphBinary type code and GremlinLang suffix sent to the server. Without wrappers, existing type-inference behavior is unchanged.

**Deserialization** — A new `numberMode: 'precise'` connection option wraps incoming numeric values in the same typed wrappers, preserving the server's original type information. Wrappers support transparent arithmetic via `valueOf`/`Symbol.toPrimitive`, and an `unwrap()` helper extracts raw values.

**Implementation** — Refactored `GraphBinary.js` IoC initialization into a `createIoc()` factory so precise mode can inject a `postDeserialize` hook via `AnySerializer`. Extended `NumberSerializationStrategy` and `GremlinLang` to route wrapper instances to the correct serializers/suffixes. Enhanced `Long` to accept `bigint` with full int64 range validation.

## Tests

- Unit: type-code routing, byte-level verification, round-trip serialize/deserialize for all six types, precise-mode coverage (nulls, nested structures, non-numeric passthrough), GremlinLang emission for all wrappers including edge cases (NaN, ±Infinity, boundary values, predicates, collections)
- Integration: precise mode against a live server verifying `Int`/`Double` wrappers on vertex properties and edge weights
- Export verification for all new public API symbols

## Documentation

- Reference docs: new "Numeric Types" section, updated connection options table and limitations
- Upgrade guide: new "JavaScript Typed Numeric Wrappers" section
- CHANGELOG entry

### DIFF
--- a/docs/src/reference/gremlin-variants.asciidoc
+++ b/docs/src/reference/gremlin-variants.asciidoc
@@ -1677,7 +1677,7 @@ can be passed in the constructor of a new `Client` or `DriverRemoteConnection` :
 |options.traversalSource |String |The traversal source. |'g'
 |options.headers |Object |Additional HTTP header key/values included with each request. |undefined
 |options.interceptors |RequestInterceptor/RequestInterceptor[] |One or more functions that can modify the HTTP request before it is sent. |undefined
-|options.numberMode |String |Set to `'precise'` to wrap deserialized numbers in typed wrappers that preserve the server's original type. |undefined
+|options.preciseNumbers |Boolean |When `true`, wraps deserialized numbers in typed wrappers that preserve the server's original type. |undefined
 |options.reader |GraphBinaryReader |The reader to use for deserializing responses. |GraphBinaryReader
 |options.writer |GraphBinaryWriter |The writer to use for serializing requests. |GraphBinaryWriter
 |options.enableUserAgentOnConnect |Boolean |Determines if a user agent header will be sent with requests. |true
@@ -1979,10 +1979,8 @@ the possibility of `null`.
 [[gremlin-javascript-numeric-types]]
 === Numeric Types
 
-JavaScript has a single `Number` type (IEEE 754 double) which cannot distinguish between JVM numeric types. The driver
+JavaScript has a single `Number` type (IEEE 754 double) which cannot distinguish between Gremlin numeric types. The driver
 provides typed wrapper classes and factory functions that give explicit control over serialization and deserialization.
-
-==== Controlling Serialization
 
 Wrapping a value selects the GremlinLang type suffix and GraphBinary type code sent to the server. Without wrappers the
 driver infers types automatically, so existing code is unaffected.
@@ -1997,25 +1995,21 @@ g.V().has('id', toLong('9007199254740993')).next();
 ----
 
 `toLong()` accepts `number`, `string`, or `bigint`. String and bigint inputs support the full signed 64-bit range;
-number inputs must be within the safe integer range.
-
-==== Preserving Types on Deserialization
+number inputs must be within the safe integer range (`RangeError` is thrown otherwise).
 
 By default, the driver deserializes all numeric values as plain `Number` (or `BigInt` for large longs). To preserve the
-server's original type, set `numberMode` to `'precise'` on the connection:
+server's original type, set `preciseNumbers` to `true` on the connection:
 
 [source,javascript]
 ----
 const g = traversal().with_(new DriverRemoteConnection('http://localhost:8182/gremlin', {
-  numberMode: 'precise'
+  preciseNumbers: true
 }));
 
 const v = await g.V(1).elementMap().next();
 const age = v.value.get('age');  // Int { value: 29, type: 'int' }
 age + 1;                         // 30 — wrappers support arithmetic via valueOf()
 ----
-
-==== Unwrapping
 
 The `unwrap()` helper extracts the raw value from any wrapper, passing non-wrapper values through unchanged:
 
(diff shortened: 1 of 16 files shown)


docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1671,6 +1671,7 @@ can be passed in the constructor of a new `Client` or `DriverRemoteConnection` :
 |options.traversalSource |String |The traversal source. |'g'
 |options.headers |Object |Additional HTTP header key/values included with each request. |undefined
 |options.interceptors |RequestInterceptor/RequestInterceptor[] |One or more functions that can modify the HTTP request before it is sent. |undefined
+|options.numberMode |String |Set to `'precise'` to wrap deserialized numbers in typed wrappers that preserve the server's original type. |undefined
```
GumpacG: How many modes are there? If it's only precise or not can we rename the option and make it a boolean?

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1969,15 +1970,67 @@ g.V().hasLabel('person').groupCount().by('age')
 Either of the above two options accomplishes the desired goal as both prevent `groupCount()` from having to process
 the possibility of `null`.
 
+[[gremlin-javascript-numeric-types]]
+=== Numeric Types
+
+JavaScript has a single `Number` type (IEEE 754 double) which cannot distinguish between JVM numeric types. The driver
+provides typed wrapper classes and factory functions that give explicit control over serialization and deserialization.
+
+==== Controlling Serialization
```
spmallette: I don't think we should use sub-sections like this without strong reason. It just blows out the TOC without specific need. One solid section on Numeric Types makes more sense to me in this particular case.

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1969,15 +1970,67 @@ g.V().hasLabel('person').groupCount().by('age')
 Either of the above two options accomplishes the desired goal as both prevent `groupCount()` from having to process
 the possibility of `null`.
 
+[[gremlin-javascript-numeric-types]]
+=== Numeric Types
+
+JavaScript has a single `Number` type (IEEE 754 double) which cannot distinguish between JVM numeric types. The driver
```
Cole-Greer: Nit: I think it's better if we distance ourself from the JVM a bit and start referring to these as gremlin types (even though they are implicitly aligned with Java)
```suggestion
JavaScript has a single `Number` type (IEEE 754 double) which cannot distinguish between gremlin numeric types. The driver
```

docs/src/reference/gremlin-variants.asciidoc
Cole-Greer: +1 to something like `options.preciseNumbers` as a boolean which defaults to false.

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1969,15 +1970,67 @@ g.V().hasLabel('person').groupCount().by('age')
...
+----
+const { toLong, toInt, toFloat, toDouble, toShort, toByte } = gremlin.structure;
+
+g.V().has('age', toInt(29)).next();
+g.V().has('score', toFloat(3.14)).next();
+g.V().has('id', toLong('9007199254740993')).next();
+----
+
+`toLong()` accepts `number`, `string`, or `bigint`. String and bigint inputs support the full signed 64-bit range;
+number inputs must be within the safe integer range.
```
Cole-Greer: What happens to numbers which are outside of the safe integer range? Does it throw an exception? This is sort of like an explicit case from double to long in Java, I think it's fair to truncate to the nearest long, even if it is inherently imprecise outside of the safe integer range.

kirill-stepanishin: If a JS number outside safe range is added, it throws a `RangeError`. Users can still pass the value as a string or bigint to get the full 64-bit range (`toLong('9007199254740993')`), so I think throwing is fair given they've opted into precise mode.
I'm not too firm on it though, I'm happy to change it if you feel it makes more sense. For now I added a sentence in the docs explaining the behaviour.

docs/src/reference/gremlin-variants.asciidoc
kirill-stepanishin: Swapped to a boolean, thanks for the suggestion

docs/src/reference/gremlin-variants.asciidoc
kirill-stepanishin: Condensed into one section, thanks for the suggestion

docs/src/reference/gremlin-variants.asciidoc
kirill-stepanishin: Applied

GumpacG:
VOTE +1

kenhuuu:
VOTE +1

docs/src/reference/gremlin-variants.asciidoc
Cole-Greer: I can see valid arguments in favour of either approach. Perhaps it's best to keep the exception for now as it's easier to take away an exception in the future compared to adding one.

Cole-Greer:
VOTE +1

xiazcy:
VOTE +1

PR #3428, opened 2026-05-15
Subgraph support for GLVs

GumpacG:
### Description
Brings gremlin-javascript, gremlin-dotnet, and gremlin-go to feature parity with gremlin-python for the GraphBinary 4.0 Graph type (0x10), so that g.X().subgraph('sg').cap('sg').next() from any of the four GLVs returns a usable Graph data container instead of throwing an unknown-type error.

Added unit tests and feature tests.


gremlin-go/driver/graphBinaryGraph_test.go
```diff
@@ -0,0 +1,239 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/text/language"
+)
+
+// TestGraphSerializerRoundTrip verifies GraphBinary 4.0 Graph (0x10)
+// serialization/deserialization round-trip preserves vertices, edges,
+// vertex-properties and properties.
+func TestGraphSerializerRoundTrip(t *testing.T) {
```
xiazcy: Let's not create a new test file for just one graph type. I'd add `TestGraphSerializerRoundTrip` into the file `graphBinarySerializer_test.go`, which already contains round trip serialization tests, and `TestGraphString` into `graph_test.go`.

gremlin-go/driver/graphBinaryGraph_test.go
```diff
@@ -0,0 +1,239 @@
...
+		g.Edges[int32(3)] = &Edge{
+			Element: Element{Id: int32(3), Label: "knows"},
+			InV:     *v2, OutV: *v1,
+		}
+		assert.Equal(t, "graph[vertices:2 edges:1]", g.String())
+	})
+}
+
+// TestGraphSerializerRegistry verifies the graph dataType is wired up correctly.
+func TestGraphSerializerRegistry(t *testing.T) {
```
xiazcy: Is this set of tests necessary? Wouldn't roundtrip tests already catches these?

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -433,6 +426,60 @@ beyond this limit will be rejected with an error.
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-3247[TINKERPOP-3247]
 
+==== Subgraph Support in GLVs
```
spmallette, line 492: I think we need to mention python here too? it's essentially all the GLVs being announced for the first time. when i did python i don't think i added any docs because i knew all these other ones had to land too.

i also think this doesn't read like Upgrade Documentation. It should be doing more to explain the relevance of this feature and less about technical details and minutiae.

gremlin-go/driver/graph.go
```diff
@@ -26,7 +26,25 @@ import (
 )
 
 // Graph is used to store the graph.
+// In-memory collections of vertices and edges populated by GraphBinary 4.0
+// Graph (0x10) deserialization so that subgraph() results can be returned as
+// a usable data container. Maps are keyed by element id.
```
Cole-Greer: Nit: Might be a bit too much context here, we probably don't want to remember to update for GraphBinary 5 and such.
```suggestion
// In-memory collections of vertices and edges, typically produced by a subgraph() traversal. Maps are keyed by element id.
```

gremlin-go/driver/graphBinaryDeserializer.go
```diff
@@ -416,6 +418,194 @@ func (d *GraphBinaryDeserializer) readEdge() (*Edge, error) {
 	return e, nil
 }
 
+// readGraph reads a GraphBinary 4.0 Graph value (after the 0x10 type code + 0x00
+// value flag have already been consumed). Mirrors org.apache.tinkerpop.gremlin.
+// structure.io.binary.types.GraphSerializer (Java).
```
Cole-Greer: Similar to above, this comment probably doesn't need to exist. I think the name is self-descriptive enough.

gremlin-go/driver/graphBinaryGraph_test.go
GumpacG: Not necessarily, I added these tests to graphBinarySerializer_test.go as that's where similar tests are as well.

docs/src/upgrade/release-4.x.x.asciidoc
GumpacG: I've changed it to mention that all GLVs support subgraphs and took out the technical details. Thanks!

gremlin-go/driver/graphBinaryDeserializer.go
GumpacG: Removed. Thanks!

kenhuuu:
What does `subgraph("sg").cap("sg").asString()` return? I noticed that you made a change to what the string output of the Graph is and I'm wondering if it impacts this step. Are there tests for this?

Cole-Greer:
VOTE +1

GumpacG:
> What does `subgraph("sg").cap("sg").asString()` return? I noticed that you made a change to what the string output of the Graph is and I'm wondering if it impacts this step. Are there tests for this?

`asString()` is untouched and should remain the same. I added a feature test for this. I just corrected `toString()` for GLVs as it used to return `"graph[]"`.

kenhuuu:
> > What does `subgraph("sg").cap("sg").asString()` return? I noticed that you made a change to what the string output of the Graph is and I'm wondering if it impacts this step. Are there tests for this?
> 
> `asString()` is untouched and should remain the same. I added a feature test for this. I just corrected `toString()` for GLVs as it used to return `"graph[]"`.

Yea, there's still something a bit odd about this to me but I guess its comes down again to the fact that Java has more capabilities than the other GLVs. The deserializer in Java actually turns the Graph into a TinkerGraph so the toString() and asString() match, but you don't have this in the other GLVs. In any case, it's fairly small and I don't think anyone will probably notice.

VOTE +1

kenhuuu:
Merging as lazy consensus has been reached.

PR #3432, opened 2026-05-21
Bump commons configuration2 -> 2.15.0

smholsen:
**WIP** This chain was longer than I expected, so when I have time to get back to this I'll evaluate whether it makes sense to just get the relevant dependencies upgraded upstream instead. 
(Consumers need to overwrite `commons-configuration2` to avoid the ghsa until it is upgraded here.) 

Bumps `commons-configuration2` from `2.9.0` to `2.15.0` to fix `GHSA-337m-mw94-2v6g`. 
Also adds an explicit override in `hadoop-gremlin` to satisfy DependencyConvergence, as Hadoop `3.3.3` pins this at `2.1.1`. 
Excludes `commons-logging` from `commons-configuration2` in `gremlin-core`, as `2.15.0` bumps it to `1.3.6` which conflicts with the `1.2` pulled in transitively via Hadoop.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -28,6 +28,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 This release also includes changes from prior 3.7.x releases.
 
 * Add missing `Configuring` interface to `GraphStepPlaceholder` and `VertexStepPlaceholder`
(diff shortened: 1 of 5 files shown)


spmallette:
Bumped to 2.15.1 in 3.7+ - b2deb67fb993e143a70052ee6b8ca20c38896238

PR #3433, opened 2026-05-25
Implement Provider Defined Types (PDT)

Cole-Greer:
Add CompositePDT (0xF0) support enabling graph providers to define custom types that serialize/deserialize seamlessly across all GLVs without driver-side configuration. Replaces the TP3 CustomTypeSerializer mechanism.

Core (gremlin-core):
- Add @ProviderDefined annotation and immutable ProviderDefinedType POJO
- Add ProviderDefinedTypeSerializer for GraphBinary with wire format: fully-qualified type string + fully-qualified fields map
- Add PdtGraphSONSerializersV4 with g:CompositePdt type tag
- Add ProviderDefinedTypeAdapter<T> and ProviderDefinedTypeRegistry with ServiceLoader discovery, recursive hydration, and graceful degradation on adapter failure
- Integrate auto-hydration into GraphBinaryReader and GraphSONMapper
- Add GraphBinaryWriter auto-conversion for @ProviderDefined objects
- Cache reflection metadata per class for performance
- Support inherited fields via superclass walking
- Remove legacy CUSTOM(0x00) type mechanism entirely

Gremlin-lang:
- Add PDT("name", ["key":value]) literal to ANTLR grammar
- Server-side parser constructs ProviderDefinedType from PDT literals
- All GLV translators emit PDT literal syntax
- Registry-based and annotation-based auto-dehydration in translators
- All TranslateVisitors handle PDT for cross-language translation

GLV support (all languages):
- Python: ProviderDefinedType, serializer, registry with @provider_defined decorator, entry_points auto-discovery, registry wired through Client/DriverRemoteConnection
- JavaScript: ProviderDefinedType, CompositePDTSerializer, registry with explicit function pair registration, client options wiring
- Go: ProviderDefinedType struct, serializer/deserializer, PDTRegistry with struct tags, RegisterFuncs, PDTProvider interface, client wiring
- .NET: ProviderDefinedType, CompositePDTSerializer, registry with IProviderDefinedTypeAdapter<T>, [ProviderDefined] attribute with IncludedFields/ExcludedFields, assembly scanning, IMessageSerializer .SetPdtRegistry() interface method, client/connection wiring

Server and testing:
- PDT flows end-to-end through gremlin-server with TinkerGraph storing original objects and conversion at serialization boundary
- Test-jar with Point, Address, Person test types for Docker server
- Integration tests in all GLVs using gremlin-lang PDT literals
- Traversal API tests covering raw PDT, registry hydration, and annotation-based auto-dehydration round-trips


docs/src/dev/provider/pdt.asciidoc
```diff
@@ -0,0 +1,295 @@
...
+        this.x = x;
+        this.y = y;
+    }
+
+    public double getX() { return x; }
+    public double getY() { return y; }
+}
+----
+
+The `name` attribute is a unique identifier for the type, typically namespaced (e.g. `"x:Point"`). By default, all
```
spmallette: i think it would be good to at least suggest guidance on the namespacing we'd like. i assume, "<graph-name>:Point" or something.

docs/src/dev/provider/pdt.asciidoc
```diff
@@ -0,0 +1,295 @@
...
+
+[source,java]
+----
+// Java - values arrive as ProviderDefinedType by default
+ProviderDefinedType pdt = (ProviderDefinedType) vertex.value("location");
+String typeName = pdt.getName();           // "x:Point"
+Map<String, Object> props = pdt.getProperties(); // {x: 1.0, y: 2.0}
+----
+
+==== Optional Hydration
```
spmallette: So many tiny subsections - necessary?

docs/src/dev/provider/pdt.asciidoc
```diff
@@ -0,0 +1,295 @@
...
+
+To convert `ProviderDefinedType` into language-native objects, register adapters with a
+`ProviderDefinedTypeRegistry`. This is optional — basic consumption works without it.
+
+===== Java
+
+Implement `ProviderDefinedTypeAdapter<T>` and register via ServiceLoader (same mechanism as the provider side). The
+driver discovers adapters on the classpath and hydrates automatically.
+
+===== Python (planned)
```
spmallette: "planned"? not actually implemented yet or is this just a mistake in the docs? same question for .NET... 

if not implemented, why is it present for some languages but not others? when will that be available?

docs/src/dev/provider/pdt.asciidoc
```diff
@@ -0,0 +1,295 @@
...
+func init() {
+    pdt.RegisterFuncs("x:Point",
+        func(p Point) map[string]any { return map[string]any{"x": p.X, "y": p.Y} },
+        func(m map[string]any) Point { return Point{X: m["x"].(float64), Y: m["y"].(float64)} },
+    )
+}
+----
+
+[[pdt-wire-format-graphbinary]]
+=== Wire Format (GraphBinary V4)
```
spmallette: isn't this just GraphBinary documentation? link to that instead of placing it here?

docs/src/dev/provider/pdt.asciidoc
```diff
@@ -0,0 +1,295 @@
...
+                 "y", {"@type": "g:Double", "@value": 2.0}]
+    }
+  }
+}
+----
+
+The `fields` value is a typed map following standard GraphSON map encoding.
+
+[[pdt-migration-from-tp3]]
+=== Migrating from TP3 Custom Types
```
spmallette: i think we need Provider Upgrade Documentation that calls attention to PDTs.

docs/src/dev/provider/pdt.asciidoc
```diff
@@ -0,0 +1,295 @@
...
+
+Provider Defined Types allow graph providers to expose custom types that drivers can serialize and deserialize without
+manual configuration on the client side. A provider annotates a class (or registers an adapter for a class it doesn't
+own), and the type flows through the wire protocol automatically. Clients receive PDT values as structured objects they
+can use directly or hydrate into language-native types.
+
+[[pdt-for-graph-providers]]
+=== For Graph Providers
+
+==== Basic Usage
```
spmallette: I think examples from the Gremlin perspective would be useful here - basically using your example types in a query to help complete the full model.

Any reason there's no Reference Docs? It's probably up to providers to provide specifics about their custom types, but isn't there a use case with TinkerGraph? Maybe use `Color` as an example as we've done in the past.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/pdt/ProviderDefinedTypeIntegrateTest.java
```diff
@@ -0,0 +1,286 @@
...
+    public void shouldRetrievePdtViaBytecodeTraversal() throws Exception {
+        final Cluster cluster = TestClientFactory.build().create();
+        final Client client = cluster.connect();
+
+        try {
+            // Store via gremlin-lang PDT literal
+            client.submit(
+                    "g.addV('location').property('point', PDT(\"Point\", [\"x\":10, \"y\":20])).iterate()").all().get();
+
+            // Retrieve via bytecode traversal (gremlin-lang)
```
spmallette: bytecode?

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/pdt/ProviderDefinedTypeIntegrateTest.java
```diff
@@ -0,0 +1,286 @@
...
+
+            final JsonNode fields = value.get("fields");
+            assertEquals("g:Int32", fields.get("x").get("@type").asText());
+            assertEquals(1, fields.get("x").get(GraphSONTokens.VALUEPROP).intValue());
+            assertEquals("g:Int32", fields.get("y").get("@type").asText());
+            assertEquals(2, fields.get("y").get(GraphSONTokens.VALUEPROP).intValue());
+        }
+    }
+
+    @Test
```
spmallette: i see we don't have a test outside of Groovy for using `Point` directly in Gremlin. In practice, can you not just do that? Is it some test environment limitations that prevents this? i've forgotten how this works maybe...

Cole-Greer:
VOTE +1

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/pdt/ProviderDefinedTypeIntegrateTest.java
Cole-Greer: I've restructured the java tests, there are now cases which demonstrate using PDT's via Traversals in all 3 tiers of support (Raw PDT only, registered types, and annotated types).

gremlin-python/src/main/python/gremlin_python/structure/graph.py
```diff
@@ -141,3 +141,114 @@ def __getitem__(self, key):
...
     def __len__(self):
         return len(self.objects)
+
+
+class ProviderDefinedType(object):
+    def __init__(self, name, properties):
+        if not name:
+            raise ValueError("name cannot be null or empty")
+        self._name = name
+        self._properties = dict(properties) if properties else {}
```
GumpacG: Should we check that keys are Strings here? Other GLVs enforce String only keys.

docs/src/dev/provider/index.asciidoc
```diff
@@ -1334,6 +1334,213 @@ can be used as a reference on how these files can be used and its
...
+conn.setPdtRegistry(MyGraphTypeRegistry.build());
+GraphTraversalSource g = traversal().with(conn);
+----
+
+With this in place, `Point` objects round-trip transparently in both directions — the annotation handles outbound
+serialization and the registry handles inbound reconstruction.
+
+For driver users consuming PDTs, see the <<gremlin-variants,Gremlin Variants>> reference documentation for
+each language driver.
+
```
GumpacG, line 1579: Can you also add contraints and error behaviour section of some sort? I think something for property map keys must be `String` and the type `name` must be non-empty.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/TypeSerializerRegistry.java
```diff
@@ -185,21 +157,6 @@ public Builder withFallbackResolver(final Function<Class<?>, TypeSerializer<?>>
...
         }
 
-        /**
-         * Add {@link CustomTypeSerializer} by way of an {@link IoRegistry}. The registry entries should be bound to
-         * {@link GraphBinaryIo}.
-         */
-        public Builder addRegistry(final IoRegistry registry) {
-            if (null == registry) throw new IllegalArgumentException("The registry cannot be null");
-
-            final List<Pair<Class, CustomTypeSerializer>> classSerializers = registry.find(GraphBinaryIo.class, CustomTypeSerializer.class);
```
GumpacG, line 195: Can you also remove GraphBinaryIo.java since all references to it are removed now?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/pdt/ProviderDefinedTypeRegistry.java
```diff
@@ -0,0 +1,214 @@
...
+        @Override
+        public Map<String, Object> toProperties(final T obj) {
+            return ProviderDefinedType.from(obj).getProperties();
+        }
+
+        @Override
+        public T fromProperties(final Map<String, Object> properties) {
+            try {
+                final java.lang.reflect.Constructor<T> ctor = targetClass.getDeclaredConstructor();
+                ctor.setAccessible(true);
```
kenhuuu, line 197: Nit: Might want to warn JPMS-compliant providers ahead of time that they are going to need to "opens" their module to gremlin-core as well now to allow reflection.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/pdt/ProviderDefinedTypeRegistry.java
```diff
@@ -0,0 +1,214 @@
...
+    private final Map<String, ProviderDefinedTypeAdapter<?>> adaptersByName = new ConcurrentHashMap<>();
+    private final Map<Class<?>, ProviderDefinedTypeAdapter<?>> adaptersByClass = new ConcurrentHashMap<>();
+
+    private ProviderDefinedTypeRegistry() {}
+
+    /**
+     * Creates a registry populated via {@link ServiceLoader} discovery.
+     */
+    @SuppressWarnings("rawtypes")
+    public static ProviderDefinedTypeRegistry build() {
```
kenhuuu: This name is very misleading as it implies the Builder pattern which is not what its doing. You should name this `create` or something else.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/remote/DriverRemoteConnection.java
```diff
@@ -214,6 +216,18 @@ public static DriverRemoteConnection using(final Configuration conf) {
...
 
+    @Override
+    public ProviderDefinedTypeRegistry getPdtRegistry() {
+        return pdtRegistry;
+    }
+
+    /**
+     * Sets the {@link ProviderDefinedTypeRegistry} for registry-based dehydration in the gremlin-lang translator.
+     */
+    public void setPdtRegistry(final ProviderDefinedTypeRegistry pdtRegistry) {
```
kenhuuu, line 230: Why is this necessary? Shouldn't the SPI discovery process handle this for the user?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/pdt/ProviderDefinedType.java
```diff
@@ -0,0 +1,179 @@
...
+            this.name = name;
+            this.fields = fields;
+        }
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public Map<String, Object> getProperties() {
```
kenhuuu: Theres naming inconsistency here. This method is `getProperties` but the annotation is "(included/excluded)fields". It should be called one or the other but not both. I prefer fields as it makes it clear that its not the same thing as properties on Elements which overloads the term. I know fields is probably too Java-specific, but I'd rather the disambiguation upfront.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GremlinLang.java
```diff
@@ -252,6 +267,19 @@ private String argAsString(final Object arg) {
...
+        if (pdtRegistry != null) {
+            final Optional<ProviderDefinedTypeAdapter<?>> adapter = pdtRegistry.getAdapterByClass(arg.getClass());
+            if (adapter.isPresent()) {
+                @SuppressWarnings("unchecked")
+                final Map<String, Object> props = ((ProviderDefinedTypeAdapter) adapter.get()).toProperties(arg);
+                return argAsString(new ProviderDefinedType(adapter.get().typeName(), props));
+            }
+        }
+
+        if (arg.getClass().isAnnotationPresent(ProviderDefined.class)) {
```
kenhuuu, line 281: Is the ordering of adapter first then annotation a specific choice? If so, either add a maintenance comment or (probably better) add a unit test for this.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/pdt/ProviderDefinedType.java
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * Returns the hydrated object if this PDT was hydrated by a {@link ProviderDefinedTypeRegistry}, or {@code null}.
+     */
+    public Object getHydrated() {
+        return hydrated;
+    }
+
+    @Override
+    public boolean equals(final Object o) {
```
kenhuuu, line 169: Does hydrated need to be taken into account here? If two PDTs have the same name and properties but reference different hydrated objects, are they the same? Should probably run an equals() check there?

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/io/binary/HeapBuffer.java
```diff
@@ -0,0 +1,230 @@
...
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+
+/**
+ * A simple heap-based {@link Buffer} implementation for unit testing in gremlin-core.
+ */
+public class HeapBuffer implements Buffer {
```
kenhuuu: Why are these tests in gremlin-core? They've normally been added in gremlin-util. If that was done, this test buffer class should just wrap `Unpooled` from netty.

gremlin-server/pom.xml
```diff
@@ -173,6 +173,16 @@ limitations under the License.
                     </archive>
                 </configuration>
             </plugin>
+            <plugin>
+                <artifactId>maven-jar-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>test-jar</goal>
```
kenhuuu, line 187: Why is this needed?

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -489,10 +489,37 @@ unwrap(toInt(29));  // 29
...
+* <<gremlin-dotnet-pdt,Gremlin.Net>>
+* <<gremlin-go-pdt,Gremlin-Go>>
+
 === Upgrading for Providers
 
 ==== Graph System Providers
 
+===== Provider Defined Types
+
+TinkerPop 4 replaces the TP3 `CustomTypeSerializer` mechanism with Provider Defined Types (PDT). The key improvement
```
kenhuuu: You might want to highlight this first sentence more. Providers need to know that their current custom types will break and they have to move to using the ProviderDefined mechanisms.

GumpacG (commented):
Not in the scope of this PR, but it may be nice to have a function for users to call to get a list of registered PDTs.

Can you also add an integration test for nested PDTs where the outer is not registered and the inner is registered and vice versa?

VOTE +1

GumpacG:
Could you also look into this issue that Kiro found?

> .NET annotation hydration is unguarded and breaks the whole response (ProviderDefinedAttribute.HydrateIfRegistered + GraphBinaryReader.cs)
> 
>   `prop.SetValue(obj, Convert.ChangeType(value, prop.PropertyType));`
> 
>   This has no try/catch and no recursion into nested PDTs, yet it is called directly from GraphBinaryReader.ReadAsync:
> 
>   `return ProviderDefinedAttribute.HydrateIfRegistered(pdt);`
> 
>   Convert.ChangeType throws InvalidCastException/FormatException for any non-IConvertible or nested-PDT property value, which propagates up and fails deserialization of the entire result stream. Every other hydration path in the PR (Java registry, Python, JS, Go, .NET ProviderDefinedTypeRegistry.Hydrate) deliberately catches and returns the raw PDT. This one path violates that contract. For the nested Person→Address annotation case, Convert.ChangeType(ProviderDefinedType, typeof(Address)) will throw. Recommend wrapping in try/catch returning the raw pdt, and recursing for nested PDT-valued properties.

gremlin-python/src/main/python/gremlin_python/structure/graph.py
Cole-Greer: Good catch, I've added a runtime check in`ProviderDefinedType.__init__` to validate keys are strings.

docs/src/dev/provider/index.asciidoc
Cole-Greer: I've folded details about the String keys and non-empty field names into the basic usage section.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/pdt/ProviderDefinedTypeRegistry.java
Cole-Greer: I've renamed to `create` in Java, python, and .Net.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/GremlinLang.java
Cole-Greer: It was deliberate so users have an ability to override an annotated class if they so choose. I've added a comment and unit test.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/pdt/ProviderDefinedType.java
Cole-Greer: I've updated to use fields consistently throughout all GLVs.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/pdt/ProviderDefinedType.java
Cole-Greer: It's intentionally excluded, equality is based just on the name and properties, it's not impacted by if one has it's hydrated form attached. I've added a comment for this.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/pdt/ProviderDefinedTypeRegistry.java
Cole-Greer: Good point. Added a "JPMS Considerations" note to the provider docs for this.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/io/binary/HeapBuffer.java
Cole-Greer: This was here to keep the tests closer to the implementing classes in gremlin-core. I agree that reusing Unpooled is simpler though. I've relocated the tests and made the switch.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/remote/DriverRemoteConnection.java
Cole-Greer: There were a few issues with the wiring of automatic cases which have been addressed, but there's still a caveat, here's the current state:

Dehydration (outbound — sending a typed object as a PDT): Fully automatic, no setPdtRegistry needed.
- @ProviderDefined-annotated classes are converted automatically — GremlinLang detects the annotation and dehydrates with no registry involved.
- Adapter-registered types are also handled automatically: DriverRemoteConnection now defaults its registry to an SPI-discovered ProviderDefinedTypeRegistry.create(), so any ProviderDefinedTypeAdapter registered via META-INF/services is picked up. (If both apply, a registered adapter takes precedence over the annotation, so a user can override annotation-derived behavior.)

Hydration (inbound — reconstructing a typed object from a response): Automatic for adapter-discoverable types.
- The default GraphBinaryMessageSerializerV4 now wires the SPI-discovered registry into the reader, so types with a ProviderDefinedTypeAdapter on the classpath hydrate back into their typed form with zero configuration.
- Types with no registered adapter come back as a raw ProviderDefinedType (name + fields) — which is the intended default and works without any setup.
- There is no mechanism to automatically pick up annotated classes on the inbound path, adapters must be registered for hydration.

When you still need setPdtRegistry (or registry.register(...)):
1. Annotation-only hydration — a @ProviderDefined class with no adapter. Annotations are passive in Java (unlike Python decorators, they don't self-register), so SPI can't discover them for the inbound path. Calling registry.register(MyType.class) synthesizes an adapter from the annotation. Note these types 
still dehydrate automatically; it's only hydration that needs the registration.
2. Per-connection or programmatic registries — e.g. connecting to multiple graphs with different PDT schemas, or building a registry by hand rather than via SPI.

So for the common case (provider ships adapters via SPI, or users only need raw ProviderDefinedType values), no manual wiring is required. setPdtRegistry remains as the override for annotation-only hydration and custom/per-connection registries.

gremlin-server/pom.xml
Cole-Greer: That jar carries the PDT test fixtures (org.apache.tinkerpop.gremlin.server.pdt.{Point,Person,Address}) that the GLV integration tests reference (e.g. new org.apache.tinkerpop.gremlin.server.pdt.Point(...) in Groovy). Without the test-jar the image build fails at that COPY, which breaks the JS/Python/.NET/Go 
integration suites.

I've put the execution back and added a comment explaining the Dockerfile dependency.

kenhuuu:
VOTE +1

xiazcy:
VOTE +1

PR #3434, opened 2026-05-28
Add explicit transaction support to non-Java GLVs

kenhuuu:
Implement Transaction classes in gremlin-python, gremlin-go, gremlin-javascript, and gremlin-dotnet that mirror the Java HttpRemoteTransaction. Transaction wraps a Client and provides begin/commit/rollback/close lifecycle plus a submit() method for plain gremlin-lang strings.

gremlin-driver's default transaction close behavior is now rollback to match all GLVs.

Also changes how requests are submitted in gremlin-go. In particular it now waits for response headers before returning. This allows users to check if the response returns an err (pretty much any non-200 OK response), before having to check the ResultSet itself.


gremlin-js/gremlin-javascript/lib/driver/remote-connection.ts
GumpacG: If sessions are being removed, should this be removed as well?

gremlin-dotnet/src/Gremlin.Net/Driver/RemoteTransaction.cs
```diff
@@ -0,0 +1,270 @@
...
+    ///
+    ///     This class is NOT thread-safe. Do not share a RemoteTransaction across
+    ///     multiple threads without external synchronization.
+    /// </summary>
+    public class RemoteTransaction : IGremlinClient, IAsyncDisposable
+    {
+        private readonly IGremlinClient _client;
+        private readonly string _traversalSource;
+        // Serializes all submissions to guarantee ordering. Matches the role of
+        // Java's synchronized on HttpRemoteTransaction.submitInternal.
```
GumpacG: I think we can remove this line.

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -291,8 +291,42 @@ re-construction machine-side.
 [[gremlin-go-transactions]]
 === Transactions
 
-IMPORTANT: Transaction support over HTTP is not yet implemented in Gremlin-Go. This will be addressed by the official
-4.0.0 release.
+Gremlin-Go supports remote transactions over HTTP. Transactions can be created from a `GraphTraversalSource` or
```
spmallette: we can't exactly say scripts vs bytecode anymore, but GraphTraversalSource vs Client doesn't sound quite right either.  or maybe the way its introduced is sounding off to me? 

Upgrade docs describe it as: 

> * **Traversal API**: `g.tx().begin()` returns a transaction-bound `GraphTraversalSource`
> * **Driver API**: `client.transact()` (or `client.Transact()` in .NET) returns a `Transaction` with `submit()` methods

seems like it should at least be consistent with that.

docs/src/reference/intro.asciidoc
```diff
@@ -446,7 +446,8 @@ regard. Prefer explicit property key names in Gremlin unless it is completely im
 The third and final point involves transactions. Under this model, one traversal is equivalent to a single transaction
 and there is no way in TinkerPop to string together multiple traversals into the same transaction.
 
-IMPORTANT: 4.0.0-beta.1 Release - Remote transactions are not supported in this beta.
+Remote transactions are supported across all GLVs. See the <<transactions,Transactions>> section and each GLV's
```
spmallette: since this section is on limitations, i think the point is to assume there is complete transaction support. we'd only document the absence of support.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -41,6 +41,34 @@ anonymized form. The original gremlator.com was a prototype built by TinkerPop c
...
+* `gtx.tx()` returns the same transaction object (enabling `gtx.tx().commit()` pattern from the Traversal API).
+
+See the <<gremlin-drivers-variants,Gremlin Drivers and Variants>> reference documentation for language-specific
+syntax and examples.
+
+==== Transaction Default Close Behavior Changed
+
+The default behavior of `close()` on a remote transaction has been changed from `commit` to `rollback` across all
+GLVs (including the Java driver). This aligns with the embedded graph transaction default and is the safer behavior:
+partial work is discarded if the user forgets to call `commit()`. In Java, the behavior can still be overridden via
```
spmallette: Might be more accurate to say that "In Java with embedded mode,"

GumpacG (commented):
VOTE +1

gremlin-go/driver/transaction.go
```diff
@@ -0,0 +1,195 @@
...
+	defer t.mutex.Unlock()
+	return t.transactionId
+}
+
+func (t *Transaction) closeTransaction(script string) error {
+	t.mutex.Lock()
+	defer t.mutex.Unlock()
+
+	if !t.isOpen {
+		return newError(err1102TransactionRollbackNotOpenedError)
```
GumpacG: These are both being called by `Commit()` and `Rollback()` and should have different errors.

gremlin-dotnet/src/Gremlin.Net/Driver/RemoteTransaction.cs
```diff
@@ -0,0 +1,270 @@
...
+                    _txConnection?.MarkClosed();
+                }
+            }
+        }
+
+        /// <summary>
+        ///     Synchronous dispose (required by IGremlinClient/IDisposable).
+        ///     Does not attempt rollback. Use <c>await using</c> for proper cleanup.
+        /// </summary>
+        public void Dispose()
```
GumpacG, line 267: Should this have rollback for sync?

gremlin-go/driver/transaction_test.go
```diff
@@ -0,0 +1,529 @@
...
+func TestTransactionDoubleBegin(t *testing.T) {
+	client := newTxClient(t)
+	defer client.Close()
+
+	tx := client.Transact()
+	_, err := tx.Begin()
+	assert.Nil(t, err)
+
+	_, err = tx.Begin()
+	assert.NotNil(t, err)
```
Cole-Greer, line 192: Nit: It would be nice to also assert the error code or part of the error message for a lot of these cases.

Cole-Greer:
I agree with above comments from Stephen and Guian, but overall LGTM.

Cole-Greer:
One extra thing to note, this PR deletes a lot of the old commented out/skipped transaction tests in favour of a fresh set of tests. I'm fine with that change but I was comparing the old and new tests and found a few minor gaps:

First of all there are still some old skipped transaction tests in python which have not been cleaned up here: https://github.com/apache/tinkerpop/blob/7c2cd79a9f4f184ea15ed8b0a5a62d7acef82348/gremlin-python/src/main/python/tests/integration/process/test_traversal.py#L95

Also the old Python and Go transaction tests seemed to have been a bit richer than the other GLVs, and contain a few cases which are not replicated in the new tests:

| Old scenario | Go old test | Python old test | In new tests? |
|---|---|---|---|
| Two concurrent txns, both rolled back, graph unchanged | Test multi rollback Transaction | test_multi_rollback_transaction | ❌ No |
| Two concurrent txns, one commits + one rolls back, only committed data persists | Test multi commit and rollback Transaction | test_multi_commit_and_rollback | ❌ No |
| Close from parent: closing the parent DriverRemoteConnection closes/rolls back child txns | Test Transaction close tx from parent | test_transaction_close_tx_from_parent | ❌ No |

gremlin-go/driver/transactionRemoteConnection.go
```diff
@@ -0,0 +1,39 @@
...
+type transactionRemoteConnection struct {
+	transaction *Transaction
+}
+
+func (t *transactionRemoteConnection) Close() {
+	// No-op: the transaction lifecycle is managed by Transaction.Close/Commit/Rollback
+}
+
+func (t *transactionRemoteConnection) submitGremlinLang(gremlinLang *GremlinLang) (ResultSet, error) {
+	if !t.transaction.isOpen {
```
xiazcy, line 33: Nit: should this use the method instead for safer access with lock?
```suggestion
	if !t.transaction.isOpen() {
```

xiazcy:

gremlin-dotnet/src/Gremlin.Net/Driver/RemoteTransaction.cs
kenhuuu: This is actually a consequence of the interface of `IGremlinClient` which is `IDisposable`. Because I wanted to extend the same interface, it meant that `RemoteTransaction` also needs to implement a `Dispose` but its not safe to rollback there because of how certain .NET runtimes with a `SynchronizationContext` behave. This means that .NET users will have to use `await using` rather than `using`.

gremlin-go/driver/transactionRemoteConnection.go
kenhuuu: Transactions are documented as being not thread safe, and `sync.Mutex` is not re-entrant, but could cause deadlock problems within the driver itself if `isOpen()` was used. I could make the change but thats the trade off that would be made. Slightly better handling if a user accidentally uses this in multiple goroutines vs potential deadlock.

docs/src/upgrade/release-4.x.x.asciidoc
kenhuuu: It's actually configurable in both. Lengthened it a bit to explain that.

kenhuuu:
VOTE +1

PR #3447, opened 2026-06-02
TINKERPOP-3250 Standarize interceptor API and JSON request amongst GLVs

kenhuuu:
https://issues.apache.org/jira/browse/TINKERPOP-3250

Redesign the request interceptor API across all GLVs to use a consistent mutate-only contract. Interceptors receive a mutable HTTP request object and modify it in place. The driver auto-serializes the request body to JSON (application/json) after all interceptors run via an idempotent serializeBody() method on the request object. Responses remain GraphBinary.

The GraphBinary request serializers are retained in each GLV but are no longer wired into the default request path. Users who need GraphBinary request bodies can write a custom interceptor that calls the serializer.

JavaScript: Add serializeBody() to HttpRequest class. Add toJSON() to RequestMessage for clean serialization without field-list duplication.

The bulkResults field is changed from a string to a boolean in gremlin-go and gremlin-dotnet to match the provider specification which defines it as a JSON boolean.

Assisted-by: Kiro:claude-opus-4-6

VOTE +1


docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -32,6 +32,33 @@ complete list of all the modifications that are part of this release.
 
 === Upgrading for Users
 
+==== Request Interceptors
```
spmallette, line 35: As Upgrade Documentation I think this section should be grounded in what we had before a little bit. "When TinkerPop supported websockets prior to 4.0.0, it offered a `RequestInterceptor` interface that allowed <stuff>. With the move to HTTP the notion of the 'interceptor' has shifted...."

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -272,6 +272,47 @@ link:https://tinkerpop.apache.org/docs/x.y.z/dev/provider/#_graph_driver_provide
 |RequestInterceptors |Functions that modify HTTP requests before sending. Used for authentication and custom headers. |empty
 |=========================================================
 
+[[gremlin-go-interceptors]]
+=== RequestInterceptor
+
+Gremlin-Go allows modification of the underlying HTTP request through `RequestInterceptor` functions. This is
+intended to be an advanced feature which means that you will need to understand how the implementation works in order
+to safely utilize it. Gremlin-Go is written in a way that you should be able to interact with a TinkerPop-enabled
+server without having to use interceptors. This is intended for cases where the server has special capabilities.
```
Cole-Greer: > Gremlin-Go is written in a way that you should be able to interact with a TinkerPop-enabled
server without having to use interceptors.

I think this is too strong of a statement. It narrows the definition of a "TinkerPop-enabled server" to essentially just gremlin-server with default settings, and gate-keeps a pretty basic feature like auth for "advanced users only".

gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphBinary4/RequestMessageSerializer.cs
```diff
@@ -31,6 +31,12 @@ namespace Gremlin.Net.Structure.IO.GraphBinary4
     /// <summary>
     ///     Serializes a <see cref="RequestMessage"/> in the GraphBinary 4.0 wire format.
     /// </summary>
+    /// <remarks>
+    ///     This serializer is no longer used by the default driver flow (which now serializes
```
Cole-Greer, line 35: I'd like to revisit if the GB Request serializers should be removed entirely following this change, however I would agree that this is not in scope of this PR.

gremlin-python/src/main/python/gremlin_python/driver/connection.py
```diff
@@ -34,17 +35,23 @@ def __init__(self, status):
...
             interceptors = [interceptors]
-        elif not (isinstance(interceptors, tuple)
-                  or isinstance(interceptors, list)
-                  or interceptors is None):
+        elif isinstance(interceptors, tuple):
+            interceptors = list(interceptors)
+        elif not (isinstance(interceptors, list) or interceptors is None):
             raise TypeError("interceptors must be a callable, tuple, list or None")
 
+        # Auth is just an interceptor. As a convenience (and for discoverability), the auth
```
Cole-Greer, line 49: Python is the only GLV which maintains a separate config for `auth` beyond the general `interceptor` configuration option. I think this is a good opportunity to simplify things and bring python inline with the other GLVs.

kenhuuu: I think we should actually do the opposite. Each GLV should have a way to add an auth to help with discoverability. Otherwise, users might not understand how to add the default Auth interceptors.

docs/src/reference/gremlin-variants.asciidoc
kenhuuu: I'll soften the language a bit by making it "with most TinkerPop-enabled servers".

gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphBinary4/RequestMessageSerializer.cs
kenhuuu: You should start a thread on the devlist for this.

xiazcy:
VOTE +1

Cole-Greer:
VOTE +1

PR #3448, opened 2026-06-03
Make Tree no longer extend HashMap

GumpacG:
## Summary
  Changes `Tree` from `extends HashMap<T, Tree<T>>` to a `final` class that
  holds a `Map` internally and exposes a tree-shaped public API. `Tree` is no
  longer a `Map`, which removes inherited `Map` methods whose behavior did not
  align with tree semantics (e.g. `size()` returned root-entry count, not total
  nodes; `containsKey` only checked immediate children).
  
  Per the dev-list discussion: https://lists.apache.org/thread/o0nqh6kmrkdht531655p351ldjll045d
  
  This is a breaking change targeted at TinkerPop 4.0.0.
  
  ## Changes
  
  `Tree` now exposes:
  - Navigation: `rootNodes()`, `childAt(key)` (throws if absent), `hasChild(key)`,
    `contains(value)` (recursive), `findSubtree(key)` (recursive, `Optional`),
    `getOrCreateChild(key)` (construction primitive).
  - Structure: `isLeaf()`, `nodeCount()`, `getNodesAtDepth(int)`,
    `getTreesAtDepth(int)`, `getLeafNodes()`, `getLeafTrees()`.
  - Composition/output: `addTree()`, `splitParents()`, `prettyPrint()`,
    structural `equals`/`hashCode`/`toString`.
  
  Behavior fixes that fall out of the rewrite:
  - `isLeaf()` returns `true` on an empty tree instead of throwing.
  - `getNodesAtDepth(0)` returns the root nodes (now 0-based).
  
  Internal callers migrated off the removed `Map` surface:
  - `TreeStep` / `TreeSideEffectStep` use `getOrCreateChild` during construction.
  - GraphBinary `TreeSerializer` and GraphSON `TreeJacksonSerializer`/
    `Deserializer` (V1–V4) iterate via `rootNodes()`/`childAt()` and rebuild via
    `getOrCreateChild()` + `addTree()`.
  - `DetachedFactory` / `ReferenceFactory` gained an explicit `Tree` branch
    (previously handled via the `instanceof Map` path).
  
  ## Breaking changes
  
  - Code treating a `Tree` as a `Map` (assignment to `Map`, `instanceof Map`,
    `get`/`put`/`keySet`/`entrySet`/`size`) no longer compiles.
  - Gremlin patterns that worked only because `Tree` was a `Map` (`select(keys)`,
    `count(local)`, `unfold()` on a tree) no longer compose; process the result
    client-side after `next()` or reshape the upstream traversal.

 ## Speed and Memory
  
  Analytical estimates (not benchmarked); the backing `HashMap` table and entries
  are identical before and after.
  
  - **Memory:** ~16 B more per node — `Tree` now holds a `HashMap` instead of
    being one, adding one wrapper object per node. The map itself is unchanged.
  - **Speed:** comparable. Construction is slightly faster (`getOrCreateChild` is
    1–2 map ops vs. the old 3); `childAt` is one op slower (collapsible to one).
    Wire-format encoding is unchanged.
  
  ## Testing
  
  - `TreeTest` rewritten against the new API (navigation, recursive
    containment/search, depth queries, leaf detection, null keys, structural
    equality, `toString`, pretty-print); `TreeSupplierTest` updated.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -27,6 +27,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 
 * Added configurable CORS `allowedOrigins` setting to Gremlin Server; warns when wildcard origin is used alongside authentication.
 * Fixed `ByteBuf` leak in `GraphBinaryMessageSerializerV4` when serialization throws an `IOException`.
+* Changed `Tree` to no longer extend `HashMap`; it is now a final class with a tree-shaped API (`childAt`, `hasChild`, `contains`, `findSubtree`, `getOrCreateChild`, `getNodesAtDepth`, `getLeafNodes`, `nodeCount`) and is no longer a `Map`.
 * Added typed numeric wrappers and `preciseNumbers` connection option to `gremlin-javascript` for explicit control over numeric type serialization and deserialization.
 * Added `NextN(n)` to `Traversal` in `gremlin-go` for batched result iteration, providing API parity with `next(n)` in the Java, Python, and .NET GLVs, and updated the Go translators in `gremlin-core` and `gremlin-javascript` to emit `NextN(n)` for the batched form.
 * Added Provider Defined Types (PDT) support — graph providers can define custom types via `@ProviderDefined` annotation that serialize/deserialize seamlessly across all GLVs without driver-side configuration. Replaces TP3 custom type mechanism.
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -28,6 +28,8 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Added configurable CORS `allowedOrigins` setting to Gremlin Server; warns when wildcard origin is used alongside authentication.
 * Fixed `ByteBuf` leak in `GraphBinaryMessageSerializerV4` when serialization throws an `IOException`.
 * Changed `Tree` to no longer extend `HashMap`; it is now a final class with a tree-shaped API (`childAt`, `hasChild`, `contains`, `findSubtree`, `getOrCreateChild`, `getNodesAtDepth`, `getLeafNodes`, `nodeCount`) and is no longer a `Map`.
+* Changed `count(local)` on a `Tree` to return the total node count (`Tree.nodeCount()`) instead of the root-entry count.
+* Changed Tree class in Java to not extend from HashMap and offered a new tree-shaped API for navigation.
 * Added typed numeric wrappers and `preciseNumbers` connection option to `gremlin-javascript` for explicit control over numeric type serialization and deserialization.
 * Added `NextN(n)` to `Traversal` in `gremlin-go` for batched result iteration, providing API parity with `next(n)` in the Java, Python, and .NET GLVs, and updated the Go translators in `gremlin-core` and `gremlin-javascript` to emit `NextN(n)` for the batched form.
 * Added Provider Defined Types (PDT) support — graph providers can define custom types via `@ProviderDefined` annotation that serialize/deserialize seamlessly across all GLVs without driver-side configuration. Replaces TP3 custom type mechanism.
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -5162,9 +5162,12 @@ The resultant tree data structure can then be navigated with its tree-shaped API
 [gremlin-groovy,modern]
 ----
 tree = g.V().out().out().tree().by('name').next()
-tree.childAt('marko')          // with the SugarPlugin: tree['marko']
-tree.childAt('marko').childAt('josh')          // with the SugarPlugin: tree['marko']['josh']
+tree.childAt('marko')
+tree.childAt('marko').childAt('josh')
 tree.getNodesAtDepth(2)
+// The sugar plugin offer convenience accessors
+tree['marko']
+tree['marko']['josh']
 ----
 
 Note that when using `by()`-modulation, tree nodes are combined based on projection uniqueness, not on the
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -5156,14 +5156,14 @@ It is important to see how the paths of all the emanating traversers are united 
 
 image::tree-step2.png[width=500]
 
-The resultant tree data structure can then be manipulated (see `Tree` JavaDoc).
+The resultant tree data structure can then be navigated with its tree-shaped API such as `childAt()`, `findSubtree()`, `getNodesAtDepth()`, and `getLeafNodes()` (see `Tree` JavaDoc).
 
 [gremlin-groovy,modern]
 ----
 tree = g.V().out().out().tree().by('name').next()
-tree['marko']
-tree['marko']['josh']
-tree.getObjectsAtDepth(3)
+tree.childAt('marko')
+tree.childAt('marko').childAt('josh')
+tree.getNodesAtDepth(2)
 ----
 
 Note that when using `by()`-modulation, tree nodes are combined based on projection uniqueness, not on the
@@ -5178,6 +5178,11 @@ g.V().has('name','josh').out('created').values('name').
 
 <1> When the `tree()` is created, vertex 3 and 5 are unique and thus, form unique branches in the tree structure.
 <2> When the `tree()` is `by()`-modulated by `label`, then vertex 3 and 5 are both "software" and thus are merged to a single node in the tree.
+
+NOTE: A `Tree` keys its children by node value, so sibling branches that resolve to the same value (whether the
+original objects or a `by()`-projection) collapse into a single node. This means structure can be lost when distinct
+paths share a value at the same depth. If every path must be preserved, use `path()`; if a queryable result that
+retains element identity, cycles, and parallel edges is required, use `subgraph()`.
 
 The `tree()` step can also take a side-effect key as an argument. When using this form, the `Tree` is is built up in a
 side-effect as each traverser passes through. The `Tree` can later be accessed by either `select()` or `cap()`.
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -1481,8 +1481,9 @@ g.V().hasLabel('person').outE('created').count().map {it.get() * 10}.path() <2>
 <1> `count()`-step is a <<a-note-on-barrier-steps,reducing barrier step>> meaning that all of the previous traversers are folded into a new traverser.
 <2> The path of the traverser emanating from `count()` starts at `count()`.
 
-IMPORTANT: `count(local)` counts the current, local object (not the objects in the traversal stream). This works for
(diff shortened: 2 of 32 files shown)


docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -501,10 +501,49 @@ It cannot be passed to `traversal().with(...)`, and mutating its collections has
...
+`isLeaf()` no longer throws on an empty tree (it returns `true`), and a `Tree` keeps the long-standing limitation
+that sibling branches resolving to the same value collapse into one node; use `path()` or `subgraph()` when full
+path structure must be preserved.
+
+
 === Upgrading for Providers
 
 ==== Graph System Providers
 
+`Tree` no longer extends `HashMap`. Provider code that inspected or rebuilt a `Tree` via `Map` methods
```
kenhuuu: This section assumes that the provider uses the reference deserializers. Providers can technically deserialize the tree bytes into whatever class they want. While this information is good as most providers will probably do that, it should probably more of a heads up to providers that do that they should expect changes.

kenhuuu: It's also missing a section header.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/CountLocalStep.java
```diff
@@ -42,6 +43,9 @@ public CountLocalStep(final Traversal.Admin traversal) {
     @Override
     protected Long map(final Traverser.Admin<S> traverser) {
         final S item = traverser.get();
+        if (item instanceof Tree)
```
kenhuuu: Why specialize case Tree here? Wouldn't the default just return a singletonIterator so the count would be 1?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/Tree.java
```diff
@@ -20,125 +20,249 @@
...
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
 
 /**
+ * A tree data structure with a tree-shaped public API that is composed of a {@link HashMap}.
+ * <p>
+ * Internal storage is a plain {@link HashMap}, so iteration order is not promised and sibling-key collisions still
```
kenhuuu: Very small nit: I'm not sure you need to describe what the internals are. You just need to explain the limitations.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/TreeTest.java
```diff
@@ -24,85 +24,233 @@
...
+        // calling again returns the same instance
+        assertSame(child, tree.getOrCreateChild("a"));
+
+        // mutating the returned subtree is observable through the parent
+        child.getOrCreateChild("a1");
+        assertTrue(tree.childAt("a").hasChild("a1"));
+    }
+
+    @Test
+    public void shouldAllowNullKeys() {
```
kenhuuu, line 145: You might want to rethink this. If you are going to implement this using maps in the other GLVs, null is not a valid key in a lot of languages.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/Tree.java
Cole-Greer: To add a bit more to that, collapsing of sibling-key collisions is described as a bug or limitation when according to TINKERPOP-1693 it is expected/intended behaviour.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/sideEffect/Tree.feature
```diff
@@ -206,36 +206,3 @@ Feature: Step - tree()
                   |--josh
                   |--peter
       """
-
-  Scenario: g_V_out_treeXaX_selectXaX_countXlocalX
-    Given the modern graph
-    And the traversal of
-      """
-      g.V().out().tree("a").select("a").count(local)
```
Cole-Greer, line 214: I guess these tests were broken as `tree().count(local)` was reliant on Tree being a map to give the count. I'd ideally like to avoid removing these tests. They weren't actually intended to test semantics of `tree().count(local)`, the pair of tests was meant to verify that TreeSideEffectStep acts as a barrier (processes all incoming traversers before producing a single result). Using `count(local)` to introspect the Tree from within gremlin was simply a nice convenience as `g.V().out().tree("a").select("a")` returns 6 references to the same Tree object, which is difficult for the tests to handle.

What are your thoughts on adding in a new case to `CountLocalStep` to handle `Tree`? It could simply call `rootNodes().size()` to retain the old semantics of counting the number of roots, although the more intuitive semantics would be if `tree().count(local)` mapped to the new `nodeCount()` method. I'm not sure how others may feel about a minor semantic break such as this.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -5049,14 +5049,14 @@ It is important to see how the paths of all the emanating traversers are united
...
-The resultant tree data structure can then be manipulated (see `Tree` JavaDoc).
+The resultant tree data structure can then be navigated with its tree-shaped API such as `childAt()`, `findSubtree()`, `getNodesAtDepth()`, and `getLeafNodes()` (see `Tree` JavaDoc).
 
 [gremlin-groovy,modern]
 ----
 tree = g.V().out().out().tree().by('name').next()
-tree['marko']
-tree['marko']['josh']
-tree.getObjectsAtDepth(3)
+tree.childAt('marko')
```
spmallette, line 5165: would be nice if we didn't lose the groovy syntax. `childAt` is better than the Java alternative but less so for Groovy. a bit of Groovy sugar would bring that back.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -501,10 +501,49 @@ It cannot be passed to `traversal().with(...)`, and mutating its collections has
...
+Replacements for the common `Map`-based access patterns:
+
+[options="header"]
+|=======================
+|3.x (`Tree` as `Map`) |4.x (`Tree` API)
+|`tree.get(key)` |`tree.childAt(key)` (throws if absent) or `tree.findSubtree(key)` (recursive, returns `Optional`)
+|`tree.containsKey(key)` |`tree.hasChild(key)`
+|`tree.keySet()` |`tree.rootNodes()`
+|`tree.size()` |`tree.rootNodes().size()` for root entries, or `tree.nodeCount()` for total nodes
+|`getObjectsAtDepth(d)` |`getNodesAtDepth(d)` (now 0-based: depth 0 returns the roots)
```
spmallette, line 653: i know we wanted more tree terminology so we moved from "Objects" to "Node", but it just occurred to me that the return here isn't exactly a "node", is it? I wonder if that will be confusing.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -501,10 +501,49 @@ It cannot be passed to `traversal().with(...)`, and mutating its collections has
...
+A few Gremlin patterns that worked only because `Tree` was a `Map` (for example `select(keys)`, `count(local)`,
+and `unfold()` applied to a `Tree`) no longer compose. Process the `Tree` result client-side after `next()`, or
+reshape the upstream traversal. The old `count(local)` (which returned the root-node count) is replaced by
+`rootNodes().size()` for the root count, or `nodeCount()` for the total node count.
+
+`isLeaf()` no longer throws on an empty tree (it returns `true`), and a `Tree` keeps the long-standing limitation
+that sibling branches resolving to the same value collapse into one node; use `path()` or `subgraph()` when full
+path structure must be preserved.
+
+
```
spmallette, line 672: No JIRA for this right? We do have a DISCUSS thread though we could do a "See: " for further details.

CHANGELOG.asciidoc
```diff
@@ -25,6 +25,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 [[release-4-0-0]]
 === TinkerPop 4.0.0 (Release Date: NOT OFFICIALLY RELEASED YET)
 
+* Changed `Tree` to no longer extend `HashMap`; it is now a final class with a tree-shaped API (`childAt`, `hasChild`, `contains`, `findSubtree`, `getOrCreateChild`, `getNodesAtDepth`, `getLeafNodes`, `nodeCount`) and is no longer a `Map`.
```
spmallette: This entry is a bit long and detailed. Maybe "Changed `Tree` class in Java to not extend from `HashMap` and offered a new tree-shaped API for navigation."

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/CountLocalStep.java
GumpacG: Tree used to return count of root nodes which is misleading. As per the [devlist post](https://lists.apache.org/thread/15lslfrs5f27hsz00k147mwc38o58w5k), it is not counting all nodes.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/step/util/TreeTest.java
GumpacG: Null will be allowed as removing it could remove valuable paths in the tree. If for example, it won't be possible in .NET, then it would be a documented limitation or have an implementation that is not in the default library of .NET.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/sideEffect/Tree.feature
GumpacG: As per the [devlist post](https://lists.apache.org/thread/15lslfrs5f27hsz00k147mwc38o58w5k), it is not counting all nodes. Thanks for the suggestion.

docs/src/reference/the-traversal.asciidoc
GumpacG: I've updated it to include tree in the SugarLoader to include support the old syntax. Docs will just have the old syntax in the comment as an option with the sugar plugin.

docs/src/upgrade/release-4.x.x.asciidoc
GumpacG: Added, thanks!

docs/src/upgrade/release-4.x.x.asciidoc
GumpacG: I agree that it's not exactly a "node" however, renaming to "objects" may make other tree functions not very tree-like. For example, `nodeCount()` would have to be `objectCount()` and `rootNodes()` would have to be `rootObjects()`. I think it would be more conducive for tree users to keep "node".

kenhuuu:
VOTE +1

Cole-Greer:
VOTE +1

xiazcy:
VOTE +1

PR #3451, opened 2026-06-07
Gremlin dart driver

kartikey321:
https://issues.apache.org/jira/browse/TINKERPOP-3251

Introduces `gremlin-dart`, a new Dart language driver for Apache TinkerPop 4.x.

## Design

Follows the HTTP-only, GraphBinary v4 model used by other TinkerPop 4.x drivers:

- **Transport**: HTTP POST via `dart:io` with chunked-encoding support for Netty HTTP trailers
- **Serialization**: GraphBinary v4 (`application/vnd.graphbinary-v4.0`) — custom reader/writer, no external binary dependencies
- **Query language**: gremlin-lang strings built by `GremlinLang`
- **Bindings**: Parameters sent as a typed GraphBinary Map field (`0x0A`)
- **DSL**: Full `GraphTraversalSource` + `GraphTraversal` (~80 steps), anonymous traversals via `__`, `P`/`TextP` predicates, all traversal strategies

## Structure

```
gremlin-dart/
  lib/
    driver/          # Connection, Client, DriverRemoteConnection, RequestMessage
    process/         # GremlinLang, GraphTraversal DSL, traversal types
    structure/       # Graph types, GraphBinary reader/writer
  test/
    feature/         # Gherkin runner against live 4.0.0-beta.2 server
    unit/            # Offline GraphBinary encode/decode, DSL string generation
  tool/
    test_live.dart   # Manual smoke test (requires docker compose up)
  docker/            # Dockerfile + server config for integration tests
```

## Test results

2121 / 2123 Gherkin feature tests pass against TinkerPop 4.0.0-beta.2.

Two known failures are server-side, not client bugs:

- `g_injectXnull_nullX_conjoinXplusX` — server throws on `conjoin(null, null)`
- `g_withSackX2147483647iX_injectX0_5fX_sackXdivX_sack` — float32 rounding: server returns `4294967296.0f` vs expected `4294967294.0f`

## Running tests

```bash
# Start server
docker compose -f gremlin-dart/docker-compose.yml up -d --wait

# Feature tests (requires server)
dart test gremlin-dart/test/feature/

# Unit tests (offline)
dart test gremlin-dart/test/unit/
```

## Checklist
- [x] Apache license header on all source files
- [x] CHANGELOG entry added under TinkerPop 4.0.0
- [x] `dart analyze` reports no issues
- [x] 2121/2123 Gherkin feature tests pass against 4.0.0-beta.2
- [x] No external dependencies beyond `uuid`


spmallette:
Hi @kartikey321 - thanks for this pull request. I see that you are heavily into Flutter. Do you use TinkerPop in your work there? Could you also offer a few words on the options that Dart developers have to work with graphs in general (I'm, of course curious about Gremlin, but would like to hear something about the wider graph ecosystem)?

With respect to this pull request itself, I think I'd like to begin a discussion on the dev list that talks about having official Dart support per your pull request, but also the larger issue of the likely trend that TinkerPop sees more contributions like this and how we handle them.  I suspect I will send that email out tomorrow. I'll post the link to the discussion on this PR when i send it. Anyway, I thought I'd at least respond here, ask a few general questions so that you'd see your work was noted and set the expectation that it may take some time to really provide much feedback here while the community discusses the topics I have in mind. Thank you!

kartikey321:
Hi @spmallette, thank you for taking a look and for considering the pr.

  Yes, we use TinkerPop indirectly in my work. We use JanusGraph for graph-based access control, and JanusGraph’s
  Gremlin/TinkerPop API is the main interface we build around. That was part of the motivation for this PR.

  More generally, from the Dart/Flutter side, the current options for working with Gremlin are limited. In practice,
  Dart teams usually end up doing one of the following:

  - call Gremlin Server directly over HTTP with hand-written request/response handling,
  - put another service in front of Gremlin using an existing Java/Go/JavaScript/Python driver,
  - or use a non-Dart driver from a separate backend component.

  Those approaches work, but they add friction when the application logic is already written in Dart. Maintaining
  closely related graph/query logic across multiple languages becomes harder than it needs to be, especially for
  teams using Dart on both client and server.

  For the wider Dart graph ecosystem, there are general-purpose graph data structure packages and some database-
  specific clients, but there does not appear to be a maintained Gremlin/TinkerPop driver that supports the TinkerPop
  4.x HTTP + GraphBinary model. As Dart continues to see more server-side usage, I think having a Gremlin driver
  would be useful both for Dart developers and for the TinkerPop ecosystem.

  I understand that accepting a new official GLV has broader maintenance and community implications, so a dev-list
  discussion makes sense. I’m happy to participate in that discussion, adjust the implementation to match TinkerPop
  expectations, and help maintain the driver if the community decides this belongs in the main repository.

spmallette:
Great reply - thank you! I personally lean toward accepting this PR in some way, but I'm curious what others think. In that sense I've not really made my post to the dev list about this PR but more about the more general issue of accepting variants that I alluded to. Here's the link to the post in case you want to follow along: https://lists.apache.org/thread/4c88gjk09vdlsnr2n7sqcndv8b2rgw9h

Cole-Greer:
Hi @kartikey321,

Thanks for putting this together, I'd love to see if we can start expanding our offerings of GLVs, now that agentic tooling can reduce some of the maintenance burden. I'll leave most of the discussion on finding a home for this work to Stephen's devlist thread (I think this would be a great fit in a revived Tinkubator).

I'm curious how you approached the development of the driver, was it derived from one of the existing GLVs or built primarily from spec/documentation? Did you have to hand-code much of it or was AI helpful in accelerating the development?

I've taken a quick look at the build and test setup, and there are a few details I wanted to make note of, although it may be best to let the devlist discussion play out for a few more days before making large changes.

1. For the non-Java GLVs, we like to run the build and tests from within Docker to limit the number of runtimes developers need to manage. The [setup from Python](https://github.com/apache/tinkerpop/blob/master/gremlin-python/docker-compose.yml) is a reasonable example to copy from. It's also worth noting how python pulls in the existing server image from `docker/gremlin-test-server/Dockerfile` instead of maintaining a bespoke test server configuration per driver.
2. I noticed that the feature tests are currently extracting the gremlin scripts from the feature test scenarios and feeding them into `Connection.submit()` as strings. That's a good starting place (gets good coverage of the transport and serialization), but ideally we want the feature tests to generate GraphTraversal's and execute those as that is most aligned with how users are expected to use the drivers. Our setup for doing such is a bit complex as we do not have a gremlin parser in most languages. The current infrastructure is we maintain [translators in gremlin-core](https://github.com/apache/tinkerpop/tree/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator), and these translators get invoked by [a groovy script](https://github.com/apache/tinkerpop/blob/master/gremlin-python/build/generate.groovy) in each GLV to parse and translate [all of the scenarios into the native language variant](https://github.com/apache/tinkerpop/blob/master/gremlin-python/src/main/python/tests/feature/gremlin.py). I'd be happy to help with this setup if it proves troublesome.
3. Regarding your failing feature tests:
    a. g_injectXnull_nullX_conjoinXplusX (`g.inject([null,null]).conjoin("+")`) I wonder if something may be going on with escaping the `"+"` argument or handling of the empty string response. This scenario is expected to work in the server.
    b. g_withSackX2147483647iX_injectX0_5fX_sackXdivX_sack, it's worth double checking the client isn't doing any unexpected lossy type conversions, but this is likely a case where it's fine to add a little tolerance to floating point equality checks in the test suite.

kartikey321:
Hi @Cole-Greer, thanks for taking a look and for the detailed notes.

  On how I approached it: the initial architecture was based on the TinkerPop 4.x HTTP transport and GraphBinary v4 format. I used the existing GLVs and TinkerPop docs as references, but the Dart driver was not copied from one specific driver. The main motivation came from my own JanusGraph/TinkerPop usage and the friction of using js driver

  I also wanted to be clear about AI usage. The initial direction and architecture were mine. I used Claude Code to help expand the implementation once the shape of the driver was in place, and I used Codex for incremental  and adverserial review, bug hunting, and test-focused checks. I reviewed the code and behaviour myself throughout

  Your build/test points make sense. I agree the Docker setup should follow the existing non-Java GLV pattern instead of having a custom server setup. I’ll look at the Python setup and the shared `docker/gremlin-test-server/Dockerfile` approach.

  For feature tests, I agree that extracting Gremlin strings is only a starting point. It helped validate the HTTP + GraphBinary path, but the better long-term setup is generated native Dart traversal tests through the translator/generate.groovy flow. I may need some help there once the dev-list direction is clearer.

  For the two failing tests, I’ll keep checking them carefully:
  - for `g.inject([null,null]).conjoin("+")`, I’ll verify whether Dart is mishandling the empty string/null result path or whether the server response differs from what the feature expects.
  - for the sack/float case, I’ll verify that the client is preserving the `0.5f` literal and then adjust the test tolerance only
  if it is just normal floating point precision behavior.

  I’ll avoid large churn on the PR while the dev-list discussion is still open, but I’m happy to keep aligning the driver with the existing GLV conventions.

Cole-Greer:
Hi @kartikey321, thanks for the detailed reply.

I hope I didn't imply any slight with my questions on AI, most of the active maintains here use AI to various extents during their work. I'm actually quite curious how useful AI was in accelerating this sort of big standalone contribution, especially from the perspective of someone who is quite experienced with TinkerPop as a user, but not necessarily as a developer in the project.

I agree it's best not to try any big changes while the discussion is still playing out. Once that's settled, I'd be happy to help out with the translator and test infrastructure.

spmallette:
There hasn't been much further discussion on the dev list. I sense we left it in a place where we're open to the idea of accepting work like this, though we have a number of details to consider around how we do that.  As it stands, there was consensus that we would accept early developmental work by bringing back the concept of the Tinkubator - basically an incubator for ideas that we'd consider experimental with limited or no support. 

I think we need a bit of structure around how Tinkubator is designed to work, but i sense it will live in this repository. I think it's just a root directory called `tinkubator/`. As such, I'd expect that this PR would need to be adjusted for that directory. That said, I'm not clear what the bar is to accept this and other work to the Tinkubator. I pushed this commit to help define it:

https://github.com/apache/tinkerpop/commit/d9c493fb3adbb427bbc3bed69aecebb1b9cc6682

Other committers can review/adjust as they see fit - that's just my initial thoughts, but what I wrote there mostly seemed in line with where we drew consensus.

kartikey321:
Hi @spmallette @Cole-Greer,

Thank you for the continued discussion and for pushing the Tinkubator structure , it makes the requirements and deliverables clear.

Since my last update I've been working on the driver to directly address the points Cole raised. The changes are on a separate branch ([kartikey321/tinkerpop@dart-driver-java-parity](https://github.com/kartikey321/tinkerpop/tree/dart-driver-java-parity)) rather than pushed here, since the target path (`gremlin-dart/` vs `tinkubator/gremlin-dart/`) wasn't clear yet. Here's what's there:

**1. Docker/build setup**
The configuration now follows the Python pattern exactly - the test server uses the shared `docker/gremlin-test-server/Dockerfile` built from source rather than a bespoke image. A Maven `pom.xml` has also been added so the driver integrates into the standard TinkerPop build and test lifecycle (mirroring the Go module structure).

**2. Feature tests → native GraphTraversal**
The feature tests no longer submit raw Gremlin strings to the server. I implemented a Dart ANTLR visitor (`GremlinAntlrToDart`) that mirrors Java's `GremlinAntlrToJava` approach - the `Gremlin.g4` grammar is used at runtime to parse each scenario's string into a native `GraphTraversal`, which is then executed through the full Dart DSL. The feature tests now exercise the DSL, transport, and serialization end-to-end.

**3. New features added**
- **Transactions** - `tx()`, `begin()`, `commit()`, `rollback()` with session-based remote connection
- **Connection pooling**- configurable pool with lifecycle management
- **Multi-host Cluster** - round-robin load balancing across contact points
- **SSL/TLS** - configurable certificate and hostname verification options
- **Retry/reconnect** - configurable backoff with idempotency-safe retry interceptor
- **AWS SigV4 auth** - request signing support for Amazon Neptune
- **16 additional DSL steps** - including `disjunct`, `withoutStrategies`, and `char` GraphBinary type

**Other improvements:**
- 7 correctness bugs fixed (found through adversarial review)
- Unit test coverage above 95% (199 tests)
- `dart analyze` clean with `analysis_options.yaml`
- Integration test suite added, env-var gated (same pattern as Python/Go)
- Two failing feature tests still being investigated - the `conjoin("+")` escaping case and the float tolerance case

**On the Tinkubator path:** Given the dev list and Stephen's commit, it sounds like `tinkubator/gremlin-dart/` is the right home rather than a root-level `gremlin-dart/` module. Should I restructure and retarget this PR for that directory? Once it  is confirmed I can make the changes. I want to make sure the framing and structure match what the community expects before making that change.

Happy to move quickly on this once there's a clear direction. Thanks again.

spmallette:
> The feature tests no longer submit raw Gremlin strings to the server. I implemented a Dart ANTLR visitor (GremlinAntlrToDart) that mirrors Java's GremlinAntlrToJava approach

That's an interesting pattern. I'm curious what other committers think about that. That runs opposite to all language variants exception Java in that python, go, .net and javascript all use Java `Translator` infrastructure to dynamically generate a static compileble source file at build time. That helps test the translators themselves which is helpful. For the case of Tinkubator, though I'm not sure what that means since there wouldn't be a `DartTranslator` in `gremlin-core` to test. I suppose it could be written in groovy and then used as part of that script that generates the test source file, but there's clearly a decision to be made there. 

That said, I think there is value in each language variant knowing more about the ANTLR grammar, so the approach is interesting. We already have that in javascript and I could see that value elsewhere.  I guess I wonder if this a better and improved way to test?

> Given the dev list and Stephen's commit, it sounds like tinkubator/gremlin-dart/ is the right home rather than a root-level gremlin-dart/ module. Should I restructure and retarget this PR for that directory? 

From my perspective, that's the right place. I don't know what that means though from a maven reactor perspective.  I suspect we would want gremlin-dart to be a standalone maven project. I guess that means some replication in the pom.xml to do that. 

Thanks for sticking with this PR while we sort out exactly how to do this. In some ways your work here is breaking some new ground on how we do things so we appreciate the patience as things get figured out.

kartikey321:
Hi @spmallette @Cole-Greer,

Following up since it's been a couple of weeks since the last exchange. I want to make sure this doesn't drift — I'm ready to make the structural changes and am just waiting for a clear direction before touching anything.

Two things I'd appreciate guidance on:

**1. ANTLR vs. Translator+generate.groovy**
I understand both approaches. For the Tinkubator, my read is that the ANTLR runtime approach is acceptable since there's no `DartTranslator` in `gremlin-core` yet and the tests still exercise the full DSL/transport/serialization pipeline. But if the preference is to eventually wire into the generate.groovy model, I'm happy to do that — it just needs a `DartTranslator` written in Java first, which feels more like a graduation step. What does the community prefer for the Tinkubator stage?

**2. Maven structure**
Based on your last comment I'll make `tinkubator/gremlin-dart/` a standalone Maven project rather than adding it to the root reactor. Does that mean no entry in the root `pom.xml` at all, or should it still be referenced as a module?

Once these two points are settled I can restructure and submit everything under `tinkubator/gremlin-dart/` quickly. Thanks again for the patience in working through the process.

spmallette:
Regarding (1), ideally, i would say that you should create a `DartTranslator` in your project in Java that we could easily port to gremlin-core in the future whenever gremlin-dart leaves Tinkubator. Use the same package/namespace as whatever gremlin-core uses now and depend on gremlin-core for the interface.  In this way, the entire gremlin-dart setup will work as close as possible to the other variants. For (2), it is a standalone maven project that builds outside the reactor so there is no entry in the pom.xml.

kenhuuu:
Hi @kartikey321 , how are the changes coming along? Do you still have any open questions you need us to answer or is there anything else you need to help you along with this?

kartikey321:
Hi @kenhuuu , thanks for checking in.

I got pulled into work commitments and wasn’t able to keep this moving as quickly as I wanted.

The direction from Stephen is clear to me now: move this under `tinkubator/gremlin-dart/`, keep it as a standalone Maven project outside the root reactor, and add a Java `DartTranslator` in the Dart project so the feature tests can follow the same generated-test path as the other GLVs.

I’ve been working through that on the `dart-driver-java-parity` branch on my fork. I don’t have any open questions right now. I’m cleaning up the remaining local changes and will rerun the checks before updating this PR.

Thanks for the patience.

kenhuuu:
Take your time, there's no rush. I was just wondering if you still had any questions but it sounds like you don't.

PR #3456, opened 2026-06-11
GValue parity across all GLVs and parameterized feature tests

Cole-Greer:
## Summary

Completes the client-side GValue (named query-parameter) feature so it behaves consistently across all GLVs, adds parameterized feature-test coverage for the GLVs, and simplifies GValue name validation.

## Issues, gaps, and inconsistencies resolved

- **GValue was incomplete/inconsistent across GLVs** — JavaScript had no GValue at all; Python/.NET/Go diverged on validation, duplicate detection, and string representation.
- **Missing parameterized test coverage** — only Python ran the shared Gherkin corpus with parameters sent as GValue bindings; .NET, Go, and JS did not.
- **JS feature tests passed vacuously** — the docker run pointed at a non-existent corpus path, so 0 scenarios actually ran.
- **GLVs were stricter than Java** — they layered an identifier-pattern check on top of Java's rules; and the leading-_ reservation was obsolete (its motivating fallback was removed).
- **Non-idiomatic driver APIs** — Go used a constructor with unexported fields/getters;

## Details

- Aligned GValue validation/behavior across Python/.NET/Go; added the JS GValue; added a nested-GValue guard everywhere; added strongly-typed GValue<T> overloads to .NET for Java parity (interleaved beside their value-typed siblings).
- Added a PARAMETERIZE Gherkin variant to .NET, Go, and JS.
- Simplified name validation to match Java: only "no nested GValue" and (non-Java drivers) "non-null name" remain; removed the leading-_ reservation in Java and all GLVs.
- Replaced Go's constructor with an idiomatic exported-field struct.

.NET parameterize design: .NET is the only driver that parameterizes at the translator level (a second generated dotnet_parameterize translation set) rather than at runtime, because it's the only variant that is both statically typed and exposes strongly-typed GValue<T> overloads—so value-vs-GValue is a compile-
time overload choice. The other GLVs wrap values at runtime and reuse one translation set. The trade-off is higher regeneration cost for .NET, inherent to the typed API.

## Tips for Reviewers

On the surface, this is a big PR, but the bulk of the lines added come from 2 generated files which can be largely ignored. The .NET feature test design required adding a parameterized mode to the translators, such that the generated steps will include casts for GValues when present. This change alone doubled the size of `gremlin.cs`, and added thousands of lines of new cases to `translations.json`. This is a lot of volume from a relatively minor change.

The most consequential decisions in this work is the introduction of GValue step overloads in GraphTraversal, GraphTraversalSource, and __ in .NET, the complete introduction of the GValue concept to JS, and the change in GValue contruction and validation in Go. The most notable testing change is that all GLVs have now copied the feature test design from python, where the suite runs twice, once with all script parameters inlined as literals, and the second time with parameters preserved as GValues.

## Limitations

- Python's parameterized-V() integration test stays skipped under TINKERPOP-3126 (This is a grammar limitation/bug, not related to the GValue implementation).
- Go does not fail fast on a nested GValue (deliberate trade-off for the struct API; surfaces downstream instead).
- .NET multi-arg hasLabel mixing literal + variable args isn't parameterized (C# overload constraint; documented in code).


Cole-Greer:
VOTE +1

gremlin-dotnet/build/generate.groovy
```diff
@@ -132,6 +147,43 @@ radishGremlinFile.withWriter('UTF-8') { Writer writer ->
     }
     writer.writeLine('            };\n')
 
+    // Generate parameterized translations
```
spmallette, line 169: The PR description points this out: 

> .NET parameterize design: .NET is the only driver that parameterizes at the translator level (a second generated dotnet_parameterize translation set) rather than at runtime, because it's the only variant that is both statically typed and exposes strongly-typed GValue overloads—so value-vs-GValue is a compile-
time overload choice. 

I think i'd like to see comments here along with the code explaining the purpose. We will need to remind ourselves someday why this is the way it is, to say nothing of agents.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/DotNetTranslateVisitor.java
```diff
@@ -46,12 +46,20 @@
...
+
     public DotNetTranslateVisitor() {
         this("g");
     }
 
     public DotNetTranslateVisitor(final String graphTraversalSourceName) {
+        this(graphTraversalSourceName, false);
+    }
+
+    public DotNetTranslateVisitor(final String graphTraversalSourceName, final boolean parameterize) {
```
spmallette, line 73: This is the only visitor that needs this "parameterize" - it's weird enough that i think the constructors need javadoc to explain

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/Translator.java
```diff
@@ -48,6 +48,11 @@ public enum Translator {
      */
     DOTNET("DotNet", DotNetTranslateVisitor::new),
 
+    /**
+     * Translates to gremlin-dotnet with parameterized GValue wrapping for variables.
+     */
+    DOTNET_PARAMETERIZE("DotNetParameterize", name -> new DotNetTranslateVisitor(name, true)),
```
spmallette: as mentioned elsewhere, this is odd enough that it deserves more javadoc explaining usage/why. Probably extends to its `DOTNET` counterpart.

as a side note, this feels unfortunate to our `Translator` enum. from a discoverability perspective, the API looks asymetric. a user might wonder how they get parameters with `JAVASCRIPT` given this naming.  Thoughts on other things we could do?

spmallette: As a separate point, assuming there aren't other options, shouldn't there be some tests for this translator. I don't see any: https://github.com/apache/tinkerpop/blob/GValueFollowupTP4/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/language/translator/GremlinTranslatorTest.java

gremlin-dotnet/build/generate.groovy
```diff
@@ -132,6 +147,43 @@ radishGremlinFile.withWriter('UTF-8') { Writer writer ->
...
+    writer.writeLine(
+            '        private static readonly IDictionary<string, List<Func<GraphTraversalSource, IDictionary<string, object>,ITraversal>>> ParameterizedTranslations = \n' +
+            '            new Dictionary<string, List<Func<GraphTraversalSource, IDictionary<string, object>, ITraversal>>>\n' +
+            '            {')
+
+    gremlins.each { k,v ->
```
spmallette, line 175: i know this is just a internal utility script, but maybe we should encapsulate some of this duplicate logic?? i could envision an the accident of changing one but not the other and that regex seems complicated enough for one to get out of sync with the other.

CHANGELOG.asciidoc
```diff
@@ -59,6 +58,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Modified HTTP API to expect gremlin-lang strings for parameters and update all GLVs to send requests in new format.
 * Added string parameter parsing to `GremlinServer` to prevent traversal injection and excessive nesting depths.
 * Modified all GLVs to detect unsupported types in `GremlinLang` and throw consistent error for that case.
+* Relaxed `GValue` name validation by removing the reserved leading underscore restriction.
```
spmallette, line 90: didn't this PR do a bit more than just this?

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1962,8 +1852,17 @@ Promise.all([
 [[gremlin-javascript-gvalue]]
 === GValue Parameterization
 
-IMPORTANT: 4.0.0-beta.2 Release - `GValue` parameterization is not yet implemented for the JavaScript driver. This
-functionality is planned for a future release.
+A `GValue` is an encapsulation of a parameter name and value. A <<traversal-parameterization,subset of gremlin steps>>
```
spmallette, line 2216: > JavaScript had no GValue at all; Python/.NET/Go diverged on validation, duplicate detection, and string representation.

shouldn't we have more docs then? like "GValue Parameterization" sections for all variants?

xiazcy:

CHANGELOG.asciidoc
Cole-Greer: This PR did more, but I'm writing the changelog from the perspective of an upgrade from 3.8. This is really the only notable changes from GValue in 3.8 Java. The other GLVs didn't have GValue before as it was tied into bytecode bindings. From the perspective of the 3.8 upgrade, there isn't really changes to GValue in the GLVs, it's a fresh start which is consistent with the existing Java pattern.

I can add a new entry detailing that GValue has been extended to all GLVs, but I don't think the modifications warrant changelog entries here.

docs/src/reference/gremlin-variants.asciidoc
Cole-Greer: Javascript was the only GLV which was missing the "GValue Parameterization" section altogether, other than the go GValue construction edit, there were no changes needed to the existing docs here.

One slight exception is some GLVs include 2 examples (has() and mergeV()), while others only have the has() example. I'll add the merge example everywhere for consistency.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/Translator.java
Cole-Greer: I opted to remove the enum for parameterized dotnet. It's a niche enough use case which really only exists for dotnet feature test generation. I don't think there's a need to confuse the enums with this new option now, generate.groovy can manually construct the parameterized translator.

We can revisit the enum later if demand arises.

PR #3458, opened 2026-06-12
Allow Steps to Accept Traversal

xiazcy:
## Summary

Allow child traversals as arguments to `has()`, `is()`, `V()`, `E()`, `property()`, `where(P)`, and all `P`/`TextP` predicates. The child traversal is evaluated per-traverser at runtime, enabling dynamic value resolution.

**JIRA:**

- [TINKERPOP-2586](https://issues.apache.org/jira/browse/TINKERPOP-2586): Traversal objects inside `within()` silently return empty instead of executing or erroring - **fully addressed** (traversals are now executed)
- [TINKERPOP-2777](https://issues.apache.org/jira/browse/TINKERPOP-2777): `V()` cannot accept traversals like `select()` for data-driven lookups - **fully addressed**
- [TINKERPOP-3005](https://issues.apache.org/jira/browse/TINKERPOP-3005): `property()` should accept a traversal producing a Map - **fully addressed**
- [TINKERPOP-1463](https://issues.apache.org/jira/browse/TINKERPOP-1463): `has(key, traversal)` doesn't work as expected - **partially addressed** (dynamic filtering works; runtime folding into index lookups is deferred as a future optimization)

### What it enables

```
// with Modern graph
// Data-driven vertex lookup (TINKERPOP-2777)
gremlin> g.inject(['1','2','3']).as('a').V(__.select('a')).values('name')
==>marko
==>vadas
==>lop

// Dynamic filtering - find people older than marko
gremlin> g.V().has('age', P.gt(__.V(1).values('age'))).values('name')
==>josh
==>peter

// Computed property mutation (TINKERPOP-3005) - set multiple properties from a Map
gremlin> g.V(4).property(__.V(1).project('friendCount','createdSoftware').by(__.out('knows').count()).by(__.out('created').values('name')))
==>v[4]
gremlin> g.V(4).valueMap('friendCount','createdSoftware')
==>[friendCount:[2],createdSoftware:[lop]]

// with AirRoutes graph
// Find airports where the city name ends with the airport code (lowercased)
gremlin> g.V().has("city", TextP.endingWith(__.values("code").toLower())).limit(10).values("code","city").fold()
==>[CUN, Cancun, ORK, Cork, OME, Nome, IEV, Kiev, AJU, Aracaju, IRA, Kirakira, OHE, Mohe, RUA, Arua, VAS, Sivas, AZD, Yazd]
```

---

## Review Guide

This is a large change (68 files, ~7k lines) but structured in layers. **Review bottom-up:**

| # | Commit | Layer | What to look for |
|---|--------|-------|-----------------|
| 1 | `5f37d11` | Predicate resolution | Core: `P.resolve()`, empty-result semantics, `AndP` short-circuit, `ConnectiveP`/`NotP` delegation |
| 2 | `6a547d2` | Step implementations | How steps use `P.resolve()`. `HasStep` single path, `GraphStep` idTraversal, `AddPropertyStep` mapForm, `AcceptsChildPredicateTraversal` marker |
| 3 | `e80423a` | Grammar | `Gremlin.g4` `nestedTraversal` alternatives. Visitors delegate to API (not `addStep()`) |
| 4 | `1b21d15` | GLVs | `within()`/`without()` serialization. .NET typed overloads |
| 5 | `793e7a8` | Tests + docs | Cucumber `.feature` files define behavioral contract. Reference doc updates |
| 6 | `400f5ae` | Cleanup | Dead code removal from HasContainer unification. Test quality fixes |

### Architecture (single resolution path)

```
                    has("age", __.V(1).values("age"))
                              │
                    DSL wraps: has("age", P.eq(traversal))
                              │
                              ▼
                    HasContainer(key, P.eq(traversal))
                              │
              ┌───────────────┴───────────────┐
              │ HasStep.resolvePredicate()     │
              │   P.resolve(traverser)         │
              │   hc.test(element)             │
              └────────────────────────────────┘
```

### Key decisions to scrutinize

1. **Single resolution path** - At the DSL level, `has(key, traversal)` wraps the traversal in `P.eq(traversal)` before creating the `HasContainer`. This means HasStep only has one code path for traversal resolution: call `P.resolve(traverser)` on the predicate, then `hc.test(element)`. The GremlinLang serialization still records the original `has(key, traversal)` form (verified by round-trip tests). This is a new 4.0.0 feature, so no existing provider code is affected. Providers implementing custom strategies must check `HasContainer.hasTraversal()` before folding into index lookups.

2. **Empty-result semantics** - When a child traversal produces zero results, behavior depends on predicate type. Collection predicates (`within`/`without`) resolve to an empty collection and delegate to `Contains.test()` which applies correct set-theoretic logic: `within([])` is always false, `without([])` is always true. Scalar predicates (`eq`/`gt`/`lt`/etc.) set `resolvedEmpty=true` and steps short-circuit to false since there is no meaningful comparison value. A null result (traversal produces `null`) is distinct from an empty result: it resolves normally and `null` is used as the comparison value, consistent with existing `P.eq(null)` semantics.

3. **AndP short-circuits, OrP doesn't** - `AndP.resolve()` overrides `ConnectiveP.resolve()` to stop resolving children at the first scalar predicate that resolves empty. This is sound because `and(unsatisfiable, X)` is always false regardless of X. `OrP` does NOT short-circuit: a non-empty resolution does not guarantee the predicate will pass `test()`, so later children must already be resolved when their turn comes. Note: short-circuiting means side-effecting child traversals (e.g., ones using `aggregate`) in later AndP operands may not execute. This matches standard boolean short-circuit semantics.

4. **`P.resolve()` mutates in place** - `resolve()` overwrites `literals`, `isCollection`, and `resolvedEmpty` fields on the `P` instance. In OLTP this is safe since resolve and test happen sequentially per traverser. In OLAP, `HasStep.clone()` deep-clones predicates per worker, providing thread isolation. The mutation pattern is a known limitation documented for future improvement (return an immutable resolved value instead). Predicate trees of arbitrary depth (AndP containing OrP containing NotP, etc.) are handled by recursive resolution and recursive cloning.

5. **Mutation guard** - `ChildTraversalValidator` rejects any child traversal containing a step that implements the `Mutating` interface (addV, addE, drop, property). This runs at construction time (DSL methods and P factory methods) and again at strategy time via `ChildTraversalVerificationStrategy`. Aggregating side-effects (`aggregate`, `store`, `group`) are NOT blocked because they don't modify graph state. Whether they should be restricted is a question for reviewers.

6. **Provider contract** - `HasContainer.hasTraversal()` is the single check point providers must use before folding. Providers that fold HasContainers into index-backed steps without this check will attempt to use an unresolved `P` value (returns `null` before `resolve()` is called) as an index key, producing wrong results. `TinkerGraphStepStrategy` demonstrates the pattern: skip traversal-bearing HasSteps with `continue` so that subsequent literal HasSteps can still be folded. Providers with custom `GraphStep` replacements must also copy the `idTraversal` field during strategy replacement, otherwise `V(traversal)` falls through to returning all vertices.

7. **`choose()` restriction** - `choose(P)` and `choose().option(P, ...)` reject traversal-bearing predicates at construction time. This is a deliberate scope decision. `PredicateTraversal` (which evaluates option-key predicates in `BranchStep`) has access to the traverser and could technically call `P.resolve()` before `P.test()`, but extending the branching infrastructure is out of scope for this feature. Users needing dynamic branching can use the `choose(Traversal, ...)` form where the branch predicate is a full traversal (e.g., `choose(__.is(P.gt(traversal)), trueChoice, falseChoice)`). This works because `IsStep` handles resolution normally. Restricting `is()` when used inside `choose()`'s branch traversal is not feasible since `choose()` receives it as an opaque `Traversal.Admin` and does not inspect internal steps.

### What's NOT in this PR

- `P.resolve()` thread-safety refactoring (future)
- `instanceof DefaultGraphTraversal` cleanup (deferred)
- Runtime index folding for deterministic child traversals
- `FilterRankingStrategy` cost awareness

### Compatibility

- Additive overloads only. No existing step or predicate signature changed.
- Serialized GremlinLang form is unchanged: `has(key, traversal)` still serializes as `has(key, traversal)`, verified by `GremlinLangTraversalRoundTripTest` and `GremlinQueryParserTraversalTest`.

---

## Testing

All `gremlin-core` unit tests pass (9132, 0 failures) and the full TinkerGraph Gherkin suite passes (2172 scenarios, 0 failures) against the current `master`.

| Category | Coverage |
|----------|----------|
| Behavioral contract | `HasTraversal.feature`, `IsTraversal.feature`, `WhereTraversal.feature`, `VETraversal.feature`, `PropertyTraversal.feature`, `ChildTraversalVerification.feature` |
| Internal mechanics | `PTraversalTest`, `CloneIndependenceTraversalTest`, `HasContainerTest`, `ChildTraversalValidatorTest`, `ChildTraversalVerificationStrategyTest` |
| Grammar + serialization | `GremlinQueryParserTraversalTest`, `GremlinLangTraversalRoundTripTest` |
| Strategy correctness | `TinkerGraphStepStrategyTraversalTest` |

## Checklist

- [x] CHANGELOG entry added
- [x] Reference documentation updated (`the-traversal.asciidoc`)
- [x] Upgrade documentation added (`release-4.x.x.asciidoc`)
- [x] All GLVs updated (Python, .NET, Go, JavaScript)
- [x] Full `mvn clean install` passes
- [x] No new external dependencies introduced

---

VOTE +1


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -49,10 +56,23 @@ public class P<V> implements Predicate<V>, Serializable, Cloneable {
...
 
     public P(final PBiPredicate<V, V> biPredicate, final V value) {
-        setValue(value);
         this.biPredicate = biPredicate;
+        // If the value is a DefaultGraphTraversal (the type created by __.xxx() anonymous traversals),
+        // treat it as a child traversal rather than a literal. This handles the case where Java's
+        // overload resolution picks P(BiPredicate, V) instead of P(BiPredicate, Traversal) when
+        // the caller passes a GraphTraversal. We specifically check for DefaultGraphTraversal
+        // rather than Traversal to avoid catching internal traversal types like ConstantTraversal,
+        // ValueTraversal, and IdentityTraversal which are used as literal values in P.
```
Cole-Greer: I'm not sure I agree with handling Traversals other than DefaultGraphTraversal as literal values. I can't imagine a case where I would want to `P.eq()` a `ConstantTraversal`, and have that mean `constantTraversal.equals(myValue)` instead of `constantTraversal().next().equals(myValue)` like any regular `GraphTraversal`. That said, I'm not sure when I would ever want to pass a ConstantTraversal into a predicate.

Even if the choice is kept to exclude most of the oddball traversals, I would think testing for `GraphTraversal` instead of `DefaultGraphTraversal` would be more reasonable.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -49,10 +56,23 @@ public class P<V> implements Predicate<V>, Serializable, Cloneable {
...
+        if (value instanceof org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.DefaultGraphTraversal) {
```
Cole-Greer: I think this logic should be folded into `setValue()` instead of kept exclusively in the constructor. I don't like the implication that `new P(Compare.eq, __.V(5))` won't behave the same as `new P(Compare.eq, 10).setValue(__.V(5))`

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -49,10 +56,23 @@ public class P<V> implements Predicate<V>, Serializable, Cloneable {
     protected Map<String, V> variables = new HashMap<>();
     protected Collection<V> literals = Collections.EMPTY_LIST;
     private boolean isCollection = false;
+    private boolean resolvedEmpty = false;
+    private Traversal.Admin<?, ?> traversalValue;
+    private List<Traversal.Admin<?, ?>> traversalValues;
 
     public P(final PBiPredicate<V, V> biPredicate, final V value) {
```
Cole-Greer, line 59: I think we should be careful to avoid the assumption that `V value` here can't also be a `List<Traversal>` simply because that explicit overload is present, or because the type parameter theoretically forbids it. This class really abuses its generics, and relies heavily on type erasure to even function. It's not exactly type safe, and generally best to treat `V` as `Object` in most instances here.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -224,6 +283,187 @@ public Set<GValue<?>> getGValues() {
...
+        trav.addStart(split);
+    }
+
+    //////////////// predicate traversal utilities
+
+    /**
+     * Recursively integrates all child traversals found in the predicate tree into the given parent step.
+     * Handles {@link ConnectiveP} (recurses into children) and {@link NotP} (recurses into wrapped predicate).
+     */
+    public static void integrateTraversals(final P<?> p, final TraversalParent parent) {
```
Cole-Greer: This feels like a bit of an odd API for `P` to surface, we already have `collectTraversals()` below. Couldn't the parent steps use this to integrate all of their children directly, instead of delegating this through P?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -49,10 +56,23 @@ public class P<V> implements Predicate<V>, Serializable, Cloneable {
     protected Map<String, V> variables = new HashMap<>();
     protected Collection<V> literals = Collections.EMPTY_LIST;
     private boolean isCollection = false;
+    private boolean resolvedEmpty = false;
+    private Traversal.Admin<?, ?> traversalValue;
+    private List<Traversal.Admin<?, ?>> traversalValues;
```
Cole-Greer: Nit: The result types of these traversals should arguably be `V` throughout this class. I expect that would help with a lot of the `@SuppressWarnings("unchecked")` annotations throughout this class. There's probably an argument that the generics are so broken in this class already that it hardly matters, but it would be more proper this way.
```suggestion
    private Traversal.Admin<?, V> traversalValue;
    private List<Traversal.Admin<?, V>> traversalValues;
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -224,6 +283,187 @@ public Set<GValue<?>> getGValues() {
...
+                    this.resolvedEmpty = false;
+                    this.isCollection = true;
+                } else {
+                    this.resolvedEmpty = true;
+                    this.isCollection = false;
+                }
+            } else {
+                this.resolvedEmpty = false;
+                final Object firstResult = trav.next();
+                if (this.biPredicate instanceof Contains) {
```
Cole-Greer: Why can't we go through the `setValue()` path for `Contains` bipredicates?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -224,6 +283,187 @@ public Set<GValue<?>> getGValues() {
...
+                CloseableIterator.closeIterator(trav);
+            }
+        }
+
+        // Multi-traversal resolution is only valid for collection predicates (within/without). An empty
+        // combined result is a legitimate empty set, so resolve to an empty collection and let Contains.test()
+        // apply the correct semantics (within(empty) -> false, without(empty) -> true) instead of short-circuiting.
+        this.resolvedEmpty = false;
+        this.isCollection = true;
+        this.literals = allResults.isEmpty()
```
Cole-Greer: Nit: Do we need the ternary here? Can we just directly feed `allResults` into `setValue()` or `this.literals`?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -453,6 +697,106 @@ public static <V> P<V> without(final Collection<V> value) {
         return new P(Contains.without, value);
     }
 
+    /**
```
Cole-Greer: Nit on organization, it would be nice to interleave these overloads with their matching pairs (all the eq's together, all the gt's together...)

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -453,6 +697,106 @@ public static <V> P<V> without(final Collection<V> value) {
...
+     * Determines if values are equal using a child traversal resolved at runtime.
+     *
+     * @since 4.0.0
+     */
+    public static <V> P<V> eq(final Traversal<?, ?> traversalValue) {
```
Cole-Greer: We shouldn't assume that traversals will never end up in the original `eq(V value)` overload, it's essentially `eq(Object value)` after erasure and java can be unreliable when selecting overloads sometimes. We ran into that problem quite a bit with GValue, the only value the GValue overloads really gave was making the generics behave a bit nicer in the average case. There's plenty of times where Java chooses to ignore the narrow overloads.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -453,6 +697,106 @@ public static <V> P<V> without(final Collection<V> value) {
...
+        return new P(Contains.without, traversalValue.asAdmin());
+    }
+
+    /**
+     * Determines if a value is not within the union of results from multiple child traversals resolved at runtime.
+     * Each traversal is evaluated independently and results are combined into a single collection.
+     *
+     * @since 4.0.0
+     */
+    public static <V> P<V> without(final Traversal<?, ?> first, final Traversal<?, ?> second, final Traversal<?, ?>... more) {
```
Cole-Greer: Why do we need special cases for these extra first, second, more overloads?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/ChildTraversalValidator.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private ChildTraversalValidator() {
+    }
+
+    /**
+     * Validates that a child traversal contains no {@link Mutating} steps at any nesting depth.
+     * Throws {@link IllegalArgumentException} if one is found.
+     */
+    public static void validate(final Traversal.Admin<?, ?> child) {
+        for (final Step<?, ?> step : child.getSteps()) {
```
Cole-Greer: Nit: this should probably just use https://github.com/apache/tinkerpop/blob/24df1d2b89e3b7d10c5f958561d05371ae1c9a75/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/TraversalHelper.java#L428

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -453,6 +697,106 @@ public static <V> P<V> without(final Collection<V> value) {
...
+        return new P(Compare.neq, traversalValue.asAdmin());
+    }
+
+    /**
+     * Determines if a value is less than another using a child traversal resolved at runtime.
+     *
+     * @since 4.0.0
+     */
+    public static <V> P<V> lt(final Traversal<?, ?> traversalValue) {
+        ChildTraversalValidator.validate(traversalValue.asAdmin());
```
Cole-Greer: It might be better to pull this validation into the P constructor or even setValue(), in most cases P's are built via these static methods but not always.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -2112,6 +2119,24 @@ the key,value pairs for those vertices.
 <9> Property key is always stored as `String` and therefore an equality check with `null` will produce no result.
 <10> An example of using `has()` with regular expression predicate.
 
+The traversal-accepting forms of `has()` allow for dynamic property comparisons. Rather than providing a literal value,
+a child traversal is supplied and its first result is used as the comparison value. This follows the same first-result
+semantics as `by(traversal)`. The child traversal must be read-only - mutating steps are rejected.
```
spmallette: > The child traversal must be read-only - mutating steps are rejected.

Don't really like dashes in our docs this way, how about:  "The child traversal cannot include mutating steps like `addV()` or `mergeE()`. If these steps are present, they will be rejected at runtime with a `XYZException`."

docs/src/reference/the-traversal.asciidoc
```diff
@@ -2467,6 +2492,18 @@ g.V().where(__.in('created').values('age').
 <2> Find projects having two or more contributors.
 <3> Find projects whose contributors average age is between 30 and 35.
 
+The `is()` step also supports predicates that contain traversal arguments for dynamic threshold comparison.
```
spmallette: any reason to not just fold this back into the previous examples? i'm not sure it needs to stand out as a separate thing.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -3989,6 +4026,21 @@ g.addV().property(set, null)
 <7> The label value can be specified as a property only at the time a vertex is added and if one is not specified in the addV()
 <8> If you pass a `null` value for the Map this will be treated as a no-op and the input will be returned
 
+The `property()` step also accepts a traversal that produces a `Map` of key-value pairs to set as properties. The
```
spmallette: again, don't like the dash this way in our docs:

> The child traversal must be read-only - mutating steps are rejected.

you could use the other verbiage i had or some variation of it.

> If the traversal does not produce a `Map`, the result is rejected.

rejected when and in with what exception?

docs/src/reference/the-traversal.asciidoc
```diff
@@ -3989,6 +4026,21 @@ g.addV().property(set, null)
...
+child traversal must be read-only - mutating steps are rejected. If the traversal does not produce a `Map`, the result
+is rejected. Each entry in the resulting Map becomes a separate property on the element, allowing multiple properties
+to be set in a single step.
+
+[gremlin-groovy,modern]
+----
+g.V(4).property(__.V(1).project('friendCount','createdSoftware').by(__.out('knows').count()).by(__.out('created').values('name'))) <1>
```
spmallette: please format the query to make this easier to read.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -3989,6 +4026,21 @@ g.addV().property(set, null)
...
+g.V(4).valueMap('friendCount','createdSoftware')
```
spmallette: maybe better to show the whole vertex with all properties to demonstrate we augmented an existing vertex that had properteis?

docs/src/reference/the-traversal.asciidoc
```diff
@@ -3989,6 +4026,21 @@ g.addV().property(set, null)
...
+----
+
+<1> Set two properties on vertex 4 (josh) from a Map produced by a child traversal. The `project()` step builds a Map
```
spmallette: if you get a better multi-line formatting as i mentioned above, then this description can be broken into smaller bits for folks to understand more easily

docs/src/reference/the-traversal.asciidoc
```diff
@@ -5187,6 +5239,19 @@ g.V().has('name', within('marko', 'vadas', 'josh')).as('person').
 <1> Normally the `V()`-step will iterate over all vertices. However, graph strategies can fold ``HasContainer``'s into a `GraphStep` to allow index lookups.
 <2> Whether the graph system provider supports mid-traversal `V()` index lookups or not can easily be determined by inspecting the `toString()` output of the iterated traversal. If `has` conditions were folded into the `V()`-step, an index - if one exists - will be used.
 
+The `V()` step also accepts a traversal argument. The child traversal is evaluated and its results are used as the
```
spmallette: again, i feel like this doesn't need to be an "also" in how this step works. just integrate it into the other examples and content more seamlessly.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -5399,6 +5464,18 @@ g.V().as('a').both().both().as('b').
 <8> Marko is younger than josh, but josh knows someone equal in age to marko (which is marko).
 <9> The "age" property is not <<by-step,productive>> for all vertices and therefore those values are filtered.
 
+The `where()` step also supports predicates that contain traversal arguments. When a predicate contains a child
```
spmallette: again, as with other comments, let's integrate this more seamlessly into the content. Also, perhaps `where` is a case where we could come up with a more complex example since it's a more advanced step.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -100,6 +69,32 @@ GLVs (including the Java driver). This aligns with the embedded graph transactio
 partial work is discarded if the user forgets to call `commit()`. In Java (both remote and embedded mode), the behavior
 can still be overridden via `tx.onClose(Transaction.CLOSE_BEHAVIOR.COMMIT)`. The non-Java GLVs do not support
 configuring close behavior and always rollback.
+==== Traversal-Accepting Steps
```
spmallette: Need a catchier title here. "Expanding Dynamic Arguments"??

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -100,6 +69,32 @@ GLVs (including the Java driver). This aligns with the embedded graph transactio
...
+
+Steps and predicates that previously only accepted literal values now accept child traversals resolved per-traverser
```
spmallette: Folks have been waiting on this feature for a long time. It needs a strong introduction with some history about the challenges folks had with the old way to do some of this stuff. Show some comparisons demonstrating the old way versus the ease of the new way. 

Stuff like the following is instructional details and doesn't really belong here: 

> Child traversals take only the first result (consistent with `by(traversal)` semantics). For `within()`/`without()`,
use `fold()` to collect multiple values into a list.

i think we should also be telling users in this section what patterns this feature allows them to replace to make their code perform better, read better, etc.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -100,6 +69,32 @@ GLVs (including the Java driver). This aligns with the embedded graph transactio
...
+// Multi-source filtering with within()
+g.V().has("name", P.within(__.V(1).out("knows").values("name").fold(), __.constant("peter")))
+----
+
+Child traversals take only the first result (consistent with `by(traversal)` semantics). For `within()`/`without()`,
+use `fold()` to collect multiple values into a list.
+
+All child traversals must be read-only. Mutating steps (`addV`, `addE`, `drop`, `property`) inside child traversals
+are rejected at construction time with `IllegalArgumentException`. A `ChildTraversalVerificationStrategy` provides
+additional safety at strategy time.
```
spmallette: Aren't there some JIRAs for this to reference?

spmallette:
There are some docs here that cover the feature, but what I don't see are updates to existing docs that use old patterns that are no longer relevant and that we should no longer promote.  Has that kind of search been done?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -414,6 +415,23 @@ public default GraphTraversal<S, Vertex> V(final Object... vertexIdsOrElements)
         return this.asAdmin().addStep(step);
     }
 
+    /**
+     * A {@code V} step that accepts a child traversal whose results are used as vertex IDs for lookup.
+     * This form is only valid as a mid-traversal step; using it as a start step will throw an
+     * {@link IllegalStateException} at runtime because there is no traverser context to evaluate the child traversal.
```
xiazcy:
```suggestion
     * A {@code V} step that accepts a child traversal whose results are used as vertex IDs for lookup.
     * This form works both as a mid-traversal step (where the current traverser seeds the child traversal)
     * and as a start step via {@link GraphTraversalSource#V(Traversal)} (where a synthetic traverser is
     * generated to seed the child traversal, consistent with {@code mergeV(traversal)} behavior).traversal.
```
Note stale javadoc

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/dsl/graph/GraphTraversal.java
```diff
@@ -437,6 +455,23 @@ public default GraphTraversal<S, Edge> E(final Object... edgeIdsOrElements) {
         return this.asAdmin().addStep(step);
     }
 
+    /**
+     * A {@code E} step that accepts a child traversal whose results are used as edge IDs for lookup.
+     * This form is only valid as a mid-traversal step; using it as a start step will throw an
+     * {@link IllegalStateException} at runtime because there is no traverser context to evaluate the child traversal.
```
xiazcy:
```suggestion
     * A {@code E} step that accepts a child traversal whose results are used as edge IDs for lookup.
     * This form works both as a mid-traversal step (where the current traverser seeds the child traversal)
     * and as a start step via {@link GraphTraversalSource#E(Traversal)} (where a synthetic traverser is
     * generated to seed the child traversal, consistent with {@code mergeE(traversal)} behavior).
```
Note stale javadoc

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
Cole-Greer: I experimented with this a bit, and it really just moving problems around instead of solving them. The real issue is the pre-existing extent to which the generics are abused in P. I think a larger P refactor would be beneficial at some point, but definitely beyond the scope of this work.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
Cole-Greer: a minimum of a `first, more...` overload is necessary to disambiguate the zero arg calls to these methods.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/GremlinLangTraversalRoundTripTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+import static org.apache.tinkerpop.gremlin.process.traversal.AnonymousTraversalSource.traversal;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests that GremlinLang serialization round-trips preserve traversal arguments.
+ * For any step containing a child traversal argument, serializing to GremlinLang and parsing back
+ * produces a structurally equivalent traversal.
+ */
+public class GremlinLangTraversalRoundTripTest {
```
spmallette, line 45: Good tests, though i can't help wondering if this is just agent generated stuff that is either:

1. already being covered elsewhere, meaning these tests should be there or
2. covered by other processes making this set of unit tests kinda redundant or
3. we don't have enough round-trip tests to cover so many other interesting cases - meaning we should improve this test dramatically to cover lots of cases.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/PTraversalTest.java
```diff
@@ -0,0 +1,531 @@
...
+
+    /**
+     * Covers the empty-result semantics of collection (within/without) versus scalar predicates and the
+     * behavior of connective predicates (AndP/OrP) and NotP when their operands carry child traversals.
+     * <p>
+     * Regression coverage for: within(empty) -> false, without(empty) -> true (collection predicates resolve
+     * to an empty collection rather than short-circuiting), scalar predicates remain resolved-empty, and the
+     * shared-state safety of resolving the same predicate across many traversers sequentially.
+     */
+    public static class EmptyResolutionAndConnectiveTest {
```
spmallette: I think there's room for more complex test cases here with embedded ands/ors. come up with some wild scenarios

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/InlineFilterStrategy.java
```diff
@@ -114,6 +114,10 @@ step instanceof OrStep && processOrStep((OrStep) step, traversal) ||
     ///////////////////////////
 
     private static boolean processHasStep(final HasStep<?> step, final Traversal.Admin<?, ?> traversal) {
+        if (step.getHasContainers().stream().anyMatch(HasContainer::hasTraversal)) return false;
```
spmallette, line 123: we changed this strategy, but i don't see any unit tests here: https://github.com/apache/tinkerpop/blob/master/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/InlineFilterStrategyTest.java

i think it would be good to add some cases demonstrating these shortcircuit points. i think i'd like more inline code comments around the short circuiting.  Update to the class javadoc also seems like a good idea.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/branch/ChooseStep.java
```diff
@@ -122,6 +123,10 @@ public void addChildOption(final M pickToken, final Traversal.Admin<S, E> traver
             }
         } else if (pickToken instanceof Traversal) {
             throw new IllegalArgumentException("Traversal is not allowed as a Pick token for choose().option()");
+        } else if (pickToken instanceof P && ((P<?>) pickToken).hasTraversal()) {
```
spmallette, line 126: I think javadoc on `choose` for `GraphTraversal` and `__` should probably make it clear we don't support this form.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java
```diff
@@ -116,21 +147,50 @@ public V getValue() {
...
-            isCollection = false;
         } else if (value instanceof Collection) {
-            isCollection = true;
-            if (((Collection<?>) value).stream().anyMatch(v -> v instanceof GValue)) {
+            final Collection<?> coll = (Collection<?>) value;
+            if (!coll.isEmpty() && coll.stream().anyMatch(v -> v instanceof GraphTraversal)) {
+                if (!coll.stream().allMatch(v -> v instanceof GraphTraversal)) {
+                    throw new IllegalArgumentException(
+                            "Cannot mix traversals and literal values. " +
+                            "Use __.constant(val) to wrap all values as traversals.");
```
spmallette: > "Use __.constant(val) to wrap all values as traversals."

maybe better as:

> "Use __.constant(val) to wrap each value as a traversals."

that's more clear, no?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/util/ReadOnlyChildValidator.java
```diff
@@ -0,0 +1,45 @@
...
+    private ChildTraversalValidator() {
+    }
+
+    /**
+     * Validates that a child traversal contains no {@link Mutating} steps at any nesting depth.
+     * Throws {@link IllegalArgumentException} if one is found.
+     */
+    public static void validate(final Traversal.Admin<?, ?> child) {
+        if (TraversalHelper.hasStepOfAssignableClassRecursively(Mutating.class, child)) {
+            throw new IllegalArgumentException("Child traversal contains a mutating step. " +
```
spmallette: would we help folks by telling them the `Mutating` step we found and in what parent?

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/PTraversalTest.java
```diff
@@ -0,0 +1,531 @@
...
+import static org.hamcrest.CoreMatchers.nullValue;
+import static org.hamcrest.CoreMatchers.sameInstance;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+/**
+ * Tests for traversal-aware behavior in {@link P}, covering traversal detection accuracy
+ * and single-value predicate rejection of multiple traversal results.
+ */
+@RunWith(Enclosed.class)
+public class PTraversalTest {
```
spmallette: While we have a good body of tests here we have none in `PTest` that carry `Traversal`. I think that if you look at `PTest` we cover cases like `clone()` and `negate` operations which I don't find coverage for here. That feels like a testing gap.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/AcceptsChildPredicateTraversal.java
```diff
@@ -0,0 +1,34 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.tinkerpop.gremlin.process.traversal.step;
+
+/**
+ * Marker interface for steps that accept user-supplied child traversals embedded as predicate or lookup
```
spmallette: > Marker interface for steps that accept user-supplied child traversals embedded as predicate or lookup arguments

I'm not sure that reads right - " Marker interface for steps that accept child traversals used to provide dynamic values to {@link P} or to the step itself"

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/AcceptsChildPredicateTraversal.java
```diff
@@ -0,0 +1,34 @@
...
+ * {@code V(traversal)}, or {@code property(traversal)}).
+ * <p>
+ * The {@link org.apache.tinkerpop.gremlin.process.traversal.strategy.verification.ChildTraversalVerificationStrategy}
+ * validates the local children of every step implementing this interface, rejecting child traversals that
+ * contain mutating steps. New steps that accept such traversals should implement this interface so they are
+ * validated automatically rather than relying on a hardcoded step list.
+ *
+ * @since 4.0.0
+ */
+public interface AcceptsChildPredicateTraversal {
```
spmallette: I don't really like this name: `AcceptsChildPredicateTraversal `. It's plainly descriptive but I'd like something more functional. How about: `DynamicSelecting`? Meaning the step is one that allows for dynamic selection of a value as an argument. This differs from a step that happens to already take `Traversal` in that the implication of "Selecting" is that it is not allowing just any `Traversal` - its is for selection only (i.e. no mutations). It's late...maybe a better name is out there along those lines, but `AcceptsChildPredicateTraversal` is not it imo.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -101,6 +70,51 @@ partial work is discarded if the user forgets to call `commit()`. In Java (both
 can still be overridden via `tx.onClose(Transaction.CLOSE_BEHAVIOR.COMMIT)`. The non-Java GLVs do not support
 configuring close behavior and always rollback.
 
+==== Expanding Dynamic Arguments to Additional Steps
+
+Prior to 4.0, comparing a traverser's value against a dynamically computed reference required verbose workarounds with
+`where()`, `select()`, and labeled steps. For example, finding all people older than "marko" previously required:
+
+[source,groovy]
```
spmallette, line 411: why are none of the examples actually executed in the Console so that we can see the results?

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -101,6 +70,51 @@ partial work is discarded if the user forgets to call `commit()`. In Java (both
...
+----
+// Old way: label marko's age, compare via where()
+g.V(1).values("age").as("markosAge").
+  V().hasLabel("person").as("p").
+  values("age").as("otherAge").
+  where("otherAge", gt("markosAge")).
+  select("p").values("name")
+----
+
+This pattern was unintuitive and could not easily compose with other predicates. With traversal-accepting arguments,
```
spmallette: "unintuitive" - better to say, "This pattern required several coordinated step labels and `where()` compartison" and let the reader decide if it is unintuitive or not.

Cole-Greer:
VOTE +1

docs/src/reference/the-traversal.asciidoc
```diff
@@ -2112,6 +2119,25 @@ the key,value pairs for those vertices.
 <9> Property key is always stored as `String` and therefore an equality check with `null` will produce no result.
 <10> An example of using `has()` with regular expression predicate.
 
+In addition to literal values, `has()` accepts child traversals for dynamic comparisons. The child traversal is
+evaluated per-traverser and its first result becomes the comparison value, consistent with `by(traversal)` semantics.
+The child traversal cannot include mutating steps like `addV()` or `property()`; if present, they are rejected with
```
spmallette: > The child traversal cannot include mutating steps like `addV()` or `property()`; if present, they are rejected with an `IllegalArgumentException`.

No semi-colons. Use a proper transition word or two sentences.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -101,6 +70,51 @@ partial work is discarded if the user forgets to call `commit()`. In Java (both
 can still be overridden via `tx.onClose(Transaction.CLOSE_BEHAVIOR.COMMIT)`. The non-Java GLVs do not support
 configuring close behavior and always rollback.
 
+==== Expanding Dynamic Arguments to Additional Steps
+
+Prior to 4.0, comparing a traverser's value against a dynamically computed reference required verbose workarounds with
```
spmallette: "verbose workarounds" - is it right to describe them as "workarounds"? i mean, these were just the semantics at hand and they were designed with purpose. I think the intro needs rewording.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -101,6 +70,51 @@ partial work is discarded if the user forgets to call `commit()`. In Java (both
...
+[source,groovy]
+----
+// Multi-source filtering with within()
+g.V().has("name", P.within(__.V(1).out("knows").values("name").fold(), __.constant("peter")))
+
+// Set multiple properties from a computed Map
+g.V(4).property(__.V(1).project("friendCount").by(__.out("knows").count()))
+----
+
+See <<has-step,has()>>, <<v-step,V()>>, <<property-step,property()>>, <<is-step,is()>>, and
```
spmallette: These should be linked as part of the "See:" below.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -101,6 +70,51 @@ partial work is discarded if the user forgets to call `commit()`. In Java (both
...
+
+[source,groovy]
+----
+// Multi-source filtering with within()
+g.V().has("name", P.within(__.V(1).out("knows").values("name").fold(), __.constant("peter")))
+
+// Set multiple properties from a computed Map
+g.V(4).property(__.V(1).project("friendCount").by(__.out("knows").count()))
+----
+
```
spmallette, line 450: This needs a summarizing finish of some sort, not to mention a description of patterns folks should be looking to replace. I would think a lot of `where` usage is obvious from the first example, but I think it is worth being explicit. Perhaps consider my comment here: https://github.com/apache/tinkerpop/pull/3458#issuecomment-4723691197 and see what might change elsewhere in our docs and then use that to inform what patterns we should advise about.

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -1525,6 +1525,37 @@ None
...
+
+*Syntax:* `E(Object... edgeIds)` | `E(Traversal idTraversal)`
+
+[width="100%",options="header"]
+|=========================================================
+|Start Step |Mid Step |Modulated |Domain |Range
+|Y |Y |N |`any` |`Edge`
+|=========================================================
+
+*Arguments:*
```
spmallette, line 1601: Makes no metnion of: 

```
gremlin> g.E([11, 12])
==>e[11][4-created->3]
==>e[12][6-created->3]
```

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -2477,6 +2743,37 @@ See: link:https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/j
...
+
+*Syntax:* `V(Object... vertexIds)` | `V(Traversal idTraversal)`
+
+[width="100%",options="header"]
+|=========================================================
+|Start Step |Mid Step |Modulated |Domain |Range
+|Y |Y |N |`any` |`Vertex`
+|=========================================================
+
+*Arguments:*
```
spmallette, line 2997: Makes no mention of:

```
gremlin> g.V([2,3])
==>v[2]
==>v[3]
```

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -1637,6 +1668,170 @@ See: link:https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/j
...
+=== has()
+
+*Description:* Filters traversers by property existence, property value, or label.
+
+*Syntax:* `has(String key)` | `has(String key, Object value)` | `has(String key, P predicate)` | `has(String key, Traversal traversal)` | `has(String label, String key, Object value)` | `has(String label, String key, P predicate)`
+
+[width="100%",options="header"]
+|=========================================================
+|Start Step |Mid Step |Modulated |Domain |Range
+|N |Y |N |`Element` / `Map` |`Element` / `Map`
```
spmallette: `Map` for domain and range?

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -1637,6 +1668,170 @@ See: link:https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/j
 link:https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/sideEffect/GroupCountSideEffectStep.java[source (sideEffect)],
 link:https://tinkerpop.apache.org/docs/x.y.z/reference/#groupcount-step[reference]
 
+[[has-step]]
```
spmallette, line 1745: i think that if we're going to add semantics, and we should, they need to be complete. `has()` in particular is lacking in details for all the nuances it has.

docs/src/dev/provider/gremlin-semantics.asciidoc
```diff
@@ -1637,6 +1668,170 @@ See: link:https://github.com/apache/tinkerpop/tree/x.y.z/gremlin-core/src/main/j
...
+
+*Syntax:* `hasId(Object id, Object... otherIds)` | `hasId(P predicate)` | `hasId(Traversal traversal)`
+
+[width="100%",options="header"]
+|=========================================================
+|Start Step |Mid Step |Modulated |Domain |Range
+|N |Y |N |`Element` |`Element`
+|=========================================================
+
+*Arguments:*
```
spmallette, line 1808:
```
gremlin> g.V().hasId([2,3])
==>v[2]
==>v[3]
```

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/map/VETraversal.feature
```diff
@@ -0,0 +1,149 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
spmallette: I'd prefer we not invent new feature test files like this. I think we want to keep these tests as step focused and oriented as possible - not feature focused. Integrate these tests into the existing Edge/Vertex.feature files.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/integrated/ReadOnlyChildVerification.feature
```diff
@@ -0,0 +1,212 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
spmallette: At a minimum, this test set does not really belong in this directory, but i feel like these tests should just be dispersed among their individual step files. That said, I wonder if we are over-testing this at the Gherkin level? There are some general problems with the tests themselves, in misuse of `@GraphComputerVerificationMidVNotSupported`. 

Maybe could go to the `integrated/` directory - probably then ok to just keep it as-is and whole.  Ensure it has the common tests we have for all strategies in that case.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/HasTraversal.feature
```diff
@@ -0,0 +1,433 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
spmallette: All of these `*Traversal.feature` tests should be integrated to their parent step files. 

For this file, the fact that every single test has `@GraphComputerVerificationMidVNotSupported` tells me something is wrong because it either means we've mis-labeled the test (as happened elsewhere) or we don't have all the right testing. 

Where are the tests for use cases with `select()` for example? I know `V()` is popular but `select()` will equally come into play. That approach and others seem necessary. What about filtering on a `sack()` value? that might be interesting.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/IsTraversal.feature
```diff
@@ -0,0 +1,273 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
spmallette: same general comments as for `has()` - https://github.com/apache/tinkerpop/pull/3458/changes#r3443937188

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/WhereTraversal.feature
```diff
@@ -0,0 +1,168 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
spmallette: same general comments as for `has()` - https://github.com/apache/tinkerpop/pull/3458/changes#r3443937188

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/HasTraversal.feature
```diff
@@ -0,0 +1,433 @@
...
+      """
+    When iterated to list
+    Then the result should be unordered
+      | result |
+      | v[vadas] |
+      | v[josh] |
+      | v[ripple] |
+      | v[lop] |
+
+  # Multi-traversal without() - exclusion from multiple relationship sources
```
spmallette: I don't believe the multi-traversal semantics are correct. In the reference docs we describe it as:

> * The multi-traversal form `within(trav1, trav2, ...)` takes the first result from each traversal and combines them into a collection for membership testing.

but `within` doesn't really work like that: 

```
gremlin> g.V().has('name',within(['josh','lop'], 'vadas'))
==>v[2]
```

It only expands the list if the first and singular argument is a collection:

```
gremlin> g.V().has('name',within(['vadas','josh','lop']))
==>v[2]
==>v[3]
==>v[4]
```

Otherwise, as shown in the first example, it will treat it like comparing on a `List` and a `String`.  `V` and `E` have similar semantics, but oddly:

```
gremlin> g.V().hasId(2, [3, 1])
==>v[2]
==>v[3]
==>v[1]
gremlin> g.V(2, [3, 1])
Expected an id that is convertible to class java.lang.Integer but received class java.util.ArrayList - [[3, 1]]
Display stack trace? [yN]
```

This goes back to https://issues.apache.org/jira/browse/TINKERPOP-2863 where I think there is a bug in the implementation which tried to get `hasId` to be like `V` and `E` but allowed a little too much to happen. I think all of this needs to be rectified.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/HasTraversal.feature
```diff
@@ -0,0 +1,433 @@
...
+
+  # Multi-traversal within() - union of relationship traversals from different sources
+  @GraphComputerVerificationMidVNotSupported
+  Scenario: g_V_hasXname_withinXVXvid1X_outXknowsX_valuesXnameX_VXvid3X_outXcreatedX_valuesXnameXXX
+    Given the modern graph
+    And using the parameter vid1 defined as "v[marko].id"
+    And using the parameter vid3 defined as "v[josh].id"
+    And the traversal of
+      """
+      g.V().has("name", P.within(__.V(vid1).out("knows").values("name").fold(), __.V(vid3).out("created").values("name").fold()))
```
spmallette: Given my  comment here; https://github.com/apache/tinkerpop/pull/3458/changes#r3443997744 i think this should be written as:

```
g.V().has("name", P.within(__.union(__.V(vid1).out("knows").values("name"), 
                                    __.V(vid3).out("created").values("name")).fold()))
```

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/HasTraversal.feature
```diff
@@ -0,0 +1,433 @@
...
+      | d[35].i |
+
+  # Multi-traversal within() - dynamic edge filtering via inV property check
+  @GraphComputerVerificationMidVNotSupported
+  Scenario: g_VXvid1X_outEXknowsX_filterXinV_hasXname_withinXV_hasXname_lopX_inXcreatedX_valuesXnameX_V_hasXname_rippleX_inXcreatedX_valuesXnameXXXX
+    Given the modern graph
+    And using the parameter vid1 defined as "v[marko].id"
+    And the traversal of
+      """
+      g.V(vid1).outE("knows").filter(__.inV().has("name", P.within(__.V().has("name","lop").in("created").values("name").fold(), __.V().has("name","ripple").in("created").values("name").fold())))
```
spmallette: The Gremlin in our tests should be formatted to match our best practices. These tests need to continue to be a solid dependable source for human curated Gremlin.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/AcceptsChildPredicateTraversal.java
Cole-Greer: I've opted to go in a different direction on this one, renaming it to `ReadOnlyTraversalParent extends TraversalParent`. I added the `extends TraversalParent` as I don't think this marker interface makes any sense for steps which are not already TraversalParents. I also think that this new framing is easier to understand, it is a `TraversalParent` with a `ReadOnly` restriction applied.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/AcceptsChildPredicateTraversal.java
Cole-Greer: I think this points to a larger branding/framing problem with this interface. I've renamed it `ReadOnlyTraversalParent` and updated the javadoc. See below thread for more details.

docs/src/upgrade/release-4.x.x.asciidoc
Cole-Greer: The upgrade docs examples should remain static such that they remain as-is following future releases. I don't think the results are particularly interesting for these traversals, however I can add static emulated console output.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/branch/ChooseStep.java
Cole-Greer: Added the javadocs

docs/src/upgrade/release-4.x.x.asciidoc
spmallette: yes, i'm referring to executed in the console with static results. i don't think we've ever used "interesting" or any other reason as criteria for including console output. it's just mostly what we do. we might truncate long results or opt to not display results for GLV related examples (because javascript for example wouldn't make sense in this context), but we generally demonstrate by way of the console.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/GremlinLangTraversalRoundTripTest.java
Cole-Greer: I've gone with option 3, and built this class out to be more well rounded. It's not a single parameterized test with 45 to give more complete coverage. The main gap that this test fills is validating GremlinLang strigification/parsing roundtrip equality. To an extent this was already covered through feature tests which are parsed from the gherkin source, and then stringified for remote execution, however that only verifies the traversal produces the expected results, which is weaker than asserting equal traversals through the round trip.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/optimization/InlineFilterStrategy.java
Cole-Greer: I've updated the javadoc and added additional test cases.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/PTraversalTest.java
Cole-Greer: I've opted to fold all of these cases directly into `PTest`. This picks up the clone and negate coverage for free and centralizes all P cases in one place. One noteworthy change is that Traversal-bearing predicates in PTest get resolved with a dummy Traverser, and they bypass the `setValue()` manipulation tests as the value is nonsense prior to calls to `resolve()`

docs/src/upgrade/release-4.x.x.asciidoc
Cole-Greer: I've updated the examples to include the console prefixes and output.

docs/src/reference/the-traversal.asciidoc
Cole-Greer: This section has been stripped down as the examples have been folded into the main `has()` examples section. The important context is now captures in the numbered callouts from the specific `has(traversal)` examples.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/HasTraversal.feature
Cole-Greer: I've migrated all of these `*Traversal.feature` to their appropriate parent step file. For has() specifically, I've removed excessive `@GraphComputerVerificationMidVNotSupported` tags and added a new case for `select()` and `sack()`

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/HasTraversal.feature
Cole-Greer: I've updated the "auto-unfolding" semantics to properly align with the literal-form of `within()`

```
gremlin> g.inject(1).is(P.within(constant([1,2]), constant(3)))
gremlin> g.inject(1).is(P.within(constant([1,2,3])))
==>1
```

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/process/traversal/PTraversalTest.java
Cole-Greer: This has been migrated to `PTest` and now includes more deeply nested ands of ors and ors of ands.

docs/src/dev/provider/gremlin-semantics.asciidoc
Cole-Greer: My agent may have gone rogue there, fixed it.

Cole-Greer:
@spmallette Regarding your comment here:
> There are some docs here that cover the feature, but what I don't see are updates to existing docs that use old patterns that are no longer relevant and that we should no longer promote. Has that kind of search been done?

I got an agent to do a scan through all of our documentation for examples in need of updates, I found a few cases in the recipes docs, notable in anti-patterns and traversal-induced-values. The anti pattern has been removed entirely, and the traversal-induced-values example has been updated to the new recommended pattern.

I haven't done any manual reads through the bulk of the docs in regards to this feature alone, but I intend to do a thorough docs review in the lead up to the 4.0.0-beta.3 release which may reveal some other examples which would benefit from refinement.

docs/src/upgrade/release-4.x.x.asciidoc
spmallette: as i read all the changes from `where()` to `has()` it makes me wonder what we should be prescribing and when.  are there cases when `where` is actually better than `has()`? or are there things that `has()` cannot do that you must use `where()`? i think this needs some thought to ensure these upgrade/reference docs are solid.

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/filter/Has.feature
```diff
@@ -620,4 +620,451 @@ Feature: Step - has()
     g.V().has("name", null)
     """
     When iterated to list
-    Then the result should be empty
\ No newline at end of file
+    Then the result should be empty
+  @GraphComputerVerificationMidVNotSupported
```
spmallette: overall the test organization looks better since i called in out in my first review, but i'm still not seeing where we've expanded tests much to cover cases that don't use mid-V.

docs/src/reference/the-traversal.asciidoc
spmallette: additional reference link for `V(Traversal)`?

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/verification/ReadOnlyChildVerificationStrategy.java
```diff
@@ -0,0 +1,60 @@
...
+        implements TraversalStrategy.VerificationStrategy {
+
+    private static final ReadOnlyChildVerificationStrategy INSTANCE = new ReadOnlyChildVerificationStrategy();
+
+    private ReadOnlyChildVerificationStrategy() {
+    }
+
+    @Override
+    public void apply(final Traversal.Admin<?, ?> traversal) {
+        for (final Step<?, ?> step : traversal.getSteps()) {
```
spmallette: have we examined if we are over-iterating here given how strategies tend to be applied plus the recursive call in `ReadOnlyChildValidator`? perhaps there is a room for some added efficiency here?

Cole-Greer: Yes, and we excessively "double up" in one case, if a `ReadOnlyTraversalParent` is itself a descendant of another `ReadOnlyTraversalParent`, the child's children will be needlessly scanned for Mutating steps a second time.

I've accepted that as a reasonable trade off for simplicity as I expect that to be a fairly rare case given the intended usage of these steps. If we were to avoid recursing into the descendants of a `ReadOnlyTraversalParent` and instead rely on the recursive application of the strategy to all children, we would instead need to include a test if `traversal` itself is a descendant of a `ReadOnlyTraversalParent` which we don't have a great way to do, and would likely negate any savings from double recursion.

Cole-Greer: Taking another pass over this code, I think it can be simplified a bit by relying on `TraversalHelper.getStepsOfAssignableClass(ReadOnlyTraversalParent.class)`. This wouldn't really impact the performance much but would be a decent simplification of the code.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/verification/ReadOnlyChildVerificationStrategy.java
```diff
@@ -0,0 +1,60 @@
...
+            if (step instanceof ReadOnlyTraversalParent) {
+                for (final Traversal.Admin<?, ?> child : ((ReadOnlyTraversalParent) step).getLocalChildren()) {
+                    try {
+                        ReadOnlyChildValidator.validate(child);
+                    } catch (final IllegalArgumentException e) {
+                        throw new VerificationException(e.getMessage(), traversal);
+                    }
+                }
```
Cole-Greer: We should recurse through global children as well. I don't think any of the current set of `ReadOnlyTraversalParent` steps actually have any global children so there's no impact right now, but it would be more future proof.
```suggestion
                for (final Traversal.Admin<?, ?> child : ((ReadOnlyTraversalParent) step).getLocalChildren()) {
                    try {
                        ReadOnlyChildValidator.validate(child);
                    } catch (final IllegalArgumentException e) {
                        throw new VerificationException(e.getMessage(), traversal);
                    }
                }
                for (final Traversal.Admin<?, ?> child : ((ReadOnlyTraversalParent) step).getGlobalChildren()) {
                    try {
                        ReadOnlyChildValidator.validate(child);
                    } catch (final IllegalArgumentException e) {
                        throw new VerificationException(e.getMessage(), traversal);
                    }
                }
```

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/strategy/verification/ReadOnlyChildVerificationStrategy.java
spmallette: I'm not completely sure it's innocuous as that, but irrespective of that point I'm now wondering why this is a separate strategy. Why wouldn't we fold this functionality under the `StandardVerificationStrategy`? This is very standard behavior that must be executed - I can't think of a reason you would opt-out of just this verification. Folding the feature into "standard" would allow the validation to happen within the same iteration of steps as the existing verification instead of a second pass. It would also remove a bunch of GLV code and unnecessary testing.

PR #3460, opened 2026-06-15
TINKERPOP-3238 Throw FailResponseException for fail() step over remote

spmallette:
https://issues.apache.org/jira/browse/TINKERPOP-3238

The Java driver now throws a FailResponseException, a subclass of ResponseException that implements Failure, when the server returns a SERVER_ERROR_FAIL_STEP (595) status code. This makes remote handling of the fail() step more consistent with embedded behavior.

Failure.format() is guarded against null traversal/traverser context since that data is not transmitted from the server, preventing a NullPointerException when formatting a remotely-reconstructed Failure.

Adds unit tests for both the defensive format() logic and the new exception, plus documentation in the reference and upgrade guides.

@kenhuuu could you say what this change means in 4.0? I don't know that we have the same kind of Exception hierarchy there anymore iirc

VOTE +1

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -26,6 +26,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 === TinkerPop 3.7.7 (Release Date: NOT OFFICIALLY RELEASED YET)
 
 * Added `NextN(n)` to `Traversal` in `gremlin-go` for batched result iteration, providing API parity with `next(n)` in the Java, Python, and .NET GLVs.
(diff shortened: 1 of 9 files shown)


kenhuuu:
> could you say what this change means in 4.0? I don't know that we have the same kind of Exception hierarchy there anymore iirc

we still use ResponseException and could probably do something very similar to what is done here. We don't have that same error code anymore but we do get a "ServerFailStepException" as the exception type in the response so we'd use that as the conditional instead.

kenhuuu:
VOTE +1

Cole-Greer:
VOTE +1

PR #3461, opened 2026-06-15
Share one connection pool across remote transactions in gremlin-driver

kenhuuu:
Each Cluster.transact() previously built its own ClusteredClient, which stood up a full set of per-host connection pools and then tore them down on every commit/rollback. That cost memory and added connection-setup latency (TCP/TLS handshake) to every transaction, even though a transaction only ever talks to one host.

Transactions don't need their own pools or a dedicated connection: the server binds each request to its transaction via the transaction id, and the HTTP model returns a connection to the pool as soon as the response completes. So a transaction only needs to pin to a host, not hold a connection. This lets all transactions share a single, eagerly created ClusteredClient and borrow a connection per request from the pinned host's warm pool, giving transaction requests the same connection occupancy as normal query traffic.

While reworking PinnedClient:
- Host selection now delegates to the cluster's LoadBalancingStrategy so transactions honor the configured strategy and skip unavailable hosts instead of picking a random (possibly dead) host.
- closeAsync() no longer tears down pools. The pools belong to the shared client and are closed with the Cluster, after open transactions are rolled back; closing one transaction must not break its siblings.

### DIFF
--- a/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
+++ b/gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Client.java
@@ -40,6 +40,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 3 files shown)


kenhuuu:
VOTE +1

Cole-Greer:
VOTE +1

xiazcy:
VOTE +1

PR #3464, opened 2026-06-17
TINKERPOP-3253 Add transaction closure methods to GLVs

kenhuuu:
https://issues.apache.org/jira/browse/TINKERPOP-3253

Adds executeInTx/evaluateInTx to GraphTraversalSource across all GLVs to wrap the begin/commit/rollback lifecycle, so a lambda receives the transaction-bound g and is auto-committed on success or rolled back on error.

The methods live on GraphTraversalSource rather than on the Transaction returned by g.tx() because the closure is a Traversal-API convenience; hosting it on the Transaction would hand Driver-API users (who submit strings) a traversal source and mix the two APIs. Keeping it on g also lets Java's existing g.tx() routing cover both embedded and remote with no interface changes. The value-returning form is named evaluateInTx rather than call because call already exists on GraphTraversalSource as the service step, and the InTx suffix keeps the transactional intent clear now that the methods no longer sit under tx().

The surface follows each language's idiom rather than forcing uniformity: dynamic languages expose a single method, while statically typed languages get a void/value pair. Go returns interface{} instead of using generics, to match the driver's existing untyped result API.

Behavior is single-shot with no retry, since Gremlin Server has no standardized retriable-error signal across providers. On commit failure a rollback is still attempted for server-side resource hygiene, but the original error stays primary and secondary cleanup failures are only logged. gtx.tx() still returns the same transaction so the commit path keeps working; only opening a second transaction errors.

VOTE +1

### DIFF
--- a/docs/src/reference/gremlin-variants.asciidoc
+++ b/docs/src/reference/gremlin-variants.asciidoc
@@ -354,6 +354,34 @@ err = tx.Commit()
 if err != nil { log.Fatal(err) }
 ----
 
+==== Managed Transaction Blocks
+
+To avoid the manual begin/commit/rollback boilerplate, `g` provides closure methods that manage the
+lifecycle: the closure receives the transaction-bound `gtx` and the transaction is committed when it returns a `nil`
+error or rolled back if it returns an error (or panics). `ExecuteInTx` is for the common no-result case; `EvaluateInTx`
+returns the value the body produces as `interface{}`, which the caller type-asserts:
+
+[source,go]
+----
+g := gremlingo.Traversal_().With(remote)
+
+// ExecuteInTx: no return value
+err := g.ExecuteInTx(func(gtx *gremlingo.GraphTraversalSource) error {
+    _, e := gtx.AddV("person").Property("name", "alice").Iterate()
+    return <-e
+})
+
+// EvaluateInTx: returns the body's value as interface{}
+v, err := g.EvaluateInTx(func(gtx *gremlingo.GraphTraversalSource) (interface{}, error) {
+    return gtx.V().Count().Next()
+})
+count := v.(int64)
+----
+
+The transaction runs exactly once (no automatic retry). The body's error is returned after the rollback; if the commit
+fails, a rollback is still attempted and the commit error is returned. A panic in the body rolls back and then
+re-panics.
+
 ==== Driver API
 
 [source,go]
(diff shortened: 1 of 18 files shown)


xiazcy:
VOTE +1

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -354,6 +354,34 @@ err = tx.Commit()
 if err != nil { log.Fatal(err) }
 ----
 
+==== Transaction Closures
```
spmallette: ngl, even though I knew what this PR was for, i still read this title as "closing a transaction". 

Maybe, "Transaction Blocks" as a language independent way to describe things? At least with "blocks" it's hard to make an assumption about what that is and you click? Or maybe, "Transaction Batching" or "Batches"?

docs/src/reference/the-traversal.asciidoc
```diff
@@ -60,8 +60,57 @@ traversal strategies may not function properly.
...
+g.executeInTx(gtx -> {
+    gtx.addV('person').iterate();
+    gtx.addV('software').iterate();
+});
+
+// use evaluateInTx when the unit of work needs to return a value
+long count = g.evaluateInTx(gtx -> gtx.V().count().next());
+----
+
+This closure form is preferred for several reasons:
```
spmallette: I personally would like docs to steer clear of heavy reliance on bullets. A smoothly written paragraph is the form we should try to stick to for "reasons". The entire readable internet is leaning into bullets like this now and I think TinkerPop is better for breaking from that. Not saying they can't ever be used, but this just doesn't seem like content to do it with.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -60,8 +60,57 @@ traversal strategies may not function properly.
...
+`g.tx()` returns a distinct transaction object each time, keyed by a server-side transaction ID, and the `gtx` spawned
+from it is a genuinely separate `GraphTraversalSource` from `g` — issuing a traversal against `g` instead of `gtx`
+silently runs as its own implicit transaction. In the *embedded* case `g.tx()` returns a thread-bound transaction (the
+traditional Java model in which the transaction is associated with the current thread), so `g` and `gtx` actually
+share the same transaction scope. Because the closure hands you the correct transaction-bound `gtx` and owns the
+lifecycle, the exact same code is correct under both paradigms; you do not have to reason about whether `g` and `gtx`
+refer to the same underlying transaction. (The non-Java GLVs are always remote, so this distinction does not arise
+there.)
+
+The closure runs the unit of work exactly once (there is no automatic retry). If the closure throws, the original
```
spmallette, line 103: please change the transition to drop the semicolon in this paragraph.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -60,8 +60,57 @@ traversal strategies may not function properly.
...
+error is re-raised after the rollback; if the `commit()` itself fails, the commit error is raised and a rollback is
+still attempted to release server-side resources.
+
+Manual transaction control (calling `begin()`, `commit()`, and `rollback()` yourself, shown below) remains fully
```
spmallette, line 107: "fit the use case — for example, when the" - just make it two sentences.

Cole-Greer (commented):
VOTE +1

docs/src/reference/gremlin-variants.asciidoc
kenhuuu: updated to "Managed Transaction Blocks"

docs/src/reference/the-traversal.asciidoc
kenhuuu: turned into paragraph

PR #3466, opened 2026-06-17
Standardized `gremlin-driver` (Java) connection options

GumpacG:
# Standardize `gremlin-driver` (Java) connection options

Implements the Java portion of the TinkerPop 4.x GLV connection-options standardization. Renames several options to canonical names (old names kept as deprecated aliases), aligns defaults, adds new options, and removes one. Reference-driver changes only; the other GLVs follow in separate PRs.

| Old | New | Default |
|-----|-----|---------|
| `maxConnectionPoolSize` | `maxConnections` | 128 |
| `resultIterationBatchSize` | `batchSize` | 64 |
| individual SSL builders | `ssl(SslContext)` | - |
| `RequestOptions.addG` | `traversalSource` | - |

- **`connectTimeout`** default lowered 15s → 5s, and it is now actually applied (it was previously validated but never enforced, so connection setup was effectively unbounded).
- **`compression`** now defaults to `DEFLATE` (on); the driver sends `Accept-Encoding: deflate` by default. Set `Compression.NONE` to disable.
- **`validationRequest`** builder default reconciled to `g.inject(0)`.

## New options

- **`readTimeout`** (default off) - streaming-safe idle-read timeout, armed per request.
- **`keepAliveTime`** (30s) - idle time before TCP keep-alive probes (`SO_KEEPALIVE` + `TCP_KEEPIDLE` where supported).
- **`maxResponseHeaderBytes`** (8192) - previously hardcoded.
- **`proxy(ProxyOptions)`** and **`url(String)`**.

## Removed (breaking)

- **`maxResponseContentLength`** - responses now stream; `readTimeout` is the partial mitigation.

## Testing

- `gremlin-driver` unit tests pass, including new `ConnectionOptionsTest` and `ReadTimeoutHandlerTest`.
- `gremlin-server` integration tests updated: a new `readTimeout` cleanup/recovery test replaces the removed cap-based tests.
- CHANGELOG, reference config table, and upgrade guide updated.


CHANGELOG.asciidoc
```diff
@@ -25,6 +25,19 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 [[release-4-0-0]]
 === TinkerPop 4.0.0 (Release Date: NOT OFFICIALLY RELEASED YET)
 
+* Standardized `gremlin-driver` (Java) connection options
+** Renamed the `connectionSetupTimeoutMillis` builder option to `connectTimeout` (default lowered from 15s to 5s) and actually wired it to `ChannelOption.CONNECT_TIMEOUT_MILLIS` on the connection bootstrap; it previously was validated but never applied. The old method is retained as a deprecated alias.
```
Cole-Greer: We've never used sub-bullets before in the changelog to my knowledge. I suppose it's not an unreasonable convention to start here.

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -991,24 +991,27 @@ The following table describes the various configuration options for the Gremlin
...
-|connectionPool.connectionSetupTimeoutMillis | Duration of time in milliseconds provided for connection setup to complete which includes the SSL handshake. |15000
+|connectionPool.compression |The wire compression algorithm negotiated with the server. Can be: `NONE` or `DEFLATE`. |DEFLATE
+|connectionPool.connectTimeout | Duration of time in milliseconds that bounds TCP connection establishment (transport setup, including the SSL handshake). Formerly `connectionSetupTimeoutMillis`, which is still accepted as a deprecated alias. |5000
 |connectionPool.enableSsl |Determines if SSL should be enabled or not. If enabled on the server then it must be enabled on the client. |false
-|connectionPool.idleConnectionTimeout | Duration of time in milliseconds that the driver will allow a channel to not receive read or writes before it automatically closes. |180000
+|connectionPool.idleTimeout | Duration of time in milliseconds that the driver will allow a channel to not receive read or writes before it automatically closes. Formerly `idleConnectionTimeoutMillis`, which is still accepted as a deprecated alias. |180000
```
kenhuuu: This is something I should probably have mentioned in the devlist and not here but sometimes its nice to just have the Ms post-fix so something like "idleTimeoutMs" so its immediately obvious what the unit is.

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -991,24 +991,27 @@ The following table describes the various configuration options for the Gremlin
...
-|connectionPool.maxWaitForConnection |The amount of time in milliseconds to wait for a new connection before timing out. |3000
+|connectionPool.maxConnections |The maximum size of a connection pool for a host. Formerly `maxConnectionPoolSize`, which is still accepted as a deprecated alias. |128
+|connectionPool.maxResponseHeaderBytes |The maximum size in bytes allowed for the HTTP response headers. |8192
+|connectionPool.maxWaitForConnection |The amount of time in milliseconds to wait for a new connection before timing out. |16000
 |connectionPool.maxWaitForClose |The amount of time in milliseconds to wait for pending messages to be returned from the server before closing the connection. |3000
 |connectionPool.reconnectInterval |The amount of time in milliseconds to wait before trying to reconnect to a dead host. |1000
-|connectionPool.resultIterationBatchSize |The override value for the size of the result batches to be returned from the server. |64
+|connectionPool.defaultBatchSize |The default value for the per-request batch size used when a request does not specify one. Formerly `resultIterationBatchSize`, which is still accepted as a deprecated alias. |64
```
kenhuuu: Again, probably devlist thing, but why does this have the word "default" in it?

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -32,6 +32,54 @@ complete list of all the modifications that are part of this release.
 
 === Upgrading for Users
 
+==== Standardizing Java Connection Options
+
+TinkerPop 4.x standardizes connection option names and defaults across the GLVs. In the Java reference driver
+(`gremlin-driver`), several `Cluster.Builder` options have been renamed for consistency, with the old names retained
+as deprecated aliases so that existing code keeps compiling, and a number of new options have been added. The notes
```
kenhuuu: For this major version change, I suggest we reduce the clutter and simply delete the old names.

CHANGELOG.asciidoc
```diff
@@ -25,6 +25,19 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
...
+** Renamed `maxConnectionPoolSize` to `maxConnections` (default 128), `idleConnectionTimeoutMillis` to `idleTimeout` (default 180s), and `resultIterationBatchSize` to `defaultBatchSize` (default 64); the old builder methods, accessors, and YAML keys are retained as deprecated aliases.
+** Added `traversalSource` to `RequestOptions.Builder` as the canonical name for `addG` (retained as a deprecated alias).
+** Added `readTimeout` (default 0/off), a per-request idle-read timeout armed when a request is written and reset on each inbound response chunk via a new `ReadTimeoutHandler` inserted after the HTTP codec, so it never fires while a pooled connection is idle between requests.
```
Cole-Greer: Overall this is a lot of changelog for work of this size, if we extend this across all GLVs, we could end up with 30% of the changelog for TP4 just being the connection options. I think it's worth trimming this as much as possible. The most important details users need to know is which options have been renamed, followed by which options have been introduced. I think most of the details of what each option does can be trimmed down/out.

The option name should be generally self-documenting on its own, and the fine details should be filled in by reference documentation and javadocs `Cluster.Builder`, `Settings` and such.

Cole-Greer (commented):
VOTE +1

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/ReadTimeoutHandler.java
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Bounds the idle time between inbound response chunks on a per-request basis. The timeout is <em>armed</em> when a
+ * request is written and rescheduled on every inbound read; it is <em>disarmed</em> once the last chunk of the
+ * response arrives. This differs from a static {@link io.netty.handler.timeout.ReadTimeoutHandler} so that it never
+ * fires while a pooled connection sits idle between requests (which is the concern of the idle-pool-close handler).
+ * <p>
+ * When the timeout elapses while awaiting a response a {@link ReadTimeoutException} is fired up the pipeline and the
+ * channel is closed.
+ */
+public class ReadTimeoutHandler extends ChannelDuplexHandler {
```
kenhuuu, line 41: This is not what I expected when I saw this. This is an idle read timeout but not a total read timeout which is what I would have expected. This overlaps with `idleTimeout`.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java
```diff
@@ -136,6 +161,59 @@ public Connection(final URI uri, final ConnectionPool pool) throws ConnectionExc
         }
     }
 
+    /**
+     * Configures the connection establishment timeout on the supplied {@link Bootstrap}. The {@code connectTimeout}
+     * is expressed in milliseconds and bounds the TCP connection setup (including the SSL handshake) performed by
+     * {@code b.connect()}. It is applied via Netty's {@link ChannelOption#CONNECT_TIMEOUT_MILLIS}.
+     */
+    static void configureConnectTimeout(final Bootstrap b, final long connectTimeout) {
+        b.option(ChannelOption.CONNECT_TIMEOUT_MILLIS, (int) connectTimeout);
```
kenhuuu: If Netty can only take an int, then connectTimeout should also be an int to prevent silent truncation.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/RequestOptions.java
```diff
@@ -150,9 +150,20 @@ public static Builder from(final RequestOptions options) {
             return builder;
         }
 
+        /**
+         * Sets the traversal source (or graph) alias to rebind to "g" on the request.
+         */
+        public Builder traversalSource(final String traversalSource) {
```
kenhuuu, line 156: Nit: I think this has traditionally been referred to as the traversal source name not just traversal source. Also the name "addG" was that because the field in the actual request is just g. So it might even be better to just leave it as "addG" although I also don't really like that name.

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
```diff
@@ -412,29 +412,6 @@ public void shouldProcessEvalInterruption() {
         }
     }
 
-    @Test
-    public void shouldEventuallySucceedAfterChannelLevelError() {
```
kenhuuu, line 416: This is a real test that is using the old max content length as a trigger. I don't think this test should be deleted, just updated.

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
```diff
@@ -139,6 +142,17 @@ public void init(final Connection connection) {
         @Override
         protected void initChannel(final SocketChannel socketChannel) {
             final ChannelPipeline pipeline = socketChannel.pipeline();
+
+            // The proxy handler must run before the SSL handler so the CONNECT tunnel is established prior to the
+            // TLS handshake.
+            final ProxyOptions proxy = cluster.getProxy();
```
kenhuuu, line 148: This doesn't need to be done for this PR, but a Jira would be nice to add proxy testing to the SimpleTestServer thing in gremlin-tools/gremlin-socket-server

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
```diff
@@ -793,20 +964,102 @@ public Builder port(final int port) {
...
+         * the scheme selects SSL ({@code https} enables SSL, {@code http} disables it; other or absent schemes leave
+         * the current SSL setting unchanged), the host is added as a contact point, an explicit port is applied via
+         * {@link #port(int)}, and a non-empty path is applied via {@link #path(String)}. This is a convenience for the
+         * common single-host case; use {@link #addContactPoint(String)} / {@link #addContactPoints(String...)} along
+         * with {@link #port(int)} and {@link #path(String)} for multi-host configurations.
+         *
+         * @param url a full URL such as {@code https://gremlin.example.com:8182/gremlin}
+         * @throws IllegalArgumentException if the URL cannot be parsed or has no host
+         */
+        public Builder url(final String url) {
```
kenhuuu, line 893: Potentially confusing. Right now, the Cluster supports multiple hosts but they must all share a single port. This method is already documented as being a convenience for a single endpoint, but I think you should actually throw if a contact point already exists to prevent misuse.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -32,6 +32,54 @@ complete list of all the modifications that are part of this release.
...
+  sets `TCP_KEEPIDLE` where supported (JDK 11+ on Linux and macOS). Set 0 to disable.
+- `maxResponseHeaderBytes` (default 8192): the maximum response header size, which was previously hardcoded.
+- `proxy(ProxyOptions)`: routes connections through an HTTP proxy.
+- `url(String)`: configures the endpoint from a single URL (scheme, host, port, and path). The existing
+  `addContactPoint`/`addContactPoints`, `port`, and `path` options remain for multi-host configurations.
+
+The `maxResponseContentLength` setting has been removed. Responses are now streamed rather than aggregated
+and size-capped. The new idle `readTimeout` is the partial mitigation for unbounded or slow responses.
+
+See: link:https://lists.apache.org/thread/yqtr2wnb1kq2pqqq4002cz511q5o0bkg[dev list discussion on standardizing GLV connection options].
```
spmallette: > [dev list discussion on standardizing GLV connection options]

We tend to just do this like: `link:https://lists....[DISCUSS] thread` - whatever we do, it should be made consistent so that agents don't get confused on how to write these things.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -32,6 +32,54 @@ complete list of all the modifications that are part of this release.
 
 === Upgrading for Users
 
+==== Standardizing Java Connection Options
```
spmallette, line 35: As I look at the other related PRs - should we just have one "Standard Connection Options" a general introduction and then language specific subsections? maybe a single table could even best convey the changes somehow if we've actually standardized everything...

CHANGELOG.asciidoc
GumpacG: It will be trimmed down for the remaining PR of the series to reduce noise in the CHANGELOG and update docs.

docs/src/upgrade/release-4.x.x.asciidoc
GumpacG: It will be trimmed down and include a table for the last PR of the series for all the GLVs to reduce noise in the CHANGELOG and update docs.

docs/src/reference/gremlin-variants.asciidoc
GumpacG: As discussed offline we will keep the millis suffix and add a separate option to follow language idiomatic units.
https://lists.apache.org/thread/nwhob2gtnkdhj64o16jlyqost1g6c7c4

docs/src/reference/gremlin-variants.asciidoc
GumpacG: Changed to just be `batchSize`

docs/src/upgrade/release-4.x.x.asciidoc
GumpacG: Removed

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/handler/ReadTimeoutHandler.java
GumpacG: I've split the responsibilities of `idleTimeout` and `readTimeout` so that there is no overlap. In flight requests is now out of `idleTimeout`'s scope

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Connection.java
GumpacG: Converted to int

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/RequestOptions.java
GumpacG: As discussed offline, we are keeping it `traversalSoruce` for parity across GLVs

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Channelizer.java
GumpacG: Created https://issues.apache.org/jira/browse/TINKERPOP-3259

gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
GumpacG: Reverted and updated

gremlin-driver/src/main/java/org/apache/tinkerpop/gremlin/driver/Cluster.java
GumpacG: Updated to throw in this case

kenhuuu:
VOTE +1

xiazcy:
VOTE +1

PR #3468, opened 2026-06-19
Standardized `gremlin-dotnet` connection options

GumpacG:
# Standardize `gremlin-dotnet` connection options

Implements the .NET portion of the TinkerPop 4.x GLV connection-options standardization. Renames several `ConnectionSettings` properties and the `Auth` factory methods to canonical names (old names kept as `[Obsolete]` deprecated aliases), aligns defaults, adds new options, and fixes two latent bugs. .NET driver changes only; the other GLVs follow in separate PRs.

| Old                       | New              | Default |
| ------------------------- | ---------------- | ------- |
| `MaxConnectionsPerServer` | `MaxConnections` | 128     |
| `IdleConnectionTimeout`   | `IdleTimeout`    | 180s    |
| `ConnectionTimeout`       | `ConnectTimeout` | 5s      |
| `KeepAliveInterval`       | `KeepAliveTime`  | 30s     |
| `EnableCompression`       | `Compression`    | -       |
| `Auth.BasicAuth`          | `Auth.Basic`     | -       |
| `Auth.SigV4Auth`          | `Auth.Sigv4`     | -       |

The old property names and auth methods are retained as `[Obsolete]` deprecated aliases (marked "As of release 4.0.0, ...") that delegate to the canonical members, so existing code keeps compiling.

* **`ConnectTimeout`** default lowered 15s → 5s.
* **`KeepAliveTime`** is now wired to a real TCP keep-alive socket option (via `SocketsHttpHandler.ConnectCallback`) rather than the inert HTTP/2 ping timeout, which had no effect on HTTP/1.1. It enables `SO_KEEPALIVE` and sets the per-socket idle time on Windows, Linux, and macOS; on other platforms keep-alive stays enabled at the OS default idle time.
* **`Compression`** is now a `{None, Deflate}` enum defaulting to `Deflate` (compression on by default); the driver sends `Accept-Encoding: deflate` by default. Set `Compression.None` to disable. The deprecated `EnableCompression` `bool` is preserved only as a compatibility shim mapping `true`/`false` to `Compression.Deflate`/`Compression.None`.

## New options
* **`Ssl`** (`SslClientAuthenticationOptions`) - adds client-certificate and custom-CA support. `SkipCertificateValidation` is applied to an internal copy rather than mutating the caller's options.
* **`DefaultBatchSize`** (64) - connection-level default that fills the per-request `batchSize` when unset.
* **`MaxResponseHeaderBytes`** - the maximum response header size, in bytes (converted internally to the handler's native kilobyte unit).
* **`ReadTimeout`** - a per-read idle timeout applied to each read of the response stream (via per-read `CancelAfter`).
* **`Proxy`** (`IWebProxy`) - routes connections through an HTTP proxy.
* **`GremlinServer.FromUrl(string)`** / **`GremlinServer(Uri)`** - configure the endpoint from a single URL (scheme, host, port, path), deriving SSL from the scheme. The existing host/port/path constructor remains.

## Removed (breaking)
* **`maxResponseContentLength`** - responses now stream; `ReadTimeout` is the partial mitigation.

## Bug fixes
* Fixed deflate response decompression, which threw on the server's zlib-framed output because it used `DeflateStream` (raw DEFLATE, RFC 1951) instead of `ZLibStream` (zlib, RFC 1950). The bug was previously masked because compression was off by default; turning it on by default would have made it a default-path failure.
* Fixed `Ssl` options cloning (used on the skip-certificate-validation path) to copy `ClientCertificateContext` and `AllowTlsResume`, which were previously dropped, breaking mTLS client certificates and silently re-enabling TLS resumption.

## Testing
* `gremlin-dotnet` unit tests pass (695 tests), including new `ConnectionSettingsTests` and the `ReadTimeout` slow-read timeout test.
* CHANGELOG, reference config table (`gremlin-variants.asciidoc`), and upgrade guide (`release-4.x.x.asciidoc`) updated for the .NET slice.

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionSettings.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionSettings.cs
@@ -49,13 +49,6 @@ namespace Gremlin.Net.Driver
         public const int DefaultMaxConnections = 128;
 
         /// <summary>
-        ///     The default maximum concurrent connections to a single server.
-        /// </summary>
-        /// <remarks>Deprecated compatibility alias; use <see cref="DefaultMaxConnections"/>.</remarks>
-        [Obsolete("As of release 4.0.0, replaced by DefaultMaxConnections.")]
-        public const int DefaultMaxConnectionsPerServer = DefaultMaxConnections;
-
-        /// <summary>
         ///     The default TCP keep-alive idle time (how long a connection is idle before the
         ///     first keep-alive probe is sent).
         /// </summary>
@@ -73,9 +66,29 @@ namespace Gremlin.Net.Driver
         public TimeSpan ConnectTimeout { get; set; } = DefaultConnectTimeout;
 
         /// <summary>
+        ///     Gets or sets <see cref="ConnectTimeout"/> in whole milliseconds. This is the millisecond
+        ///     view of the same setting; <see cref="ConnectTimeout"/> is the idiomatic <see cref="TimeSpan"/> form.
+        /// </summary>
+        public int ConnectTimeoutMillis
+        {
+            get => (int) ConnectTimeout.TotalMilliseconds;
+            set => ConnectTimeout = TimeSpan.FromMilliseconds(value);
+        }
+
+        /// <summary>
         ///     Gets or sets how long idle connections stay in the pool before being closed.
         /// </summary>
         public TimeSpan IdleTimeout { get; set; } = DefaultIdleTimeout;
+
+        /// <summary>
+        ///     Gets or sets <see cref="IdleTimeout"/> in whole milliseconds. This is the millisecond
+        ///     view of the same setting; <see cref="IdleTimeout"/> is the idiomatic <see cref="TimeSpan"/> form.
+        /// </summary>
+        public int IdleTimeoutMillis
+        {
+            get => (int) IdleTimeout.TotalMilliseconds;
+            set => IdleTimeout = TimeSpan.FromMilliseconds(value);
+        }
 
         /// <summary>
         ///     Gets or sets the maximum concurrent connections to a single server.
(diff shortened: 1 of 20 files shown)


gremlin-dotnet/src/Gremlin.Net/Driver/ConnectionSettings.cs
```diff
@@ -85,10 +106,107 @@ public class ConnectionSettings
...
+        ///     whole-request deadline. <see cref="System.Threading.Timeout.InfiniteTimeSpan"/>
+        ///     (the default) disables it.
+        /// </summary>
+        public TimeSpan ReadTimeout { get; set; } = System.Threading.Timeout.InfiniteTimeSpan;
+
+        /// <summary>
+        ///     Gets or sets the HTTP proxy used for connections. When set, it is applied to the
+        ///     underlying handler explicitly.
+        /// </summary>
+        public IWebProxy? Proxy { get; set; }
```
kenhuuu: Maybe too specific to the HttpClient? It seems like some other GLVs create specific options.

GumpacG: IWebProxy is the idiomatic general-purpose .NET proxy abstraction in System.Net and it's exactly what SocketsHttpHandler.Proxy consumes.
  
Regarding other GLVs, Java is the only driver that defines a custom proxy type (ProxyOptions). The others all use their platform's native proxy representation, same as .NET does:
  - .NET: IWebProxy (BCL interface) -> handler.Proxy
  - Go: func(*http.Request) (*url.URL, error) -> http.Transport.Proxy
  - Python: proxy URL string -> aiohttp post(proxy=...)
  - JS: proxy URI string -> undici ProxyAgent
  - Java: custom ProxyOptions -> Netty HttpProxyHandler

Java needed its own class because Netty is low-level (you insert an HttpProxyHandler with an address + credentials, there's no ready-made proxy object), whereas .NET/Go/Python/JS each sit on a client that already exposes a proxy type or URL. So IWebProxy keeps .NET consistent with the majority pattern (native proxy representation per language).

kenhuuu:
VOTE +1

Cole-Greer (commented):
VOTE +1

xiazcy:
VOTE +1

PR #3469, opened 2026-06-19
Standardized `gremlin-python` connection options

GumpacG:
# Standardize `gremlin-python` connection options

Implements the Python portion of the TinkerPop 4.x GLV connection-options standardization. Renames two options to canonical names (old names kept as deprecated aliases), adds a set of new options, removes two, and raises the minimum `aiohttp` floor. Python driver changes only; the other GLVs follow in separate PRs.

| Old           | New               | Default      |
| ------------- | ----------------- | ------------ |
| `pool_size`   | `max_connections` | 128 (was 8)  |
| `ssl_options` | `ssl`             | -            |

The old names still work but emit a `DeprecationWarning` (marked "As of release 4.0.0, ...") and should be migrated. `max_connections` is now also applied to the aiohttp `TCPConnector` `limit`, so the transport layer reflects the option in addition to sizing the connection pool.

* **`compression`** is a new option defaulting to `'deflate'` (on); the driver advertises `Accept-Encoding: deflate` by default. Set `compression='none'` to disable, which also suppresses aiohttp's automatic `Accept-Encoding` injection so compression is not silently negotiated. Defaulting on is a deliberate deviation from the proposal's agreed default-off, applied consistently across the GLVs by later agreement.
* **`connect_timeout`** defaults to 5s. Combined with the existing `read_timeout`, it is wired into a single aiohttp `ClientTimeout` via the socket-level `sock_connect`/`sock_read` knobs (not a whole-request `total`), so long but legitimate streaming responses are not aborted while a stalled server no longer hangs forever.
* The minimum supported **`aiohttp`** is raised to `3.11` (required for the `socket_factory` used by `keep_alive_time`).

## New options
* **`connect_timeout`** (5s) - socket-connect timeout (the existing `read_timeout` was rewired into the unified `ClientTimeout`).
* **`idle_timeout`** (180s) - mapped to the aiohttp `TCPConnector` keep-alive timeout.
* **`keep_alive_time`** (30s) - enables TCP keep-alive probes via the connector socket factory (`TCP_KEEPIDLE`/`TCP_KEEPALIVE`, guarded by platform availability).
* **`compression`** (`'none'`/`'deflate'`, default `'deflate'`) - the wire compression negotiated with the server.
* **`default_batch_size`** (64) - connection-level default that fills the per-request `batchSize` when unset.
* **`proxy`** and **`trust_env`** - explicit HTTP proxy and environment-trust options for the aiohttp transport.
* **`auth.sigv4`** gains a credentials-provider variant accepting an optional credentials provider or callable, falling back to the AWS environment variables.

## Removed (breaking)
* **`max_content_length`** - previously accepted but discarded.
* **`headers`** kwarg on `Client`/`DriverRemoteConnection` - custom headers must now be set via an interceptor.

## Bug fix
* Fixed `Client.submit`/`submit_async` mutating a caller-supplied `RequestMessage` in place; the message fields are now cloned before applying `request_options`/`default_batch_size`, so resubmitting the same message no longer accumulates state (matching the no-mutate contract of the .NET/JS drivers). This is required for the new `default_batch_size` to fill safely.

## Testing
* `gremlin-python` unit tests pass, including new `test_client_options`, `test_connection_options`, and `test_transport_compression` suites. The compression suite uses an in-process aiohttp loopback server to verify real `Accept-Encoding` negotiation and transparent deflate decompression over a socket.
* CHANGELOG, reference config table (`gremlin-variants.asciidoc`), and upgrade guide (`release-4.x.x.asciidoc`) updated for the Python slice.

## Notes
* The minimum `aiohttp` is raised from `3.8` to `3.11` (`pyproject.toml`). This is a dependency-floor bump, required for the `TCPConnector(socket_factory=...)` API that backs `keep_alive_time`; users pinned to `aiohttp` 3.8-3.10 must upgrade.


gremlin-python/src/main/python/gremlin_python/driver/client.py
```diff
@@ -38,21 +38,34 @@ def cpu_count():
 
 class Client:
 
-    def __init__(self, url, traversal_source, pool_size=None, max_workers=None,
+    def __init__(self, url, traversal_source, max_connections=None, max_workers=None,
```
kenhuuu: This is bit of a legacy question, but any reason why we don't just set the defaults here rather than checking for None later and setting them after?

docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -3188,13 +3190,21 @@ can be passed to the `Client` or `DriverRemoteConnection` instance as keyword ar
...
-|interceptors |The request interceptors to run after request serialization.|`None`
+|interceptors |The request interceptors to run before the request body is serialized.|`None`
 |auth |An authentication interceptor. Always appended to the end of the interceptor list so it runs last. |`None`
-|pool_size |The number of connections used by the pool. |4
+|max_connections |The maximum number of connections used by the pool. `pool_size` is accepted as a deprecated alias. |128
+|connect_timeout |Timeout in seconds for establishing the connection (TCP connect plus TLS handshake). |5
+|read_timeout |Per-read idle timeout in seconds applied while streaming a response. Resets per chunk. |`None`
+|write_timeout |Timeout in seconds for writing a request to the transport. |`None`
+|ssl |An `ssl.SSLContext` used for TLS connections. `ssl_options` is accepted as a deprecated alias. |`None`
+|idle_timeout |How long in seconds an idle connection remains in the pool before being closed. |180
```
kenhuuu: This works differently from how it does in Java. May need to make this more consistent.

gremlin-python/src/main/python/tests/unit/driver/test_transport_compression.py
```diff
@@ -0,0 +1,178 @@
...
+        except asyncio.IncompleteReadError as e:
+            out.extend(e.partial)
+            break
+        if not chunk:
+            break
+        out.extend(chunk)
+    return bytes(out)
+
+
+class TestCompressionWireBehavior:
```
kenhuuu: I'm not sure what value these tests bring. The GremlinServer already supports deflate so if we just enabled compression during integration tests, it already verifies that the driver can properly handle deflate compression. The only thing that needs to be verified is that the request header requests for compression.

gremlin-python/src/main/python/gremlin_python/driver/client.py
GumpacG: Updated now to set the default here as old deprecated options are removed

gremlin-python/src/main/python/tests/unit/driver/test_transport_compression.py
GumpacG: Removed as these cases are covered in test_client_deflate_compression_round_trip and test_connection_options.py.

docs/src/reference/gremlin-variants.asciidoc
GumpacG: As discussed offline, Java had the inconsistent behaviour and is now fixed. This will remain for Python.

kenhuuu:
VOTE +1

Cole-Greer (commented):
VOTE +1

xiazcy:
VOTE +1

PR #3471, opened 2026-06-19
TINKERPOP-3252 Replace Transaction.open() with idempotent begin()

kenhuuu:
https://issues.apache.org/jira/browse/TINKERPOP-3252

open() and begin() were redundant ways to start a transaction, and the strict "throw if already open" contract was incompatible with the embedded AUTO behavior: a read opens the transaction implicitly, so a later explicit begin() would throw even though the caller did nothing wrong. Collapsing to a single idempotent begin() makes explicit and implicit opens compose, and gives one consistent transaction-start verb across embedded, remote, and all GLVs.

close() is made idempotent for the same reason — so the common try-with-resources / double-close patterns are safe rather than surprising.

The base AbstractTransaction.begin() now opens via a guarded doOpen() so the contract holds for every provider (not just TinkerGraph) and MANUAL mode is no longer broken in the base class.

VOTE +1

### DIFF
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -135,9 +135,10 @@ in relation to the usage convention and graph provider caveats alluded to earlie
 
 Focusing on remote contexts first, note that it is still possible to issue traversals from `g`, but those will
 operate as implicit transactions (as opposed to the explicit transaction opened by `gtx`) and simply behave as
-self-contained units of work (i.e. one traversal is one implicit transaction). Each explicit transaction requires its
-own `Transaction` object. Multiple `begin()` calls on the same `Transaction` object are not permitted and will throw
-an `IllegalStateException`:
+self-contained units of work (i.e. one traversal is one implicit transaction). Each independent explicit transaction
+requires its own `Transaction` object obtained from `g.tx()`. Calling `begin()` more than once on the same
+`Transaction` object is idempotent - it does not start a new transaction and does not throw, returning a
+`GraphTraversalSource` bound to the already-open transaction:
 
 [source,java]
 ----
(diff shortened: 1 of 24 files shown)


docs/src/reference/the-traversal.asciidoc
```diff
@@ -213,8 +214,25 @@ occurs. `Transaction.READ_WRITE_BEHAVIOR` contains pre-defined `Consumer` functi
...
+does not start a new transaction and does not throw - it returns a `TraversalSource` bound to the transaction that is
+already open. This behavior is what allows `begin()` to coexist with `AUTO`. Under `AUTO`, a read or write implicitly
+opens a transaction, so an explicit `begin()` issued afterward would otherwise be operating on an already-open
+transaction; because `begin()` is idempotent, that call is safe rather than an error. Likewise, `close()` is
+idempotent - closing a transaction that is not open is a no-op.
+
+How `begin()` behaves once a transaction has been closed depends on the transaction model, and the two reference
+models differ here intentionally:
+
+* *Embedded* transactions are typically thread-bound and reusable. After a `commit()` or `rollback()`, the same
```
spmallette: "typically" - using this word because providers might have different semantics in embedded mode? if so, i think that should be spelled out clearly.

docs/src/reference/the-traversal.asciidoc
```diff
@@ -276,7 +294,8 @@ or other read operations executed in the context of that open transaction.
 <4> Calling `commit` finalizes the transaction.
 <5> Change transaction behavior to require manual control.
 <6> Adding a vertex now results in failure because the transaction was not explicitly opened.
-<7> Explicitly open a transaction.
+<7> Explicitly begin a transaction with `begin()`, which returns a `GraphTraversalSource` bound to it. Calling
```
spmallette: "begin" 2x - "start a transaction with `begin()`"? Do we need that second sentence here as part of the callout to describe this line of code. it's sort of answering a question no one asked over here....

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -151,6 +151,22 @@ Key behaviors consistent across all GLVs:
 See the <<gremlin-drivers-variants,Gremlin Drivers and Variants>> reference documentation for language-specific
 syntax and examples.
 
+==== `Transaction.open()` Replaced by `begin()`
+
+The `open()` method has been removed from the `Transaction` API. Use `begin()` instead, which is now the single
+transaction-start method for both embedded and remote contexts. Replace any `tx.open()` or `g.tx().open()` calls with
+`begin()`. This is a compile-time break and is straightforward to find and fix.
+
+In addition, `begin()` is idempotent: calling it when a transaction is already open does not start a new transaction
```
spmallette: the colon and semicolon use is sort of weird. use multiple sentences or find some good transition words.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -151,6 +151,22 @@ Key behaviors consistent across all GLVs:
...
+and does not throw, returning a `TraversalSource` bound to the existing transaction. For embedded graphs this replaces
+the previous behavior where opening an already-open transaction threw `transactionAlreadyOpen()`; that exception
```
spmallette: Do we need to say: 

> that exceptionfactory (`Transaction.Exceptions.transactionAlreadyOpen()`) has been removed,

That detail points to something internal to TinkerPop. Doesn't really seem relevant to me how the exception raises - it's just relevant that the exception isn't thrown.

xiazcy:
VOTE +1

Cole-Greer:
VOTE +1

PR #3472, opened 2026-06-19
Standardized `gremlin-js` connection options

GumpacG:
# Standardize `gremlin-javascript` connection options

Implements the JavaScript portion of the TinkerPop 4.x GLV connection-options standardization. Adopts `undici` as a pinned dependency and builds an explicit dispatcher so the standardized options take effect, adds a set of new options, removes dead WebSocket-era fields, and removes `headers`. JavaScript driver changes only; the other GLVs follow in separate PRs.

## New options
* **`maxConnections`** (128) - caps concurrent connections per origin on the default dispatcher (previously uncapped). *(breaking)*
* **`readTimeoutMillis`** - per-read idle (body) timeout in ms (undici `bodyTimeout`); resets per chunk, so it is streaming-safe.
* **`maxResponseHeaderBytes`** - maximum response header size in bytes (undici `maxHeaderSize`).
* **`keepAliveTimeMillis`** (30000 ms) - idle time before TCP keep-alive probes begin; enables `SO_KEEPALIVE` via a custom undici connector. Set `0` to disable.
* **`proxy`** - HTTP proxy URI; routes requests through an undici `ProxyAgent`.
* **`compression`** (`'none'`/`'deflate'`, default `'deflate'`) - the wire compression negotiated with the server. *(breaking)*
* **`batchSize`** (64) - connection-level default that fills a request's `batchSize` when unset.
* **`bulkResults`** (default false) - connection-level default for `bulkResults`, applied to every request unless overridden per-request. The `DriverRemoteConnection` traversal path defaults to `true` regardless of this setting.
* **`logger`** - a logger object (with `debug`/`info`/`warn`/`error` methods) or a `(level, message, ...args)` callback. Disabled when unset.

* Adopted **`undici`** as a pinned dependency (`6.27.0`). The driver now builds and owns an explicit undici dispatcher from the discrete options above; previously the options were declared but not wired to anything.
* **`compression`** defaults to `'deflate'` (on), so the driver sends `Accept-Encoding: deflate` by default. Set `compression: 'none'` to disable; this also suppresses the auto-injected `Accept-Encoding` so compression is not silently negotiated. Defaulting on is a deliberate deviation from the proposal's agreed default-off, applied consistently across the GLVs by later agreement.
* **`maxConnections`** now caps concurrent connections at 128 (was uncapped).

## Removed (breaking)
* **`agent`**, **`ca`**, **`cert`**, **`pfx`**, and **`rejectUnauthorized`** TLS fields - they were declared (WebSocket-era leftovers) but never wired to the HTTP transport. TLS is now configured through the Node/undici runtime (`NODE_EXTRA_CA_CERTS`, `NODE_TLS_REJECT_UNAUTHORIZED`).
* **`headers`** - removed as a deprecated option.

## Bug fix
* Fixed `Client.submit()` so that an explicit `bulkResults: false` request option is forwarded to the server instead of being silently dropped.

## Notes
* Per the proposal, JS does not expose `connectTimeout`/`idleTimeout`/`writeTimeout` as driver options (the undici runtime defaults are used), and TLS is runtime-delegated rather than a driver option. No public `dispatcher` option is exposed; the dispatcher is built internally from the discrete options.

## Testing
* `gremlin-javascript` builds (dual CJS/ESM) and all unit tests pass, including new `client`, `connection`, `dispatcher`, and `logger` suites covering the option-to-undici mapping, compression negotiation, keep-alive wiring, and `bulkResults` precedence.

### DIFF
--- a/docs/src/upgrade/release-4.x.x.asciidoc
+++ b/docs/src/upgrade/release-4.x.x.asciidoc
@@ -32,6 +32,49 @@ complete list of all the modifications that are part of this release.
 
 === Upgrading for Users
 
+==== Standardizing JavaScript Connection Options
+
+TinkerPop 4.x standardizes connection option names and defaults across the GLVs. In `gremlin-javascript`, the driver
+now adopts `undici` as a pinned dependency that ships a default dispatcher, several dead options have been removed, the
+standalone `headers` option has been deprecated, and a number of new options have been added. The notes below describe
+the JavaScript changes. See <<glv-driver-changes, GLV Driver Changes>> for the equivalent changes in the other
+drivers.
+
+Renames (deprecated aliases). The following option has been deprecated. It still works but should be migrated:
+
+- `headers` is deprecated. The standalone `headers` option is now implemented via a synthesized interceptor, so custom
+headers continue to work without re-introducing dead configuration. Prefer setting custom headers via an interceptor
+directly, e.g. `interceptors: (req) => { req.headers['X-Custom'] = 'value'; }`. When the `headers` option is set, it is
+applied by an interceptor that runs before the auth interceptor, so SigV4 still signs over those headers.
+
+Behavior changes. These change runtime behavior on upgrade, even if you do not change your configuration:
+
+- `compression` now defaults to `'deflate'` (on), so the driver sends `Accept-Encoding: deflate` by default. It is a
+`'none'`/`'deflate'` string union. Set `'none'` to disable it.
+- `maxConnections` now caps concurrent connections per origin at 128 by default, where the number of connections was
+previously uncapped.
+
+New options:
+
+- `readTimeout`: a per-read idle (body) timeout in milliseconds applied to the default dispatcher (undici
+`bodyTimeout`). It resets per chunk, so it is safe for streaming.
+- `maxResponseHeaderBytes`: the maximum size of the response headers in bytes applied to the default dispatcher (undici
+`maxHeaderSize`).
+- `keepAliveTime` (default 30000 ms): the idle time before TCP keep-alive probes begin, applied to the default
+dispatcher via a `SO_KEEPALIVE` connector. Set 0 to disable.
+- `proxy`: an HTTP proxy URI that routes requests through an undici `ProxyAgent`.
+- `defaultBatchSize` (default 64): a connection-level default that fills a request's `batchSize` when it is left unset.
+- `bulkResults` (default false): a connection-level default for `bulkResults` applied to every request unless
+overridden per-request. The `DriverRemoteConnection` traversal path defaults to `true` regardless of this setting.
+- `logger`: a logger object (with `debug`/`info`/`warn`/`error` methods) or a `(level, message, ...args)` callback.
+Logging is disabled when unset.
+
+The dead `agent`, `ca`, `cert`, `pfx`, and `rejectUnauthorized` fields have been removed. They were declared but never
+wired. TLS is now configured through the Node/undici runtime (for example `NODE_EXTRA_CA_CERTS` or
+`NODE_TLS_REJECT_UNAUTHORIZED`). *(breaking)*
+
+See: link:https://lists.apache.org/thread/yqtr2wnb1kq2pqqq4002cz511q5o0bkg[dev list discussion on standardizing GLV connection options].
+
 ==== Standardizing .NET Connection Options
 
 TinkerPop 4.x standardizes connection option names and defaults across the GLVs. In `gremlin-dotnet`, several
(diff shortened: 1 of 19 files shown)


gremlin-js/gremlin-javascript/package.json
```diff
@@ -44,7 +44,8 @@
   "dependencies": {
     "antlr4ng": "3.0.16",
     "buffer": "^6.0.3",
-    "eventemitter3": "^5.0.1"
+    "eventemitter3": "^5.0.1",
+    "undici": "6.27.0"
```
kenhuuu, line 52: A community member spent quite a bit of effort trying to make the javascript driver work in the browser. Will this dependency make it impossible for it to work in the browser moving forward?

spmallette: not sure what's possible, but can we produce a smoke test of some sort that would help defend that functionality?

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -32,6 +32,49 @@ complete list of all the modifications that are part of this release.
 
 === Upgrading for Users
 
+==== Standardizing JavaScript Connection Options
+
+TinkerPop 4.x standardizes connection option names and defaults across the GLVs. In `gremlin-javascript`, the driver
+now adopts `undici` as a pinned dependency that ships a default dispatcher, several dead options have been removed, the
```
Cole-Greer: Nit: I would avoid framing the removed options as "dead options". They were indeed unwired in 4.0.0-beta.2, but from the perspective of an upgrade from 3.8.x, they were active options which are now removed.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -32,6 +32,49 @@ complete list of all the modifications that are part of this release.
...
+- `keepAliveTime` (default 30000 ms): the idle time before TCP keep-alive probes begin, applied to the default
+dispatcher via a `SO_KEEPALIVE` connector. Set 0 to disable.
+- `proxy`: an HTTP proxy URI that routes requests through an undici `ProxyAgent`.
+- `defaultBatchSize` (default 64): a connection-level default that fills a request's `batchSize` when it is left unset.
+- `bulkResults` (default false): a connection-level default for `bulkResults` applied to every request unless
+overridden per-request. The `DriverRemoteConnection` traversal path defaults to `true` regardless of this setting.
+- `logger`: a logger object (with `debug`/`info`/`warn`/`error` methods) or a `(level, message, ...args)` callback.
+Logging is disabled when unset.
+
+The dead `agent`, `ca`, `cert`, `pfx`, and `rejectUnauthorized` fields have been removed. They were declared but never
```
Cole-Greer: Building on my comment above, `ca`, `cert`, `pfx` and `rejectUnauthorized` were active and used in 3.8, they should be described as removed or replaced. `agent` was introduced during my work on 4.0.0-beta.2 as part of the migration to the fetch API. I think that can be removed cleanly without any acknowledgement in the docs as it never existed in any stable release. There's also options such as `PingInterval` and `PongTimeout` which had previously been removed for beta.2, but it might be nice to acknowledge how new options such as `keepAliveTime` replace that role.

I had left the JS connection options in an unfortunate state in beta.2. It would be nice if this ignored that messy intermediate state and just gave the clean picture of the 3.8->4 change.

gremlin-js/gremlin-javascript/lib/driver/connection.ts
```diff
@@ -99,10 +163,39 @@ export default class Connection extends EventEmitter {
       throw new TypeError('interceptors must be a function, array, or undefined');
     }
 
+    // The deprecated `headers` option is implemented as a synthesized interceptor so custom
+    // headers still work without re-introducing dead configuration. It is appended after user
```
Cole-Greer: It may be preferable to push the headers interceptor before the users interceptors, it would prevent us from worrying about any changes to the request message format made by the users custom interceptors, and it also would ensure that the new headers are visible from within the users interceptors in case they need them for any purpose.

gremlin-js/gremlin-javascript/package.json
GumpacG: I added dispatcher.browser.ts for browsers and throw if users try to set the unsupported connection options in a browser. 

I tested this manually with example in examples/browser. As for a smoke test, I created a follow up ticket https://issues.apache.org/jira/browse/TINKERPOP-3260 to implement this.

gremlin-js/gremlin-javascript/lib/driver/connection.ts
GumpacG: Headers are deprecated and removed so this is no longer applicable.

kenhuuu:
VOTE +1

Cole-Greer:
According to the build failures, it looks like there's some minor updates missing in Gremlin-MCP. Other than that, this LGTM.

VOTE +1 (pending mcp updates/build fix)

spmallette:
VOTE +1

PR #3479, opened 2026-06-22
Remove references to GraphSON4.

kenhuuu:
GraphSON4 was enabled in gremlin-driver/gremlin-python for the beta1 release but should be removed prior to the actual release as 4.x will only officially support GraphBinary.

VOTE +1


Cole-Greer:
VOTE +1

spmallette:
VOTE +1

PR #3485, opened 2026-06-26
Implement PrimitivePDT

Cole-Greer:
## Summary

Implements **PrimitivePDT**, the provider-defined *primitive* type reserved in the
GraphBinary (`0xF1`) and GraphSON (`g:PrimitivePdt`) V4 specs, completing the
Provider Defined Types feature begun with CompositePDT in #3433.

A PrimitivePDT represents a value as a single **opaque stringified value** plus a
provider type name — for types that have no native TinkerPop representation but can
be expressed as one string (e.g. an unsigned 32-bit integer, a WKT geometry, a
provider-specific identifier). It complements CompositePDT (a type name + a map of
fields) for structured types.

Wire/text forms:
- **GraphBinary** `0xF1`: `{type}{value}` — two fully-qualified Strings; the value is
  opaque and never parsed by TinkerPop.
- **GraphSON** `g:PrimitivePdt`: `{"type": "...", "value": "..."}` with an untyped
  string value (response-only in V4).
- **gremlin-lang**: `PDT("name", "value")` — an overload of the composite
  `PDT("name", [map])` literal (the second argument's type selects the variant; this
  is unambiguous to the parser).

Provider integration is **adapter-based**: a provider registers a primitive adapter
(`toValue`/`fromValue`) so the driver auto-dehydrates a native object on send and
auto-hydrates it on receive, with no per-call configuration. A registered adapter
takes precedence over the `@ProviderDefined` annotation / `[ProviderDefined]`
attribute on dehydration, consistent across GLVs.

## What's included

**gremlin-core / gremlin-util (Java)**
- `PrimitiveProviderDefinedType` value type and `PrimitiveProviderDefinedTypeSerializer`
  (`DataType.PRIMITIVE_PDT` `0xF1`).
- New `PrimitivePDTAdapter`; extracted a common `ProviderDefinedTypeAdapter` supertype
  with `CompositePDTAdapter` / `PrimitivePDTAdapter` implementations and explicit
  composite/primitive registry accessors.
- Registry hydration (incl. primitive-nested-in-composite), GraphBinary reader/writer
  dispatch, and GraphSON `g:PrimitivePdt` ser/deser.
- gremlin-lang grammar overload + all translator visitors + `GremlinLang` text emission.
- **Prerequisite fix**: made the GraphBinary/GraphSON *write* path registry-aware so a
  type registered via an adapter (but not annotated) can be serialized on the
  server→client response path — a pre-existing CompositePDT gap.

**GLVs** — full support in **Python, JavaScript, Go, .NET** (and the Java driver):
value type, `0xF1` serializer, registry primitive path, gremlin-lang `PDT("name","value")`
emission + adapter auto-dehydration, and client wiring (reusing the existing PDT
registry plumbing).

**Server** — test fixtures (`Uint32`, `TinkerId`, and a composite `Measurement`
containing a primitive) and end-to-end integration coverage.

**Docs** — provider guide section, gremlin-lang literal reference, upgrade note, and
CHANGELOG entry. The GraphBinary/GraphSON spec sections already matched the
implementation.

## Key design decisions

- **Adapter-only** provider integration for primitives (no annotation path); a common
  adapter supertype unifies discovery while keeping composite/primitive distinct.
- **Grammar overload** of `PDT(...)` rather than a new keyword, following the existing
  literal-alternation convention.
- **Opaque value**: TinkerPop never parses the value; round-trip fidelity is preserved
  (leading zeros, large/non-numeric values), with the provider adapter owning
  parse/format.
- **Adapter precedence over annotation/attribute** on dehydration, matching the request
  and response paths.

## Testing

- **Unit tests** across all languages (serializer round-trip incl. opaque-value
  fidelity, registry hydration with graceful degradation, dual-registration rejection,
  gremlin-lang text emission, adapter-over-annotation/attribute precedence).
- **Per-GLV integration tests** against the Docker test server, covering the
  unregistered/raw round-trip, registered auto de/hydration, and
  primitive-nested-in-composite: Java, Python, JavaScript, Go, .NET.

VOTE +1


docs/src/dev/provider/index.asciidoc
```diff
@@ -1771,6 +1577,117 @@ serialization and the registry handles inbound reconstruction.
 For driver users consuming PDTs, see the <<gremlin-variants,Gremlin Variants>> reference documentation for
 each language driver.
 
+[[primitive-provider-defined-types]]
+==== Primitive Provider Defined Types
```
kenhuuu: Nit: this section doesn't flow well in the PDT section. It needs to be better integrated. The main problem I get reading this is the lack of introduction of primitive and composite PDT. That should be done right at the top as the rest of the section applies to both.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/DotNetTranslateVisitor.java
```diff
@@ -1255,11 +1209,19 @@ public Void visitDurationLiteral(final GremlinParser.DurationLiteralContext ctx)
...
-        visitGenericMapLiteral(ctx.genericMapLiteral());
-        sb.append(")");
+        if (ctx.genericMapLiteral() != null) {
+            sb.append("new ProviderDefinedType(");
+            sb.append(ctx.stringLiteral(0).getText());
+            sb.append(", ");
+            visitGenericMapLiteral(ctx.genericMapLiteral());
+            sb.append(")");
+        } else {
+            sb.append("new PrimitiveProviderDefinedType(");
```
kenhuuu: Its odd for composite PDT to be `ProviderDefinedType` and primitive to be `PrimitiveProviderDefinedType`. This suggests a IS-A relationship, which doesn't exist since primitive PDT is not a composite PDT.

gremlin-util/src/test/java/org/apache/tinkerpop/gremlin/util/ser/binary/GraphBinaryWriterPdtTest.java
```diff
@@ -22,8 +22,8 @@
 import org.apache.tinkerpop.gremlin.structure.io.Buffer;
 import org.apache.tinkerpop.gremlin.structure.io.binary.GraphBinaryReader;
 import org.apache.tinkerpop.gremlin.structure.io.binary.GraphBinaryWriter;
-import org.apache.tinkerpop.gremlin.structure.io.pdt.ProviderDefined;
-import org.apache.tinkerpop.gremlin.structure.io.pdt.ProviderDefinedType;
+import org.apache.tinkerpop.gremlin.structure.io.binary.TypeSerializerRegistry;
+import org.apache.tinkerpop.gremlin.structure.io.pdt.*;
```
kenhuuu: Nit: don't use wildcard imports

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/translator/DotNetTranslateVisitor.java
Cole-Greer: I'll rename these to `CompositePDT` and `PrimitivePDT` (or equivalents) in all GLVs.

kenhuuu:
VOTE +1

docs/src/dev/provider/index.asciidoc
Cole-Greer: I've updated the provider docs to better integrate composite and primitive PDTs together.

PR #3283, opened 2025-11-30
Add Java 25 build compatibility for TinkerPop 3.8

jrepp:
## Summary

Enables building TinkerPop with Java 25 by updating Groovy and fixing annotation processor configuration.

### Build Configuration
1. **`pom.xml`**: Updated `groovy.version` to `4.0.30-SNAPSHOT` (includes ASM 9.9 for Java 25 support)
2. **`gremlin-groovy/pom.xml`**: Added explicit `annotationProcessorPaths` for `GremlinDslProcessor`

### Test Fixes
3. **`LambdaRestrictionStrategy.java`**: Use case-insensitive lambda detection (Java 25 changed class naming from `$$Lambda$N/...` to `$$Lambda/0x...`)
4. **`ImportGremlinPluginTest.java`**: Handle `Math.TAU` field added in Java 19

## Build Instructions

```bash
mvn clean install -DskipTests -Denforcer.skip=true -Dmaven.javadoc.skip=true
```

**Prerequisites:** Local Groovy 4.0.30-SNAPSHOT with ASM 9.9 - see https://github.com/apache/groovy/pull/2346

## Test Results

| Module | Tests | Failures | Skipped |
|--------|-------|----------|---------|
| gremlin-core | 8500 | 0 | 45 |
| gremlin-server | 8723 | 0 | 24 |
| hadoop-gremlin | 846 | 3 | 700 |

**Note:** hadoop-gremlin failures are pre-existing Hadoop + Java 25 incompatibility issues, not related to this PR.

## Blocking Dependency

⚠️ **This PR cannot be merged until [apache/groovy#2346](https://github.com/apache/groovy/pull/2346) lands in a Groovy 4.x release.**

The linked Groovy PR updates ASM from 9.7 to 9.9, which is required to parse Java 25 class files (class file version 69). Without this ASM update, Groovy cannot compile or load classes built with Java 25, causing `IllegalArgumentException` errors during bytecode analysis.

Once that Groovy PR is released (likely as 4.0.25 or later), update the `groovy.version` in `pom.xml` from `4.0.30-SNAPSHOT` to the actual release version.

## Notes for Maintainers

- The `annotationProcessorPaths` change is backward compatible and follows Maven best practices
- The enforcer plugin's Java version restriction `[11,18)` will need updating for official Java 25 support

PR #3488, opened 2026-06-28
Partial Java 21/25 support for TinkerPop 3.8.x

porunov:
Add **partial** (experimental) support for building and running with Java 21 and Java 25.
All modules except `spark-gremlin` become compatible with Java 21 and Java 25, and both versions get their own GitHub Actions build job. The reference docs (supported-runtime table), developer docs (system configuration), 3.8.2 upgrade docs, and CHANGELOG note the new experimental support.

This may potentially help graph providers that don't use `spark-gremlin` to upgrade to Java 21 or 25.
Providers that do require `spark-gremlin` will not be able to upgrade past Java 17 on TinkerPop 3.x versions. Full Java 21/25 upgrade requires dropping Java 11 support. You can find the full upgrade in PR #3489 which targets TinkerPop 4.0.0.

The reason Spark can't be upgraded to a newer version on the `3.8-dev` branch is that no Spark version is compatible with both Java 11 and Java 21/25 — Java 21 support arrived in Spark 4.0 ([SPARK-43831](https://issues.apache.org/jira/browse/SPARK-43831)) and Java 25 support in Spark 4.1, and Spark 4.x requires Java 17 as a minimum (and Scala 2.13). Spark 3.3.x itself does not run on Java 21 or newer: its off-heap `Platform` class reflects on the private `DirectByteBuffer(long, int)` constructor that JDK 21 removed, so `Platform.<clinit>` fails with `NoSuchMethodException` (verified by running the `spark-gremlin` tests on JDK 21: 28 of 36 tests fail with `NoClassDefFoundError: Could not initialize class org.apache.spark.unsafe.array.ByteArrayMethods`).

Validated locally with a full `mvn clean install` (the same module set as the new CI job) on Java 21; all modules pass.

There was the first attempt to add Java 25 support in the PR #3283
This PR can be considered as continuation

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -37,6 +37,39 @@ jobs:
           distribution: 'temurin'
       - name: Build with Maven
         run: mvn clean install -pl $EXCLUDE_MODULES -Dci --batch-mode -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
+  java-jdk21:
+    name: mvn clean install - jdk21
+    timeout-minutes: 45
+    needs: smoke
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - name: Set up JDK 21
+        uses: actions/setup-java@v5
+        with:
+          java-version: '21'
+          distribution: 'temurin'
+      # spark-gremlin is excluded because Spark 3.3.x does not run on Java 21: its off-heap Platform initializer
+      # reflects on the private DirectByteBuffer(long, int) constructor that JDK 21 removed. Java 21 support arrived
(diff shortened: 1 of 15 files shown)


Cole-Greer:
I think we should also add a GitHub actions config for Java 21 as well, considering it is also an LTS. Would you be able to add that additional job?

I also think we should update the [reference docs](https://tinkerpop.apache.org/docs/current/reference/#gremlin-java) and [developer docs](https://tinkerpop.apache.org/docs/current/dev/developer/#system-configuration) to list "Experimental support for Java 21 and 25"

VOTE +1 (pending Java 21 testing and docs)

porunov:
Thanks @Cole-Greer for the review! Done and updated.

kenhuuu:
VOTE +1

PR #3489, opened 2026-06-29
Support Java 21/25 and Spark 4.1, raise minimum to Java 17

porunov:
Modernizes TinkerPop 4.0 to build and run on Java 21 and Java 25 and upgrades the Spark OLAP integration to Spark 4.1. Because Spark 4 requires Java 17 as a minimum, the project-wide minimum Java version is raised from 11 to 17. Every LTS release from the new minimum upward (17, 21, 25) is covered by CI.

Java 21/25 support:
- Groovy 4.0.25 -> 4.0.32 (bundled ASM parses Java 25 bytecode)
- StringFactory: detect the Java 21+ lambda class name format ($$Lambda/)
- LambdaRestrictionStrategy: match "lambda" case-insensitively
- ImportGremlinPluginTest: account for Math.TAU added in Java 19
- gremlin-groovy: declare annotationProcessorPaths explicitly (JDK 23+ no longer runs classpath-discovered processors) and widen the javadoc groovy-stubs profile to all supported JDKs
- gremlin-console: apply the neo4j-gremlin --add-opens flags for JDK 17+

Dependency upgrades:
- Hadoop 3.4.2 -> 3.4.3 (uses the Subject.current()/callAs() replacement APIs, so hadoop-gremlin runs on Java 25)
- Spark 3.5.4 -> 4.1.2, which is Scala 2.13 only, the first Spark line to support Java 25, and the second to support Java 21 (SPARK-43831 landed in 4.0). Switched spark-core_2.12 -> _2.13, ported the Scala 2.13 changes (JavaConversions -> CollectionConverters, WrappedArray -> immutable ArraySeq) and aligned Netty, jackson (via jackson-bom) and other transitive versions that Spark 4 requires
- Netty 4.1.125 -> 4.2.7 (required by Spark 4)

Java 17 baseline (breaking):
- enforcer requireJavaVersion [11,18) -> [17,26); compiler release 11 -> 17 (and gremlin-groovy/gremlin-annotations release 8 -> 17)
- build-test.yml: all jobs run on Java 17, plus new Java 21 and Java 25 jobs; the Java 11 job is removed and coverage runs on the Java 17 job
- Dockerfiles updated to Java 17; developer, reference, and upgrade docs list Java 17 as the minimum with Java 21 and 25 supported

Validated with a full `mvn clean install` on Java 17, Java 21, and Java 25 (all JVM modules, including hadoop-gremlin and spark-gremlin).

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -16,10 +16,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
-      - name: Set up JDK 11
+      - name: Set up JDK 17
         uses: actions/setup-java@v5
         with:
-          java-version: '11'
+          java-version: '17'
           distribution: 'temurin'
       - name: Build with Maven
         run: mvn clean install -DskipTests -Dci --batch-mode -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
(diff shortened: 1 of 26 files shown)


Cole-Greer:
I think we should also add a GitHub actions config for Java 21 as well, considering it is also an LTS. Would you be able to add that additional job?

VOTE +1 (pending Java 21 testing)

porunov:
Thanks @Cole-Greer for the review! Both JDK 21 tests and the documentation updates were added.

kenhuuu:
VOTE +1

PR #3490, opened 2026-06-29
Stabilize the flaky shouldWaitForAllResultsToArrive test

porunov:
`shouldWaitForAllResultsToArrive` asserted that its polling loop ran at least once (checked > 0). With HTTP streaming over a fast loopback, all nine results can already be available on the first allItemsAvailable() check, so the loop
body never runs and the assertion fails intermittently. Drop that timing-dependent assertion; the test now only asserts that all results ultimately arrive.

### DIFF
--- a/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
+++ b/gremlin-server/src/test/java/org/apache/tinkerpop/gremlin/server/GremlinDriverIntegrateTest.java
@@ -645,15 +645,15 @@ public class GremlinDriverIntegrateTest extends AbstractGremlinServerIntegration
         try {
             final Client client = cluster.connect();
 
-            final AtomicInteger checked = new AtomicInteger(0);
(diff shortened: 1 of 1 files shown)


kenhuuu:
Thanks for fixing these.  The fix for `shouldOpenAndCloseManyTransactionsSequentially()` is actually showing that there is a bug in the `gremlin-driver` transaction implementation as its supposed to be blocking so calling `all()` shouldn't be necessary. I'm actually going to fix this right now so please revert the change for that test.

VOTE +1

kenhuuu:
https://github.com/apache/tinkerpop/pull/3491 should solve flakiness in `shouldOpenAndCloseManyTransactionsSequentially`

porunov:
Thanks for the review @kenhuuu ! Reverted changes in `shouldOpenAndCloseManyTransactionsSequentially `, so this PR is addressing only `shouldWaitForAllResultsToArrive `

Cole-Greer:
VOTE +1

I'm going to merge as a CTR, I don't see the need for additional reviewers.

PR #3491, opened 2026-06-29
Make Java RemoteTransaction submit() wait for server response headers

kenhuuu:
This is a case where implementation didn't match documented and expected behavior. Transactions need to be semi-blocking in order for them to reach the server in the exact same order. It was incorrectly assumed that the Java GLV already did this so this just changes it to actually do that.

VOTE +1


Cole-Greer (commented):
VOTE +1

PR #3492, opened 2026-06-29
Rename configuration to responseSerializer.

kenhuuu:
This makes it consistent among all the GLVs and ensures that users know that is for response size deserialization only and not request serialization.

VOTE +1


docs/src/reference/gremlin-variants.asciidoc
```diff
@@ -1997,7 +1997,7 @@ can be passed in the constructor of a new `Client` or `DriverRemoteConnection` :
 |options.interceptors |RequestInterceptor/RequestInterceptor[] |One or more functions that can modify the HTTP request before it is sent, run in order. |undefined
 |options.auth |RequestInterceptor |An auth interceptor that is always appended to the end of the interceptor list so it runs last. |undefined
 |options.preciseNumbers |Boolean |When `true`, wraps deserialized numbers in typed wrappers that preserve the server's original type. |undefined
-|options.reader |GraphBinaryReader |The reader to use for deserializing responses. |GraphBinaryReader
+|options.responseSerializer |GraphBinaryReader |The reader to use for deserializing responses. |GraphBinaryReader
```
spmallette: i suppose we say somewhere else that "serializing requests isn't configurable because...." - that might be useful information to folks who come from 3.x and are used to `serializer` working the way it used to.

CHANGELOG.asciidoc
```diff
@@ -26,12 +26,12 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
...
-** Java: renamed `maxConnectionPoolSize`/`connectionSetupTimeoutMillis`/`idleConnectionTimeoutMillis`/`resultIterationBatchSize` and `RequestOptions.addG`->`traversalSource`; removed `maxResponseContentLength` (responses now stream); added `readTimeoutMillis`, `keepAliveTimeMillis`, `maxResponseHeaderBytes`, `proxy`, `url(String)`, `ssl(SslContext)`. *(breaking)*
+** Aligned option names across drivers: `maxConnections` (128), `connectTimeoutMillis` (5000), `readTimeoutMillis` (off), `idleTimeoutMillis` (180000), `keepAliveTimeMillis` (30000), `compression` (on/`deflate`), `batchSize` (64), `bulkResults` (false), `maxResponseHeaderBytes`, `proxy`, `ssl`, and `responseSerializer`. Timeouts use a millisecond-suffixed canonical name with an idiomatic duration companion (Java `Duration`, Go `time.Duration`, .NET `TimeSpan`, Python seconds); JavaScript uses milliseconds only.
```
spmallette, line 33: not your change, but i'd rather we not use multi-level bullets and multiple sentences per bullet in CHANGELOG. something to generally fix before release cc/ @Cole-Greer as the presumed release manager for 4.0.0-beta.3

spmallette:
Called out a couple of items, but not blocking: VOTE +1

CHANGELOG.asciidoc
Cole-Greer: I can ensure these get de-nested and slimmed out prior to release.

Cole-Greer (approved):
VOTE +1

docs/src/reference/gremlin-variants.asciidoc
kenhuuu: This is actually a good find. The PR that changed this was actually missing this information in the documentation (it only mentioned the usage of RequestInterceptor but didn't explicitly say that serializer options were going away). I'll add it to upgrade/reference documentation to make it clear.

PR #3493, opened 2026-06-29
Speed up Python GraphBinary deserialization

kirill-stepanishin:
The GraphBinary reader built a `DataType` enum member from the type byte for every object it decoded. That per-object enum construction heavily degrades deserialization performance on large result sets.

The reader now builds a `{type code: deserializer}` lookup table once up front and dispatches on the raw integer instead, avoiding per-object enum construction. Behavior is unchanged: an unknown type code still raises `ValueError("... is not a valid DataType")`.

### Performance

Benchmarked on two cross-region EC2 instances (server in US-EAST-2, client in US-WEST-2) to capture realistic network latency, against the Modern graph over GraphBinary V4 on Python 3.11. Each query was run with and without this change, alternating back to back across 3 sweeps, reporting the median.

| Query | Before | After | Change |
|---|---|---|---|
| `g.V().repeat(both()).times(12)` (~200k results) | 7.97 s | 5.85 s | 26% faster |
| `g.V()` (6 results) | 0.107 s | 0.109 s | no change |

The improvement is significant on large result sets, where per-object deserialization cost dominates, and scales with the number of objects returned.

### DIFF
--- a/gremlin-python/src/main/python/gremlin_python/structure/io/graphbinaryV4.py
+++ b/gremlin-python/src/main/python/gremlin_python/structure/io/graphbinaryV4.py
@@ -154,7 +154,7 @@ class GraphBinaryReader(object):
         self.pdt_registry = pdt_registry
         # Mirror of self.deserializers keyed by int type code instead of DataType.
         # Avoids the per-read DataType(bt) call, whose aenum construction negatively affects performance on large results.
-        self._objectify_by_code = {dt.value: des.objectify for dt, des in self.deserializers.items()}
+        self._deserializer_by_type_code = {dt.value: des.objectify for dt, des in self.deserializers.items()}
 
     def read_object(self, b):
         if b is None:
@@ -171,7 +171,7 @@ class GraphBinaryReader(object):
                     buff.read(1)
                 return None
             try:
-                objectify = self._objectify_by_code[bt]
+                objectify = self._deserializer_by_type_code[bt]
             except KeyError:
                 raise ValueError("%r is not a valid DataType" % bt) from None
             result = objectify(buff, self, nullable)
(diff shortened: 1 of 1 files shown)


gremlin-python/src/main/python/gremlin_python/structure/io/graphbinaryV4.py
```diff
@@ -149,6 +152,9 @@ def __init__(self, deserializer_map=None, pdt_registry=None):
         if deserializer_map:
             self.deserializers.update(deserializer_map)
         self.pdt_registry = pdt_registry
+        # Mirror of self.deserializers keyed by int type code instead of DataType.
+        # Avoids the per-read DataType(bt) call, whose aenum construction negatively affects performance on large results.
+        self._objectify_by_code = {dt.value: des.objectify for dt, des in self.deserializers.items()}
```
kenhuuu: Nit: `_objectify_by_code` naming was hard for me to understand. Without reading through the entire implementation, I couldn't tell this was a typecode to deserialization function dispatch table. Maybe that could be clearer in the name? Something like `code_to_objectify_function` or `objectify_function_dispatch_table`? This is really small so if you feel like it doesn't need to be changed then don't change it.

kenhuuu:
VOTE +1

gremlin-python/src/main/python/gremlin_python/structure/io/graphbinaryV4.py
kirill-stepanishin: Renamed the map to `_deserializer_by_type_code`, I think it makes more sense now. Thanks for the suggestion!

Cole-Greer (approved):
VOTE +1

kenhuuu:
Reached lazy consensus

PR #3500, opened 2026-07-01
Restrict trigger push branch for GitHub Workflow

apupier:
Feature branches rarely need their own CI runs: the code is already tested when a pull request is opened against a release branch. If the push trigger has no branch restriction and pull_request is also configured, every push to a branch with an open PR runs the workflow twice: once for the push and once for the PR synchronisation.

Always give the push trigger an explicit list of branches: this stops branches created from a release branch from inheriting its workflow runs.

see https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=430408443#GitHubActionsRecommendedPractices-Restrictthepushtriggertospecificbranches

ignoring only dependabot branches as it seems that several feature branches are created on the main Apache tinkerpop repository. The dependabot case should already cut most if the duplicate workload

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -1,5 +1,9 @@
 name: build-test
-on: [push, pull_request]
+on:
+  push:
+    branches-ignore:
+      - dependabot/**
+  pull_request:
 env:
   # modules commonly excluded from builds as they have their own independent non-JVM setups and can be run in parallel.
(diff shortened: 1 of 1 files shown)


Cole-Greer:
Thanks @apupier, I don't have concerns with restricting the `on push` actions from only dependabot branches.

VOTE +1

kenhuuu:
VOTE +1

PR #3501, opened 2026-07-01
Fix doFinalReduction to remove keys 

theneelshah:
Fix doFinalReduction to remove keys when post-barrier steps produce nothing. This fixes the empty keys in the output. Only the keys with values show up.


Cole-Greer:
Hi @theneelshah,

If I'm understanding the change correctly, the new behaviour somewhat contradicts the currently documented [group() step semantics](https://github.com/apache/tinkerpop/blob/5aae2577c29c7af7e295ef0860eb6eb5d4816cb8/docs/src/dev/provider/gremlin-semantics.asciidoc?plain=1#L1665-L1704). These docs will need updating if we are to proceed. Could you share some extra details on what use case is motivating this change?

Also if we proceed here, I think this change likely warrants adding a new scenario to [Group.feature](https://github.com/apache/tinkerpop/blob/master/gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/sideEffect/Group.feature) as well as a quick `CHANGELOG` entry.

Cole-Greer:
After taking a closer look and running some test queries in this branch, I think I misunderstood the scope of this change. It's more narrow than I previously thought and I no longer think it's in violation of the documented semantics. I now understand it to be a bug fix for a specific edge case.

```
// 3.8.1
gremlin> g.V().group().by("name").by("age")
==>[ripple:[],peter:[35],vadas:[27],josh:[32],lop:[],marko:[29]]
gremlin> g.V().group().by(values("name")).by(values("age"))
==>[peter:35,vadas:27,josh:32,marko:29]
gremlin> g.V().group().by(values("name")).by(values("age").fold().unfold())
==>[ripple:[],peter:[35],vadas:[27],josh:[32],lop:[],marko:[29]]

// This Branch
gremlin> g.V().group().by("name").by("age")
==>[ripple:[],peter:[35],vadas:[27],josh:[32],lop:[],marko:[29]]
gremlin> g.V().group().by(values("name")).by(values("age"))
==>[peter:35,vadas:27,josh:32,marko:29]
gremlin> g.V().group().by(values("name")).by(values("age").fold().unfold())
==>[peter:35,vadas:27,josh:32,marko:29]
```

I no longer think any docs changes are warranted here, I think all that is needed is a feature test for `g.V().group().by(values("name")).by(values("age").fold().unfold())` (or something less contrived than my fold().unfold() if you have something like that), as well as a quick changelog entry.

PR #3502, opened 2026-07-01
Remove Kerberos support

spmallette:
Kerberos (SASL) authentication is not supported in 4.x. Removes the Krb5Authenticator, the kerby-based KDC test fixture, related configs, and the kerby/wildfly-common dependencies. Also drops the now-dead Kerberos server port (45942) and KDC port (4588) from the GLV docker test setup, and the libkrb5 build packages.

VOTE +1


Cole-Greer:
VOTE +1

kenhuuu:
VOTE +1

PR #3503, opened 2026-07-02
TINKERPOP-3262 Rename request "bindings" field to "parameters"

kenhuuu:
https://issues.apache.org/jira/browse/TINKERPOP-3262

Standardize on "parameter" for the gremlin-lang query parameter map across the wire (JSON/GraphBinary/GraphSON), server internals and error messages, and the public driver APIs of all five GLVs.

Bounded to that one concept. "binding" is kept where it names something else, since renaming those would be wrong rather than churn: javax.script.Bindings and its ecosystem (BindingsCustomizer, script-engine and global TraversalSource bindings), where the server still hands parameters to a javax.script.Bindings for evaluation as a boundary with a foreign API; GQL pattern-match bindings in gql-gremlin; the gremlin-lang grammar's "variable" vocabulary (VariableResolver); and gremlin-core's step Parameters, disambiguated in docs as "step parameters".

Review Note:
This is mostly just a search and replace of the word "binding" for "parameter". If you are reviewing this, all you need to really do is decide whether the boundaries of the renaming were correct. Did it go too far or not far enough.

VOTE +1


spmallette:
VOTE +1

Cole-Greer:
VOTE +1

PR #3504, opened 2026-07-02
Speed up Python GraphBinary deserialization (3.7)

kirill-stepanishin:
The GraphBinary reader built a `DataType` enum member from the type byte for every object it decoded. That per-object enum construction heavily degrades deserialization performance on large result sets.

The reader now builds a `{type code: deserializer}` lookup table once up front and dispatches on the raw integer instead, avoiding per-object enum construction. Behavior is unchanged: an unknown type code still raises `ValueError("... is not a valid DataType")`.

### Performance

Benchmarked on two cross-region EC2 instances (server in US-EAST-2, client in US-WEST-2) to capture realistic network latency, against the Modern graph over WebSocket on Python 3.11. Each query was run with and without this change, alternating back to back across 3 sweeps, reporting the median.

| Query | Before | After | Change |
|---|---|---|---|
| `g.V().repeat(both()).times(12)` (~200k results) | 6.81 s | 5.36 s | 21% faster |
| `g.V()` (6 results) | 0.181 s | 0.181 s | no change |

The improvement is significant on large result sets, where per-object deserialization cost dominates, and scales with the number of objects returned.

### DIFF
--- a/gremlin-python/src/main/python/gremlin_python/structure/io/graphbinaryV1.py
+++ b/gremlin-python/src/main/python/gremlin_python/structure/io/graphbinaryV1.py
@@ -118,6 +118,9 @@ class DataType(Enum):
 
 NULL_BYTES = [DataType.null.value, 0x01]
 
+# null type code as a plain int, so the per-read null check skips the aenum lookup
+_NULL = DataType.null.value
+
 
 def _make_packer(format_string):
     packer = struct.Struct(format_string)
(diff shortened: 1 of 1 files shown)


kenhuuu:
VOTE +1

NOTE: a master branch version of this has already been merged. Make sure to use use `-s ours` on a merge up to master.

Cole-Greer:
VOTE +1

PR #3507, opened 2026-07-03
Consolidate gremlin-python read timeout into a single ReadTimeoutError

GumpacG:
gremlin-python armed two read-timeout mechanisms at the same value - the driver's
`async_timeout` wrapper (raising `asyncio.TimeoutError`) and aiohttp's `sock_read`
(raising `SocketTimeoutError`). Whichever fired first was a race, so a read timeout
surfaced nondeterministically as one type or the other, which also made
`test_client_side_timeout_set_for_aiohttp` flaky (it asserted on the exception message).

Change: remove the redundant `async_timeout` read wrapper and keep aiohttp `sock_read`
as the single read-timeout source (it also covers the initial-response wait and is
streaming-safe). Normalize its `SocketTimeoutError`/`ServerTimeoutError` into a new
driver-owned `ReadTimeoutError` so a read timeout always surfaces as one deterministic,
transport-agnostic type. `ReadTimeoutError` subclasses the builtin `TimeoutError`, so it
stays catchable via `except TimeoutError` while leaking neither asyncio nor aiohttp types
(addresses review feedback). It is documented as temporary: once the driver is fully
async it should revert to `asyncio.TimeoutError` (TINKERPOP-2774).
  

gremlin-python/src/main/python/gremlin_python/driver/aiohttp/transport.py
```diff
@@ -71,6 +71,17 @@ def _normalize_compression(compression):
...
 
 
+def _run_read(loop, read_timeout, coro):
+    """Run a response-read coroutine on ``loop``, normalizing aiohttp's read timeout
+    (SocketTimeoutError / ServerTimeoutError) into a plain ``asyncio.TimeoutError`` so a
+    read timeout always surfaces as one deterministic, library-agnostic type."""
+    try:
+        return loop.run_until_complete(coro)
+    except aiohttp.ServerTimeoutError as e:
+        raise asyncio.TimeoutError(
```
kenhuuu: I'm not sure this is the right decision. The driver is currently synchronous, so have users "except/catch" or see an asyncio error might be a bit odd. That said, returning aiohttp internals also probably not great. So maybe we just make our own TimeoutError?

GumpacG: Good suggestion. I created a custom error.

kenhuuu:
VOTE +1

kenhuuu:
merging as lazy consensus reached

PR #3509, opened 2026-07-08
Add The Zoo graph to demo Multi-Label

Cole-Greer:
### Summary

This PR adds a new toy graph, "the zoo," and uses it to replace bespoke empty-graph-plus-initializer fixtures in the multi-label test suite and reference documentation with a shared, deterministic dataset.

### Changes

New toy graph (TinkerFactory.createTheZoo() / generateTheZoo())
- 13 vertices (animals, habitats, a person) and 20 edges (livesIn, careFor, friendsWith, eats, avoids), built with LabelCardinality.ZERO_OR_MORE.
- Exercises multi-label vertices alongside a variety of property types (string, int, double, boolean, list-cardinality properties), giving multi-label scenarios a richer, realistic fixture instead of ad hoc addV("a", "b")-style setups.

Note that this PR is meant to stack on top of the #3483 which introduces multilabel as part of https://issues.apache.org/jira/browse/TINKERPOP-3261


docs/src/reference/the-graph.asciidoc
```diff
@@ -206,6 +206,34 @@ link:https://tinkerpop.apache.org/docs/x.y.z/dev/provider/#_label_cardinality[pr
 details and the <<label-step,label()>>, <<labels-step,labels()>>, <<addlabel-step,addLabel()>>, and
 <<droplabel-step,dropLabel()>> steps for traversal access.
 
+[[the-zoo-toy-graph]]
+TIP: A toy graph demonstrating multi-label vertices alongside a variety of property types is available at
+`TinkerFactory.createTheZoo()` and `data/tinkerpop-zoo*`. Unlike "The Crew" above, which requires a graph configured
```
spmallette: doesn't this sentence read strangely? 

> Unlike "The Crew" above, which requires a graph configured for vertex multi-properties and meta-properties, "The Zoo" requires a `LabelCardinality` of `ONE_OR_MORE` or `ZERO_OR_MORE`.

We don't "configure" graphs for multi/meta-properties. A graph "supports" those things. You might say that you "configure" label cardinality for "TinkerGraph" but not every graph will have a "configuration" for that (i.e. some graphs will just support it or not) so you'd have to be specific.

docs/src/reference/the-graph.asciidoc
```diff
@@ -206,6 +206,34 @@ link:https://tinkerpop.apache.org/docs/x.y.z/dev/provider/#_label_cardinality[pr
...
+for vertex multi-properties and meta-properties, "The Zoo" requires a `LabelCardinality` of `ONE_OR_MORE` or
+`ZERO_OR_MORE`.
+
+.The Zoo
+image::the-zoo-graph.png[width=685]
```
spmallette, line 221: Can't help wondering if this graph carries too much to display nicely for folks to follow as an image. I won't let it block the PR but maybe we just need do showcase the multilabels for the graph visually or a core subgraph...

Cole-Greer: I'm considering that image a bit of a placeholder at the moment. I think with a bit of effort to find a more planar embedding and some different colours for nodes, it will be a lot more legible.

Cole-Greer:
VOTE +1

PR #3511, opened 2026-07-11
Add gremlin/io package export exposing GraphBinary serializers

danielbodart:
Adds a `gremlin/io` package export exposing the GraphBinary type serializers.

They are currently only reachable by importing past the `exports` map into `build/`, which couples consumers to internal file layout and breaks under bundlers that honour `exports` (e.g. esbuild/wrangler). This lets server-side v4 protocol implementations reuse them cleanly:

```js
import ioc from 'gremlin/io';
```

Follows the `./language` subpath precedent (`import` + `types`). CHANGELOG updated. Happy to rename the subpath or add a CJS `require` entry if you'd prefer.

Not tied to a JIRA yet (glad to file one). Scope is the export map + changelog only; not run through the full Maven build locally. Drafted with AI assistance (`Assisted-by` trailer on the commit).


Cole-Greer:
Thanks @danielbodart for the contribution. My vote is VOTE +1. Merging as CTR.

PR #3512, opened 2026-07-13
Added coverage for gremlin-console and gremlin-language

GumpacG:
# Test coverage improvements

Two related but independent changes that improve test coverage in TinkerPop, each self-contained:

1. **Grammar parser coverage** - raises `GremlinParser` coverage in `gremlin-language`.
2. **gremlin-console coverage** - adds unit + end-to-end (pexpect) tests for the console.

Both changes also remove dead code found while raising coverage. Removals are listed first, followed by a consolidated coverage summary with the before/after numbers and an explanation of why coverage doesn't go higher.

---

## Dead code removed

**gremlin-language / gremlin-core (grammar)**
- Removed the orphaned `booleanArgument`, `dateArgument`, and
  `nullableGenericLiteralMap` grammar rules (unreachable from `queryList` after
  their references were dropped in earlier changes), plus the coupled
  `gremlin-core` base-visitor stubs, `ArgumentVisitor`/`TranslateVisitor`
  methods, and obsolete `ArgumentVisitorTest` cases. Regenerated the `gremlin-js`
  parser and removed its dead `visitBooleanArgument`.

**gremlin-console**
- `Console.groovy`: unused private `writeTraverserToErrorLines`.
- `PluggedIn.groovy`: never-instantiated `GroovyGremlinShellEnvironment` inner class.
- Deleted the unused `MockGroovyGremlinShellEnvironment` test helper.

---

### gremlin-console
**Caveat:** the standard build runs the pexpect tests in Docker without the JaCoCo agent, so CI's JaCoCo run won't reflect the pexpect-driven portion; the 83% figure was obtained by attaching the agent to the console JVM the pexpect tests spawn.

before (~50%):

coverage-console-before

after (~83%):

coverage-screenshot-2

---

### gremlin-language (grammar / GremlinParser)

- gremlin-language grammar package: 56.98% -> 60.25% (instructions).
- `GremlinParser` main rule class: 76.2% -> 81.2% instructions, 62.8% -> 69.2% branches.
- Aggregate `GremlinParser` (gremlin-language + full gremlin-core suite): 57.5% -> 61.3%.
- ~33 gremlin-core grammar visitor methods moved from 0% to covered.

before (~57%):
coverage-screenshot-7
before (~60%):
coverage-screenshot-5

`GremlinParser` before:
coverage-screenshot-8
`GremlinParser` after:
coverage-screenshot-6

### Why coverage doesn't go higher

**gremlin-console**
- `InstallCommand`'s successful dependency-download branch (the real `Grape.grab` + post-download plugin load) needs Grape + network, so it can't run in the offline test environment.
- `RemoteGremlinPlugin` and remote query submission only do real work against a live Gremlin Server. Covering them from `gremlin-console` would require standing up an embedded server or adding a test dependency on `gremlin-server`/`gremlin-driver` - out of scope for this module, and already covered by those modules' own integration tests - so those paths are intentionally left unexercised here.
- The `StackOverflowError` hint branch in `handleError` stays uncovered: triggering a real `StackOverflowError` on demand is flaky and depends on the JVM `-Xss` stack size, so a test for that single hint message would be brittle. (The other interactive error branches - the "Display stack trace? [yN]" y/n answers, the `Failure` formatter, and the non-interactive `-e` script-error exit - are covered by the pexpect suite.)

**gremlin-language (`GremlinParser`)**

`GremlinParser` is ANTLR-generated, so the remaining uncovered code is either
unreachable by valid input or only coverable by tests that assert nothing
meaningful. These additions already cover essentially all the reachable,
behaviour-bearing productions:

- ANTLR error-recovery paths - every generated rule method has a
  `catch (RecognitionException)` recovery block. The grammar test harness installs
  `BailErrorStrategy`, which aborts on the first error, so recovery is never
  entered; reaching those blocks would mean re-parsing malformed input under a
  non-bail error strategy the harness deliberately doesn't use. This is the single
  largest uncovered bucket.
- Grammar-shadowed alternatives - some labeled alternatives are always
  pre-empted by an earlier alternative that matches the same input (e.g.
  `by(Order)`/`by(T)`, and the nested-traversal `has`/`property`/`hasValue`
  forms, which the `*_Object` alternatives subsume). ANTLR never builds those
  context nodes, so no input can cover them; making them reachable would require
  reordering the grammar (out of scope).
- Unused generated accessors - `*Context` accessor methods such as
  `KeywordContext` and `TraversalGTypeContext` token getters are never called
  (consumers read `ctx.getText()`), and the generated no-op listener is unused.
  The only way to cover them is to invoke the generated getters directly, which
  asserts nothing about behaviour, so such a test would only game the metric.
- Generated infrastructure - `getSerializedATN`, `getATN`, `sempred`, etc. are
  ANTLR boilerplate; invoking them in a test would colour the lines without
  verifying anything, so covering them would only inflate the number.


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/ArgumentVisitor.java
```diff
@@ -112,15 +112,6 @@ public Object parseLong(final GremlinParser.IntegerArgumentContext ctx) {
         }
     }
 
-    @Override
-    public Object visitBooleanArgument(final GremlinParser.BooleanArgumentContext ctx) {
```
kenhuuu, line 116: I'd probably lean toward keeping these as they are real things in the grammar. Candidate for removal if we remove it from the `Gremlin.g4` file as well

gremlin-language/src/main/antlr4/Gremlin.g4
```diff
@@ -1544,11 +1544,6 @@ ioOptionsStringConstant
     : K_IOU
     ;
 
-booleanArgument
```
kenhuuu, line 1547: I believe the intention is to keep the 4.x and 3.8.x grammar aligned. Given that 3.8 has already been released, we shouldn't do this deletion until 5.x

gremlin-console/src/main/groovy/org/apache/tinkerpop/gremlin/console/PluggedIn.groovy
```diff
@@ -77,35 +77,4 @@ class PluggedIn {
     void deactivate() {
         this.activated = false
     }
-
-    public class GroovyGremlinShellEnvironment implements GremlinShellEnvironment {
```
Cole-Greer, line 81: This is a part of the codebase I'm not too familiar with, but as I understand it this code only existed to support the old remote-console infrastructure. If so, I agree it should be removed.

gremlin-console/src/test/groovy/org/apache/tinkerpop/gremlin/console/PreferencesTest.groovy
```diff
@@ -0,0 +1,198 @@
...
+        Preferences.colors = true
+
+        // Installs the change listeners AND loads default values.
+        Preferences.expandoMagic()
+    }
+
+    @AfterClass
+    static void tearDownClass() {
+        // These tests write to the real per-user java.util.prefs store. Clear it so developer prefs aren't polluted.
+        ShellPreferences.clear()
```
xiazcy: Just confirming that these won't clear existing user preferences?

xiazcy:
VOTE +1 pending comments resolution

GumpacG:
This should be merged after beta-3 as to not add risk and other coverage tests in other areas will be after beta-3 as well.

gremlin-console/src/test/groovy/org/apache/tinkerpop/gremlin/console/PreferencesTest.groovy
GumpacG: I included an in memory preference for testing so that it doesn't conflict with user preferences.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/language/grammar/ArgumentVisitor.java
GumpacG: Reverted, thanks!

gremlin-language/src/main/antlr4/Gremlin.g4
GumpacG: Reverted, thanks!

PR #3513, opened 2026-07-15
Fix doFinalReduction to remove keys when post-barrier steps produce nothing

theneelshah:


Cole-Greer:
Following [my comment in the original PR](https://github.com/apache/tinkerpop/pull/3501#issuecomment-4870160248), I think it's nonsensical that a non-productive by() would result in different behaviours (filter vs empty list) depending on if a barrier happens to be present. I agree this is a bug, as the semantics of group().by() should not have special cases for barriers, it would be inconsistent and unpredictable. For this reason, I think it's suitable for inclusion in 3.8-dev.

VOTE +1

xiazcy:
VOTE +1

kenhuuu:
VOTE +1

PR #3516, opened 2026-07-16
Stop sending traversal source as GremlinLang parameter

kenhuuu:
Remote traversals should select the traversal source with the top-level HTTP request `g` field. The `parameters` field is only for query values referenced by the Gremlin string, so adding `g` there was unused by Gremlin Server.

VOTE +1


Cole-Greer:
VOTE +1

xiazcy:
VOTE +1

PR #3517, opened 2026-07-16
Reduce async overhead in .NET GraphBinary reads

kirill-stepanishin:
### Background
The GraphBinary response reader in Gremlin.Net used a `BufferedStream` so it could refill from the network in 8 KiB chunks. But even when the next few bytes were already sitting in that buffer, every primitive read still went through an async state machine and allocated a small scratch array. On large responses, that per-read overhead adds up to a large share of deserialization time.

### The change
`BufferedStream` is replaced with a small custom `GraphBinaryReadBuffer`. Refill behavior is the same, but when the bytes are already buffered the typed readers complete synchronously with no allocation. When they are not, a slow path handles the refill and keeps the previous EOF and cancellation behavior.

Variable-length payloads (Binary, String, Char) also switch to `ReadExactlyAsync`, since `Stream.ReadAsync` may return fewer bytes than requested and the old code assumed it wouldn't.

Unit tests cover the common case where bytes are already buffered, cases where a value straddles a refill boundary, end of stream, cancellation, and reading a full response one small chunk at a time through `ResponseSerializer`.

### Performance

Benchmarked with the client and server on separate cross-region EC2 instances (server in US-EAST-2, client in US-WEST-2) so results reflect realistic network latency.

| Workload | Before | After | Change |
|---|---|---|---|
| `g.V().repeat(both()).times(2)` (~200k results, latency) | 2.229 s | 1.612 s | **28% faster** |
| `g.V()` (6 results, latency) | 0.0578 s | 0.0580 s | no change |
| `g.V()` (throughput @ concurrency 64) | 1,145 req/s | 1,151 req/s | no change |
| `g.V()` (throughput @ concurrency 256) | 4,597 req/s | 4,595 req/s | no change |
| `g.V()` (throughput @ concurrency 1000) | 17,402 req/s | 17,423 req/s | no change |

Large responses benefit the most, since that is where async overhead accounts for most of the client-side time. Throughput is unchanged, and the small query confirms no regression.

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphBinary4/GraphBinaryReadBuffer.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Structure/IO/GraphBinary4/GraphBinaryReadBuffer.cs
@@ -1,0 +1,305 @@
+#region License
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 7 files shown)


xiazcy:
VOTE +1

Cole-Greer:
VOTE +1

kenhuuu:
VOTE +1

PR #3518, opened 2026-07-16
Honor per-request options embedded in a script via with options

kenhuuu:
The server now applies per-request options set inside a submitted Gremlin string's with(): evaluationTimeout/timeoutMillis, materializeProperties, batchSize, bulkResults, and language, with a single precedence: script embedded with() > request field > (header, bulkResults only) > default. This fixes a regression where a script embedded timeoutMillis was silently ignored on the raw string path.

Options are resolved once in Context; batchSize is parsed and validated there, returning 400 Bad Request for a non positive or out of range value rather than stalling result iteration or surfacing an uncaught 500. Validation runs before any transaction begin side effects. g, parameters, and transactionId remain request field/header only, and the obsolete requestId scrape is removed from GremlinScriptChecker.

VOTE +1

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
@@ -36,14 +36,16 @@ public class GremlinScriptChecker {
     /**
      * An empty result whose properties return as empty.
      */
-    public static final Result EMPTY_RESULT = new Result(null, null, null);
+    public static final Result EMPTY_RESULT = new Result(null, null, null, null, null);
 
     /**
      * At least one of these tokens should be present somewhere in the Gremlin string for {@link #parse(String)} to
      * take any action at all.
      */
     private static final Set<String> tokens = new HashSet<>(Arrays.asList("timeoutMillis", "TIMEOUT_MILLIS",
-            "requestId", "REQUEST_ID", "materializeProperties", "ARGS_MATERIALIZE_PROPERTIES"));
+            "materializeProperties", "ARGS_MATERIALIZE_PROPERTIES",
+            "language", "ARGS_LANGUAGE", "batchSize", "ARGS_BATCH_SIZE",
+            "bulkResults", "BULK_RESULTS"));
 
     /**
      * Matches single line comments, multi-line comments and space characters.
@@ -85,16 +87,6 @@ public class GremlinScriptChecker {
     private static final String timeoutTokens = "[\"']timeoutMillis[\"']|(?:Tokens\\.)?TIMEOUT_MILLIS";
 
     /**
-     * Regex fragment for the timeout tokens to look for. There are basically four:
-     * <ul>
-     *     <li>{@code requestId} which is a string value and thus single or double quoted</li>
-     *     <li>{@code REQUEST_ID} which is a enum type of value which can be referenced with or without a {@code Tokens} qualifier</li>
-     * </ul>
-     * See {@link #patternWithOptions} for a full explain as this regex is embedded in there.
-     */
-    private static final String requestIdTokens = "[\"']requestId[\"']|(?:Tokens\\.)?REQUEST_ID";
-
-    /**
      * Regex fragment for the materializeProperties to look for. There are basically four:
      * <ul>
      *     <li>{@code materializeProperties} which is a string value and thus single or double quoted</li>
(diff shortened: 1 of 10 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
```diff
@@ -110,7 +129,7 @@ public class GremlinScriptChecker {
      * From https://regex101.com/
      *
      * \.with\((((?:["']materializeProperties["']|["']scriptEvaluationTimeout["']|(?:Tokens\.)?ARGS_EVAL_TIMEOUT|(?:Tokens\.)?ARGS_SCRIPT_EVAL_TIMEOUT),(?<to>\d*)(:?L|l)?)|((?:["']materializeProperties["']|(?:Tokens\.)?ARGS_MATERIALIZE_PROPERTIES),["'](?<mp>.*?)["']?)|((?:["']requestId["']|(?:Tokens\.)?REQUEST_ID),["'](?<rid>.*?)["']))\)
```
xiazcy: Nit: update or perhaps even remove this outdated javadoc

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
```diff
@@ -178,7 +197,9 @@ public class GremlinScriptChecker {
             Pattern.compile("\\.with\\((((?:"
                     + timeoutTokens + "),(?<to>\\d*)(:?L|l)?)|((?:"
                     + materializePropertiesTokens + "),[\"'](?<mp>.*?)[\"']?)|((?:"
-                    + requestIdTokens + "),[\"'](?<rid>.*?)[\"']))\\)");
+                    + languageTokens + "),[\"'](?<lang>.*?)[\"'])|((?:"
+                    + batchSizeTokens + "),(?<bs>\\d+))|((?:"
```
xiazcy: Confirming that we want the regex to ignore negative values here? I believe this just silently drops the option when user puts a negative values like `g.with('batchSize', -5)` (though not sure if anyone would) and falls back to default. Just wondering if we want to explicitly capture negative values and give error.

xiazcy:

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
kenhuuu: Yea, I noticed this too, but I figured giving the default was better than returning a 400 for this particular case.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
kenhuuu: Yea, I wasn't sure what to do about this so I kept it the same.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
```diff
@@ -178,7 +197,9 @@ public class GremlinScriptChecker {
...
+                    + bulkResultsTokens + "),(?<br>true|false)))\\)");
```
Cole-Greer: Nit: It might be worth allowing this to be `.*` instead of `true|false` and relying on the downstream `Boolean.parseBoolean()` to reject invalid inputs. `parseBoolean()` is case-insensitive which I think gives nice flexibility.

The same argument could potentially be applied to the `d+` for batchSize, although it's weaker in that case. Expanding that one to `.*` would presumably allow for things like `+` prefixes and `L` suffixes.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
xiazcy: I think falling to default makes sense for negative values, but perhaps we should note it in the reference doc that if a user puts a negative value it'll get ignored and falls to default?

Cole-Greer:

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
kenhuuu: I'm going to slightly loosen bulkResult only because something like `g.with('bulkResults', 'someinvalidvalue')` gets interpreted as false because of parseBooleans behavior. So its not quite right to completely loosen.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
kenhuuu: Made the script recognize negatives.

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/jsr223/GremlinScriptChecker.java
kenhuuu: shortened it.

PR #3520, opened 2026-07-17
Fix gremlin-python GremlinLang equality and submitAsync bugs

GumpacG:
# gremlin-python bug fixes

Two bugs introduced in `4.0.0-beta.1` during the Bytecode-to-GremlinLang
migration.

## 1. `GremlinLang.__eq__` ignores the query

`gremlin_python/process/traversal.py`

`__eq__` compared `self.gremlin` to itself instead of to `other.gremlin`, so
equality depended only on the parameters and ignored the actual steps.

### Problem use case

```python
a = g.V().count().gremlin_lang
b = g.V().drop().gremlin_lang
a == b
```

### Behaviour

| | Result |
|---|---|
| Before | `True` - two completely different queries (with equal/empty parameters) compare equal |
| After | `False` - equality now considers the query steps as well as the parameters |

## 2. `DriverRemoteConnection.submitAsync` is broken (removed)

`gremlin_python/driver/driver_remote_connection.py`

The deprecated `submitAsync` alias forwarded three positional arguments to
`submit_async`, which had been narrowed to a single `gremlin_lang` argument in
4.x, and it never returned the future - so it could not work at all. Rather than
repair a deprecated method, it has been removed (per PR review), with an upgrade
note directing users to `submit_async()`.

### Problem use case

```python
conn = DriverRemoteConnection('http://localhost:8182/gremlin', 'g')
future = conn.submitAsync(gremlin_lang)
result = future.result()
```

### Behaviour

| | Result |
|---|---|
| Before | Raises `TypeError: submit_async() takes 2 positional arguments but 4 were given`; even setting that aside it returned `None` instead of a future |
| After | `submitAsync` is removed; use `submit_async(gremlin_lang)`, which returns the `Future` (resolving to a `RemoteTraversal`) |

## Tests

- `tests/unit/process/test_gremlin_lang.py`: equality of traversals with
  different steps / equal params, equal objects, differing params, and
  non-GremlinLang comparison.
- `tests/unit/driver/test_remote_connection.py`: `submitAsync` emits the
  deprecation warning, forwards a single argument, returns a `Future` resolving
  to a `RemoteTraversal`, and no longer raises `TypeError`.


xiazcy:
VOTE +1

gremlin-python/src/main/python/gremlin_python/driver/driver_remote_connection.py
```diff
@@ -75,12 +75,12 @@ def submit(self, gremlin_lang):
                                          request_options=self.extract_request_options(gremlin_lang))
         return RemoteTraversal(result_set)
 
-    def submitAsync(self, message, parameters=None, request_options=None):
+    def submitAsync(self, gremlin_lang):
```
Cole-Greer: We should arguably just take TinkerPop 4 as the opportunity to delete this deprecated overload instead of continuing to adjust it. It's been deprecated for 5 years at this point. We can kick that into a new JIRA to followup with later if you think it needs more discussion, or to take a more holistic view of the python deprecation situation when we aren't in a release cycle.

Cole-Greer:
Building on my comment above, I'd like to either remove `submitAsync()` now, or open a JIRA to track it. I'm happy with either of those directions for this PR.

VOTE +1

gremlin-python/src/main/python/gremlin_python/driver/driver_remote_connection.py
GumpacG: Removed, thanks!

kenhuuu:
VOTE +1

PR #3522, opened 2026-07-17
Fix gremlin-go/dotnet P.outside() and gremlin-go negative BigInteger

GumpacG:
# gremlin-go bug fixes

Two gremlin-go bugs introduced in 4.0.0-beta.2

## 1. Negative BigInteger / BigDecimal decoded incorrectly

Problem: `readBigInt` reconstructed negative two's-complement values with a bit
width one byte too wide (`(len(b)+1)*8` instead of `len(b)*8`), so it subtracted a
power of two that was 256x too large. Positive values were unaffected. Negative
`BigDecimal` values were affected too, because their unscaled value decodes through
the same path. Go was the only GLV with this bug because it hand-rolls the
conversion (`math/big` has no signed-bytes constructor), while the other GLVs
delegate to a correct standard-library primitive.

Use case: a ledger service stores account balances as arbitrary-precision integers
so they never overflow. Account `acct-1` is overdrawn by 500, stored as the
`BigInteger` `-500`.

Before:

```go
r, _ := g.V("acct-1").Values("balance").Next()
balance := r.GetInterface() // *big.Int
// want:  -500
// got:   -16712180   (silently wrong, no error returned)
```

The application reads a balance of -16,712,180 for an account that is only 500
overdrawn. The error is silent, so it can flow into downstream calculations,
alerts, or stored data before anyone notices.

After:

```go
r, _ := g.V("acct-1").Values("balance").Next()
balance := r.GetInterface() // *big.Int
// -500, matching the server and every other GLV
```

## 2. P.outside() generated invalid gremlin-lang

Problem: `translatePValue` list-wrapped the arguments of any multi-value predicate
except `between` and `inside`, but `outside` was omitted from that exclusion.
`outside` takes exactly two arguments in the grammar (like `between`/`inside`), so
they must be comma-separated, not wrapped in a list.

Use case: an anomaly-detection query finds sensor readings outside the normal
operating range of 18 to 65 (too cold or too hot).

Before:

```go
readings, err := g.V().Has("temperature", gremlingo.P.Outside(18, 65)).ToList()
// generated gremlin-lang: g.V().has("temperature",outside([18,65]))
// err: server-side parse error - "outside([18,65])" is not valid gremlin-lang
// readings: nil
```

The query never runs. Because the identical query works from the Java, Python,
.NET, and JavaScript drivers, the failure looks like a server or environment
problem rather than a driver bug, and `P.Outside` is effectively unusable from Go.

After:

```go
readings, err := g.V().Has("temperature", gremlingo.P.Outside(18, 65)).ToList()
// generated gremlin-lang: g.V().has("temperature",outside(18,65))
// err: nil
// readings: the out-of-range vertices, as expected
```


xiazcy:
VOTE +1

Cole-Greer:
VOTE +1

kenhuuu:
VOTE +1

PR #3523, opened 2026-07-17
Multilabel Docs Update

xiazcy:
Slightly bigger changes in the docs so opted for a PR. This can also be merged as CTR after quick reviews.

VOTE +1

### DIFF
--- a/docs/src/dev/provider/index.asciidoc
+++ b/docs/src/dev/provider/index.asciidoc
@@ -192,11 +192,51 @@ The following bullets provide some tips to consider when implementing the struct
 `Graph open(Configuration)` method where the `Configuration` is an Apache Commons class of that name. Alternatively, the
 `Graph` implementation can have the `GraphFactoryClass` annotation which specifies a class with that static
 `Graph open(Configuration)` method.
+* `Vertex`
+** By default a vertex has exactly one immutable label. Graphs that want to support multiple, mutable, or zero
+labels should follow the <<multi-label-support,Multi-Label Support>> section below.
 * `VertexProperty`
 ** This interface is both a `Property` and an `Element` as `VertexProperty` is a first-class graph element in that it
 can have its own properties (i.e. meta-properties). Even if the implementation does not intend to support
 meta-properties, the `VertexProperty` needs to be implemented as an `Element`. `VertexProperty` should return empty
 iterable for properties if meta-properties is not supported.
+
+[[multi-label-support]]
+==== Multi-Label Support
+
+By default a vertex has exactly one immutable label, matching the 3.x behavior. Providers that want to support
+multiple, mutable, or zero labels declare their supported label cardinality via
+`Graph.Features.VertexFeatures#getLabelCardinality()`, which defaults to `LabelCardinality.ONE`. To support
+multi-label, return `ONE_OR_MORE` or `ZERO_OR_MORE` from this method. This value may be a static constant for
+graphs which do not offer configurable label cardinality. The label cardinality should reflect the current `Graph`
+instance.
+
+The `LabelCardinality` enum exposes `min()`, `max()`, and `supportsMutation()` for programmatic introspection of
+constraints. Constraint enforcement is handled by the `LabelCardinalityValidator` utility class, which providers may
+use directly or replace with their own validation logic tailored to their storage backend.
+
+Providers implementing multi-label must:
+
+- Store and return a `Set<String>` from `Element.labels()`
+- Implement `Element.addLabel(String, String...)`, `Element.dropLabel(String, String...)`, and `Element.dropLabels()`
+- Ensure `hasLabel("a", "b")` uses OR semantics (matches vertices with label "a" or "b")
+- Serialize/deserialize the label set via the V4 GraphBinary format (which sends labels as a list)
+
+The default `elementMap()` and `valueMap()` implementations determine label output format solely from the
+`with("multilabel")` source option, not from the graph's `LabelCardinality`. This means a multi-label graph still
+returns a single label string from these steps unless `with("multilabel")` is explicitly configured. The intent is
+to eventually deprecate the single-string path entirely and always return labels as a set, aligning `elementMap()`
+and `valueMap()` with `labels()`. Until then, `with("singlelabel")` lets users force single-string output on sources
+that default to multi-label semantics. Because the two options request opposite formats, they cannot both be set on
+the same source — doing so raises a `VerificationException`.
+
+Providers who want `elementMap()`/`valueMap()` to return the full label set by default (without requiring users to
+set `with("multilabel")`) should override `PropertyMapStep` and `ElementMapStep` to tie label output to their
+server-side label cardinality configuration. In that case, `with("singlelabel")` should still be respected as an
+explicit user override back to single-string output.
+
+See the link:https://tinkerpop.apache.org/docs/x.y.z/reference/#vertex-labels[Vertex Labels] reference and the
+<<gremlin-semantics,Gremlin Semantics>> section for the behavioral details.
 
 [[olap-implementations]]
 ==== OLAP Implementations
--- a/docs/src/dev/provider/index.asciidoc
+++ b/docs/src/dev/provider/index.asciidoc
@@ -193,50 +193,16 @@ The following bullets provide some tips to consider when implementing the struct
 `Graph` implementation can have the `GraphFactoryClass` annotation which specifies a class with that static
 `Graph open(Configuration)` method.
 * `Vertex`
-** By default a vertex has exactly one immutable label. Graphs that want to support multiple, mutable, or zero
-labels should follow the <<multi-label-support,Multi-Label Support>> section below.
+** Providers must opt into a specific `VertexLabelCardinality` of `ONE`, `ONE_OR_MORE`, or `ZERO_OR_MORE`. Any graph
+supporting a `VertexLabelCardinality` of `ONE_OR_MORE`, or `ZERO_OR_MORE` must use a `Vertex` implementation which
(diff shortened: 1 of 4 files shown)


docs/src/dev/provider/index.asciidoc
```diff
@@ -192,12 +192,52 @@ The following bullets provide some tips to consider when implementing the struct
...
+
+The `LabelCardinality` enum exposes `min()`, `max()`, and `supportsMutation()` for programmatic introspection of
+constraints. Constraint enforcement is handled by the `LabelCardinalityValidator` utility class, which providers may
+use directly or replace with their own validation logic tailored to their storage backend.
+
+Providers implementing multi-label must:
+
+- Store and return a `Set<String>` from `Element.labels()`
+- Implement `Element.addLabel(String, String...)`, `Element.dropLabel(String, String...)`, and `Element.dropLabels()`
+- Ensure `hasLabel("a", "b")` uses OR semantics (matches vertices with label "a" or "b")
```
Cole-Greer: I don't think this line needs to be included here, it's already covered in the semantics docs and enforced in feature tests.

docs/src/dev/provider/index.asciidoc
```diff
@@ -192,12 +192,52 @@ The following bullets provide some tips to consider when implementing the struct
...
+- Serialize/deserialize the label set via the V4 GraphBinary format (which sends labels as a list)
```
Cole-Greer: I don't think this is relevant either, our serializer implementations are already correctly wired to `Element.labels()`. Anyone implementing their own serializers should follow the io docs, which are already aligned for label sets.

docs/src/dev/provider/index.asciidoc
```diff
@@ -192,12 +192,52 @@ The following bullets provide some tips to consider when implementing the struct
...
+instance.
+
+The `LabelCardinality` enum exposes `min()`, `max()`, and `supportsMutation()` for programmatic introspection of
+constraints. Constraint enforcement is handled by the `LabelCardinalityValidator` utility class, which providers may
+use directly or replace with their own validation logic tailored to their storage backend.
+
+Providers implementing multi-label must:
+
+- Store and return a `Set<String>` from `Element.labels()`
+- Implement `Element.addLabel(String, String...)`, `Element.dropLabel(String, String...)`, and `Element.dropLabels()`
```
Cole-Greer: I'm not sure this is the best presentation for these items. How a provider chooses to store labels is not significant, they may store them as a Set behind the scenes which is fine.

The critical piece here which is unique to multilabelled graphs, is that they MUST override all of `Element.labels()`, `Element.addLabel(String, String...)`, `Element.dropLabel(String, String...)`, and `Element.dropLabels()`. What's unique about this situation compared to before is that none of these methods are abstract in `Element` (they all have default implementations which follow single-label semantics). That is a bit of a trap and really should be the focus here.

docs/src/dev/provider/index.asciidoc
```diff
@@ -192,12 +192,52 @@ The following bullets provide some tips to consider when implementing the struct
...
 * `VertexProperty`
 ** This interface is both a `Property` and an `Element` as `VertexProperty` is a first-class graph element in that it
 can have its own properties (i.e. meta-properties). Even if the implementation does not intend to support
 meta-properties, the `VertexProperty` needs to be implemented as an `Element`. `VertexProperty` should return empty
 iterable for properties if meta-properties is not supported.
 
+[[multi-label-support]]
+==== Multi-Label Support
+
+By default a vertex has exactly one immutable label, matching the 3.x behavior. Providers that want to support
```
Cole-Greer: I think we should be careful with using the word default here, we don't want to imply that that any choice is "more correct" than the other. TinkerGraph defaults to a cardinality of ONE, but TinkerPop itself shouldn't have any bias towards any of the choices. The [reference docs](https://github.com/apache/tinkerpop/blob/19112cc479f332ce352b77df36cbac120fb7a451/docs/src/reference/the-graph.asciidoc?plain=1#L192-L240) were carefully structured to avoid calling it the default.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -97,9 +97,8 @@ See: link:https://tinkerpop.apache.org/docs/4.0.0/reference/#match-step[match()]
...
 release introduces configurable label cardinality, allowing vertices to carry multiple labels that can be added and
 removed over their lifetime. This is an optional feature for graph providers and not all will support it in the same
-way. TinkerGraph offers a configurable approach to enabling the the feature using one of the `LabelCardinality` settings
-as a graph-level setting that defaults to `ONE`, preserving the existing single-label behavior. To enable multi-label
-in TinkerGraph, set the vertex label cardinality in the graph properties:
+way. TinkerGraph enables it through a graph-level `LabelCardinality` setting that defaults to `ONE`, preserving the
+existing single-label behavior:
 
 ```
 gremlin.tinkergraph.vertexLabelCardinality=ZERO_OR_MORE
```
Cole-Greer:
```suggestion
gremlin.tinkergraph.vertexLabelCardinality=ONE
```

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -108,14 +107,13 @@ gremlin.tinkergraph.vertexLabelCardinality=ZERO_OR_MORE
 The three modes are `ONE` (single, immutable, the 3.x default), `ONE_OR_MORE` (mutable, minimum one), and
 `ZERO_OR_MORE` (fully flexible). Edge labels remain fixed at `ONE`.
 
-The following examples demonstrate how Gremlin behaves with multi-label enabled:
+With multi-label enabled, `addLabel()` and `dropLabel()` mutate a vertex's labels, and the new `labels()` step
+reliably returns every label — unlike `label()`, which returns only one and is now deprecated:
```
Cole-Greer:
```suggestion
which produces all labels, while the legacy `label()` step continues to produce a single label only, and is now deprecated:
```

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -108,14 +107,13 @@ gremlin.tinkergraph.vertexLabelCardinality=ZERO_OR_MORE
...
 
 ```text
```
Cole-Greer: This should be using Java or groovy syntax highlighting.

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -124,43 +122,17 @@ gremlin> g.V().has('name','marko').dropLabel('employee').labels()
...
-==>graphtraversalsource[tinkergraph[vertices:1 edges:0], standard]
-gremlin> gml.V().has('name','marko').elementMap()
-==>{id=0, label=[manager, person, employee], name=marko}
-gremlin> gml.V().has('name','marko').valueMap(true)
-==>{id=0, label=[manager, person, employee], name=[marko]}
-```
+`mergeV()` accepts a list for `T.label` to match or create multi-label vertices, with append-only `onMatch`
+semantics. By default `elementMap()` and `valueMap()` still return a single label string; use `with("multilabel")`,
+per-traversal or as a persistent source, to receive the full set. The reference documentation covers each of these
+behaviors in detail.
```
Cole-Greer:
```suggestion
`mergeV()` now optionally accepts a list for `T.label` to match or create multi-label vertices. By default
`elementMap()` and `valueMap()` still return a single label string and a new traversal-source level
`with("multilabel")` config alters the results to wrap labels in lists. The reference documentation covers each of these
behaviors in detail.
```

docs/src/upgrade/release-4.x.x.asciidoc
```diff
@@ -1094,33 +1066,16 @@ expected for the 4.0.0 major release.
...
-returns a single label string from these steps unless `with("multilabel")` is explicitly configured. The intent is
-to eventually deprecate the single-string path entirely and always return labels as a set, aligning `elementMap()`
-and `valueMap()` with `labels()`. Until then, `with("singlelabel")` allows users to override a source-level
-`with("multilabel")` back to single-string output.
-
-Providers who want `elementMap()`/`valueMap()` to return the full label set by default (without requiring users to
-set `with("multilabel")`) should override `PropertyMapStep` and `ElementMapStep` to tie label output to their
-server-side label cardinality configuration. In that case, `with("singlelabel")` should still be respected as an
-explicit user override back to single-string output.
+Providers declare their supported label cardinality via `Graph.Features.VertexFeatures#getLabelCardinality()`, which
```
Cole-Greer: Could use more sensitivity around the concept of "default".

docs/src/dev/provider/index.asciidoc
```diff
@@ -192,12 +192,52 @@ The following bullets provide some tips to consider when implementing the struct
...
+returns a single label string from these steps unless `with("multilabel")` is explicitly configured. The intent is
+to eventually deprecate the single-string path entirely and always return labels as a set, aligning `elementMap()`
+and `valueMap()` with `labels()`. Until then, `with("singlelabel")` lets users force single-string output on sources
+that default to multi-label semantics. Because the two options request opposite formats, they cannot both be set on
+the same source — doing so raises a `VerificationException`.
+
+Providers who want `elementMap()`/`valueMap()` to return the full label set by default (without requiring users to
+set `with("multilabel")`) should override `PropertyMapStep` and `ElementMapStep` to tie label output to their
+server-side label cardinality configuration. In that case, `with("singlelabel")` should still be respected as an
+explicit user override back to single-string output.
```
Cole-Greer: It might be nice for this section to reference the `@SingleLabelDefault` and `@MultiLabelDefault` gherkin tags in relation to the providers choices here.

Cole-Greer:
VOTE +1

PR #3525, opened 2026-07-17
[gremlin-dotnet] Handle non-bulked results in remote traversals

xiazcy:
The `gremlin-dotnet` remote traversal path (`DriverRemoteConnection` / `TransactionRemoteConnection`) assumed every server result was a `Traverser`. It submitted as `SubmitAsync<Traverser>`, so `ResultSet<Traverser>` performed an unconditional `(Traverser)item` cast during iteration.

When a compatible HTTP endpoint returns non-bulked results, e.g. a raw `Int64` for `g.V().count()`, or any traversal run with `bulkResults=false`, the values arrive as raw scalars rather than `Traverser` instances, and the cast throws `InvalidCastException`. In effect, `bulkResults=false` was unusable on the remote traversal path.

### Fix

Submit as `SubmitAsync<object>` and adapt the stream at the iteration boundary: pass `Traverser` instances through unchanged, and wrap raw values in `new Traverser(item, 1)`. This mirrors the other GLVs. The generic `ResultSet<T>` and the plain `client.SubmitAsync<T>` path are unchanged.

### Testing

- Unit tests for the adaptation: raw scalar, bulked `Traverser` expansion, and a mixed stream (`DriverRemoteConnectionTests`).
- Remote-traversal integration tests with `bulkResults=false` for a single scalar (`count`) and multi-result (`values`) case (`GraphTraversalTests`).
- Client-path integration tests confirming `bulkResults` true/false behavior (`GremlinClientTests`), matching the Go driver's coverage.

---
VOTE +1

### DIFF
--- a/gremlin-dotnet/src/Gremlin.Net/Driver/Remote/DriverRemoteConnection.cs
+++ b/gremlin-dotnet/src/Gremlin.Net/Driver/Remote/DriverRemoteConnection.cs
@@ -142,7 +142,7 @@ namespace Gremlin.Net.Driver.Remote
                 requestMsg.AddField(Tokens.ArgsBulkResults, true);
             }
 
-            var resultSet = await _client.SubmitAsync<Traverser>(requestMsg.Create(), cancellationToken)
+            var resultSet = await _client.SubmitAsync<object>(requestMsg.Create(), cancellationToken)
                 .ConfigureAwait(false);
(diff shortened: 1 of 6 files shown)


Cole-Greer:
VOTE +1

kenhuuu:
VOTE +1

PR #3526, opened 2026-07-18
Update GraphBinary/GraphSON model test coverage

kenhuuu:
These changes update the model tests by removing stale tests, adding more tests for existing types and new tests for newly supported types.

For review, the tests themselves can probably largely be ignored. I was going to CTR this but I did change how CompositePDT works (uses typed Map now rather than plain JSON object) so I'm going to keep this open for a couple days in case anyone wants to look at that part in particular. If not, I'll likely just merge this in as a CTR.

VOTE +1


Cole-Greer:
VOTE +1

xiazcy:
VOTE +1

PR #3531, opened 2026-07-21
Fix gremlin-go PartitionStrategy panic and gremlin-python ProductiveByStrategy

GumpacG:
# Fix gremlin-go PartitionStrategy panic and gremlin-python ProductiveByStrategy productiveKeys

## 1. gremlin-go: `PartitionStrategy` panics when `ReadPartitions` is unset

`gremlin-go/driver/strategies.go`

`PartitionStrategy` called `config.ReadPartitions.ToSlice()` with no nil check. `ReadPartitions` is a `Set` interface, so leaving it unset (a valid, documented configuration - "zeroed values are ignored") caused a nil-interface method call and crashed the caller.

### Use case

Configure a write-only partition (write to partition `a`, read from everywhere), leaving `ReadPartitions` unset:

```go
g = g.WithStrategies(gremlingo.PartitionStrategy(gremlingo.PartitionStrategyConfig{
    PartitionKey:   "_partition",
    WritePartition: "a",
    // ReadPartitions intentionally not set
}))
```

### Behaviour

| | Result |
|---|---|
| Before | Panics at strategy construction: `runtime error: invalid memory address or nil pointer dereference` |
| After | Constructs normally; `readPartitions` is simply omitted from the configuration when unset |

## 2. gremlin-python: `ProductiveByStrategy` drops `productiveKeys`

`gremlin-python/src/main/python/gremlin_python/process/strategies.py`

`ProductiveByStrategy.__init__` accepted a `productiveKeys` argument but never stored it in `self.configuration`, so it was silently discarded before the request was built. The strategy was therefore always sent with no keys (the empty singleton), and the caller's `productiveKeys` had no effect. Every other GLV (Java, Go, JavaScript, .NET) carries `productiveKeys` correctly, so gremlin-python was the lone outlier.

### Use case

Scope the strategy to specific keys:

```python
g = traversal().withStrategies(ProductiveByStrategy(productiveKeys=["name", "age"]))
```

### Behaviour

| | Result |
|---|---|
| Before | `productiveKeys` is ignored; the strategy serializes as the bare `ProductiveByStrategy` singleton (empty configuration), so `by()`-modulation productivity does not match what was configured and differs from every other driver |
| After | `productiveKeys` is carried in the configuration (`{"productiveKeys": ["name", "age"]}`) and serialized to the server, matching the other GLVs |

## Tests
- `gremlin-go/driver/strategies_test.go`: `PartitionStrategy` with `ReadPartitions` unset no longer panics and produces the expected configuration.
- `gremlin-python/.../tests/unit/process/test_strategies.py` (`test_configurable`): `ProductiveByStrategy(productiveKeys=[...])` stores the keys in its configuration, and `ProductiveByStrategy()` yields an empty configuration.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -35,6 +35,8 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Expanded `gremlin-python` CI matrix to test against Python 3.9, 3.10, 3.11, 3.12, and 3.13.
 * Add Node 26 support for `gremlin-javascript` and `gremlint`.
 * Corrected numerous inaccuracies in the reference documentation, including wrong default values (connection pool sizes, buffer sizes, ports, timeouts), stale serializer class names, removed options documented as available, and broken code examples across the JVM, Python, `.NET`, Go, and JavaScript drivers.
+* Fixed a panic in `gremlin-go` `PartitionStrategy` when `ReadPartitions` was left unset.
+* Fixed `gremlin-python` `ProductiveByStrategy` to pass through the `productiveKeys` argument, which was previously accepted but never serialized to the server.
 
 [[release-3-7-6]]
 === TinkerPop 3.7.6 (Release Date: April 1, 2026)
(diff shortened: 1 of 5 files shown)


Cole-Greer:
VOTE +1

PR #3539, opened 2026-07-22
TINKERPOP-3275 Fix gremlin-python GraphBinary BigInteger/BigDecimal serialization

GumpacG:
## Summary

`gremlin-python`'s GraphBinary `BigIntIO` writer computed the byte length from the
value's magnitude bit-length but encoded the bytes as signed two's-complement. Negative
boundary values (e.g. `-129`, `-255`) raised `OverflowError`, and positive/zero values
produced non-canonical bytes that diverge from the Java reference (`BigInteger.toByteArray()`).

The fix computes the minimal signed two's-complement length, unifying the positive, negative,
and zero paths. `BigDecimal` is fixed transitively (its unscaled value is serialized as a
`BigInteger`).

## Before vs After

| Value | Before | After |
|---|---|---|
| `BigInteger(-129)` | `OverflowError` | `00 00 00 02 ff 7f` |
| `BigDecimal(-2.55)` (unscaled `-255`) | `OverflowError` | `00 00 00 02 ff 01` |
| `BigInteger(127)` | `00 00 00 02 00 7f` (non-minimal) | `00 00 00 01 7f` |
| `BigInteger(0)` | `00 00 00 00` (Java rejects) | `00 00 00 01 00` |

## Changes

- `gremlin-python/.../structure/io/graphbinaryV1.py` - rewrote `BigIntIO.write_bigint`.
- `gremlin-python/.../tests/unit/io/test_graphbinaryV1.py` - added boundary round-trip tests
  and exact wire-byte assertions.


xiazcy:
VOTE +1

Cole-Greer:
VOTE +1

Note to whoever ends up merging this, to re-evaluate the impact on GraphBinary 4 when merging into master. The GB4 write path in python is borderline dead code at the moment so perhaps we're safe to just keep the new feature tests and drop the code change, otherwise, this fix should be adapted into GB4 upon merging.

PR #3540, opened 2026-07-22
TINKERPOP-3277 Fix gremlin-go GraphBinary serialization of zero BigInteger/BigDecimal

GumpacG:
## Summary

`gremlin-go` serialized a zero `BigInteger` as a GraphBinary length of `0` with no value
bytes. Java's `new BigInteger(new byte[0])` throws `NumberFormatException`, so a Java server
could not read zero values sent by Go. The fix encodes zero as the canonical single `0x00`
byte (length 1), matching Java's `BigInteger.toByteArray()`. `BigDecimal` with a zero unscaled
value is fixed transitively (it serializes its unscaled value as a `BigInteger`).

## Before vs After

| Value | Before | After |
|---|---|---|
| `BigInteger(0)` | `00 00 00 00` (Java rejects) | `00 00 00 01 00` |
| `BigDecimal` with unscaled `0` | `...00 00 00 00` (Java rejects) | `...00 00 00 01 00` |

### Example

```go
// Sending a zero-valued BigInteger property to a Java Gremlin Server
g.AddV("event").Property("count", big.NewInt(0)).Iterate()
```

Before: the Java server fails to deserialize the request (`NumberFormatException: Zero length BigInteger`).
After: zero serializes to canonical bytes and is read correctly.

## Changes

- `gremlin-go/driver/graphBinary.go` - `getSignedBytesFromBigInt` returns `[]byte{0}` for zero.
- `gremlin-go/driver/graphBinary_test.go` - added zero exact-bytes, zero round-trip, and
  zero-unscaled `BigDecimal` tests.

## Testing

Added `TestGraphBinaryV1` subtests for the zero cases.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -39,6 +39,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 * Fixed `gremlin-python` `ProductiveByStrategy` to pass through the `productiveKeys` argument, which was previously accepted but never serialized to the server.
 * Deprecated `ProductiveByStrategy` which was introduced as a temporary way to mimic pre-3.5.0 null processing behavior.
 * Fixed `gremlin-python` GraphBinary serialization of `BigInteger`/`BigDecimal` negative boundary values (e.g. `-129`) that raised `OverflowError`.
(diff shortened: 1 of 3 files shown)


xiazcy:
VOTE +1

Cole-Greer:
VOTE +1

PR #3544, opened 2026-07-22
TINKERPOP-3042 Deprecate ProductiveByStrategy

GumpacG:
`ProductiveByStrategy` was introduced in 3.5.0 as a temporary way to mimic the `null` processing
behavior of versions prior to 3.5.0. That compatibility need has passed, so the strategy is
deprecated as of 3.7.7 (not replaced) and will be removed in the next major version.

This is deprecation only: no behavior change, nothing removed, and the strategy is still applied
where it was before.

### DIFF
--- a/docs/src/upgrade/release-3.7.x.asciidoc
+++ b/docs/src/upgrade/release-3.7.x.asciidoc
@@ -110,6 +110,15 @@ example `g.V().hasId(1, 2, 4)`).
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-3273[TINKERPOP-3273]
 
+==== Deprecated ProductiveByStrategy
+
+`ProductiveByStrategy` is deprecated as of 3.7.7 and is not replaced. The strategy was introduced in 3.5.0 as a
+temporary way to mimic the `null` processing behavior of versions prior to 3.5.0, where a `by()` modulator would
+always produce a result. That compatibility need has passed, so the strategy is no longer recommended and will be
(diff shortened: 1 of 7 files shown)


docs/src/upgrade/release-3.7.x.asciidoc
```diff
@@ -76,6 +76,15 @@ Note that this change is for Java only and designed to better align embedded and
 
 See: link:https://issues.apache.org/jira/browse/TINKERPOP-3238[TINKERPOP-3238]
 
+==== Deprecated ProductiveByStrategy
+
+`ProductiveByStrategy` is deprecated as of 3.7.7 and is not replaced. The strategy was introduced in 3.5.0 as a
+temporary way to mimic the `null` processing behavior of versions prior to 3.5.0, where a `by()` modulator would
+always produce a result. That compatibility need has passed, so the strategy is no longer recommended and will be
+removed in the next major version. This deprecation applies across all Gremlin Language Variants.
```
xiazcy: nit: next major from 3.7.x's perspective is 3.8.x, which I don't think we are removing yet until 4.0.0

xiazcy:
VOTE +1

Cole-Greer:
VOTE +1

Cole-Greer:
Merging as CTR

docs/src/upgrade/release-3.7.x.asciidoc
GumpacG: Updated to say it will be removed for 4.0.0

PR #3545, opened 2026-07-22
TINKERPOP-3185: Fix PeerPressure.property_name token in gremlin-python

kirill-stepanishin:
https://issues.apache.org/jira/browse/TINKERPOP-3185

In the Python driver, `PeerPressure.property_name` pointed at the wrong value, a `pageRank` token instead of a `peerPressure`.

Because of this, `peerPressure().with_(PeerPressure.property_name, ...)` silently ignored the setting. This fix corrects the value to match Java and the other GLVs.

I also checked the other `with()` constants across all GLVs, and this was the only mismatch.


Cole-Greer:
VOTE +1, merging as CTR.

PR #3546, opened 2026-07-22
[TINKERPOP-3273] Align hasId() collection unrolling with V()/E()

xiazcy:
### Summary
`hasId()` and `g.V()`/`g.E()` disagreed on how collection arguments are unrolled. `hasId()` unrolled a collection in any argument position, while `V()`/`E()` only unroll a single collection argument. This made `g.V().hasId(1, [2,4])` return `v[1], v[2], v[4]` whereas `g.V(1, [2,4])` rejected the list. Reported as an incomplete fix from TINKERPOP-2863.

### Change
`hasId(Object, Object...)` now unrolls a collection/array only when it is the single argument. With multiple arguments, each argument is a literal id (no unrolling), matching `V()`/`E()`. A single collection argument is still unrolled, preserving TINKERPOP-2863 behavior.

Behavior after this change:
| Call                  | Predicate         |
|-----------------------|-------------------|
| `hasId([2,4])`        | `within(2,4)`     |
| `hasId(1, 2, 4)`      | `within(1,2,4)`   |
| `hasId(1, [2,4])`     | `within(1,[2,4])` |

Because a start-step `hasId()` folds into the graph step, `g.V().hasId(1, [2,4])` now takes the same lookup path as `g.V(1, [2,4])` (e.g. an id-conversion error on TinkerGraph), giving the consistency the ticket asked for.

### Tests
- `HasStepTest`: removed the now-invalid assertion that `hasId([1], [2])` equals `hasId(1, 2)`.
- `TinkerGraphTest`: added a mid-traversal test (a collection among multiple args is not unrolled) and a start-step test asserting `g.V().hasId(1, [2,4])` is consistent with `g.V(1, [2,4])`.
- `HasId.feature`: added a portable mid-traversal scenario. The start-step error path is provider/engine-specific (it does not occur under OLAP or on shuffle-order graphs), so it is covered by the unit test rather than the cross-provider feature suite.

---

VOTE +1


spmallette:
VOTE +1

Cole-Greer:
VOTE +1

PR #3547, opened 2026-07-22
Fix automatic GLV version propagation

Cole-Greer:
Previously, updating the project version required manual edits to propagate into some language variants: Python's `__version__` was fully manual, and JavaScript/Gremlint never updated their `package-lock.json`. This extends the Maven build (`generate-sources`) to automatically rewrite these files, translating the version suffix per language convention (Python `.dev1`, JS/npm `-alpha1`, .NET/Go raw `-SNAPSHOT`).

With this change, after `mvn versions:set` + `mvn clean install -DskipTests`, all GLV files pick up the new version with no manual steps. 

VOTE +1

WARNING: When merging this PR, Javascript sections will need to be reworked to reflect the gremlin-js restructuring. If complications arise, a `-s ours` merge followed by a new followup PR may be warranted.

### DIFF
--- a/gremlin-python/pom.xml
+++ b/gremlin-python/pom.xml
@@ -82,6 +82,42 @@ limitations under the License.
                     </replacements>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.codehaus.gmavenplus</groupId>
+                <artifactId>gmavenplus-plugin</artifactId>
+                <dependencies>
+                    <dependency>
+                        <groupId>org.apache.groovy</groupId>
+                        <artifactId>groovy-all</artifactId>
+                        <version>${groovy.version}</version>
+                        <type>pom</type>
+                        <scope>runtime</scope>
+                    </dependency>
+                </dependencies>
+                <executions>
+                    <execution>
+                        <id>update-version</id>
+                        <phase>generate-sources</phase>
+                        <goals>
+                            <goal>execute</goal>
+                        </goals>
+                        <configuration>
+                            <scripts>
+                                <script>
+                                    def mavenVersion = "${project.version}"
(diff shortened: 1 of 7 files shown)


gremlin-python/pom.xml
```diff
@@ -82,6 +82,42 @@ limitations under the License.
...
+                            <scripts>
+                                <script>
+                                    def mavenVersion = "${project.version}"
+                                    // Translate Maven version to PEP 440:
+                                    //   "x.y.z-SNAPSHOT" -> "x.y.z.dev1"
+                                    //   "x.y.z"          -> "x.y.z"
+                                    def pythonVersion = mavenVersion.replace("-SNAPSHOT", ".dev1")
+                                    def platformAgnosticBaseDirPath = project.basedir.getAbsolutePath().replace("\\", "/")
+                                    def file = new File(platformAgnosticBaseDirPath + "/src/main/python/gremlin_python/__init__.py")
+                                    file.write(file.getText("UTF-8").replaceFirst(/__version__ = '.*'/, "__version__ = '" + pythonVersion + "'"))
```
xiazcy: Nit: this regex may fail silently if the `__init__.py`'s format gets changed (spacing, quotes, etc.), though I don't see a reason it would, so not a big concern, just a note here.

xiazcy:
VOTE +1

gremlin-python/pom.xml
Cole-Greer: I added a comment to `__init__.py` indicating this is managed by maven and shouldn't be touched. Additionally I added a check to the replacement script such that it will crash the build if the regex match ever fails.

Cole-Greer:
Merging as CTR

PR #3550, opened 2026-07-22
TINKERPOP-3187 Add propertyMap() helper to Element in GLVs

GumpacG:
## Summary

Adds a `propertyMap()` convenience helper that views an element's properties as a
map keyed by property key (`key -> [property objects]`), complementing the flat
`properties` list. The helper is provided on all element types (`Vertex`, `Edge`,
`VertexProperty`) in the four GLVs and on the `Element` structure API in
`gremlin-core`. Also fixes a latent `gremlin-go` GraphBinary deserialization bug
where `VertexProperty.Key` was never populated.

Related: [TINKERPOP-3187](https://issues.apache.org/jira/browse/TINKERPOP-3187)

### gremlin-core (Java)
- `structure/Element.java`: new `default <V> Map<String, List<Property<V>>> propertyMap(String... propertyKeys)`.
  Because `Vertex`, `Edge`, and `VertexProperty` all extend `Element`, they inherit
  it automatically with no provider code changes. Follows the varargs key-filter
  convention of the existing `properties(...)` / `values(...)` methods, preserves
  encounter order via `LinkedHashMap`, and returns an empty (non-null) map when
  there are no matching properties. This is a public API addition to the structure
  interface, implemented as a `default` method so existing implementations are
  unaffected.

### GLVs
- `gremlin-javascript`: `propertyMap()` defined on the `Element` base class
  (inherited by Vertex, Edge, VertexProperty). Returns a plain object.
- `gremlin-python`: `property_map()` on the `Element` base class. Returns a `dict`.
- `gremlin-dotnet`: typed `PropertyMap()` per subclass, mirroring the existing
  `Property(key)` accessors: `Vertex` returns `Dictionary<string, List<VertexProperty>>`;
  `Edge` and `VertexProperty` return `Dictionary<string, List<Property>>`.
- `gremlin-go`: typed `PropertyMap()` per subclass: `Vertex` returns
  `map[string][]*VertexProperty`; `Edge` and `VertexProperty` return
  `map[string][]*Property`.

The typed per-subclass form in .NET/Go and the single base-class method in
JS/Python each follow the language's established convention.

### Bug fix (gremlin-go)
- `driver/graphBinary.go`: `vertexPropertyReader` now sets `vp.Key` from the
  deserialized label (it previously set only `vp.Label`). Without this, a
  server-deserialized `VertexProperty` had an empty `Key`, which caused
  `Vertex.PropertyMap()` to group everything under the empty key. This is a
  behavioral bug fix, not an API change (the `Key` field already existed and is
  public).

## Testing

- `gremlin-javascript`: mocha unit tests for Vertex/Edge/VertexProperty (all passing).
- `gremlin-python`: unit tests for Vertex/Edge/VertexProperty grouping.
- `gremlin-dotnet`: xUnit tests added to `VertexTests`, `EdgeTests`, and
  `VertexPropertyTests` (full unit suite passing).
- `gremlin-go`: driver unit tests for each element type plus
  `TestVertexPropertyReaderSetsKey` exercising the deserialization path that the
  `Key` fix corrects.
- `gremlin-core` / `tinkergraph-gremlin`: five tests in `TinkerGraphTest` covering
  multi-property grouping, single-valued keys, empty element, edge properties, and
  explicit key filtering (all passing against the reference TinkerGraph).

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -33,6 +33,7 @@ This release also includes changes from prior 3.7.x releases.
 * Add missing `Configuring` interface to `GraphStepPlaceholder` and `VertexStepPlaceholder`
 * Fixed bug in `group()` value traversal where keys were retained with stale barrier state instead of being filtered when steps following a `Barrier` in the second `by()` produced no output (e.g. `by(values("age").fold().unfold())` or `by(__.out().fold().count(local).is(P.gt(0)))` for vertices with no out-edges).
 * Corrected numerous inaccuracies in the reference documentation, including wrong default values (connection pool sizes, buffer sizes, ports, timeouts), stale serializer class names, removed options documented as available, and broken code examples across the JVM, Python, `.NET`, Go, and JavaScript drivers.
+* Added `propertyMap()` helper to `Vertex` in `gremlin-javascript`, `gremlin-python`, and `gremlin-dotnet` to view a vertex's properties as a map keyed by property key.
 
 [[release-3-8-1]]
 === TinkerPop 3.8.1 (Release Date: April 1, 2026)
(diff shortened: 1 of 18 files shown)


Cole-Greer:
VOTE +1, merging as CTR

PR #3553, opened 2026-07-23
Add server-start prerequisite to Connecting via Console section

spmallette:
The `Connecting via Console` section of the reference docs shows Gremlin traversals against the modern toy graph (`marko`, `vadas`, `lop`, ...), but it never tells the reader that a compatible Gremlin Server has to be running first, nor which configuration to use. A newcomer who follows the section as written—or who starts the server with the default configuration—gets an empty or incompatible graph, so every traversal shown returns nothing with no explanation.

This adds a short prerequisite before the console instructions:

- Start Gremlin Server with `conf/gremlin-server-modern.yaml`, with a cross-link to the existing *Starting Gremlin Server* section.
- A NOTE clarifying that a different (or the default) configuration serves a different or empty graph, so the examples would not return the results shown.

The existing properties-file, `DriverRemoteConnection`, and client examples are unchanged; only the prerequisite was added. The docs were rendered locally (full live build) and the new content appears correctly in the Connecting via Console section.

### DIFF
--- a/docs/src/reference/gremlin-applications.asciidoc
+++ b/docs/src/reference/gremlin-applications.asciidoc
@@ -732,8 +732,19 @@ and details in the <<gremlin-drivers-variants,Gremlin Drivers and Variants>> Sec
 [[connecting-via-console]]
 === Connecting via Console
 
-Gremlin Console can be used to send remote traversals to a running Gremlin Server. Start Gremlin Console as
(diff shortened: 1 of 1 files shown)


spmallette:
merged on 4b29b5f361bf04cde9a4533db3cf78052c4f31ac

PR #3556, opened 2026-07-23
TINKERPOP-3271 `subgraph()` throws a descriptive error for non-Edge input

GumpacG:
## Summary

`subgraph()` produces an edge-induced subgraph and must be called at an edge step.
When the traversal instead feeds it a non-Edge value, `SubgraphStep` passed that
value straight into an `Edge`-typed method, producing a raw `ClassCastException`
with no indication of what went wrong.

This change validates the traverser value at runtime and throws a descriptive
`IllegalStateException` naming the required `Edge` input, consistent with the
existing pattern in `AddPropertyStep` and `AddEdgeStep`. Valid queries are
unaffected; only the failure message for already-invalid queries changes.

This is a backward-compatible diagnostics fix (runtime validation only, no API or
generics change).

### Non-edge (primitive) input
```groovy
g.inject(1).subgraph('sg').cap('sg').iterate()
```
- Before: `ClassCastException: class java.lang.Integer cannot be cast to class ...Edge`
- After: `IllegalStateException: subgraph() requires Edge input but encountered Integer; use an edge step such as outE(), inE(), or bothE()`

### Vertex input
```groovy
g.addV().subgraph('sg').cap('sg').iterate()
```
- Before: `ClassCastException: class ...TinkerVertex cannot be cast to class ...Edge`
- After: `IllegalStateException: subgraph() requires Edge input but encountered TinkerVertex; use an edge step such as outE(), inE(), or bothE()`

### Correct usage (unchanged)
```groovy
g.V().has('v',1).outE('X').subgraph('sg').cap('sg')  // edges in -> builds subgraph
```
- Before and after: works identically

## Changes
- `gremlin-core`: runtime `Edge` validation in `SubgraphStep.sideEffect(...)`
- `gremlin-test`: two negative tests added to the `SubgraphTest` suite
- `CHANGELOG.asciidoc`: entry under 3.7.7

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,6 +25,7 @@ image::https://raw.githubusercontent.com/apache/tinkerpop/master/docs/static/ima
 [[release-3-7-7]]
 === TinkerPop 3.7.7 (Release Date: NOT OFFICIALLY RELEASED YET)
 
+* Fixed `subgraph()` to throw a descriptive error identifying the required `Edge` input instead of an internal `ClassCastException` when the traversal produces a non-edge value.
 * Fixed `PeerPressure.property_name` in `gremlin-python` incorrectly mapping to the `pageRank` property name token.
 * Added `NextN(n)` to `Traversal` in `gremlin-go` for batched result iteration, providing API parity with `next(n)` in the Java, Python, and .NET GLVs.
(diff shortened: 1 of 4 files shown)


xiazcy:
VOTE +1

Cole-Greer:
VOTE +1, merging as CTR.

PR #3558, opened 2026-07-23
[TINKERPOP-3151] Allow floats with no leading digits

kirill-stepanishin:
https://issues.apache.org/jira/browse/TINKERPOP-3151

Makes the leading digit optional in float literals, so forms like `.5`, `.5f`, and `.5m` parse, aligning the Gremlin grammar with Groovy.

**Changes**:

- Grammar: added a `DecimalFloatingPointLiteral` alternative for the leading-dot form. Added a `RANGE` : `'..'` token so `1..5` still tokenizes correctly (otherwise `..` gets swallowed into a `.5` float), and pointed `genericLiteralRange` at it, with a matching index fix in `GenericLiteralVisitor`.
- Tests: added `ValidFloatLiteralTest` cases and Gherkin scenarios `g_injectXpoint5X/fX/dX`, plus `g_injectX1to5X` to cover the new `RANGE` token. Regenerated the Python/JS/Go/.NET GLV feature files.
- .NET: ignored `g_injectXpoint5fX` because the .NET translator emits a bare `0.5`, dropping the float type on round-trip (a pre-existing limitation only .NET's strict typing surfaces, not fixing it here).


gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/sideEffect/Inject.feature
```diff
@@ -235,3 +235,51 @@ Feature: Step - inject()
...
+    Then the result should be unordered
+      | result |
+      | d[0.5].d |
+
+  @GraphComputerVerificationInjectionNotSupported
+  Scenario: g_injectX1to5X
+    Given the modern graph
+    And the traversal of
+      """
+      g.inject(1..5)
```
xiazcy, line 280: I wonder if it's worth to also do a string range test like `g.inject("a".."c")` here

xiazcy:

gremlin-test/src/main/resources/org/apache/tinkerpop/gremlin/test/features/sideEffect/Inject.feature
kirill-stepanishin: Added, thanks

Cole-Greer:
VOTE +1, merging as CTR

PR #3559, opened 2026-07-23
TINKERPOP-3270 `where(P)` throws a descriptive error for non-String scope keys

GumpacG:
## Summary

`where(P)` expects a `String` scope key (e.g. `where(eq('a'))`). Given a non-String
predicate value like `gt(5)`, `WherePredicateStep` cast it to `String` unconditionally,
throwing a raw `ClassCastException` at step construction. This change validates the
value first and throws a descriptive `IllegalArgumentException` pointing to `is(P)` for
value comparisons. Backward-compatible diagnostics fix; valid queries are unaffected.

### Non-String (numeric) predicate value

```
__.values('v').where(gt(5))
```
* Before: `ClassCastException: class java.lang.Integer cannot be cast to class java.lang.String`
* After: `IllegalArgumentException: where(P) requires a String scope key but encountered Integer; use is(P) to compare values`

### Nested inside a parent `where(Traversal)`

```
g.V().hasLabel('X').where(__.values('v').where(gt(5)))
```
* Before: same `ClassCastException`, buried in a nested traversal
* After: same descriptive `IllegalArgumentException` pointing to `is(P)`

### Correct usage (unchanged)

```
g.V().as('a').out().where(eq('a'))   // String scope key comparison
__.values('v').is(gt(5))             // value comparison via is(P)
```
* Before and after: works identically

## Changes
* `gremlin-core`: runtime `String` validation in `WherePredicateStep.getSelectKey(...)`
* `gremlin-test`: negative tests added to the `WhereTest` suite (the exception is thrown at step construction, so they pass across all engines)
* `CHANGELOG.asciidoc`: entry under 3.7.7

### DIFF
--- a/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/WhereTest.java
+++ b/gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/WhereTest.java
@@ -24,6 +24,7 @@ import org.apache.tinkerpop.gremlin.process.GremlinProcessRunner;
 import org.apache.tinkerpop.gremlin.process.traversal.P;
 import org.apache.tinkerpop.gremlin.process.traversal.Path;
 import org.apache.tinkerpop.gremlin.process.traversal.Traversal;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.__;
 import org.apache.tinkerpop.gremlin.structure.T;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/WhereTest.java
```diff
@@ -552,5 +583,19 @@ public Traversal<Vertex, String> get_g_VX1X_asXaX_out_hasXageX_whereXgtXaXX_byXa
         public Traversal<Vertex, String> get_g_VX3X_asXaX_in_out_asXbX_whereXa_eqXbXX_byXageX_name(final Object v3Id) {
             return g.V(v3Id).as("a").in().out().as("b").where("a", eq("b")).by("age").values("name");
         }
+
+        @Override
+        public Traversal<Vertex, Integer> get_g_V_valuesXageX_whereXgtX5XX() {
+            // gt(5) yields a P whose value is an Integer rather than a String scope key; the cast mimics the
+            // erased P<String> that reaches where(P) via raw types, bytecode, or deserialization
+            final P<String> nonStringKeyPredicate = (P<String>) (P) gt(5);
```
xiazcy: Curious, why the double cast for (P) here? Doesn't `(P) gt(5)` gets you the integer form that you wanted? Or am I misreading the test?

xiazcy:

gremlin-test/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/filter/WhereTest.java
GumpacG: Addressed. It was redundant and removed.

Cole-Greer:
VOTE +1, merging as CTR

PR #3567, opened 2026-07-24
Preserve float type in DotNetTranslator

kirill-stepanishin:
`DotNetTranslator` emitted a Java Float as a bare 0.5, which C# reads as a double. Now it emits 0.5f so the type round-trips.

This re-enables the `g_injectXpoint5fX` .NET Gherkin scenario, previously ignored (`FloatLiteralTypeNotPreserved`).


Cole-Greer:
Thanks @kirill-stepanishin.
VOTE +1, merging as CTR

PR #3570, opened 2026-07-25
Add expected output and explanatory callouts to values()-step docs

spmallette:
The reference documentation for the values()-step listed three examples against the TinkerPop Crew graph but rendered no output and offered only a single sentence of prose, so the behavior of the step was difficult to predict from the docs alone.

This adds:

- An introductory paragraph explaining that values() returns the values of the requested properties, or of every property when called with no arguments, and noting that a property may be multi-valued or carry meta-properties.
- Numbered callouts on the three examples clarifying that a no-argument values() returns every property value across all keys, that the sample vertex has a multi-valued location property, and that properties('location').values() extracts meta-property values (the startTime and endTime years) rather than the city-name values returned by values('location').
- A note that the order in which property values are returned is not guaranteed.

The examples remain executable, so their output is rendered at build time rather than hand-written. No behavior changes; this is a documentation-only improvement.

### DIFF
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -5768,13 +5768,29 @@ link:++https://tinkerpop.apache.org/javadocs/x.y.z/core/org/apache/tinkerpop/gre
 === Values Step
 
 The `values()`-step (*map*) extracts the values of properties from an `Element` in the traversal stream.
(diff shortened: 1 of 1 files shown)


spmallette:
merged on 45bd9c67d74dad137ac8ae18e073747f89edc7c2

PR #3573, opened 2026-07-27
Add runnable count(local) examples to the Count Step docs

spmallette:
The `count(local)` behavior in the Count Step section of the reference documentation was described only in an `IMPORTANT` prose callout, with no executable examples. The callout also omitted the `Path` case: it stated that a count of `1` is returned "for any other object," but `CountLocalStep` returns the path length for a `Path`.

This change:

- Removes the prose-only callout and repurposes its content as explanatory prose that correctly covers all object kinds the step handles: `Collection` (number of elements), `Map` (number of entries), `Path` (number of objects in the path), and any other object (`1`).
- Adds a runnable example block against the modern graph so the per-object-kind counts render as live output at build time and are self-verifying:
  - `g.V().fold().count(local)` → `6` (Collection)
  - `g.V().group().by(label).count(local)` → `2` (Map)
  - `g.V(1).out().path().count(local)` → `2, 2, 2` (Path)
  - `g.V(1).count(local)` → `1` (any other object)
- Cross-links the examples to `fold()`, `group()`, `path()`, and the path data structure so a reader can follow the behavior in context.

The reference book builds and the count-step section reads coherently with the rendered live output matching the documented counts.

### DIFF
--- a/docs/src/reference/the-traversal.asciidoc
+++ b/docs/src/reference/the-traversal.asciidoc
@@ -1311,8 +1311,23 @@ g.V().hasLabel('person').outE('created').count().map {it.get() * 10}.path() <2>
 <1> `count()`-step is a <<a-note-on-barrier-steps,reducing barrier step>> meaning that all of the previous traversers are folded into a new traverser.
 <2> The path of the traverser emanating from `count()` starts at `count()`.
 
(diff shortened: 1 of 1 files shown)


spmallette:
Merged with 06e12a84587b28f9044aac60e0235bb928774f7e

PR #3578, opened 2026-07-28
TINKERPOP-3278 Disable Java serialization on Gryo IO reads

GumpacG:
## Summary

Several Gryo type registrations, mostly `TraversalStrategy` implementations and `TraversalExplanation`,
were bound to Kryo's `JavaSerializer`, which deserializes via `java.io.ObjectInputStream.readObject()`.
Reading a Gryo document from an untrusted source could therefore reconstruct and run an arbitrary
`Serializable` object graph during decoding, before the graph layer accepts or rejects anything.

This change stops the graph-document IO paths from reaching that sink, while leaving full fidelity
available for trusted, in-process use.

## What changed

**New API**

- `GryoMapper.Builder.javaSerializationAllowed(boolean)`, default `true`. When `false`, every registration
  whose serializer resolves to a `JavaSerializer` is dropped, including registrations contributed through an
  `IoRegistry` or `addCustom(...)`, whether the serializer is supplied directly, as a `Function`, or via a
  class default serializer.

**Paths hardened (build their mappers with `javaSerializationAllowed(false)`)**

- The `io()` step (read and write branches)
- `GryoReader` and `GryoWriter` default mappers
- `GryoIo` (covers `graph.io(IoCore.gryo())` and `gryo` graph persistence)
- The Hadoop Gryo input/output formats (`GryoRecordReader`, `GryoRecordWriter`)

Reads and writes are hardened together so these paths cannot produce a document they will not read back.

**Unchanged (full fidelity retained for trusted bytes)**

- A directly built `GryoMapper.build().create()`
- `GryoPool`, and the `spark-gremlin` and Hadoop object pools (which additionally run unlocked by design)
- GraphSON and GraphBinary (never used native Java serialization)

## Behavior change

A stream that presents one of the affected type ids now fails with an unregistered class id instead of
being deserialized:

```
org.apache.tinkerpop.shaded.kryo.KryoException: Encountered unregistered class ID: 187
```

Affected types: `PartitionStrategy`, `SubgraphStrategy`, `SeedStrategy`, `VertexProgramStrategy`,
`ProductiveByStrategy`, `OptionsStrategy` and `TraversalExplanation` in both Gryo 1.0 and 3.0, plus
`GroupStep.GroupBiOperator` and `OrderGlobalStep.OrderBiOperator` in Gryo 1.0.

## Breaking changes and capability impact

**Breaking:** reading a Gryo document that embeds one of the affected types now fails on the default IO paths
(`io()`, `GryoReader`, `GryoWriter`, `GryoIo`, and the Hadoop formats) instead of deserializing it. Restore the
old behavior where the bytes are trusted with `javaSerializationAllowed(true)` (see below). Additive API only,
no signatures changed; `GryoMapper.build().create()` and graph-structure documents are unaffected.

A graph document produced by `writeGraph`/`writeVertices` contains only vertices, edges, properties and
their values, never strategies, so in practice nothing is lost on the graph-document paths. Nothing in the
repository round-trips these types through Gryo: OLAP ships strategies to workers via JDK serialization in
the job configuration, not through the Gryo registrations.

The capability is removed from the default, not from the product. A directly built mapper keeps full
fidelity, and `javaSerializationAllowed(true)` restores it on any builder. `GryoIo` exposes it through the
`onMapper` consumer where the bytes are trusted:

```java
GryoIo.build().graph(graph).onMapper(m -> ((GryoMapper.Builder) m).javaSerializationAllowed(true)).create();
```

Migration: a caller holding an older `.kryo` document that embeds one of the affected types can read it by
supplying a mapper with `javaSerializationAllowed(true)` to `GryoReader`, or via the `onMapper` hook above.

## Testing

- `GryoMapperTest` (parameterized over Gryo 1.0 and 3.0): a canary proving `readObject()` does not run on a
  hardened read, a positive control proving the crafted bytes do reach the sink on a full-fidelity mapper
  (so the negative assertions are meaningful), graph-structure round-trips, and registration checks against
  the constructed `Kryo` for default, direct-custom, `Function`-custom and class-default-serializer forms.
- `GryoRecordReaderWriterTest`: asserts the Hadoop reader and writer build hardened mappers.

## Docs

CHANGELOG entry, an upgrade note in `docs/src/upgrade/release-3.7.x.asciidoc`, and scoped updates to
`THREAT_MODEL.md` (sections 8 and 9) describing which mappers are hardened.

### DIFF
--- a/THREAT_MODEL.md
+++ b/THREAT_MODEL.md
@@ -337,14 +337,17 @@ Per-surface trust table:
   Groovy string by concatenation, which is the calling application's concern (§9). *Violation symptom:*
   grammar breakout / step injection from a value that should stay a literal. *Severity:* critical.
  
-- **Deserializer integrity.** The wire deserializers (GraphSON, GraphBinary) and **Gryo in its locked default
-  (`registrationRequired=true`)** reading attacker bytes do not lead to arbitrary object instantiation / code
-  execution beyond the registered type set. Because `inject()` and value arguments let a request carry any
+- **Deserializer integrity.** The wire deserializers (GraphSON, GraphBinary) and **the hardened Gryo mappers the
+  IO paths build** (`registrationRequired=true` plus `javaSerializationAllowed=false`, i.e. `io()`, `GryoReader`,
+  `GryoWriter`, `GryoIo`) reading attacker bytes do not reach native Java deserialization
+  (`ObjectInputStream.readObject()`). Because `inject()` and value arguments let a request carry any
   supported type, a bug in a **registered** type's (de)serializer that crashes/OOMs the reader is also
   in-model, on **both** the server (request) and the GLV (response) side. The GraphML reader disables
   external entities and DTDs by default (XXE-safe). *Violation symptom:* deserialization gadget / RCE / XXE,
   or a registered-type serializer crashing/OOMing either end. *Severity:* critical. Gryo is not on the wire,
-  and unlocked Gryo or a caller-supplied unhardened XML factory is out-of-model (user responsibility, §9).
+  and unlocked Gryo (as run by `spark-gremlin` and the Hadoop object pools), a directly built `GryoMapper` or
+  `GryoPool` that keeps Java serialization, or a caller-supplied unhardened XML factory is out-of-model
+  (user responsibility, §9).
  
 - **Resource bounds — split.** Malformed/pre-auth input that crashes/OOMs/hangs the server is **in-model**
   (above). Ordinary expensive traversals / large results are **operator capacity**, NOT in-model, unless a
@@ -394,7 +397,13 @@ Per-surface trust table:
   allow-list (`registrationRequired=true`), so it is not an arbitrary-instantiation sink, and a break within
   that locked config is a `VALID` bug like any deserializer. **Running Gryo unlocked
   (`registrationRequired=false`) is not a safe boundary against untrusted bytes and is the user's
-  responsibility.** (A few registered types use Java native serialization, a gadget caveat even when locked.)
+  responsibility.** A few registered types are also serialized with Kryo's `JavaSerializer`, which reads by way of
+  `ObjectInputStream.readObject()` and is a gadget caveat even when locked. As of 3.7.7 the mappers the IO paths
+  build (`io()`, `GryoReader`, `GryoWriter`, `GryoIo`) drop those registrations
+  (`GryoMapper.Builder.javaSerializationAllowed(boolean)` selects the behavior), so a break there is `VALID`. A
+  directly built `GryoMapper` and `GryoPool` keep them, as do the `spark-gremlin` and Hadoop object pools that
+  additionally run unlocked; those remain the user's responsibility. The Hadoop Gryo input/output formats still
+  keep them while locked, which is a known gap rather than a disclaimed one.
   Gryo is not on the wire, so this is an IO/file-surface concern (`io()` step, persistence, OLAP).
  
 - **A `TraversalStrategy` is not an access-control boundary on its own.** A remote request can remove or
--- a/THREAT_MODEL.md
+++ b/THREAT_MODEL.md
@@ -247,7 +247,7 @@ Per-surface trust table:
 | Gremlin Server — bytecode/traversal request | serialized traversal bytecode | **yes**, within privileges | auth; traversal-step allow-list; resource limits |
 | Gremlin Server — session id (`SessionOpProcessor`) | client-supplied session string | **yes** | keyed by the string with no owning-user check, so any client presenting the id shares the session (see §11b) |
 | Request deserialization (GraphSON / GraphBinary) | serialized bytes | **yes** (pre-auth) | robustness of the wire serializers |
-| Graph IO — Gryo/GraphSON/GraphML files (`io()` step, persistence, OLAP) | on-disk / cluster bytes | only if the caller loads untrusted files | GraphSON, locked Gryo (`registrationRequired=true`), and GraphML with the default XML factory owe deserializer integrity. Unlocked Gryo and a caller-supplied unhardened `XMLInputFactory` (XXE) are the caller's responsibility |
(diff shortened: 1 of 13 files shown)


gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoMapperTest.java
```diff
@@ -456,4 +785,27 @@ public Duration read(final Kryo kryo, final Input input, final Class<Duration> d
         }
     }
 
+    /**
```
spmallette: orphaned javadoc comment

THREAT_MODEL.md
```diff
@@ -394,7 +398,12 @@ Per-surface trust table:
   allow-list (`registrationRequired=true`), so it is not an arbitrary-instantiation sink, and a break within
   that locked config is a `VALID` bug like any deserializer. **Running Gryo unlocked
   (`registrationRequired=false`) is not a safe boundary against untrusted bytes and is the user's
-  responsibility.** (A few registered types use Java native serialization, a gadget caveat even when locked.)
+  responsibility.** A few registered types are also serialized with Kryo's `JavaSerializer`, which reads by way of
+  `ObjectInputStream.readObject()` and is a gadget caveat even when locked. As of 3.7.7 the mappers the IO paths
```
Cole-Greer: > As of 3.7.7

I don't think we should encode the development history in the threat model. I think these edits should simply ensure that the current state of the model will accurately reflect 3.7-dev:head once the PR has merged.

hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/structure/io/gryo/GryoRecordReader.java
```diff
@@ -73,7 +73,7 @@ public void initialize(final InputSplit genericSplit, final TaskAttemptContext c
         if (configuration.get(Constants.GREMLIN_HADOOP_GRAPH_FILTER, null) != null)
             this.graphFilter = VertexProgramHelper.deserialize(ConfUtil.makeApacheConfiguration(configuration), Constants.GREMLIN_HADOOP_GRAPH_FILTER);
         this.gryoReader = GryoReader.build().mapper(
-                GryoMapper.build().addRegistries(IoRegistryHelper.createRegistries(ConfUtil.makeApacheConfiguration(configuration))).create()).create();
+                GryoMapper.build().javaSerializationAllowed(false).addRegistries(IoRegistryHelper.createRegistries(ConfUtil.makeApacheConfiguration(configuration))).create()).create();
```
spmallette, line 78: for our documentation sake, could you please add a comment as to why we disable java serialization here - something that say why it isn't necessary basically. my take is that as a record reader (and you need a similar comment in `GryoRecordWriter`) it is meant to serialize a graph `Element` which would not serialize things like `OptionsStrategy` and classes that use `JavaSerializer`. 

this is in contrast to the `MapReduce` shuffle path (which we didn't change): `VertexWritable.write()`/`readFields()` go through `KryoShimServiceLoader`, which uses the `GryoPool` from `HadoopPools`. That pool is built with `initializeMapper(m -> m.registrationRequired(false))` — i.e. deliberately unlocked, and not hardened. That's consistent with the threat model, which explicitly places "the Hadoop object pools that additionally run unlocked" out-of-model (they only carry intra-job graph elements between mapper and reducer, not attacker-controlled bytes). I think you should probably add a clarifying comment there about why we didn't disable the `JavaSerializer` there.

docs/src/upgrade/release-3.7.x.asciidoc
```diff
@@ -32,6 +32,47 @@ complete list of all the modifications that are part of this release.
...
+
+Callers that supply their own mapper to `GryoReader` or `GryoWriter` are unaffected and should add
+`javaSerializationAllowed(false)` if they read bytes they do not control. `GryoIo` accepts an `onMapper` consumer, so
+full fidelity can be restored where the bytes are trusted:
+
+[source,java]
+----
+GryoIo.build().graph(graph).onMapper(m -> ((GryoMapper.Builder) m).javaSerializationAllowed(true)).create();
+----
+
```
spmallette, line 75: i think we should tie this to a JIRA.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoMapperTest.java
GumpacG: Fixed, thanks!

THREAT_MODEL.md
GumpacG: Fixed, thanks!

hadoop-gremlin/src/main/java/org/apache/tinkerpop/gremlin/hadoop/structure/io/gryo/GryoRecordReader.java
GumpacG: Added comments, thanks!

docs/src/upgrade/release-3.7.x.asciidoc
GumpacG: Added a Jira ticket, thanks! https://issues.apache.org/jira/browse/TINKERPOP-3278

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/gryo/GryoMapper.java
```diff
@@ -78,11 +79,15 @@ public final class GryoMapper implements Mapper<Kryo> {
...
     private final boolean referenceTracking;
+    private final boolean javaSerializationAllowed;
     private final Supplier<ClassResolver> classResolver;
     private final GryoVersion version;
 
     private GryoMapper(final Builder builder) {
-        this.typeRegistrations = builder.typeRegistrations;
+        this.javaSerializationAllowed = builder.javaSerializationAllowed;
+        this.typeRegistrations = builder.typeRegistrations.stream().
+                filter(tr -> javaSerializationAllowed || !(tr.getShadedSerializer() instanceof JavaSerializer)).
```
xiazcy: Is there a reason we are doing the filtering of direct `JavaSerializer` here in the constructor, instead of inside `createMapper()` via the `resolvesToJavaSerializer` helper with the rest?

GumpacG: It was a miss. I've moved it to the helper function. Thanks!

spmallette:
VOTE +1 - recommend squashing commits in this case. not sure there's much value or history to preserve in this little change.

xiazcy:
VOTE +1

Cole-Greer:
VOTE +1

PR #3583, opened 2026-07-29
Bump Netty to 4.1.136.Final and jackson-databind to 2.18.9

Cole-Greer:
Dependency upgrades to address accumulated CVEs.

### Netty 4.1.136.Final
- Bumps `bouncycastle.version` 1.69 → 1.70 to match Netty's own pinned `bcpkix-jdk15on` version.
- Fixes `HttpHandlerUtilTest` (Netty's stricter HTTP request-line validation now rejects the test's placeholder URI containing a space).
- Updates `NOTICE` files in `gremlin-console`, `gremlin-driver`, `gremlin-server`.

### jackson-databind 2.18.9
- Fixes compile breaks from removed/changed Jackson APIs (`DefaultSerializerProvider.withCaches`, `DeserializationContext.mappingException`).
- Updates `GraphSONStreamConstraintsTest` assertions for Jackson's reworded exception messages (behavior unchanged).
- Bumps `maven-shade-plugin` 3.4.1 → 3.6.2 (required for jackson-databind 2.18.9's Java 21 multi-release JAR classes; no relocation/behavior changes for our use case).

**Breaking change:** jackson-databind 2.18.9 hardens `InetAddress` deserialization to reject hostnames, accepting only literal IP addresses (closes a DNS-lookup-on-untrusted-input vector, jackson-databind#6058, related to CVE-2026-54514). No upstream opt-out exists. This restriction is adopted as-is rather than reintroducing hostname resolution via a custom deserializer. Any GraphSON data with `InetAddress` values stored as hostnames (e.g. `"localhost"`) will fail to deserialize after this upgrade and must be migrated to literal IP addresses. Documented in `docs/src/upgrade/release-3.7.x.asciidoc`.

VOTE +1

### DIFF
--- a/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONStreamConstraintsTest.java
+++ b/gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONStreamConstraintsTest.java
@@ -48,8 +48,8 @@ public class GraphSONStreamConstraintsTest extends AbstractGraphSONTest{
             serializeDeserializeAuto(mapper, serializedData);
         });
         assertTrue("Expected StreamConstraintsException for exceeding max number length, found: "+exception.getMessage(),
-                exception.getMessage().contains("org.apache.tinkerpop.shaded.jackson.core.exc.StreamConstraintsException: Number length")
-                && exception.getMessage().contains("exceeds the maximum length (2)"));
+                exception.getMessage().contains("org.apache.tinkerpop.shaded.jackson.core.exc.StreamConstraintsException: Number value length")
+                && exception.getMessage().contains("exceeds the maximum allowed (2, from `StreamReadConstraints.getMaxNumberLength()`)"));
     }
 
     @Test
@@ -66,8 +66,8 @@ public class GraphSONStreamConstraintsTest extends AbstractGraphSONTest{
             serializeDeserializeAuto(mapper, serializedData);
         });
         assertTrue("Expected StreamConstraintsException for exceeding max String length, found: "+exception.getMessage(),
-                exception.getMessage().contains("String length")
-                        && exception.getMessage().contains("exceeds the maximum length (20)"));
+                exception.getMessage().contains("String value length")
+                        && exception.getMessage().contains("exceeds the maximum allowed (20, from `StreamReadConstraints.getMaxStringLength()`)"));
     }
 
     @Test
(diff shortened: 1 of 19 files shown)


gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONStreamConstraintsTest.java
```diff
@@ -48,8 +48,8 @@ public void testMaxNumberLengthConfig() throws Exception {
             serializeDeserializeAuto(mapper, serializedData);
         });
         assertTrue("Expected StreamConstraintsException for exceeding max number length, found: "+exception.getMessage(),
-                exception.getMessage().contains("org.apache.tinkerpop.shaded.jackson.core.exc.StreamConstraintsException: Number length")
-                && exception.getMessage().contains("exceeds the maximum length (2)"));
+                exception.getMessage().contains("org.apache.tinkerpop.shaded.jackson.core.exc.StreamConstraintsException: Number value length")
+                && exception.getMessage().contains("exceeds the maximum allowed (2, from `StreamReadConstraints.getMaxNumberLength()`)"));
```
xiazcy, line 52: It seems like our assertions depends on the Jackson error message wording, is there a way to just assert the exception type + limit value instead?

pom.xml
```diff
@@ -174,12 +174,12 @@ limitations under the License.
         <logback.version>1.2.13</logback.version>
         <metrics.version>3.0.2</metrics.version>
         <mockito.version>3.10.0</mockito.version>
-        <netty.version>4.1.125.Final</netty.version>
+        <netty.version>4.1.136.Final</netty.version>
```
xiazcy, line 177: Nit: there's a stale comment inside `gremlin-server/pom.xml` referring to `https://github.com/netty/netty/blob/netty-4.1.125.Final/pom.xml#L963` for the bouncycastle pin rationale, given the version bumps, comment should be updated.

gremlin-shaded/pom.xml
```diff
@@ -48,7 +48,7 @@ limitations under the License.
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.15.2</version>
+            <version>2.18.9</version>
```
xiazcy, line 51: I see your commit message said about deferring the issue of transient dependencies for hadoop/spark/sparql still pulling in the older Jackson, should that point be mentioned more visibly somewhere in the docs for now? And perhaps opening a Jira for that?

xiazcy (approved):
Generally looks good to me. 

VOTE +1 pending comment resolution.

gremlin-core/src/test/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONStreamConstraintsTest.java
Cole-Greer: I think I'd rather continue asserting the full message even if it leaves us prone to changing these tests again in the future. It's useful to know when jackson is changing the messages on us.

gremlin-shaded/pom.xml
Cole-Greer: That's a case of my agent leaking a bit too much context into the commit message. Transient jackson dependencies in hadoop/spark/sparql have always been managed somewhat independently from the main dependency in gremlin-shaded. Those upgrades have a much larger blast-radius which I don't want to commit to at this time. I don't think there's any need to document that anywhere.

PR #3586, opened 2026-07-31
TINKERPOP-3279 Restrict GraphSON 1.0 embedded-type deserialization

GumpacG:
## Summary

GraphSON 1.0 with embedded types (`TypeInfo.PARTIAL_TYPES`) configured Jackson default typing with `JsonTypeInfo.Id.CLASS` and no `PolymorphicTypeValidator`. Reading a document reconstructed whatever class was named in its `@class` property, so a crafted document could name and construct (or merely load, running its static
initializer) any class on the classpath. It is reachable pre-auth via the typed GraphSON 1.0 wire serializer
`GraphSONMessageSerializerV1` (`application/vnd.gremlin-v1.0+json`), and through `io()` reads and GraphSON 1.0
persistence. GraphSON 2.0/3.0 are unaffected: they resolve types through a fixed registry, not by arbitrary class name.

This change constrains that default typing while leaving legitimate value types working.

## What changed

**New API**
- `GraphSONMapper.Builder.addAllowedTypeIdPrefix(String...)` trusts additional class-name prefixes for GraphSON 1.0 embedded-type deserialization, in addition to the safe defaults. Intended for provider/application types read from trusted input.

**Hardened (GraphSON 1.0 `PARTIAL_TYPES` only)**
- A custom `PolymorphicTypeValidator` decides a simple type id from its **name**, so a disallowed class is refused in `validateSubClassName` **before it is loaded** (its static initializer never runs). Array descriptors are unwrapped and allowed only when the component type is allowed; primitive arrays are allowed.
- The allow-list is **derived from the types GraphSON 2.0/3.0 register** (reduced to package prefixes, since V1 names concrete implementations): `java.lang`, `java.util`, `java.math`, `java.time`, `java.sql`, `org.apache.tinkerpop`, minus the network packages `java.net`/`java.nio`.
- A few safe value types in those excluded packages are allowed **by exact name**: `java.net.InetAddress` and `java.net.URI` (string-backed, no DNS, unlike `java.net.URL`).
- `java.lang.Class` is **exact-denied** even though `java.lang` is allowed, so a `java.lang.Class` *value* cannot name and load an arbitrary class (Jackson would otherwise resolve it with `initialize=true`).
- A `GraphSON1dScreeningIdResolver` refuses a **parameterized type id** (one containing `<`) before Jackson resolves it. This closes a bypass where a disallowed class hidden as a generic type argument was loaded before validation, and where **enum** type arguments were not validated at all (Jackson skips them). GraphSON 1.0 never emits a parameterized `@class`, so nothing legitimate is affected.

## Behavior change

A GraphSON 1.0 typed document whose `@class` names a type outside the allowed set now fails on read:

```
org.apache.tinkerpop.shaded.jackson.databind.exc.InvalidTypeIdException:
Could not resolve type id ...: Configured `PolymorphicTypeValidator` ... denied resolution
```

## Breaking changes and capability impact

**Breaking:** reading a GraphSON 1.0 typed document now fails, instead of deserializing, when its `@class`:
- names a type outside the allow-list (a provider type registered through an `IoRegistry`, an application POJO, most `java.net.*`/`java.nio.*` types),
- is `java.lang.Class` used as a value, or
- is a parameterized/generic id.

A document written by an older version that embeds such a type may no longer be readable.

Restore a specific type for trusted input by trusting its package:

```java
GraphSONMapper.build().version(GraphSONVersion.V1_0).typeInfo(TypeInfo.PARTIAL_TYPES)
        .addAllowedTypeIdPrefix("com.example.").create();
```

or migrate to GraphSON 3.0 / GraphBinary. Standard scalar, collection and array values, `java.math`/`java.time`/
`java.sql` values, `java.net.InetAddress`/`java.net.URI`, and TinkerPop graph types are unaffected. The restriction
gates reads; the write path still emits `@class` for any type, so a writer and reader should be configured consistently. (`java.nio.ByteBuffer` is written by V1 as a concrete subtype Jackson cannot reconstruct, so it was never round-trippable in V1 and remains refused on read.)

## Testing

`GraphSONMapperPartialEmbeddedTypeTest` covers, on the default V1 mapper:
- a disallowed `@class` is refused, and is not class-loaded (a static-init canary never fires);
- arrays round-trip, and an array of a disallowed component type is refused;
- a disallowed class named as a generic **type argument** is refused before it is loaded, and an **enum** type
  argument is refused (regression for the parameterized-id / enum bypass);
- a `java.lang.Class` value is refused and not loaded;
- a value in a denied network package (`java.net.URL`) is refused, while `java.net.InetAddress` and `java.net.URI` round-trip;
- `java.sql`/`java.util` value types round-trip;
- `addAllowedTypeIdPrefix(...)` re-enables an otherwise-denied package.

GraphSON 2.0/3.0 embedded-type and wire-serializer suites continue to pass unchanged.

### DIFF
--- a/THREAT_MODEL.md
+++ b/THREAT_MODEL.md
@@ -232,7 +232,7 @@ Knobs that change which security properties hold (Gremlin Server, `gremlin-serve
   `gremlin-server-secure.yaml`. The docs are explicit that TinkerPop offers no complete out-of-the-box
   protection against nefarious scripts. *(documented — `gremlin-applications.asciidoc` "Protecting Script
   Execution", the two sample configs)*
-- **Enabled serializers** — wire set is GraphSON 3.0 + GraphBinary. Gryo is IO-format-only, not on the wire,
+- **Enabled serializers** — the default wire set is GraphSON 3.0 + GraphBinary. The typed GraphSON 1.0 serializer (`vnd.gremlin-v1.0+json`) is also shippable and constrains its `@class` default typing to an allow-list (§8). Gryo is IO-format-only, not on the wire,
   and defaults to a locked registration allow-list (`registrationRequired=true`). Disabling that lock removes
   the untrusted-input protection (§9). *(documented — sample configs, `gremlin-applications.asciidoc`
   "Serialization")*
@@ -341,7 +341,8 @@ Per-surface trust table:
   IO paths build** (`registrationRequired=true` plus `javaSerializationAllowed=false`, i.e. `io()`, `GryoReader`,
   `GryoWriter`, `GryoIo`, and the Hadoop Gryo input/output formats) reading attacker bytes do not reach native Java
   deserialization
-  (`ObjectInputStream.readObject()`). Because `inject()` and value arguments let a request carry any
+  (`ObjectInputStream.readObject()`), and do not instantiate an arbitrary class named in the document: GraphSON 1.0
+  embedded types constrain Jackson default typing to a validated type allow-list. Because `inject()` and value arguments let a request carry any
   supported type, a bug in a **registered** type's (de)serializer that crashes/OOMs the reader is also
   in-model, on **both** the server (request) and the GLV (response) side. The GraphML reader disables
   external entities and DTDs by default (XXE-safe). *Violation symptom:* deserialization gadget / RCE / XXE,
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONMapper.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONMapper.java
@@ -28,8 +28,12 @@ import org.apache.tinkerpop.shaded.jackson.core.StreamReadConstraints;
 import org.apache.tinkerpop.shaded.jackson.databind.ObjectMapper;
 import org.apache.tinkerpop.shaded.jackson.databind.SerializationFeature;
 import org.apache.tinkerpop.shaded.jackson.databind.JavaType;
+import org.apache.tinkerpop.shaded.jackson.databind.DatabindContext;
+import org.apache.tinkerpop.shaded.jackson.databind.DeserializationContext;
 import org.apache.tinkerpop.shaded.jackson.databind.cfg.MapperConfig;
 import org.apache.tinkerpop.shaded.jackson.databind.jsontype.PolymorphicTypeValidator;
+import org.apache.tinkerpop.shaded.jackson.databind.jsontype.TypeIdResolver;
+import org.apache.tinkerpop.shaded.jackson.databind.jsontype.NamedType;
 import org.apache.tinkerpop.shaded.jackson.databind.jsontype.TypeResolverBuilder;
 import org.apache.tinkerpop.shaded.jackson.databind.jsontype.impl.StdTypeResolverBuilder;
 import org.apache.tinkerpop.shaded.jackson.databind.module.SimpleModule;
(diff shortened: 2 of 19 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONMapper.java
```diff
@@ -288,6 +492,18 @@ public Builder typeInfo(final TypeInfo typeInfo) {
...
         }
 
+        /**
+         * Adds a class-name prefix that GraphSON 1.0 embedded-type deserialization will accept in a {@code @class}
+         * property, in addition to the safe defaults ({@code java.lang.}, {@code java.util.}, {@code java.math.},
+         * {@code java.time.}, {@code org.apache.tinkerpop.} and array types). Use this to re-enable a provider or
+         * application type read from trusted input; a graph document from an untrusted source should not be granted
+         * additional prefixes. Has no effect on GraphSON 2.0 or 3.0, which resolve types through a fixed registry.
+         */
+        public Builder addAllowedTypeIdPrefix(final String... prefixes) {
```
kenhuuu: Should this have a matching denyTypeIdPrefix() in case the current allowed list is too permissive?

gremlin-core/src/test/java/com/example/gadget/GraphSONTestGadgets.java
```diff
@@ -0,0 +1,101 @@
...
+        public static final String FIRED_PROPERTY = "tinkerpop.test.graphson.staticInitCanaryEnum";
+        static {
+            System.setProperty(FIRED_PROPERTY, "fired");
+        }
+    }
+
+    /**
+     * Canary used to verify a disallowed class named as a generic type argument is not loaded when refused.
+     */
+    public static class StaticInitCanaryArg {
```
kenhuuu, line 109: Nit: maybe a short comment about why this `StaticInitCanaryArg` and `StaticInitCanaryValue` both need to exist. From a first glance it looks like the same class, but I'm guessing this has to do with static initialization?

kenhuuu:
Some potential missing test cases:

1. Try cloning a GraphSONMapper to see if the allowedTypeIdPrefixes are copied over.
2. Test some invalid prefixes like null, empty and blank. Also make sure something like `com.example` doesn't allow something like `com.exampleextendedddddddd` through.
3. Add some test for the MessageSerializers or IO serializers and not just the ObjectMapper directly.

THREAT_MODEL.md
```diff
@@ -232,7 +232,7 @@ Knobs that change which security properties hold (Gremlin Server, `gremlin-serve
   `gremlin-server-secure.yaml`. The docs are explicit that TinkerPop offers no complete out-of-the-box
   protection against nefarious scripts. *(documented — `gremlin-applications.asciidoc` "Protecting Script
   Execution", the two sample configs)*
-- **Enabled serializers** — wire set is GraphSON 3.0 + GraphBinary. Gryo is IO-format-only, not on the wire,
+- **Enabled serializers** — the default wire set is GraphSON 3.0 + GraphBinary. The typed GraphSON 1.0 serializer (`vnd.gremlin-v1.0+json`) is also shippable and screens its `@class` default typing against an allow-list of exact class names before loading the class (§8). Gryo is IO-format-only, not on the wire,
```
Cole-Greer: I think we can probably drop any threat model changes from this PR, such updates are probably best handled separately.

Cole-Greer (commented):
VOTE +1

gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONMapper.java
kenhuuu: Will add in a later version if needed. For now all types are known to be good.

kenhuuu:
VOTE +1

kenhuuu:
Merging as its well past lazy consensus

PR #3610, opened 2026-08-04
TINKERPOP-3281 Reject malformed GraphBinary length prefixes

GumpacG:
GraphBinary sized heap allocations from an unvalidated 4-byte length prefix before reading the payload, so a tiny
pre-auth frame (~6-21 bytes) could demand gigabytes and OOM Gremlin Server or a connecting driver.

## What changed

- Validate each length/count against the bytes actually remaining before it sizes an allocation or bounds a loop;
  reject negative or oversized with `IOException`.
- Cap eager pre-sizing of growable collections/maps (grow on demand instead).
- Reject truncated length prefixes and negative `BulkSet` bulk values.
- Covers `String`, `ByteBuffer`, `BigInteger`, `InetAddress`, list, set, map, `Tree`, `BulkSet`, `Bytecode`, `P`
  and graph. Well-formed messages are unaffected.

## Testing

`GraphBinaryLengthPrefixTest` (rejection + large round-trip cases); full GraphBinary compatibility suite green.

 <br>
See: https://issues.apache.org/jira/browse/TINKERPOP-3281

### DIFF
--- a/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/types/SimpleTypeSerializer.java
+++ b/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/types/SimpleTypeSerializer.java
@@ -23,6 +23,7 @@ import org.apache.tinkerpop.gremlin.structure.io.binary.GraphBinaryReader;
 import org.apache.tinkerpop.gremlin.structure.io.binary.GraphBinaryWriter;
 import org.apache.tinkerpop.gremlin.structure.io.binary.TypeSerializer;
 import org.apache.tinkerpop.gremlin.structure.io.Buffer;
+import org.apache.tinkerpop.gremlin.structure.io.SerializationException;
 
 import java.io.IOException;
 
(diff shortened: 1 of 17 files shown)


gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/structure/io/binary/types/SimpleTypeSerializer.java
```diff
@@ -68,6 +68,39 @@ public T readValue(final Buffer buffer, final GraphBinaryReader context, final b
...
 
+    /**
+     * Reads a length or element-count prefix and validates it before it is used to size an allocation. A negative
+     * value, or one larger than the number of bytes actually remaining in the buffer, cannot be legitimate since
+     * every counted element or byte needs at least one byte on the wire, so it is rejected rather than allowed to
+     * drive a large allocation from a small message.
+     */
+    protected static int readSizePrefix(final Buffer buffer) throws IOException {
+        if (buffer.readableBytes() < Integer.BYTES)
+            throw new IOException(String.format(
```
kenhuuu, line 87: We should probably throw SerializationException from serializers as that is already a type of IOException

GumpacG: SerializationException lives in gremlin-util which depends on gremlin-core, not the other way around. To do this, SerializationException would have to be duplicated, which is not recommended. Won't do as SerializationException is not in the right place.

kenhuuu:
VOTE+1

kenhuuu:
Merging as lazy consensus reached.

PR #3630, opened 2026-08-26
Reduce async overhead in JS GraphBinary reads

kirill-stepanishin:
Adds a synchronous fast path to gremlin-javascript StreamReader

Every primitive read in `StreamReader` awaits `#ensure(n)` first. For a buffered `submit()` response a chunk of
the payload is already in memory, so that await often does nothing but still costs a Promise and a
microtask per read, which dominates client CPU on large responses.  Reads now return right away when the bytes are already buffered and only await when the buffer needs refilling.

### Performance

Cross-region EC2 (server US-EAST-2, client US-WEST-2) over HTTP, Modern graph.

| Workload | Before | After | Change |
| --- | --- | --- | --- |
| `g.V().repeat(both()).times(12)` (~200k results) | 2.714 s | 2.085 s | **~23% faster** |
| `g.V()` (6 results) | 0.054 s | 0.056 s | no change |
| `g.V()` throughput @ 64 / 256 / 1000 | 1226 / 2490 / 2396 req/s | 1212 / 2633 / 2438 req/s | no change |

### DIFF
--- a/gremlin-js/gremlin-javascript/lib/structure/io/binary/internals/StreamReader.js
+++ b/gremlin-js/gremlin-javascript/lib/structure/io/binary/internals/StreamReader.js
@@ -23,7 +23,8 @@ import { Buffer } from 'buffer';
  * Async byte reader that provides a uniform interface over both a complete Buffer
  * (for non-streaming submit()) and a ReadableStream (for streaming HTTP responses).
  *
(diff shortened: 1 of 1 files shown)


kenhuuu:
VOTE +1

Cole-Greer:
VOTE +1

PR #3636, opened 2026-09-01
inject() should accept bound variables

danielbodart:
## What

`inject()` is the only variadic-generic step in the Gremlin grammar that rejects a bound variable. This changes both `inject` productions from `genericLiteralVarargs` to `genericArgumentVarargs`, so `inject(<variable>)` parses and resolves like every sibling step.

## Why (consistency)

A `…Argument` rule is exactly `literal | variable` — the `variable` alternative is the bound parameter. Counting who uses which variadic-generic rule in `Gremlin.g4`:

- **`genericArgumentVarargs`** (accepts a bound variable) is used by `V()`, `E()` (spawn and mid-traversal), `hasId()`, `hasValue()`, `property()`, and — the closest structural twin — `within()` / `without()`.
- **`genericLiteralVarargs`** (literal-only) is referenced by **nothing but the two `inject` productions**.

So `inject` is the lone holdout: its nearest sibling `within()`, which does the identical "spray a list of values in" job, already accepts a bound parameter. This aligns `inject` with the rest of the language rather than adding anything new to it.

## Compatibility

Strict superset. `genericArgument` includes `genericLiteral` (maps included), so every existing `inject(...)` call parses and behaves exactly as before; only the previously-rejected `inject(<variable>)` becomes valid. Verified against the existing negative grammar corpus (`incorrect-gremlin-values.txt`) — it contains no bare identifiers, so nothing moves from "correctly rejected" to "now accepted."

## Changes

- **Grammar** (`Gremlin.g4`): both `inject` productions → `genericArgumentVarargs`.
- **Reference visitors** (`TraversalSourceSpawnMethodVisitor`, `TraversalMethodVisitor`): route through `ArgumentVisitor.parseObjectVarargs(ctx.genericArgumentVarargs())`, identical to how `V()`/`E()` already work.
- **Groovy translators** (Java + JavaScript): the `inject(x, null)` Groovy-closure disambiguation walked a `genericLiteralExpr` layer that `genericArgumentVarargs` doesn't have (the args sit directly under the varargs node); adapted to the flatter shape. Output is unchanged for all existing cases including `inject(1, null)`.
- **Tests**: grammar-level parse cases for `g.inject(x)` / `g.V().inject(x)` (`BasicGrammarTest`), and an end-to-end variable-resolution assertion in `GremlinQueryParserTest.shouldParseVariablesInVarargs` mirroring the existing `g.V(x, y, 300)` case.
- **CHANGELOG** entry.

The now-orphaned `genericLiteralVarargs` rule is left in place (harmless; its context class is still generated, keeping `GenericLiteralVisitor`/`DefaultGremlinBaseVisitor` compiling) — happy to remove it if preferred.

## Notes

- No JIRA filed yet — glad to open one if that's preferred for tracking.
- Generated ANTLR parsers (all language targets) are build-time artifacts and will regenerate from the grammar; no committed generated code changed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


danielbodart:
How do you get code coverage to rerun?

spmallette:
hello and thanks for the contribution. i was wondering if you had a specific need/use case for this feature that you could share or if it was just a point of syntax consistency that was driving it. any details you could share with us on that?

danielbodart:
Actually it's a super cool feature (well I think so), just by adding
support for this into the grammar we get federated graph calls.

I am building a clean room Gremlin server where each graph is fully
isolated. To query across graphs we have a federate step that runs a
sub-traversal on a sibling graph. The hard case is a mid-traversal federate:
every traverser on the local side carries a value that has to reach the
remote side, and we very much do not want a round trip per traverser.

So we batch them at the barrier. All the parents cross as one {parentId:
parentValue} map, a single bound parameter, and the remote side runs plain
Gremlin over it:

inject($map).unfold().group().by(Column.keys).by(<per-parent sub-traversal
over the entry value>)

That is the entire transport. group().by(Column.keys) gives us correlation
for free, and it stays correct when two different parents inject equal
values, because they are still two distinct keys. Results come back keyed
by parent, so we scatter them locally. The remote server needs no knowledge
whatsoever that it is being federated to. It is just a Gremlin query with a
bound parameter.

What the user writes is standard as well. An alias before the boundary, a
select after it:

g.V().hasLabel("person").values("name").as("e")
  .call("federate", ["graph": "crew"])
  .V().has("name", select("e"))

The one thing blocking it was inject() taking genericLiteralVarargs, so the
map could not arrive as a parameter at all. Hence the patch. Passing it as
a single bound map rather than per-parent arguments also matters more than
it looks: Our runtime caps a statement at 100 bound parameters, so we have
to make sure we don't hit it

The reason we're keen on it is where it goes next. Once the remote side can
return vertices and edges rather than only values and counts, federation is
complete. You can pull a subgraph from another graph and keep traversing
it, and since it is all standard Gremlin over HTTP, that works across
implementations, not just between two of ours.

Cole-Greer:
Hi @danielbodart, thanks for the submission and the response. Your federated queries use case sounds quite interesting, although there may be a connection there that I am missing. I'm not quite understanding why the map cannot be inlined into the InjectStep as a `genericMapLiteral` instead of passing via a `variable`. I'll share some context below on reasonings behind `literal`/`argument` divide with steps in the first place, but in short I'm not understanding the need for map variables in inject() at this time.

The entire purpose of passing `argument`'s to steps instead of `literal`'s is to take advantage of query caching mechanisms which providers may optionally implement. The intent is that the raw query string (including the variable name) can be used as a key for this cache, and then providers can swap in specific parameter values later upon cache hits. Unless a provider leveraging such a caching strategy, there's generally not any advantage to passing step arguments as a `variable` instead of directly in-lining it as a `literal`.

With this in mind, supporting variables in a step is actually a much more involved process than simply enabling the functionality in the grammar and parser. There needs to be an ability for the variables to be preserved inside the parsed `GraphTraversal`, and those variables must be able to survive optimizing `TraversalStrategy` application. The established pattern for this is to box variables into a `GValue`, which gets passed into a special `GValueHolder` placeholder step. You can reference a class like [GraphStepPlaceholder](https://github.com/apache/tinkerpop/blob/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/step/map/GraphStepPlaceholder.java) as an example.

The preservation of GValues in the parser is controlled by the `VariableResolver` which is configured. We typically use `DirectVariableResolver` for most of our testing, which immediately reduces all variables and replaces them with their literals. In this configuration, there is no meaningful distinction between an argument being passed as a `variable` or a `literal`. If you were to instead configure `DefaultVariableResolver`, you would find that these variables will instead manifest as `GValue` objects in the parsed `GraphTraversal`, which would break the semantics of `InjectStep` without further modifications to follow the `GValueHolder` pattern. As-is, this would result in a `InjectStep<GValue<Map>>` instead of an `InjectStep<Map>` as desired. A GValue should never survive to the point of traversal execution.

The set of steps which are currently permitted to accept variables was carefully curated to target steps which showed the greatest need and upside for this query caching use case. Due to the complexity of the `GValueHolder` pattern, we deliberately withheld such behaviour from steps which we did not find justified the complexity.

danielbodart:
Interesting, so the reason I can't inline them is because the values from the parent traversal can contain whole sub graphs of data that are projected into the sibling graph (so can be pretty big / unbounded) and exactly as you say I can now benefit from query caching. My implementation uses SQLlitre as the backing store and any bindings just become prepare statement bindings 1-2-1 with no extra complexity introduced. So inlining (for my implementation) can actually stop the query running (I lower the whole gremlin traversal into a relational algebra and then compile to SQL) and with unbound inlining of a graph that could easily hit the max statement size of SQLlite, it's not cachable as you already mentioned but it's also a lot slower to parse as the map would need to be parsed by gremlin parser (antlr-ng in my case) rather than as native code inside SQLite.  

But I am a realist and understand none of these are your concerns! I didn't realise the inconsistency was intentional and though this was an easy win for everyone and just wanted to upstream my patch. I can carry on shipping an enhanced gremlin library for the federated use cases. I'll close this but have a few more upstream patches I'd like to see if any of them are interest to you all

